PR #13, opened 2020-05-09
remove node_modules

jasonLaster:


Sawtaytoes:
Can we create a `package.json` off this? It could have `private: true`, but the goal would be to visually see all dependencies rather than relying on `package-lock.json`.

Sawtaytoes:
Another reason we should remove `node_modules/` from the repo:
image

It converts the scripts on different OSes.

PR #26, opened 2020-05-19
Support searching for a file

jasonLaster:
It's not pretty, but it works

Screen Shot 2020-05-18 at 9 20 10 PM

### DIFF
--- a/src/devtools/client/debugger/dist/vendors.js
+++ b/src/devtools/client/debugger/dist/vendors.js
@@ -4722,7 +4722,7 @@ var _react = __webpack_require__(6);
 
 var _react2 = _interopRequireDefault(_react);
 
-var _reactDom = __webpack_require__(112);
+var _reactDom = __webpack_require__(103);
 
 var _reactDom2 = _interopRequireDefault(_reactDom);
 
--- a/src/devtools/client/debugger/src/components/App.js
+++ b/src/devtools/client/debugger/src/components/App.js
@@ -27,7 +27,7 @@ import type { Source } from "../types";
 
 import { KeyShortcuts } from "devtools-modules";
 import Services from "devtools-services";
-const shortcuts = new KeyShortcuts({ window });
+const shortcuts = new KeyShortcuts({ window, target: document });
 
 const { appinfo } = Services;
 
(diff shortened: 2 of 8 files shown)


src/devtools/client/debugger/dist/vendors.js
```diff
@@ -4722,7 +4722,7 @@ var _react = __webpack_require__(6);
 
 var _react2 = _interopRequireDefault(_react);
 
-var _reactDom = __webpack_require__(112);
```
jasonLaster, line 4725: this is no longer needed because, we're transitioning away from vendors.js, but it is nice to have incase we want to revert back to this artifact later.

src/devtools/client/debugger/src/components/App.js
```diff
@@ -27,7 +27,7 @@ import type { Source } from "../types";
 
 import { KeyShortcuts } from "devtools-modules";
 import Services from "devtools-services";
-const shortcuts = new KeyShortcuts({ window });
+const shortcuts = new KeyShortcuts({ window, target: document });
```
jasonLaster, line 30: this fixes KeyShortcuts

src/devtools/client/debugger/src/components/App.js
```diff
@@ -115,6 +115,10 @@ class App extends Component<Props, State> {
     verticalLayoutBreakpoint.addListener(this.onLayoutChange);
     this.setOrientation();
 
+    shortcuts.on(L10N.getStr("symbolSearch.search.key1"), (_, e) =>
+      this.toggleQuickOpenModal(_, e)
+    );
```
jasonLaster, line 120: adds cmd+p

src/devtools/client/debugger/src/components/shared/Modal.js
```diff
@@ -8,7 +8,7 @@ import PropTypes from "prop-types";
 import React from "react";
 import type { Node as ReactNode } from "react";
 import classnames from "classnames";
-import Transition from "react-transition-group/Transition";
+import { Transition } from "react-transition-group";
```
jasonLaster, line 11: This switches us to using the node_module, which has a couple of advantages

- we no longer rely on building a vendors bundle, which will be annoying to maintain
- we now have a versioned module which we can bump if we want improvements

PR #61, opened 2020-06-04
Remove moz.build + mochitest files

jasonLaster:


jasonLaster:
fun fact, this removes ~45 of the files in src/devtools

PR #65, opened 2020-06-04
Loosen some prop-types

jasonLaster:


jasonLaster:
I'm not sure about the execution point one

PR #72, opened 2020-06-05
refactor toolbox

jasonLaster:
- moves gToolbox into the Toolbox component because the toolbox class was always a nightmare
- eventually all of the panels should be part of the same react tree 
- a bunch of the bootstrap logic is still moving around

### DIFF
--- a/src/protocol/thread.js
+++ b/src/protocol/thread.js
@@ -39,10 +39,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // performed on the state at different points in the recording. This layer
 // helps with adapting the devtools to the WRP.
 
-const {
-  sendMessage,
-  addEventListener,
-} = require("./socket");
+const { sendMessage, addEventListener } = require("./socket");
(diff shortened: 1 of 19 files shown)


src/protocol/thread.js
```diff
@@ -1418,6 +1501,27 @@ const ThreadFront = {
     );
     return this.getPreferredLocationRaw(mappedLocation);
   },
+
+  async getDescription() {
```
jasonLaster, line 1505: this is the only thing i changed.

src/protocol/thread.js
```diff
@@ -1354,7 +1437,7 @@ const ThreadFront = {
       pause.instantiate(message.pauseId, message.data);
       if (message.argumentValues) {
         message.argumentValues = message.argumentValues.map(
-          v => new ValueFront(pause, v)
+          (v) => new ValueFront(pause, v)
```
jasonLaster, line 1440: I'll file an issue for adding a prettier config so we keep things consistent and don't add unecessary parens.

jasonLaster:
I disabled `log` i think this should be configurable

PR #97, opened 2020-06-06
Fix inspector loading

jasonLaster:


jasonLaster:
loads the panel, but does not load the markup tree

PR #132, opened 2020-06-16
restyle conditional panel

jasonLaster:


jasonLaster:
the input is now the same color as the conditional panel

PR #167, opened 2020-06-21
Remove node_modules directory

jasonLaster:


jasonLaster:
this is an artifact of our rushed setup

PR #176, opened 2020-06-23
Move files into gecko

jasonLaster:
I'd like to have one place where all of the gecko/mc lives. This will make it easy to separate code that is new and code that is imported from the old world.

What do you think?

by the way, this is not done, i just wanted to share it before i sunk some time into it.


jasonLaster:
Screen Shot 2020-06-22 at 6 38 22 PM

PR #191, opened 2020-07-03
Resume should use selected frame

jasonLaster:
This is close, but i think stepping might be weird

http://localhost:8080/index.html?id=b7a518bb-3556-4d8a-b6aa-eb692b920529

https://webreplay.io/view?id=b7a518bb-3556-4d8a-b6aa-eb692b920529

by the way, i was pausing here an stepping with different frames. when i stepped i saw something weird with `a`

Screen Shot 2020-07-03 at 4 27 10 PM

### DIFF
--- a/src/devtools/client/debugger/src/reducers/pause.js
+++ b/src/devtools/client/debugger/src/reducers/pause.js
@@ -12,7 +12,8 @@
 
 import { prefs } from "../utils/prefs";
 import { getSelectedSourceId } from "./sources";
-import { getSelectedFrame } from "../selectors/pause";
+import { getSelectedFrame, getFramePositions } from "../selectors/pause";
+import { findLast, find } from "lodash";
 
 import type { Action } from "../actions/types";
 import type { State } from "./types";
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/reducers/pause.js
```diff
@@ -685,4 +692,23 @@ export function getPausePreviewLocation(state: State) {
   return state.pause.previewLocation;
 }
 
+export function getResumePoint(state, type) {
+  const thread = getCurrentThread(state);
+  const framePoints = getFramePositions(state, thread)?.positions.map(p => p.point);
```
bhackett1024, line 697: This should await until the positions are loaded so behavior is consistent even if things haven't finished loading yet.

src/devtools/client/debugger/src/reducers/pause.js
```diff
@@ -685,4 +692,23 @@ export function getPausePreviewLocation(state: State) {
...
+  const executionPoint = getThreadExecutionPoint(state, thread);
+
+  if (!framePoints || isTopFrame(state, thread)) {
+    return executionPoint;
+  }
+
+  if (type == "reverseStepOver" || type == "rewind") {
+    return findLast(framePoints, p => +p < +executionPoint);
```
bhackett1024, line 705: What is the rationale behind changing the origin of the step based on the kind of step being performed?  Also, this should use BigInt instead of + to coerce the point correctly.

src/devtools/client/debugger/src/reducers/pause.js
jasonLaster: good call

src/devtools/client/debugger/src/reducers/pause.js
jasonLaster: 👍

PR #196, opened 2020-07-04
Fix showing multiple markup views

gabrielluong:


jasonLaster:
It's okay to show an empty markup view while loading... I'd like to revisit how we do in-panel loading

PR #234, opened 2020-07-11
Add helper for switching between prod and local

jasonLaster:
small helper for jumping between localhost and prod


jasonLaster:
`app.prod()` vs `app.local()`

PR #249, opened 2020-07-13
Setup fullstory

jasonLaster:
Here's a quick walk through of [fullstory](https://www.loom.com/share/9b8daadedfee4fcfae42b3c4c521011e)

And our beta [analytics](https://www.notion.so/User-Analytics-18356ea8ab4247caaaa96d183856af85) goals.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -50,21 +50,13 @@ if (test) {
 // when running automated tests, which we don't want to happen. It would
 // be good if this was less fragile...
 //
-// *** WARNING ***
 
-require("./styles.css");
-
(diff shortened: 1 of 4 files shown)


src/main.js
```diff
@@ -51,20 +38,24 @@ if (test) {
 // be good if this was less fragile...
 //
 // *** WARNING ***
+// During testing, make sure we clear local storage before importing
+// any code that might instantiate preferences from local storage.
```
bhackett1024, line 42: This comment is in the wrong place.

PR #257, opened 2020-07-14
fix async call stack styling

jasonLaster:


jasonLaster:
I looked at async stacks and grouped stacks here

http://localhost:8080/index.html?id=6dbd70cc-8af6-42fe-9519-bda5b6b5f82e
http://localhost:8080/index.html?id=5a1ef7ec-bd74-4980-9cdb-646b515b6cdf

PR #268, opened 2020-07-22
Move close button to right to no longer cover watch expressions

vickimo:


bhackett1024:
Is this change still necessary after https://github.com/WebReplay/webreplay-devtools/pull/269 ?

vickimo:
> Is this change still necessary after #269 ?

Yup, the close button is still misaligned (#269 only contained the one commit).

bhackett1024:
Cool, thanks!

PR #269, opened 2020-07-22
Removing line height css that causes the close button to overflow

vickimo:
Checked to see that the row does not shift upon expansion


princiya:
looks good!

bhackett1024:
Thanks!

PR #275, opened 2020-07-24
[WIP] 223 outline ui scroll fixed

princiya:
**Before**:

Screenshot 2020-07-24 at 14 55 48

**After**:

ui-scroll-fix

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
@@ -32,6 +32,9 @@ import type { Symbols } from "../../reducers/ast";
 import type { SymbolDeclaration, FunctionDeclaration } from "../../workers/parser";
 import type { Source, Context, SourceLocation } from "../../types";
 
+const outlineHeaderHeight = 29;
+const outlineFooterHeight = 25;
+
 type OwnProps = {|
   alphabetizeOutline: boolean,
   onAlphabetizeClick: Function,
@@ -282,16 +285,21 @@ export class Outline extends Component<Props, State> {
     }
 
     return (
-      <ul ref="outlineList" className="outline-list devtools-monospace" dir="ltr">
-        {namedFunctions.map(func => this.renderFunction(func))}
-        {classes.map(klass => this.renderClassFunctions(klass, classFunctions))}
-      </ul>
+      <div
+        className="outline-list-wrapper"
+        style={{ height: `calc(100% - ${outlineHeaderHeight + outlineFooterHeight}px)` }}
+      >
+        <ul ref="outlineList" className="outline-list devtools-monospace" dir="ltr">
+          {namedFunctions.map(func => this.renderFunction(func))}
+          {classes.map(klass => this.renderClassFunctions(klass, classFunctions))}
+        </ul>
+      </div>
     );
   }
 
   renderFooter() {
     return (
-      <div className="outline-footer">
+      <div className="outline-footer" style={{ height: `${outlineFooterHeight}px` }}>
         <button
           onClick={this.props.onAlphabetizeClick}
           className={this.props.alphabetizeOutline ? "active" : ""}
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
```diff
@@ -282,16 +285,21 @@ export class Outline extends Component<Props, State> {
...
+          {classes.map(klass => this.renderClassFunctions(klass, classFunctions))}
+        </ul>
+      </div>
     );
   }
 
   renderFooter() {
     return (
-      <div className="outline-footer">
+      <div className="outline-footer" style={{ height: `${outlineFooterHeight}px` }}>
```
jasonLaster, line 302: why do this in JS instead of CSS? Totally fine, just curious.

src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
```diff
@@ -282,16 +285,21 @@ export class Outline extends Component<Props, State> {
     }
 
     return (
-      <ul ref="outlineList" className="outline-list devtools-monospace" dir="ltr">
-        {namedFunctions.map(func => this.renderFunction(func))}
-        {classes.map(klass => this.renderClassFunctions(klass, classFunctions))}
-      </ul>
+      <div
+        className="outline-list-wrapper"
+        style={{ height: `calc(100% - ${outlineHeaderHeight + outlineFooterHeight}px)` }}
```
jasonLaster, line 290: why do this in JS instead of CSS? Totally fine, just curious.

src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
```diff
@@ -32,6 +32,9 @@ import type { Symbols } from "../../reducers/ast";
 import type { SymbolDeclaration, FunctionDeclaration } from "../../workers/parser";
 import type { Source, Context, SourceLocation } from "../../types";
 
+const outlineHeaderHeight = 29;
+const outlineFooterHeight = 25;
```
jasonLaster, line 36: if it were me, i'd probably use CSS vars, but I don't feel strongly either

princiya: it is inside Outline.js there is OutlineFilter.js. The scroll ui and outline-footer are inside outline.js, so it felt easier to define them here and control the styles through JS which will also answer your previous comment.

jasonLaster: got it. thanks!

jasonLaster:
Just read through the code and it looks like the outline UI has a terrible mix of flexbox and absolute positioning :/ 

Do you think you could update it to *just* be flexbox where the list grows to be the full height, but the input and footer are fixed and always on the top bottom? That should be a lot cleaner...

jasonLaster:
Also, minor note, but the selection seems a bit funky when it is not the widest element...

Screen Shot 2020-07-28 at 9 24 31 AM


happy to file a separate issue for this...

princiya:
> Just read through the code and it looks like the outline UI has a terrible mix of flexbox and absolute positioning :/
> 
> Do you think you could update it to _just_ be flexbox where the list grows to be the full height, but the input and footer are fixed and always on the top bottom? That should be a lot cleaner...

yes, i can change it in this PR to only use flexbox.

princiya:
> Also, minor note, but the selection seems a bit funky when it is not the widest element...
> 
> Screen Shot 2020-07-28 at 9 24 31 AM
> 
> happy to file a separate issue for this...

i did not understand this. if it is related to the outline, I'd rather fix in this PR itself...

jasonLaster:
Sounds good!

here's a new issue
https://github.com/WebReplay/webreplay-devtools/issues/283

jasonLaster:
re: selected ui... not sure...

src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
Garbee: I think this is one area where BEM style naming for classes could alleviate any concerns on scope widening when doing things in CSS.

That's a discussion for another time though. This PR should certainly not be going there.

jasonLaster:
closing this as it has gone stale.

PR #284, opened 2020-07-28
[WIP] Landing page

princiya:
I need inputs on styling this landing page itself. This PR is an initial draft of what I have in mind and to align with what we want to achieve.

Initial landing page:

Screenshot 2020-07-29 at 00 57 09

Error state:

Screenshot 2020-07-29 at 00 56 24

### DIFF
--- a/src/ui/components/App.js
+++ b/src/ui/components/App.js
@@ -37,6 +37,7 @@ import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
 import { connect } from "react-redux";
 import { actions } from "../actions";
 import { selectors } from "../reducers";
+import { messages } from "../utils/messages";
 
 import "styles.css";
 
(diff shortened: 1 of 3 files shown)


src/ui/components/App.js
```diff
@@ -111,22 +132,68 @@ class App extends React.Component {
     return (
       <>
         <div id="header">
-          <div className="logo"></div>
-          <div id="status"></div>
+          <div className="logo" />
+          <div id="status" ref={this.statusRef} />
```
princiya, line 136: I tried to have the error state in the landing page itself as an error message above the input element. But the error message comes from `protocol.socket` :

```
function setStatus(text) {
  document.getElementById("status").innerText = text;
}
```

This function uses `innerText`, and I was not able to get the `innerText` value in `componentDidMount` or `componentDidUpdate`. I will try a bit more, else will revert all unnecessary changes like `ref`, `this.state.recordingInputError`

src/ui/components/App.js
```diff
@@ -45,23 +46,43 @@ function setTheme(theme) {
...
+    const params = new URLSearchParams(document.location.search.substring(1));
+    this.setState({
+      recordingId: params.get("id"),
+    });
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const { theme } = this.props;
+    const { recordingId, recordingInputError } = this.state;
```
princiya, line 70: Trying to have a better error message, else will remove this unnecessary additional state variable `recordingInputError`.

src/ui/components/App.js
```diff
@@ -111,22 +132,68 @@ class App extends React.Component {
...
+            minSize="20%"
+            maxSize="80%"
+            vert={vert}
+            onResizeEnd={num => {}}
+            startPanel={startPanel}
+            endPanelControl={false}
+            endPanel={endPanel}
+          />
+        ) : (
+          <section className="landing-page-wrapper">
```
jasonLaster, line 153: lets move this to its own `renderLocalLandingPage()` function

src/ui/components/App.js
jasonLaster: hmm, feels like we should move this to react :)

src/ui/components/App.js
```diff
@@ -45,23 +46,43 @@ function setTheme(theme) {
...
   };
 
   componentDidMount() {
     const { theme } = this.props;
     setTheme(theme);
+
+    const params = new URLSearchParams(document.location.search.substring(1));
+    this.setState({
+      recordingId: params.get("id"),
+    });
```
jasonLaster, line 65: I wonder if we need state here or if we can just do something with an uncontrolled input... not sure

src/ui/utils/messages.js
```diff
@@ -0,0 +1,4 @@
+export const messages = {
+  noRecording: "Recording ID not specified",
+  errorRecording: "Error: Invalid recording ID",
```
jasonLaster, line 3: I think it's okay to have text in the JSX. We're not translating anytime soon

src/ui/components/App.js
princiya: yes.................... but it was a nice exercise for me to understand to workflow..

princiya:
> Overall love it!
> 
> One interesting point is that it would be nice if this _only_ happened locally. basically a dev-mode thing. I think we have a helper for this somewhere...

got it. yes the helper function is there in bootstrap.js.

src/ui/components/App.js
princiya: yes, this is also part of trying to get that error message within the landing page. i will revert these changes if it is ok to show the error message like it appears on the 2nd attached screenshot.

jasonLaster:
Closing as we now have a reasonable landing page

PR #294, opened 2020-08-02
Timeline refactor

jasonLaster:


jasonLaster:
They'll be several other refactor tickets here, but the gist of it is that the Timeline component is 

1. a legacy gecko component with a ton of cruft
2. doing way too much
3. keeping tons of internal state that should be in redux and available to the Toolbox

PR #327, opened 2020-08-13
remove the context menu items for debugger statements (#193)

### LINKED ISSUE #193: "Ignore debugger" does not work
it's possible to negate a debugger statement by right clicking on the gutter and selecting "never pause here".

This is not implemented yet in WR

hbenl:
Fixes #193.
There are 3 places where such context menu items show up:
- right clicking in the (empty) gutter
- right clicking on a breakpoint icon in the gutter
- right clicking on a breakpoint in the Breakpoints panel

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/menus/breakpoints.js
+++ b/src/devtools/client/debugger/src/components/Editor/menus/breakpoints.js
@@ -172,12 +172,12 @@ export function breakpointItems(
     toggleDisabledBreakpointItem(cx, breakpoint, breakpointActions),
   ];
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks. I don't think i was clear with this bug. The goal is to implement the support in the backend so the debugger statement is actually negated.

Here is a good example:
https://replay.io/view?id=ec471d64-480d-4e1f-8d25-4929bf7ddbcb

@bhackett1024 what do you think needs to be done here?

jasonLaster:
can we close this @hbenl?

PR #354, opened 2020-08-15
don't wait for a panel's initialization before switching to it (fixes #258)

### LINKED ISSUE #258: Clicking inspector doesn't work when it hasn't loaded
On a slow connection, clicking the inspector button doesn't do anything for several seconds.  The inspector should open immediately, even if markup hasn't loaded yet.

hbenl:


jasonLaster:
Looks great

PR #369, opened 2020-08-17
Adds elegant spacing between toolbar buttons

darkwing:
Makes it such that toolbar button item texts are no longer flush against each other.

Just want to make a small contribution to this future legendary project :)


jasonLaster:
Thanks @darkwing !

PR #431, opened 2020-08-21
WIP: setup jest

thymikee:
Early and experimental, trying to run Jest from root.


jasonLaster:
closing as we can now do something simpler now that we're no longer supporting more complex tests

thymikee:
Ugh, turned out I didn't find time to contribute more to it. Sorry about that, but glad you found a simpler solution :)

jasonLaster:
Yep, simpler is often better :)

PR #500, opened 2020-08-25
Save the logged in user's picture and name for use in comments

jaril:
Follow up to #442.

When the user logs in using auth0, this saves their associated picture and name in the user metadata. Now, we can display the user's picture alongside their comments in the right sidebar.

Note that the login doesn't persist after refresh when using Auth0's "Sign up with Google" feature. However, it does persist if you sign up for an Auth0 account and use that instead. 

image

### DIFF
--- a/src/ui/components/Avatar.js
+++ b/src/ui/components/Avatar.js
@@ -1,0 +1,44 @@
+import React, { useState } from "react";
+import { useAuth0 } from "@auth0/auth0-react";
+
+import { connect } from "react-redux";
+import { selectors } from "ui/reducers";
+import { actions } from "ui/actions";
+import { getAvatarColor } from "ui/utils/user";
+
+const Avatar = props => {
+  let { player, isFirstPlayer, updateUser, isLoggedIn, setLoggedIn, setLoggedOut } = props;
+  let auth = useAuth0();
+
+  if (auth.isAuthenticated && isFirstPlayer) {
+    if (!player.name) {
+      // Check if the user has just logged out. If so, update and add the associated
+      // picture and name from the user metadata.
+      updateUser(auth.user);
+    }
+
+    return (
+      <div className={`avatar authenticated first-player`}>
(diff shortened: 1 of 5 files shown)


src/ui/components/Avatar.js
```diff
@@ -0,0 +1,44 @@
...
+const Avatar = props => {
+  let { player, isFirstPlayer, updateUser, isLoggedIn, setLoggedIn, setLoggedOut } = props;
+  let auth = useAuth0();
+
+  if (auth.isAuthenticated && isFirstPlayer) {
+    if (!player.name) {
+      // Check if the user has just logged out. If so, update and add the associated
+      // picture and name from the user metadata.
+      updateUser(auth.user);
+    }
```
jasonLaster, line 18: Hmm, i'm wondering if this should be in a `useEffect`. Not too sure, i don't know hooks :)


```js
useEffect(() => updateUser(auth.user), [auth, isFirstPlayer])
```

src/ui/components/RightSidebar/Comment.js
```diff
@@ -81,16 +81,15 @@ class Comment extends React.Component {
 
   renderAvatar() {
     const { comment } = this.props;
-
-    // Comments that have been made prior to adding the users feature don't
-    // have an associated user. We just give those comments a grey circle.
-    // This should be removed eventually.
+    const { picture, name } = comment.user;
```
jasonLaster, line 84: Is it possible to have a comment without an avatar?

```
comment?.user.avatarID`
```

makes it look like it is, but who knows

jaril: In a way, yes. 

Old comments from before we enabled the users feature don't have a user property. Because of that, there's no corresponding avatarID for that comment. We use the avatarID to generate the avatar color, so in a way the old comments don't have an avatar.

I'll rewrite this as `comment.user?.avatarID` since `comment.user` can be undefined.

PR #508, opened 2020-08-26
Remove unused components in the Inspector

gabrielluong:
Cleaning up some of the unused components in the Inspector. We can always undo some of these changes when the time comes.


jasonLaster:
Feeling lighter!

Issue #486, opened 2020-08-24
[Comments] Can't add a comment close to a comment near the end of the timeline


jaril:
Adding a comment to a point near the end of the timeline (>95% progress) prevents a comment from being added to a point that's within ~5% of it on both sides.

https://replay.io/view?id=69eea65a-2d2a-44b0-941c-315ddc768cfe

PR #533, opened 2020-08-28
Fixed the interactions between multiple comments in the timeline

jaril:
- Nearby timeline comments to the hovered time are now styled as if they were hovered on
- The create comment button is hidden after creating a comment to show the new comment marker underneath
- Newly created comments in the timeline with no content will now close properly after clicking the "Cancel" button

image


jasonLaster:
Two things I'm noticing, which could be follow ups are

1. styling of comments that are close together. Perhaps we need a fill?
2. adding events in certain areas does not work... https://replay.io/view?id=97e3f2ca-937a-4bcb-a11b-f1cf2a96bb34

jaril:
Fixes #486

PR #534, opened 2020-08-29
Add scroll zooming in the timeline

jaril:
image

### DIFF
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -17,6 +17,7 @@ import PropTypes from "prop-types";
 import { SVG } from "image/svg";
 import { log } from "protocol/socket";
 import FullStory from "ui/utils/fullstory";
+import ScrollContainer from "ui/components/Timeline/ScrollContainer";
 
 const {
   screenshotCache,
@@ -67,7 +68,6 @@ function CommandButton({ img, className, onClick, active }) {
     previous: SVG.NextButton,
     pause: SVG.ReplayPause,
     play: SVG.ReplayResume,
-    zoomout: SVG.ZoomOut,
     comment: SVG.Comment,
   };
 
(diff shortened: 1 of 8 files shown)


src/ui/components/Timeline/index.js
```diff
@@ -761,6 +766,49 @@ export class Timeline extends Component {
...
+    let mid = zoomRegion.mid || length / 2;
+    let scale = zoomRegion.scale;
+
+    const newLength = (length * scale) / newScale;
+    const newMid = mid + (hover - mid) * ((newScale - scale) / newScale);
+    const newStart = newMid - newLength / 2;
+    const newEnd = newMid + newLength / 2;
+
+    return { newLength, newMid, newStart, newEnd };
+  }
```
jasonLaster, line 799: couple of high level thoughts:

1. it'd be nice if this function returned the zoomRegion type, basically drop the `new` :) 
2. it looks like we don't need to return length
3. it'd be nice if this helper function could live in a util. Given that, i'd probably just pass in the props as well  `calculateNewSelection({hover, newScale, recordingDuration, zoomRegion})`

src/ui/components/Timeline/index.js
```diff
@@ -761,6 +766,49 @@ export class Timeline extends Component {
...
+  renderScrollContainer() {
+    return div(
+      { className: "scroll-container", onScroll: this.onScrollContainerScroll },
+      div(
+        { className: "scroll-content" },
+        div({ className: "start" }, "start"),
+        div({ className: "end" }, "end")
+      )
+    );
+  }
```
jasonLaster, line 810: I wonder if the Scroll Container would look good as its own component

```js
function calculateNewSelection() {
}

function ScrollContainer({}) {
  function onScrollContainerScroll() {}
}

export default connect(
  () => {},
  () => {}
)(ScrollContainer)
```

PR #541, opened 2020-08-30
(WIP) Add title

jasonLaster:
Screen Shot 2020-08-29 at 6 21 11 PM

### DIFF
--- a/src/ui/actions/metadata.js
+++ b/src/ui/actions/metadata.js
@@ -7,12 +7,15 @@ import FullStory from "ui/utils/fullstory";
 // Metadata key used to store comments.
 const CommentsMetadata = "devtools-comments";
 const UsersMetadata = "devtools-users";
+const TitleMetadata = "devtools-title";
+
 let heartbeatPing = Date.now();
 const refreshRate = 10000;
 
 export function setupMetadata(_, store) {
   ThreadFront.watchMetadata(CommentsMetadata, args => store.dispatch(onCommentsUpdate(args)));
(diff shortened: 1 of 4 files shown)


src/ui/actions/metadata.js
```diff
@@ -181,3 +184,29 @@ export function updateUser(authUser = {}) {
...
 }
+
+function onTitleUpdate(title) {
+  return async ({ dispatch, getState }) => {
+    if (title === selectors.getTitle(getState())) {
+      return;
+    }
+
+    if (!title) {
+      await new Promise(r => setTimeout(r, 5000));
```
jasonLaster, line 195: this is a little gross, but it makes it work reliably :)

src/ui/actions/metadata.js
```diff
@@ -181,3 +184,29 @@ export function updateUser(authUser = {}) {
...
+      const titleVal = await ThreadFront.evaluate(0, null, "window.location.host");
```
jasonLaster, line 196: We'll be able to replace this in the future when we share data with the recordings app.

src/ui/components/Header.css
```diff
@@ -15,9 +15,16 @@
   background-size: contain;
 }
 
+.recording-title {
+  flex-grow: 1;
+  text-align: center;
```
jasonLaster, line 20: This approach doesn't actually center it, but we can definitely iterate

jasonLaster:
This will need editing soon too

PR #547, opened 2020-08-30
Make the React-based Rules view functional for Replay

gabrielluong:
This patch set makes the React-based Rules view functional for webreplay. To test, you will need to comment out the pref check:

```
diff --git a/src/devtools/client/inspector/inspector.js b/src/devtools/client/inspector/inspector.js
index 7058645e..0da1a676 100644
--- a/src/devtools/client/inspector/inspector.js
+++ b/src/devtools/client/inspector/inspector.js
@@ -920,12 +920,12 @@ Inspector.prototype = {
       */
     ];

-    if (Services.prefs.getBoolPref("devtools.inspector.new-rulesview.enabled")) {
-      sidebarPanels.push({
-        id: "newruleview",
-        title: INSPECTOR_L10N.getStr("inspector.sidebar.ruleViewTitle"),
-      });
-    }
+    // if (Services.prefs.getBoolPref("devtools.inspector.new-rulesview.enabled")) {
+    sidebarPanels.push({
+      id: "newruleview",
+      title: INSPECTOR_L10N.getStr("inspector.sidebar.ruleViewTitle"),
+    });
+    // }
```

### DIFF
--- a/src/devtools/client/inspector/rules/models/element-style.js
+++ b/src/devtools/client/inspector/rules/models/element-style.js
@@ -86,11 +86,11 @@ class ElementStyle {
 
     this.onRefresh = this.onRefresh.bind(this);
 
-    if (this.ruleView.isNewRulesView) {
-      this.pageStyle.on("stylesheet-updated", this.onRefresh);
-      this.ruleView.inspector.styleChangeTracker.on("style-changed", this.onRefresh);
-      this.ruleView.selection.on("pseudoclass", this.onRefresh);
-    }
+    // if (this.ruleView.isNewRulesView) {
+    //   this.pageStyle.on("stylesheet-updated", this.onRefresh);
+    //   this.ruleView.inspector.styleChangeTracker.on("style-changed", this.onRefresh);
+    //   this.ruleView.selection.on("pseudoclass", this.onRefresh);
+    // }
   }
 
   get unusedCssEnabled() {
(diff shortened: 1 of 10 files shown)


src/devtools/client/inspector/rules/models/element-style.js
```diff
@@ -86,11 +86,11 @@ class ElementStyle {
 
     this.onRefresh = this.onRefresh.bind(this);
 
-    if (this.ruleView.isNewRulesView) {
-      this.pageStyle.on("stylesheet-updated", this.onRefresh);
-      this.ruleView.inspector.styleChangeTracker.on("style-changed", this.onRefresh);
-      this.ruleView.selection.on("pseudoclass", this.onRefresh);
-    }
+    // if (this.ruleView.isNewRulesView) {
```
jasonLaster, line 89: should we keep this and pref it off via a new `edit` flag or just remove it in another pass and circle back?

src/devtools/client/inspector/rules/new-rules.js
```diff
@@ -565,14 +560,14 @@ class RulesView {
...
-      this.store.dispatch(updateAddRuleEnabled(false));
-      this.store.dispatch(updateClasses([]));
+      // this.store.dispatch(disableAllPseudoClasses());
+      // this.store.dispatch(updateAddRuleEnabled(false));
+      // this.store.dispatch(updateClasses([]));
       this.store.dispatch(updateRules([]));
       return;
     }
 
-    this.pageStyle = element.inspectorFront.pageStyle;
```
jasonLaster, line 575: I think this might have been causing some issues...

src/devtools/client/inspector/rules/models/element-style.js
gabrielluong: isNewRulesView isn't a pref. It's a parameter that only appears in new-rules.js to trigger new codepaths that was being implemented for the React-based Rules view. The goal was to move some of the business logic to the model where it belonged. 

In Replay, we don't have this.pageStyle anymore, so we will remove that probably in the future. The StyleChangeTracker and Pseudoclass might be important when we implement those features in Replay.

src/devtools/client/inspector/rules/new-rules.js
gabrielluong: We don't have this.pageStyle anymore in Replay. This would've represented the StyleRuleActor, which provides methods for editing CSS, which we haven't implemented in Replay. In the existing Rules view, this is also undefined. We will want to clean up this.pageStyle in a followup.

PR #551, opened 2020-08-31
Disable SplitBox resizing from the console

### LINKED ISSUE #381: [Console] The split console resizes from the console
Right now, it's possible to resize the height of the split console from the console panel, regardless of whether or not the split console is open.

Users should only be able to resize the split console if they have the split console open in the Debugger/Inspector (any panel that's not the Console).

jaril:
Fixes #381 .

### DIFF
--- a/src/devtools/client/shared/components/splitter/SplitBox.css
+++ b/src/devtools/client/shared/components/splitter/SplitBox.css
@@ -91,3 +91,10 @@
 .split-box.dragging > .uncontrolled {
   pointer-events: none;
 }
+
+/* This disables splitter bar dragging when the user is 
+in the console. Dragging/resizing should only work in the debugger
+or the inspector where the split console can be displayed.  */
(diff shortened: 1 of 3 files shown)


src/devtools/client/shared/components/splitter/SplitBox.css
```diff
@@ -91,3 +91,10 @@
 .split-box.dragging > .uncontrolled {
   pointer-events: none;
 }
+
+/* This disables splitter bar dragging when the user is 
+in the console. Dragging/resizing should only work in the debugger
+or the inspector where the split console can be displayed.  */
+#toolbox.console .split-box.horz > .splitter {
+  pointer-events: none;
+}
```
jasonLaster, line 100: I wonder if we need to add it to all 3 places if the CSS is *just* going to be redundant.

I think this is fine to do here, but we should probably remove the redundant CSS eventually...

PR #552, opened 2020-08-31
Display multiline comments in the console and the timeline

### LINKED ISSUE #539: [Commnets] Can not press enter while editing
I believe we recently regressed the ability to press enter and create a new line while editing a comment.

I think the issue might be in the markup tree's keyshortcut handling.

jaril:
Fixes #539
image

### DIFF
--- a/src/ui/components/Comments/CommentEditor.js
+++ b/src/ui/components/Comments/CommentEditor.js
@@ -24,7 +24,7 @@ class CommentEditor extends React.Component {
   onKeyDown = e => {
     if (e.key == "Escape") {
       this.stopEditingComment();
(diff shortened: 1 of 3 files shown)


src/ui/components/Comments/CommentEditor.js
```diff
@@ -24,8 +24,6 @@ class CommentEditor extends React.Component {
   onKeyDown = e => {
     if (e.key == "Escape") {
       this.stopEditingComment();
-    } else if (e.key == "Enter") {
```
jasonLaster, line 27: lets change this to cmd+enter which is nice :)

PR #554, opened 2020-08-31
Use class syntax

jasonLaster:


src/devtools/client/inspector/inspector.js
```diff
@@ -218,66 +197,66 @@ Inspector.prototype = {
...
 
     return this._is3PaneModeEnabled;
-  },
+  }
 
   set is3PaneModeEnabled(value) {
     this._is3PaneModeEnabled = value;
-    Services.prefs.setBoolPref(THREE_PANE_ENABLED_PREF, this._is3PaneModeEnabled);
-  },
+    prefs.threePane = this._is3PaneModeEnabled;
```
gabrielluong, line 228: s/threePane/threePaneEnabled would be a bit more meaningful.

PR #555, opened 2020-09-01
Removed close split console button when in the console panel

### LINKED ISSUE #370: [Console] Split console button shows up in Console panel
The Close Split Console button shows up on the console panel, even when the user is not using split console.

image

jaril:
Fixes #370

### DIFF
--- a/src/devtools/client/webconsole/webconsole-ui.js
+++ b/src/devtools/client/webconsole/webconsole-ui.js
@@ -361,8 +361,8 @@ class WebConsoleUI {
     }
   }
 
-  _onChangeSplitConsoleState() {
-    this.wrapper.dispatchSplitConsoleCloseButtonToggle();
+  _onChangeSplitConsoleState(selectedPanel) {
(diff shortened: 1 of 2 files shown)


src/devtools/client/webconsole/webconsole-ui.js
```diff
@@ -361,8 +361,8 @@ class WebConsoleUI {
     }
   }
 
-  _onChangeSplitConsoleState() {
-    this.wrapper.dispatchSplitConsoleCloseButtonToggle();
+  _onChangeSplitConsoleState(selectedPanel) {
```
jasonLaster, line 364: `selectedPanel ` => `shouldDisplayButton`

jaril: `selectedPanel` is correct here since this is where it's being called:

https://github.com/RecordReplay/devtools/blob/f030aa2561aea1e082f2167a759dc51c4f5e8bf2/src/ui/components/Toolbox.js#L111

So it's value is the panel e.g. `debugger`, `console`, etc

PR #559, opened 2020-09-01
Hide comment markers outside of the zoom region

### LINKED ISSUE #528: [Timeline] Comments outside of the zoom selection are still displayed
Zooming in the timeline still displays the comments outside of its zoom selection. Instead, they're just pushed all the way to the timeline's edge.

STR here: https://replay.io/view?id=0a2a99af-5255-4dc3-a6e5-1999a9e82c06

jaril:
Fixes #528.

### DIFF
--- a/src/ui/components/Comments/CommentMarker.js
+++ b/src/ui/components/Comments/CommentMarker.js
@@ -52,15 +52,15 @@ class CommentMarker extends React.Component {
   }
 
   renderCreateCommentButton() {
-    const { createComment, currentTime } = this.props;
+    const { createComment, currentTime, zoomRegion } = this.props;
 
-    if (this.getCommentAtTime(currentTime)) {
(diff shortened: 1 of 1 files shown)


src/ui/components/Comments/CommentMarker.js
```diff
@@ -52,15 +52,17 @@ class CommentMarker extends React.Component {
...
     if (this.getCommentAtTime(currentTime)) {
       return null;
     }
 
     return (
       <button
-        className={classnames("create-comment")}
+        className={classnames("create-comment", {
+          overflowed: currentTime > zoomRegion.endTime,
+        })}
```
jasonLaster, line 65: why not return null above?

the benefit might be that it's fewer divs in the dom?

jaril: Good catch, thanks. Fixed!

PR #596, opened 2020-09-06
WIP: header uploading state

Rennzie:
This adds the `uploadedMB` and `totalMB` to the app reducer on an `uploading` property. It's then passed to the `Header` component as props in `App`. 

I'm not sure how to test the uploading so I don't know if the implementation works yet. Any pointers would be awesome. 

It's not clear how to tell when the uploading is completed if there is no `totalMB`. Does there need to be an `onUploadingCompleted` method added in socket.js to set the loading back to null or something else in this case?

Secondly, I've left the status to be updated via `element.innerText` however would it be worth setting that into the app reducer too? This should give us a little more control over it, and allow for conditional styling when the status is an error or uploading etc.


Rennzie:
Some overlap with the [Landing Page PR](https://github.com/RecordReplay/devtools/pull/284) regarding moving the status into react. What about having a `status` property in the app reducer that is either null, or is an object with a type like: 

```
{
  type: 'error' | 'message' | 'uploading';
  message?: string;
}
```

src/ui/components/Header.js
```diff
@@ -42,13 +43,20 @@ class Header extends React.Component {
...
     const isLoaded = loading == 100;
+    const upload = uploading?.totalMB
+      ? `${(uploading?.uploadedMB / uploading?.totalMB) * 100}%`
+      : `${uploading?.uploadedMB}MB`;
+
     return (
       <>
         <div id="header">
           <div className="logo"></div>
+          {/* NOTE: Should we use the app state to set this status rather than relying on `element.innerText`? */}
```
jasonLaster, line 56: lets do the same thing with status as well

src/ui/components/Header.js
```diff
@@ -42,13 +43,20 @@ class Header extends React.Component {
...
           <div id="status"></div>
+          {/* NOTE: This needs a way to determine when the upload is completed: specifically when there is no `totalMB` */}
```
jasonLaster, line 58: when would we not have a total upload?

Perhaps this starts as uploaded and only changes if we learn about an uploading status?

jasonLaster (commented):
Thanks!

Looking great

src/ui/components/Header.js
Rennzie: Thoughts ont the interface for the status type above?

src/ui/components/Header.js
Rennzie: In `main.js` in the `onUploadedData` function `lengthMB` can be undefined. I'm not the circumstances under which that could happen.

In that case it not possible to know when the download is completed from the information provided. The has two downstream effects, 

1. Knowing when to remove the progress indicator based on upload completion. 
2. The antD progress indicator requires a percentage to now how much of the circle to colour in. If there is no `lengthMB` then we would need to use a spinner or some other component. I think the material-ui option could work but that requires adding their theme to the app and comes with a lot of bloat for a single component.

jasonLaster:
Hi @Rennzie i'm closing this as we now how this working in redux with a simple UI

Screen Shot 2020-10-01 at 4 24 57 PM

Issue #323, opened 2020-08-13
Fix unique key issue with ObjectInspectorItem


jasonLaster:
Recording
https://replay.io/view?id=7c0c16e9-52b4-4d5f-a4ae-166f190e89ee

STR:
1. go to http://localhost:8080/index.html?id=320d1de6-81df-4505-8e9c-44494b6c8cd0

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ObjectInspectorItem`. See https://fb.me/react-warning-keys for more information.
    in span (created by ObjectInspectorItem)
    in ObjectInspectorItem (created by TreeNode)
    in div (created by TreeNode)
    in TreeNode (created by Tree)
    in div (created by Tree)
    in Tree (created by ObjectInspector)
    in ObjectInspector (created by Connect(ObjectInspector))
    in Connect(ObjectInspector)
    in Unknown (created by MessageContainer)
    in span (created by Message)
    in span (created by Message)
    in span (created by Message)
    in div (created by Message)
    in Message (created by MessageContainer)
    in MessageContainer (created by ConsoleOutput)
    in div (created by ConsoleOutput)
    in ConsoleOutput
    in VisibilityHandler
    in Unknown (created by Connect(Component))
    in Connect(Component) (created by App)
    in div (created by App)
    in div (created by App)
    in App
    in VisibilityHandler
    in Unknown (created by Connect(Component))
    in Connect(Component)
    in Provider
```

jasonLaster:
seeing some undefineds

<img width="738" alt="Screen Shot 2020-09-04 at 5 35 30 PM" src="https://user-images.githubusercontent.com/254562/92293028-461d6e00-eed5-11ea-9f08-4636efe67819.png">

PR #597, opened 2020-09-06
Fix React warning about unique key props

jasonLaster:
Fixes (#323)


jasonLaster:
Here's a recording with a comment of propRep, swapping `props.key` for `name` should work

https://replay.io/view?id=9461be71-6133-4526-a692-6e978e71208c

PR #598, opened 2020-09-07
Add tabIndex to messages so they are tab navigable

ryanjduffy:
Adds `tabIndex` (because the `<a>` lacks an `href` which would make them focusable by default) and replaces `onClick` handler with `onFocus` handler to make message points tab-navigable.

I looked at adding the same to comments so that users could stop at those as well but they're rendered in a different subtree and so won't as easily participate in natural tab order.

### DIFF
--- a/src/ui/components/Timeline/Message.js
+++ b/src/ui/components/Timeline/Message.js
@@ -85,6 +85,7 @@ class Message extends React.Component {
 
     return (
       <a
+        tabIndex={0}
         className={classnames("message", {
           future: isFuture,
           highlighted: isHighlighted,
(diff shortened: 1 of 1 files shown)


jasonLaster:
Thanks!

I'm actually not sure how to reproduce this :) Could you make a recording?

ryanjduffy:
Discovered that it doesn't work out of the box on Macs with Firefox because links aren't tab-navigable without some additional configuration. See https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox/11713537.

With that enabled and a bit more code to focus the node on click produces the following.

https://replay.io/view?id=9d751a43-2d73-4028-a217-b3f343931fda

ryanjduffy:
Appears the extra `onClick` code isn't required once those settings are enabled actually.

ryanjduffy:
While taking advantage of built-in tab navigation is nice, it might be better to define some global key handlers to navigate the timeline so you can ensure consistency across platforms and not require every entry be sequential in the DOM.

jasonLaster:
Thanks @ryanjduffy!

PR #605, opened 2020-09-07
Hide toolbox while loading

### LINKED ISSUE #600: Improve loading UX
1. we should hide the devtools while loading
2. we should show the session died error more prominently
3. we should show a nice animation graphic and useful tip while loading

jaril:
Fixes #600. This hides the toolbox until the recording is fully loaded. 

This patch also delays hiding the loading screen. That means the user will always see the loading bar get to 100% before it disappears. Before, the loading bar would not render the 100% loading state before disappearing. This is less noticeable if it went from 99% - 100%, but can be jarring if it goes from 4% - 100%. [Loom demo](https://www.loom.com/share/2020b870f7d54644815582e48bc28e98)

### DIFF
--- a/src/ui/components/App.js
+++ b/src/ui/components/App.js
@@ -25,8 +25,9 @@ class App extends React.Component {
   };
 
   componentDidMount() {
-    const { theme } = this.props;
+    const { theme, initialize } = this.props;
 
+    initialize();
     setTheme(theme);
   }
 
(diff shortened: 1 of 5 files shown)


src/ui/components/App.js
```diff
@@ -22,19 +23,27 @@ class App extends React.Component {
...
+    const { theme, loading } = this.props;
     if (theme !== prevProps.theme) {
       setTheme(theme);
     }
+
+    // We want the loading bar to show the 100% progress before closing it. This
+    // has the effect of hiding the loading bar 1 second after loading is at 100%.
+    if (loading === 100 && this.state.displayLoadingScreen) {
+      setTimeout(() => this.setState({ displayLoadingScreen: false }), 1000);
+    }
```
jasonLaster, line 46: i don't think we need this :P

src/ui/components/App.js
```diff
@@ -51,17 +60,33 @@ class App extends React.Component {
...
   }
 
+  renderLoadingScreen() {
+    const { loading } = this.props;
+
+    return (
+      <>
+        <Header />
+        <LoadingScreen loading={loading} />
+      </>
```
jasonLaster, line 70: lets inline this

src/main.js
```diff
@@ -131,7 +135,7 @@ setTimeout(async () => {
     await new Promise(resolve => setTimeout(resolve, 100));
   }
 
-  store = await bootstrapApp({ initialize }, { recordingId });
+  store = await bootstrapApp({ initialize, initializeToolbox }, { recordingId });
```
jasonLaster, line 138: i dont think we need a callback here

jasonLaster (approved):

PR #630, opened 2020-09-13
Improve test runner

jasonLaster:
1. stop using custom devtools server
2. extract createTestScript

### After this change

* we will need to have the dev server running `npm start`
* We'll no longer need to do a webpack build every time we change the code 🏃 

This should save a lot of time when we're fixing tests

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,16 +11,23 @@ module.exports = {
     publicPath: "dist",
   },
   devServer: {
+    before: app => {
+      app.get("/view", (req, res) => {
+        res.sendFile("index.html", { root: "." });
+      });
+
+      app.get("/test", (req, res) => {
+        const testFile = req.url.substring(6);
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -11,16 +11,23 @@ module.exports = {
     publicPath: "dist",
   },
   devServer: {
+    before: app => {
+      app.get("/view", (req, res) => {
+        res.sendFile("index.html", { root: "." });
+      });
+
+      app.get("/test", (req, res) => {
```
jasonLaster, line 19: out of curiosity, where does this fetch come from?

bhackett1024: When running tests, the thread front dynamically creates a script element whose source will be fetched via this url.  See initializeToolbox in thread.js

bhackett1024:
This looks good, feel free to merge when tests pass.

PR #640, opened 2020-09-14
Improve test runner

jaril:


test/run.js
```diff
@@ -1,44 +1,48 @@
...
-        process.exit(0);
-    }
+  const argv = minimist(process.argv.slice(2));
+
+  if (argv._.length == 0 && Object.keys(argv).length == 1) {
+    const usage = `
+      Usage: run.js arguments
+      Arguments:
+        --count N: Run tests N times
+        --pattern PAT: Only run tests matching PAT
```
jasonLaster, line 33: lets add these fields `--record-viewer`, `--record-examples`, `--record-all`

test/run.js
```diff
@@ -1,44 +1,48 @@
...
-      case "--pattern":
-        patterns.push(next);
-        break;
-      default:
-        console.log(usage);
-        process.exit(0);
-    }
+  const argv = minimist(process.argv.slice(2));
+
+  if (argv._.length == 0 && Object.keys(argv).length == 1) {
```
jasonLaster, line 28: It'd be nice if this is shown for `-h` and `--help` as well

test/run.js
```diff
@@ -1,44 +1,48 @@
 /* Copyright 2020 Record Replay Inc. */
 
 // Harness for end-to-end tests. Run this from the devtools root directory.
-
 const fs = require("fs");
 const http = require("http");
 const os = require("os");
 const { spawnSync, spawn } = require("child_process");
 const { defer } = require("../src/protocol/utils");
+const url = require("url");
```
jasonLaster, line 9: We should update  `main.js` so that run is called with `run --record-all`
https://github.com/RecordReplay/devtools/blob/master/.github/actions/test/main.js#L34

jasonLaster (commented):
Couple small tweaks, but overall works great and brings my average run down the ~11sec per test as opposed to ~20-30sec

PR #646, opened 2020-09-15
Remove project root

jasonLaster:


src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
```diff
@@ -188,66 +161,3 @@
...
-
-.source-outline-panel.has-root > div {
-  height: 100%;
-}
-
-.source-outline-panel.has-root .thread-header {
-  margin-top: 4px;
-}
-
-.sources-list .img.blackBox {
```
jaril, line 200: Do we not have the blackbox feature anymore?

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -101,32 +101,7 @@ class SourceTreeItem extends Component {
       }
     }
 
-    if (isDirectory(item)) {
-      this.addCollapseExpandAllOptions(menuOptions, item);
```
jaril, line 105: Don't we still need the collapse all/expand all context menu item?

If not we can take the method out as well:

https://github.com/RecordReplay/devtools/blob/74d3a8d8a2f0fab17ed37d36720c7efbcee3cd50/src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js#L148-L164

src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
jasonLaster: good call we should keep this

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
jasonLaster: oops. good catch

jaril:
It looked like the permalink was being added to the diff 😬 Might have been a github bug.

Just needs this change: https://gist.github.com/jaril/f7869af7e591db719bc8b9ccbea12b06

PR #648, opened 2020-09-15
Don't display an evaluation error's stack trace

### LINKED ISSUE #238: [console] evaluations with a type error should not have a stack
It's nice to have an error stack when an exception is thrown in the code. Console evaluations should not have one... 

notice the top frame is `debugger eval`

Screen Shot 2020-07-12 at 7 32 59 AM

jaril:
Fixes #238


jasonLaster:
Yes!

PR #649, opened 2020-09-15
Pass in a replay-all flag when tests are running as a Github Action

jaril:


jasonLaster (commented):
This feels a bit off 

* where's runTestRunner
* can we remove the wrapping function?

here's a quick sketch (https://gist.github.com/2d0571e1239b2c654736b04001ab2b20)

jaril:
Continued in #651

PR #651, opened 2020-09-15
Fix test runner so that it runs correctly in github actions

jaril:


jasonLaster (approved):

jaril:
https://github.com/RecordReplay/devtools/actions/runs/255943504
https://github.com/RecordReplay/devtools/actions/runs/255943503

PR #652, opened 2020-09-15
Add apolloClient

jasonLaster:
This deprecates #650

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "typecheck": "tsc --watch"
   },
   "dependencies": {
+    "@apollo/client": "^3.1.5",
     "@auth0/auth0-react": "^1.0.0",
     "@babel/core": "^7.11.0",
     "@babel/plugin-proposal-class-properties": "^7.8.3",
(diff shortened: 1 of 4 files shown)


jasonLaster:
doing this mostly for git reasons.

jasonLaster:
it looks like httpLink is not being created

```
diff --git a/src/ui/utils/apolloClient.js b/src/ui/utils/apolloClient.js
index 6d7ba5d8..3ea4ff64 100644
--- a/src/ui/utils/apolloClient.js
+++ b/src/ui/utils/apolloClient.js
@@ -2,10 +2,15 @@ import { gql, ApolloClient, InMemoryCache } from "@apollo/client";
 import { HttpLink } from "apollo-link-http";

 export const createApolloClient = async auth0Client => {
+  console.log(`createApolloClient`);
+  window.queryApollo = query =>
+    client.query({ query: gql(query) }).then(result => console.log(result));
+
   if (auth0Client.isLoading === true) {
     return;
   }
   const createHttpLink = async headers => {
+    console.log(`>>> fails to create http link`);
     const hasuraToken = await auth0Client.getAccessTokenSilently({
       audience: "hasura-api",
     });
@@ -26,8 +31,6 @@ export const createApolloClient = async auth0Client => {
     link,
     cache: new InMemo
```

jaril:
This diff gets this working locally for both first-party and third-party logins:

https://gist.github.com/jaril/fbd1ac3a59f7c04677acac23e37d7dfa

PR #669, opened 2020-09-17
improve typescript setup

hbenl:
- automatically start the typechecking watcher when the webpack-dev-server is started in VS Code
- remove 2 sample typescript files (that contain typechecking errors in strict mode)


src/devtools/client/debugger/src/workers/parser/tests/fixtures/scopes/ts-sample.ts
```diff
@@ -1,41 +0,0 @@
-
```
jasonLaster, line 1: i think these files ere important for tests.

could we exclude them?

jasonLaster (approved):
👍 with one comment

src/devtools/client/debugger/src/workers/parser/tests/fixtures/scopes/ts-sample.ts
hbenl: What tests? Our own tests do not need them (AFAIK). And I don't want to enable the typecheck watcher as long as these files are there because they contain errors and that would train people to ignore errors from the typecheck watcher.

jasonLaster: These fixtures are for the unit tests  in the parser worker.

If it's hard to exclude the files from the watcher, then remove the fixtures and we'll add them later

hbenl: OK, I've excluded them instead

Issue #581, opened 2020-09-03
[Protocol] optimize getMappedLocation


jasonLaster:
There are two ways of getting mappedLocations `getPreferredMappedLocation` and `getPreferredLocation`

* `getPreferredMappedLocation` queries the server
* `getPreferredLocation` uses existing data

We should be able to switch `getPreferredMappedLocation` in logpoint.js to `getPreferredLocation`, which will reduce a lot of network traffic.

In general, we should do a better job of avoiding network calls, by caching mapped locations fetched from the server and first reading from the new location cache and scripts cache.

```js
diff --git a/src/protocol/thread.js b/src/protocol/thread.js
index d7493bdb..f3bb1fd7 100644
--- a/src/protocol/thread.js
+++ b/src/protocol/thread.js
@@ -1153,6 +1153,9 @@ const ThreadFront = {
   // Map scriptId to info about the script.
   scripts: new Map(),

+  // map scriptId to locations array
+  mappedLocations: new Map(),
+
   // Resolve hooks for promises waiting on a script ID to be known.
   scriptWaiters: new ArrayMap(),

@@ -1859,11 +1862,20 @@ const ThreadFront = {

   // Get the location to use for a generated location without alternatives.
   async getPreferredMappedLocation(location) {
+    // check to see if the location is available by
+    let existingMappedLocation = await this.getPreferredLocation(location);
+    if (existingMappedLocation) {
+      return existingMappedLocation;
+    }
+
     const { mappedLocation } = await sendMessage(
       "Debugger.getMappedLocation",
       { location },
       this.sessionId
     );
+
+    // cache the results s
+
     return this.getPreferredLocation(mappedLocation);
   },
```

hbenl:
We can't switch to `getPreferredLocation` in logpoint.js because (in my experiments at least) it always gets an array with only one location (for which the server will return an array with several locations).
I've added a cache for `MappedLocation`s in #680.

PR #680, opened 2020-09-18
add MappedLocationCache (#581)

hbenl:


src/protocol/mapped-location-cache.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    this.cache.set(cacheKey, mappedLocation);
+    return mappedLocation;
+  }
+
+  /**
+   * Encode a location as a string for use as a cache key
+   */
+  private encodeLocation(location: Location): string {
+    return `${location.scriptId}|${location.line}|${location.column}`;
```
jasonLaster, line 52: I wonder if this could just be `JSON.stringify`

src/protocol/mapped-location-cache.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface Location {
+  scriptId: string;
+  line: number;
+  column: number;
+}
+
+type MappedLocation = Location[];
+
+export class MappedLocationCache {
```
jasonLaster, line 12: This looks great. I wonder if down the road, we would want to make this a more generic feature of the protocol layer, where you could opt to make certain messages cacheable...

jasonLaster (commented):
Looks good to me.

CC @bhackett1024

src/protocol/mapped-location-cache.ts
hbenl: The critical point here is making sure that the string created for 2 locations is the same if `scriptId`, `line` and `column` are the same - I wouldn't trust `JSON.stringify` here: the order of the properties could change and a location could contain additional properties, which we want to ignore here.

jasonLaster:
Screen Shot 2020-09-21 at 6 29 00 PM

jasonLaster (approved):

PR #682, opened 2020-09-18
Refactor App.js

jaril:
This patch:
- Refactors `App.js` to simplify its structure
- Wraps `App.js` children with a new `ApolloProvider` that then passes down the correct `ApolloClient` context
- Adds an error page for when Auth0 needs to display a popup to get the user's consent, but is not able to

### DIFF
--- a/src/ui/components/Account.js
+++ b/src/ui/components/Account.js
@@ -1,0 +1,36 @@
+import React, { useEffect } from "react";
+import Recordings from "./Recordings/index";
+import { useAuth0 } from "@auth0/auth0-react";
+import Header from "./Header";
+
+function WelcomePage() {
+  const { loginWithRedirect } = useAuth0();
+
+  return (
+    <div className="welcome-screen">
+      <div className="welcome-panel">
+        <img className="logo" src="images/logo.svg" />
+        <img className="atwork" src="images/computer-work.svg" />
+        <button onClick={() => loginWithRedirect()}>Log In</button>
+      </div>
+    </div>
+  );
+}
+
+function AccountPage() {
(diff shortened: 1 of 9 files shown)


src/ui/components/DevTools.js
```diff
@@ -0,0 +1,100 @@
...
+      />
+    );
+  }
+
+  render() {
+    const { hideComments, loading, commentVisible } = this.props;
+    const recordingIsLoading = loading < 100;
+
+    if (recordingIsLoading) {
+      return renderLoadingBar();
```
jasonLaster, line 75: https://gist.github.com/jasonLaster/a6d64ae10454b2247332d2446faf3b39

^ small fix

src/ui/components/shared/Error.js
```diff
@@ -0,0 +1,22 @@
...
+export default function Error() {
+  const browser = getBrowser();
+
+  return (
+    <div className="popup-error">
+      <h3>Uh oh</h3>
+      <p>Please turn off your pop-up blocker and refresh this page.</p>
+    </div>
+  );
+}
```
jasonLaster, line 22: Nice, i think this is great now that we know this is the problem

src/ui/components/Account.js
```diff
@@ -0,0 +1,36 @@
...
+    return <WelcomePage />;
+  }
+
+  return (
+    <>
+      <Header />
+      <AccountPage />
+    </>
+  );
+}
```
jasonLaster, line 36: So much cleaner

jasonLaster (approved):

Issue #438, opened 2020-08-21
Support (BigInt, ResizeObserver)


jasonLaster:
https://sentry.io/organizations/replay/issues/1852751139/?project=5399075&query=is%3Aunresolved
https://sentry.io/organizations/replay/issues/1852461334/?query=is%3Aunresolved&statsPeriod=14d

hbenl:
Polyfilling is impossible (BigInt) or not recommended (ResizeObserver), so I used a "ponyfill" for ResizeObserver in #685 and replaced BigInt with JSBI in #687.

hbenl:
Fixed in #685 and #688

PR #687, opened 2020-09-20
use JSBI instead of the native BigInt (#438)

hbenl:
there is one more place where BigInt is used: https://github.com/RecordReplay/devtools/blob/3019a433a308451405f949c3c9a43955493ea5b9/src/protocol/thread.js#L341
In this case the BigInt is stored in a ValueFront. More investigation would be needed to find out how to replace it.
However, I think this only affects recordings where the recorded page uses BigInt itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5436,6 +5436,11 @@
         "esprima": "^4.0.0"
       }
     },
+    "jsbi": {
+      "version": "3.1.4",
+      "resolved": "https://registry.npmjs.org/jsbi/-/jsbi-3.1.4.tgz",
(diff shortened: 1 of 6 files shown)


hbenl:
Actually since we only compare BigInts, we could also create comparison functions that compare string length and then (if the length is the same) compare as strings, which would be faster.

bhackett1024:
> there is one more place where BigInt is used:
> 
> https://github.com/RecordReplay/devtools/blob/3019a433a308451405f949c3c9a43955493ea5b9/src/protocol/thread.js#L341
> 
> In this case the BigInt is stored in a ValueFront. More investigation would be needed to find out how to replace it.
> However, I think this only affects recordings where the recorded page uses BigInt itself.

Yeah, this is for BigInt values from the recorded process itself.  I'm not aware of any places where we actually need this to be a bigint.  It should be fine to just return the string here and see if any tests break.

hbenl:
Superseded by #688

PR #714, opened 2020-09-24
[Console] Add the source location and delete button for each console message

gabrielluong:
It's probably easier to review individual commits.

Before:

Screen Shot 2020-09-24 at 12 11 54 AM

After:

Screen Shot 2020-09-24 at 12 12 41 AM

I experimented with showing the Delete button for every message since there's no longer a way to just clear all the messages. I think this will help with filtering unwanted messages from a user's perspective. Let me know what yall think.

### DIFF
--- a/src/devtools/client/webconsole/components/Output/Message.js
+++ b/src/devtools/client/webconsole/components/Output/Message.js
@@ -427,7 +427,7 @@ class Message extends Component {
           repeat,
           " ",
           location,
-          type == "command" ? this.renderDeleteButton() : null
+          type == MESSAGE_TYPE.COMMAND ? this.renderDeleteButton() : null
         ),
         attachment,
         ...notesNodes
--- a/src/devtools/client/webconsole/components/Output/Message.js
+++ b/src/devtools/client/webconsole/components/Output/Message.js
@@ -191,10 +191,10 @@ class Message extends Component {
   }
 
   renderDeleteButton() {
-    const { dispatch, message } = this.props;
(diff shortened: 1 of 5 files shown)


src/devtools/client/webconsole/components/Output/Message.js
```diff
@@ -427,7 +426,8 @@ class Message extends Component {
           repeat ? " " : null,
           repeat,
           " ",
-          type == "command" ? this.renderDeleteButton() : null
+          location,
```
jasonLaster, line 429: Nice find

src/styles.css
```diff
@@ -98,6 +99,7 @@
 
 @import url("devtools/client/themes/variables.css");
 @import url("devtools/client/themes/webconsole.css");
+@import url("devtools/client/themes/components-frame.css");
```
jasonLaster, line 102: 👍

jasonLaster (commented):
Going to sleep on some of the user questions.

bhackett1024:
This looks really nice, I've missed having a delete button for old console evaluations before.  Maybe a question for followup, but would it make sense to have a restore button that undeletes all messages?  I could imagine deleting messages (intentionally or accidentally) and wanting to restore them later.

jaril (approved):
Looks good!

Let's pass in the message type in the delete action so we can track it in logrocket

https://gist.github.com/jaril/c3fcf0f8c1254ea6db55029067bdd872

jaril (approved):

PR #717, opened 2020-09-24
Show previews of the recording while it's loading

### LINKED ISSUE #705: [Viewer] Show recording screen sooner
**Describe the bug**

I went to view a recording and waited 30 seconds to see the the recording UI

I think we can speed this up by talking to Hasura directly and getting the recording title, and screen and showing that before the server has gotten setup.

### LINKED ISSUE #711: [Accounts] speedup initial UI
**Describe the bug**

At the moment we fetch all of the recording data (including images) to show at once. We should break it up so that we can quickly show something and then fill it in.

jaril:
Fixes #705

Sep-24-2020 13-21-31

This adds a new loading screen for when users are opening a recording. Currently, we only display a white blank screen with a loading bar by the header. This is not a good experience for somebody who, for example, might have to wait 30s for a recording to load.

To keep them from clicking away, this patch adds more obvious visual cues to indicate that things are progressing:
1) Moves the loading bar from the header to the center of the screen.
2) Shows a preview of the recording that displays a preview of the current loading progress

### DIFF
--- a/src/ui/actions/app.js
+++ b/src/ui/actions/app.js
@@ -10,13 +10,15 @@ export function setupApp(recordingId, store) {
     }
   );
 
-  const loadingInterval = setInterval(() => store.dispatch(bumpLoading()), 500);
+  const loadingInterval = setInterval(() => store.dispatch(bumpLoading()), 1000);
 }
 
 function bumpLoading() {
   return ({ dispatch, getState }) => {
     const loading = selectors.getLoading(getState());
-    dispatch({ type: "loading", loading: Math.min(loading + 1, 99) });
+    const increment = Math.random() * 4;
+
+    dispatch({ type: "loading", loading: Math.min(loading + increment, 99) });
   };
 }
 
--- a/src/ui/components/DevTools.js
+++ b/src/ui/components/DevTools.js
@@ -1,4 +1,4 @@
-const React = require("react");
+import React, { useState, useEffect } from "react";
 import { connect } from "react-redux";
 
 import Toolbox from "./Toolbox";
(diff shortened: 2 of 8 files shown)


jaril:
Follow-up thoughts:
#### Loading tips
We should add loading tips. The effort would be trivial, and would actually be really useful for new users.
#### Misdirection
It appears that the big bottleneck here, processing the recording, happens once. After that, refreshing the page seems snappy. One way we can decrease the _perceived_ loading time is if we pretend that it's a prompt. 

Say a user just finished saving the recording, and the link is opened immediately. Instead of waiting for the recording to be processed and ready, they're asked for a Recording Title and Description. By the time they're done, we're much closer to finishing processing their recording.

It won't change the perceived loading time if they copy that link and open it on another browser. Nor will it do anything for other users viewing that recording. 

But for that one user's first open, it will appear snappy. Well, snappier.

src/ui/actions/app.js
```diff
@@ -10,13 +10,15 @@ export function setupApp(recordingId, store) {
...
 
-  const loadingInterval = setInterval(() => store.dispatch(bumpLoading()), 500);
+  const loadingInterval = setInterval(() => store.dispatch(bumpLoading()), 1000);
 }
 
 function bumpLoading() {
   return ({ dispatch, getState }) => {
     const loading = selectors.getLoading(getState());
-    dispatch({ type: "loading", loading: Math.min(loading + 1, 99) });
+    const increment = Math.random() * 4;
```
jaril, line 19: In the case that there's no progress, this keeps the loading bar artificially progressing for ~30s. The hope, of course, is that the recording would have been ready way before that 30s mark.

jasonLaster:
Love it!

All for the misdirection. Even if they just save the defaults that would buy us time to do some of the post processing. 

I wonder if the progressbar below the screenshot might be valuable realestate for the form... I think it looks better, but in the future might get in our way.

src/ui/components/DevTools.js
```diff
@@ -1,102 +1,132 @@
...
-    const { updateTimelineDimensions } = this.props;
-    const { orientation } = this.state;
+  const image = `url(data:${preview.mimeType};base64,${preview.data})`;
+  return <div className="preview-container" style={{ backgroundImage: image }}></div>;
+}
 
-    let startPanel, endPanel;
-    const vert = orientation != "bottom";
-    const toolbox = <Toolbox />;
+function RecordingLoadingScreen({ loading, recordingDuration }) {
```
jasonLaster, line 61: This might warrant being in its own file... 

Also i think we might want to ask apollo for the recording image, title, etc

jasonLaster (approved):
Lets land and iterate

jasonLaster:
moved this to #705 :)

PR #740, opened 2020-09-26
Refactor getPreview to avoid typeError

jasonLaster:
This avoids a typeError when nextPaintPoint returns a non-object. It also removes the try/catch based on my assumption that closest paint point could return a non-object too.


jasonLaster:
@jaril i'm landing to keep Sentry from 😢 and depleting our error budget. feel free to tweak modnay

jaril:
Nope - makes sense. Thanks for doing it!

PR #743, opened 2020-09-26
use the client from the protocol repository

hbenl:
I considered several options to reference code from the protocol repository in the devtools:
1. add it to `package.json` in the normal way
  this is currently not possible because we don't publish the protocol package in the npm registry (yet?)
  there is also one possible downside to this approach: any changes in the protocol repository would need to be published in the npm registry before they can be used in the devtools (making developing changes that span both repositories harder to do, although `npm link` can help with that)
2. clone the protocol repository next to the devtools and reference it directly in the imports (e.g. `require("../../../../protocol/js/client")`)
  downsides: ugly imports in many files, setting up and building devtools becomes more complicated, typescript tooling has some issues with referencing files from outside of the current package
3. clone the protocol repository next to the devtools and reference them in both `compilerOptions.paths` in `tsconfig.json` and `resolve.modules` or `resolve.alias` in `webpack.config.js`
  downsides: same as 2. minus the ugly imports plus the need to express this dependency twice (in `tsconfig.json` and `webpack.config.js`)
4. clone the protocol repository next to the devtools and reference it in `package.json` using a local path (i.e. `"record-replay-protocol": "../protocol"`)
  downsides: same as 2. minus the ugly imports
5. reference the protocol repository in `package.json` as a git url
  downsides: the result of the typescript compilation of the protocol needs to be checked into git, changes in the protocol repository need to be pushed to github before they can be used in the devtools

Option number 5 seems to be the cleanest solution by far, so I went for that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7399,6 +7399,20 @@
         "readable-stream": "^2.0.2"
       }
     },
+    "record-replay-protocol": {
+      "version": "git://github.com/RecordReplay/protocol.git#81d7fdd99c5a216674594e861a6a7713a51a09a9",
+      "from": "git://github.com/RecordReplay/protocol.git",
+      "requires": {
+        "ws": "^7.3.1"
+      },
+      "dependencies": {
+        "ws": {
+          "version": "7.3.1",
+          "resolved": "https://registry.npmjs.org/ws/-/ws-7.3.1.tgz",
+          "integrity": "sha512-D3RuNkynyHmEJIpD2qrgVkc9DQ23OrN/moAwZX4L8DfvszsJxpjQuUq3LMx6HoYji9fbIOBY18XWBsAux1ZZUA=="
+        }
+      }
+    },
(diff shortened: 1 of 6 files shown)


jasonLaster (approved):
Thanks for the writeup on the different import options. I agree that 5 is the best for now. eventually we will want to version the protocol, but i agree that we should hold off on that for now.

jasonLaster:
This is exciting. Great job!

hbenl:
> Thanks for the writeup on the different import options.

I've run into this issue several times in recent years: javascript lacks the concept of multi-package projects (similar to "Solutions" in Visual Studio, "Workspaces" in Eclipse or "Multi-root workspaces" in VS Code) where the packages reference each other via the npm registry (Option 1 in my comment above) but when you work in the context of the multi-package project, you want the packages to import each other locally (like what `npm link` does) so that you can easily develop changes that span multiple packages.

PR #755, opened 2020-09-29
Convert DevTools from a functional to a class component

gabrielluong:
This will be needed for further work with making the Toolbox toggleable. We want to be able to call the toolbox initialization in componentDidMount().


jaril (approved):

jaril:
Looks good!

We can do the same thing in a functional component with `useEffect(() => { initializeToolbox() }, [])`, which behaves the same way as componentDidMount. But seeing as we're not using any other hooks in DevTools, reverting it to a class component makes more sense.

PR #756, opened 2020-09-29
Add a isToolboxOpen pref for the app state

gabrielluong:
This adds a toggeable isToolboxOpen app state for https://github.com/RecordReplay/devtools/issues/734.

### DIFF
--- a/src/ui/reducers/app.js
+++ b/src/ui/reducers/app.js
@@ -4,6 +4,8 @@ function initialAppState() {
   return {
     recordingId: null,
     theme: "theme-light",
+    // Whether or not the developer tools toolbox is opened.
+    isToolboxOpen: prefs.isToolboxOpen,
     splitConsoleOpen: prefs.splitConsole,
     selectedPanel: prefs.selectedPanel,
     tooltip: null,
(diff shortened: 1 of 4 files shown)


src/ui/reducers/app.js
```diff
@@ -4,6 +4,8 @@ function initialAppState() {
   return {
     recordingId: null,
     theme: "theme-light",
+    // Whether or not the developer tools toolbox is opened.
```
jasonLaster, line 7: This is open to debate, but I'd prefer not leaving comments on individual state fields unless we think there's something that might be confusing.

For instance, it's possible some of the loading states could be confusing ...

jasonLaster (approved):
Looks good. One small discussion item, but feel free to land with or without

src/ui/reducers/app.js
gabrielluong: Personally, I would prefer to see every state field annotated. I wouldn't easily be able to tell you if `selectedPanel` is a `String` or possibly even a `Number` for the index of the selected panel or the expected object for `tooltip` or the possible values for `theme`.

Boolean states such as splitConsoleOpen could use a bit more structure such as a `is` prefix to make it clear that it's a `Boolean`. Undoubtedly, using typescript for all app state will help as well, but I think leaving the comments like these will at least help somebody in the future whether it is doing the eventual typescript conversion and figuring out the types or someone looking to add a new state field in the future and seeing some existing pattern in the code to follow.

PR #759, opened 2020-09-29
Make sure that the user is flushed from the browser pref upon logout

### LINKED ISSUE #758: Replay browser continues to show the Record button even though the user has logged out
**Describe the bug**
On the Replay browser, the record button is shown even though the user has already logged out.

**To Reproduce**
1. Open the Replay browser.
2. Login.
3. Logout.
4. Notice that the record button is still displayed.

**Expected behavior**
The record button should be hidden upon user logout.

jaril:
Fixes #758.

We usually tell the Replay browser's prefs about the user's ID through the Avatar component. 

https://github.com/RecordReplay/devtools/blob/86a74874efcd0a739c6f1623e4467ea2cb27d7e0/src/ui/components/Avatar.js#L8-L20

The problem is that Avatar component is not rendered if the user is not logged in. So when a user logs in, logs out, then is redirected to the welcome page, the Replay browser's prefs are not updated.

This makes it so we tell the browser that the user is logged out.

### DIFF
--- a/src/ui/components/Account/index.js
+++ b/src/ui/components/Account/index.js
@@ -5,6 +5,7 @@ import Header from "../Header";
 import Loader from "../shared/Loader";
 import { gql, useQuery } from "@apollo/client";
 import classnames from "classnames";
+import { setUserInBrowserPrefs } from "../../utils/browser";
 
 import "./Account.css";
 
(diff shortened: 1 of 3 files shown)


src/ui/components/Account/index.js
```diff
@@ -63,6 +64,10 @@ function AccountPage() {
 function WelcomePage() {
   const { loginWithRedirect } = useAuth0();
 
+  useEffect(() => {
+    setUserInBrowserPrefs("");
```
jasonLaster, line 68: `setUserInBrowserPrefs(null)`

jasonLaster (approved):
One tweak. otherwise great

PR #762, opened 2020-09-29
Display the error message for "Session died unexpectedly" more prominently

### LINKED ISSUE #745: Handle "Session died unexpectedly" cases better
Right now, when a user loads a recording and receives the "Session died unexpectedly" error, the page just stops there. 

It's not clear what they should do next, or how they should do it. Should they simply reload? Or are the servers down and they should just come back 30 minutes from now?

If it's the first one, we can save them a step and just reload the page ourselves when we find out that session died unexpectedly with `location.reload()`. If it's the latter, there's not much we can do, but we can at least let them know that it's *not* the first case, and that they should come back later.

### LINKED ISSUE #765: Style the error modal
It'd be great to show a more styled error UI

https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/Replay%3A%3ASketeches?node-id=786%3A878

jaril:
Fix #745, Fix #765 

Demo:
https://www.loom.com/share/a4e7185e660244759959fb6c3b3331ca

Screenshot:
image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -68,8 +68,8 @@ function onUploadedData({ uploaded, length }) {
   }
 }
 
-function onSessionError({ message }) {
-  setStatus(`Error: ${message}`);
+function onSessionError(error) {
+  store.dispatch({ type: "set_error_message", message: error.message });
 }
 
 let initialized = false;
--- a/src/ui/components/shared/UserPrompt.css
+++ b/src/ui/components/shared/UserPrompt.css
@@ -1,0 +1,130 @@
+.user-prompt {
+  display: flex;
+  flex-direction: column;
+  padding: 5rem;
+  margin: auto;
+  align-items: center;
+  font-size: 1.25rem;
+  text-align: center;
+  width: 630px;
+  --light-grey: #969696;
+  --dark-grey: #696969;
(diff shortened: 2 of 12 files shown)


src/main.js
```diff
@@ -68,8 +68,8 @@ function onUploadedData({ uploaded, length }) {
   }
 }
 
-function onSessionError({ message }) {
-  setStatus(`Error: ${message}`);
```
jasonLaster, line 72: we should have an action for this 

`store.dispatch(setErrorMessage(error.message))`

src/ui/components/shared/UserPrompt.css
```diff
@@ -0,0 +1,130 @@
...
+.user-prompt button .content {
+  padding-top: 0.5px; 
+}
+
+.user-prompt button.clicked {
+  background: linear-gradient(152.34deg, #3398ff 25.66%, #007dfd 78.07%);
+}
+
+.user-prompt button.clicked .img.refresh {
+  animation: spin .7s linear infinite;
```
jasonLaster, line 90: I love this attention to detail!

jasonLaster (approved):

PR #771, opened 2020-10-01
Make the console timeline navigation button more noticeable

jaril:
This patch does a few things:
1) It makes the previously subtle seek button in the timeline more prominent by replacing it with a big blue button that pops up from the left when a user hovers on a message. It's now very hard to miss.
2) It adds a label to the seek button when the user hovers on the button (e.g. rewind, fast-forward).
3) Once a user seeks to a message, we leave a purple paused button displayed next to the message they're paused on. This indicates that the user is paused there, and more importantly, is clickable to look at that source line in the debugger.

Demo: https://www.loom.com/share/4a5601aadeee4a46913b8789e6b557b0

Screenshot:
image

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -959,20 +959,6 @@ a.learn-more-link.webconsole-learn-more-link {
   margin-top: calc(var(--console-output-vertical-padding) - 1px);
 }
 
-.webconsole-app .message:hover > .icon.rewindable {
-  cursor: pointer;
-  transform: rotate(180deg);
-}
-
-/*
- * we flip the next.svg icon by default because when we're
- * not paused, we would jump back. We remove the transform here
- * because we want to jump forward.
- */
-.webconsole-app .message.paused ~ .message:hover .icon.rewindable {
-  transform: none;
-}
-
 .webconsole-output .paused ~ .message:last-of-type,
 .webconsole-output .paused:last-of-type {
   border-bottom-width: 0;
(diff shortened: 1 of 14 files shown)


src/devtools/client/themes/webconsole.css
```diff
@@ -994,3 +980,112 @@ a.learn-more-link.webconsole-learn-more-link {
...
+  margin-left: calc(0px - var(--message-left-padding));
+}
+
+.message:hover .button .img {
+  opacity: 100%;
+}
+
+.overlay-container:hover .button {
+  background: linear-gradient(152.2deg, #67DAFF 6.75%, #4382FF 93.14%);
+  box-shadow: 1px 0px 3px rgb(0,0,0,0.3);
```
gabrielluong, line 1077: nit: let's add a space after each comma in rgb()

src/ui/components/Timeline/index.js
```diff
@@ -255,7 +255,8 @@ export class Timeline extends Component {
   };
 
   onPlayerMouseUp = e => {
-    const { hoverTime, hoveringOverMarker } = this.props;
+    const { hoverTime } = this.props;
+    const { hoveringOverMarker } = this.state;
```
gabrielluong, line 259: It would be nice to add this to the default state in the constructor()

jasonLaster: Good catch.

jasonLaster (approved):
Pushing a couple of visual tweaks

PR #788, opened 2020-10-02
Change markers from circle to diamond, and add a timeline scrubber

jaril:
This patch is mainly concerned with changing the timeline message markers from circles to diamonds. The main reasons for this change are:
- **Diamonds are more visually precise**: Instead of visually estimating the middle position of a circle to figure out the position of a message, a diamond has a pointy end that tells you exactly that (_Example 4_). As seen in the example, with the diamond, you can get to 1px precision, lining up the diamond tip with the paused line.
- **Diamonds are more distinct for overlapping markers**: Take for example two timeline markers that are 2px away from each other (_Example 2_). Diamond markers with their pointy ends makes each marker distinct - the circle markers just blend together into a blob.

image
image

Other things this patch handles:
- **Removed the hover animation for timeline message markers**: Before, it would scale 125% upon hovering on the marker. The reason is that the marker hover state is already indicated by the cursor changing from an arrow to a pointer. The scaling on hover is redundant, and more importantly, has the negative side effect of covering up other markers that are close to the hovered marker. 
- **Changed the color of timeline areas after a pause point to blue**: Before, timeline areas after the pause point would be white. This meant that if a user was paused at 50% progress, the timeline would look half blue and half white. By making it all blue, the timeline is stronger visually even if the pause point is modified. We distinguish between the pre-pause areas and post-pause areas by making the post-pause areas lighter. This can be seen in _Example 4_ where both the background and the progress line is lighter after the purple pause line.
- **Highlights the message that the user is hovering in the console**: We already did this before, but I've changed it to be more subtle. The main objective for this effect was for the user to draw the conclusion that the console messages correspond to timeline markers. The way we do it is by applying a highlighted state to the marker, and showing the screenshot as a tooltip over that marker. The tooltip was already noisy and attention grabbing, so I've made the highlighted message state identifiable, but "pop" less.
- **Added a timeline scrubber**: Before, to indicate the pause point we would simply have a purple line within the progress bar. To make the timeline look even more like a timeline, we've added a scrubber visual to that line (_Example 5_). Scrubbers, which usually look like this one or just a triangle pointing down, are a given in most editor timeline UIs. For us, it's nice to have because it makes pause point more prominent and easier for the user to find. What I like most about it is when it's combined with diamond markers, it becomes VERY obvious to users that you're paused on a marker. This is something that users would always point out in live onboarding - that it's not obvious from the timeline that they're paused on a console message. With this, the scrubber points at the diamond, the diamond points at the marker, and the paused line lines both of them up. If you have the console open, you'll even see the overlay button to the left of the message. 

image

Things this patch doesn't do, and that we should discuss before filing feature requests for:
- **We should get the horizontal positioning precision for the markers/scrubber/paused line down to the exact 1px**: Right now there's a 1px tolerance. The message marker width is an intentionally an odd number (11px) so that theoretically, any horizontal marker can be center aligned to 1px at its middle (6px). Right now it shifts between the 5px/6px position. Even if we get this right, the alignment logic is fragile since it depends on so many variables defined in multiple places, so we should centralize it and define the width of the marker/message/scrubber width in one place. And to keep it from shifting, ideally we would write unit tests that makes sure everything is in its right place. This might sound tedious for some shifty pixel values, but Replay is in the timeline business, and we don't have an excuse for the timeline being anything but perfect.
- **We should rethink how we implement the timeline tooltip**: The timeline tooltip is good at its current job - showing the user a preview of that hover point in time. The preview feature is a must have, but it's almost _too_ good at its job and has its pitfalls. It's too big: on a 1080p screen the viewer is 460px high and the viewer takes up 180px(40%) of that. It has a noticeable lag. And the one that pains me the most, it activates when a user hovers on a console message - while well-intentioned, what ends up happening is that as the user hovers **up and down** in the console, their eyes are moves from **left to right** at an area _above_ the console. This steals focus from console message actions, like the rewind button which is to the left of a message. It's made even worse if for example, the user hovers on a message that activates a tooltip all the way to the right of the timeline. Good luck making them notice any UI on the left - they'll go cross-eyed.
- **We should add some way for users to notice a newly added marker**: This is a small step towards differentiation of markers in the timeline. At the very least, when the user performs an action that might modify the timeline (e.g. adding a logpoint), it should be obvious to a user that the timeline has been modified and it should be easy to point to that new modification. Right now, when a user adds a logpoint that adds 1 marker to a timeline with 0 markers, it's fairly easy for them to tell that it's been modified (it went from 0 to 1 markers), and to point to the new modification (literally the only one marker on the timeline). This gets more difficult as the number of markers in the timeline increase, and it becomes downright impossible very quickly. We can easily fix this by making any new message marker to have a newly added state, where we could distinguish it by any number of effects for ~2-3 seconds. The more subtle options that are easy to implement include changing its color and making it bigger. The more explicit options, and ones I particularly would enjoy, is if we started exploring popups. It comes in two forms. One can be a tooltip just above each marker with information (e.g. Logpoint script.js:50:1). The other one can be a small popup modal that tells the user about an action they just performed (e.g. Added 10 timeline markers from Logpoint in script.js:50:1).
- **We should make it easier for users to zoom in/out**: We can already zoom in the timeline by hovering on the scrollbar and scrolling up/down. However, this UI is not discoverable - I have yet to see a user find out about the zoom feature by themselves. It also doesn't make it obvious to a user if they're zoomed in/out. Scroll to zoom is nice for power users, but we should also have a foolproof zoom UI that's discoverable. An effective UI would be something similar to video editors like Adobe Premiere/AfterEffects where there is a timeline summary above the timeline with a start pointer and end pointer. Those start/end pointers are draggable, and affects the zoom in the timeline. [Vue devtools](https://headwayapp.co/vue-js-devtools-changelog/timeline-and-api-166716) recently implemented a stripped down version of this and it looks really good. With this, we also won't have to indicate zoom level (100%/150%/200%) somewhere because it the user's zoom status would be easily inferred from the start/end pointers. Another option is to add a zoom slider outside of the actual timeline, that increases the zoom relative to the current pause point, similar to [Sonuum](https://sonuum.com/) where they have a slider that goes from a minus magnifying glass to a plus magnifying glasses.
- **We should handle overlapping markers better**: Right now, we simply check each marker as we render them. If we find that there's already a rendered marker within 2.5px of our new marker, we don't render the new marker. This isn't great since we're hiding markers. Ideally all the marker should be on the timeline, and we deal with overlapped markers in a different way. This is an interesting problem because there's no obvious prior art. Video editor keyframes don't get nearly as crowded as Replay's, and their solution to crowding is to display everything and let them overlap. A simple solution might be to group markers that are close together and present them as a "blobbed" selection for the user to click on and zoom into. We could also communicate how many markers are in the blob.

Things we should discuss but are not urgent and should not be prioritized:
- **We should add some way for users to differentiate between different markers**: I don't feel like this should be prioritized, because it will require adding more complexity in the timeline before we nail down the basics. But at some point, it should be easy for users to tell different types of markers apart from each other. There are a lot of ways we can do this as markers can be differentiated based on how it was added (inline console.log, warning, error, logpoint, event logpoint, etc), their source (foo.js vs bar.js), the function they're in (Foo() vs Bar()). This is a long discussion as it also shoehorns into a multiple timeline feature, but again, this should be done later.
- **We should add time indicators to the timeline**: A few months ago we had tick marks in the timeline that was a good way of indicating _some_ measure of time, although it didn't actually tell you which minute/second each tick corresponded to. It would be nice to have ticks and timestamps - it's an expected visual element in editor timelines. But I'm not sure that we need it, and even if we did, that we need to do it now. For one, users I've talked to have never once referred to a time while using debugging a recording (e.g. "I wish I could see what happened at the 43 second mark"). My sample is limited, but my theory is that in a recording, a user's proxy for time passing is either the viewer screenshot or the code execution itself. They don't think in minutes and seconds, rather they think in terms of what events have happened. I found this interesting and it makes sense. For example, when somebody is editing a podcast together, a lot of their actions (splicing audio) is based on the words being said at that time. They don't think "I should extract the audio from this clip starting at 2.2s and end at 10.1s", instead they think "I should extract the audio from this clip starting when the person starts talking, and end when they stop talking". In fact, [businesses like Descript](https://www.descript.com/) are built around navigating a timeline by words, rather than timestamps. This is hardly uncommon - the same pattern is seen in almost every video player (Youtube, Netflix) where they don't have tick marks in the timeline. I think we have the same phenomenon in Replay. It would be _nice_ to have ticks and timestamps, but it's hardly important. I think there might be value in displaying the current paused time and total recording time, but that's outside of ticks and timestamps.

### DIFF
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -31,12 +31,11 @@ const { assert } = require("protocol/utils");
 import { actions } from "../../actions";
 import { selectors } from "../../reducers";
 import Message from "./Message";
+import { timelineMarkerWidth } from "../../utils/timeline";
 
 const { div } = dom;
 
 import "./Timeline.css";
-
-const markerWidth = 11;
 
 function classname(name, bools) {
   for (const key in bools) {
@@ -485,7 +484,7 @@ export class Timeline extends Component {
   // Get the percent value for the left offset of a message.
   getLeftOffset(message) {
     const messagePosition = this.getVisiblePosition(message.executionPointTime) * 100;
-    const messageWidth = (markerWidth / this.overlayWidth) * 100;
+    const messageWidth = (timelineMarkerWidth / this.overlayWidth) * 100;
 
     return Math.max(messagePosition - messageWidth / 2, 0);
   }
@@ -532,7 +531,7 @@ export class Timeline extends Component {
     }
 
     const middlePercent = this.getVisiblePosition(comment.time) * 100;
-    const widthPercent = (markerWidth / this.overlayWidth) * 100;
+    const widthPercent = (timelineMarkerWidth / this.overlayWidth) * 100;
     const percent = Math.max(middlePercent - widthPercent / 2, 0);
 
     return dom.a({
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -229,12 +229,15 @@ export class Timeline extends Component {
     const isHovered = window.elementIsHovered(this.$progressBar);
     if (!isHovered) {
       clearInterval(this.hoverInterval);
+      this.hoverInterval = null;
       hideTooltip();
     }
   };
 
   onPlayerMouseEnter = async e => {
-    this.hoverInterval = setInterval(this.hoverTimer, 100);
+    if (!this.hoverInterval) {
+      this.hoverInterval = setInterval(this.hoverTimer, 100);
+    }
   };
 
   onPlayerMouseMove = async e => {
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -36,7 +36,7 @@ const { div } = dom;
 
 import "./Timeline.css";
 
-const markerWidth = 7;
+const markerWidth = 11;
 
 function classname(name, bools) {
   for (const key in bools) {
@@ -488,7 +488,7 @@ export class Timeline extends Component {
   }
 
   renderMessages() {
-    const { messages, currentTime, pausedMessage, highlightedMessage, zoomRegion } = this.props;
+    const { messages, currentTime, highlightedMessage, zoomRegion } = this.props;
     let visibleIndex;
 
     return messages.map((message, index) => {
@@ -499,7 +499,6 @@ export class Timeline extends Component {
           index={index}
           messages={messages}
           currentTime={currentTime}
-          pausedMessage={pausedMessage}
           highlightedMessage={highlightedMessage}
           zoomRegion={zoomRegion}
           overlayWidth={this.overlayWidth}
@@ -556,21 +555,43 @@ export class Timeline extends Component {
     return comments.map(comment => this.renderCommentMarker(comment));
   }
 
-  renderHoverPoint() {
-    const { hoverTime, hoveredMessage } = this.props;
+  renderHoverScrubber() {
+    const { hoverTime, hoveredMessage, currentTime } = this.props;
+
     if (hoverTime == null || hoveredMessage) {
       return [];
     }
-    const offset = this.getPixelOffset(hoverTime);
-    return [
-      dom.span({
-        className: "hoverPoint",
+
+    let offset = this.getPixelOffset(hoverTime);
+
+    return dom.span(
+      {
+        className: "scrubber hovered",
         style: {
           left: `${Math.max(offset, 0)}px`,
-          zIndex: 1000,
         },
-      }),
-    ];
+      },
+      dom.img({
+        src: "images/scrubber.svg",
+      })
+    );
+  }
+
+  renderPausedScrubber() {
+    const { hoverTime, hoveredMessage, currentTime } = this.props;
+    let offset = this.getPixelOffset(currentTime);
+
+    return dom.span(
+      {
+        className: "scrubber",
+        style: {
+          left: `${Math.max(offset, 0)}px`,
+        },
+      },
+      dom.img({
+        src: "images/scrubber.svg",
+      })
+    );
   }
 
   renderUnprocessedRegions() {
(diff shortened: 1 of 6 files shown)


gabrielluong:
I feel your comments are very valuable here, and I would hate to see the comments that weren't addressed in this PR to be lost when it is merged. I think filing some kind of [meta] issue or adding them to some kind of doc could be super valuable from a more open discussion pov.

Garbee:
I think when it comes to keeping a lot of great information like this around, it boils down to us figuring out a structured way of keeping a reference of user stories associated with the app functionality.

i.e. We could have a Notion section for each feature, a sub-page with implemented user stories and a sub-page with desired ones. Or individual GH issues per user story as an enhancement/feature request. Where only implemented ones end up in some document/specification of the feature.

src/ui/utils/timeline.js
```diff
@@ -1,4 +1,4 @@
-const timelineMarkerWidth = 7;
+const timelineMarkerWidth = 11;
```
Garbee, line 1: This number is repeated in so many places... Do we have nowhere to export this kind of thing from so it is shared across wherever it is needed?

I feel like any changes here could make it pretty easy to get out-of-sync.

Garbee (commented):
Just a cursory look over but everything seems to be in order with the patch and no immediate points of concern jump out. Just one repeated number shared around that I'd like us to try and get into one place.

Nice work. I especially like how a noticeable UX improvement comes with a net-removal of lines of code.

src/ui/utils/timeline.js
gabrielluong: I would recommend adding a constants.js file in ui/ if it makes sense.

src/ui/components/Timeline/index.js
```diff
@@ -36,7 +36,7 @@ const { div } = dom;
 
 import "./Timeline.css";
 
-const markerWidth = 7;
+const markerWidth = 11;
```
jasonLaster, line 39: perhaps we should import the marker width from `util/timeline`

jasonLaster (approved):
Looks great!

src/ui/utils/timeline.js
jaril: Good call - went ahead with creating a constants. Thanks for the suggestions!

Issue #704, opened 2020-09-22
[Timeline] playing the recording could be smoother


jasonLaster:
**Describe the bug**
Playing a recording starts and stops at different spots

**To Reproduce**

1. go to http://localhost:8080/view?id=81a8a24c-14dd-49f9-8856-c58e64ae6910
2. hit play and watch the recording

![](http://g.recordit.co/FY0gX3Zkxo.gif)

jasonLaster:
@hbenl what do you think we could do here?

jasonLaster:
I imagine we could use JS to render intermediate locations inbetween frames... not really sure

hbenl:
I think the marker should simply run smoothly through the timeline "in real time" (i.e. it traverses one recorded second in one second). Of course it has to pause when it's still downloading data for the next frame - but in that case it should keep running until it is immediately before that frame in the timeline and only pause there if the data still hasn't arrived.

jasonLaster:
@hbenl that sounds right. perhaps we can show a loading indicator if we're waiting on something :)

PR #793, opened 2020-10-03
Reimplement playback to make it smoother (#704)

hbenl:


jaril (commented):
This looks **really** good and appears to work well, thanks for doing the work Holger!

Issue #789, opened 2020-10-02
[Sentry] FETCHED_FRAMES fails for undefined frames


jasonLaster:
https://sentry.io/organizations/replay/issues/1928690485/?project=5399075&referrer=alert_email

https://app.logrocket.com/4sdo4i/replay/s/4-17c7b3c6-1618-4991-8130-f7547f05207e/0/8e49d574-f0ee-4d2b-a247-35871cfdc4b0?t=1601629700415.9827

<img width="859" alt="Screen Shot 2020-10-02 at 12 05 15 PM" src="https://user-images.githubusercontent.com/254562/94960331-902f4a80-04a7-11eb-9e33-60bd3c75980a.png">

PR #804, opened 2020-10-09
add guard for undefined value (#789)

hbenl:
I couldn't reproduce this issue but the fix seemed obvious enough...

### DIFF
--- a/src/devtools/client/debugger/src/reducers/pause.js
+++ b/src/devtools/client/debugger/src/reducers/pause.js
@@ -76,7 +76,7 @@ function update(state = createPauseState(), action) {
 
     case "FETCHED_FRAMES": {
       const { frames } = action;
-      const selectedFrameId = frames.length ? frames[0].id : undefined;
+      const selectedFrameId = frames && frames.length ? frames[0].id : undefined;
       return {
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/reducers/pause.js
```diff
@@ -76,7 +76,7 @@ function update(state = createPauseState(), action) {
 
     case "FETCHED_FRAMES": {
       const { frames } = action;
-      const selectedFrameId = frames.length ? frames[0].id : undefined;
+      const selectedFrameId = frames && frames.length ? frames[0].id : undefined;
```
jasonLaster, line 79: perhaps we don't want to dispatch `FETCHED_FRAMES` if we don't have frames.

So basically put the guard in the action.

hbenl: we still need to set `framesLoading` to `false`, so we still need to dispatch `FETCHED_FRAMES`.

jasonLaster: good call.

jasonLaster (approved):

PR #805, opened 2020-10-09
remove unused imports from Sharing.js

hbenl:
One of the imports resulted in typescript being added to the webpack bundle, almost doubling its size (and showing some webpack warnings).


jaril (approved):
Gotcha - thanks Holger!

PR #814, opened 2020-10-13
Handle 'This recording format is no longer supported' errors

### LINKED ISSUE #813: Handle backend error when recordings are not supported
When loading a V1 recorder recording, the backend now generates an error.  For  example, load https://replay.io/view?id=851897e1-e88c-414c-9933-4dad9d5337a3 and look in the console for a `Message failed Recording.createSession 
Object { code: 31, message: "This recording's format is no longer supported" }` message.  This error should be handled and the load stopped with an appropriate error message.  (We'll still want this error even when the V1 recorder is a distant memory, as we're not going to commit to supporting all old recordings forever.)

I tried changing things here https://github.com/RecordReplay/devtools/blob/ad3cd4d6550e8b4bae28ce2d71566af1902cfdd3/src/main.js#L54 to show an error, but the error only showed up very briefly before the page resumed showing a loading indicator.

jaril:
Fix #813. 

This does the trick, though we probably need to revisit how errors are handled in the frontend. I didn't take the time to plan out z-indices in advance and they're only going to get more convoluted from here.

### DIFF
--- a/src/ui/components/shared/Error.js
+++ b/src/ui/components/shared/Error.js
@@ -30,7 +30,17 @@ export function PopupBlockedError() {
   );
 }
 
-export function SessionError({ error }) {
+export function Error({ error }) {
+  if (error != "Session died unexpectedly") {
+    return (
+      <Modal opaque error>
+        <h1>Whoops!</h1>
+        <p>Looks like something went wrong with this page.</p>
+        <p className="tip">Error: {error}</p>
+      </Modal>
+    );
+  }
+
   return (
     <Modal translucent error>
       <h1>Whoops!</h1>
--- a/src/main.js
+++ b/src/main.js
@@ -53,6 +53,8 @@ async function createSession() {
   } catch (e) {
     if (e.code == 9) {
       store.dispatch(setErrorMessage("Invalid recording ID"));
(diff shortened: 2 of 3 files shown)


src/ui/components/shared/Error.js
```diff
@@ -30,7 +30,17 @@ export function PopupBlockedError() {
...
 }
 
-export function SessionError({ error }) {
+export function Error({ error }) {
+  if (error != "Session died unexpectedly") {
+    return (
+      <Modal opaque error>
+        <h1>Whoops!</h1>
+        <p>Looks like something went wrong with this page.</p>
+        <p className="tip">Error: {error}</p>
```
jasonLaster, line 39: Hmm, minor detail but I think we can drop the `Error: `. It should be read slightly better without it

jasonLaster (approved):
One small tweak, but otherwise great

jasonLaster:
What do you think of moving the copy to the right? Not sure...

I just adjusted padding and made the `p` widths 100%


Screen Shot 2020-10-13 at 1 21 22 PM
Screen Shot 2020-10-13 at 1 16 37 PM



Screen Shot 2020-10-13 at 1 21 34 PM

PR #818, opened 2020-10-14
Split protocol/thread.js into multiple files

hbenl:


hbenl:
I tried to make as little changes as possible (because the diff can't relate the code in the new files to the one in the old file), only imports and exports had to change.

bhackett1024 (approved):
Looks great, thanks!

PR #819, opened 2020-10-14
guard against an empty recordings array

hbenl:
I got some exceptions while debugging because `data.recordings[0]` was undefined. Perhaps the reason was that I used a recording that I didn't create.


jaril (approved):
Looks good! This most likely happened because the recording was set to private. 

If a recording is private and you had private feature toggled on (`app.features.private = true`), then the Header component wouldn't be rendered in the first place. But because we still have the `private` feature toggled off by default (`app.features.private = false`), the Header is rendered and tries to fetch a title that it doesn't have access. That's what causes the error.

This fixes it in the meantime which is great - thanks for doing this Holger!

PR #821, opened 2020-10-14
Redesign and fix the Sharing Modal

jaril:
This refactors the old `Modal` component to make a distinction between modals and user prompts. User prompts such as the recording loading screen and the first recording information now use the new `Prompt` component.

It also fixes the weird behavior in the `Sharing` modal whenever you try to add/remove a collaborator. Before, we would execute a Mutation (add/remove) and try to refetch the collaborators. The problem with this approach is that the refetch was done synchronously, which wasn't reliable. I've changed it so that we only refetch the collaborators after the mutation has completed.

## Screenshots
<details><summary>Modals</summary>
<br>

</details>
<details><summary>Prompts</summary>
<br>

</details>
<details><summary>Errors</summary>
<br>

</details>


src/ui/components/RecordingLoadingScreen.js
```diff
@@ -1,7 +1,8 @@
 import React, { useState, useEffect } from "react";
 import { connect } from "react-redux";
 import Header from "./Header/index";
-import Modal from "./shared/Modal";
+// import Modal from "./shared/Modal";
```
jasonLaster, line 4: :/

jasonLaster (approved):

PR #823, opened 2020-10-15
GitHub actions

hbenl:
- fix the lint action which succeeded even when there were linting errors
- add a typecheck action


jasonLaster (approved):

jasonLaster:
Lets fix the `Devtools` prettier fail in a followup

hbenl:
> Lets fix the Devtools prettier fail in a followup

That is fixed in #822

PR #829, opened 2020-10-16
Toggle privacy features on

jaril:
This patch:
1) Toggles the privacy features on 🎉
2) Makes it so that only the recording owner can set the recording privacy (private/public) and edit collaborators

### DIFF
--- a/src/ui/components/DevTools.js
+++ b/src/ui/components/DevTools.js
@@ -69,7 +69,7 @@ function getUploadingMessage(uploading) {
 function getIsAuthorized({ data, error, isAuthenticated }) {
   // We let Hasura decide whether or not the user can view a recording. The response to our query
   // will have a recording if they're authorized to view the recording, and will be empty if not.
-  return data.recordings.length > 0 ? true : false;
+  return data.recordings.length;
 }
 
 function DevTools({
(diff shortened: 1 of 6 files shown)


src/ui/utils/prefs.js
```diff
@@ -20,7 +20,7 @@ pref("devtools.toolbox-opened", true);
 pref("devtools.features.comments", true);
 pref("devtools.features.users", true);
 pref("devtools.features.auth0", true);
-pref("devtools.features.private", false);
+pref("devtools.features.private", true);
```
jasonLaster, line 23: can we get rid of this pref :)

src/ui/components/DevTools.js
```diff
@@ -65,15 +67,8 @@ function getUploadingMessage(uploading) {
...
-  if (!features.private) {
-    return true;
-  }
-
   // We let Hasura decide whether or not the user can view a recording. The response to our query
   // will have a recording if they're authorized to view the recording, and will be empty if not.
-  // What this doesn't explicitly tell us is *why* that user is allowed to view the recording: is the
-  // recording public, or is the user the original creator for that recording? You can check that
-  // by checking the recording's user field, which will be null for non-creator users.
   return data.recordings.length > 0 ? true : false;
```
jasonLaster, line 72: data can be undefined, so i recommend doing

```
return data?.recordings.length > 0
```

the ternary isn't needed 😉

jasonLaster (approved):
Looks great. Ship it!

bhackett1024:
Hi, sorry, but I had to revert this change because it broke all of our tests both when running locally and as a github action.

PR #830, opened 2020-10-16
New markup view

hbenl:
There's a new feature flag `devtools.inspector.features.old-markupview.enabled` for disabling the old markup view.
When this is set to `false`, the new markup view will be shown in its place.
When it is set to `true`, the old markup view will be shown and the new markup view will appear in the side panel if `devtools.inspector.features.new-markupview.enabled` is set to `true`.

There are still a couple of bugs to fix, but since this is happening behind a feature flag, I think we can merge this already.

### DIFF
--- a/src/devtools/client/inspector/markup/new-markup.js
+++ b/src/devtools/client/inspector/markup/new-markup.js
@@ -107,7 +107,7 @@ class MarkupView {
    * @param  {Boolean} isExpanded
    *         Whether or not the node is expanded.
    */
-  async addNodeToTree(node, { isExpanded = false } = {}) {
+  addNodeToTree(node, { isExpanded = false } = {}) {
     const parentNode = node.parentNode();
     const id = node.objectId();
 
@@ -169,6 +169,7 @@ class MarkupView {
    */
   async expandNode(nodeId) {
     await this.getChildren(this.nodes.get(nodeId));
+    if (!this.tree) return;
     this.tree[nodeId].isExpanded = true;
     this.store.dispatch(updateTree(this.tree));
   }
@@ -182,7 +183,7 @@ class MarkupView {
    */
   async getChildren(node) {
     const children = await node.childNodes();
-
+    if (!this.tree) return;
     this.tree[node.objectId()].children = children.map(child => child.objectId());
 
     // Adds the children into the tree model. No need to fetch their children yet
@@ -194,7 +195,7 @@ class MarkupView {
       }
 
       if (!this.tree[childNodeId]) {
-        await this.addNodeToTree(childNodeFront);
+        this.addNodeToTree(childNodeFront);
       }
     }
   }
@@ -230,11 +231,11 @@ class MarkupView {
       }
 
       this.nodes.set(currentNode.objectId(), currentNode);
-      await this.addNodeToTree(currentNode, { isExpanded: true });
+      this.addNodeToTree(currentNode, { isExpanded: true });
       await this.getChildren(currentNode);
 
       // Walk up the parent nodes until the known root node.
-      if (currentNode === ThreadFront.getKnownRootDOMNode()) {
+      if (currentNode === ThreadFront.getKnownRootDOMNode() || !this.tree) {
         break;
       }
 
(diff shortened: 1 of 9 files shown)


src/devtools/client/webconsole/webconsole.js
```diff
@@ -229,25 +229,26 @@ class WebConsole {
...
-    const onNodeFront = this.toolbox.target
-      .getFront("inspector")
-      .then(inspectorFront => inspectorFront.getNodeFrontFromNodeGrip(grip));
+    const onNodeFront = valueFront
+      .getPause()
+      .ensureDOMFrontAndParents(valueFront._object.objectId)
+      .then(async nf => {
+        await nf.ensureParentsLoaded();
+        return nf;
+      });
```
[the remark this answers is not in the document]
hbenl: This isn't just a style thing: my code returns `Promise<NodeFront>` but yours returns `Promise<undefined[]>`.

And yes, this is independent of all the other stuff in this PR (and also applies to the old markup view), just tell me if you want me to move it into a separate PR.

src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -219,13 +241,13 @@ class MarkupView {
...
     while (currentNode) {
-      if (this.nodes.has(currentNode.objectId())) {
+      if (!this.nodes || this.nodes.has(currentNode.objectId())) {
         // We can stop importing once we encounter a node that is already known.
         break;
       }
 
       this.nodes.set(currentNode.objectId(), currentNode);
-      await this.addNodeToTree(currentNode, { isExpanded: true });
+      this.addNodeToTree(currentNode, { isExpanded: true });
```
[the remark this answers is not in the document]
hbenl: `addNodeToTree(...)` is synchronous, so no need to `await` it. It was previously marked as `async` but I removed that because it doesn't `await` anything.

src/devtools/client/webconsole/webconsole.js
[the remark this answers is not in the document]
jasonLaster: @hbenl i think mine does as well :) 

I think this would be better in its own PR for tracking purposes. The markup view is big so itd be easy to get confused in the future

src/devtools/client/inspector/markup/new-markup.js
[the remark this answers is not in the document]
jasonLaster: Nice

src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -97,6 +101,15 @@ class MarkupView {
     return this._eventTooltip;
   }
 
+  /**
+   * An object representing the markup tree. The key to the object represents the object
+   * ID of a NodeFront of a given node. The value of each item in the object contains
+   * an object representing the properties of the given node.
+   */
+  getTree() {
```
jasonLaster, line 109: is this used? I checked because it seems a bit fishy to be accessing the state directly, but in the right circumstances it could be fine

src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -168,8 +183,7 @@ class MarkupView {
    */
   async expandNode(nodeId) {
     await this.getChildren(this.nodes.get(nodeId));
-    this.tree[nodeId].isExpanded = true;
-    this.store.dispatch(updateTree(this.tree));
+    this.store?.dispatch(updateNodeExpanded(nodeId, true));
```
jasonLaster, line 186: when would store not be defined?

jasonLaster (commented):
Looking good.

jasonLaster:
I saw some exceptions in setupSidebar from addRule and queueTab so i assume you're still ironing it out.

src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -219,13 +241,13 @@ class MarkupView {
     // starting from the selected node itself since we want to display it expanded with
     // its children visible.
     while (currentNode) {
-      if (this.nodes.has(currentNode.objectId())) {
+      if (!this.nodes || this.nodes.has(currentNode.objectId())) {
```
gabrielluong, line 244: Probably can simplify to use optional chaining with `this.nodes?`

src/devtools/client/inspector/markup/new-markup.js
hbenl: I used it initially, but now it isn't used anymore.

src/devtools/client/inspector/markup/new-markup.js
hbenl: If `destroy()` was called during the `await` in the previous line.

src/devtools/client/inspector/markup/new-markup.js
hbenl: No, `this.nodes?.has(...)` would return `undefined` (which is falsy) if `this.nodes` is undefined, but we need it to return `true` (or something truthy) in that case.

hbenl:
I've split this into #839 and #840.

PR #834, opened 2020-10-19
Toggle privacy

jaril:
Re-do of #829.

The old patch broke the tests because it blocked the test runner from opening the test recording.

The main problem it that new test recordings are not being inserted as new rows in the Hasura recordings table. When the viewer tries to fetch the test recording from Hasura for authorization purposes, Hasura responds with an empty array. We then infer from the empty array that the user does not have permission to view the recording, and we display a "You need to sign in to view this recording" error instead.

This patch works around this problem by skipping the authorization step if we're in a test.

### DIFF
--- a/src/ui/components/DevTools.js
+++ b/src/ui/components/DevTools.js
@@ -69,7 +69,7 @@ function getUploadingMessage(uploading) {
 function getIsAuthorized({ data, error, isAuthenticated }) {
   // We let Hasura decide whether or not the user can view a recording. The response to our query
   // will have a recording if they're authorized to view the recording, and will be empty if not.
-  return data.recordings.length > 0 ? true : false;
+  return data.recordings.length;
 }
 
 function DevTools({
--- a/src/ui/components/DevTools.js
+++ b/src/ui/components/DevTools.js
@@ -26,9 +26,6 @@ const GET_RECORDING = gql`
       title
       recordingTitle
       is_private
-      user {
-        auth_id
-      }
     }
   }
 `;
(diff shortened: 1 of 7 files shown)


src/ui/components/Header/ShareDropdown.js
```diff
@@ -104,46 +130,47 @@ function Collaborators({ recordingId, setExpanded, setSharingModal }) {
...
+          setSharingModal={setSharingModal}
+        />
+      ) : null}
+    </>
+  );
+}
 
+function ShareDropdown({ recordingId, setSharingModal }) {
+  const [expanded, setExpanded] = useState(false);
+  const isOwner = getIsOwner(recordingId);
```
jasonLaster, line 160: should this also be a hook since it uses hooks?  e.g. `useIsOwner`

src/ui/components/DevTools.js
```diff
@@ -65,16 +64,19 @@ function getUploadingMessage(uploading) {
...
 
 function getIsAuthorized({ data, error, isAuthenticated }) {
-  if (!features.private) {
+  const test = new URL(window.location.href).searchParams.get("test");
+
+  // Ideally, test recordings should be inserted into Hasura. However, test recordings are currently
+  // not being inserted as a Hasura recordings row, so the GET_RECORDING query will respond with an
+  // empty recordings array. To temporarily work around this for now, we return `true` here so
+  // the test can proceed.
+  if (test) {
```
jasonLaster, line 73: Good call. My bet is that Hasura does not accept a new recording without a user_id. We can probably get around this by having tests use a saved Test account...

jasonLaster (approved):
Wow. Great job!

PR #849, opened 2020-10-20
Bootstrap the Debugger store when the application bootstraps

jaril:


src/devtools/client/debugger/packages/devtools-reps/src/object-inspector/actions.js
```diff
@@ -166,6 +159,7 @@ async function releaseActors(state, client, dispatch) {
 
 function invokeGetter(node, receiverId) {
   return async ({ dispatch, client, getState }) => {
+    return
```
jaril, line 162: Is this supposed to be a bailout?

jasonLaster: yeah - i don't want to think about that now :)

PR #859, opened 2020-10-22
fix a couple of bugs from the console cleanup

hbenl:


src/devtools/client/webconsole/components/Output/GripMessageBody.js
```diff
@@ -55,10 +55,11 @@ function GripMessageBody(props) {
       transformEmptyString: true,
       escapeWhitespace,
       style: styleObject,
-      dispatch,
     });
   }
 
+  Object.assign(objectInspectorProps, { dispatch });
```
jasonLaster, line 61: dispatch could be added directly to the object above

```
@@ -47,6 +47,7 @@ function GripMessageBody(props) {
     autoExpandDepth: shouldAutoExpandObjectInspector(props) ? 1 : 0,
     mode,
     maybeScrollToBottom,
+    dispatch
   };
```

jasonLaster (approved):
One small nit, otherwise great. And thanks for the help!

PR #861, opened 2020-10-22
Move message events out of hud and into redux

jasonLaster:


src/protocol/utils.ts
```diff
@@ -131,8 +131,8 @@ export const EventEmitter = {
       );
     };
 
-    obj.emit = (name, value) => {
-      (obj.eventListeners.get(name) || []).forEach(handler => handler(value));
+    obj.emit = (name, ...values) => {
+      (obj.eventListeners.get(name) || []).forEach(handler => handler(...values));
```
jaril, line 135: Is this going to be okay? Won't we have to change all handler function paramaters to expect an array because of the spread operator change?

jaril, line 135: Is this going to be okay? Won't we have to change all handler function paramaters to expect an array because of the spread operator change?

jaril (commented):
Also needs a quick `npm run lint --fix` :)

src/devtools/client/webconsole/actions/autocomplete.js
```diff
@@ -31,7 +31,7 @@ function autocompleteUpdate(force, getterPath, expressionVars) {
     }
 
     const inputValue = hud.getInputValue();
-    const frameActorId = await hud.getFrameActor();
+    const frameActorId = await toolbox.getPanel("debugger")?.getFrameId();
```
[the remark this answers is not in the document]
jasonLaster: I think so... we don't always start up the debugger... I don't think.

src/protocol/utils.ts
jasonLaster: yep, if values is [a,b,c], this will lead to `handler(a,b,c)`

PR #862, opened 2020-10-23
fix debugger thunks

hbenl:
This fixes a bug introduced in 827a5a0b9b8426628d654a1bc4f1dd5d0215bc75


jasonLaster (approved):
Thanks. We can clean this up after @gabrielluong extracts the toolbox panels from the Toolbox component.

PR #864, opened 2020-10-23
Inspector improvements

hbenl:


src/devtools/client/debugger/panel.js
```diff
@@ -101,18 +101,22 @@ export class DebuggerPanel {
     this.toolbox.selectTool("inspector");
   }
 
-  async openElementInInspector(gripOrFront) {
+  async openElementInInspector(valueFront) {
     const onSelectInspector = this.toolbox.selectTool("inspector");
-    const onGripNodeToFront = getNodeFront(gripOrFront, this.toolbox);
 
-    const [front, inspector] = await Promise.all([onGripNodeToFront, onSelectInspector]);
+    const onNodeFront = valueFront
```
jasonLaster, line 107: could we simplify this further to be 

```
this.toolbox.openNodeInInspector(valueFront, 'debugger');
```

jasonLaster (approved):

src/devtools/client/debugger/panel.js
hbenl: No, because that method doesn't exist there. `this.toolbox` is an instance of the class defined in `src/ui/components/Toolbox.js`, you may be thinking of the function in `src/devtools/client/webconsole/actions/toolbox.js`, but that is a thunk and I *think* it is in a different store (with our redux setup I never know what ends up where).

hbenl:
I merged the master branch into this branch because the fixes can't be tested without the debugger thunks fix.

PR #865, opened 2020-10-23
New markup improvements

hbenl:


src/devtools/client/inspector/markup/components/TextNode.js
```diff
@@ -7,14 +7,14 @@ const { getStr, getFormatStr } = require("../utils/l10n");
...
-      type: PropTypes.string.isRequired,
+      type: PropTypes.number.isRequired,
       value: PropTypes.string.isRequired,
     };
   }
 
   render() {
     const { type, value } = this.props;
-    const isComment = type === "comment";
+    const isComment = type === 8;
```
jasonLaster, line 17: lets use NODE.COMMENT

https://github.com/jasonLaster/devtools/blob/master/src/devtools/shared/dom-node-constants.js#L14

src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -326,7 +326,7 @@ class MarkupView {
       this.expandNode(nodeId);
     }
 
-    this.store.dispatch(updateSelectedNode(nodeId));
+    this.selection.setNodeFront(this.nodes.get(nodeId));
```
jasonLaster, line 329: why do we not need to update the redux store?

src/devtools/client/inspector/markup/components/TextNode.js
hbenl: Yes!

src/devtools/client/inspector/markup/new-markup.js
hbenl: The selection will send a `new-node-front` event, which results in a call to `this.update()`, which will update the redux store.

jasonLaster (approved):

src/devtools/client/inspector/markup/new-markup.js
jasonLaster: Nice

PR #874, opened 2020-10-26
Use seeking for "continue to here" feature

jaril:
This changes the implementation for `continueToHere`.

Instead of adding a hidden breakpoint on the line that the user selects and using rewind/resume to hit it, we seek to the first point on that line instead. This way, it doesn't end up pausing on other breakpoints on the way to the target line.

https://www.loom.com/share/8dbbb4f616d94656ad28bcb803f86527


jasonLaster:
Lets get rid of hidden breakpoints entirely. We can always get a breakpoints hits with an Analysis call, which will be more useful.

https://gist.github.com/jasonLaster/b9a798131b8d1505cc0e53debb9776e7

PR #892, opened 2020-10-28
Move breakpoints panel to the right sidebar

jaril:
This patch gets the ball rolling for #891. 

It shows the breakpoints pane as a sidebar item so users can open the panel even if they don't have the debugger panel one.

This also goes a step further and changes the name from Breakpoints to Logpoints. 

### Screenshots
image
image
image


jaril:
Follow ups should be:
- Remove breakpoint pane from the debugger panel
- Design and style the logpoint pane
- Expand the logpoint pane whenever the user adds a logpoint
- Allow the user to edit logpoint messages from the logpoint pane
- Highlight the timeline and console messages when the user hovers on a logpoint in the logpoint pane

PR #896, opened 2020-10-29
typescriptify protocol/thread

hbenl:


src/protocol/thread/pause.js
```diff
@@ -1,4 +1,4 @@
-const { sendMessage } = require("../socket");
+const { client } = require("../socket");
 const { defer, assert } = require("../utils");
 const { ValueFront } = require("./value");
```
jasonLaster, line 3: is there any reason not to import these as well?

jasonLaster (approved):

src/protocol/thread/pause.js
hbenl: I just removed the unused import: the calls to `sendMessage()` were replaced by the corresponding client calls...

PR #900, opened 2020-10-30
move the location check from mapStateToProps() to the appropriate render function (#826)

hbenl:
This is a weird issue: the `ConditionalPanel` component is only used after checking that `conditionalPanelLocation` is defined (see [here](https://github.com/RecordReplay/devtools/blob/1cb548ba48f840874e33687a582a6b49b9391bb0/src/devtools/client/debugger/src/components/Editor/index.js#L529)). So how is it possible that it then throws an error because the location is undefined? My guess is that this happens because the check occurs in `mapStateToProps()`, which may be called more often than expected. So I moved the check to the `renderToWidget()` method, where the location is needed.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/ConditionalPanel.js
+++ b/src/devtools/client/debugger/src/components/Editor/ConditionalPanel.js
@@ -120,6 +120,10 @@ export class ConditionalPanel extends PureComponent {
     }
     const { location, editor } = props;
 
+    if (!location) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
@jaril could you take a look at this? My hunch is that we're getting rid of this panel.

jaril:
Yeah – I'm currently working on a patch that gets rid of the conditional panel entirely so this shouldn't be an issue after today.

Issue #901, opened 2020-10-30
TypeError: can't access property "querySelector", consoleOutput is undefined


hbenl:
This error message appears when clicking on a marker in the timeline.

PR #904, opened 2020-10-30
fix clicking on a marker in the timeline (#901)

hbenl:


src/devtools/client/webconsole/webconsole.js
```diff
@@ -56,6 +56,8 @@ export class WebConsole {
   }
 
   async init() {
+    this.outputNode = this.document.getElementById("toolbox-content-console");
```
jasonLaster, line 59: why do we need `this.document` as opposed to `document`

jasonLaster (approved):
Nice catch

PR #925, opened 2020-11-03
Add syntax highlighting to breakpoint panel editor

jaril:


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
```diff
@@ -0,0 +1,190 @@
+import React, { useEffect, useState, PureComponent } from "react";
+import classnames from "classnames";
+import { connect } from "devtools/client/debugger/src/utils/connect";
+import { getContext } from "devtools/client/debugger/src/selectors";
+import { PanelInput } from "./PanelInput";
+import actions from "devtools/client/debugger/src/actions";
+
+class PanelEditor extends PureComponent {
```
jaril, line 8: Used the editor name here instead since it's more appropriate.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelInput.js
```diff
@@ -0,0 +1,83 @@
+import React, { useRef, useEffect } from "react";
+import ReactDOM from "react-dom";
+import CodeMirror from "codemirror";
+
+export function PanelInput({
```
jaril, line 5: Changed the component name to PanelInput since the original name, BreakpointEditor, was confusing. 

The name made sense at first since this generates a codeMirror "editor" that allows for syntax highlighting. But it didn't make sense since there can be multiple editors (logValue, condition) for a single breakpoint.

Also, it looks like an input element, so I was in favor of just pretending it's an input.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
```diff
@@ -0,0 +1,92 @@
...
+  }, []);
+
+  if (!node) {
+    return null;
+  }
+
+  return ReactDOM.createPortal(<>{children}</>, node);
+}
+
+function Panel({ breakpoint, editor }) {
```
jaril, line 63: Refactored the Widget and PanelSummary, but kept it in the same file since 1) they're not that big, and 2) they're not being used elsewhere.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
jaril: This is intentionally kept as a class component, because using a functional component with the current implementation of PanelInput doesn't work out very well. See [stale props inside functional components](https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function).

We only assign the event listeners (e.g. `handleKeyDown`) to PanelInput when it mounts. The problem with functional components is that a callback like `handleKeyDown` “sees” the props and state from the render it was created in. In this case, the `handleKeyDown` callback attached to codeMirror when it mounts is never updated, and will keep using the old state/props as if they were never updated. 

For example, editing a logValue from "Hello" to "Hello World" and pressing the Enter key to save the new log message will call `onKeyDown` and `handleSave`. But both will end up referring to stale state, where the logValue is still "Hello". The logValue then ends up being saved as "Hello".

This problem simply goes away if we just use a class component instead.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
```diff
@@ -0,0 +1,190 @@
...
+      // logValue and conditionValue until after the component mounts.
+      return null;
+    }
+
+    return (
+      <div
+        className={classnames("panel-editor", { conditional: hasCondition })}
+        ref={node => (this.panelEditorNode = node)}
+      >
+        {this.renderHeader()}
```
jaril, line 181: I tried splitting out the Form and Header into their own components, but there was so much plumbing that it wasn't worth it.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
```diff
@@ -0,0 +1,190 @@
...
+        )}
+      </div>
+    );
+  }
+
+  render() {
+    const { conditionValue, logValue } = this.state;
+    const hasCondition = conditionValue !== null;
+
+    if (!logValue) {
```
jaril, line 170: Not sure about this. The other way to do it is with [componentWillMount](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) but it doesn't seem advisable.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
```diff
@@ -0,0 +1,92 @@
...
+    const _node = document.createElement("div");
+    const editorLine = toEditorLine(location.sourceId, location.line || 0);
+    const _widget = editor.codeMirror.addLineWidget(editorLine, _node, {
+      noHScroll: true,
+    });
+
+    setNode(_node);
+  }, []);
+
+  if (!node) {
```
jaril, line 56: Bail on first render since we haven't added a widget to codeMirror yet,  and we don't have a `node` to use for the portal.

PR #934, opened 2020-11-04
render doctype nodes correctly

hbenl:
The "unknown" node often shown at the top of the markup view was a doctype node that wasn't rendered correctly (and not the document node as I had guessed previously).

### DIFF
--- a/src/protocol/thread/node.ts
+++ b/src/protocol/thread/node.ts
@@ -106,7 +106,7 @@ export class NodeFront {
   }
 
   get doctypeString() {
-    return `<!DOCTYPE ${this.nodeName}>`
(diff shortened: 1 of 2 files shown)


gabrielluong (approved):

src/protocol/thread/node.ts
```diff
@@ -106,8 +106,7 @@ export class NodeFront {
   }
 
   get doctypeString() {
-    // NYI
-    return "unknown";
+    return this.nodeName;
```
gabrielluong, line 109: We can probably actually simplify it a bit more by 

```suggestion
    return `<!DOCTYPE ${this.nodeName}>`
```

PR #950, opened 2020-11-05
Ensure split console is closed upon opening a replay

### LINKED ISSUE #729: [Console] The splitconsole is open when a recording is opened
**Describe the bug**
During my user interview with Janelle where she installed Replay, created a new account and created her first recording, I noticed that her first recording opened with the split console. This was unexpected. It may have been user error, but let's keep an eye on this. The same behavior appeared when I asked her to copy and navigate to the same URL in Chrome.

**To Reproduce**
1. Install and open the Replay browser
2. Create a new account
3. Create a new recording
4. Open that recording, check that the split console is not showing

**Expected behavior**
The toolbox should by default show the console.

jaril:
This makes sure that whenever a replay is opened, the splitconsole is toggled off.

Fix #729

### DIFF
--- a/src/ui/utils/prefs.js
+++ b/src/ui/utils/prefs.js
@@ -8,7 +8,7 @@ import Services from "devtools-services";
 const { pref } = Services;
 
 // app prefs.
-pref("devtools.split-console", true);
+pref("devtools.split-console", false);
 pref("devtools.selected-panel", "debugger");
 pref("devtools.user", "{}");
 pref("devtools.recording-id", "");


src/ui/reducers/app.ts
```diff
@@ -10,7 +10,7 @@ function initialAppState(): AppState {
     unexpectedError: null,
     theme: "theme-light",
     // Whether or not the developer tools toolbox is opened.
-    isToolboxOpen: prefs.isToolboxOpen,
+    isToolboxOpen: false,
```
jasonLaster, line 13: hmm, i think it is nice to use the pref here. In general, i'm a fan of persisting our UI settings

jaril: Gotcha!

PR #961, opened 2020-11-05
Keep gutter breakpoint markers from collapsing

### LINKED ISSUE #889: [Debugger] Breakpoint markers can be collapsed
I've noticed occasionally that if breakpoints can shift to the left and not be resized.

I believe the steps to reproduce are

1. set a breakpoint
2. select the console
3. refresh
4. select the debugger

Screen Shot 2020-10-28 at 8 18 07 AM

jaril:
Fix #889. Demo: https://share.descript.com/view/Meb61S6bJBu

This fix makes the breakpoints behave pretty much the same–I tested it up to [6 digit line numbers](https://share.descript.com/view/VlHP44pM81p). 

At 8 or 9 digits it starts going wonky, but that's unrelated. The breakpoint SVG is 60px wide and we can't keep pushing it to the right.

---

Update: It works for code lines up to 7 digits. It definitely won't for 8 digits–then again, if somebody's loading a file that big we've got other problems...
image


jasonLaster:
looks like #965 can land. Lets see if this still happens once we're on latest CM...

PR #967, opened 2020-11-06
Revert "Fix logrocket import"

hbenl:
These imports were written using `require()` on purpose: the imported files have no typings and typescript won't accept that (because it is set to "strict" typechecking), unless you use `require()` which implies that the import has type `any`.
I wonder, however, how it is possible that this wasn't noticed: when I change the imports from `require()` to `import` in VS Code, I immediately get error markers for them. The typescript watcher (which is automatically started when you start the webpack-dev-server task in VS Code) also complains and when you include this in a PR, the typecheck workflow also complains. Yet all these checks were apparently ignored, which begs the question if we need to make them "louder"?
Screenshot_2020-11-06_10-35-14

### DIFF
--- a/src/ui/actions/metadata.ts
+++ b/src/ui/actions/metadata.ts
@@ -3,10 +3,10 @@ import { Action } from "redux";
 import { selectors } from "ui/reducers";
 import { UIStore, UIThunkAction } from ".";
 import { Comment } from "ui/state/metadata";
-import { prefs, features } from "ui/utils/prefs";
+const { prefs, features } = require("ui/utils/prefs");
 import { seek } from "./timeline";
 import { User } from "ui/state/metadata";
-import LogRocket from "ui/utils/logrocket";
+const LogRocket = require("ui/utils/logrocket");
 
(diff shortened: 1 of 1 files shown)


src/ui/actions/metadata.ts
```diff
@@ -3,10 +3,10 @@ import { Action } from "redux";
 import { selectors } from "ui/reducers";
 import { UIStore, UIThunkAction } from ".";
 import { Comment } from "ui/state/metadata";
-import { prefs, features } from "ui/utils/prefs";
+const { prefs, features } = require("ui/utils/prefs");
 import { seek } from "./timeline";
 import { User } from "ui/state/metadata";
-import LogRocket from "ui/utils/logrocket";
```
jasonLaster, line 9: the impetus for using import was that Logrocket has an `export default` we could either tack a default at the end `require("ui/utils/logrocket").default;` or add a type definition

hbenl: OK, got it. I've added `default` to the end of the `require`.

PR #980, opened 2020-11-09
Show breakpoint panel for empty log expressions

jasonLaster:
Also adds check for codeMirror focusing

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
@@ -8,18 +8,14 @@ import actions from "devtools/client/debugger/src/actions";
 class PanelEditor extends PureComponent {
   panelEditorNode;
 
-  state = {
-    logValue: null,
-    conditionValue: null,
-  };
+  constructor(props) {
+    const { breakpoint } = props;
+    super(props);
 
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
```diff
@@ -169,12 +165,6 @@ class PanelEditor extends PureComponent {
     const { conditionValue, logValue } = this.state;
     const hasCondition = conditionValue !== null;
 
-    if (!logValue) {
-      // Skip the first render. The state doesn't have the default values for
-      // logValue and conditionValue until after the component mounts.
-      return null;
-    }
```
jasonLaster, line 176: this was the problem, but i think we can get around it by setting the state in the constructor...

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelInput.js
```diff
@@ -94,9 +94,9 @@ export function PanelInput({
 
   useEffect(() => {
     if (autofocus) {
-      setTimeout(() => codeMirrorNode.current.focus(), 0);
```
jasonLaster, line 97: guards against an error i was seeing in the console

jaril (approved):

PR #988, opened 2020-11-09
Handle hasura errors when fetching recordings

### LINKED ISSUE #881: TypeError: can't access property "recordings", t is undefined
Sentry Issue: [REPLAY-83](https://sentry.io/organizations/replay/issues/1929550472/?referrer=github_integration)

```
TypeError: can't access property "recordings", t is undefined
  at AccountPage (./src/ui/components/Account/index.js:70:12)
...
(9 additional frame(s) were not displayed)
```

jaril:
Fix #881


jaril:
Thought about it again and I don't know what kind of error could be causing this bug. We could just handle the error, but then we still wouldn't know what it is and what's causing it in the first place.

I've never seen it before so I want to just throw an error when it happens, so that we'll know exactly what's happening when it hits Sentry.

jasonLaster (approved):

PR #997, opened 2020-11-11
Fix gutter clicks on space occupied by a widget

### LINKED ISSUE #996: Clicking on the area under a breakpoint re-adds the breakpoint
When we add a breakpoint, the editor's gutter is expanded to make space for the breakpoint panel. Clicking on the extra space appears resets the breakpoint's options to the default log message. I think we might be re-adding the breakpoint on that gutter click.

Before adding a breakpoint:
```
| 1|
| 2|
| 3|
| 4|
```

After adding a breakpoint:
```
| 1|
| 2| <-- Breakpoint added to line 2
|  | <-- Clicking here re-adds breakpoint to line 2
|  | <-- Clicking here re-adds breakpoint to line 2
| 3|
| 4|
```

jaril:
Fix #996.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/index.js
+++ b/src/devtools/client/debugger/src/components/Editor/index.js
@@ -13,7 +13,6 @@ import classnames from "classnames";
 import { debounce } from "lodash";
 
 import { isFirefox } from "devtools-environment";
-import { getLineText } from "./../../utils/source";
 import { features } from "../../utils/prefs";
 import { getIndentation } from "../../utils/indentation";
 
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/actions/breakpoints/index.js
```diff
@@ -198,6 +198,13 @@ export function addBreakpointAtLine(cx, line, shouldLog = false, disabled = fals
     if (!source) {
       return;
     }
+
+    const breakpointExists = getBreakpointAtLocation(state, { line, column: undefined });
```
jasonLaster, line 202: this seems counterintuitive and might need a comment describing why we do it.

jasonLaster (approved):

PR #999, opened 2020-11-11
Move vendored libs to npm

jasonLaster:


jasonLaster:
NOTE: whatwg-url 8.2.0 switches to a webidl-conversions lib that uses SharedArrayBuffer, which is not currently supported in the Replay browser

PR #1007, opened 2020-11-13
Timewarp to an existing pause

hbenl:


src/protocol/logpoint.ts
```diff
@@ -66,7 +66,7 @@ client.Analysis.addAnalysisResultListener(({ analysisId, results }) => {
     results.forEach(
       async ({ key, value: { time, pauseId, location, values, data, frameworkListeners } }) => {
         const pause = new Pause(ThreadFront.sessionId!);
-        pause.instantiate(pauseId);
+        pause.instantiate(pauseId, key, time, true);
```
bhackett1024, line 69: Can this either have a comment like `pause.instantiate(pauseId, key, time, /* hasFrames */ true)` or pass an object literal to improve readability?

bhackett1024 (approved):
Looks great!

src/protocol/logpoint.ts
hbenl: Done

PR #1008, opened 2020-11-13
Refactor inspector, part 2

hbenl:


src/devtools/client/inspector/markup/new-markup.js
```diff
@@ -64,6 +70,44 @@ class MarkupView {
...
+    if (this.selection.nodeFront) {
+      this.store.dispatch(selectionChanged(this.selection, false));
+    } else {
+      const defaultNode = await rootNode.querySelector("body");
+      await defaultNode?.ensureLoaded();
+      if (defaultNode && !this.selection.nodeFront && ThreadFront.currentPause === pause) {
+        this.selection.setNodeFront(defaultNode, { reason: "navigateaway" });
+      }
+    }
+  }
```
jasonLaster, line 106: We don't need to do this now, but it would be nice if this business logic were in a redux action. The reason for that, is that it helps us keep state out of the code and encapsulate some of this business logic. 

see https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/webconsole/actions/messages.js#L30-L44

src/ui/actions/timeline.ts
```diff
@@ -174,8 +174,18 @@ export function setZoomRegion(region: ZoomRegion): SetZoomRegionAction {
...
+  point: ExecutionPoint,
+  time: number,
+  hasFrames: boolean,
+  pauseId?: PauseId
+): UIThunkAction {
   return () => {
-    ThreadFront.timeWarp(point, time, hasFrames);
+    const pause = pauseId !== undefined ? Pause.getById(pauseId) : undefined;
+    if (pause) {
+      ThreadFront.timeWarpToPause(pause);
```
jasonLaster, line 186: hmm, we seek to a lot of different types of places... Checking in advance if we have a pause first might be a bit cumbersome, but i can see why you need it now and we can revisit this over time.

src/ui/actions/timeline.ts
```diff
@@ -174,8 +174,18 @@ export function setZoomRegion(region: ZoomRegion): SetZoomRegionAction {
...
+    } else {
+      ThreadFront.timeWarp(point, time, hasFrames);
```
jasonLaster, line 188: now that we're sharing redux actions, we can probably get rid of the other timeWarpy actions

jasonLaster (approved):
Looking good. Feel free to land when the tests are happy

gabrielluong (approved):
Mostly reviewing the inspector changes. I think we can probably go with removing the commented out code next.

PR #1012, opened 2020-11-13
New dashboard

jaril:
Things needed in a follow up:
- Add batch actions in grid view.
- Add sorting with other columns (privacy, duration).
- Add search.
- Align columns better since the current implementation's columns break once resized to a width <~1200.
- Look into not hiding some columns if the viewport width is narrow.
- Merge Title and Page title (maybe).

image
image
image
image
image
image


jasonLaster (commented):
Couple of minor things, but then this is ready to go

1. scrolling in grid view feels broken
2. the grid view card dropdown is missing (probably related)
3. the bulk actions dropdown needs some more padding

PR #1016, opened 2020-11-14
Highlight messages on hover

jasonLaster:
this repurposes the same highlight effect we use now when hovering on an individual message in the console. My hunch is that is a bit much, but we can always tone it down


src/ui/components/Timeline/Message.js
```diff
@@ -97,7 +102,8 @@ class Message extends React.Component {
       }) >
       timelineMarkerWidth / 2;
 
-    const isHighlighted = highlightedMessage == message.id;
+    const isHighlighted =
```
jaril, line 105: This can be confusing. We probably should add a comment explaining why we need to have a fallback in case `highlightedMessage != message.id`.

jaril: Also small nit, but we should probably name it `highlightedMessageId` to be more specific?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Breakpoint.js
```diff
@@ -42,12 +41,24 @@ class Breakpoint extends PureComponent {
...
 
     return bp;
   }
 
+  onMouseEnter = event => {
+    this.props.highlightLocation(this.props.breakpoint.location);
+  };
+
+  onMouseLeave = () => {
+    console.log("onmouseleave");
```
jaril, line 58: log

jaril (approved):
Just had two things but looks good!

Also, if anything, the highlighted state blends in and is actually hard to notice right away.

PR #1029, opened 2020-11-17
finish typescriptifying protocol/thread

hbenl:


src/protocol/thread/thread.ts
```diff
@@ -673,7 +778,9 @@ export const ThreadFront = {
       // Determine the kind of this source, or its minified version.
       let kind = info.kind;
       if (kind == "prettyPrinted") {
-        const minifiedInfo = this.sources.get(info.generatedSourceIds[0]);
+        const minifiedInfo = info.generatedSourceIds
+          ? this.sources.get(info.generatedSourceIds[0])
+          : undefined;
```
jasonLaster, line 783: we do `const minifiedInfo = this.sources.get(info.generatedSourceIds![0]);` in another place. Is there any reason not to do that here too?

jasonLaster (approved):
looks good assuming the tests are green, which they look like they eventually will be.

Screen Shot 2020-11-17 at 8 03 51 PM

src/protocol/thread/thread.ts
hbenl: No, there's no reason, I now changed that place to the safer version as well. Good catch!

PR #1030, opened 2020-11-17
Markup view: scroll the selected node into view

hbenl:
...except if it was selected in the markup view itself

### DIFF
--- a/src/devtools/client/inspector/markup/actions/markup.js
+++ b/src/devtools/client/inspector/markup/actions/markup.js
@@ -1,7 +1,14 @@
 import { assert } from "protocol/utils";
 import { ThreadFront } from "protocol/thread";
 import { DOCUMENT_TYPE_NODE } from "devtools/shared/dom-node-constants";
-import { RESET, NEW_ROOT, ADD_CHILDREN, UPDATE_NODE_EXPANDED, UPDATE_SELECTED_NODE } from "./index";
+import {
+  RESET,
+  NEW_ROOT,
+  ADD_CHILDREN,
+  UPDATE_NODE_EXPANDED,
+  UPDATE_SELECTED_NODE,
+  UPDATE_SCROLL_INTO_VIEW_NODE,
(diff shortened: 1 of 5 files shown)


src/devtools/client/inspector/markup/actions/markup.js
```diff
@@ -106,9 +123,17 @@ export function selectionChanged(selection, expandSelectedNode) {
 
     // expand each ancestor, loading its children if necessary
     for (const ancestor of ancestors) {
+      if (shouldScrollIntoView) {
+        dispatch(scrollIntoView(ancestor.objectId()));
+      }
```
gabrielluong, line 128: I am kinda curious if we need this since we are scrolling to the selectedNode at the end.

gabrielluong (approved):

src/devtools/client/inspector/markup/components/Node.js
```diff
@@ -174,6 +180,7 @@ class Node extends PureComponent {
           role: "treeitem",
           onMouseEnter: this.onMouseEnter,
           onMouseLeave: this.onMouseLeave,
+          ref: shouldScrollIntoView ? this.scrollIntoView : undefined,
```
gabrielluong, line 183: This is a neat usage of `ref`.

hbenl: Thanks. One interesting aspect is that the callback will only get called once when it is first set, on subsequent renders it will not get called - and that is important, because otherwise, it would keep scrolling to this node on every render...

src/devtools/client/inspector/markup/actions/markup.js
hbenl: When selecting a node with the node picker, it can take some time until all ancestors are loaded and expanded - in that case it's nice to scroll the node that is currently loading into view. That's why I added `scrollIntoViewNode` because otherwise we could indeed just use `selectedNode`.

PR #1034, opened 2020-11-17
Test runner: support multiple patterns

hbenl:
The test runner accepts multiple `--pattern` arguments, but tests had to match all of the given patterns to be run.
With this patch, tests only need to match one of the given patterns to be run (my guess is that this was the intention all along).


jasonLaster:
Nice

jasonLaster (approved):

PR #1047, opened 2020-11-19
Fix missing L10N property

jasonLaster:


jasonLaster:
Fixes a crash we were seeing in sentry

PR #1049, opened 2020-11-19
Setup thread event listeners before the app renders

### LINKED ISSUE #1046: [Console] rewind buttons are not shown in the console
STR:
1. go to http://localhost:8080/view?id=cd4d7b22-da98-430e-9923-dc963fa8a0c7
2. select the console
3. hover on a console message

It looks like when we load the replay, the pausedExecutionPoint is missing. It comes back when you click around in the timeline

jaril:
Fix #1046

### DIFF
--- a/src/devtools/client/debugger/panel.js
+++ b/src/devtools/client/debugger/panel.js
@@ -2,8 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-import { defer, assert } from "protocol/utils";
-import { resizeBreakpointGutter } from "./src/utils/ui";
+import { assert } from "protocol/utils";
 import { openDocLink } from "devtools/client/shared/link";
 import { onConnect } from "devtools/client/debugger/src/client";
 import { getCodeMirror } from "devtools/client/debugger/src/utils/editor";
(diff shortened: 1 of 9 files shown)


src/protocol/thread/thread.ts
```diff
@@ -173,6 +173,7 @@ class _ThreadFront {
     const { endpoint } = await client.Session.getEndpoint({}, sessionId);
 
     // Make sure the debugger has added a pause listener before warping to the endpoint.
+    // can we get rid of this? -jaril
     await gToolbox.startPanel("debugger");
```
jaril, line 177: There are no pause event listeners being set up by the debugger panel.

jaril: Moved it to `devtools-toolbox.js` so we start the default panels immediately.

jasonLaster (approved):

src/devtools/client/debugger/panel.js
```diff
@@ -25,41 +24,6 @@ export class DebuggerPanel {
     return this;
   }
 
-  refresh() {
```
jaril, line 28: Moved this to the `Devtools` component to avoid race conditions where we refresh CodeMirror before the `DevTools` component is expanded on the screen.

src/ui/utils/devtools-toolbox.js
```diff
@@ -65,7 +65,6 @@ export class DevToolsToolbox {
     await panel.open();
 
     this.panels[name] = panel;
-    store.dispatch(actions.setSelectedPanel(name));
```
jaril, line 62: We don't need to open up the panel every time we start it. Since we front load opening the console and the debugger, this would cause the toolbox to flicker very quickly before setlling on the actual selected panel. (console -> debugger -> selectedPanel)

src/devtools/client/debugger/src/components/App.js
```diff
@@ -114,6 +114,17 @@ class Debugger extends Component {
     shortcuts.off("Escape", this.onEscape);
   }
 
+  componentDidUpdate(prevProps) {
+    const { selectedPanel, refreshCodeMirror } = this.props;
+
+    // Only refresh CodeMirror when moving from a non-debugger panel to the debugger panel. Otherwise,
+    // the gutter will keep errantly resizing between refreshes.
+    if (selectedPanel == "debugger" && prevProps.selectedPanel != selectedPanel) {
+      console.log(actions);
```
jasonLaster, line 123: ^

jasonLaster (approved):

PR #1052, opened 2020-11-20
Improve markup view performance

hbenl:
The markup view used to render the entire tree whenever anything in the markup state changed.
With this patch, every markup component receives only the information it needs from the markup state, so that only the nodes that have actually changed are rendered.
Furthermore, when a new node is selected, its ancestors are only scrolled into view when their children still need to be downloaded and added to the tree.

### DIFF
--- a/src/devtools/client/inspector/markup/components/MarkupApp.js
+++ b/src/devtools/client/inspector/markup/components/MarkupApp.js
@@ -3,6 +3,8 @@ const dom = require("react-dom-factories");
 const PropTypes = require("prop-types");
 const { connect } = require("react-redux");
 
+const { getNode, getRootNodeId } = require("../reducers/markup");
+
 const Node = createFactory(require("./Node"));
 
 const Types = require("../types");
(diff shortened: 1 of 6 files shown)


src/devtools/client/inspector/markup/components/MarkupApp.js
```diff
@@ -3,25 +3,28 @@ const dom = require("react-dom-factories");
...
 class MarkupApp extends PureComponent {
   static get propTypes() {
     return {
-      markup: PropTypes.shape(Types.markup).isRequired,
+      node: PropTypes.shape(Types.node).isRequired,
       onSelectNode: PropTypes.func.isRequired,
       onShowEventTooltip: PropTypes.func.isRequired,
       onToggleNodeExpanded: PropTypes.func.isRequired,
+      onMouseEnterNode: PropTypes.func.isRequired,
+      onMouseLeaveNode: PropTypes.func.isRequired,
```
jasonLaster, line 20: Minor comment, but i wonder if we want to continue using proptypes? 

I don't think it gets us much when we're using a type system.

jasonLaster (approved):

src/devtools/client/inspector/markup/components/MarkupApp.js
hbenl: Yeah, I usually leave them in to check calls from javascript files. But when all callers are already converted to Typescript, the proptypes probably don't add any value anymore. Furthermore, I noticed that they can actually get in the way: Typescript somehow tries to derive type information about the component from proptypes, but the result didn't work for me and I had to remove the proptypes for Typescript to respect my manual typings.

PR #1054, opened 2020-11-20
Workaround for duplicate applied rules

hbenl:
We sometimes receive duplicate rules from `CSS.getAppliedRules`, as a result React writes a lot of error messages to the console warning about duplicate keys in `RulesApp`.
This patch works around the issue by removing duplicate rules from the result of `CSS.getAppliedRules`. It can be reverted when RecordReplay/backend#384 is resolved.

### DIFF
--- a/src/protocol/thread/node.ts
+++ b/src/protocol/thread/node.ts
@@ -215,7 +215,14 @@ export class NodeFront {
       this._pause
         .sendMessage(client.CSS.getAppliedRules, { node: this._object.objectId })
         .then(({ rules, data }) => {
-          this._rules = rules;
+          this._rules = [];
+          for (const rule of rules) {
+            if (
+              !this._rules.some(r => rule.rule === r.rule && rule.pseudoElement === r.pseudoElement)
+            ) {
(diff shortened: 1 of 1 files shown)


src/protocol/thread/node.ts
```diff
@@ -215,7 +215,14 @@ export class NodeFront {
...
         .then(({ rules, data }) => {
-          this._rules = rules;
+          this._rules = [];
+          for (const rule of rules) {
+            if (
+              !this._rules.some(r => rule.rule === r.rule && rule.pseudoElement === r.pseudoElement)
+            ) {
+              this._rules.push(rule);
+            }
+          }
```
jasonLaster, line 225: could this be simplified with https://lodash.com/docs/4.17.15#uniqBy

```
this._rules = uniqBy(rules, rule => rule.pseudoElement)
```

hbenl: more like `uniqBy(rules, rule => `${rule.rule}|${rule.pseudoElement})`

jasonLaster (approved):

PR #1061, opened 2020-11-24
[WIP] Re-implement the event tooltip

hbenl:
This currently lacks the necessary CSS. One issue is that I can't get the Dropdown component to mask the content behind it when I use it in the markup view. The reason could be a Firefox bug:

Firefox / Replay browser:
Screenshot_2020-11-24_20-00-41

Chrome:
Screenshot_2020-11-24_20-01-05

### DIFF
--- a/src/devtools/client/inspector/markup/actions/eventTooltip.ts
+++ b/src/devtools/client/inspector/markup/actions/eventTooltip.ts
@@ -1,0 +1,64 @@
+import { Action } from "redux";
+import { ThreadFront } from "protocol/thread";
+import { assert } from "protocol/utils";
+import { UIThunkAction } from "ui/actions";
+import { EventInfo } from "../state/eventTooltip";
+
+export type SetEventTooltipAction = Action<"set_event_tooltip"> & {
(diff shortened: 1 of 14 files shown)


hbenl:
I had placed the event badge inside a stacking context that is created for that markup node, so its `z-index` didn't have any effect relative to elements outside of that stacking context (i.e. other markup nodes). It looks like Firefox rendered this right and Chrome has a bug.
So the dropdown masks the markup content now, but here's the next problem: it is rendered inside the markup panel, which is set to `overflow: auto`...
Perhaps we need to change the Dropdown component so that the dropdown container can float above the entire app instead of being confined to this panel - that would require moving the dropdown container somewhere outside (e.g. as a direct child of `document.body`), this could be achieved using a React Portal.
Screenshot_2020-11-25_16-55-49

jasonLaster:
The code looks good. Lets land this and then add support for the portal and perhaps viewport aware logic so it doesn't get clipped by the window.

jasonLaster (approved):

PR #1063, opened 2020-11-25
Disable 3 panel elements toggle

jasonLaster:
Disabling the 3panel toggle because the elements panel will be smaller in the right sidebar


hbenl (approved):

jaril:
The tests don't look right–re-running them! 

image

PR #1064, opened 2020-11-25
sources accordion view

jasonLaster:


jaril (approved):
Small nit: persisting the expand/collapse state of the sources panel worries me. 

It makes it possible for somebody to unintentionally keep their sources collapsed, and think that their sources haven't loaded properly and/or are missing. My preference would be to keep it open by default.

PR #1080, opened 2020-11-26
[Inspector][Rules] Show the Color swatch and linkable URL in the rules view

gabrielluong:


jasonLaster:
This looks great, but it's not clear what to look for. Can you quickly summarize the new functionality?

Also, can you explain what is going on with the output parsers?

src/devtools/client/inspector/rules/components/DeclarationValue.js
```diff
@@ -0,0 +1,56 @@
...
+  require("devtools/client/inspector/rules/components/value/FontFamily")
+);
+const Url = createFactory(require("devtools/client/inspector/rules/components/value/Url"));
+
+class DeclarationValue extends PureComponent {
+  static get propTypes() {
+    return {
+      colorSpanClassName: PropTypes.string,
+      colorSwatchClassName: PropTypes.string,
+      values: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.string, PropTypes.object]))
```
hbenl, line 16: The `propTypes` should also specify `fontFamilySpanClassName`

hbenl (approved):

gabrielluong:
> This looks great, but it's not clear what to look for. Can you quickly summarize the new functionality?
> 
> Also, can you explain what is going on with the output parsers?

The OutputParser is responsible for parsing a CSS declaration (ex: `color: red`) and returning an AST-like array that describes the declaration value (ex: `red`). You can imagine there are many color taking functions and color values can also appear in shorthand properties. So, OutputParser basically parses that and returns token of values that we are interested in. The old OutputParser would return a document fragment with all the color swatches and DOM handling for filter(), shapes, etc directly. I have reduced the OutputParser to outputting an array of objects that describes the token into COLOR, URL, etc along with their value. I opted to exclude some values such as Flex, Grid swatch next to `display: [flex|grid]` because we are missing those functionalities.

gabrielluong:
In the patch queue, you first see me ripping out all the DOM fragments that are being appended to a document fragment that is originally returned and reducing it to just parsing the value and tagging them with COLOR, URI, etc and returning an array at the end. We feed this array of tokens to the React frontend to control how we want to output the declaration value in the Rules view. The end result is that you can see color swatches in the Rules view and links are parsed and can be navigated.

hbenl:
I have rebased this PR [here](https://github.com/hbenl/webreplay-devtools/tree/output-rebased).

hbenl:
Closing in favor of #1114

PR #1082, opened 2020-11-26
Revise icons on left nav

ienjoy:
image

The previous icons were quite thick, so I made them thinner. Here's how they look now:

image


ienjoy:
image

The previous icons were quite thick, so I made them thinner. Here's how they look now:

image

jaril (approved):

PR #1083, opened 2020-11-27
Fix "Not Allowed" messages triggered by LogRocket

hbenl:
- Removes the `RuleFront`s from the actions for the Rules view
- Adds sanitizers to the LogRocket middleware that remove all `*Front` objects from state and actions before they are inspected by LogRocket
- Adds a sanity checking middleware in development mode which logs `*Front` objects found in state and actions (without flooding the console with log messages)

### DIFF
--- a/src/ui/utils/sanitize.ts
+++ b/src/ui/utils/sanitize.ts
@@ -1,0 +1,54 @@
+import { Pause, ValueFront } from "protocol/thread";
+import { NodeFront } from "protocol/thread/node";
+import { StyleSheetFront } from "protocol/thread/styleSheet";
+import { StyleFront } from "protocol/thread/style";
+import { RuleFront } from "protocol/thread/rule";
+import { NodeBoundsFront } from "protocol/thread/bounds";
+import { Dispatch, Middleware } from "redux";
+import { UIState } from "ui/state";
+import { UIAction } from "ui/actions";
+
+const forbiddenClasses: Record<string, any> = {
+  Pause,
+  ValueFront,
+  NodeFront,
+  StyleSheetFront,
+  StyleFront,
+  RuleFront,
+  NodeBoundsFront,
+};
+
+const loggedCategories = new Set<string>();
+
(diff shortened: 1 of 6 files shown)


src/ui/utils/sanitize.ts
```diff
@@ -0,0 +1,54 @@
...
+  NodeFront,
+  StyleSheetFront,
+  StyleFront,
+  RuleFront,
+  NodeBoundsFront,
+};
+
+const loggedCategories = new Set<string>();
+
+export function sanitize(obj: any, path: string, category: string, logSanitized: boolean): any {
```
jasonLaster, line 23: this is a small nit, but i find it hard to remember what these params do, perhaps this might be cleaner

```js
sanitize(action, { path: "", category: `action[type=${action.type}]`, logSanitized: true });
```

src/ui/utils/sanitize.ts
```diff
@@ -0,0 +1,54 @@
...
+  return obj;
+}
+
+export const sanityCheckMiddleware: Middleware<
+  {},
+  UIState,
+  Dispatch<UIAction>
+> = store => next => action => {
+  sanitize(store.getState(), "", "state", true);
+  sanitize(action, "", `action[type=${action.type}]`, true);
```
jasonLaster, line 52: nit: `action[${action.type}]` is probably fine

jasonLaster (approved):
Looks great.

My only concern is that two inspector tests failed.

src/ui/utils/sanitize.ts
hbenl: One downside with Typescript is that it makes writing a function that takes its parameters as an object a bit cumbersome:
```
function sanitize(obj: any, { path, category, logSanitized }: { path: string; category: string; logSanitized: boolean }) {...}
```
So I tend to avoid it if there aren't many callers, because it doesn't seem worth making the function declaration less readable. After all, Typescript makes it possible to peek at the function declaration with one keystroke, so it's not necessary to make the call more explicit.

hbenl:
All failed tests passed locally.

PR #1086, opened 2020-11-27
Left align the header and move the user avatar to lower left

jaril:
image
image


src/ui/components/Header/Header.js
```diff
@@ -41,9 +42,9 @@ function Avatars({ user, getActiveUsers }) {
 function Links({ user, getActiveUsers, recordingId, setSharingModal }) {
   return (
     <div className="links">
-      <Avatars user={user} getActiveUsers={getActiveUsers} />
+      {/* <Avatars user={user} getActiveUsers={getActiveUsers} /> */}
```
jasonLaster, line 45: what do we want to do with this?

PR #1109, opened 2020-12-01
Add PortalDropdown component and use it in EventTooltip

hbenl:


src/ui/components/shared/PortalDropdown.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { CSSProperties, PureComponent } from "react";
+import { createPortal } from "react-dom";
+import { DropdownProps } from "./Dropdown";
+import "./PortalDropdown.css";
+
+interface PortalDropdownProps extends DropdownProps {
+  distance?: number;
+}
+
+export default class PortalDropdown extends PureComponent<PortalDropdownProps> {
```
jasonLaster, line 10: minor thing here, but this looks like an opportunity for a functional component. We've been using a lot of them in the `src/ui` directory and it has felt cleaner.

hbenl: But where can I store the `buttonRef`? And how do I replace the callbacks without accidentally triggering lots of renders?

src/ui/components/shared/PortalDropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+  private buttonRef = React.createRef<HTMLButtonElement>();
```
jasonLaster, line 11: useRef

jasonLaster (approved):

PR #1110, opened 2020-12-01
implement the "Open in Debugger" button in the new EventTooltip

hbenl:


src/devtools/client/inspector/markup/actions/eventTooltip.ts
```diff
@@ -62,3 +62,9 @@ export function showEventTooltip(nodeId: string): UIThunkAction {
     dispatch(setEventTooltip(nodeId, listenerInfo));
   };
 }
+
+export function viewSourceInDebugger(event: EventInfo): UIThunkAction {
+  return ({ toolbox }) => {
+    toolbox.viewSourceInDebugger(event.url, event.line, event.column, event.sourceId);
+  };
+}
```
jasonLaster, line 70: Looks good. Over time we'll find ways to consolidate some of this logic:

https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/webconsole/actions/toolbox.js#L64

For instance, in the near term we should be able to remove the entire notion of a toolbox object and panel objects: https://github.com/RecordReplay/devtools/issues/1096

jasonLaster (approved):

PR #1121, opened 2020-12-02
Fix the performance of the sanitizer for the LogRocket middleware

hbenl:


src/ui/utils/sanitize.ts
```diff
@@ -43,12 +43,22 @@ export function sanitize(obj: any, path: string, category: string, logSanitized:
...
+}
+
 export const sanityCheckMiddleware: Middleware<
   {},
   UIState,
   Dispatch<UIAction>
 > = store => next => action => {
-  sanitize(store.getState(), "", "state", true);
-  sanitize(action, "", `action[type=${action.type}]`, true);
+  sanitizeState(store.getState(), true);
```
jasonLaster, line 61: I'm worried about sanitizing the state for every action. This feels like redundant work i.e. we just checked the a source in the sources store in the last round, we shouldn't need to do it again.

This is also unnecessary because the fronts would be arriving in the store via an action and we just checked them on the way in.

src/ui/utils/sanitize.ts
```diff
@@ -43,12 +43,22 @@ export function sanitize(obj: any, path: string, category: string, logSanitized:
   return obj;
 }
 
+export function sanitizeState(state: any, logSanitized: boolean) {
+  return sanitize({ ...state, ast: undefined }, "", "state", logSanitized);
+}
+
+export function sanitizeAction(action: any, logSanitized: boolean) {
+  return action.type === "SET_SYMBOLS"
```
jasonLaster, line 51: What's this doing again? 

It seems like this will naively return an empty action for SET_SYMBOLS. Wouldn't we want to do something like

```js
if (EXCLUDED_ACTIONS.include(action.type) {
  return action
}

return sanitize(action, "", `action[${action.type}]`, logSanitized);
```

src/ui/utils/sanitize.ts
hbenl: The `sanitize()` function was primarily written for the prod case where we need to sanitize state and actions before they are processed by LogRocket. Note that all the fronts will remain in the redux state, only the copies of actions and state that we pass to LogRocket are modified.

src/ui/utils/sanitize.ts
hbenl: Yes, I could just return that action - but then LogRocket would immediately complain it is too large and stop processing the redux state and actions.

karina-y:
no change, however it does load quicker. i did get another "whoops something went wrong" message but wasn't able to get that one recorded. i did record a hard reload with emptied cache.. in the minute of typing this actually i just saw the whoops message appear on that window. i wasn't interacting with it, just typing here.
anyway here is a recording with a hard reload and emptied cache
https://www.loom.com/share/b1a47c3cfbd245cda9898e7c2615a319

src/ui/utils/sanitize.ts
jasonLaster: got it - good catch.

Perhaps we minimize the LogRocket middleware because we're not using it too much at the moment:

```js
actionSanitizer: action => ({ type: action.type }
stateSanitizer: state => () => ({})
```

This should help us with perf and then we can slowly add data in that would be useful

src/ui/utils/sanitize.ts
jasonLaster: Thanks, i got confused between the two usecases. Also, didn't realize that our middleware continued to use action, which is not affected.

hbenl:
What's the ID for that recording? And what's the URL for the site you were recording?

src/ui/utils/sanitize.ts
hbenl: I'd like to take one more look at this using Karina's recording but if it keeps impacting our performance, only handing parts of the state and actions to LogRocket sounds like a good idea.

karina-y:
same as before, this is the id and i was on localhost 9e683a8c-c342-43f4-a3d6-cc7e7df111c3

PR #1122, opened 2020-12-02
Persist the selected tab when switching between view modes

jaril:
This PR makes it so that when the user switches between the dev view and non-dev view, their selected editor tab is maintained in the dev view.


karina-y:
looks good to me other than some errors in the console. functionality works as expected.

Issue #1025, opened 2020-11-17
[Console] scroll to evaluation


jasonLaster:
STR:
1. pause at some location
2. scroll to the bottom of the console
3. evaluate an expression

ER: console scrolls to evaluation
AR: nada

PR #1123, opened 2020-12-02
works, but can we pull gtoolbox

karina-y:
issue #1025 in progress
- the scrolling issue when typing something in the console at a breakpoint is fixed however there are 3 things to do still:
1. there is a bug that causes the console to scroll to the bottom after you've done the above then select any new breakpoint
2. removal of gstoolbox use
3. clean up console logs and whitespacing etc


karina-y:
clsoing this one, going to open a new pr with the rest of the fixes

PR #1134, opened 2020-12-03
Convert the redux store and model classes for the rules view to typescript

hbenl:


src/devtools/client/inspector/css-properties.ts
```diff
@@ -7429,38 +7436,43 @@ const FIRST_CHAR = ["[_a-z]", NON_ASCII, ESCAPE].join("|");
...
   return !!input.match(IS_VARIABLE_TOKEN);
 }
 
-module.exports = {
-  supportsType(property, type) {
-    return properties[property] && properties[property].supports.includes(type);
+const CSSProperties = {
+  supportsType(property: string, type: string) {
+    const prop = properties[property];
+    return prop && prop.supports.includes(type);
```
gabrielluong, line 7446: Curious if we could 
```suggestion
    return properties[property]?.supports.includes(type);
```

hbenl: Good idea, see my next commit

jasonLaster (approved):

PR #1136, opened 2020-12-03
Scroll to eval

karina-y:


src/devtools/client/webconsole/components/Output/ConsoleOutput.js
```diff
@@ -69,12 +72,55 @@ class ConsoleOutput extends Component {
     };
   }
 
+  constructor(props) {
+    super(props);
+    this.maybeScrollToBottom = this.maybeScrollToBottom.bind(this);
+  }
```
jasonLaster, line 78: we can drop `this.maybeScrollToBottom.bind` in favor of 

```js
maybeScrollToBottom = () => {
}
```

https://babeljs.io/docs/en/babel-plugin-proposal-class-properties

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
```diff
@@ -69,12 +72,55 @@ class ConsoleOutput extends Component {
...
+    } else {
+      this.maybeScrollToBottom();
+    }
+  }
+
+  maybeScrollToBottom() {
+    if (this.outputNode && this.shouldScrollBottom) {
+      scrollToBottom(this.outputNode);
+    }
+  }
```
jasonLaster, line 122: why does this need to be a function?

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
```diff
@@ -69,12 +72,55 @@ class ConsoleOutput extends Component {
...
       scrollToBottom(this.outputNode);
     }
   }
 
+  componentDidUpdate(prevProps) {
+    const messagesDelta = this.props.messages.size - prevProps.messages.size;
+
+    // When evaluation results are added, scroll to them.
+    this.shouldScrollMessageId = null;
+    this.shouldScrollMessageNode = null;
```
jasonLaster, line 91: could these be local variables to the function?

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
karina-y: good call, maybescroll doesn't even do anything. this.shouldScrollBottom doesn't exist and was something i grabbed from the old code without realizing it wasn't necessary

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
karina-y: yes, something i should've changed after removing gtoolbox

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
jasonLaster:
```js

componentDidUpdate(prevProps) {
  this.maybeScrollToMessage(prevProps);
}

maybeScrollToMessage(prevProps) {
  const messagesDelta = this.props.messages.size - prevProps.messages.size;

  // When evaluation results are added, scroll to them.
  let shouldScrollMessageId;
  let shouldScrollMessageNode;

  if (messagesDelta <= 0 || lastMessage.type != MESSAGE_TYPE.RESULT) {
    return;
  }

    // [...this.props.messages.values()] seems slow
    // we should have a separate messageList somewhere we can check OR
    // use a memoization function to be able to get the last message quickly
    const lastMessage = [...this.props.messages.values()][this.props.messages.size - 1];

    const node = ReactDOM.findDOMNode(this);
    const resultNode = node.querySelector(`div[data-message-id='${lastMessage.id}']`);

  if (!resultNode) {
    return;
  }

  // Scroll to the previous message node if it exists. It should be the
  // input which triggered the evaluation result we're scrolling to.
  const previous = resultNode.previousSibling;
  (previous || resultNode).scrollIntoView();
}
```

jasonLaster (approved):

Issue #1023, opened 2020-11-17
Selecting a frame does not highlight the location in the editor


jasonLaster:
STR:

1. go to https://replay.io/view?id=e5cba9d2-58ec-4bc9-ba02-c7051984d9cf&test=1
2. click on the marker with the comment
3. select other frames in the call stack

https://www.loom.com/share/891df22ac6bb4c748b0e9e4529e9e081

karina-y:
also when selecting call stack, the line should be highlighted in blue, ie
![image](https://user-images.githubusercontent.com/19538307/100946209-b5e0cc00-34b7-11eb-994d-93fb6a0581c9.png)

karina-y:
recording added by @jasonLaster  https://replay.io/view?id=db1e5523-4497-4510-b47b-f0d47124bd4b

PR #1144, opened 2020-12-03
Frame highlight

karina-y:
issue #1023 resolved.
reviewed this with @jaril and confirmed the code that was removed is not being used anywhere else in the codebase and is a remnant of a feature we no longer need https://bugzilla.mozilla.org/show_bug.cgi?id=1580334

### DIFF
--- a/src/devtools/client/debugger/panel.js
+++ b/src/devtools/client/debugger/panel.js
@@ -5,7 +5,6 @@
 import { assert } from "protocol/utils";
 import { openDocLink } from "devtools/client/shared/link";
 import { onConnect } from "devtools/client/debugger/src/client";
-import { getCodeMirror } from "devtools/client/debugger/src/utils/editor";
 
 export class DebuggerPanel {
   constructor(toolbox) {
@@ -36,10 +35,6 @@ export class DebuggerPanel {
 
   _getState() {
     return this._store.getState();
-  }
-
-  getToolboxStore() {
-    return this.toolbox.store;
   }
 
   openLink(url) {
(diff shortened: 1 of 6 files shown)


src/devtools/client/debugger/src/actions/pause/previewPausedLocation.js
```diff
@@ -2,32 +2,8 @@
...
-
-    const sourceLocation = {
-      line: location.line,
-      column: location.column,
-      sourceId: source.id,
-    };
-    dispatch(selectLocation(cx, sourceLocation));
-
-    dispatch({
-      type: "PREVIEW_PAUSED_LOCATION",
```
jasonLaster, line 26: I think there are some additional places where `PREVIEW_PAUSED_LOCATION` is used

src/devtools/client/debugger/src/actions/pause/previewPausedLocation.js
```diff
@@ -2,32 +2,8 @@
...
-      location: sourceLocation,
-    });
-  };
-}
+import { getContext } from "../../selectors";
 
 export function previewPausedLocationBySourceId(location) {
```
jasonLaster, line 8: should we get rid of this too?

src/devtools/client/debugger/panel.js
```diff
@@ -123,10 +123,6 @@ export class DebuggerPanel {
     return this._actions.selectSourceURL(cx, url, { line, column });
   }
 
-  previewPausedLocation(location) {
-    return this._actions.previewPausedLocation(location);
-  }
-
   clearPreviewPausedLocation() {
```
jasonLaster, line 126: should we kill this?

jasonLaster (approved):
Thanks!

src/devtools/client/debugger/src/actions/pause/previewPausedLocation.js
karina-y: i'll have to do some digging, it looks like there were two methods with the same dispatch type

karina-y:
renamed methods to accurately describe what they do and set previewLocation to null once done scrubbing to clean code.

jasonLaster (approved):

jasonLaster (approved):

Issue #1143, opened 2020-12-03
Console evaluation results should always show next to the input


jasonLaster:
It's now possible to get the console evaluation results to show up in the wrong spot.

STR:

1. go to the console and do many evaluations
2. quickly change pause locations

ER: results show up next to inputs
AR: they show up next to recent pause

Loom: https://www.loom.com/share/dca4d08ed6a14de5bba5311cb647c6d8
^ view 0:33

karina-y:
one edge case is left, https://www.loom.com/share/fb4aca38105742d5b17403b8719c2600

if you type something and hit enter really really fast they will end up out of order

PR #1169, opened 2020-12-08
Console eval

karina-y:
fixes issue #1143 
instead of placing the evaluation results at the last pause point, we check if the type is of 'result' and if so it gets placed under its command

### DIFF
--- a/src/devtools/client/webconsole/reducers/messages.js
+++ b/src/devtools/client/webconsole/reducers/messages.js
@@ -1068,14 +1068,8 @@ function getDefaultFiltersCounter() {
 
 //get the point for the corresponding command, regardless of where we're currently paused
 function getPausePoint(newMessage, state) {
-  if (
-    newMessage.type === constants.MESSAGE_TYPE.RESULT &&
-    newMessage.parameters &&
-    newMessage.parameters.length > 0 &&
-    newMessage.parameters[0]._pause &&
-    newMessage.parameters[0]._pause.point
-  ) {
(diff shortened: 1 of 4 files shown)


karina-y:
removed old commits not related to this pr

src/devtools/client/webconsole/reducers/messages.js
```diff
@@ -1066,6 +1066,21 @@ function getDefaultFiltersCounter() {
...
 
+//get the point for the corresponding command, regardless of where we're currently paused
+function getPausePoint(newMessage, state) {
+  if (
+    newMessage.type === constants.MESSAGE_TYPE.RESULT &&
+    newMessage.parameters &&
+    newMessage.parameters.length > 0 &&
+    newMessage.parameters[0]._pause &&
+    newMessage.parameters[0]._pause.point
+  ) {
```
jasonLaster, line 1077: we can clean this up even further with  optional chaining.

```js
newMessage.parameters.[0]?._pause.point
```

we could add more `?` but i don't think we need them...

```js
newMessage?.parameters?.[0]?._pause?.point
```

why not:

* parameters should always be an array
*  if there is a _pause there should be a point

last tweak, it makes me uncomfortable to access a private property `_pause`. Perhaps we should add an accessor

```js
newMessage.parameters.[0]?.getExecutionPoint()
```

jasonLaster (approved):

karina-y:
@jasonLaster i ran both failing tests locally and can't find an error
all i see is this from breakpoints-04

image

Issue #1105, opened 2020-12-01
[DevTools] Better React error handling


jaril:
Currently we mix view logic with error handling. It would be nice if we could separate all the error handling into an [Error Boundary](https://reactjs.org/docs/error-boundaries.html) to keep the code from being illegible.

jasonLaster:
Definitely agree we should have an Error Boundary so we never just go white.  Perhaps at that point we can dispatch the right error action too

PR #1170, opened 2020-12-08
Save session data

jasonLaster:
we might want to also have an on conflict clause to prevent errors for dup sessions, but couldnt figure that out
```
      on_conflict: { constraint: sessions_session_id_key, update_columns: [] }
```

### DIFF
--- a/src/ui/components/Avatar.js
+++ b/src/ui/components/Avatar.js
@@ -1,19 +1,10 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
 import { useAuth0 } from "@auth0/auth0-react";
-
-import { connect } from "react-redux";
-import { actions } from "ui/actions";
 import { getAvatarColor } from "ui/utils/user";
-import { setUserInBrowserPrefs } from "../utils/browser";
 
 const Avatar = props => {
-  let { player, isFirstPlayer, updateUser } = props;
+  let { player, isFirstPlayer } = props;
   let auth = useAuth0();
-
-  useEffect(() => {
-    setUserInBrowserPrefs(auth.user);
-    updateUser(auth.user);
-  }, [auth.user]);
 
   if (auth.isAuthenticated && isFirstPlayer) {
     return (
@@ -31,6 +22,4 @@ const Avatar = props => {
   );
 };
 
-export default connect(null, {
-  updateUser: actions.updateUser,
-})(Avatar);
+export default Avatar;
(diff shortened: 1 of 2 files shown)


src/ui/components/Avatar.js
```diff
@@ -1,20 +1,43 @@
...
   let { player, isFirstPlayer, updateUser } = props;
   let auth = useAuth0();
+  const [AddSession] = useMutation(CREATE_SESSION);
 
   useEffect(() => {
     setUserInBrowserPrefs(auth.user);
     updateUser(auth.user);
   }, [auth.user]);
 
+  useEffect(() => {
```
jaril, line 30: Was there a specific reason for putting this in the Avatar component? I expected it to be higher in the tree, like `<Devtools />` maybe.

src/ui/components/Avatar.js
```diff
@@ -1,20 +1,43 @@
...
+    if (auth.user?.sub && ThreadFront.sessionId) {
+      const object = {
+        id: ThreadFront.sessionId,
+        recording_id: ThreadFront.recordingId,
+        controller_id: ThreadFront.sessionId.split("/")[0],
+      };
+      AddSession({ variables: { object } });
+    }
+  }, [auth.user?.sub]);
```
jaril, line 39: I think we can just subscribe to changes to `auth.user` instead of `auth.user.sub` here, since once we get the user from auth0, their `sub` doesn't change.

jaril (approved):

src/ui/components/Avatar.js
jasonLaster: That's a good call. i was look for auth data

src/ui/components/Avatar.js
jasonLaster: yeah that's fair

src/ui/components/DevTools.js
```diff
@@ -62,12 +73,31 @@ function DevTools({
...
+
+  useEffect(() => {
+    if (auth.user && sessionId) {
+      const object = {
+        id: sessionId,
+        recording_id: recordingId,
+        controller_id: sessionId.split("/")[0],
+      };
+      CreateSession({ variables: { object } });
+    }
```
jaril, line 99: Looks good!

jaril (commented):
Looks great. At some point this week we should probably refactor `<DevTools />` since it's getting crowded and move the error handling elsewhere (see #1105)

PR #1171, opened 2020-12-08
display the shorthand CSS properties in the rules view

hbenl:
This was a lot easier than expected because all the necessary functionality was already there, we just need to use it.


jasonLaster:
This change is great!

The one thing I'd add now is a test to make sure we don't regress


Screen Shot 2020-12-08 at 12 43 05 PM


Screen Shot 2020-12-08 at 12 43 24 PM

gabrielluong (approved):

PR #1193, opened 2020-12-10
Use function name as logpoint expression

### LINKED ISSUE #1009: [Debugger] use function name as a default expression
It'd be nice to experiment with using the function name as a breakpoint prefix

Screen Shot 2020-10-30 at 2 40 11 PM

https://www.loom.com/share/78c572df3d39492b9aecd4475f5fc8f2

chujunlu:
Close #1009

LogValue is using function name as an expression if I add a breakpoint or a column breakpoint within a function.
Screen Shot 2020-12-10 at 8 29 44 AM

Screen Shot 2020-12-10 at 8 29 53 AM

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/index.js
+++ b/src/devtools/client/debugger/src/actions/breakpoints/index.js
@@ -14,7 +14,9 @@ import {
   getSelectedSource,
   getBreakpointAtLocation,
   getBreakpointsForSource,
+  getSymbols,
 } from "../../selectors";
+import { findClosestEnclosedSymbol } from "../../utils/ast";
 import {
   addBreakpoint,
   removeBreakpoint,
@@ -176,6 +178,7 @@ export function addBreakpointAtLine(cx, line, shouldLog = false, disabled = fals
     if (!source) {
       return;
     }
+
     const breakpointLocation = {
       sourceId: source.id,
       sourceUrl: source.url,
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
```diff
@@ -86,26 +86,7 @@ export class Outline extends Component {
...
-    );
-
-    if (enclosedItems.length == 0) {
-      return this.setState({ focusedItem: null });
-    }
-
-    // Find the closest item to the selected location to focus
-    const closestItem = enclosedItems.reduce((item, closest) =>
-      positionAfter(item.location, closest.location) ? item : closest
-    );
```
chujunlu, line 108: Most of this part is doing the same thing as [`findClosestofSymbol`](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/utils/ast.js#L47). I think we can keep only one implementation.

src/devtools/client/debugger/src/actions/breakpoints/index.js
```diff
@@ -151,6 +152,7 @@ export function addBreakpointAtLine(cx, line, shouldLog = false, disabled = fals
     if (!source) {
```
chujunlu, line 152: [Outline](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js#L322) component uses [`getSelectedSourceWithContent`](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/reducers/sources.js#L544) to retrieve the selected source. I wonder should we use the same thing instead of [`getSelectedSource`](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/reducers/sources.js#L532)?

jasonLaster: i am not sure if we need the source content here.

PR #1197, opened 2020-12-10
Add loading param + screen

jasonLaster:


src/ui/components/DevTools.js
```diff
@@ -109,7 +110,7 @@ function DevTools({
     return null;
   }
 
-  if (queryIsLoading) {
+  if (isLoading() || queryIsLoading) {
```
loganfsmyth, line 113: If the whole URL is `/view?loading=true`, I don't think we'll be rendering the `DevTools` component at all, so I think this needs to be handled in `App.js`?

PR #1204, opened 2020-12-11
Show inline styles inherited from the selected node's ancestors

hbenl:


gabrielluong:
@hbenl Do you have a replay to demonstrate inline styles?

hbenl:
@gabrielluong go to https://replay.io/view?id=87e54827-f477-4632-bc3e-e30839b20dd7 and select the child of the second div with `class="parent"`.

jasonLaster (approved):

hbenl:
New recording id for testing: 746a315d-631c-44cd-beeb-69386fbb3741

PR #1205, opened 2020-12-11
Reimplement the computed view

hbenl:


hbenl:
Note that the test I have added will only work when #1204 has also landed.

jasonLaster:
Can we preemptively pull out some components from ComputedApp? 

https://gist.github.com/jasonLaster/ac3bc3d7be1905f041cc22ec95ed7262

src/devtools/client/inspector/computed/panel.ts
```diff
@@ -1,15 +1,17 @@
...
 export class ComputedPanel {
   private tool: any;
   provider?: ReactElement;
 
-  constructor(inspector: any) {
+  constructor(inspector: Inspector) {
     this.tool = new ComputedViewTool(inspector, window);
-    this.provider = createElement(InspectorTabPanel, { id: "sidebar-panel-computedview" });
+    // this.provider = createElement(InspectorTabPanel, { id: "sidebar-panel-computedview" });
+    this.provider = createElement(ComputedApp);
```
jasonLaster, line 14: this is super weird. Lets prioritize adding these panels to the react tree and removing these tool classes `ComputedViewTool`.

jasonLaster: https://github.com/RecordReplay/devtools/issues/1210

src/devtools/client/inspector/computed/reducers/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (expanded) {
+      if (expandedProperties.indexOf(property) < 0) {
+        expandedProperties.push(property);
+      }
+    } else {
+      const index = expandedProperties.indexOf(property);
+      if (index >= 0) {
+        expandedProperties.splice(index, 1);
+      }
+    }
```
jasonLaster, line 37: should expanded properties be a Set?

jasonLaster (commented):
The app looks good, just a couple of opinionated comments

jasonLaster (approved):

src/devtools/client/inspector/computed/reducers/index.ts
hbenl: I have extracted the the components from `ComputedApp` and switched `expandedProperties` to `Set<string>`, as suggested. 
The main changes compared to your gist:
- I put the extracted components in separate files
- I don't connect `ComputedApp` to the redux store, only to pass down the stuff I get from the store to the new components. Instead, I connect those components to the store directly.
- I put the "No CSS properties found" div inside the `ComputedProperties` component
- instead of hiding the "No CSS properties found" div and the list of properties, you remove them from the DOM. I decided to stick to hiding them because it should be slightly better for performance and I didn't see any benefits from removing instead of hiding them.

Issue #1129, opened 2020-12-02
need to check messages exist before accessing them


karina-y:
**Describe the bug**
when clearing all bugs from console, error appears due to no breakpoints existing

**To Reproduce**
1. Go to a recording
2. add breakpoints
3. remove all the breakpoints
4. see error appear in console

**Screenshots**
![image](https://user-images.githubusercontent.com/19538307/100943940-10c3f480-34b3-11eb-936d-76243e6ac431.png)

![image](https://user-images.githubusercontent.com/19538307/100943951-19b4c600-34b3-11eb-979c-04c5b43a8c5f.png)

PR #1208, opened 2020-12-11
checking messages exist

karina-y:
fixes issue #1129 
when removing all console messages we'd get an error that none exist, now checking they exist before running code related to messages


jasonLaster (approved):

karina-y:
after rebase:
- breakpoints-02 & breakpoints-06 time out in ci and locally
- stepping-01 times out in ci, passes locally before declaring it failed

@jasonLaster safe to ship?

PR #1220, opened 2020-12-12
Jon's messed up color patch -- to ignore

jonbell-lot23:
1. We were using different shades of blue so I standardised on one
2. Incorporated primary and secondary accent colors
3. Removed the white stroke on the event circles


jonbell-lot23:
Confirmed -- I messed up the event circles: 
image
image

So we should chat about that @jaril

jasonLaster:
accent colors: I think the primary colors look good. I wonder if the purple token highlight could pop more... If we make it more pronounced would we need to switch to white text on purple background?  If so, we might need to change our approach to highlighting and use codemirror's tokenizer...

Play buttons: i'm a bit nervous about the size of the play buttons given their simple geometries. I wonder if we might want to put them in a circular button? Refactoring UI discusses the principle in #5 ([link](https://medium.com/refactoring-ui/7-practical-tips-for-cheating-at-design-40c736799886)).

jonbell-lot23:
This PR had too much going on, so check here instead: https://github.com/RecordReplay/devtools/pull/1228

PR #1225, opened 2020-12-14
checking if we should scroll in console panel by whether or not node is in view now, not by node type

karina-y:


karina-y:
tests:
stepping-06 times out in ci and locally
console_warp-01 times out in ci and passes locally
logpoint-04 times out in ci and locally

karina-y:
is the eval being appended to the top of the console?

karina-y:
@jasonLaster just tested this branch and evals aren't being appended to the top. i also went back to the loom and if you look closely you can see the scrollbar move as well, i was just already close to the top on that recording.

karina-y:
latest test update:
inspector-03 times out in ci but passes locally on master and eval-scroll

src/devtools/client/webconsole/components/Output/ConsoleOutput.js
```diff
@@ -101,6 +101,15 @@ class ConsoleOutput extends Component {
       return;
     }
 
+    //check if node is already in view
```
jaril, line 104: _Very_ minor nit - I'd be more descriptive here with something like:
```js
// Don't scroll to the evaluation result if it's already in view.
```

jaril (approved):
Just had one small nit but apart from that this looks good! There's one failing test but it appears to pass locally. Let's ship it!

PR #1228, opened 2020-12-14
Adding primary and secondary accent colours

jonbell-lot23:
I'm standardising the UI a bit more to prepare for accessibility/dark mode/personalisation in the future. With these changes, there should be a single place to update the whole UI.

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.css
@@ -94,3 +94,10 @@
   stroke: var(--replaying-paused-marker-stroke);
 }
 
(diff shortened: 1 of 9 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.css
```diff
@@ -94,10 +94,3 @@
   stroke: var(--replaying-paused-marker-stroke);
 }
 
-.breakpoint-navigation-timeline-pause-marker {
```
jasonLaster, line 97: Hmm, this seems to delete the pause line, which we should probably evaluate in the context of another PR

jonbell-lot23: True. Fixed, thanks!

PR #1233, opened 2020-12-15
Show number of hits when the user hovers on a line number

jaril:
image

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.js
+++ b/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.js
@@ -17,23 +17,18 @@ export function StaticTooltip({ targetNode, children }) {
   );
 }
 
-export function LineNumberTooltip({ editor, cx, runAnalysisOnLine, analysisPoints }) {
+export function LineNumberTooltip({
+  editor,
+  cx,
+  runAnalysisOnLine,
+  analysisPoints,
+  setHoveredLineNumberLocation,
+}) {
   const [lineNumberNode, setLineNumberNode] = useState(null);
 
-  const setHoveredLineNumber = event => {
-    let target = event.target;
-    const isBreakpointMarkerNode = target.closest(".new-breakpoint");
-
-    // If hovered on a breakpoint marker, get the corresponding linenumber element.
-    if (isBreakpointMarkerNode) {
-      const gutterNode = target.closest(".Codemirror-gutter-elt");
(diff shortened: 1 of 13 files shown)


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.js
```diff
@@ -0,0 +1,65 @@
...
+
+  const setHoveredLineNumber = event => {
+    let target = event.target;
+    const isBreakpointMarkerNode = target.closest(".new-breakpoint");
+
+    // If hovered on a breakpoint marker, get the corresponding linenumber element.
+    if (isBreakpointMarkerNode) {
+      const gutterNode = target.closest(".Codemirror-gutter-elt");
+      target = gutterNode?.previousElementSibling;
+    }
```
jasonLaster, line 31: could we move this logic to gutter-events too so we just emit the line number? This is cute, but I don't think we want to think about other gutter markers in this api, just the gutter el and line


```
const setHoveredLineNumber = (line, target) => {
    runAnalysisOnLine(cx, line);
    setLineNumberNode(target);
}
```

src/protocol/logpoint.ts
```diff
@@ -251,11 +263,12 @@ export function setLogpointByURL(
   line: number,
   column: number,
   text: string,
-  condition: string
+  condition: string,
+  showInConsole: boolean = true
```
jasonLaster, line 267: i wonder if `setLogpoint` should take an object? these long lists of args gets sketchy

```
logGroupId, mapper, null, true
```

src/protocol/logpoint.ts
```diff
@@ -300,7 +313,7 @@ function eventLogpointMapper(getFrameworkListeners: boolean) {
 
 export async function setEventLogpoint(logGroupId: string, eventTypes: string[]) {
   const mapper = eventLogpointMapper(/* getFrameworkListeners */ true);
-  const analysisId = await createLogpointAnalysis(logGroupId, mapper, null);
+  const analysisId = await createLogpointAnalysis(logGroupId, mapper, null, true);
```
jasonLaster, line 316: should condition be null? probably, but we need to update the type..

jasonLaster (commented):
couple comments. otherwise great

src/protocol/logpoint.ts
jaril: did you mean the logpoint parameter?

PR #1234, opened 2020-12-15
Remove breakpoint line

jonbell-lot23:
Minor tweak to remove the vertical line from the breakpoint panel.

Split out from a larger PR yesterday.


jonbell-lot23:
I now disagree with this PR

PR #1238, opened 2020-12-15
Add keyboard shortcuts to the markup panel

hbenl:


jasonLaster:
the keyboard shortcuts should not move the panel as well

https://www.loom.com/share/9ae1508a756a409994340d150dccc795

jasonLaster (approved):
One small ux detail, but otherwise great

PR #1239, opened 2020-12-15
checking elem exists on timeframe check before accessing it

karina-y:


src/test/harness.js
```diff
@@ -374,7 +374,7 @@ async function executeInConsole(value) {
 function waitForFrameTimeline(width) {
   return waitUntil(() => {
     const elem = document.querySelector(".frame-timeline-progress");
-    return elem.style?.width == width;
+    return elem?.style?.width == width;
```
jasonLaster, line 377: looks like you only need one `?`

`elem?.style.width == width`

karina-y: fixed!

jasonLaster (approved):

PR #1245, opened 2020-12-15
scrolling node into view when revealing in tree and its scrolled out of view

karina-y:
changed the container we were using to check if a node was within view, to the '.controlled' div as it's the only one who's height respects the scrollable space's height. also removed logic to snap the node to the top or bottom and is now scrolling happily centered.

### DIFF
--- a/src/devtools/client/debugger/packages/devtools-components/src/tree.js
+++ b/src/devtools/client/debugger/packages/devtools-components/src/tree.js
@@ -753,17 +753,16 @@ class Tree extends Component {
           return closestScrolledParent(node.parentNode);
         };
 
-        const scrolledParentRect = treeElement?.getBoundingClientRect();
-        const isVisible =
-          !treeElement || (top >= scrolledParentRect.top && bottom <= scrolledParentRect.bottom);
+        //this is the one that contains the scrollable contents without expanding its height past what is visible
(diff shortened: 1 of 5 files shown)


src/devtools/client/debugger/packages/devtools-components/src/tree.js
```diff
@@ -753,17 +753,15 @@ class Tree extends Component {
...
-        const isVisible =
-          !treeElement || (top >= scrolledParentRect.top && bottom <= scrolledParentRect.bottom);
-
-        if (!isVisible) {
-          const { alignTo } = options;
-          const scrollToTop = alignTo
-            ? alignTo === "top"
-            : !scrolledParentRect || top < scrolledParentRect.top;
-          element.scrollIntoView(scrollToTop);
+        const parentElem = document.querySelector(".debugger .controlled");
```
jasonLaster, line 756: `.debugger .controlled` might not be the parent...

karina-y:
notes: make the sources and outline panels' contents scrollable

karina-y:
failing ci tests run locally:
breakpoints-04 fails in master and local, says video format is no longer supported
logpoint-03 fails in master and local, says video format is no longer supported
console_warp-02 times out in local branch and master
object_preview-03 times out in local branch and master

PR #1247, opened 2020-12-15
Opening a source link from a console message should be reflected in the source tree

### LINKED ISSUE #1181: clicking on a source file link in console doesn't navigate to it in source tree
**Describe the bug**
when clicking on a source file from the console panel, the source tree panel doesn't navigate to the file. instead it highlights the parent-most open directory it lives in.

**To Reproduce**
1. Go to any recording with breakpoints
2. Click on link in console to source file
3. look at source tree panel, notice nothing happens but a highlight

**Recording**
https://www.loom.com/share/20fd7fa2ec26457e8e35eee53bcf34ea

jaril:
Fix #1181.

This updates the store, specifically the `shownSource` value, so that the SourceTree component is responsive to the user opening a new source.

### DIFF
--- a/src/devtools/client/debugger/panel.js
+++ b/src/devtools/client/debugger/panel.js
@@ -117,6 +117,7 @@ export class DebuggerPanel {
     const cx = this._selectors.getContext(this._getState());
     const location = { sourceId, line, column };
 
+    this._actions.showSource(cx, sourceId);
     await this._actions.selectSource(cx, sourceId, location);
   }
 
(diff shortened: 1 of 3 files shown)


karina-y (commented):
only works if the sources panel is expanded and in view. if the sources panel is collapsed it doesn't get expanded and if the explorer panel isn't selected to be in view, it doesn't open up. this pr has a related fix for that when selecting 'reveal in tree' https://github.com/RecordReplay/devtools/pull/1240

jasonLaster:
Love it!

jasonLaster (approved):

PR #1248, opened 2020-12-16
Playback background pattern

jonbell-lot23:
I designed this a few weeks ago when exploring how to make the UI pop a little more in a subtle way. Old version on the left, new version on the right:

image

Here's how it looks in the product:

image


jasonLaster:
Two thoughts:
1. the background feels a bit scheumorphic 
2. it feels very prominent

This was the error UI i saw in the netlify deploy when the image didn't load and I kinda liked the soft background. The primary reason is that it doesn't pop that much.  

Screen Shot 2020-12-15 at 6 53 54 PM

jonbell-lot23:
1. Skeu is back in! 😂
2. I'm trying to design for this moment:

image

The first time someone loads Replay, they're getting a big black rectangle, which looks broken. We used to use a light grey background, which looks broken/kind of cheap. I think moving to black was a good call, because it sends a strong message that this is a video player, like YouTube or Netflix:

image

But because we don't have a nicely defined video region the way YouTube does, we need a bit of contrast with the flat black. So here's how I see each:

1. Black is a good start but doesn't frame the video region well enough
2. Dark grey is ok, but video players don't really do that
3. The light grey feels even further away from video, and cheap
4. A pattern/texture signals a bit of bling -- even if it's not this precise png, I think a subtle pattern feels more premium

jasonLaster:
Lets give it a shot. I guess youtube also shows static sometimes 😉

PR #1252, opened 2020-12-16
Remove the debug label from the console overlay buttons

### LINKED ISSUE #1188: [Console] Revise the purple "debug" button
When you're paused on a console messsage, we show a purple bookmark to the left to indicate that you're paused there. Hovering on that bookmark shows the "Debug" label and clicking on it opens up that source in the editor.

That was designed back when we had a separate console and debugger panel. Now, pressing the button doesn't feel so much like going to "debug" mode. Instead, it feels more like an "open in editor" button. 

We should change the label from "Debug" to something like "Show in editor"/"Open in editor".

image
image

jaril:
Fix #1188

image

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -977,7 +977,6 @@ a.learn-more-link.webconsole-learn-more-link {
   z-index: 1;
   border-top-right-radius: var(--border-radius);
   border-bottom-right-radius: var(--border-radius);
-  transition-duration: var(--slide-in-duration);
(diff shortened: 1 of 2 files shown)


jasonLaster:
Screen Shot 2020-12-15 at 6 49 48 PM

Looks good except we should keep the purple hover effect

jaril:
Should we? It's no longer a button and now only serves as a visual indicator. I took it out so it doesn't look like you can interact with it, only to press it and think that it might be broken.

jasonLaster:
re-added the selectLocation behavior too

PR #1256, opened 2020-12-16
Make sure only one console message has the paused marker on it

### LINKED ISSUE #1217: Duplicate pause buttons
**Describe the bug**
It looks like when a logpoint and an event listener are at the same location, we show two pasued badges

**To Reproduce**
1. go to https://replay.io/view?id=4a2c1694-dd87-4e86-9df3-93292f19bfa8
2. select click events
3. add a logpoint at the same spot

https://replay.io/view?id=d1122315-33ae-4f69-ac2e-419a99791859

Screen Shot 2020-12-11 at 3 50 21 PM

jaril:
Fix #1217.

This makes sure that even if there are multiple logs (e.g. breakpoint + event breakpoint) that is associated with a particular execution point, we only show one pause marker in the console message. This is different from what we have now, where we would show a pause marker for each of those console messages if we were paused at that point.

This gets the job done but it's a half measure. Ideally, if a user is paused at a point where there are multiple messages, it should still be obvious that there are multiple messages associated with that pause. 

```
Scenario: Paused on Pause group 1

Before:
⏸ Event breakpoint message (Pause group 1)
⏸ Breakpoint message (Pause group 1)
   Breakpoint message (Pause group 2)
   Breakpoint message (Pause group 3)

After:
⏸ Event breakpoint message (Pause group 1)
   Breakpoint message (Pause group 1)
   Breakpoint message (Pause group 2)
   Breakpoint message (Pause group 3)
```

In the `After` case, the breakpoint message in pause group 1 is no longer visually associated with the current pause. This makes it easy to mistake it for being in a different pause group altogether.

image

### DIFF
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.js
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.js
@@ -117,6 +117,23 @@ class ConsoleOutput extends Component {
     (previous || resultNode).scrollIntoView();
   }
 
+  getIsFirstMessageForPoint(index, visibleMessages) {
+    const { messages } = this.props;
+
+    if (index == 0) {
+      return true;
+    }
+
+    let previousMessage = messages.get(visibleMessages[index - 1]);
+    let currentMessage = messages.get(visibleMessages[index]);
+
+    if (!previousMessage || !currentMessage) {
+      return false;
+    }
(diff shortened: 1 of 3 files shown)


jasonLaster (approved):
Nice!

PR #1268, opened 2020-12-16
Start the timeline from the beginning upon open

jaril:


jasonLaster:
So, the trade-off here is that the initial screen by definition will be black because every app begins in a navigation state.

I wonder if the best option would be to show the last graphic and have the timeline / play button at the beginning. A feature we could add on top is letting users set a thumbnail. This is common in video tools too :)

jonbell-lot23:
I think this is a good way forward for now. Bring the playhead to the beginning (maybe 1 second in, because that means the timeline has a little blue and mostly grey) and use the last frame.

In a perfect world, that black screen would be clipped out. It's not useful for users, it's just a technical thing we haven't gotten around to fixing just yet. But we can do that later.

Broader playability notes here: https://www.notion.so/replayio/Playability-Rationale-Doc-781baff5abda475589ebdc8eb499ca27

jasonLaster:
lets land this once we have support for showing the last graphic.

jasonLaster:
lets revisit this once we have navigation landed. at that point we can start from a better spot. 

this code is super stale too

PR #1269, opened 2020-12-16
Add playback autoplay

jaril:
This adds an autoplay when a user opens a replay and behaves as follows:
- The autoplay playback starts immediately after the replay is loaded and ends when it hits the end
- Toggling between the player/debug view doesn't stop the autoplay playback – it continues where it left off in the previous view
- Pressing pause during the autoplay playback stops the playback
- Once the autoplay playback is stopped, toggling between player/debug view behaves as it did before


jonbell-lot23:
I think this is a good feature, and I have some nuanced things to say on the topic, but let's hold off for now. The other playability items help us move things forward and I'm happy to focus on them first.

Playability write-up here: https://www.notion.so/replayio/Playability-Rationale-Doc-781baff5abda475589ebdc8eb499ca27

jasonLaster:
closing as we focus on other playability wins

PR #1271, opened 2020-12-16
Show timeline hover preview on playback for Play mode

jaril:
This PR changes the way we show video previews while hovering on the timeline in Play mode. 

Instead of showing the preview within a tooltip just above the timeline, this now shows the preview on the actual playback screen itself. As soon as the user moves their cursor away from the timeline, the playback screen resets to whatever the current paused point is.

I immediately noticed a few things that we haven't thought through:
1. This sucks for "long hover" cases, where the user might start at 0% and hover slowly all the way through to 100%. With the preview being shown on the playback screen, the user now has to look away from the timeline to see what changes are happening in the preview. This makes it much harder to hover in a straight line from 0% -> 100%. In my experience, my cursor kept going too far up/down.
2. It's not obvious when the screen being shown in the playback is the a) current paused screen or b) the preview screen. This makes item no.1 even worse, because there's no immediate feedback that you've stopped hovering if your mouse drifts away from the timeline. It might obvious if the current paused screen is very different from the preview screen, but apart from that, it becomes confusing very quickly.

We can dismiss 1 as a concern if we are confident that people won't do the "long hover" cases. This isn't that unreasonable of an assumption given that people can just press the play button to do the same thing. But if we think or we presume that people will still do long hovers, then we should rethink whether we want to add this preview feature.

I'd balance that decision with this feature's obvious benefit for "short hover" cases, where you're trying to pinpoint where a specific happened. For those cases, having the preview screen much larger than the tooltip is a very noticeable improvement.

If we do proceed with it, we should definitely do something about 2 and visually indicate that what they're looking at is a preview. Perhaps something like a small popup at the top+center of the screen that says "Showing preview at 6s" that remains there while the user is hovering on the timeline.

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.css
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.css
@@ -2,6 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
+.outlines-pane {
+  overflow: auto;
+  display: flex;
+  flex-direction: column;
+  min-height: 25px;
+}
+
 .outline {
   overflow-y: hidden;
 }
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
@@ -17,9 +17,9 @@
 
 .sources-pane {
   display: flex;
-  flex: 1;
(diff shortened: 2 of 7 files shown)


jonbell-lot23:
Pulling forward something I said in Slack: I recommend doing the other Playability bugs and holding off on this. Will help us see what we think about this with other variables controlled for.

jonbell-lot23:
Broader playability write-up and plan here: https://www.notion.so/replayio/Playability-Rationale-Doc-781baff5abda475589ebdc8eb499ca27

PR #1283, opened 2020-12-18
Show recordings shared with you

jasonLaster:


src/ui/components/Account/index.js
```diff
@@ -56,6 +77,20 @@ function FirstRecordingPrompt() {
   );
 }
 
+function getRecordings(data) {
+  if (!data?.users.length) {
```
jaril, line 81: Do we need the optional chaining here? We check for `error`/`loading` in the AccountPage component, so by the time we pass the data here it _should_ be truthy.

jaril (approved):
Look good!

src/ui/components/Account/index.js
jasonLaster: I guess not...

PR #1295, opened 2020-12-18
Make tabs scrollable horizontally

jaril:
Chrome:
image
Firefox:
image

https://share.descript.com/view/KDseaBY6Q7A

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Tab.js
+++ b/src/devtools/client/debugger/src/components/Editor/Tab.js
@@ -16,7 +16,6 @@ import { copyToTheClipboard } from "../../utils/clipboard";
 import actions from "../../actions";
 
 import {
-  getDisplayPath,
   getFileURL,
   getRawSourceURL,
   getSourceQueryString,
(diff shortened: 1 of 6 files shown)


karina-y (commented):
looks like the tabs are showing the pathname now as well instead of just the source file name

image

jaril:
It appears that the same thing happens without the patch. 

@jasonLaster is there a reason for keeping the path there? We can either 1) leave it alone, 2) manage the overflow and limit the tab to some n pixels, or 3) get rid of the path altogether.

image

jasonLaster (approved):

jasonLaster:
So we show pathnames to disambiguate tabs... The goal is only show them when we have two conflicting tabs with the same filename. I'd be happy getting rid of it though and just showing paths on hover...

karina-y (approved):

jaril:
Two failing tests right before being merged: 
-`inspector-05` is intermittent and passes locally.
-`console_warp-02` is flakey, see #1298 

image

PR #1302, opened 2020-12-21
WIP: Feature/display selector

camacho:
This PR adds a new toolbar with an inspector to the non-dev view.

The majority of the work completed so far is around a refactor to how the `NodePicker` works. Originally, setting a tool and setting a panel were two separate calls. For the functionality of the new dev-view `NodePicker`, we wanted the view to change when a node is clicked, and we wanted the active panel to be the inspector panel.

To create this effect, we moved the assumption that `gTool.selectTool` would dispatch a Redux action that would result in both the `panel` and `viewMode` values in the store being set at the same time.

Todo:
- [ ] get the selected node from the non-dev view to be selected in the dev view
- [ ] properly handle the behavior of using the non-dev view `NodePicker` twice (events seem to not be fired)
- [ ] cleanup CSS/styling
- [ ] tests (and test fixes)

### DIFF
--- a/src/ui/components/NodePicker.js
+++ b/src/ui/components/NodePicker.js
@@ -91,7 +91,15 @@ export default class NodePicker extends React.Component {
   };
 
   nodePickerMouseClick = e => {
-    this.nodePickerMouseClickInCanvas(this.mouseEventCanvasPosition(e));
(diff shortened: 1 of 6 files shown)


camacho:
cc/ @jaril @jasonLaster

PR #1303, opened 2020-12-21
Improvements to the computed view

hbenl:
- fix dom nesting (fixes a warning from react)
- don't show CSS variables as computed properties
- improve performance

### DIFF
--- a/src/devtools/client/inspector/computed/actions/index.ts
+++ b/src/devtools/client/inspector/computed/actions/index.ts
@@ -46,6 +46,8 @@ function createComputedProperties(elementStyle: ElementStyle): ComputedPropertyS
 
   const properties: ComputedPropertyState[] = [];
   for (const [name, value] of computed) {
+    if (name.startsWith("--")) continue;
+
     let inheritanceCounter = 1;
     const selectors: MatchedSelectorState[] = [];
     for (const rule of elementStyle.rules) {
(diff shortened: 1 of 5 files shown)


src/devtools/client/inspector/computed/actions/index.ts
```diff
@@ -46,6 +46,8 @@ function createComputedProperties(elementStyle: ElementStyle): ComputedPropertyS
 
   const properties: ComputedPropertyState[] = [];
   for (const [name, value] of computed) {
+    if (name.startsWith("--")) continue;
```
jasonLaster, line 49: Can you add a comment here?

src/devtools/client/inspector/computed/components/ComputedProperties.tsx
```diff
@@ -33,20 +33,18 @@ function ComputedProperties(props: PropsFromRedux) {
...
+    if (hidden) return null;
+    dark = !dark;
+    allPropertiesHidden = false;
     const isExpanded = expandedProperties.has(property.name);
     const toggleExpanded = () => setComputedPropertyExpanded(property.name, !isExpanded);
 
     return (
       <ComputedProperty
-        key={index}
+        key={property.name}
```
jasonLaster, line 46: is it possible to have duplicate properties? perhaps not for the computed properties

jasonLaster (approved):

src/devtools/client/inspector/computed/components/ComputedProperties.tsx
hbenl: No, the properties are generated from a Map that uses the `name` property as key.

PR #1321, opened 2020-12-22
Display clicks in the non-dev view transcript

jaril:
This adds a "Show clicks" checkbox for Play mode, in place of the old event listener breakpoint dropdown. Toggling the checkbox toggles the visibility of click events as entries in the transcript.

image
image

### DIFF
--- a/src/ui/components/Views/NonDevView.js
+++ b/src/ui/components/Views/NonDevView.js
@@ -15,7 +15,7 @@ import { selectors } from "../../reducers";
 import "./NonDevView.css";
 
 export function EventsFilter() {
-  const [expanded, setExpanded] = useState(false);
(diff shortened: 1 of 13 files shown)


src/ui/components/Views/NonDevView.js
```diff
@@ -33,7 +33,28 @@ export function EventsFilter() {
...
+        id="show-clicks"
+        checked={showClicks}
+        onChange={() => setShowClicks(!showClicks)}
+      />
+    </div>
+  );
+}
+
+function NonDevView({ updateTimelineDimensions, narrowMode, clickEvents }) {
+  const [showClicks, setShowClicks] = useState(true);
```
jaril, line 56: Using redux state to hold this for the first pass. This should be moved to redux in some form for the very next patch that works on this.

PR #1329, opened 2020-12-22
Make the timeline and its command buttons larger

jaril:
This accomplishes the same thing but varies from the original requirements which asked for a bigger Play button and 10px to the bottom. 

The first pass (see below) looked too large (50px vs 64px). I've adjusted it to 56px, and added most of the new height asymmetrically so it's lifted 4px from the bottom of the viewport. Any more than this gets into noticeably asymmetrical territory.

--- | Before | Proposed | Implemented
--- | --- | --- | ---
timeline height | 50px | 60px | 56px
control icon height/width | 20px | 30px | 28px
control button clickable height/width | 20px | 30px | 36px

image
image
image

----------

###  First pass (Old)

--- | Before | Proposed | Implemented
--- | --- | --- | ---
timeline height | 50px | 60px | 64px
control icon height/width | 20px | 30px | 28px
control button clickable height/width | 20px | 30px | 36px

image


jasonLaster (approved):

jasonLaster:
The restart icon feels too large. I wonder if we could try using a filled in button for play and restart to give a perceived size / visability win wihthout going with buttons that are actually this large...

CC @jonbell-lot23 

Screen Shot 2020-12-23 at 2 16 42 PM

comments are no longer anchored :P 
Screen Shot 2020-12-23 at 2 16 39 PM

jonbell-lot23:
After using the large one for a while on a branch, going back to the small version didn't just feel too small. It felt like an error in the design. It's swimming in white space, and only as large as the avatar beside it, despite being the singular most important action on the screen.

Additionally, people in testing often don't notice the timeline or play button.

I would like to make this large in master and try it. As with everything, we can modify further as we learn more. But I don't want to stop small changes like this in GitHub without a chance to test.

Issue #1096, opened 2020-11-30
[Code cleanup] remove devtool-toolbox


jasonLaster:
We have this concept of a "Toolbox" with "Panels" (inspector/console/debugger) in Replay.

`src/devtools/client/debugger/panel.js`

`src/devtools/client/webconsole/panel.js`

`src/devtools/client/inspector/panel.js`

These panels are required for the actual React components for the debugger/inspector/console to work properly. But the logic is handled separately apart from the React lifecycle for mostly legacy reasons (?). This has led to race conditions whenever we change the order in which we initialize the panels relative to each other and the actual React components.

The panels are used primarily in the `devtools-toolbox` (`src/ui/utils/devtools-toolbox.js`). The toolbox is what's responsible for handling/initializing the panels, keeping a record of what's been initialized and what's not, and selecting a panel when one of the panel icons are clicked on.

Ideally this would all be much **simpler**. We should be able to move the react components for the panels around without having to worry about race conditions. 

So it would be nice if we could rely less on panels less. It would be even better if we can completely get rid of them, and move all of the panel initialization logic to their respective React components instead. If we're able to do that, perhaps we could even get rid of `devtools-toolbox` entirely.

karina-y:
issue closed automatically because of a pr merge but there are still two parts to this, reopening.

PR #1332, opened 2020-12-22
Cleanup remove console

karina-y:
fixes part of issue #1096
removes `src/devtools/client/webconsole/panel.js`
toolbox is now being initialized from main.js until we remove toolbox entirely


karina-y:
tests timing out in ci:
breakpoints-02 passes locally
**logpoint-04 times out locally on branch but passes on master**
stepping-02 passes locally
stepping-01 passes locally
object_preview-02 passes locally
breakpoints-03 passes locally
stepping-05 passes locally
breakpoints-01 passes locally
**console_stacks  times out locally on branch but passes on master**

going to dig into the two timing out on master.

karina-y:
ok regarding the latest push: i readded some of the code i removed from devtools-toolbox, initializing the console from within the startPanel rather than the init(). i also removed init from main.js as it was already being called in DevTools.js which is necessary to allow elements to render when the elements tab is already selected on load (ie you select the elements tab, then refresh). i've tested with console, elements, and comments open on refresh and tested each other panel that wasn't open on refresh, all open and load properly.

src/ui/utils/devtools-toolbox.js
```diff
@@ -37,6 +39,13 @@ export class DevToolsToolbox {
     await this.selectTool(selectedPanel);
   }
 
+  initConsole() {
+    initOutputSyntaxHighlighting();
```
jaril, line 43: We should be able to move this to the JSTerm component for once it mounts so it just takes care of it itself.

https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/webconsole/components/Input/JSTerm.js#L54-L57

src/ui/utils/devtools-toolbox.js
```diff
@@ -37,6 +39,13 @@ export class DevToolsToolbox {
...
+
+    // TODO: the store should be associated with the toolbox
+    setupMessages(store);
```
jaril, line 46: Is there a reason why we can't just move `setupMessages` to the end of main.js? 

https://github.com/RecordReplay/devtools/blob/master/src/main.js#L81

src/ui/utils/devtools-toolbox.js
jaril: Also in the same file, line 30-32 we can get rid of this code since don't need to start the console panel anymore.

```
    // The console has to be started immediately on init so that messages appear
    // on the timeline.
    await this.startPanel("console");
```

src/devtools/client/debugger/panel.js
```diff
@@ -42,8 +42,7 @@ export class DebuggerPanel {
   }
 
   async openConsoleAndEvaluate(input) {
-    const consolePanel = await this.toolbox.selectTool("console");
-    consolePanel.evaluateExpression(input);
+    this._actions.evaluateExpression(input);
```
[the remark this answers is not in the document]
karina-y: i didn't touch anything else in debugger because i'm doing each panel in separate pr's, i only touched those two lines because it was necessary to remove webconsole panel

jaril (approved):

karina-y:
stepping-05 times out in ci but passes locally

PR #1348, opened 2020-12-23
WIP: Show Syntax errors in the breakpoints panel

jasonLaster:


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
```diff
@@ -127,6 +127,10 @@
   align-items: center;
 }
 
+.breakpoint-panel .form-row.invalid .CodeMirror {
+  border-color: #dc75c5;
```
jaril, line 131: If it's an error let's stick with some standard error color. The console ones are nice and not firetruck red.

https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/themes/webconsole.css#L31-L32

src/devtools/client/debugger/src/actions/breakpoints/modify.js
```diff
@@ -302,3 +279,16 @@ export function setBreakpointOptions(cx, location, options = {}) {
     await clientSetBreakpoint(client, getState(), breakpoint);
   };
 }
+
+export function validateSyntax(options) {
+  return async ({ parser }) => {
```
jaril, line 284: Can we reuse this logic in setBreakpointOptions?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
```diff
@@ -38,16 +44,22 @@ class PanelEditor extends PureComponent {
...
-
     this.setBreakpoint();
-    toggleEditingOff();
   };
 
   onEnter = () => {
     this.handleSave();
   };
 
+  onChange = async newOption => {
```
jaril, line 54: This might get noisy if it keeps toggling states, but it's a good tradeoff since the immediate feedback's nice.

src/devtools/client/debugger/src/actions/breakpoints/modify.js
```diff
@@ -272,19 +254,14 @@ export function disableBreakpoint(cx, initialBreakpoint) {
...
- * @memberof actions/breakpoints
- * @static
- * @param {SourceLocation} location
- *        @see DebuggerController.Breakpoints.addBreakpoint
- * @param {Object} options
- *        Any options to set on the breakpoint
- */
 export function setBreakpointOptions(cx, location, options = {}) {
-  return async ({ dispatch, getState, client, sourceMaps }) => {
+  return async ({ dispatch, getState, client, parser }) => {
```
jaril, line 258: We can keep this clean and remove the validation here if we disable the save button when there's an error in state.

jaril (approved):
Looking good! 

Not sure what the babel parser would return as an error, but if it's actionable enough for the user we should display it. Could be a followup.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelEditor.js
jasonLaster: yeah, unfortunately we need to go from error back to good on keystroke...

src/devtools/client/debugger/src/actions/breakpoints/modify.js
jasonLaster: good call

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
jasonLaster: went with `--red-50`

PR #1360, opened 2020-12-24
Update the created at contextual timestamp for a replay without refre…

### LINKED ISSUE #1341: “Captured x minutes ago” is better than "x minutes ago"
It is vague to see X minutes ago when it is a low number.... 

It would be nice say Captured if it is minutes or update it every minute so it is up to date...

jaril:
Fix #1341

This updates the "created" time every 30s so it's doesn't become stale.

That said, it doesn't solve the problem where somebody might look at this time and not be sure what it's referring to. Last saved? Last activity? Last comment? For that, we probably should add "Created ...." right at the beginning of that sentence

### DIFF
--- a/src/ui/components/Header/Header.js
+++ b/src/ui/components/Header/Header.js
@@ -71,9 +71,23 @@ function HeaderTitle({ recordingId, editingTitle, setEditingTitle }) {
         editingTitle={editingTitle}
         recordingId={recordingId}
       />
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
Good call, lets add Created.

Tempted to drop the subtitle entirely too... I think we'll likely want a more developed recording information section with creator, created, url similar to the accounts page.

jaril:
Yeah, likely. I'll just merge this for now and we can circle back to it once we want to commit to having the detailed recording information page.

PR #1361, opened 2020-12-24
Removed actions and unified it under seek()

### LINKED ISSUE #1334: We appear to force-switch to the breakpoints pane whenever we stop playing?
1. Open a recording
2. Ensure you are at the end of the recording
3. Select the sources pane
4. Press play to start from the beginning

Expected: Continue showing sources pane
Actual: Force-switched to the breakpoint pane as soon as recording starts playing

Similarly:

1. Open a recording
2. Jump to the middle of a recording
3. Select the sources pane
4. Press play
5. Wait until player reaches the end

Expected: Continue showing sources pane
Actual: Force-switched to the breakpoint list as soon as playing hits end of the timeline

Third:

1. Open a recording
2. Jump to somewhere early enough you can hit pause
3. Select the sources pane
4. Press play
5. Press pause

Expected: Continue showing sources pane
Actual: Force-switched to the breakpoint list as soon as playing is paused

jaril:
Fix #1334

### DIFF
--- a/src/devtools/client/debugger/src/actions/ui.js
+++ b/src/devtools/client/debugger/src/actions/ui.js
@@ -75,13 +75,8 @@ export function toggleFrameworkGrouping(toggleValue) {
   };
 }
 
-export function showSource(cx, sourceId) {
+export function ensureSourcesIsVisible() {
   return ({ dispatch, getState }) => {
-    const source = getSource(getState(), sourceId);
-    if (!source) {
-      return;
-    }
-
     // Make sure the explorer/pause information panel is open so that the user
     // sees those panels
     if (getPaneCollapse(getState())) {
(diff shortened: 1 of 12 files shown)


src/devtools/client/debugger/src/actions/ui.js
```diff
@@ -100,14 +95,21 @@ export function showSource(cx, sourceId) {
...
     if (getSourcesCollapsed(getState())) {
       dispatch({
         type: "TOGGLE_SOURCES",
         sourcesCollapsed: false,
       });
     }
+  };
+}
+
+export function showSource(cx, sourceId) {
```
jaril, line 107: I moved the panel opening logic away from `showSource` and into `ensureSourcesIsVisible`. This is because we don't always want to force-open the sources panel when we're showing the source in the editor. 

For example, when we rewind to a console message, we want to show that source in the editor, but it's more important to show the pause information panel. 

Instead, this makes it so we can just call `ensureSourcesIsVisible` where we need it (e.g. clicking a console message source link, clicking reveal in tree).

src/devtools/client/debugger/src/components/Editor/Tab.js
```diff
@@ -13,7 +13,7 @@ import SourceIcon from "../shared/SourceIcon";
 import { CloseButton } from "../shared/Button";
 import { copyToTheClipboard } from "../../utils/clipboard";
 
-import actions from "../../actions";
+import { actions } from "ui/actions";
```
jasonLaster, line 16: 👍

jasonLaster (approved):

PR #1370, opened 2020-12-25
Make breakpoint panel widths fluid

### LINKED ISSUE #1345: Make breakpoint widget adjust to existing editor width
The breakpoint widget should have a min width and a max width. Between those two widths, it should resize to fill the editor width.

jasonLaster:
fixes #1345 

https://www.loom.com/share/cedc1d6894e9435cb5c427c462a8ed12

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -12,8 +12,12 @@ import BreakpointNavigation from "devtools/client/debugger/src/components/Second
 
 import "./Panel.css";
 
-function getEditorWidth({ editor }) {
-  return Math.max(editor.getScrollInfo().clientWidth - 60, 300);
+function getPanelWidth({ editor }) {
+  // The indent value is an adjustment for the distance from the gutter's left edge
+  // to the panel's left edge, which is approximately ~60.
+  const panelIndent = 60;
+
+  return editor.getScrollInfo().clientWidth - panelIndent;
 }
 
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
```diff
@@ -12,6 +12,10 @@ import BreakpointNavigation from "devtools/client/debugger/src/components/Second
 
 import "./Panel.css";
 
+function getEditorWidth({ editor }) {
+  return Math.max(editor.getScrollInfo().clientWidth - 60, 300);
```
jaril, line 16: Let's add a 500px max value here. It gets too wide otherwise.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
```diff
@@ -71,14 +75,21 @@ function Widget({ location, children, editor, insertAt }) {
...
   const [editing, setEditing] = useState(false);
+  const [width, setWidth] = useState(getEditorWidth(editor));
   const [inputToFocus, setInputToFocus] = useState("logValue");
 
   const toggleEditingOn = () => setEditing(true);
   const toggleEditingOff = () => setEditing(false);
+  const setEditorWidth = () => setWidth(getEditorWidth(editor));
+
+  useEffect(() => {
+    editor.editor.on("refresh", setEditorWidth);
```
jaril, line 86: Ideally there'd be a resize event we can listen for since refresh gets noisy, but I don't think CodeMirror has that outside of refresh. This should be okay.

jaril (approved):

PR #1386, opened 2020-12-28
Bring back timelime marker highlighting when the corresponding console message is hovered

jaril:
This fixes the connection between the console and the timeline, so that hovering on a console message makes the corresponding timeline marker for that message react.

> Wait for #1382 to land before this one.


jasonLaster (approved):
Looks great!


I filed this issue as a follow up (https://github.com/RecordReplay/devtools/issues/1399). This was what highlight location tried to do, but i think we could do it better now.

Issue #1357, opened 2020-12-24
"created" date highlights on hover when it shouldn't


karina-y:
this was introduced when i made the original recorded url clickable, both these have the same ".page-url" class and took the same styling. the "Created x days ago" bit becomes blue, underlined, and cursor set to pointer. none of that should be happening.

![image](https://user-images.githubusercontent.com/19538307/103104794-39389d80-45de-11eb-9b19-cfb273fae269.png)

PR #1387, opened 2020-12-28
changed classname for created from page-url to more applicable name, …

karina-y:
…fixes the created date from turning blue and underlined on hover
fixes issue #1357 

before:
image

after:
image

### DIFF
--- a/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
+++ b/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
@@ -86,7 +86,7 @@ function ItemTitle({ data, editing, editingTitle, setEditingTitle }) {
           </div>
         ) : null}
       </div>
-      <div className="page-url">Created {moment(data.date).fromNow()}</div>
+      <div className="replay-created-at">Created {moment(data.date).fromNow()}</div>
     </div>
   );
 }
--- a/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
+++ b/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
@@ -86,7 +86,7 @@ function ItemTitle({ data, editing, editingTitle, setEditingTitle }) {
           </div>
         ) : null}
       </div>
-      <div className="replay-created-at">Created {moment(data.date).fromNow()}</div>
+      <div>Created {moment(data.date).fromNow()}</div>
     </div>
   );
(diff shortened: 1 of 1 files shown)


src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
```diff
@@ -86,7 +86,7 @@ function ItemTitle({ data, editing, editingTitle, setEditingTitle }) {
           </div>
         ) : null}
       </div>
-      <div className="page-url">Created {moment(data.date).fromNow()}</div>
```
jasonLaster, line 89: do we need to carry over the styling for `page-url`?

src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
```diff
@@ -86,7 +86,7 @@ function ItemTitle({ data, editing, editingTitle, setEditingTitle }) {
...
+      <div className="replay-created-at">Created {moment(data.date).fromNow()}</div>
```
jasonLaster, line 89: perhaps we don't need a className, if we're not going to have any styling

jaril: Yeah we should be able to just take out the classname here. The styling's being handled [here](https://github.com/RecordReplay/devtools/blob/master/src/ui/components/Dashboard/RecordingItem/RecordingListItem.css#L130-L133) without having to refer to a classname.

karina-y: i don't see my comment i posted before hitting resolve.. good point, i'll go ahead and remove it altogether

src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
jaril: Yeah we should be able to just take out the classname here. The styling's being handled here without having to refer to a classname.

karina-y: no i checked with and without and it doesn't alter the styling at all. the only styling attached to that classname is changing the color to blue and underlining the text on hover

karina-y:
tests timing out in ci:
stepping-05, passes locally

jasonLaster (approved):

Issue #1077, opened 2020-11-26
[Debugger] Add an empty visual state for the breakpoints panel


jaril:
Right now, when somebody opens a replay and the breakpoints panel is empty, it looks off-putting since it's a lone panel that doesn't change appearance when you toggle expand/collapse.

It'd be good to use that space more productively. Maybe some text indicating to the user that there's 0 breakpoints and that they have to click on the line gutter to add one.

![image](https://user-images.githubusercontent.com/15959269/100301790-ccce6e00-2f4d-11eb-886c-23c1b0f053dd.png)

PR #1389, opened 2020-12-28
displaying message now when there are no breakpoints to show in the breakpoint panel

karina-y:
fixes issue #1077 
i followed the same format message in the comments panel for the empty breakpoints panel. the text will also wrap to fit the width of the panel and hide when the breakpoint panel is collapsed.

before:
image

after:
image

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
@@ -46,7 +46,7 @@ class Breakpoints extends Component {
 
     if (!breakpointSources.length) {
       return (
-        <div className="no-breakpoints">
+        <div className="empty-state-content">
           There is nothing here yet. Try adding a breakpoint in a source file.
         </div>
       );
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
@@ -45,7 +45,11 @@ class Breakpoints extends Component {
     const { breakpointSources } = this.props;
 
     if (!breakpointSources.length) {
-      return null;
+      return (
+        <div className="no-breakpoints">
+          There is nothing here yet. Try adding a breakpoint in a source file.
+        </div>
+      );
     }
 
     const sources = [...breakpointSources.map(({ source }) => source)];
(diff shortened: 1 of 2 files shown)


jonbell-lot23:
Copy and visual looks good to me, thanks

karina-y:
tests timing out in ci:
stepping-05, passes locally

jonbell-lot23:
Another note on this, inspired by a chat I just had with Jason.

There are sometimes two modes for thinking about design: "good enough" versus "wouldn't it be cool if?" At Microsoft, we sometimes talked about gold/silver/bronze for the level of design support each feature would get.

So from a "good enough" / bronze level of support, this looks good and ready to go! But if I were thinking more broadly about this issue I'd be thinking about illustrations, loading states/edgecases, adding some whimsy to the copy, etc.

So that's something I'm thinking about. We might want a way to quantify the level of design support on tickets so we have a vocabulary for saying "Looks good! But let's open a separate ticket to make it even better later!"

karina-y:
yeah i saw that chat, i liked the idea of using the space for more instructional opportunity. i suppose my thought process is to launch the bronze as the gold is being developed and designed, what are your thoughts @jonbell-lot23 and @jasonLaster ?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
```diff
@@ -45,7 +45,11 @@ class Breakpoints extends Component {
     const { breakpointSources } = this.props;
 
     if (!breakpointSources.length) {
-      return null;
+      return (
+        <div className="no-breakpoints">
```
jaril, line 49: Small nit, it's better to have classnames that describe the element. 

`no-breakpoints` mostly describes the state. For this, something like `empty-state-content` is more descriptive.

jaril (approved):

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/index.js
karina-y: all done

jonbell-lot23:
Closing the loop here -- yup, let's do the bronze level for now and make a new task for making this better, as part of the great onboarding experience work we want to do.

jasonLaster (approved):

karina-y:
tests timing out in ci:
stepping-05 passes locally
object_preview-05 times out in master and branch

Issue #1350, opened 2020-12-23
[UI] Make it more obvious you can edit a title


karina-y:
It's currently not obvious how to edit a title unless you're told how to. A better solution for now would to make it highlight on hover and change the cursor from a pointer to a pencil or text icon.

example:
highlight the text on hover
![image](https://user-images.githubusercontent.com/19538307/103025626-0a9bc380-4507-11eb-8950-9ac4054ad432.png)

show a text cursor on hover rather than a pointer
![image](https://user-images.githubusercontent.com/19538307/103025667-1f785700-4507-11eb-961f-0ba2be631776.png)

karina-y:
discussed with @jonbell-lot23 and will be doing the following:
- add title property to element saying "Click to edit the title."
- changing cursor to text icon on hover
- changing from double click to single click to edit the title

PR #1393, opened 2020-12-29
made editing title of replay more obvious

karina-y:
fixes issue #1350 
- added title property to element saying "Click to edit the title."
- changed cursor to text icon on hover
- changed from double click to single click to edit the title
- also removed extra color styling as it wasn't being used and was overridden by new color style added recently

(not pictured in the screenshots is the cursor changed from 'pointer' to 'text' on hover)
before:
image

after:
image

### DIFF
--- a/src/ui/components/Header/Header.css
+++ b/src/ui/components/Header/Header.css
@@ -81,13 +81,12 @@
 
 #header .title {
   max-width: 320px;
-  color: var(--grey-90);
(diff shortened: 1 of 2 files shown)


karina-y:
i'm anticipating stepping-05 to time out in the ci so i ran it locally and it passe

jasonLaster (approved):

Issue #1392, opened 2020-12-29
remove ability to edit title in accounts page


karina-y:
currently there is a feature to edit the title from your dashboard, however due to another click event, before you're able to edit the title it navigates directly to the replay.

i discussed with @jonbell-lot23 about fixing this and he'd like to remove the ability to edit by title click from the dashboard altogether.

PR #1401, opened 2020-12-29
cant edit title by clicking title in dashboard anymore

karina-y:
fixes issue #1392 
previously, clicking on the title in the dashboard would cause an input box to appear that allows the user to edit, however we weren't calling stopPropogation so before the user got a chance to edit the title it'd just navigate them to the replay page. after discussing with @jonbell-lot23 the ability to edit the title by clicking on it has been removed from the dashboard.
to accomplish this i've added a new property to the Title element to determine whether or not you can edit on click, i chose to do this instead of removing the Title element from the dashboard because we still need the Title element to edit the title via the three dots on the right of the replay row.


karina-y:
tests timing out in ci:
object_preview-05 times out in branch and master
stepping-05 passes in branch and master

karina-y:
reran both tests timing out in ci and both pass in local

jasonLaster (approved):

PR #1403, opened 2020-12-29
Move the avatar to the header

### LINKED ISSUE #1376: Improve login/avatar for the viewer
Currently we neither show you your avatar or prompt you to login in the viewer. We should do both of these things :)

jaril:
Fix #1376.

image
image
image
image


jasonLaster:
I think user options should be next to the other avatars, ideally in the stack on the right or left.

Screen Shot 2020-12-29 at 12 15 53 PM

Issue #1026, opened 2020-11-17
Debugger scrolling jitters with breakpoint panels open


jasonLaster:
STR:
1. add some breakpoints in a file
2. scroll for awhile

notice how occasionally the editor jumps back to the breakpoint panel

PR #1411, opened 2020-12-29
scrolling no longer jitters when adding breakpoint then scrolling in source file

karina-y:
fixes issue #1026 
there isn't much documentation on why _widget.clear() was being called so i tested different scenarios to make sure nothing would break.
- adding a breakpoint to the same line number in two different files
- giving each one different custom logs
- ff and rewinding to each
- removing and re-adding each
- refreshing page to be sure the breakpoints were still there

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -62,12 +62,9 @@ function Widget({ location, children, editor, insertAt }) {
       return;
     }
     const editorLine = toEditorLine(location.sourceId, location.line || 0);
-    const _widget = editor.codeMirror.addLineWidget(editorLine, node, {
+    editor.codeMirror.addLineWidget(editorLine, node, {
       insertAt,
     });
-    return () => {
-      _widget.clear();
-    };
   }, [loading]);
 
   if (!node) {
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -62,9 +62,16 @@ function Widget({ location, children, editor, insertAt }) {
       return;
     }
     const editorLine = toEditorLine(location.sourceId, location.line || 0);
-    editor.codeMirror.addLineWidget(editorLine, node, {
+    const _widget = editor.codeMirror.addLineWidget(editorLine, node, {
       insertAt,
     });
+
+    // We are clearing the widget here because without
+    // doing so causes breakpoints to show up on the
+    // wrong line number when clicking in the gutter.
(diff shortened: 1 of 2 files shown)


jaril:
I don't think it's documented anywhere, but the reason appears to be that without it, adding breakpoints to line numbers break.

This is the closest I got to a proper STR. Note that I add a breakpoint to line 12 towards the end, but the breakpoint is added to line 9 instead. https://share.descript.com/view/LAfAtMkMNZK

karina-y:
@jaril thanks for catching that! i'll convert this to draft and dig more.

karina-y:
@jasonLaster would love your input. here are some discoveries on this issue with help from @jaril 

**problem:**
when widgets would disappear as the user scrolls away from them, the height of the source file would adjust to accommodate.
ie: the widget is 90px in height, and the editor is 800px in height. when the user scrolls a widget into view the source file's height becomes 890px. when the user scrolls away from the widget and it is removed, the source file goes back to 800px. this height adjustment causes a secondary scroll not intentionally triggered by the user. the secondary scroll is always equal to the same height as the widget(s) removed.

**why were we hiding widgets to begin with?:**
this seems to stem from an issue a couple years ago with firefox devtools where it would significantly slow down if you had a minified file you were trying to breakpoint. if all the code is on one line, codemirror would try to inject hundreds of column breakpoints in that line and it would cause a lag.
https://github.com/firefox-devtools/debugger/pull/7311

**solution:**
seeing as we don't have that case with replay, widgets are now shown on load at all times.

**performance tests:**
with 28 breakpoints in one file, the editor component loads in 56.9ms
with only 1 breakpoint in one file, the editor component loads in 49.5ms
with a few dozen in one file and a few dozen in another file, the editor component loads in 51.3ms

**one performance issue:**
there is one performance issue i've noticed, if i add too many breakpoints to separate files, there is a slight lag in switching between the two. when you switch to a tab with many breakpoints they all render at once and are removed when you've navigated to another tab.
this is only apparent when clicking a tab where you have dozens of breakpoints
https://www.loom.com/share/771ea95cb18f488ca3e43ac048def502

when switching to a tab with no breakpoints, there is no lag
https://www.loom.com/share/0a8ddb17d378448690251025bda76e1b

karina-y:
some more perf tests regarding the one performance issue in my comment above:

**THIS PR:**
in the tests below i've replaced all Widget components with an empty div:
- switching from a tab with no breakpoints to a tab with ~30-40 breakpoints
image

- switching from a tab with ~30-40 breakpoints to another tab with about the same amount of breakpoints
image

in the tests below all Widget components are rendering:
- switching from a tab with no breakpoints to a tab with ~30-40 breakpoints
image

- switching from a tab with ~30-40 breakpoints to another tab with about the same amount of breakpoints
image

karina-y:
reran test timing out in ci and it passes in local

jasonLaster (approved):

PR #1415, opened 2020-12-30
Retain the splitbox proportions when resizing/zooming the viewport

jaril:
This keeps the splitbox proportions the same when the user:
1) Resizes the browser viewport, or,
2) Zooms in the browser viewport (`Cmd +` / `Cmd -`)

Fix #1415.

### DIFF
--- a/src/devtools/client/debugger/src/components/App.js
+++ b/src/devtools/client/debugger/src/components/App.js
@@ -41,7 +41,7 @@ import "devtools/client/themes/common.css";
 
 import "./shared/menu.css";
 
-import SplitBox from "devtools-splitter";
+import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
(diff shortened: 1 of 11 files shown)


jaril:
This currently resizes the sidebar panel when moving the splitter between the editor and the video/toolbar. Needs to be fixed.

jaril:
Update: Resizing three panes with splitbox is hard. My wish list was below:
1) Splitbox proportions remain the same on zoom and resize
2) Resizing with the splitter between sidebar and editor should not affect the viewer+toolbar panel width
3) Resizing with the splitter between the editor and viewer+toolbar panel should not affect the sidebar width

Looks like getting 1 is easy. But we can't have both 2 and 3. At least, not with the current splitbox implementation where resizing is only aware of 2 panels, and doesn't consider the 3rd panel.

PR #1424, opened 2020-12-30
Allow the timeline playback to continue playing without interruptions

### LINKED ISSUE #1390: Pressing play should not pause at breakpoints
one thing i've seen in research and have been coming around to, is that pressing play should not pause at breakpoints.

playing is all about watching a video, and pausing for the user is a pretty surprising moment for most folks

jaril:
As it is, we stop the ongoing playback whenever a breakpoint is hit. This PR makes playback continuous. Once a user presses play on the timeline, the timeline will continue to play until it hits the end of the recording.

This fixes #1390.

### DIFF
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -281,13 +281,8 @@ export class Timeline extends Component {
 
   /**
    * Playback the recording segment from `startTime` to `endTime`.
-   * Optionally a `pauseTarget` may be given that will be seeked to after finishing playback.
    */
-  async playback(startTime, endTime, pauseTarget) {
-    if (pauseTarget) {
-      assert(endTime <= pauseTarget.time);
-    }
-
+  async playback(startTime, endTime) {
     let startDate = Date.now();
     let currentDate = startDate;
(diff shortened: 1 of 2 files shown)


src/ui/components/Timeline/index.js
```diff
@@ -298,10 +293,12 @@ export class Timeline extends Component {
...
       nextGraphicsPromise = getGraphicsAtTime(nextGraphicsTime);
     };
-
     prepareNextGraphics();
 
-    while (this.props.playback) {
+    // This is a `do ... while` instead of `while` because startPlayback is synchronous.
+    // It's possible that the playback in state hasn't been updated yet by the time
+    // this is first called. So we just force it to run the first time.
+    do {
```
jasonLaster, line 301: i can see why you wanted to clean this up and move it to redux :)

jasonLaster (commented):
overall looks much cleaner, but i couldn't test. Feel free to land and then extract to redux

src/ui/components/Timeline/index.js
jaril: Will do!

PR #1429, opened 2020-12-31
Occupy the widget bottom row when there are no hits

### LINKED ISSUE #1369: "No hits" should be more obvious
the breakpoint panel "no hits" label should be more obvious

jaril:
Fix #1369. Let's circle back to this when we have a better idea about loading state and allowing/disallowing users to add breakpoints on lines that don't have any corresponding points.

Loaded with hits and loaded with no hits: 
image
Loading:
image


jasonLaster:
perhaps we should drop the bar because the new timeline is more minimal and just use a font bold treatment

jasonLaster (approved):

PR #1430, opened 2020-12-31
Force guests to sign in before leaving a comment

### LINKED ISSUE #1374: [Commnts] Guests should be prompted to login
Guests should be prompted to login before they can add a comment

jaril:
Fixes #1374.

This makes it so that clicking on the add a new comment button as an anonymous user shows a screen that prompts the user to sign in.

image


jasonLaster:
hmm, i would prefer something more subtle like the share modal

jasonLaster:
I wonder if we should show the comment tooltip, but instead of showing the text box, just have a log in button...

PR #1432, opened 2020-12-31
Make sure the the focused source remains focused when toggling the sidebar panel

### LINKED ISSUE #1213: collapsing sources panel loses highlight on active source file
**Describe the bug**
if you select a file from the sources panel it gets highlighted in blue. but if you collapse and reopen the panel you lose that blue highlight.

**To Reproduce**
1. Go to any recording
2. select file in sources panel, notice it gets highlighted in blue
3. collapse sources panel
4. reopen sources panel
5. notice active file is no longer highlighted

**Expected behavior**
active file retains highlight

**Recording**
https://www.loom.com/share/50fcbe67f96e4a7d9d75eb4b406a900c

jaril:
Fixes #1213.

Demo: https://share.descript.com/view/tJBsykQZToV

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
@@ -19,6 +19,7 @@ import {
   getContext,
   getExtensionNameBySourceUrl,
   getSourceContent,
+  getFocusedSourceItem,
 } from "../../selectors";
 import actions from "../../actions";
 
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -264,13 +273,14 @@ function isExtensionDirectory(depth, extensionName) {
...
 const mapStateToProps = (state, props) => {
-  const { source, item } = props;
+  const { source, item, focused } = props;
   return {
     cx: getContext(state),
     hasMatchingGeneratedSource: getHasMatchingGeneratedSource(state, source),
     hasSiblingOfSameName: getHasSiblingOfSameName(state, source),
     hasPrettySource: source ? checkHasPrettySource(state, source.id) : false,
     sourceContent: source ? getSourceContentValue(state, source) : null,
+    focused: source ? focused || getFocusedSourceItem(state)?.contents.id == source.id : false,
```
jasonLaster, line 283: `focused || (source && getFocusedSourceItem(state)?.contents.id == source.id)`

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
```diff
@@ -30,16 +31,33 @@ import {
...
+  _node;
+
   componentDidMount() {
-    const { autoExpand, item } = this.props;
+    const { autoExpand, item, focused } = this.props;
     if (autoExpand) {
       this.props.setExpanded(item, true, false);
     }
+    if (focused) {
+      this._node.parentElement.classList.add("focused");
```
jasonLaster, line 46: reaching outside of the current element is a bit risky... Why do we need to do that here?

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
jaril: Better. Thanks!

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.js
jaril: Very risky.

The SourcesTreeItem is just the tree node. It doesn't include the entire row. Without this. the focus persists when toggling the panel, but the indent doesn't have the focus styling. I added screenshots below.

The proper way to do it would be to apply the focused styling to the entire row. To do that, we'd have to edit code in the `Tree` component. But [we're currently importing it from npm](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/components/shared/ManagedTree.js#L9), so we can't.

**With parentElement styling:**
image

**Without:**
image

jasonLaster: ooh... actually 😸 

haven't tested this yet 
https://github.com/RecordReplay/devtools/pull/1439

jasonLaster:
This should be much easier now :)

PR #1434, opened 2020-12-31
Update the video in DevTools when the user hovers on console messages

### LINKED ISSUE #1125: [Viewer] Show a preview in the viewer when hovering on a console message
Hovering on a log message used to show the timeline tooltip to give the user some idea of what the viewport look like at the time that that console message was logged. To keeps things simple, we got rid of it  with the new replay layout.

When I'm debugging failing tests, my first instinct is to check the console for the `ERROR: Test failed` message. I find myself wanting to hover on that message, and hovering on previous messages to see what led to the error. But because we no longer have the preview functionality, I have to pause at each of those messages to see what the viewport looked like at that time.

We should bring it back to make that experience much smoother. Instead of showing the preview as a tooltip in the timeline, it would be nice if we could show the preview in the viewer when a user hovers on a console message.

> Note: This runs the risk of being very noisy visually. Just because a user is hovered on a message doesn't mean that they want to see a preview. But given our current feature set, if a user is hovered on a message, it's usually for navigation purposes. Because of that I don't think it should be a problem.

jaril:
Fixes #1125.

Demo: https://share.descript.com/view/ALoiRY3QZhj

### DIFF
--- a/src/ui/components/Timeline/index.js
+++ b/src/ui/components/Timeline/index.js
@@ -115,13 +115,17 @@ export class Timeline extends Component {
 
   // Called when hovering over a message in the console.
   onConsoleMessageHover = async (type, message) => {
-    const { setTimelineState } = this.props;
+    const { setTimelineState, currentTime } = this.props;
 
     if (type == "mouseenter") {
       setTimelineState({ hoveredMessageId: message.id });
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
I think this is going to test really well.

PR #1437, opened 2020-12-31
Re-enable user select

### LINKED ISSUE #1339: Let me copy text out of the scopes panel
It would be nice if you could select text in a rep in the scopes pane

Additionally, if you right click on a node, there should be context menu for copying the value

jaril:
Selecting text in the scopes panel works on Firefox. The difference was that we're using a css file (`markup.css`) that they aren't, and there's a `user-select: none` in that file. I took it out.

Fixes #1339.

### DIFF
--- a/src/devtools/client/themes/markup.css
+++ b/src/devtools/client/themes/markup.css
@@ -32,10 +32,6 @@
   -moz-control-character-visibility: visible;
 }
 
-body {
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):
we might need to add new `user-select`s but that's fine

PR #1442, opened 2021-01-01
resizing icons to 28px rather than 26px

jonbell-lot23:
1. Moved the icon to 28px as designed
2. Exported from Figma rather than Sketch


jasonLaster:
Looks great

PR #1460, opened 2021-01-04
removed inspector panel.js

karina-y:
- no longer initializing inspector from panel.js, this is happening in toolbox now (where it will eventually be moved out of once toolbox has been removed entirely)
- removed some unused imports
- tested with:
	- elements panel open on startup
	- elements panel open on startup then navigating to console
	- tested with console panel open on startup then navigating to elements
	- tested with comments panel open on startup then navigating to elements
	- various other combinations of clicking around panels


src/ui/utils/devtools-toolbox.js
```diff
@@ -54,11 +54,14 @@ export class DevToolsToolbox {
 
     const panels = {
       debugger: DebuggerPanel,
-      inspector: InspectorPanel,
+      inspector: Inspector,
```
jasonLaster, line 57: could we also remove this Inspector as well?

karina-y: the entire inspector.ts file?

karina-y:
tests timing out in ci:
breakpoints-05 passes locally
inspector-02 times out locally and on master

jasonLaster (approved):

PR #1466, opened 2021-01-05
informing user of dead session

karina-y:
- letting user know when their session has been closed due to inactivity
- had to refactor logpoints a bit due to a circular reference from socket.ts, only setting up log points now if we have a recording
- force testing this works fine but i'm going to leave a window open tonight and check it in the morning to confirm

### DIFF
--- a/src/protocol/logpoint.ts
+++ b/src/protocol/logpoint.ts
@@ -7,6 +7,7 @@
 // messages associated with the logpoint atomically.
 
 import {
+  AnalysisEntry,
   AnalysisId,
   createAnalysisResult,
   ExecutionPoint,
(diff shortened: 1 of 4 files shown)


src/protocol/logpoint.ts
```diff
@@ -55,75 +55,77 @@ export const PointHandlers: {
   onPoints?: (points: PointDescription[], info: LogpointInfo) => void;
 } = {};
 
-client.Analysis.addAnalysisResultListener(({ analysisId, results }) => {
-  log(`AnalysisResults ${results.length}`);
+export function setupLogpoints() {
```
jasonLaster, line 58: why did this need to be in a setup function?

could we move the handler bits out `({ analysisId, results }) => {})`

so it reads

```js
export function setupLogpoints() {
  client.Analysis.addAnalysisResultListener((onAnalysisResult))
}
```

src/protocol/socket.ts
```diff
@@ -113,6 +114,12 @@ function onSocketMessage(evt: MessageEvent<any>) {
 function onSocketClose() {
   setStatus("Disconnected.");
   log("Socket Closed");
+
+  store.dispatch(
```
jasonLaster, line 118: is the store global? Is there any way to pass it in... 

perhaps ... https://gist.github.com/73b610bde300ea6e50c5ef58ad18ac90

karina-y: it is global and it does work when i test so i'm not sure why it's throwing an error on store. but i'll add your changes in and see if it fixes that error

src/protocol/logpoint.ts
```diff
@@ -123,7 +124,13 @@ client.Analysis.addAnalysisPointsListener(({ analysisId, points }) => {
       LogpointHandlers.onPointLoading!(logGroupId, point, time, location);
     });
   }
-});
+};
+
+export function setupLogpoints() {
+  client.Analysis.addAnalysisResultListener(({ analysisId, results }) => onAnalysisResult);
```
jasonLaster, line 130: shouldn't this be?

```js
client.Analysis.addAnalysisResultListener(({ analysisId, results }) => onAnalysisResult(analysisId, results))
```

karina-y: you're absolutely right, i'll start leaving late changes for the morning so i don't miss things like this.

karina-y:
inspector-02 is timing out in ci, local branch, and master

PR #1496, opened 2021-01-08
Update the replay's page title

### LINKED ISSUE #1490: Show recording titles in the title
At the moment Replay's title is just "Replay", but it should be the recording's title

jaril:
Fixes #1490

### DIFF
--- a/src/ui/components/DevTools.js
+++ b/src/ui/components/DevTools.js
@@ -127,6 +127,10 @@ function DevTools({
     return null;
   }
 
+  if (document.title != data.recordings[0].title) {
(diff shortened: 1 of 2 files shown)


src/ui/components/DevTools.js
```diff
@@ -127,6 +127,10 @@ function DevTools({
     return null;
   }
 
+  if (document.title != data.recordings[0].title) {
+    document.title = data.recordings[0].title;
+  }
```
jasonLaster, line 132: We should put this into a `useEffect`. It's not necessary, but I think it's a bit more proper

PR #1503, opened 2021-01-08
Avoid race between `.click()` and mapped_minified.js load.

loganfsmyth:
There's nothing guaranteeing the execution ordering between the `setTimeout` callback and the `mapped_minified.js` script here, and the tests assume that the minified code runs first, so this was causing test flakiness.


jasonLaster (approved):

jasonLaster:
Thanks for looking into this

PR #1512, opened 2021-01-11
Remove comment tooltip

jaril:
Chrome demo: https://share.descript.com/view/wZiAC0QPqR3
Replay browser demo: https://share.descript.com/view/K26fyBCdN0l

In addition to taking out the comment tooltip, this simplifies how we add new comments and replies. This is implemented primarily with a `pendingComment` – a shim that's stored in the redux store that we used to mock the new comment before the user fills in its contents and we save it in Hasura. 

That's different from the previous approach where every new comment are added to Hasura immediately. That introduced a lot of complexity because by default we had to involve Hasura with every interaction with the "pending" new comment.


jasonLaster:
Looks great!

PR #1515, opened 2021-01-12
Show "Rules not available" when getAppliedRules fails

hbenl:


src/devtools/client/inspector/rules/components/RulesApp.js
```diff
@@ -183,7 +183,7 @@ class RulesApp extends PureComponent {
     const pseudoElementRules = [];
     const styleRules = [];
 
-    for (const rule of rules) {
+    for (const rule of rules || []) {
```
jasonLaster, line 186: could we ensure the rules are always an array?

src/devtools/client/inspector/rules/actions/rules.ts
```diff
@@ -56,10 +56,10 @@ export function updateHighlightedSelector(
  * @param  {Array} rules
  *         Array of Rule objects containing the selected element's CSS rules.
  */
-export function updateRules(rules: Rule[]): UpdateRulesAction {
+export function updateRules(rules: Rule[] | null): UpdateRulesAction {
   return {
     type: "UPDATE_RULES",
-    rules: rules.map(rule => getRuleState(rule)),
+    rules: rules?.map(rule => getRuleState(rule)),
```
jasonLaster, line 62: perhaps here we should do  `rules ? rules.map(rule => getRuleState(rule)) : []`

src/protocol/thread/node.ts
```diff
@@ -213,12 +213,15 @@ export class NodeFront {
             this._computedStyle.set(name, value);
           }
         }),
-      this._pause
-        .sendMessage(client.CSS.getAppliedRules, { node: this._object.objectId })
-        .then(({ rules, data }) => {
+      this._pause.sendMessage(client.CSS.getAppliedRules, { node: this._object.objectId }).then(
```
jasonLaster, line 216: stylistically I'd prefer a try/catch

```js
try {
  const{rules, data} =  await this._pause.sendMessage(client.CSS.getAppliedRules, { node: this._object.objectId })
  ....
} catch (e) {
  this._rules = null;
}
```

jasonLaster (commented):
A bunch of minor style comments, but otherwise looks good

src/devtools/client/inspector/rules/components/RulesApp.js
hbenl: No, that's the point of this PR: we need to distinguish between the situation where `getAppliedRules` succeeds and returns 0 rules versus the situation where it fails. But in this place we don't need that distinction.

src/devtools/client/inspector/rules/actions/rules.ts
hbenl: No, see my comment above.

src/protocol/thread/node.ts
hbenl: So do I, but in this case we'd have to wrap it in an async IIFE (because we need the resulting Promise), so it would be
```
      (async () => {
        try {
          const { rules, data } = await this._pause.sendMessage(client.CSS.getAppliedRules, {
            node: this._object.objectId,
          });
          this._rules = uniqBy(rules, (rule: AppliedRule) => `${rule.rule}|${rule.pseudoElement}`);
          this._pause.addData(data);
        } catch (e) {
          this._rules = null;
        }
      })(),
```
versus
```
      this._pause.sendMessage(client.CSS.getAppliedRules, { node: this._object.objectId }).then(
        ({ rules, data }) => {
          this._rules = uniqBy(rules, (rule: AppliedRule) => `${rule.rule}|${rule.pseudoElement}`);
          this._pause.addData(data);
        },
        () => {
          this._rules = null;
        }
      ),
```
I think I prefer the second version here.

jasonLaster: ugh, i begrudgingly agree... This code feels awkward, like we probably are missing ...

```js
  async ensureLoaded() {
    ....
    const stylesLoading = this.loadComputedStyles(...)
    const  rulesLoading = this.loadAppliedRules(...)
    await Promise.all([stylesLoading, rulesLoading])
    ...
 }
```

Up to you though... I trust your judgement.

src/devtools/client/inspector/rules/components/RulesApp.js
jasonLaster: got it, that makes more sense. 

Another way to do this would be to introduce a `groupRules` function that returns `inherited`, `pseudoElements`, and `styleRules`  and factor out the branching into `renderPane`:

that would look like this:

```js
renderPane() {
  if (!rules) { 
     return <div className="devtools-sidepanel-no-result" >No Rules</div>
  }

  if (rules.length == 0) {
   return <div className="devtools-sidepanel-no-result" >Rules not available</div>
  }

  const { styled, pseudoElements, inherited } = this.groupRules(rules)
  return <>
   {this.renderPseudoElementRules(pseudoElementRules)}
   {this.renderStyleRules(styleRules)}
   {this.renderInheritedRules(inheritedRules)}
  </>
}

render() {
  return <div id="sidebar-panel-ruleview" className="theme-sidebar inspector-tabpanel ">
    {this.renderPane()}
  </div>
}
```

This is cosmetic, but i think it is nice to move branching into functions with their own returns. 

Also, i'd like to start using JSX where we can in the elements panel... It can be gradual, but it is much nicer to read.

Also, I'd like to get rid of getStr and l10n globally. The properties files are big and are making our build process much more complicated.

src/devtools/client/inspector/rules/actions/rules.ts
jasonLaster: got it!

src/devtools/client/inspector/computed/elementPageStyle.ts
```diff
@@ -19,7 +19,7 @@ export class ElementPageStyle {
   getMatchedSelectors(property: string) {
     const selectorInfos: SelectorInfo[] = [];
 
-    for (const rule of this.store.getState().rules.rules) {
+    for (const rule of this.store.getState().rules.rules || []) {
```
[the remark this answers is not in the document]
jasonLaster: it'd still be nice to have a selector here

```js
const rules = getAppliedRules(getState()) || []
for (const rule of appliedRules) {
```

jasonLaster (approved):
Stylistic points, but feel free to land when it's ready.

hbenl:
I agree with the stylistic improvements but decided to defer them to a future PR.

All failed tests ran locally except inspector-02.js (which currently also doesn't run without this PR).

PR #1520, opened 2021-01-12
Fix console object inspector to sort arrays correctly

### LINKED ISSUE #1507: Arrays should be sorted numerically
**Description:**
Arrays are sorted alphabetically and not numerically. 

**Steps to Reproduce:**
1. go to https://replay.io/view?id=8326fb65-fcb0-4364-bc53-e8647d6d18ee
2. evaluate `s.points.map(p => p.point)` I believe

**Expected Result:** the array should be sorted numerically

**Screenshots:**

Screen Shot 2021-01-09 at 8 57 06 AM

ryanjduffy:
When both keys are numbers, sort numerically instead of alphabetically.

image

This proposed fix will sort the keys of objects too like the screenshot below.

image

Another alternative is to typecheck the object as an array and apply a different sort for those vs objects. I can change to this approach if it is preferred but I think the proposed solution is reasonable.

Fixes #1507

### DIFF
--- a/src/protocol/thread/value.ts
+++ b/src/protocol/thread/value.ts
@@ -270,6 +270,14 @@ export class ValueFront {
       contents,
     }));
     rv.sort((a, b) => {
+      // if both element names are numbers, sort them numerically instead of
+      // alphabetically.
+      const aN = Number.parseInt(a.name);
+      const bN = Number.parseInt(b.name);
+      if (!isNaN(aN) && !isNaN(bN)) {
+        return aN - bN;
+      }
+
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):

jasonLaster (approved):
I like this approach.

jasonLaster:
I kicked off another run of the tests, so hopefully things will be greener (https://github.com/RecordReplay/devtools/actions/runs/481486234)

ryanjduffy:
Eek! Lots of failed tests :/

Issue #283, opened 2020-07-28
[Outline] selection should fill the full width


jasonLaster:
STR:

1. go to https://webreplay.io/view?id=d8dbdd03-1ee8-4779-af83-995d69eb1c11
2. open `https://firebugs.dev/static/js/main.466d878e.chunk.js`
3. select the outline pane

<img width="308" alt="Screen Shot 2020-07-28 at 9 24 31 AM" src="https://user-images.githubusercontent.com/254562/88705060-db9e3480-d0c3-11ea-9aef-fa51da5c55a3.png">

princiya:
Thanks for creating this issue. I will address this in my PR.

PR #1523, opened 2021-01-13
fix the rendering of the Outline view in Chrome

hbenl:


jasonLaster:
Lets make sure don't regress this: https://github.com/RecordReplay/devtools/commit/8f50b1ea79b724176155f09a98d57720a1a68d16

hbenl:
I don't understand why that patch solved #283 or why it doesn't seem to be needed anymore, but that seems to be the case:
From #283:

From my test with this PR:
Screenshot_2021-01-14_10-39-00

jasonLaster (approved):

PR #1524, opened 2021-01-13
Fix console messages to wrap around location

### LINKED ISSUE #1519: Console messages should look better in narrow consoles
**Description:**

When the console is really narrow, the console location fills almost the entire width, which forces the message to be a single column.

**Steps to Reproduce:**

1. go to https://replay.io/view?id=b27b7ee9-19cf-4efe-aaf6-d276688a3d46
2. view the console when its super narrow

**Additional Context:**

Screen Shot 2021-01-12 at 3 26 51 PM

ryanjduffy:
## Issue

In narrow viewports, the message body width is limited by the location and can cause the content to be rendered vertically one character at a time.

## Analysis

The flex box layout reserves space for the location at the expense of the body. Furthermore, flex reserves all the vertical space below the location so the body can't wrap into that space either.

## Resolution

* Move the location into the body and float it right so it only consumes the vertical and horizontal space it requires.
* Add `word-break: break-all` to the content so single word content wraps mid-word rather than be forced to the next line.

image

## Additional Considerations

I'm not sure how to test this with all possible message types so it's possible there are some regressions but the DOM structure is consistent for all so I expect it to be okay.

Fixes #1519

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -324,16 +324,14 @@
 }
 
 .webconsole-app .message-location {
-  max-width: 40vw;
-  flex-shrink: 0;
+  float: right;
   color: var(--frame-link-source);
-  margin: 0 1ch;
   /* Makes the file name truncated (and ellipsis shown) on the left side */
   direction: rtl;
+  margin-inline-end: 1ch;
   white-space: nowrap;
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):

jasonLaster:
looks great!

PR #1572, opened 2021-01-25
refactor authentication and bootstraping

hbenl:


src/ui/actions/session.ts
```diff
@@ -21,19 +21,24 @@ declare global {
...
+export async function createSession(store: UIStore, recordingId: string) {
   addEventListener("Recording.uploadedData", (data: uploadedData) =>
     store.dispatch(onUploadedData(data))
   );
   addEventListener("Recording.sessionError", (err: sessionError) =>
     store.dispatch(onSessionError(err))
   );
   try {
-    ThreadFront.setTest(getTest());
+    ThreadFront.setTest(getTest() || undefined);
```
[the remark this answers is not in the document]
hbenl: The thing is that we're being inconsistent in our code about using `null` vs. `undefined`. Typescript warned me about it here and I decided to go for the least intrusive fix.

src/ui/utils/tokenManager.tsx
```diff
@@ -0,0 +1,142 @@
...
+    } catch (e) {
+      if (e.error !== "login_required" && e.error !== "consent_required") {
+        throw e;
+      }
+
+      return await this.auth0Client.getAccessTokenWithPopup({ audience, ignoreCache: refresh });
+    }
+  }
+
+  private setState(state: TokenState, deferredState: Deferred<TokenState>) {
```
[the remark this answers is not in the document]
hbenl: `this.deferredState` could be replaced while we're fetching the token. As a result, the promise in the old Deferred would not get resolved and we'd eventually try to resolve the promise in the new Deferred a second time. Admittedly this is a _very_ theoretical possibility.

src/ui/utils/tokenManager.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+        this.setState({ token }, deferredState);
+        if (deferredState === this.deferredState) {
+          this.setupTokenRefresh(token);
+        }
+      } catch (e) {
+        this.setState({ error: e }, deferredState);
+      }
+    } else {
+      this.setState({}, this.deferredState);
```
loganfsmyth, line 115: Should this and the other calls in here be this.deferredState.resolve? Does `setState` accept a deferred?

src/ui/utils/tokenManager.tsx
```diff
@@ -0,0 +1,150 @@
...
+        }
+      }
+
+      this.isTokenRequested = true;
+      const deferredState = this.deferredState;
+      try {
+        const token = await this.fetchToken(refresh);
+
+        this.setState({ token }, deferredState);
+        if (deferredState === this.deferredState) {
```
loganfsmyth, line 108: Is there concern that this will change as a side-effect of triggering the listeners in `setState`, or only that it will change during fetching? If it's just fetching, it might be clearer to have
```
const deferredState = this.deferredState;
try {
  const token = await this.fetchToken(refresh);
  if (deferredState === this.deferredState) {
    return;
  }
  this.setState({ token });
  this.setupTokenRefresh(token):
} catch (e) {
  this.setState({ error: e });
}
assert(deferredState === this.deferredState);
```
to remove the need for the `deferredState` param.

loganfsmyth (approved):
Left a comment but your call whether it's worth changing.

src/ui/utils/tokenManager.tsx
hbenl: I was only concerned about it changing during fetching. But in that case I don't want to simply return because that would mean a promise returned by `getToken()` in the meantime would never be settled. Also, your code could still end up trying to resolve a Deferred twice if `this.deferredState` changed while fetching and the fetching then throws.

PR #1579, opened 2021-01-25
Refresh codemirror when going in/out of breakpoint editing mode, fixes #1568

### LINKED ISSUE #1568: Problems selecting text while editing a logpoint
**Description:**
While editing a logpoint, rendering issues appear to be causing copy and paste functions to copy and paste text other than the text that was selected.
This is an intermittent bug that clearly manifests in the replay, but can be difficult to reproduce otherwise.

**Steps to Reproduce:**
1. Go to breakpoints/debugging screen
2. Add logpoint
3. Edit logpoint
4. Under edit box, select text and copy. See that the selection is not correctly aligned with the text.
5. Try to paste that text, it is different from what was selected. Copying and pasting into the logpoint edit box is something a user is likely to do when adding code to a logpoint.

**Expected Result:**
Selected text copied
**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:
https://replay.io/view?id=47d08792-b711-4ce5-9718-81ffa6a62913

bhackett1024:
It's hard to tell from the recording in #1568 why the selection doesn't refresh properly here, since the issue is in CodeMirror which I don't know well.  We've had other similar issues with CodeMirror though, and forcing it to refresh fixed those problems.  That works here as well --- we need CodeMirror to update both when going in and out of editing mode on the breakpoint in order for code selection to work as expected.  Sending a resize event to the window is the best way I've found for doing that.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -89,6 +89,10 @@ function Panel({ breakpoint, editor, insertAt, hoveredPoint }) {
   const [width, setWidth] = useState(getPanelWidth(editor));
   const [inputToFocus, setInputToFocus] = useState("logValue");
 
+  // Force CodeMirror to refresh when changing the size of the breakpoint
+  // panel so that code selection works properly.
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
Works well. Thanks!

PR #1582, opened 2021-01-26
Don't use cx in setBreakpointPositions

### LINKED ISSUE #1567: Number of hits does not show when hovering over a line number in pause mode
**Description:**
When in the paused state in the breakpoints/debugging screen, the number of hits does not show. when hovering over a line number in the gutter.
This is an intermittent bug that clearly manifests in the replay, but can be difficult to reproduce otherwise.

**Steps to Reproduce:**
1. Go to breakpoints/debugging screen in Devtools
2. Click on a message in the console log to pause
3. Hover over a line number to see number of hits, the ellipsis remains, number of hits is never shown.
**Expected Result:**

**Additional Context:**

* Screenshots:
Screen Shot 2021-01-23 at 10 24 31 AM

* Video (Loom):
* Replay:
https://replay.io/view?id=b13a688a-53eb-478b-a904-e8932136673e

bhackett1024:
Fixes #1567

From the recording in that bug it's pretty easy to see that we threw an exception inside updateHoveredLineNumber and execution didn't get far enough along to actually call actions.setHoveredLineNumberLocation.

Debugger action contexts are a little tricky to use.  I added these to the firefox devtools back in 2019 to fix a raft of problems with stale async actions updating debugger state after the user resumed, paused, or navigated.  The replay devtools don't need the context as much, however, because they can't navigate and change the set of loaded sources.  They're still needed when the action is updating state related to the current pause, but that isn't needed when setting breakpoint positions.  There's a lot more cx removal that can be done here and in related code, but this is enough to get the hits to load as expected.

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/breakpointPositions.js
+++ b/src/devtools/client/debugger/src/actions/breakpoints/breakpointPositions.js
@@ -59,7 +59,7 @@ function groupByLine(results, sourceId, line) {
   return positions;
 }
 
-async function _setBreakpointPositions(cx, sourceId, line, thunkArgs) {
+async function _setBreakpointPositions(sourceId, line, thunkArgs) {
   const { client, dispatch, getState, sourceMaps } = thunkArgs;
   let generatedSource = getSource(getState(), sourceId);
   if (!generatedSource) {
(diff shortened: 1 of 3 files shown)


jasonLaster (approved):
Thanks Brian. I like the idea of limiting our usage of `cx` to just where it is needed

jasonLaster:
I think these tests were created an hour ago before the hasura fix was deployed.

- Failure stepping-06.js https://replay.io/view?id=a48a79c7-d1dd-4a31-adbb-125330c39d82&test=stepping-06.js |
- Failure breakpoints-03.js https://replay.io/view?id=074e8f6e-04bc-4706-a1f0-cc61322bedd9&test=breakpoints-03.js

PR #1601, opened 2021-01-27
Fix object inspector stuck in loading state, fixes #1033

### LINKED ISSUE #1033: Adding a condition to a breakpoint doesn't update its total point count
Conditions are working as they should now. But when we add a condition to a breakpoint, its total count remains the same.

We should either:
1) Simple version - Update the total count to reflect the number of points under the new condition.
2) Complicated version - Update the total count, but find a way to display the original total count as well. Not sure what this might look like, but it would be some shorthand for indicating "8 hits for line 12, and 2 hits for line 12 under the new condition".

bhackett1024:
I tried to write a test for this but unfortunately it doesn't repro on our simpler test cases.  There are several different maps used for caching in the object inspector and I feel this could be simplified greatly.  The protocol values do their own caching.

### DIFF
--- a/src/devtools/client/debugger/packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
+++ b/src/devtools/client/debugger/packages/devtools-reps/src/object-inspector/components/ObjectInspector.js
@@ -103,8 +103,9 @@ class ObjectInspector extends Component {
       return;
     }
 
-    for (const [path, properties] of nextProps.loadedProperties) {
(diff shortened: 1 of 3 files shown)


bhackett1024:
For https://github.com/RecordReplay/backend/issues/1033

jasonLaster:
@hbenl could you take a look?

hbenl (approved):
LGTM

Issue #1549, opened 2021-01-20
Icons are blank on first load


jasonLaster:
I'm not sure what the steps are to reproduce this, but it seems like when I come back to replay after a day or so, the icons are missing on the first load.

## Missing icons
<img width="1497" alt="Screen Shot 2021-01-20 at 10 09 27 AM" src="https://user-images.githubusercontent.com/254562/105216895-082d8900-5b08-11eb-9424-4ae7695a71e7.png">

## styles with the icon
<img width="420" alt="Screen Shot 2021-01-20 at 10 12 02 AM" src="https://user-images.githubusercontent.com/254562/105216885-049a0200-5b08-11eb-87df-d44b1dd3beaa.png">

## styles without the icon
<img width="462" alt="Screen Shot 2021-01-20 at 10 10 53 AM" src="https://user-images.githubusercontent.com/254562/105216890-0663c580-5b08-11eb-8e65-e65f647e61ba.png">

jasonLaster:
I wonder if the fix might be a webpack config that builds a CSS file instead of creating the inline styles on the fly.

hbenl:
I couldn't reproduce this but I think this would be a good chance to move to a standard webpack mechanism. See #1615.

PR #1615, opened 2021-01-28
Use webpack for creating our image styles (#1549)

hbenl:
This PR replaces [our own mechanism](https://github.com/RecordReplay/devtools/blob/565af73073bbfa22e67993d861deb1ef0709df15/src/image/image.js#L7) for creating styles for bundled images with webpack's standard mechanism.

One notable difference is the use of the `mask-image` property: Chrome-based browsers currently only support `-webkit-mask-image`, so our mechanism used to create both properties. We could
* include both properties in the new `image.css` file, but that would mean a lot of repetition
* only use the `-webkit-mask-image` property, which is supported in all browsers, but is non-standard
* use the `mask` shorthand property, which is surprisingly supported by chrome-based browsers. The only downside is that we have to add `center / cover` after the url because otherwise the `mask-size` property is reset to `auto`.

In order for webpack's `css-loader` to use [our `modules` configuration](https://github.com/RecordReplay/devtools/blob/565af73073bbfa22e67993d861deb1ef0709df15/webpack.config.js#L32), the urls need to start with `~`. See [here](https://github.com/webpack-contrib/css-loader#url)

Our webpack configuration [disables resolving urls](https://github.com/RecordReplay/devtools/blob/565af73073bbfa22e67993d861deb1ef0709df15/webpack.config.js#L73) in css files, I'm guessing this was done in order not to bundle large images. I reenabled it for the new css file only. I think we should switch to enabling it by default and only exclude the large images, either placing them in one folder (e.g. `src/image/images/unbundled`) or referencing them in the same css file.


jasonLaster:
Thanks it's nice to get rid of our custom image.js script.

Out of curiosity, do you know what the bug was with the old implementation? And how this might fix it?

Also, do you think it is a good idea to bundle the CSS with the JS or should we move to a separate CSS file? We don't need to do it now, just curious.

hbenl:
> Out of curiosity, do you know what the bug was with the old implementation?

No idea. I was never able to reproduce it.

> do you think it is a good idea to bundle the CSS with the JS or should we move to a separate CSS file?

I'd stick to the simplest solution (one bundle that includes everything) until we find a problem with that.

jasonLaster (approved):

PR #1621, opened 2021-01-28
Bring back avatars

### LINKED ISSUE #1592: Re-add Avatars
Now that we track sessions in hasura it will be easier to implement them

1. write a query for the recording's sessions
2. if the session has a user we can use their photo (new feature)
3. when the session is destroyed, we can remove the circle

voila!

jaril:
Fix #1592.

Behavior:
- Each session is represented by one avatar circle, except for the user's current session (tab).
- In case of multiple sessions per user, we still represent each session (tab) as one circle

image

### DIFF
--- a/src/ui/components/Header/Header.js
+++ b/src/ui/components/Header/Header.js
@@ -2,9 +2,7 @@ import React, { useEffect, useState } from "react";
 
 import { connect } from "react-redux";
 import { selectors } from "ui/reducers";
-import { actions } from "ui/actions";
 import Avatar from "ui/components/Avatar";
-import { groupBy } from "lodash";
 import { useGetActiveSessions } from "ui/hooks/sessions";
 import Title from "ui/components/shared/Title";
 import IconWithTooltip from "ui/components/shared/IconWithTooltip";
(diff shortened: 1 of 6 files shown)


src/ui/components/Header/Header.js
```diff
@@ -25,9 +27,37 @@ const GET_RECORDING_TITLE = gql`
...
+
+  const uniqueUsers = groupBy(data.sessions, session => session.user?.auth_id);
+  const displayedSessions = Object.keys(uniqueUsers).reduce((acc, key) => {
+    if (key == "undefined") {
+      return [...acc, ...uniqueUsers[key]];
+    }
+
+    return [...acc, uniqueUsers[key][0]];
+  }, []);
+  const displayedUsers = displayedSessions.map(session => session.user);
```
jasonLaster, line 45: this feels overly complicated...

why can't I do `uniqBy(sessions.map(session => session.user), user => user.id)`

jasonLaster: we'll also need a nice way to remove the logged in user

jasonLaster (approved):
Two small things, but looks great!

src/ui/components/Header/Header.js
jaril: Yes - this was done in a roundabout way because there's a way to differentiate session for users, but no way to differentiate sessions for guests.

There's three cases to consider:
1) Recording has me, you and two guests looking at it. 
- A: [Jaril, Jason, Guest1, Guest2]
- B: [Jason, Guest1, Guest2]
2) Recording has me, you (but opened in two tabs), and two guests are looking at it. 
- A: [Jaril, Jason, Jason, Guest1, Guest2]
- B: [Jason, Guest1, Guest2]
3) Recording has me, you, and one guest (but opened in two tabs) looking at it 
- A: [Jaril, Jason, Guest1, Guest1]
- B: [Jason, Guest1]

As a general rule, it's obvious that we shouldn't show two avatars for one user just because they have multiple sessions (1B, 2B). Problem is when it comes to guests – we can't tell the difference between 1 guest looking at 2 tabs, or 2 guests looking at 1 tab each. Both of those sessions come back as `user: null`.

I thought that we should just treat each guest session as a separate avatar (3B). Should we do differently?

PR #1622, opened 2021-01-28
Use token base preview matching

jasonLaster:
This uses the simpler expression matching and unfortunately updates a lot of the token location logic which was pretty unstable

### DIFF
--- a/src/devtools/client/debugger/src/actions/preview.js
+++ b/src/devtools/client/debugger/src/actions/preview.js
@@ -2,49 +2,27 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-//
-
 import { isConsole } from "../utils/preview";
-import { findBestMatchExpression } from "../utils/ast";
 import { getExpressionFromCoords } from "../utils/editor/get-expression";
 import { createPrimitiveValueFront } from "protocol/thread";
 
 import {
   getPreview,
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/actions/preview.js
```diff
@@ -2,49 +2,27 @@
...
-  }
-  return match;
-}
-
 export function updatePreview(cx, target, tokenPos, codeMirror) {
-  return ({ dispatch, getState, client, sourceMaps }) => {
+  return ({ dispatch, getState }) => {
     const cursorPos = target.getBoundingClientRect();
 
-    if (!isSelectedFrameVisible(getState()) || !isLineInScope(getState(), tokenPos.line)) {
```
jasonLaster, line 43: I preemptively got rid of `isLineInScope` because chrome doesn't use it and it is mostly just confusing. Also, we can drop the dependency on babel for previewing and simplify things.

src/devtools/client/debugger/src/utils/editor/token-events.js
```diff
@@ -17,25 +17,38 @@ function isValidToken(target) {
...
+    return false;
   }
 
   if (target.closest(".popover")) {
-    return true;
+    return false;
   }
 
-  return invalidTarget || invalidToken || invalidType;
+  return true;
```
jasonLaster, line 51: this just seems wrong... not sure what happened here

jaril (approved):

PR #1625, opened 2021-01-29
Use webpack for creating our image styles - second attempt (#1549)

hbenl:
The mask approach in #1615 created some styling regressions, so here's another attempt using `-webkit-mask-image`.


jasonLaster (approved):
Thanks!

PR #1660, opened 2021-02-02
Clean up logged out ux fixes #1630

### LINKED ISSUE #1630: Guests should not have a back button.
We currently show a back button in the top left that doesn't do anything.

We should instead encourage folks to sign in.

jasonLaster:
1. Remove the back button
2. Put a Replay logo there instead
3. Make the login link at the top right into a button
4. Sharing options shouldn't have an "invite collaborators" option

I paid a lot of attention to copying the button styles from the toggle :) 

Screen Shot 2021-02-01 at 6 51 34 PM

### DIFF
--- a/src/ui/components/Header/Header.css
+++ b/src/ui/components/Header/Header.css
@@ -99,13 +99,23 @@
 }
 
 button.login {
+  color: var(--primary-accent);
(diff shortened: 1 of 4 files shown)


jaril (approved):
Looks good to me!!

PR #1662, opened 2021-02-02
Typescriptify the Timeline components

hbenl:


src/ui/components/Timeline/Marker.tsx
```diff
@@ -36,17 +46,7 @@ class Marker extends React.Component {
     return true;
   }
 
-  getIsSecondaryHighlighted(hoveredPoint) {
```
jaril, line 39: Removing this would break similar secondary highlighting for related hits (from the same breakpoint) to the hovered/highlighted timeline marker.

Which I just noticed is already broken on master. Would you mind putting this  back and then applying this diff? That should fix secondary highlighting (demo below):

https://gist.github.com/jaril/f7787217e3d81721cf512b6a4a07d6b4

https://user-images.githubusercontent.com/15959269/106607608-8428c780-6531-11eb-9ffd-c68d837a47af.mov

hbenl: Done. Typescript had complained that this couldn't work (because `frame` was always undefined and hence the condition was always false) and so I had removed it.

jaril: Thanks!

jasonLaster (approved):

PR #1669, opened 2021-02-02
Don't show an Error when the socket closes because we're redirecting

hbenl:


src/main.js
```diff
@@ -69,6 +69,8 @@ async function initialize() {
   window.gToolbox = new DevToolsToolbox();
   store = await bootstrapStore();
 
+  window.addEventListener("beforeunload", socketWillClose);
```
jasonLaster, line 72: nit: i'd probably just put the listener is socket.js and have several listeners if we need it... not a big deal though

jasonLaster (approved):

src/main.js
hbenl: Good idea.

PR #1679, opened 2021-02-03
Move graphic logic to the timeline action

### LINKED ISSUE #1680: Hovering on a breakpoint timeline marker should update the video
**Description:**

When I was talking with Romain, he noticed that hovering on a breakpoint marker wouldn't update the video. To be honest, I thought he was crazy because we've obsessed on just this interaction, but he was right!

The fix is super easy too and will make the code more descriptive.

jasonLaster:
fixes #1680

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -88,7 +88,6 @@ class Timeline extends Component<PropsFromRedux> {
     const mouseTime = this.getMouseTime(e);
 
     if (hoverTime != mouseTime) {
-      paintGraphicsAtTime(mouseTime);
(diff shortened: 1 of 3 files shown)


src/ui/components/Timeline/index.tsx
```diff
@@ -88,7 +88,6 @@ class Timeline extends Component<PropsFromRedux> {
     const mouseTime = this.getMouseTime(e);
 
     if (hoverTime != mouseTime) {
-      paintGraphicsAtTime(mouseTime);
```
jasonLaster, line 91: moved an incidental graphics call to a timeline action so the logic would be in one place

jasonLaster:
I got a little carried away with the cleanup

Issue #1627, opened 2021-01-29
Devtools stay interactive after socket has closed


bhackett1024:
In some cases I've seen a message from the devtools after the remote connection has closed due to inactivity, but when I looked at a tab I left open overnight it still seemed responsive, though nothing requiring a remote connection worked and I could tell the backend session had closed long ago.

jasonLaster:
It looks like we are handling one of two socket events

- onSocketClose - we should 
- onSocketError - we assumed error implied inactivity, which feels wrong

https://github.com/RecordReplay/devtools/blob/master/src/protocol/socket.ts#L115-L130

PR #1690, opened 2021-02-03
Check if elem is defined (#1525)

hbenl:


hbenl:
This only happened in one session and, if I understand the sentry issue correctly, only after hours of inactivity, so this could be fixed with #1627. But adding a safety check doesn't hurt.

jasonLaster (approved):

PR #1695, opened 2021-02-03
protocol: if websocket connection errors show correct error message

### LINKED ISSUE #1465: If dispatch address is misconfigured when viewing a recording you get redirected to the recordings list
**Description:**
If devtools can't connect to the dispatcher when you are viewing a recording, it silently redirects you to the recordings list. 

**Steps to Reproduce:**
1. Visit https://replay.io/view?id=bbb1a8d2-3e34-4e17-bf5f-d67b0a252ae7&dispatch=ws://dispatch.replay.io

(note how I am overriding the dispatcher URL to use `ws` instead of `wss`, causing the issue)

**Expected Result:**
Some error is displayed saying that it can't connect to the dispatcher.

**Additional Context:**
I used Firefox for this, both the Repaly version and regular Firefox.

These messages are displayed in the JS console:

```
Assertion failed! instrument.js:109:45
Uncaught (in promise) DOMException: The operation is insecure. main.js:23
Uncaught (in promise) Error: Assertion failed!
    l utils.ts:53
    N apolloClient.ts:16
    i metadata.ts:89
    y metadata.ts:106
    t thunk.js:19
    c Redux
    nc App.js:69
    React 2
    unstable_runWithPriority scheduler.production.min.js:19
    React 3
    F scheduler.production.min.js:17
    onmessage scheduler.production.min.js:14
    EventHandlerNonNull* scheduler.production.min.js:13
    Webpack 13
utils.ts:53:10
```

I believe the "operation is insecure" error is the problem because if you [disable that check in Firefox](https://discord.com/channels/779097926135054346/795692282053132299/796119652202971187) the page loads just fine.

Here's [a recording](https://replay.io/view?id=e10913d6-161a-4697-929b-1f455ee02d99), but it doesn't seem to exhibit the same behavior.

jazzdan:
Fixes #1465

I think a lot of stuff has changed since the initial report. Now devtools does show an error when it can't connect to the backend, but it shows the `onSocketClose` error not the `onSocketError` error.

I made `onSocketError` also set `willClose` so that `onSocketClose` will not fire after it, overwriting the error message in the UI.

I also added a `console.error` that dumps the event, which I found helpful while debugging.

I tried to write a test but I had a really hard time mocking out `UIStore`

### DIFF
--- a/src/protocol/socket.ts
+++ b/src/protocol/socket.ts
@@ -135,6 +135,8 @@ function onSocketClose() {
 
 function onSocketError(evt: Event) {
   console.error("Socket Error", evt);
+  // If the socket has errored, the connection will close. So let's set `willClose`
+  // so that we show _this_ error message, and not the `onSocketClose` error message
   willClose = true;
   return ({ dispatch }: { dispatch: Dispatch<Action> }) => {
     log("Socket Error");
--- a/src/protocol/socket.ts
+++ b/src/protocol/socket.ts
@@ -41,7 +41,7 @@ export function initSocket(store: UIStore, address?: string) {
 
   socket.onopen = makeInfallible(onSocketOpen);
   socket.onclose = makeInfallible(() => store.dispatch(onSocketClose()));
-  socket.onerror = makeInfallible(() => store.dispatch(onSocketError()));
+  socket.onerror = makeInfallible((evt: Event) => store.dispatch(onSocketError(evt)));
   socket.onmessage = makeInfallible(onSocketMessage);
 }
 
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):

src/protocol/socket.ts
```diff
@@ -133,7 +133,9 @@ function onSocketClose() {
   };
 }
 
-function onSocketError() {
+function onSocketError(evt: Event) {
+  console.error("Socket Error", evt);
+  willClose = true;
```
jaril, line 138: Looks good! Can you leave a comment so it's obvious why we have to set `willClose = true` twice?

jaril (approved):

src/protocol/socket.ts
jazzdan: Yes!

Issue #1355, opened 2020-12-23
expanding different instance of same object in an evaluation expands all instances of that object's evaluations


karina-y:
**Description:**
when you log the same objects in two different places (ie app.store and store), expanding one expands both and collapsing one collapses both.

**Steps to Reproduce:**
1. go to a recording
2. pause anywhere the store variable is available and log it as an evaluation
3. pause elsewhere and log app.store as an evaluation
4. try expanding one of those evaluations
5. notice both expand
6. try collapsing one of those evaluations
7. notice both collapse

**Expected Result:**
only the evaluation you are clicking on should expand/collapse

**Replay:**
https://replay.io/view?id=286bdff0-a274-4212-ad6e-5e973d3db042

jasonLaster:
should be fixed now that we're not using the OI

PR #1727, opened 2021-02-07
Simplify object inspector

jasonLaster:
At the moment, the Object Inspector is heavily intertwined with redux. There were reasons to do that in firefox devtools, but those reasons no longer exist. Instead we should leverage the fact that the protocol helpers are much better here and talk directly with them.

One of the benefits of this approach is that loading one object will not affect other objects!


jasonLaster:
This is getting closer. The console mostly works... although at somepoint after expanding a node ~5 times it stops loading.

The scopes pane and editor preview need work, but probably not that much... 

I'm happy with how much simpler this is and how much faster the frontend will be simply because we can avoid unecessary react renders and redux selectors

jasonLaster:
This will fix #1355

hbenl:
Superseded by #1865

Issue #1702, opened 2021-02-04
Breakpoint panel shows 0 hits, but many console messages are shown


jasonLaster:
**Description:**
When I add a breakpoint, the panel shows 0 hits, but many console messages were shown.

1. go to http://localhost:8080/view?id=0f64a8ae-27df-47ec-9ec3-e57ffcbf7f7e
2. set a breakpoint in cursor.ts#639

**Expected Result:**

I see a breakpoint timeline with many hits, one per console message.

**Additional Context:**

* Screenshots:
* Video (Loom): https://www.loom.com/share/7fe03864e07a48a6bc91f9622ad0f0e0
* Replay: https://replay.io/view?id=d9215cc2-6591-4d96-88f7-fe4024a5c0e1

<img width="1498" alt="Screen Shot 2021-02-04 at 8 16 01 AM" src="https://user-images.githubusercontent.com/254562/106921820-3fcc3180-66c1-11eb-98fa-108aa271e8fb.png">

hbenl:
The reason for this is the same as for #1701, both components get their data from the same selector.

PR #1735, opened 2021-02-08
Refactor logpoint analyses

### LINKED ISSUE #1581: Logpoints should not be evaluated if the log is a string
At the moment the logpoint logic always does a `frame.evaluate` call regardless of what the logpoint is... This is silly since our default logpoint is just the name of the function. We don't need to evaluate `"foo"` to know the result is `"foo"`.

### LINKED ISSUE #1701: Breakpoint tooltip shows 4 different values (3 hits, 0 hits, 20 hits, 0 hits)
**Description:**
When I hover on the gutter the tooltip flashes several different values

**STR:**
1. go to http://localhost:8080/view?id=0f64a8ae-27df-47ec-9ec3-e57ffcbf7f7e
2. set a breakpoint in cursor.ts#639

**Expected Result:**

I see loading and then 0 hits, which is what the panel eventually shows

**Additional Context:**

* Screenshots:
* Video (Loom): https://www.loom.com/share/7fe03864e07a48a6bc91f9622ad0f0e0
* Replay: https://replay.io/view?id=d9215cc2-6591-4d96-88f7-fe4024a5c0e1

hbenl:
- fixes #1701, #1702
- runs only one analysis when there is more than one source for a URL

TODO:
- the second stage of the event logpoint analysis is not implemented yet because it currently doesn't work
- fix #1581

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1356,9 +1356,9 @@
       }
     },
     "@recordreplay/protocol": {
-      "version": "0.2.2",
-      "resolved": "https://registry.npmjs.org/@recordreplay/protocol/-/protocol-0.2.2.tgz",
-      "integrity": "sha512-TuMoL6R5nJT66fXqbyaxrMPGtX0uj7ZVPJh7RJuBY506dIuu22BA6NhuFdXtATR2PGsfMQ//pDjzfu89saRA6g=="
+      "version": "0.3.0",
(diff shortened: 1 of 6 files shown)


hbenl:
Issue for the second stage of the event logpoint analysis: RecordReplay/backend#1185

jasonLaster (approved):

jasonLaster:
I pushed a couple of small code tweaks, feel free to revert as well.

PR #1740, opened 2021-02-08
Extract random_walk

jasonLaster:
I extracted `randomItem`, but then realized I didn't know how to run the script at all 😉 

@bhackett1024 how do you use it?


bhackett1024:
See https://github.com/recordreplay/devtools/issues/1739

Issue #1743, opened 2021-02-09
Investigate memoizing MessageContainer + Timeline Marker


jasonLaster:
**Description:**

While looking into #1742 we found two other components that update when the user hovers on a message.

<img width="1493" alt="Screen Shot 2021-02-09 at 8 35 05 AM" src="https://user-images.githubusercontent.com/254562/107395591-cf068a00-6ab1-11eb-87f4-9ea00333c725.png">

**Steps to Reproduce:**

hover on a console message

PR #1748, opened 2021-02-10
Minimize the amount of rendering done when the user hovers over the console (#1743)

hbenl:


src/devtools/client/webconsole/components/Output/Message.js
```diff
@@ -17,17 +17,6 @@ const CollapseButton = require("devtools/client/webconsole/components/Output/Col
...
-  if (!message?.frame || !hoveredPoint?.location) {
-    return false;
-  }
-
-  const keyOne = getLocationKey(hoveredPoint.location);
-  const keyTwo = getLocationKey(message.frame);
-  return keyOne == keyTwo;
-}
 
 class Message extends Component {
```
jasonLaster, line 21: I wonder if we can promote Message to a `PureComponent` and drop `shouldComponentUpdate`

https://reactjs.org/docs/react-api.html#reactpurecomponent

src/devtools/client/webconsole/components/Output/MessageContainer.js
```diff
@@ -61,14 +63,16 @@ class MessageContainer extends Component {
...
+      "isSecondaryHighlighted",
+      "shouldScrollIntoView",
     ];
 
     return triggeringUpdateProps.some(prop => this.props[prop] !== nextProps[prop]);
   }
 
   render() {
-    const message = this.props.getMessage();
+    const message = this.props.message;
```
jasonLaster, line 75: `const {message} = this.props;

jasonLaster (approved):
Looks great

src/devtools/client/webconsole/components/Output/Message.js
hbenl: I was wondering that too, but I decided to try with minimal changes for now. We're passing a lot of props to the message components that I think may trigger unnecessary renders. But we should revisit this when we refactor the code.

src/ui/components/Timeline/index.tsx
```diff
@@ -184,20 +209,26 @@ class Timeline extends Component<PropsFromRedux> {
...
+              point={point.point}
+              time={point.time}
+              hasFrames={!!point.frame}
+              location={point.frame?.[0]}
+              currentTime={currentTime}
+              isPrimaryHighlighted={isPrimaryHighlighted}
+              isSecondaryHighlighted={isSecondaryHighlighted}
+              zoomRegion={zoomRegion}
+              overlayWidth={this.overlayWidth}
+              onSeek={() => {}}
```
jasonLaster, line 228: this seems dangerous for re-renders. But i assume it is okay

hbenl: Dangerous how?

jasonLaster: its a new object everytime, so could trigger a re-render

hbenl: Ah, yes, we should definitely fix this before making `Marker` a `PureComponent`.

PR #1778, opened 2021-02-11
Add error handling for Apollo queries/mutations

jaril:
Right now, we don't have a way to see query/mutation errors in Sentry. 

Those errors are not handled, which leads to bugs in the UI. Those bugs produce some type of error which ends up being caught by Sentry. But it makes it difficult to figure out what the root cause of the bug was in the first place.

This logs any errors so that when we see a query/mutation related bug in Sentry, we can check the logs to see what the original error was.

### DIFF
--- a/src/ui/hooks/comments.ts
+++ b/src/ui/hooks/comments.ts
@@ -92,8 +92,6 @@ export function useGetComments(
     pollInterval: 5000,
   });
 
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):

jasonLaster:
good call!

PR #1827, opened 2021-02-15
Seek to first comment or first meaningful paint

hbenl:
With this PR Replay will seek to the first paint point with a screen larger than 30000 bytes.
* we really want to seek to the first "meaningful paint". Choosing it by its jpeg size works for some recordings, but not all
* the current logic will wait for the recording to be scanned completely and then start looking at the screen sizes, this would be faster if we started while the recording is still being scanned

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -156,15 +156,12 @@ class _ThreadFront {
   off!: (name: string, handler: (value?: any) => void) => void;
   emit!: (name: string, value?: any) => void;
 
-  async setSessionId(sessionId: SessionId) {
+  setSessionId(sessionId: SessionId) {
     this.sessionId = sessionId;
     this.mappedLocations.sessionId = sessionId;
     this.sessionWaiter.resolve(sessionId);
 
     log(`GotSessionId ${sessionId}`);
-
-    const { endpoint } = await client.Session.getEndpoint({}, sessionId);
-    this.emit("endpoint", endpoint);
   }
 
   async initializeToolbox() {
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -166,9 +166,7 @@ class _ThreadFront {
 
   async initializeToolbox() {
     const sessionId = await this.waitForSession();
-    // const { endpoint } = await client.Session.getEndpoint({}, sessionId);
 
-    // this.timeWarp(endpoint.point, endpoint.time, /* hasFrames */ false, /* force */ true);
     await this.initializedWaiter.promise;
     this.ensureCurrentPause();
 
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -169,7 +169,7 @@ class _ThreadFront {
     // const { endpoint } = await client.Session.getEndpoint({}, sessionId);
 
     // this.timeWarp(endpoint.point, endpoint.time, /* hasFrames */ false, /* force */ true);
-    this.initializedWaiter.resolve();
+    await this.initializedWaiter.promise;
     this.ensureCurrentPause();
 
     if (this.testName) {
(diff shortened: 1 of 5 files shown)


jonbell-lot23:
See also: https://github.com/RecordReplay/devtools/issues/1670

So if there’s a comment, jump there. If not, jump to first meaningful paint.

src/protocol/thread/thread.ts
```diff
@@ -156,23 +156,20 @@ class _ThreadFront {
...
   }
 
   async initializeToolbox() {
     const sessionId = await this.waitForSession();
-    const { endpoint } = await client.Session.getEndpoint({}, sessionId);
+    // const { endpoint } = await client.Session.getEndpoint({}, sessionId);
 
-    this.timeWarp(endpoint.point, endpoint.time, /* hasFrames */ false, /* force */ true);
-    this.initializedWaiter.resolve();
+    // this.timeWarp(endpoint.point, endpoint.time, /* hasFrames */ false, /* force */ true);
```
jasonLaster, line 171: lets remove these comments

src/ui/actions/timeline.ts
```diff
@@ -32,14 +37,79 @@ export type TimelineActions =
...
+    });
+    const firstComment = firstCommentResult?.data?.comments?.[0];
+    if (firstComment) {
+      const { point, time, has_frames } = firstComment;
+      ThreadFront.timeWarp(point, time, has_frames);
+      return;
+    }
+
+    await paintPointsWaiter;
+    for (const paintPoint of gPaintPoints.slice(0, 10)) {
```
jasonLaster, line 88: this feels very soupy...

jasonLaster: i think graphics should hide most of this from redux... also others who import the protocol might want it.

jasonLaster:
This is fairly opinionated. Also untested, so feel free to push back

src/ui/utils/apolloClient.ts
```diff
@@ -2,10 +2,14 @@ import { ApolloClient, InMemoryCache, NormalizedCacheObject } from "@apollo/clie
 import { HttpLink } from "apollo-link-http";
 import { DocumentNode } from "graphql";
 import { assert } from "protocol/utils";
+import { getTest } from "ui/utils/environment";
```
hbenl, line 5: The intention was not to disable some graphql queries in tests but to to always seek to the endpoint in tests to make them more predictable.

src/ui/utils/apolloClient.ts
```diff
@@ -2,10 +2,14 @@ import { ApolloClient, InMemoryCache, NormalizedCacheObject } from "@apollo/clie
...
 
 let apolloClient: ApolloClient<NormalizedCacheObject> | null = null;
 
 export function query({ variables = {}, query }: { variables: any; query: DocumentNode }) {
+  if (!getTest()) {
```
hbenl, line 10: The intention was not to disable some graphql queries in tests but to to always seek to the endpoint in tests to make them more predictable.

src/ui/actions/timeline.ts
```diff
@@ -99,7 +149,7 @@ function setRecordingDescription({ duration, lastScreen }: RecordingDescription)
 
     // Paint the last screen to get it up quickly, even though we don't know yet
     // which execution point this is and have warped here.
-    paintGraphics(lastScreen);
+    // paintGraphics(lastScreen);
```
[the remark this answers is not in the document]
hbenl: Well, I thought jumping to the end of the recording and then jumping back to the first comment or meaningful paint could be confusing. But this `paintGraphics()` call doesn't do anything anyway because the description currently doesn't contain a `lastScreen` property

src/ui/actions/timeline.ts
hbenl: I like the cleanup you did there.

src/ui/utils/apolloClient.ts
jasonLaster: That makes sense

src/ui/actions/timeline.ts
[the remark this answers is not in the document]
jasonLaster: cool. then perhaps we should delete the code

jasonLaster (approved):

Issue #1769, opened 2021-02-11
Wrong screenshot is shown


hbenl:
When the mouse hovers over the timeline or the console, the screenshot for the hovered point is shown and when the mouse leaves the timeline or console the screenshot should revert to the one for the current pause. But sometimes this doesn't happen and the screenshot for the last hovered point remains.
I have a hunch this happens when the mouse leaves the timeline or console before the screenshot download finishes (I think I saw the screenshot for the current pause flash shortly before being replaced by the one for the last hovered point).
To reproduce this, open a recording and quickly move the mouse to a console message and then out of the console. Note that this only seems to happen the first time you hover over a message, for all subsequent hovers over the same message, the screenshot is correctly reverted.

jasonLaster:
Good find @hbenl. 

I've noticed this too. Lets record it!

hbenl:
https://replay.io/view?id=5e536de5-56ac-44c0-8f71-b362a0c9f072

PR #1837, opened 2021-02-16
Remove paintGraphicsAtTime (#1769)

hbenl:


src/ui/actions/timeline.ts
```diff
@@ -360,17 +366,8 @@ export function setHoveredPoint(hoveredPoint: HoveredPoint | null): UIThunkActio
...
-      return;
-    }
-
-    if (hoveredPoint) {
-      paintGraphicsAtTime(hoveredPoint.time);
-    } else {
-      paintGraphicsAtTime(selectors.getCurrentTime(getState()));
-    }
+    const updateGraphics = !selectors.getPendingComment(getState());
+    dispatch(setTimelineToTime(hoveredPoint?.time || null, updateGraphics));
```
jasonLaster, line 371: when would hoveredPoint not have a time?

src/ui/state/timeline.ts
```diff
@@ -33,8 +33,8 @@ export interface TimelineState {
 
 export interface HoveredPoint {
   target: "timeline" | "console" | "widget" | "transcript";
-  point: string;
-  time: number;
+  point?: string;
+  time?: number;
```
jasonLaster, line 37: i think all hovered points should have these properties unless there's a really good reason

src/ui/actions/timeline.ts
hbenl: https://github.com/RecordReplay/devtools/blob/3d8fb302472837ed29aa92b96193c2b28684a4d8/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js#L101-L106

src/ui/state/timeline.ts
hbenl: Well, if you hover over a breakpoint panel, you have a location but no singular time.

jasonLaster: got it... i don't think we should consider that a point.

hbenl: Yeah, the name `HoveredPoint` seems wrong here, but we still want the functionality when hovering over a breakpoint panel, right?

jasonLaster: haha we're debating that actually. But either way we want it when hovering on the gutter. 

I propose introducing a new hoveredLocation type and state. It's nice to keep this strict.

hbenl: > I propose introducing a new hoveredLocation type and state.

I don't think that's a good idea:
* location and point/time in `HoveredPoint` come from the same thing in the UI (the one you're hovering over)
* most parts of the code that use `HoveredPoint` read or set both location and point/time at the same time
* the `target` property of `HoveredPoint` would have to be duplicated - or stored separately, in which case we would have three separate entities in redux (coming from one thing in the UI) instead of one entity with 4 properties

PR #1847, opened 2021-02-17
Comment dropdown

### LINKED ISSUE #1774: Comments: Delete / Edit reply
**Description:**

It should be easy to delete/edit a reply

### LINKED ISSUE #1841: Comments: cannot delete replies to a comment
**Description:**
The option to delete a comment exists only on the top level comment. Replies should also have a delete option. It would make sense users can only delete their own replies, which is how it appears to currently work on top level comments.

**Steps to Reproduce:**
1. Open a recording
2. Enter a comment, save
3. Reply to that comment by clicking directly on it, save
4. Try to delete your reply

**Expected Result:**
Should the user find an error in their reply, they should have the option to delete it without removing the entire comment thread
**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:

jaril:
Fix #1841. Fix #1774.

Demo: https://share.descript.com/view/tzASNZz6z1P

image

### DIFF
--- a/src/ui/components/Comments/CommentEditor.js
+++ b/src/ui/components/Comments/CommentEditor.js
@@ -136,7 +136,6 @@ function CommentEditor({
 
   const handleSave = () => {
     if (editing) {
-      console.log("handleSave editing");
       handleExistingSave();
     } else if (isNewComment) {
       handleNewSave();
--- a/src/ui/components/Comments/CommentEditor.js
+++ b/src/ui/components/Comments/CommentEditor.js
@@ -8,6 +8,27 @@ import { actions } from "ui/actions";
 import CommentTool from "ui/components/shared/CommentTool";
 import "draft-js/dist/Draft.css";
 
+const moveSelectionToEnd = (editorState, DraftJS) => {
+  const { EditorState, SelectionState } = DraftJS;
+  const content = editorState.getCurrentContent();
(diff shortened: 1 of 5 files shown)


src/ui/components/Comments/CommentEditor.js
```diff
@@ -8,17 +8,41 @@ import { actions } from "ui/actions";
 import CommentTool from "ui/components/shared/CommentTool";
 import "draft-js/dist/Draft.css";
 
+const moveSelectionToEnd = (editorState, DraftJS) => {
```
jasonLaster, line 11: could this function live outside of the class?

src/ui/components/Comments/CommentEditor.js
```diff
@@ -86,16 +124,21 @@ function CommentEditor({
...
   const addComment = hooks.useAddComment(clearPendingComment);
-
-  const isNewComment = comment.content === "";
+  const updateComment = hooks.useUpdateComment(clearPendingComment);
+  const isNewComment = comment.content == "";
 
   const handleSave = () => {
-    if (isNewComment) {
+    if (editing) {
+      console.log("handleSave editing");
```
jasonLaster, line 139: 🧂

jasonLaster (approved):

PR #1878, opened 2021-02-22
Speed up analysis for gutter tooltips and logpoints

hbenl:


jasonLaster:
It looks like we kick off new requests when we add the breakpoint. This makes sense perhaps because it is difficult to cache given the filters are separate, but perhaps we can do some caching at the manager level?

jasonLaster (approved):

PR #1892, opened 2021-02-23
Remove console isSecondaryHighlighted

jasonLaster:


hbenl (changes requested):
There are more places containing `isSecondaryHighlighted` - should those also be removed?
Screenshot_2021-02-23_15-45-08

jasonLaster:
nope, it's fine to have them in the global timeline, just not in the console

hbenl (approved):

PR #1899, opened 2021-02-24
Show the correct hits for conditional breakpoints (#1033)

hbenl:
There is one issue left: when you change the condition, the breakpoint panel shows strange markers for a short while before changing to the correct markers.


jasonLaster:
this looks like part of the first PR

hbenl:
I made the other PR on top of this one because they're touching the same files. GitHub isn't very good at this, so this PR shows up as part of the other one.

jasonLaster:
got it, i'll review this first then

Issue #1880, opened 2021-02-22
Cache analysis for the same criteria


jasonLaster:
**Description:**

It looks like the client will make a separate analysis for gutter points and breakpoint timeline points

STR:

1. hover on the gutter (observe requests)
2. click on the gutter (observe new requests)
3. click on the gutter a couple more times (observe new requests)

PR #1906, opened 2021-02-24
Don't run an analysis if we already have the points and don't need the results (#1880)

hbenl:


src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.js
```diff
@@ -87,32 +82,32 @@ function BreakpointTimeline({
     <div className="breakpoint-navigation-timeline-container">
       <div
         className={classnames("breakpoint-navigation-timeline", { dimmed: shouldDim })}
-        ref={timelineNode}
```
jasonLaster, line 90: this is great

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.js
```diff
@@ -25,45 +25,17 @@ function hasSecondaryHighlighted({ hoveredItem, breakpoint }) {
...
-  const startTime = zoomRegion.startTime;
-  const endTime = zoomRegion.endTime;
-
-  // Percent offset values are [0, 100].
-  const unadjustedLeftPercentOffset = ((point.time - startTime) / (endTime - startTime)) * 100;
-  const unadjustedLeftPixelOffset = (unadjustedLeftPercentOffset / 100) * timelineWidth;
-  const leftPixelOffset = unadjustedLeftPixelOffset - markerWidth / 2;
-  const leftPercentOffset = (leftPixelOffset * 100) / timelineWidth;
-  return leftPercentOffset;
-}
```
jasonLaster, line 40: don't know how you're getting rid of this math, but i like it

src/protocol/logpoint.ts
```diff
@@ -93,14 +98,20 @@ async function showPrimitiveLogpoints(
...
-    PointHandlers.onPoints(points, { locations });
+function saveLogpointHits(
+  points: PointDescription[],
+  results: AnalysisEntry[],
+  locations: Location[],
+  condition: string
+) {
+  if (condition) {
+    points = points.filter(point =>
+      results.some(result => result.key === point.point && result.value.time === point.time)
```
jasonLaster, line 109: what is this doing?

src/ui/actions/app.ts
```diff
@@ -92,19 +90,6 @@ export function setupApp(recordingId: RecordingId, store: UIStore) {
...
-    locations.forEach(
-      (location: Location | null) => location && store.dispatch(setAnalysisPoints(points, location))
-    );
-  };
-
-  PointHandlers.addPendingNotification = (location: any) => {
-    store.dispatch(setPendingNotification(location));
-  };
-}
-
```
jasonLaster, line 107: glad we're moving this

src/ui/reducers/app.ts
```diff
@@ -85,7 +85,7 @@ export default function update(
     }
 
     case "set_analysis_points": {
```
jasonLaster, line 87: do you think it's time to add an analysis reducer?

jasonLaster (commented):
Left some thoughts. Will test soon

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.js
hbenl: [That's how](https://github.com/hbenl/webreplay-devtools/blob/55ffa9c433bd82eb2e31faa85ff5b6fe6c06c904/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.js#L67). CSS is awesome.

src/protocol/logpoint.ts
hbenl: Filtering out the `PointDescription`s referring to points where the condition is false.

src/ui/reducers/app.ts
hbenl: Let's do that when it gets more complex.

jasonLaster:
works for me

Sent via Superhuman ( https://sprh.mn/?vip=jason.laster.11@gmail.com )

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.js
```diff
@@ -89,7 +64,7 @@ function BreakpointTimelinePoint({
       })}
       title={`${index + 1}/${analysisPoints.length}`}
       onClick={() => seek(point.point, point.time, true)}
-      style={{ left: `${leftPercentOffset}%` }}
+      style={{ left: `calc(${leftPercentOffset}% - ${pointWidth / 2}px)` }}
```
jaril, line 67: I think we could do this even simpler. To get the same half-of-width-displacement effect:
```
.breakpoint-navigation-timeline-point {
    transform: translateX(60px);
}
```
That way JS doesn't ever need to know about `pointWidth` in the first place. Reduces the code here to just `style={{ left: `${leftPercentOffset}%` }}`

jaril (commented):
Left two comments on minor nits, but looks good!

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.js
hbenl: That was my first idea, but then I saw that the javascript variable is used in 2 other places, so it's probably a good idea to have a single source of truth here.

Issue #1843, opened 2021-02-16
The video player is slow the first time


jasonLaster:
**Description:**

When users go to view the recording for the first time, the video player is slower the first time while the cache is filing up

jasonLaster:
https://github.com/RecordReplay/devtools/issues/1843

jasonLaster:
We have logic to keep the timeline moving, but in practice it looks like it is failing

https://github.com/RecordReplay/devtools/blob/master/src/ui/actions/timeline.ts#L286-L303

I think we should either keep the progressbar moving... even if the backend can't find graphics...
or if we're lagging show an indicator in the video in the bottom right...

jonbell-lot23:
If I'm understanding correctly, we have three options:

1. Don't even load the video until we know everything can load smoothly
2. Optimise for progress bar always being smooth, but explain why the playback region is sometimes not loaded
3. Optimise for playback region always being accurate, but explain why the progress bar is sometimes lagging

Do we have the data to show streaming progress the way YouTube does? If so, I propose this:

<img width="1222" alt="image" src="https://user-images.githubusercontent.com/9154902/108560338-76f93e80-7361-11eb-876d-ea44232f93e4.png">

* Introduce #F8F8F8 as a light grey to use for unloaded timeline content
* Keep the same grey for the loaded part of the timeline (#D7D7DB I believe?)
* Keep the same primary-accent blue for the current playhead

jasonLaster:
Following up based on a call with @hbenl 

1.  Knowing how much we buffered from where we are could be tricky. it might be easier to show a spinner when the video player is waiting for frames.
2. On the backend, we might be able to make paints consistently fast by reading from disk

hbenl:
@jaril With PR #1910 you can query whether playback is stalled and show this to the user.

hbenl:
#1911 is something I had prototyped last week and I'm putting out now in case someone wants to play with it: It precaches screenshots for playback (but stops when the user switches to "dev" mode because the downloads could degrade the performance of other interactions) and reports the progress to the redux store, so that [this UI](https://github.com/RecordReplay/devtools/issues/1843#issuecomment-782342002) becomes possible.
Note that this PR has some limitations: the precaching isn't ever resumed once it is stopped because the user changed to non-dev mode. Also, the progress variable only indicates the precaching progress, more screenshots may already have been downloaded.

jonbell-lot23:
[...]

PR #1911, opened 2021-02-25
Precache graphics (#1843)

hbenl:
This will start precaching graphics for playback as long as the user stays in non-dev mode. The precaching progress can be queried with the `getPlaybackPrecachedTime()` selector.

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -12,8 +12,9 @@ import {
   findPaintsResult,
 } from "@recordreplay/protocol";
 import { client } from "./socket";
-import { actions, UIStore } from "ui/actions";
+import { actions, UIStore, UIThunkAction } from "ui/actions";
 import { Canvas } from "ui/state/app";
+import { selectors } from "ui/reducers";
 
 export const screenshotCache = new ScreenshotCache();
 
@@ -225,14 +226,6 @@ export async function getGraphicsAtTime(
     ? screenshotCache.getScreenshotForPlayback(point, paintHash)
     : screenshotCache.getScreenshotForPreview(point, paintHash);
 
-  if (forPlayback) {
-    // Start loading graphics at nearby points.
-    for (let i = paintIndex; i < paintIndex + 5 && i < gPaintPoints.length; i++) {
-      const { point, paintHash } = gPaintPoints[i];
-      screenshotCache.getScreenshotForPlayback(point, paintHash);
-    }
-  }
-
   const screen = await screenPromise;
 
   let mouse: MouseAndClickPosition | undefined;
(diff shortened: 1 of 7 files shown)


jasonLaster:
@hbenl are you planning on picking this up? I think video perf improvements could be nice. 

Also I wonder if brian would have insights into why it might be slow. Especially if we have finished the basic scan...

src/protocol/graphics.ts
```diff
@@ -423,3 +416,43 @@ export async function getFirstMeaningfulPaint(limit: number = 10) {
...
+    const startIndex = mostRecentIndex(gPaintPoints, startTime);
+    if (startIndex === undefined || precacheStartIndex === startIndex) {
+      return;
+    }
+    precacheStartIndex = startIndex;
+
+    let endIndex = mostRecentIndex(gPaintPoints, startTime + precacheMinTime) || startIndex;
+    endIndex = Math.max(endIndex, startIndex + 1 + precacheMinCount);
+    endIndex = Math.min(endIndex, gPaintPoints.length);
+    for (let i = startIndex + 1; i < endIndex; i++) {
```
kannanvijayan, line 440: The single-paint increment here is going to issue a lot of fetches during precaching.  I want to suggest computing the indexes of the loop by doing a fixed time-offset (e.g. 10ms increments), and recomputing the gPaintPoints index for each of those times, and then caching the images at those times only.

The only problem with that is synchronizing this with the playback engine so that later precache attempts for overlapping time segments also query the same points.

Can we have a global PLAYBACK_SNAP variable for now, set to 50ms?  Then, within `getGraphicsAtTime`, we snap the time as follows:

```
if (forPlayback) {
  time += PLAYBACK_SNAP - (time % PLAYBACK_SNAP);
}
...
```

This will ensure that for playback, every recording of some length will always request the same paint indexes.

kannanvijayan (changes requested):

src/protocol/graphics.ts
```diff
@@ -423,3 +416,66 @@ export async function getFirstMeaningfulPaint(limit: number = 10) {
...
+    if (precachedTime > selectors.getPlaybackPrecachedTime(getState())) {
+      dispatch(actions.setPlaybackPrecachedTime(precachedTime));
+    }
+  };
+}
+
+// Snap time to 50ms intervals, snapping up.
+const snapInterval = 50;
+export function snapTimeForPlayback(time: number) {
+  return time + snapInterval - (time % snapInterval);
```
kannanvijayan, line 480: I just realized I made the same error in my code, but this will align `0` up to `50`, and we probably want `0` to map to `0`.  I can't think if a super-clever expression for this, so my best idea is:

```
const rem = time % snapInterval;
return time + (rem > 0 ? snapInterval - rem : 0);
```

kannanvijayan (commented):
Looks good aside from noted minor issue.

kannanvijayan:
I meant to approve this review, but it seems to have left it at "change requested".

PR #1919, opened 2021-02-26
Add a Privacy note

jasonLaster:
https://github.com/RecordReplay/backend/issues/1320

Screen Shot 2021-02-25 at 6 50 15 PM


jaril (approved):
One last nit — if a non-author opens up the dropdown, we don't see the public/private toggle, but we see the notice. Is this intentional? If not, we can make it so that the notice only shows up when there's a public/private toggle.

Demo: https://share.descript.com/view/axtMiRk6doh

jasonLaster:
good call.

PR #1920, opened 2021-02-26
Show example replays by in the library fixes #1915

### LINKED ISSUE #1915: Improve empty library UI
The empty library ui is causing almost everyone to spend time reading....

Worse yet, if they dont use replay, they wont be able to create a replay.

Instead, I propose we show folks a couple of good default replays.

### LINKED ISSUE #1921: Library: Only let creators delete a recording
**Description:**

I think anyone can delete a recording from the library... We should lock that down.

jasonLaster:
* shows example recordings
* hides the settings dropdown for non-owners fixes #1921 

Screen Shot 2021-02-25 at 7 32 29 PM


jaril (approved):
Looks good!

Demo: https://share.descript.com/view/PphzMNwyEph

PR #1923, opened 2021-02-26
Stop using auth_id to identify users

hbenl:


src/ui/utils/tokenManager.tsx
```diff
@@ -133,7 +138,10 @@ class TokenManager {
       try {
         const token = await this.fetchToken(refresh);
 
-        this.setState({ token }, deferredState);
+        const decodedToken = jwt_decode<any>(token);
+        const userId = decodedToken?.["https://hasura.io/jwt/claims"]?.["x-hasura-user-id"];
+
+        this.setState({ token, userId }, deferredState);
```
jasonLaster, line 144: I think it might be nice to make this a bit more general. My thinking is that in the future there might be other claims and it is nice for the tokenManager, to not expose arbitrary fields

```
const claims = { hasura: { userId:  decodedToken?.["https://hasura.io/jwt/claims"]?.["x-hasura-user-id"] } }
this.setState({ token, claims }, deferredState);
```

src/ui/components/Header/ShareDropdown.js
```diff
@@ -27,32 +27,30 @@ const GET_RECORDING_PRIVACY = gql`
...
-      }
+      user_id
     }
   }
 `;
 
 function useIsOwner(recordingId) {
-  const { user, isAuthenticated } = useAuth0();
-  const { data, loading } = useQuery(GET_OWNER_AUTH_ID, {
+  const { userId } = useToken() || {};
```
jasonLaster, line 39: could `useToken` always return an object to be safer?

src/ui/components/Header/ShareDropdown.js
```diff
@@ -113,17 +111,16 @@ function Privacy({ isPrivate, toggleIsPrivate }) {
 }
 
 function Collaborators({ setExpanded, setModal }) {
-  const { isAuthenticated } = useAuth0();
+  const tokenState = useToken();
+  if (!tokenState || !tokenState.userId) {
+    return null;
+  }
```
jasonLaster, line 117:
```
const {userId} = useToken()
if (!userId) {
  return null;
}
```

jasonLaster (approved):
A couple of small tweaks, but otherwise great

src/ui/components/Header/ShareDropdown.js
hbenl: Right now `undefined` means auth0 is still loading, whereas `{}` means the user didn't try to authenticate (and `{ error: ... }` means authentication failed), but that could be changed of course.

src/ui/utils/tokenManager.tsx
hbenl: I like the idea

PR #1926, opened 2021-02-27
Update devtools to new per-session analysis commands.

loganfsmyth:
We'll have to coordinate landing this at the same time as the backend update, but I think that's fine, it seems too difficult to try to make this change in a backward-compatible way.


jasonLaster (approved):

jasonLaster:
It's nice to see TS complain

PR #1929, opened 2021-02-28
Fix sharing modal fixes #1864

### LINKED ISSUE #1864: Open Sharing preferences is empty for accounts

jasonLaster:


src/ui/state/app.ts
```diff
@@ -32,6 +32,7 @@ export interface AppState {
   expectedError: ExpectedError | null;
   unexpectedError: sessionError | null;
   modal: ModalType | null;
+  modalOptions: Object | null;
```
hbenl, line 35: Why not `modalOptions: { recordingId: RecordingId } | null;` ?

hbenl (changes requested):
Please use a proper type for `modalOptions`.

Issue #1876, opened 2021-02-20
Publish protocol directory to NPM


jasonLaster:
**Description:**

We currently publish the protocol to NPM as a package, which provide low-level types. It would be nice to also publish our devtools' protocol to NPM so that folks could easily build their own clients that have similar functionality.

**Benefits:**

1. Create a socket
2. Fetch graphics
3. Perform analysis 
4. Fetch data from fronts
5. map locations

**Logistics**

1. Add a package.json, with a name. Perhaps `devtools-client`?
2. Add a readme.md, with instructions on how to use the client. At some point, we can add an examples directory that can be exercised in CI.

jasonLaster:
closing as this is not a priority

PR #1935, opened 2021-02-28
Publish protocol as a separate NPM package that can be re-used #1876

jasonLaster:
One of the things that comes up when i talk with others is that many people would like to experiment with building their own frontends.

This is one of our goals and is something we should definitely encourage. I think there are several things we can do to make this process easy

1. Build a community + ecosystem (discord and friends)
2. Publish documentation and examples
3. Publish packages that streamline the process
4. Make it easy for folks to hack on the devtools frontend so they can quickly prototype ideas

**Devtools Client**

Our devtools client does a bunch of things pretty well. 

1. wrap graphics so that it is easy to show a video
2. wrap analysis so that it is easy to add logpoints
3. ...

If we publish this as a package others, can easily get started. It will also help us ensure a nice separation of concerns between our redux store for the app and what the protocol needs.

This PR is mostly cleanup, which helps regardless:

1. move the redux logic out, which arguably should never have been in there
2. small import tweaks

### DIFF
--- a/src/protocol/thread/style.ts
+++ b/src/protocol/thread/style.ts
@@ -1,7 +1,11 @@
 import { StyleDeclaration } from "@recordreplay/protocol";
 import { Pause, WiredObject } from "./pause";
 import { assert, DisallowEverythingProxyHandler } from "../utils";
-const { ELEMENT_STYLE } = require("devtools/client/inspector/rules/constants");
+
+// The PageStyle actor flattens the DOM CSS objects a little bit, merging
+// Rules and their Styles into one actor. For elements (which have a style
+// but no associated rule) we fake a rule with the following style id.
+const ELEMENT_STYLE = 100;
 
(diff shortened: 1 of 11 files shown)


src/protocol/thread/style.ts
```diff
@@ -1,7 +1,11 @@
 import { StyleDeclaration } from "@recordreplay/protocol";
 import { Pause, WiredObject } from "./pause";
 import { assert, DisallowEverythingProxyHandler } from "../utils";
-const { ELEMENT_STYLE } = require("devtools/client/inspector/rules/constants");
+
+// The PageStyle actor flattens the DOM CSS objects a little bit, merging
+// Rules and their Styles into one actor. For elements (which have a style
+// but no associated rule) we fake a rule with the following style id.
+const ELEMENT_STYLE = 100;
```
jasonLaster, line 8: this probably should be exposed by the protocol since its coming from the backend

PR #1941, opened 2021-03-01
Use the user_id_by_email function to fetch the user_id for an email

hbenl:
We want to restrict access to the users table, but when a user wants to add another user as a collaborator, these restrictions should not apply (otherwise he could never add collaborators that he hasn't collaborated with before). For this purpose I have created a user defined function called user_id_by_email in the database that allows us to fetch the user_id for a given email.
When this PR has landed, we can add the restrictions to the users table.

### DIFF
--- a/src/ui/components/shared/SharingModal.js
+++ b/src/ui/components/shared/SharingModal.js
@@ -35,9 +35,8 @@ const GET_OWNER_AND_COLLABORATORS = gql`
 
 const GET_COLLABORATOR_ID = gql`
   query GetCollaboratorId($email: String = "") {
-    users(where: { email: { _eq: $email } }) {
+    user_id_by_email(args: { email: $email }) {
       id
-      email
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/SharingModal.js
```diff
@@ -171,12 +170,16 @@ function EmailForm({ recordingId }) {
   if (status.type === "fetched-user") {
     if (status.error) {
       return <ErrorHandler message={"We can not fetch that collaborator right now."} />;
-    } else if (status.data.users.length === 0) {
+    } else if (status.data.user_id_by_email.length === 0) {
```
jasonLaster, line 173: As a follow up, i'm not thrilled with how the graphql data structures leak into our UI. It would be nice to have tighter formatting closer to the queries so that we can normalize this data a bit more.

So for instance, by the time we're in this component it would be nice to expect a user record that had an ID.

jasonLaster (approved):

PR #1948, opened 2021-03-02
Fix the permission check in RecordingGridItem

hbenl:
The permission check concluded that the user is the owner of all recordings that he can see because `undefined == undefined` (we don't use `auth_id` to identify users anymore).


jasonLaster (approved):
Yikes. Guess this was in flight during the auth_id change

PR #1949, opened 2021-03-02
Add inline replies

### LINKED ISSUE #1887: [UX] Add inline replies (⚙ Code ready)
image

Figma page: Inline replies

Notes:
* This isn't pixel perfect yet -- let's just get this in code so we can see how it feels with the 8px grid work
* This behaviour should work for all events, not just replies. For example, if I click on a Mouse Click event, I should see this same pattern for attaching a comment to it. We might even end up removing the add comment button entirely!

jaril:
Fix #1887.

Demo: https://share.descript.com/view/dPioAlJFZd0


jasonLaster (approved):
Couple of small questions.


Lets also drop the Adding Comment UI... It's distracting and causes a jump
Screen Shot 2021-03-02 at 2 58 35 PM

jaril:
Updated demo: https://share.descript.com/view/jKdlSvhP8U0

Issue #1814, opened 2021-02-14
[Playwright] Breakpoint messages do not appear


bhackett1024:
While working on a new experimental playwright test I ran into a problem where if we simulate a click on the gutter to add a breakpoint with no hover first, the hovering logic that runs afterwards will clear the logpoint ID for the line and the breakpoint messages do not get added to the console.  See the recording below, made using playwright.  Unfortunately there are some problems figuring out the problem from the replay, I'll file separate issues for those.

https://replay.io/view?id=36dc99a2-9bf0-4314-b743-0ccda8c549c8

bhackett1024:
https://github.com/RecordReplay/devtools/issues/1815 covers some of the many problems I ran into while looking into this.

jasonLaster:
@bhackett1024 what priority do you think this should be?

bhackett1024:
While I've only seen this with playwright, it certainly seems like it could be triggered by a user and lead to weird behavior.  It's also a problem with how the devtools are using logpoint IDs which could manifest in other ways, so it would be nice if it was fixed soon.

PR #1980, opened 2021-03-04
Fix breakpoint messages not appearing in playwright tests (#1814)

hbenl:
To be honest, I don't fully understand what's going on here, but this seems to fix the issue.

### DIFF
--- a/src/devtools/client/debugger/src/client/commands.js
+++ b/src/devtools/client/debugger/src/client/commands.js
@@ -150,7 +150,6 @@ function removeBreakpoint(location) {
 }
 
 function runAnalysis(location, options) {
-  maybeClearLogpoint(location);
   options = maybeGenerateLogGroupId(options);
   const { condition, logValue, logGroupId } = options;
   const { line, column, sourceUrl, sourceId } = location;


jasonLaster:
Hmm, this feels like a potentially risky fix. Could you look into it a bit more, perhaps with the failing replay test?

hbenl:
I don't know what the removed line was supposed to do: it is only called when the user hovers over a line number and then it checks if there is already a logpoint on that line and, if so, removes all console messages from that logpoint. That would be a bug and the only reason it was never noticed is that there is a check that this function shouldn't be called if we already have the analysis points for that line - which should *always* be the case if there is a logpoint on that line.
To summarize: it seems to me the removed code would never do anything due to a performance optimization in the calling function and if that optimization wasn't there, it would lead to a bug (I have actually checked this by commenting out [this line](https://github.com/RecordReplay/devtools/blob/2839270d32e0ac98bdd64e7b790c3fe8b7ec47a1/src/devtools/client/debugger/src/actions/breakpoints/modify.js#L177) and recording the result: https://replay.io/view?id=219dd9eb-9cf7-48a3-aad6-8bffb6bc1ae8).

jasonLaster:
👍 great writeup. Sorry i missed this earlier

jasonLaster (approved):

PR #2007, opened 2021-03-05
Refactor Transcript components

jaril:
https://share.descript.com/view/80Fgeyo6TNQ

Screenshot
image


jasonLaster (approved):

src/ui/components/Transcript/TranscriptItem/PauseTranscriptItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+    };
+    getSourceLocation();
+  }, [item]);
+
+  if (sourceLocation) {
+    const filename = getFilenameFromURL(sourceLocation.sourceUrl);
+    icon = "document-text";
+    label = `${filename}:${sourceLocation.line}`;
+    secondaryLabel = getTextAtLocation(state, sourceLocation.sourceId, sourceLocation);
+  }
```
jasonLaster, line 41: this causes a bit of a delay some times... so i think i would probably choose a different set of defaults like `loading...` but that is a nit

PR #2016, opened 2021-03-06
Adjust ViewToggle border radius and do some component clean up

achou11:
Notes:

- Adjusts the border radius for the sliding pill to better match the container's border radius
  - Before:
  
    image
  
  - After:
    
    image

- There was a bug that caused the pill's border radius to be unset after clicking the first time
    - Before:
    
      view-toggle-border-radius-disappear
      
    - After:
 
      view-toggle-border-radius-no-disappear
      
- Prevents a class name of `null` from rendering for the inactive view mode button
- Only attempt to load Framer  on component mount instead of on every rerender

### DIFF
--- a/src/ui/components/Header/ViewToggle.js
+++ b/src/ui/components/Header/ViewToggle.js
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from "react";
 import { connect } from "react-redux";
+import classnames from "classnames";
 import "./ViewToggle.css";
 import { setViewMode } from "../../actions/app";
 import { getViewMode } from "../../reducers/app";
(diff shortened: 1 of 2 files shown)


src/ui/components/Header/ViewToggle.js
```diff
@@ -39,13 +39,7 @@ function ViewToggle({ viewMode, setViewMode }) {
...
   const { motion, AnimateSharedLayout } = framerMotion;
 
-  const onClick = () => {
-    if (viewMode == "dev") {
-      setViewMode("non-dev");
-    } else {
-      setViewMode("dev");
-    }
-  };
+  const onClick = () => setViewMode(viewMode === "dev" ? "non-dev" : "dev");
```
achou11, line 42: Could add a useCallback here but seems like overkill 🤷

jasonLaster: 😉

jonbell-lot23:
High five, @achou11 ! We should make a little badge for people like you who contribute! We can call it the Andrew Badge 😂

PR #2017, opened 2021-03-07
Add initial emoji support to comment renderer

ryanjduffy:
Adds emoji renderer to comment markdown.

I added an initial map of "regular" emojis to the text version used by `emoji-dictionary`. Should be expandable to more values.


jasonLaster:
First of all, I love this. I wonder if we could take it a couple of steps further

1. remark-emoji looks similar, but perhaps better abstracted (https://github.com/rhysd/remark-emoji)
2. I'd love to include an emoji autocomplete to encourage the practice

This is probably similar to other features like cmd+b adding `**<cursor>**`

ryanjduffy:
Nice! I should have looked for an existing lib :) I was lured in by the call of the regex. I'll take another look at it. might do a bit of refactoring to move draft-js and the markdown renderer into a shared component so it can be used throughout the app (e.g. in a description field for a replay, user profile stuff)

jasonLaster:
heh - i'd be happy to first get it into comments and then see where we go

ryanjduffy:
Quick update: I have draft-js + emoji autocomplete working but the styling overrides aren't complete yet. Will probably get back to it later tonight.

jasonLaster:
Thanks @ryanjduffy out of curiosity. Where did things land with draft-js's emoji plugin? it'd be nice to just use their defaults...

ryanjduffy:
It's mostly working. I have everything rendering but the popup for autocomplete is being clipped by the comment region scroller so I need to do a bit more work to ensure it stays visible.

ryanjduffy:
Ok. Sorted it out. Ready for review!

Issue #1683, opened 2021-02-03
Console evaluations do not respect selected async frame


jasonLaster:
**Description:**
I just looked into a problem where the selected frame, should have worked, but instead i had to seek to that location t do a proper evaluation. When I looked at the replay, it was clear we're passing the frameActor to the appropriate protocol methods, but it did not work.

STR:
1. go to https://replay.io/view?id=6d8b6839-1afa-478c-8d5d-9b587bdec371
2. set a breakpoint in actions/preview.js#25
3. select frame 2 `thunk.js:19`
4. evaluate `action` in the console

Replay:

https://replay.io/view?id=6c145aef-52a9-48a0-8297-7d417675821f

loganfsmyth:
@jasonLaster Is the issue here that the consume hard-codes `asyncIndex: 0` here: https://github.com/RecordReplay/devtools/blob/ed00251d7b3d90245e7a47213d1fa07740d69241/src/devtools/client/webconsole/actions/input.js#L64

Since you're selecting a frame across an async boundary, it's not just another frame that you're evaling in, it's an earlier pause too.

jasonLaster:
good call. Just left a comment in the replay at that callsite (
https://replay.io/view?id=6c145aef-52a9-48a0-8297-7d417675821f)

Do you think if we pass the right asyncIndex it will work? We could also
get the relevant pauseId, but i guess that's the point of the asyncIndex...

bhackett1024:
One thing to keep in mind here is that the pauses for the async parents are at an earlier point from the thread's current pause point, which is used to update the timeline, console position etc.  Adding evaluated messages at the current point in the console would put them in the wrong order wrt other messages.  This isn't entirely a new issue as e.g. the scopes panel also reflects that earlier time.  It's not an easy problem to fix from a UX perspective and might be too minor to care about right now.

loganfsmyth:
> Do you think if we pass the right asyncIndex it will work? We could also get the relevant pauseId, but i guess that's the point of the asyncIndex...

Yeah either way works, at the end of the day you need a `pauseId`/`frameId` pair to know where/when to evaluate.

PR #2030, opened 2021-03-08
Fix evaluations in async frames (#1683)

hbenl:


jasonLaster:
@hbenl This looks good. Could you write a couple of end-to-end tests to recreate these scenarios for async frames and evaluations without frames?

hbenl:
I added a test for evaluations in async frames, including the ordering of these evaluations in the console. Evaluations without frames are already tested in `console_eval.js` (that's how I noticed the bug).

Issue #2047, opened 2021-03-09
Scopes not updated when changing frames 


hbenl:
Scopes are not always updated when switching between frames (especially when the frames have the same variables, e.g. when switching between calls of a recursive function): http://localhost:8080/view?id=436bf418-fcdc-420b-852f-105e3025e680

PR #2048, opened 2021-03-09
Ensure that scopes are rerendered when switching between frames (#2047)

hbenl:


jasonLaster:
This would be another area where we should add an e2e test if it is reproducible.

Generally speaking, we should try and ensure the core debugging functionality is robust.

PR #2054, opened 2021-03-09
Add test that scopes are rerendered when switching frames

hbenl:


hbenl:
The failed test passes for me locally.

PR #2064, opened 2021-03-10
Include pauses in the current url fixes #993

### LINKED ISSUE #993: Replay urls should include the current pause
I'd love for the current pause location to be in the URL... i think we might support part of this now

jasonLaster:


src/ui/actions/timeline.ts
```diff
@@ -59,23 +60,44 @@ export async function setupTimeline(recordingId: RecordingId, store: UIStore) {
...
-      ({ point, time } = firstComment);
-      hasFrames = firstComment.has_frames;
-    } else {
-      const firstMeaningfulPaint = await getFirstMeaningfulPaint(10);
-      if (firstMeaningfulPaint) {
-        ({ point, time } = firstMeaningfulPaint);
-      }
+  const initialPausePoint = await getInitialPausePoint(recordingId);
+  if (initialPausePoint) {
+    if (initialPausePoint) {
```
hbenl, line 65: :smile:

src/ui/actions/timeline.ts
```diff
@@ -186,6 +208,22 @@ export function setZoomRegion(region: ZoomRegion): SetZoomRegionAction {
...
+}: {
+  point: ExecutionPoint;
+  time: number;
+  hasFrames: boolean;
+}) {
+  const url = new URL(window.location.toString());
+  url.searchParams.set("point", point);
+  url.searchParams.set("time", `${time}`);
+  url.searchParams.set("hasFrames", `${hasFrames}`);
+  window.history.pushState({}, "", url.toString());
```
hbenl, line 224: we're using `react-router`, is it safe to mix that with direct access to `window.history`?

src/ui/utils/environment.ts
```diff
@@ -29,3 +29,20 @@ export function isDeployPreview() {
 export function hasLoadingParam() {
   return url.searchParams.get("loading") != null;
 }
+
+export function getPausePointParams() {
+  const pointParam = url.searchParams.get("point");
+  const point = `${pointParam}`;
```
hbenl, line 35: You'll potentially end up with a param `point=null`, I'm pretty sure a URL with such a param (and hence trying to timeWarp to the ExecutionPoint `null`) would cause problems.
I think you should change the condition in line 43 to `pointParam && timeParam` instead.

hbenl (changes requested):
LGTM except for some small changes.

src/ui/actions/timeline.ts
jasonLaster: I think so... I looked into a tighter integration briefly, but nothing was obvious...

PR #2075, opened 2021-03-10
Use function name as transcript entry label where applicable

jaril:
Demo: https://share.descript.com/view/3wMqRcezGVn

Refresh demo: https://share.descript.com/view/38wpxyu7jha
- Refers to the changes in commit: `Handle possible stale floatingItem state due to async code`

### DIFF
--- a/src/ui/components/Transcript/TranscriptItem/FloatingTranscriptItem.tsx
+++ b/src/ui/components/Transcript/TranscriptItem/FloatingTranscriptItem.tsx
@@ -1,43 +1,44 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
 import { connect, ConnectedProps } from "react-redux";
 
 import CommentThread from "ui/components/Comments/TranscriptComments/CommentThread";
 import TranscriptItem from "./TranscriptItem";
-import { PendingComment, PauseItem, SourceLocation } from "ui/state/comments";
-import { ThreadFront } from "protocol/thread";
+import { FloatingItem } from "ui/state/comments";
 import { UIState } from "ui/state";
+import { selectors } from "ui/reducers";
 
 const { getFilenameFromURL } = require("devtools/client/debugger/src/utils/sources-tree/getURL");
 const { getTextAtLocation } = require("devtools/client/debugger/src/reducers/sources");
 const { getFormattedTime } = require("ui/utils/timeline");
+const { findClosestFunction } = require("devtools/client/debugger/src/utils/ast");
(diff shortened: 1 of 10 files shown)


src/ui/components/Transcript/TranscriptItem/FloatingTranscriptItem.tsx
```diff
@@ -1,43 +1,44 @@
...
-  if (sourceLocation) {
-    const filename = getFilenameFromURL(sourceLocation.sourceUrl);
+  if (item.location) {
+    // const { sourceUrl, line } = item.location;
+    // const filename = getFilenameFromURL(sourceUrl);
+    // label = `${filename}:${line}`;
     icon = "document-text";
-    label = `${filename}:${sourceLocation.line}`;
-    secondaryLabel = getTextAtLocation(state, sourceLocation.sourceId, sourceLocation);
+    label = closestFunction?.name || "(anonymous)";
```
jaril, line 40: `findClosestFunction` function either returns 1) an object that represents the function, or 2) null.

This just makes sure that in null cases, we just refer to it as `(anonymous)`.

src/ui/components/Transcript/TranscriptItem/FloatingTranscriptItem.tsx
```diff
@@ -1,43 +1,44 @@
...
-import { ThreadFront } from "protocol/thread";
+import { FloatingItem } from "ui/state/comments";
 import { UIState } from "ui/state";
+import { selectors } from "ui/reducers";
 
 const { getFilenameFromURL } = require("devtools/client/debugger/src/utils/sources-tree/getURL");
 const { getTextAtLocation } = require("devtools/client/debugger/src/reducers/sources");
 const { getFormattedTime } = require("ui/utils/timeline");
+const { findClosestFunction } = require("devtools/client/debugger/src/utils/ast");
+const { getSymbols } = require("devtools/client/debugger/src/reducers/ast");
```
jasonLaster, line 14: in my initial patch, I made `findClosestFunction` a selector so that we could query it directly. Not a big deal tho

src/ui/components/Transcript/TranscriptItem/FloatingTranscriptItem.tsx
jasonLaster: How about we use the function name if we have it otherwise fall back to the source filename

jasonLaster (approved):

PR #2081, opened 2021-03-10
Tweaks to header to align closer to our Figma doc

jonbell-lot23:
image

Yee-haw, I'm getting back into the code a bit. This was a small tweak to make things closer to the Figma designs I cleaned up yesterday.


jonbell-lot23:
Hm, I may have screwed up the scrollbar in the transcript. It should be flush against the right but my padding might have messed it up.

Issue #2084, opened 2021-03-11
Infinite loop in Transcript.tsx


hbenl:
Currently some recordings can't be opened because the devtools run into an infinite loop:
[This effect](https://github.com/RecordReplay/devtools/blob/a2755285eabeddd11b4e9b5b7ce881ec559b5396/src/ui/components/Transcript/Transcript.tsx#L55-L67) dispatches a redux thunk that dispatches a redux action that changes the `floatingItem` prop and hence triggers a rerender, which triggers the effect, ...
One might think that this is mitigated by the `[currentTime, comments]` dependency list on the effect. However, the `comments` come from the `useGetComments()` hook, which [creates a new empty array](https://github.com/RecordReplay/devtools/blob/a2755285eabeddd11b4e9b5b7ce881ec559b5396/src/ui/hooks/comments.ts#L85) every time it is called until the comments are received from hasura. On some recordings, this will break the loop when comments are received, whereas on others devtools get stuck in the loop because the comments seemingly are never received (I haven't investigated this difference any further).

PR #2085, opened 2021-03-11
Don't dispatch set_floating_item when the floating item isn't changing (#2084)

hbenl:


hbenl:
I'm merging this because Replay is currently broken and we're nearing the Deploy hold period.

PR #2092, opened 2021-03-11
Turning blue icons to light grey, and changing background from black to ... also light grey

jonbell-lot23:
Before:
before - Screen Shot 2021-03-12 at 11 06 36 AM

After:
after - Screen Shot 2021-03-12 at 11 20 36 AM

Aaahhh, doesn't that feel less intense?


src/ui/components/Transcript/TranscriptItem/TranscriptItem.css
```diff
@@ -74,7 +74,7 @@
 
 .event-timestamp {
   font-size: 12px;
-  background: var(--grey-20);
+  background: var(--gray-20);
```
jaril, line 77: This one's really meant to be var(--grey-20). There's some gray/grey conflict with the naming schemes :)

jonbell-lot23: Oh! Thanks!

jasonLaster (commented):
Love this!

PR #2123, opened 2021-03-16
Sync package-lock with package.json.

loganfsmyth:
Looks like https://github.com/RecordReplay/devtools/commit/17d9e99acf57d28d2a1ea9b24372fecec1a1e42d landed some new deps without updating the lockfile.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1363,6 +1363,48 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


hbenl:
This is also part of #2120, I'd prefer to just land that.

PR #2125, opened 2021-03-17
Handle evaluation failures correctly (#2100)

### LINKED ISSUE #2100: TypeError: can't access property "isString", t is undefined
Sentry Issue: [DEVTOOLS-J7](https://sentry.io/organizations/replay/issues/2270301337/?referrer=github_integration)

```
TypeError: can't access property "isString", t is undefined
  at GripMessageBody (./src/devtools/client/webconsole/components/Output/GripMessageBody.js:54:7)
  at a (./src/devtools/client/webconsole/components/Output/CollapseButton.js:25:42)
  at render (./src/devtools/client/webconsole/components/Output/ConsoleOutput.js:120:9)
  at next (./src/devtools/client/debugger/src/actions/utils/middleware/promise.js:34:14)
  at next (./src/devtools/client/debugger/src/actions/utils/middleware/context.js:66:12)
...
(41 additional frame(s) were not displayed)
```

hbenl:
Fixes #2100

### DIFF
--- a/src/devtools/client/webconsole/actions/input.js
+++ b/src/devtools/client/webconsole/actions/input.js
@@ -47,18 +47,29 @@ function evaluateExpression(expression) {
     );
     dispatch({ type: EVALUATE_EXPRESSION, expression });
 
-    // Even if the evaluation fails,
-    // we still need to pass the error response to onExpressionEvaluated.
-    const onSettled = res => res;
+    try {
+      const response = await evaluateJSAsync(expression, {
+        asyncIndex,
+        frameId,
+        forConsoleMessage: true,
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
Does it make sense to add an end-to-end test for this?

hbenl:
> Does it make sense to add an end-to-end test for this?

This is not about failures in the evaluated expression (that case was already handled correctly), but about failures somewhere in the backend that lead to the evaluation request failing (instead of returning information about the error thrown in the evaluated code). I don't think it is possible for a devtools user to trigger this reliably.

PR #2129, opened 2021-03-17
Render promises as regular objects (#2097)

### LINKED ISSUE #2097: Promises are shown as invalid objects
**Description:**
 
I've seen promises come up as invalid objects pretty consistently. This was especially apparent when i was pairing with a new user and we were looking into a promises bug.

**Steps to Reproduce:**
1. view https://replay.io/view?id=8d9fa1a8-dee1-4024-ac4c-b37d7f6c55f6

The original app is here https://obwxr.csb.app/ and https://codesandbox.io/s/divine-https-obwxr?file=/src/index.js:103-123

**Replay:** https://replay.io/view?id=eac1476a-2d41-45c5-a906-ea3ce32c7f30

hbenl:
`PromiseRep` expects special handling of promises by the backend, surfacing a promise's state and value. No such special handling currently exists in our backend (AFAIK), so with this PR promises are treated like regular objects.
Fixes #2097

### DIFF
--- a/src/devtools/client/debugger/packages/devtools-reps/src/reps/rep.js
+++ b/src/devtools/client/debugger/packages/devtools-reps/src/reps/rep.js
@@ -54,7 +54,6 @@ const reps = [
   TextNode,
   Attribute,
   Func,
-  PromiseRep,
   ArrayRep,
   Document,
   DocumentType,


jasonLaster:
Lets get @loganfsmyth and @bhackett1024 thoughts. My take on it is that we should prioritize getting the promise state, like we do in firefox (https://searchfox.org/mozilla-central/source/devtools/server/actors/object.js#247-264).


Here's a replay with a promise packet (https://replay.io/view?id=eac1476a-2d41-45c5-a906-ea3ce32c7f30)
Screen Shot 2021-03-18 at 6 33 53 AM

This version _does_ show that there is a promise, but it is sad to show it empty

Screen Shot 2021-03-18 at 6 36 26 AM

loganfsmyth:
I'm happy to look into adding promise state to the recording next week. Since older recordings won't have that data, it seems like we'll need the frontend to be backward-compatible. Can we change `PromiseRep` to gracefully fall back to standard Object behavior if these fields are missing?

hbenl:
> I'm happy to look into adding promise state to the recording next week

Backend issue: RecordReplay/backend#1568

> Can we change PromiseRep to gracefully fall back to standard Object behavior if these fields are missing?

Of course.

loganfsmyth:
Fixed by #2175

PR #2132, opened 2021-03-17
Swap the stepIn button to be before stepOut

loganfsmyth:
I once again hit the wrong button because our buttons are in a different order than Chrome and FF.


jasonLaster (approved):
lol

Issue #1753, opened 2021-02-10
Make Codemirror lazy


jasonLaster:
**Description:**

Codemirror is our second biggest library and we only need it when devtools are open, so it would be nice to make it lazy.

<img width="1025" alt="Screen Shot 2021-02-10 at 2 33 05 PM" src="https://user-images.githubusercontent.com/254562/107581734-5684f380-6bad-11eb-8f7b-4b52e6fc1d6e.png">

PR #2140, opened 2021-03-18
Import CodeMirror dynamically (#1753)

hbenl:


src/devtools/client/debugger/src/components/App.js
```diff
@@ -284,6 +285,20 @@ Debugger.childContextTypes = {
   l10n: PropTypes.object,
 };
 
+function DebuggerWaitingForEditor(props) {
+  const [hasEditor, setHasEditor] = useState(false);
+
+  useEffect(() => {
+    async function awaitEditor() {
```
jasonLaster, line 292: style thing, but an iife would be cleaner. and perhaps `setLoading` instead of editor, we might want to lode other things as well.

```js
useEffect(() => {
  (async () => {
    await waitForEditor();
    setLoading(false)
  })()
}, [])
```

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelInput.js
```diff
@@ -37,6 +38,10 @@ export function PanelInput({
         extraKeys.Esc = onEscape;
       }
 
+      const CodeMirror = await import("codemirror");
```
jasonLaster, line 41: This feels a bit gross, i wonder if there's a way to simplify this further...

perhaps use `setState` for CodeMirror or pass it down from the editor...

src/devtools/client/debugger/src/components/App.js
hbenl: Actually I think an IIFE is less readable. But I already noticed this is a minority opinion among javascript developers.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelInput.js
hbenl: I created a `getCodeMirror()` function with which I can get CodeMirror synchronously once it has been imported (and this component won't get rendered before that). Passing it down would have required passing it through some half dozen components, so I decided against that.

src/devtools/client/debugger/src/components/App.js
jasonLaster: hehe

jasonLaster (approved):

Issue #2159, opened 2021-03-21
The console should only scroll when the message is not visible


jasonLaster:
https://replay.io/view?id=75a881e5-d810-4f6c-b1ae-b7f664104f1e

PR #2163, opened 2021-03-22
Fix the visibility check in ConsoleOutput.scrollToClosestMessage() (#2159)

hbenl:


jasonLaster (approved):
I like how much simpler this is

PR #2167, opened 2021-03-23
Ensure that stepping through breakpoint hits works in order.

loganfsmyth:
See https://github.com/RecordReplay/devtools/issues/2157#issuecomment-803497028

We've got this case where hitting next/previous on a breakpoint will sometimes jump more than one point. This looks to be because the frontend doesn't do any sorting of the results.

We could do _some_ of this sorting in the server, but since `Analysis.analysisPoints` is an event, and there could be multiple events (though I think right now we only send one with the full result), even if we sorted the individual response points, points in the various response packets could still be interleaved with points in other events, so it seems like this should be up to the client.


hbenl (approved):
Almost identical to #2164 :laughing:

Issue #2062, opened 2021-03-10
Remove recording_id jwt depenednecy


jasonLaster:
**Description:**
The JWT dependency is causing a couple issues to creep in

1. popups are appearing on initial load
2. brave is upset
3. perhaps others...

It sounds like we can explore another option, which might behave better

PR #2172, opened 2021-03-23
Remove the recordingId from the access token (#2062)

hbenl:


src/ui/components/Header/UserOptions.js
```diff
@@ -14,7 +14,7 @@ function UserOptions({ recordingId, setModal }) {
   const [expanded, setExpanded] = useState(false);
   const { isAuthenticated } = useAuth0();
 
-  const showShare = hooks.useIsOwner(recordingId);
+  const showShare = recordingId ? hooks.useIsOwner(recordingId) : false;
```
jasonLaster, line 17: yikes... i dont think we should put a condition around a hook... is there another way of doing this like?

```js
const showShare = hooks.useIsOwner(recordingId ? recordingId : false);
```

hbenl: Whoops, of course we shouldn't. It feels wrong to send a dummy query to Hasura because of this, but I can't think of a way to avoid this...

Issue #1903, opened 2021-02-24
After signing in, user should return to page they were previously on


lisahackett1978:
**Description:**
When logged out of the Replay browser and attempting to add a comment, Replay prompts the user to log in. After logging in user always arrives at the recordings page, not the screen they were on previously.

**Steps to Reproduce:**
1. Log out of Replay
2. Quit Replay
3. Open Replay, copy URL of a recording into the address bar, go
4. Attempt to add a comment
5. Log in
6. Arrive at recordings page

**Expected Result:**
After signing in, return user to page they were previously on
**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:

jonbell-lot23:
100%

jasonLaster:
Thanks Lisa for the writeup. I had filed another issue #1712, which was a duplicate but your version is clearer.

My hope is that we can fix this when we change Authentication providers. We're currently exploring clerk.dev and AWS cognito.

hbenl:
We can fix this without changing our auth provider, see https://github.com/RecordReplay/devtools/issues/1761#issuecomment-792765142

hbenl:
depends on #2062

PR #2174, opened 2021-03-23
Return to the previous page after login (#1903)

hbenl:


hbenl:
This should only be merged _after_ #2172

jasonLaster (approved):

PR #2175, opened 2021-03-23
Expose promise state in object preview.

### LINKED ISSUE #2097: Promises are shown as invalid objects

loganfsmyth:
I'm not totally clear on if this is the right way to do it, but I did this while testing out the backend changes so figured I'd post it. If there's a better way, feel free to take over.

Depends on https://github.com/RecordReplay/backend/issues/1568 and https://github.com/RecordReplay/gecko-dev/pull/324

Fixes #2097

### DIFF
--- a/src/protocol/thread/value.ts
+++ b/src/protocol/thread/value.ts
@@ -127,6 +127,11 @@ export class ValueFront {
     return [];
   }
 
+  previewPromiseState() {
+    // Older recordings did not set promiseState so this could be null.
+    return (this.hasPreview() && this._object!.preview!.promiseState) || null;
+  }
+
   className() {
     if (this._object) {
       return this._object.className;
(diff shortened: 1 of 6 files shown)


bhackett1024:
Can we update the object_preview-01.js test to exercise this?

src/protocol/thread/value.ts
```diff
@@ -127,6 +127,14 @@ export class ValueFront {
     return [];
   }
 
+  previewPromiseState() {
+    if (this.hasPreview()) {
+      // Older recordings did not set promiseState so this could be null.
+      return this._object!.preview!.promiseState || null;
+    }
+    return null;
```
hbenl, line 135: Could be simplified to `return this._object?.preview?.promiseState || null;`

hbenl (approved):
LGTM

src/protocol/thread/value.ts
loganfsmyth: Yeah I can change it. It's this way because I copy-pasted `previewContainerEntries` and never bothered to reformat it.

PR #2181, opened 2021-03-24
Expand sources by default

### LINKED ISSUE #2004: Source Explorer should be expanded by default
**Description:**

I recently changed the UI to have outline view expanded by default, now it seems like the source explorer is collapsed which is weird

jaril:
Fix #2004.


jasonLaster (approved):
so embarrassing. I flipped this in my head

Issue #2006, opened 2021-03-05
Add comment @mentions


jasonLaster:
**Description:**

@mentions will let folks easily collaborate.

https://codesandbox.io/s/8n3zl4n9pl?file=/src/MyEditor.js:193-216
https://www.draft-js-plugins.com/plugin/mention

ryanjduffy:
Some questions that come to mind around this:
* Should it be limited to team members or active collaborators?
* If not, maybe order by collaborators, team members, then all users?
* How are users notified of mentions? Email? Notification page a la GitHub?

First thoughts around the design:

* The plugin seems like it'd do the trick. The architecture of it is a bit weird but workable.
* When the user adds/edits a comment, fire off a request for the mention suggestions. Ideally, this would be cached and refreshed on future edits.
* When the commit is saved, the comment would be parsed for mentions and those would be inserted into a notifications table (linking comment, commenter, and mention and possibly recording for ease of access to that data).
* The notifications table would be the source for a standalone view and could be periodically queried to generate email.

jasonLaster:
Great questions. I think we can keep the MVP fairly limited:

1. Only show recording collaborators, for most replays that actually means you wont use this feature unless you're viewing someone else's recording :)

2. Fetch the collaborators when the recording loads. We do this anyways for the share modal so not a big deal.

PR #2198, opened 2021-03-26
Pre-Load DraftJS from CommentThread

### LINKED ISSUE #2191: Load draftjs earlier so that the comment editor shows up faster

ryanjduffy:
Fixes #2191 and re-fixes: #2006

* Adds back mention support
* Refactors the draftjs code to allow pre-loading of modules
* Preloads modules from CommentThread on 1s timeout

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@babel/types": "^7.12.0",
     "@draft-js-plugins/editor": "^4.1.0",
     "@draft-js-plugins/emoji": "^4.2.0",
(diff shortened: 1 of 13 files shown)


ryanjduffy:
Comment submit doesn't work yet due to the current structure so a bit more refactoring is required.

ryanjduffy:
I think this is ready for review. I've confirmed that draft-js and the plugins are not properly split into their own chunks by webpack and that they are loaded async from the CommentThread.

jasonLaster:
landed via #2223

PR #2215, opened 2021-03-29
Use dispatcher commands to access user settings

bhackett1024:
Experimental, don't merge.


bhackett1024:
This works, except that while we invalidate old queries after changing user settings, we don't refetch those queries or update affected react components.  I looked into that a bit but I'm not very familiar with react.  It looks like Apollo Client uses useReducer() for dealing with this.

PR #2226, opened 2021-03-31
Add invite system

### LINKED ISSUE #2229: Add basic invite system

jaril:
Fix #2229.

This adds basic invitation support to Replay.
- New users can register for a Replay account provided that they have access to the web app link, which prompts them to register for one.
- New users start off as non-invited. Non-invited users can create replays, but the replays they create are not viewable until the user is invited.
- To invite a user, an invited user can go into the Settings panel and access the Invitations tab where they can invite users using their email addresses. Note that this action doesn't currently send that user an e-mail.
- Non-invited users see an error page when they try to open their settings panel's invitations tab.
- There is currently no verification for the email input. So its easy to invite an invalid email, e.g. yourself, an existing Replay user, an invalid email address, etc. This should be done as a follow up.
- Regular users have 5 invitations. Replay employees have infinite invitations.
- We cross-check non-invited users with outstanding invitations from the client. Once a non-invited user has an invitation that matches their e-mail, we mark their account as invited.

image

### DIFF
--- a/src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
+++ b/src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
@@ -6,7 +6,7 @@ import "./ReplayInvitations.css";
 export default function ReplayInvitations() {
   const [inputValue, setInputValue] = useState("");
   const { invitations, loading: inviteLoading } = hooks.useGetInvitations();
-  const { isInternal, loading: userLoading } = hooks.useGetUserInfo();
+  const { isInternal, invited, loading: userLoading } = hooks.useGetUserInfo();
   const addInvitation = hooks.useAddInvitation();
   const userId = getUserId();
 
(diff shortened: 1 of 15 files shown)


src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -0,0 +1,66 @@
...
+    return <li className="replay-invitations" />;
+  }
+
+  const inviteCount = invitations.length;
+
+  return (
+    <li className="replay-invitations">
+      <label className="setting-item">
+        <div className="label">
+          You have {5 - inviteCount} invite{5 - inviteCount == 1 ? "" : "s"} left
```
jasonLaster, line 36: lets move 5 to a variable

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+        <div className="description">
+          Replay is meant for sharing. Invite your team and friends below so they can start creating
+          their own replays.
+        </div>
+      </label>
+      {inviteCount < 5 && (
```
jasonLaster, line 43: lets add a follow up so that internal users can invite infinitely many folks.

jasonLaster (approved):

PR #2234, opened 2021-04-02
Add new welcome screen for users who haven't created a replay yet

### LINKED ISSUE #2194: RPT-1 Welcome screen
- [x] Default people to this welcome screen
- [x] Place to put text

jaril:
Fix #2194.

image
image


jasonLaster:
Is it possible to detect that we're in the Replay browser and can record?

If so, it would be nice to only show it then and not in another browser where you can only view your library.

Screen Shot 2021-04-04 at 8 04 42 PM

jaril:
Let's detect the Replay browser in a follow up by adding `__IS_RECORD_REPLAY_RUNTIME__`.

jonbell-lot23:
@jaril and @jasonLaster it looks like this got stopped from going live? It's a big part of RPT, can we resolve conflicts and push it live?

PR #2275, opened 2021-04-06
Visual adjustments to the header in preparation for teams feature

jonbell-lot23:
We'll soon have our Teams feature, so I've been designing ahead to prepare for that. These are some small tweaks to prepare for that future.

1. Removed outdated blue color
2. Moved logo (it'll conflict with our IA when we add teams, but we can discuss)
3. Added pointer and color highlight to menu highlight
4. Made the menu more subtle. It's big and bold and stands out from our wispy other icons.

### DIFF
--- a/src/ui/components/Account/Library.tsx
+++ b/src/ui/components/Account/Library.tsx
@@ -9,7 +9,7 @@ function Header() {
   return (
     <div id="header">
       <div className="header-left">
-        <div className="title-label">Your Replay library</div>
+        <div className="title-label">Your library</div>
       </div>
       <UserOptions mode="account" />
     </div>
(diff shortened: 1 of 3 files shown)


src/ui/components/Account/Library.tsx
```diff
@@ -9,8 +9,7 @@ function Header() {
   return (
     <div id="header">
       <div className="header-left">
-        <div className="logo" />
-        <div className="title-label">Replay</div>
+        <div className="title-label">Your Replay library</div>
```
jasonLaster, line 12: do we need Replay? `Your library` might be punchier...

jonbell-lot23: I could see that. Will remove, thanks!

jonbell-lot23: Ok, change is in

PR #2290, opened 2021-04-07
Add error and retry link to Apollo

### LINKED ISSUE #2225: TypeError: can't access property "users", e is undefined
Sentry Issue: [DEVTOOLS-MA](https://sentry.io/organizations/replay/issues/2306757799/?referrer=github_integration)

```
TypeError: can't access property "users", e is undefined
  at getRecordings (./src/ui/hooks/recordings.ts:169:3)
  at useGetMyRecordings (./src/ui/hooks/recordings.ts:311:22)
  at getRecordings (./src/ui/components/Dashboard/Dashboard.tsx:20:43)
  at c (./src/ui/components/Dashboard/Dashboard.tsx:27:35)
...
(12 additional frame(s) were not displayed)
```

jaril:
Fix #2225.

### DIFF
--- a/src/ui/utils/apolloClient.ts
+++ b/src/ui/utils/apolloClient.ts
@@ -1,7 +1,8 @@
-import { ApolloClient, InMemoryCache, NormalizedCacheObject } from "@apollo/client";
-import { HttpLink } from "apollo-link-http";
+import { ApolloClient, InMemoryCache, NormalizedCacheObject, from, HttpLink } from "@apollo/client";
 import { DocumentNode } from "graphql";
 import { defer } from "protocol/utils";
+import { onError } from "@apollo/client/link/error";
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
How does this fix it?

jaril:
Technically, it doesn't fully fix it. This specific error is caused by a GraphQL network error. 

This just adds a retry link so that in the case of a network error, we only throw after having tried it again a few times. Specifically, it re-attempts the query 5 times before giving up and throwing the error.

https://www.apollographql.com/docs/react/api/link/apollo-link-retry/

The extra mile would be to fail gracefully if an error throws anyway after retrying 5x. But I'm putting this in for now to see if the occurrence of these errors becomes trivial with just this fix.

hbenl (approved):

PR #2292, opened 2021-04-07
Add back privacy link

ryanjduffy:
When I was messing around with tailwind, I noticed the privacy link was dropped so this adds it back.

Was:
image

Now:
image


jaril (approved):
Good catch. Thanks Ryan!

PR #2312, opened 2021-04-09
Revising the library in preparation for teams and filters

jonbell-lot23:
Previous design:
image

Revision:
image

This is a pretty big visual change! Here's what's going on.

* We need to prepare for Teams and a fantastic filter/search experience
* The old design was cutting off Replay description
* The new design works to have better information density ... while removing low priority information
* Under consideration: what if Teams were listed along the left-hand side? It'd be a fantastically clear mental model and also make Teams more prominent. To discuss.

This is not the final design, but living with it for a few days is going to help us make the next step forward on the UI.

### DIFF
--- a/src/ui/components/Dashboard/DashboardViewerContent.js
+++ b/src/ui/components/Dashboard/DashboardViewerContent.js
@@ -108,7 +108,7 @@ function DashboardViewerContentHeader({
       </th>
       <th>PREVIEW</th>
       <th>TITLE</th>
-      <th>DURATION</th>
+      <th>LENGTH</th>
       <th className="sorter" onClick={() => setAscOrder(!ascOrder)}>
         <span className="label">CREATED</span>
         {ascOrder ? <div className="img arrow-up-2" /> : <div className="img arrow-down-2" />}
--- a/src/ui/components/Dashboard/DashboardViewerContent.js
+++ b/src/ui/components/Dashboard/DashboardViewerContent.js
@@ -108,13 +108,13 @@ function DashboardViewerContentHeader({
       </th>
       <th>PREVIEW</th>
       <th>TITLE</th>
-      <th>LENGTH</th>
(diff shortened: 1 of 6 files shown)


jasonLaster:
Wow! I like a lot of this. 

I think dropping the URLs might be a bit early for me without an appropriate replacement. 

Also the teams dropdown might be better in the top right header? thoughts?

jonbell-lot23:
1. The average user has one domain. I don’t think that justifies a filter. I’d rather that space be empty then add visual complexity.
2. My favourite placement idea is putting it in the top left, where it currently says “Your Library.” But I can see it going in the top right too. Worthy discussion for the next iteration.

jasonLaster:
👍 I think you're right about both of these. 

It's just tough because I'm not the average user 😉

jasonLaster:
- 👍 I think you're right about both of these. 
- It's just tough because I'm not the average user!
- I would move the copy link button to the dropdown. It's not clear to me what we're copying actually. I had to stop and think
- The padding for the row feels a bit tight.
- i'm now excited to drop the left sidebar :)

jonbell-lot23:
Roger! I agree we can loosen the padding. I was using “auto” in a new way so I’ll figure that out.

I totally know how I want copy link to work, but don’t know how to write the code. Maybe we can make a separate ticket for that.

jonbell-lot23:
Oh and the left sidebar. Yeah, once we put teams somewhere (maybe top left?) then it’ll be fully empty. I’d be happy to explore dropping it outright. Talk about an easy to understand page!

jasonLaster:
totally. Drop it all

jonbell-lot23:
Alright, I changed the padding. Which leaves these other items I can steer right into a new ticket:

* Put the dropdown somewhere (I like the top left, we'll see)
* Figure out how to deal with copy link
* Explore dropping the left side nav entirely

PR #2330, opened 2021-04-12
Fix loading order so that workspaces and invitations appear immediately

### LINKED ISSUE #2265: Teams (aka workspaces) don't appear in library until about 5-10 seconds go by
Replay: https://share.descript.com/view/Fy7ABUeqKNt

We need to make sure that UI appears immediately, and can never have a delay like this.

jaril:
Fix #2265.

Before: https://replay.io/view?id=ba9c60ab-cdce-4716-9a91-dcdb23875555&point=33749929583581750599676199591478407&time=15123.4404551201&hasFrames=true
After: https://share.descript.com/view/PLepn1E8MSM


hbenl (approved):
LGTM

PR #2332, opened 2021-04-12
Bumped react-devtools-inline to 4.12

bvaughn:
`createBridge` and `createStore` have been added to stable release 4.12.0

Note that the version of the `react-devtools-core` embedded back that's used should be bumped to 4.12.0 as well to make sure that message formats are in-sync. Longer term, we should defer running the backend code until the profile is being viewed. Might not want to land this change without also landing that one.

cc @jasonLaster

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4490,6 +4490,7 @@
         "icss-utils": "^4.1.1",
         "loader-utils": "^1.2.3",
         "normalize-path": "^3.0.0",
+        "postcss": "^7.0.32",
         "postcss-modules-extract-imports": "^2.0.0",
         "postcss-modules-local-by-default": "^3.0.2",
         "postcss-modules-scope": "^2.2.0",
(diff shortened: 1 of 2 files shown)


jasonLaster:
Thanks @bvaughn linking to the react-devtools gecko-dev [directory](https://github.com/RecordReplay/gecko-dev/tree/webreplay-release/devtools/server/actors/replay/react-devtools). 

Are you saying, we'll also need to re-apply the the backend files as well: hook.js, contentScript.js, react_devtools_backend.js?

bvaughn:
Looks like just the `react_devtools_backend.js` I think

jasonLaster:
kicking off a clone --> build for gecko-dev

jasonLaster:
also installing Java :P

bvaughn:
Oh wait what do you need? Just the built backend file? or what?

PR #2333, opened 2021-04-12
Adding a download link to the dropdown

jonbell-lot23:
Addresses https://github.com/RecordReplay/devtools/issues/2213 (sort of ... we're doing this as a short term fix, and the broader fix will be later)

Follow-ups:

1. The link should say "Launch Replay" and launch the app if it's installed
2. We might show things differently in library versus devtools
3. If you're using the Replay browser already, the link should go away


jasonLaster (approved):

jasonLaster:
Awesome

PR #2359, opened 2021-04-15
Update the Devtools Node layout

jasonLaster:
Here's a replay with some thoughts. Theres at least one thing i will tidy up tomorrow

https://replay.io/view?id=0cdf1b09-d245-4be7-9c2e-e4ef45097bb9&point=14278816362184167063060290937553346&time=7196.719843290892&hasFrames=true

Screen Shot 2021-04-14 at 5 22 00 PM

### DIFF
--- a/src/ui/components/Header/Header.js
+++ b/src/ui/components/Header/Header.js
@@ -43,11 +43,11 @@ function Avatars({ recordingId, sessionId }) {
   );
 }
 
-function Links({ recordingId, sessionId }) {
+function Links({ recordingId, sessionId, recordingTarget }) {
   return (
     <div className="links">
       <Avatars recordingId={recordingId} sessionId={sessionId} />
-      {!prefs.video && <ViewToggle />}
+      {!prefs.video && recordingTarget != "node" && <ViewToggle />}
       <UserOptions />
     </div>
   );
@@ -92,7 +92,7 @@ function HeaderTitle({ recordingId, editingTitle, setEditingTitle }) {
   );
 }
 
-function Header({ recordingId, sessionId }) {
+function Header({ recordingId, sessionId, recordingTarget }) {
   const [editingTitle, setEditingTitle] = useState(false);
 
   return (
@@ -104,7 +104,7 @@ function Header({ recordingId, sessionId }) {
           editingTitle={editingTitle}
         />
       </div>
-      <Links recordingId={recordingId} sessionId={sessionId} />
+      <Links recordingId={recordingId} sessionId={sessionId} recordingTarget={recordingTarget} />
     </div>
   );
 }
@@ -112,4 +112,5 @@ function Header({ recordingId, sessionId }) {
 export default connect(state => ({
   recordingId: selectors.getRecordingId(state),
   sessionId: selectors.getSessionId(state),
+  recordingTarget: selectors.getRecordingTarget(state),
 }))(Header);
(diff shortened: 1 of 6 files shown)


src/ui/components/Header/Header.js
```diff
@@ -78,6 +78,7 @@ function HeaderTitle({ recordingId, editingTitle, setEditingTitle }) {
   }
 
   const { recordingTitle, title } = data.recordings?.[0] || {};
+  console.log(data, recordingTitle, title);
```
sergeichestakov, line 81: did you mean to leave this console.log in here?
```suggestion
```

src/ui/components/SecondaryToolbox/index.js
```diff
@@ -88,17 +88,19 @@ function InspectorPanel() {
   );
 }
 
-function SecondaryToolbox({ selectedPanel, setSelectedPanel, narrowMode }) {
+function SecondaryToolbox({ selectedPanel, setSelectedPanel, narrowMode, recordingTarget }) {
   const {
     userSettings: { show_react },
   } = hooks.useGetUserSettings();
+  const isNode = recordingTarget == "node";
```
sergeichestakov, line 95: should this be using triple equals?
```suggestion
  const isNode = recordingTarget === "node";
```

src/ui/components/SecondaryToolbox/index.js
```diff
@@ -88,17 +88,19 @@ function InspectorPanel() {
...
   return (
-    <div className="secondary-toolbox">
+    <div className={`secondary-toolbox ${isNode && "node"}`}>
```
sergeichestakov, line 97: I typically use`classnames` when I need to conditionally applying class names and thoroughly enjoy it: https://www.npmjs.com/package/classnames

Just a suggestion!

sergeichestakov (approved):
Left a couple comments/suggestions but this looks great! 😄 

Only other thing I'd say here is that there seems to be a lot of special casing for Node. I saw you mentioned this in your replay as well. Don't have enough context on which languages you support and how their use cases differ but is there anything here that can be generalized? That way you can conditionally render based on a language "type" or specific use case rather than on the language itself because I imagine it can get pretty messy if each language has to render something different.

At Replit, for example, we have a config for each language with a bunch of fields, (e.g. `hasGit`, `hasInterpreter`, etc) each of which have implications for what gets shown in the IDE:

Screen Shot 2021-04-15 at 10 44 59 AM

You can see the full list here: http://eval.global.replit.com/languages

Curious to hear your thoughts! Not a blocker by any means for this PR, you can always follow up anyways

jaril (approved):
Apart from the items that @sergeichestakov brought up, looks good to me!

src/ui/components/Header/Header.js
jasonLaster: haha - sloppy :)

src/ui/components/SecondaryToolbox/index.js
jasonLaster: yep - we use classnames too, just moving quickly

src/ui/components/SecondaryToolbox/index.js
jasonLaster: hehe

jasonLaster:
Thanks - I like the idea of having target attributes like `hasGraphics` as opposed to specific language checks. Another example of this will be libraries + frameworks such as React, Redux, React Profiling

PR #2365, opened 2021-04-15
Fix opening share modal from the library

### LINKED ISSUE #2348: Cannot view sharing preferences from the library
**Description:**

Looks like the sharing modal in the library view is causing a crash

Here is a replay 
https://replay.io/view?id=7f2cc017-cfe7-41c1-a910-af93eb471a93

jaril:
Fixes #2348.

This restores the old way we would pass a recordingId to the share modal in a way that works for both the library and a replay. Instead of referring to `state.app.recordingId`, the modal refers to a special `recordingId` in `state.app.modalOptions`.

Demo: https://share.descript.com/view/5KgEcgwgb8u


jasonLaster:
transferring to @hbenl

hbenl (approved):

PR #2381, opened 2021-04-16
mixpanel: track breakpoints

jazzdan:
Use mixpanel to track who sets breakpoints, in prod only.

Also add userEmail to mixpanel and sentry context.

We're just using mixpanel as an experiment right now, it might be
removed in the future. We want to see how useful it is to capture
user behavior that we can't just infer from rows written to the DB.


jazzdan:
Hmm not sure if that test failure is real, any thoughts?

jaril (approved):

hbenl (approved):

hbenl:
@jazzdan The test failure was introduced by another PR

PR #2390, opened 2021-04-19
mixpanel: add workspace name to context when viewing a recording

jazzdan:


src/ui/hooks/recordings.ts
```diff
@@ -30,6 +30,9 @@ export function useGetRecording(recordingId: RecordingId) {
           user {
             invited
           }
+          workspace {
```
jazzdan, line 33: Can I just modify graphQL queries now given the changes to the API?

jasonLaster: This is fine. The nice thing about graphql is that once we have workspaces supported it should just work

src/ui/hooks/recordings.ts
```diff
@@ -45,7 +48,8 @@ export function useGetRecording(recordingId: RecordingId) {
   const recording = data?.recordings[0];
   // Tests don't have an associated user so we just let it bypass the check here.
   const isAuthorized = isTest() || recording;
-  return { recording, isAuthorized, loading };
+  const workspaceName: string | null = data?.workspace?.name;
```
loganfsmyth, line 51: This is in the wrong place given the query you are doing, it would be `recording?.workspace?.name`

src/ui/hooks/recordings.ts
loganfsmyth: Yup!

src/ui/components/DevTools.tsx
```diff
@@ -50,9 +51,13 @@ function DevTools({
...
-  const { recording, isAuthorized, loading: recordingQueryLoading } = hooks.useGetRecording(
-    recordingId
-  );
+  const {
+    recording,
+    isAuthorized,
+    loading: recordingQueryLoading,
+    workspaceName,
+  } = hooks.useGetRecording(recordingId);
+  mixpanel.register({ workspaceName });
```
loganfsmyth, line 60: Is this going to complain if the value is `undefined`, like on the recording list page?

src/ui/hooks/recordings.ts
jazzdan: argh thanks!

src/ui/components/DevTools.tsx
loganfsmyth: You're right, not sure what I was thinking there. I guess the question still stands for `undefined` when there is no workspace though.

src/ui/hooks/recordings.ts
```diff
@@ -45,7 +48,8 @@ export function useGetRecording(recordingId: RecordingId) {
   const recording = data?.recordings[0];
   // Tests don't have an associated user so we just let it bypass the check here.
   const isAuthorized = isTest() || recording;
-  return { recording, isAuthorized, loading };
+  const workspaceName: string | null = recording?.workspace?.name;
```
loganfsmyth, line 51: To me it feels like this belongs more at the call-site of this function e.g.
```
if (recording?.workspace) {
  mixpanel.register({ 
    workspaceName: recording.workspace.name
  });
}
```

src/ui/components/DevTools.tsx
```diff
@@ -53,6 +54,9 @@ function DevTools({
   const { recording, isAuthorized, loading: recordingQueryLoading } = hooks.useGetRecording(
     recordingId
   );
+  if (recording?.workspace) {
+    mixpanel.register({ workspaceName: recording.workspace.name });
```
jazzdan, line 58: This is probably not what we want because it will associate this workspace with this user until the next time they view a recording, which is not accurate.

jazzdan:
I'm going to close in favor of the approach that Jaril and I outlined in #2418

PR #2412, opened 2021-04-22
Only start a logrocket session for external users

### LINKED ISSUE #2367: Logrocket: exclude internal sessions
**Description:**

It's not that helpful for logrocket to have a bunch of internal sessions. We'll also need the client to fetch the internal field for users

```diff
diff --git a/src/ui/utils/environment.ts b/src/ui/utils/environment.ts
index dbd93a52..ca91c0b0 100644
--- a/src/ui/utils/environment.ts
+++ b/src/ui/utils/environment.ts
@@ -1,3 +1,5 @@
+import { User } from "ui/types";
+
 const url = new URL(window.location.href);

 export function isDevelopment() {
@@ -16,8 +18,8 @@ export function isTest() {
   return getTest() != null;
 }

-export function skipTelemetry() {
-  return isTest() || isDevelopment();
+export function skipTelemetry(user: User) {
+  return user.internal || isTest() || isDevelopment();
 }

 export function isDeployPreview() {
diff --git a/src/ui/utils/logrocket.ts b/src/ui/utils/logrocket.ts
index 8c53ee96..a452bb45 100644
--- a/src/ui/utils/logrocket.ts
+++ b/src/ui/utils/logrocket.ts
@@ -7,7 +7,7 @@ let setup = false;

 export default {
   createSession: (auth: any) => {
-    if (skipTelemetry()) {
+    if (skipTelemetry(auth.user)) {
       return;
     }
```

jazzdan:
Fixes #2367

### DIFF
--- a/src/ui/components/App.tsx
+++ b/src/ui/components/App.tsx
@@ -129,7 +129,7 @@ function App({ theme, recordingId, modal, updateNarrowMode }: AppProps) {
 
   useEffect(() => {
     setUserInBrowserPrefs(auth.user);
-    if (auth.user && !userData?.internal) {
+    if (auth.user && userData && !userData.internal) {
       LogRocket.createSession(auth);
     }
   }, [auth.user]);
--- a/src/ui/components/App.tsx
+++ b/src/ui/components/App.tsx
@@ -26,6 +26,7 @@ import { ModalType } from "ui/state/app";
 import useToken from "ui/utils/useToken";
 var FontFaceObserver = require("fontfaceobserver");
 import "styles.css";
+import { useGetUserInfo } from "ui/hooks/users";
 
 function AppModal({ modal }: { modal: ModalType }) {
   switch (modal) {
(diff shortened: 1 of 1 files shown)


src/ui/components/App.tsx
```diff
@@ -103,7 +129,7 @@ function App({ theme, recordingId, modal, updateNarrowMode }: AppProps) {
 
   useEffect(() => {
     setUserInBrowserPrefs(auth.user);
-    if (auth.user) {
+    if (auth.user && !userData?.internal) {
```
hbenl, line 132: This will probably not work as intended: when you use the `useQuery` hook in a component, that component is usually rendered twice (the first time while the query is still loading and the second time with the loaded result data, unless the query has already been cached): during the first render, `userData` is `null` and hence `!userData?.internal` is `true`, so you will create a Logrocket session before finding out if this is an internal user or not. Use something like `if (auth.user && userData && !userData.internal)` instead.

hbenl (changes requested):

src/ui/components/App.tsx
jazzdan: Ah good call @hbenl, I was unclear on how `?.` would behave in this situation.

jaril:
There's a `GetUserInfo` query available in `users.ts` that should simplify this a bit!

https://github.com/RecordReplay/devtools/blob/master/src/ui/hooks/users.ts#L5-L23

jazzdan:
Ah thanks @jaril!

src/ui/components/App.tsx
```diff
@@ -79,7 +81,10 @@ function App({ theme, recordingId, modal, updateNarrowMode }: AppProps) {
   const auth = useAuth0();
   const { claims } = useToken();
 
-  setTelemetryContext(claims?.hasura.userId, auth.user?.email);
+  const userId = claims?.hasura.userId;
+  const { isInternal, email } = useGetUserInfo();
+
+  setTelemetryContext(userId, email);
```
jasonLaster, line 87: I wonder if we should have`setTelemetryContext` here or next to  `createSession`

jazzdan: It _is_ a bit confusing but they're separate things. Set telemetry context doesn't determine whether the telemtry gets sent, only what context is attached.

jaril (approved):

PR #2425, opened 2021-04-26
Show replays that users have commented on their library

### LINKED ISSUE #2406: [UX] No way to see Replays I've commented on in the Library
1. We need to change the query to show all Replays (yours, ones you're invited to, and ones you've commented on)
2. With that in place, we can close this ticket and use https://github.com/RecordReplay/devtools/issues/2407 for other UX features we want in the library like a search box and filter dropdowns.

Read the design guidance here: https://www.notion.so/replayio/Library-Filtered-View-eb7d0cc6e7894d3ea6ebdf696d46b823

jaril:
Fix #2406 .

### DIFF
--- a/src/ui/components/Dashboard/Dashboard.tsx
+++ b/src/ui/components/Dashboard/Dashboard.tsx
@@ -39,11 +39,9 @@ function PersonalDashboard() {
     return <Loader />;
   }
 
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):
This looks like a good short term solution

PR #2429, opened 2021-04-27
Use the new GraphQL API (#2329)

hbenl:


src/ui/components/shared/SettingsModal/SettingsBodyItem.tsx
```diff
@@ -66,10 +62,9 @@ function Dropdown({
       setShowRefresh(true);
     }
 
-    updateUserSetting({
+    updateDefaultWorkspace({
       variables: {
-        newValue: selectedValue,
-        userId,
```
loganfsmyth, line 72: So glad we don't have to pass this all over now.

src/ui/hooks/recordings.ts
```diff
@@ -303,25 +312,28 @@ export function useGetWorkspaceRecordings(
...
-          last_screen_mime_type
-          duration
-          description
-          date
-          is_private
-          user {
-            name
-            email
-            picture
+      query GetWorkspaceRecordings($workspaceId: ID!) {
```
loganfsmyth, line 315: Interesting, I was assuming we'd do this query using `uuid` from the workspace, and `workspace(uuid: $workspaceId) {`. It does work either way though. How do you feel about the `id` vs `uuid` stuff?

src/ui/components/Comments/VideoComments/VideoComment.tsx
```diff
@@ -8,6 +8,7 @@ import { Canvas } from "ui/state/app";
 import { Comment } from "ui/state/comments";
 
 function inCenter(canvas: Canvas, { position }: Comment) {
+  if (!position) return true;
```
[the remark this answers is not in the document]
hbenl: No, I haven't investigated this any further.

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -37,11 +38,12 @@ export default function ReplayInvitations() {
     );
   }
 
-  const inviteCount = invitations.length;
-  // Replay folks get unlimited invites.
-  const maxInvites = isInternal ? Number.POSITIVE_INFINITY : 5;
-  const label = `You have ${maxInvites - inviteCount} invite${
-    maxInvites - inviteCount == 1 ? "" : "s"
+  if (availableInvitations === 1234567890) {
```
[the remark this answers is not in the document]
hbenl: I preferred to have that number in the code rather than in the UI. Since it's only visible to us, that preference isn't strong though.

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -77,12 +79,12 @@ export default function ReplayInvitations() {
         </form>
       )}
       <div className="invitations-list">
-        {invitations.map((invite, i) => (
+        {invitations!.map((invite, i) => (
           <div className="invitations" key={i}>
-            <div className={`material-icons ${invite.invited_user?.invited && "finished"}`}>
-              {invite.invited_user?.invited ? "check_circle" : "pending"}
+            <div className={`material-icons ${invite.pending || "finished"}`}>
+              {invite.pending ? "pending" : "check_circle"}
```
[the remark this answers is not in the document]
hbenl: Currently, `pending` is always set to `false`, but I left the code there in case we want to bring back the ability to see the pending state in the UI.

src/ui/hooks/recordings.ts
hbenl: I use only one kind of ID for each table, `uuid` for recordings and `id` for everything else. I think it's a bit confusing for API clients that there are two kinds of IDs, but not a big problem.

src/ui/components/DevTools.tsx
```diff
@@ -102,15 +95,6 @@ function DevTools({
   }
 
   if (!loaderResult) {
-    if (deleted_at) {
-      expectedError = { message: "This replay has been deleted." };
-    }
-
-    // Test recordings don't have a user, so we skip this check in that case.
-    if (user && user.invited == false) {
-      expectedError = { message: "The author of this Replay has not activated their account yet." };
```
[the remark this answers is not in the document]
hbenl: If I understand this correctly, it is actually a bug: it assumes that `invited` will be `true` for every user that has logged in at least once ("activated their account"), but that doesn't make any sense because users are only created when they login for the first time. Instead, `invited` should be `false` for users that logged in without receiving an invitation, but I don't think we want to block those users' recordings.
This bug was never noticed because it was neutralized by another bug that set `invited` to `true` even for users that didn't receive an invitation. That bug actually also neutralized another bug: instead of looking at invitations that the user has received, the code for setting `invited` looks at invitations that the user has sent...
So no, I don't think we need `recording.ownerNeedsInvite` (at least not for now).

src/ui/components/App.tsx
```diff
@@ -108,7 +104,7 @@ function App({ theme, recordingId, modal, updateNarrowMode }: AppProps) {
 
   useEffect(() => {
     setUserInBrowserPrefs(auth.user);
-    if (auth.user && !isInternal) {
+    if (auth.user && availableInvitations === 1234567890) {
```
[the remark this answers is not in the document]
hbenl: @jazzdan What do you think?
@loganfsmyth Perhaps we want to expose `isInternal` on `viewer.user` (i.e. you can't see other users' `isInternal` flag)?

src/ui/hooks/invitations.ts
```diff
@@ -1,135 +1,79 @@
...
+          availableInvitations
         }
       }
-    `,
-    { refetchQueries: ["GetRecording"] }
+    `
   );
 
-  return claimInvitation;
+  return { availableInvitations: data?.viewer.availableInvitations, loading, error };
```
loganfsmyth, line 61: `viewer?.`

loganfsmyth: Looks like `viewer` is assumed to exist in a few of the hooks

src/ui/components/DevTools.tsx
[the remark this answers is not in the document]
loganfsmyth: >  it assumes that invited will be true for every user that has logged in at least once ("activated their account"), but that doesn't make any sense because users are only created when they login for the first time. Instead, invited should be false for users that logged in without receiving an invitation, but I don't think we want to block those users' recordings.

My understanding is that, if a user logs in and has not been invited, their account would have `invited: false` until someone sends them an invitation, at which point their account will have `invited: true` set. The goal being that if you log in, you should be able to comment, but if you make a recording, you're not able to view it until you've been invited?

I'm fine to remove that logic, but should that be a bigger conversation?

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -27,7 +28,7 @@ export default function ReplayInvitations() {
     );
   }
 
-  if (!invited) {
+  if (availableInvitations === null) {
```
loganfsmyth, line 31: How do you feel about this check? I admit I'm tempted to add `internal` and `invited` booleans on `viewer` so if you haven't been invited, it's just `availableInvitations === 0`

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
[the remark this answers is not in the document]
loganfsmyth: Overall I don't mind if we want to transform it for readability, but I think I'd rather have it be a more meaningful value if that is the case, so I think I'll change the 'internal' value to be `2**32-1` so it's a clearly-defined int-max value.

src/ui/hooks/recordings.ts
```diff
@@ -119,18 +147,23 @@ export function useGetOwnersAndCollaborators(recordingId: RecordingId) {
     console.error("Apollo error while getting owners and collaborators", error);
   }
 
-  const collaborators: CollaboratorDbData[] = data.collaborators;
-  const recording = data.recordings_by_pk;
+  const collaborators: CollaboratorDbData[] = data.recording.collaborators.edges.map(
```
loganfsmyth, line 150: This needs a `?` on `collaborators` because the new API returns `null` if you aren't a collaborator/workspace member/owner, e.g. `data.recording.collaborators?.edges.map(`

src/ui/components/DevTools.tsx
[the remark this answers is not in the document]
hbenl: I have reinstated the old logic around invitations, which will mark any user as invited as soon as they login.
We should definitely clarify what's the intention here.

src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
hbenl: > I admit I'm tempted to add internal and invited booleans on viewer

Yes please

loganfsmyth: I forgot to ping you on the issue, but I landed this last night https://github.com/RecordReplay/backend/pull/1827

loganfsmyth: That change also means this is never going to pass right now, since doesn't return `null` anymore.

PR #2433, opened 2021-04-28
Hide material icons if the icon font is not loaded yet

jaril:


src/ui/components/shared/MaterialIcon.css
```diff
@@ -0,0 +1,7 @@
+.material-icons {
+  color: var(--theme-icon-color);
+}
+
+.material-icons.invisible {
+  visibility: hidden;
```
ryanjduffy, line 6: You might use `color: transparent` instead of hiding the entire element. As it stands, if the fonts fail to load, the user menu is completely hidden rather than just the three dots being hidden. You might always want to set some default bounds in this case since the natural width will be much larger with font ligatures when the font doesn't load.

PR #2435, opened 2021-04-28
getLoggedInUser sometimes returns nothing

jazzdan:
Let's handle that case and encode it in TypeScript

### DIFF
--- a/src/ui/components/App.tsx
+++ b/src/ui/components/App.tsx
@@ -62,8 +62,12 @@ function installViewportObserver({ updateNarrowMode }: Pick<AppProps, "updateNar
   observer.observe(viewport!);
 }
 
-function setTelemetryContext(userId: string | undefined, userEmail: string | undefined) {
-  let sentryContext: Record<string, string> = {};
+function setTelemetryContext(
+  userId: string | undefined,
+  userEmail: string | undefined,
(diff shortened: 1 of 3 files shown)


jazzdan:
To be clear: this is because the person viewing the recording isn't logged in and the recording is public :D

src/ui/components/App.tsx
```diff
@@ -81,8 +82,13 @@ function App({ theme, recordingId, modal, updateNarrowMode, setFontLoading }: Ap
   const { claims } = useToken();
   const { availableInvitations } = hooks.useGetAvailableInvitations();
 
-  setTelemetryContext(claims?.hasura.userId, auth.user?.email);
-  // const { loading } = hooks.useMaybeClaimInvite();
+  const userId = claims?.hasura.userId;
+  const userInfo = useGetUserInfo();
+  let isInternal = false;
```
jaril, line 87: Looks like we're not really using `isInternal` anywhere here—are we?

jaril (approved):

src/ui/components/App.tsx
jazzdan: Yeah, I meant to set that in the telemetryContext. Will fix. Thanks!

PR #2437, opened 2021-04-28
Allow collaborators to access the share modal

### LINKED ISSUE #2400: Collaborators should be able to add other collaborators
**Description:**

At the moment, only replay owners can add collaborators. This in practice an be quite limiting. I think the next steps should be to allow other collaborators to be able to add folks. At some point we'll have more fine grained permissions, but this is a reasonable next step since it unblocks some use cases

jaril:
Fix #2400.

Demo: https://share.descript.com/view/lkxr8aPJmRa


jaril:
Demo after rebase: https://share.descript.com/view/3YEEZa4K6Tl

PR #2438, opened 2021-04-28
Add URL parameters to identify users coming from an email link

### LINKED ISSUE #2416: Our invite link needs a param so we can pass along a better login message
Here's the current flow:

image

We need to make sure the login screen can show special text like "Almost there! First we need you to log in." To do that, we should pass a param in the URL like emailinvite=true.

jaril:
Fix #2416.

image


jasonLaster (approved):

jonbell-lot23:
Manifique! I like how you designed this screen on your own by leaning on existing precedent. Looks great.

PR #2459, opened 2021-04-30
Allow users to add a comment from the breakpoint widget

### LINKED ISSUE #2358: [UX] Add comment to a line of code
We need an entry point for adding a comment to a line of code:

jaril:
Fix #2358.

This adds an "Add a comment" button to the breakpoint widget. This button is disabled except for when the user is currently paused on one of the breakpoint widget's hits. In that case, clicking the button forces the transcript panel to open and opens the comment editor at for that hit in the transcript.

Demo: 
https://share.descript.com/view/n4NyFQzS1Ra

Screenshot:
image

### DIFF
--- a/src/devtools/client/debugger/src/components/App.css
+++ b/src/devtools/client/debugger/src/components/App.css
@@ -12,16 +12,6 @@ button {
   font-size: inherit;
 }
 
-.debugger button:hover,
-.debugger button:focus {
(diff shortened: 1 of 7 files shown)


jaril:
Stripe 0 was failing so reran the tests to make sure it's not busted.

jaril:
1 cancelled after 15m

PR #2471, opened 2021-04-30
Adds video player support

### LINKED ISSUE #2489: Add streaming video player to front end
* Integrate with back end video encoder to enable video playback on the front end
* Paint screenshots when paused and jumping to time

jasonLaster:
Adds a `<video>` player which is activated instead of the `<canvas>` screen paints when playing. When the player is paused / warped to a time, the `<canvas>` screen paint takes over and the video's playhead is moved to the correct time though hidden by the canvas.

Fixes #2489


src/protocol/graphics.ts
```diff
@@ -122,6 +125,100 @@ function onMouseEvents(events: MouseEvent[], store: UIStore) {
   store.dispatch(actions.setEventsForType(gMouseClickEvents, "mousedown"));
 }
 
+export const Video = {
```
hbenl, line 128: I would prefer a real class here.

src/protocol/graphics.ts
```diff
@@ -122,6 +125,100 @@ function onMouseEvents(events: MouseEvent[], store: UIStore) {
...
+    //     this.sourceBuffer.addEventListener("updateend", () => this.mediaSource.endOfStream());
+    //     // Try to flush the first buffered fragment if it exists
+    //     this.flush();
+    //   });
+    // }
+  },
+  onVideoEvent(fragment: string, store: UIStore) {
+    this.render(fragment);
+  },
+  render(fragment: string) {
```
hbenl, line 156: we could remove `onVideoEvent()` and `render()` and just use `append()`

src/ui/utils/prefs.js
```diff
@@ -43,6 +43,7 @@ pref("devtools.features.transcriptHover", false);
 pref("devtools.features.widgetHover", false);
 pref("devtools.features.reactDevtools", false);
 pref("devtools.features.smoothPlayback", true);
+pref("devtools.features.videoPlayback", true);
```
hbenl, line 46: I don't think we want to enable this by default (yet)

hbenl (commented):
- I don't see any video when trying this - should I or is there some backend stuff still missing?
- during playback, the timeline is still bound to the old screenshot download logic, so it will stall if the screenshots can't be downloaded fast enough and the timeline and the video playback will be out of sync

src/protocol/graphics.ts
ryanjduffy: I left these in intentionally to aid in migrating to the original architecture using `MediaSource` but I'm content to remove them.

src/ui/utils/prefs.js
ryanjduffy: Works for me!

ryanjduffy:
Thanks for the feedback, @hbenl! 

You should see the video with the current production back end. That said, it isn't very noticeable (by design) since it is only visible when playing and is obscured by the canvas when paused. Happy to review it together tomorrow if that's helpful so I can see what you are seeing.

Could use some pointers on the screenshot loading comment. I understand what you mean but I'm not sure yet how to reproduce it. If you could share a replay, that'd be handy!

ryanjduffy:
Pushed an update for the specific comments. Let me know what else you see!

PR #2493, opened 2021-05-03
sentry: set release version

jazzdan:
1) Use earthly to build + deploy JS in ci
2) Pass through the git SHA as a compile time constant
3) Use the git sha as the release version for Sentry

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+const webpack = require("webpack");
 const Visualizer = require("webpack-visualizer-plugin");
 
 module.exports = {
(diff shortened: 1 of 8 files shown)


webpack.config.js
```diff
@@ -30,6 +31,7 @@ module.exports = {
   plugins: [
     new MiniCssExtractPlugin(),
     process.env.REPLAY_BUILD_VISUALIZE && new Visualizer(),
+    new webpack.EnvironmentPlugin({ REPLAY_RELEASE: undefined }),
```
jasonLaster, line 34: should the release by undefined?

hbenl: `undefined` is just the default value that will be used if `REPLAY_RELEASE` is not defined in `process.env`. However, maybe we should use `"development"` as the default value (which is also used as the default in another place in this PR).

hbenl (approved):

webpack.config.js
jazzdan: According to the webpack docs if you make it undefined Webpack yells at you when you don't define it, which I thought was a nice side-effect. We handle it being falsey in the code and default to "development" there.

Issue #2285, opened 2021-04-07
React Devtools: Use latest Devtools frontend


jasonLaster:
**Description:**

We currently use a custom version from a couple weeks ago. Once we have a new version to use, we can try using the frontend in an iframe, which should fix some of the issues we're seeing.

PR #2501, opened 2021-05-04
Update react-devtools-inline to version 4.13.0 (#2285)

hbenl:
Note that this also updates React to an experimental version, because the latest react-devtools-inline version uses experimental React APIs that are not available in a stable release yet.


hbenl:
PR for the backend: RecordReplay/gecko-dev#346

PR #2510, opened 2021-05-04
Add optimistic delete

### LINKED ISSUE #2448: Library: replays should immediately disappear
**Description:**
When I move replays, they should immediately disappear from the UI (Optimistic updates)

jaril:
Fix #2448.

Not quite there yet. This shows optimistic delete as long as you 1) only mutate one thing at a time, and 2) do it very slowly.

Optimistic responses work by editing the cache directly immediately after a mutation request is sent and inserting the expected response for a particular query+variables combination. 

This approach appears to introduce race conditions when multiple requests are in flight. This happens when you perform multiple mutations back-to-back, like performing batch actions (which currently just applies the same mutation n number of times) or deleting 3 recordings one after another quickly.

It's _better_ than what we have now, but I'd like to spend more time on it to figure out how to make it more robust.

Latest Demo: https://share.descript.com/view/xnzTaxmqmYs


hbenl:
What's the goal here? Is it to update the cache before the server responds (which usually takes less than 100ms, so is that really worth it?) or to update the cache without having to refetch the data? If it's the latter, then that should already work for the `updateRecordingWorkspace` mutation because that mutation fetches the changed data with the mutation result. I don't know how to tell Apollo about the result of the `deleteRecording` mutation (or any mutation that deletes stuff), perhaps that would need to be done in javascript.
In any case, this should not be done by the caller(s) of the `updateRecordingWorkspace`/`deleteRecording` methods, but by those methods themselves.

jasonLaster:
The concern at a high level is that it can take awhile for the deleted recordings to actually be removed. 

Here is a loom video, but you probably will see something similar
https://www.loom.com/share/a412c7c8041a420bab04549c7127925d

hbenl:
> The concern at a high level is that it can take awhile for the deleted/moved recordings to actually be removed.

Yeah, I saw those slow updates as well, that's also why I made the mutations fetch their own result. It can still take a while for the UI to update and I didn't investigate any further. I just wanted to say that the optimistic updates may not be what we need to fix this issue because they can only speed things up by some 100ms compared to what we have now.

jasonLaster:
Thanks.

- I think you're right that the problems are deeper than networking. I just checked the network monitor and confirmed your findings. Also, the performance profile is quite alarming
- I think it would be nice to land the optimistic update. In general, we should aim to be under 100ms and this is a good pattern.

Will follow a couple follow up items for general perf

PR #2518, opened 2021-05-05
Updated timeout screen to "Replays disconnect after five minutes to reduce server load. Ready when you are!"

jonbell-lot23:
Was "Apologies! A refresh should do the trick" which implied failure.

Now is "Replays disconnect after five minutes to reduce server load. Ready when you are!" which explains what happened and that it's all a perfectly normal and healthy thing.


jonbell-lot23:
Messed up, trying again

Issue #1673, opened 2021-02-02
Use arrow keys to go forward/backward a frame


jasonLaster:
**Description:**
A couple of months ago we removed the ability to forward/backward a frame at a time. This is actually a nice feature to have when you want to see what react will paint next or just see what the ui will look like next in general.

I propose we add support for arrow key (left/right) nav and then slowly add context menus and other UIs...

jasonLaster:
@jonbell-lot23 what do you think of this? my hunch is we probably just need to be sensitive to do it on focus state and maybe think tooltips...

jonbell-lot23:
I think this is great. Here's my proposal.

* Add the keys with no UI.
* Obviously they shouldn't do anything in the console, but should do things elsewhere.
* While we're at it, clicking on an event in the transcript should allow up/down arrow keys to navigate as well. The up/down is a far more visible thing since it's such a pillar of the UI, and it's basic accessibility thing. So I think they should live together.

Then we can work out HUD/tooltips/etc based on what we learn from that.

jonbell-lot23:
It's been a month, dusting off this item.

@jaril I think the current takeaway is that we should add left/right arrow keys to step forward and back.

hbenl:
How do I get the ExecutionPoint for the next/previous frame?

PR #2528, opened 2021-05-06
Use arrow keys to navigate in time (#1673)

hbenl:


hbenl:
Note that this seeks to the previous/next paint instead of the previous/next frame as stated in the issue.

jasonLaster (approved):

PR #2530, opened 2021-05-06
Use a checkbox for the upload form

### LINKED ISSUE #2513: UploadScreen Enter should confirm
**Description:**

This is a minor detail, but I think that pressing enter while editing the title should confirm the Replay. At the moment it toggles the publicity field which feels a bit weird. 

I've made this mistake several times
Screen Shot 2021-05-04 at 5 27 48 PM

jaril:
Fix #2513.


jasonLaster:
Oh sad. I liked the toggle, i think we could handle this in JS right?

PR #2537, opened 2021-05-06
deploy: pass git SHA as input param, take 2

jazzdan:


.github/workflows/deploy.yml
```diff
@@ -7,17 +7,17 @@ on:
...
         uses: actions/checkout@v2
-      - name: Install Dependencies
-        run: npm install
+      - name: Download earthly
+        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.5.11/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
       - name: Build webpack
-        run: ./node_modules/.bin/webpack --mode=production
-      - name: Tar up output files
-        run: tar -czf dist.tgz index.html dist node_modules
+        run: "earthly --build-arg REPLAY_RELEASE=$GIT_SHA +dist"
```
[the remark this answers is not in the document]
loganfsmyth: So I understand right, this now means the deploy workflow will build devtools twice, once to create `dist.tgz` and once to actually deploy, so we're relying on the earthly cache to not duplicate too much work?

package.json
```diff
@@ -95,6 +95,7 @@
     "cypress": "^6.5.0",
     "cypress-localstorage-commands": "^1.4.0",
     "cypress-real-events": "^1.2.0",
+    "esbuild-loader": "^2.12.0",
```
loganfsmyth, line 98: Looks like this is leftover from some earlier esbuild attempt.

webpack.config.js
```diff
@@ -30,6 +31,7 @@ module.exports = {
   plugins: [
     new MiniCssExtractPlugin(),
     process.env.REPLAY_BUILD_VISUALIZE && new Visualizer(),
+    new webpack.EnvironmentPlugin({ REPLAY_RELEASE: undefined }),
```
loganfsmyth, line 34: Hmm, I'd expect this to mean that
```
release: process.env.REPLAY_RELEASE ? process.env.REPLAY_RELEASE : "development",
```
becomes
```
release: undefined ? undefined : "development",
```
and thus is always development, but I'm not totally sure on that.

.github/actions/deploy/main.js
```diff
@@ -30,7 +30,7 @@ function invalidateCloudFront() {
 
 spawnChecked("npm", ["install"]);
 
-spawnChecked("./node_modules/.bin/webpack", ["--mode=production"]);
+spawnChecked("earthly", ["--build-arg", "REPLAY_RELEASE=$INPUT_GIT_SHA", "+dist"]);
```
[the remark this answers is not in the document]
loganfsmyth: And also here for the ARG being `GIT_SHA`, not `REPLAY_RELEASE`

.github/actions/deploy/action.yml
```diff
@@ -1,6 +1,11 @@
 name: "Deploy"
 description: "Deploy to S3"
 
+inputs:
+  git_sha:
+    description: "git SHA of code being deployed"
+    required: true
+
 runs:
   using: "node12"
```
loganfsmyth, line 10: Should we change this to node14 too to match the new Earthly environment? Looks like there's one other reference in another action too.

loganfsmyth (changes requested):

.github/workflows/deploy.yml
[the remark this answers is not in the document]
jazzdan: Yes, this is the way it works in master too except it never caches. The deploy task kicks off two jobs:

1) Builds webpack
2) Builds webpack and deploys

2 only runs if 1 succeeds, I think

A little wasteful but I'm already changing a ton of things. :P

webpack.config.js
loganfsmyth: Oh dang I'm wrong on this one, I guess this uses the env var instead if it is present, missed that.

.github/actions/deploy/action.yml
jazzdan: I might be more inclined to change the earthly environment to be node 12, to reduce the number of things I'm changing in this patch

.github/actions/deploy/main.js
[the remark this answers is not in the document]
jazzdan: >Does github auto-populate INPUT_GIT_SHA from the git_sha input?

That's my understanding, yeah. I can add a check in here that enforces that that environment variable is present.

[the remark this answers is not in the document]
loganfsmyth: Oh also, is this actually doing to expand? Since this is in Node, it may need to be

    `REPLAY_RELEASE=${process.env.INPUT_GIT_SHA}`

package.json
jazzdan: Yeah thanks

.github/actions/deploy/main.js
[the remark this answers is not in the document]
jazzdan: Yeah that's what I did in the most recent version

package-lock.json
```diff
@@ -5453,6 +5453,70 @@
         "is-symbol": "^1.0.2"
       }
     },
+    "esbuild": {
```
loganfsmyth, line 5456: This can be reverted now that you remove the `package.json` line.

loganfsmyth (approved):

PR #2548, opened 2021-05-07
Add contenthash to bundle filenames

hbenl:


hbenl:
We keep seeing errors in Sentry that look like they could be fixed by this. But first we need to ensure that the content hash for the main bundle is added in our `index.html`.

ryanjduffy:
Would a querystring content hash provide the same cache-busting while being backwards compatible with `main.js`?

hbenl:
Yes, putting the hash in a query string should work, I updated the PR accordingly. However, I think we can't use this if we want to deliver our bundles using a CDN, because it won't cache URLs containing query strings.
I experimented a bit more with putting the content hash in the filename: getting our `index.html` rewritten by webpack is pretty simple with HtmlWebpackPlugin. But then we also need to teach webpack about loading the workers - this is very different in webpack 5 vs. webpack 4, so I tried out webpack 5 and found that its bundle splitting seems to work differently (our main bundle is a lot bigger with webpack 5 vs. webpack 4), so that will require more investigation. Furthermore, we'll need to change our code for loading the workers so that webpack understands it.
@jlast Perhaps we want to use the content-hash-in-query-string solution for now? If we want to deliver via CDN, we should replace it with the content-hash-in-filename solution before we go to market, however.

jasonLaster:
@hbenl `content-hash-in-query-string` is fine for now as long as we can follow up. I'm also okay switching to Webpack 5 and then fixing the perf later.

Generally, lets fix the issue asap and then improve down the line

PR #2553, opened 2021-05-07
Add retryImport to add resiliency to dynamic imports

### LINKED ISSUE #2547: Dynamic Imports can occasionally fail
Sentry Issue: [DEVTOOLS-QH](https://sentry.io/organizations/replay/issues/2387206807/?referrer=github_integration)

```
TypeError: Cannot read property 'call' of undefined
  at __webpack_require__ (webpack/bootstrap:63:22)
  at editor (/src/devtools/client/webconsole/components/Input/JSTerm.js:14:22)
  at async d.componentDidMount (/src/devtools/client/webconsole/components/Input/JSTerm.js:56:10)
```

ryanjduffy:
This will add some resiliency to dynamic imports by retrying 5 (by default) times. It might not fix #2547 just yet because it appears that in that case the promise would resolve but with an empty object. However, this fix will give us an entry point to further patch for that scenario too by checking the response and then retrying (after possibly cleaning up webpack cache ...).

### DIFF
--- a/src/devtools/client/debugger/src/utils/editor/create-editor.js
+++ b/src/devtools/client/debugger/src/utils/editor/create-editor.js
@@ -5,6 +5,7 @@
 //
 
 import { assert } from "protocol/utils";
+import retryImport from "utils/retryImport";
 import { features, prefs } from "../prefs";
 
 let editorWaiter;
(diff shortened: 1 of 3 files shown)


ryanjduffy:
I looked into this further and I'm not sure it'll fix the partially loaded module issue. Once webpack fetches a chunk, it flags it as fetched in a module-scoped variable (`installedChunks`) and there doesn't seem to be a way to force it to refetch it.

PR #2558, opened 2021-05-07
Add optimistic comment mutations

jaril:


ryanjduffy:
This may not be new information but sharing nonetheless. I think optimistic updates are a great idea and really help with a "snappy" UX. One side effect I noticed when testing this was that the user's avatar "blinks" when the update comes in. This is caused by the `key` set around the `CommentCard` changes causing React to dump and re-create the node and its ``.

I explored adding some non-db metadata that maintains the `id` across the `update()`. This worked well but then was lost when the apollo subscription fires and replaces the non-db data. I think this approach has promise assuming we can hook into that subscription process.

Here's the code for reference. Ignore the extra commits from basing it on the PR.

https://github.com/jaril/webreplay-devtools/compare/optimistic-comments...ryanjduffy:add-opt-model?expand=1

jasonLaster:
Thanks @ryanjduffy. That's a great suggestion

hbenl (approved):

ryanjduffy:
@jaril - Still seeing the same flashing of the avatar

jaril:
@ryanjduffy Opened a separate issue for the avatar flash: https://github.com/RecordReplay/devtools/issues/2566

PR #2559, opened 2021-05-08
Tightening this div to make the text look less wide

jonbell-lot23:
Old:
image

New:
image


ryanjduffy:
This is better, I think. As an aside, I think the "Ready when you are" feels a bit out of place there. I like that copy as the heading text above the icon perhaps but the tone feels incongruent with the preceding sentence. That said, it's better color that was there before. 😁

jasonLaster (approved):

PR #2565, opened 2021-05-10
Fix more bugs in the ReactDevTools integration

hbenl:


src/ui/components/SecondaryToolbox/ReactDevTools.js
```diff
@@ -95,6 +112,7 @@ async function sendRequest(event, payload) {
 // TODO Pass custom bridge
 // TODO Use portal containers for Profiler & Components
 export function ReactDevtoolsPanel() {
+  ensureIsSetup();
```
jasonLaster, line 115: I think hooks should come first

src/ui/components/SecondaryToolbox/ReactDevTools.js
```diff
@@ -42,11 +49,17 @@ function InitReactDevTools() {
...
+          listener(msg);
+        } catch (err) {
+          console.warn("Error in ReactDevTools frontend", err);
+        }
+      };
     },
     async send(event, payload) {
       if (event == "inspectElement") {
-        if (inspected.has(payload.id)) {
+        if (inspected.has(payload.id) && !payload.path) {
```
jasonLaster, line 62: can you explain what is happening with `payload.path`? 

It looks like we might be able to add a guard at the top

```js
// <explain why we're skipping payload.path>
if (pathload.path) {
  return;
}
```

src/ui/components/SecondaryToolbox/ReactDevTools.js
hbenl: OK

src/ui/components/SecondaryToolbox/ReactDevTools.js
hbenl: Simply skipping if `payload.path` is set would break deep inspection of components' props and state: Only the top levels are initially transmitted to the ReactDevTools frontend and when you go deeper, more data is retrieved using an `inspectElement` message with `payload.path` set to identify the object that you want to inspect. The `inspected` Set should only be used for the `inspectElement` messages without `payload.path`.

PR #2581, opened 2021-05-11
use hooks

jasonLaster:


hbenl:
I have updated #2240 using the ideas from this PR

PR #2586, opened 2021-05-12
Add an association filter

jaril:
### Screenshot
image


jaril:
@hbenl I tried doing the filtering client side for now. How bad of a perf hit do you think this might be?

hbenl (approved):
I didn't see a measurable performance impact in my tests.

PR #2590, opened 2021-05-13
Show number of comments for each recording in the library

### LINKED ISSUE #2410: RPT-3 Library: New columns that include comment activity
Github has comment activity, and we should do a similar thing. Like this:

image

Notice how in this design we don't have an "activity" column, and instead we just tuck it in at the end. That's good for now, and the same thing that Github does.

I assume we get the talk button icon for free with Material icons, please let me know if not!

jaril:
Fix #2410.

image

### DIFF
--- a/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
+++ b/src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
@@ -6,6 +6,7 @@ import { AuthAvatar } from "ui/components/Avatar";
 import formatDate from "date-fns/format";
 import formatDistanceToNow from "date-fns/formatDistanceToNow";
 import hooks from "ui/hooks";
+import { ChatAltIcon } from "@heroicons/react/outline";
 import "./RecordingListItem.css";
 
 function getDurationString(durationMs) {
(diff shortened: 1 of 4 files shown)


jaril:
I have the same perf concern here as #2586 because of the change to the query. Thoughts? @hbenl

src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
```diff
@@ -70,6 +71,7 @@ function ItemDuration({ duration }) {
 }
 
 function ItemTitle({ data, editing, editingTitle, setEditingTitle, handleClickUrl }) {
+  console.log({ data });
```
hbenl, line 74: this should be remoevd

src/ui/hooks/recordings.ts
```diff
@@ -65,6 +65,11 @@ const GET_WORKSPACE_RECORDINGS = gql`
                 name
                 picture
               }
+              comments {
+                user {
+                  id
+                }
+              }
```
hbenl, line 72: you could use `comments { id }` here and in `GET_MY_RECORDINGS` instead - that would involve one table less in this query, so it could help with performance (although I didn't see a measurable impact in my tests)

hbenl (approved):
I didn't see a noticeable performance impact by this change

src/ui/hooks/recordings.ts
hbenl: I just saw that you need the user's id in #2586

src/ui/components/Dashboard/RecordingItem/RecordingListItem.js
jaril: fixed!

PR #2592, opened 2021-05-13
Visually consistent error screens

### LINKED ISSUE #2538: Whoops screen should be consistent with other login screens
image

This is how it looks now. Can we make it look like our new login screen?

jaril:
Fix #2538.

1) When a user has just signed up and their browser is blocking the auth0 popup.
image
2) When a user opens a private recording but they haven't signed in.
image
3) When a user opens a private recording that they don't have permissions for.
image
4) For unexpected session errors.
image
5) For unexpected socket errors.
image
6) For when the socket closes after 5 minutes.
image


src/devtools/client/themes/variables.css
```diff
@@ -384,16 +384,16 @@ certain that you **have** to use a z-index (e.g. element with position: absolute
...
+  --z-index-1: 10;
+  --z-index-2: 20;
+  --z-index-3: 30;
+  --z-index-4: 40;
+  --z-index-5: 50;
+  --z-index-6: 60;
+  --z-index-7: 70;
+  --z-index-8: 80;
+  --z-index-9: 90;
+  --z-index-10: 100;
```
jaril, line 396: This didn't play well with tailwind's default z-indices so I brought it down closer to their scale.

jasonLaster:
Wow! Looks so much better

jasonLaster (approved):

PR #2595, opened 2021-05-13
Url nullable

jasonLaster:


jasonLaster:
we can always revert when we have a better approach, but i'd like to unbreak the library for me and potentially others

Issue #2482, opened 2021-05-02
Invalid breakpoint locations


jasonLaster:
**Description:**
I noticed while playing with a replay of checkly that breakpoints in one of the files kept throwing invalid location

orignal replay: https://replay.io/view?id=447862fb-8716-454c-9df5-d9170561fbea&point=36995115119955335297599016873754794&time=13407.019675125219&hasFrames=true

Replay: https://replay.io/view?id=9a84a016-45d2-4637-a6cd-f69a942365c1&point=86646453860603916231275196140488022&time=39661.53658274155&hasFrames=true

loganfsmyth:
This appears to be because the frontend is breaking assumptions we make on the backend around round-tripping source file locations. The client is only supposed to send the backend original locations if the locations came from the backend originally, which means that it is unsafe to for instance get a position on one original source, and apply it to an different original source (even if they have the same URL and same content). That is exactly what is happening in this case with https://github.com/RecordReplay/devtools/blob/78165d6295d4fc95a9b8cadefc0c5503854880d3/src/protocol/logpoint.ts#L318

because a given URL may have multiple source actors for the given URL, but you can't give an arbitrary position for a URL unless the position came from the server first. That function in its present form seems to be inherently unsafe. 

When getting possible breakpoint positions, the frontend should query _all_ server-side sources for associated with the frontend's "source", and when you click on a line to analyze, we should be getting the set of `locations` to send in the analysis by getting the position for each server-side source, instead of getting the list of IDs and creating the list of positions based on the response for just one of the sources.

The quick fix here is probably that we can send `getPossibleBreakpointPositions` for every source but ignore the result and hope it is the same and it'll probably work, but it's a bit ugly and is very much relying on side-effects

In general the fact that the frontend tries to unify the views of multiple sources into one here is causing this confusion. Even if the frontend wants to present multiple files as a single file, it should still be individually performing operations on the individual sources or else things are going to crop up like this.

loganfsmyth:
I don't know if I'm the right person to dig into this more or if you want to look at it @hbenl

hbenl:
The root cause for this seems to be that we throw away the location's `sourceId` and then all we have to work with is the `sourceUrl`, and there are only 2 options in this case: try with all `sourceId`s for the given `sourceUrl` (even if that means potentially running into some errors) or give up straight away.
Luckily, we have the `sourceId` initially, it just gets thrown away here if we have a `sourceUrl`:
[...]

PR #2599, opened 2021-05-13
Request breakpointPositions for all identical sources (#2482)

hbenl:


src/devtools/client/debugger/src/actions/breakpoints/index.js
```diff
@@ -198,7 +199,25 @@ export function updateHoveredLineNumber(line) {
     // Set the initial location here as a placeholder to be checked after any async activity.
     dispatch(actions.setHoveredLineNumberLocation(initialLocation));
 
-    await dispatch(setBreakpointPositions({ sourceId: source.id, line }));
```
jasonLaster, line 201: This feels like something where we either want to update `setBreakpointPositions` to always do this for all of the sources that match the URL or add a new `setBreakpointPositionsForURL`.

src/devtools/client/debugger/src/actions/breakpoints/index.js
```diff
@@ -198,7 +199,25 @@ export function updateHoveredLineNumber(line) {
...
+    const promises = [];
+    promises.push(dispatch(setBreakpointPositions({ sourceId: source.id, line })));
+    if (source.url) {
+      // there may be multiple identical sources for the same URL (due to navigations),
+      // in this case we need to request breakpoint positions for all of them so that
+      // we can set logpoints in all of them later on
+      const sourceIds = ThreadFront.getSourceIdsForURL(source.url);
+      // sourceIds will only contain "preferred" sources, check if the current source
+      // is one of them (otherwise it's not identical to them)
+      if (sourceIds.contains(source.id)) {
```
jasonLaster, line 211: when is this not the case?

src/devtools/client/debugger/src/actions/breakpoints/index.js
```diff
@@ -198,7 +199,25 @@ export function updateHoveredLineNumber(line) {
...
+        for (sourceId of sourceIds) {
+          if (sourceId !== source.id) {
```
jasonLaster, line 213: I think we can probably simplify this by dropping the IF.


```js
if (source.url) {
  const sourceIds = ThreadFront.getSourceIdsForURL(source.url);
  return Promise.all(sourceIds.map(sourceId => Thingy(sourceId))
} else {
 return Thingy(sourceId)
}
```

jasonLaster (commented):
Looking good. Can we also add some end-to-end tests that cover these navigation cases?

hbenl:
Superseded by #2628

PR #2607, opened 2021-05-13
Add back early loading for DraftJS

jaril:
Put back pre-loading for DraftJS that used to be in `CommentThread` [here](https://github.com/Recordreplay/devtools/commit/40e18f80#diff-5cdf3eb46dcbc25d1f674aedfc2c77b186f72d47d692d726525044078b2332e2R29).

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { connect, ConnectedProps } from "react-redux";
 import classNames from "classnames";
 import { UIState } from "ui/state";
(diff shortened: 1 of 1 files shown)


src/ui/components/Comments/TranscriptComments/CommentCard.tsx
```diff
@@ -87,6 +88,18 @@ function CommentCard({
     pendingComment?.comment.time === currentTime &&
     pendingComment?.comment.point === executionPoint;
 
+  const load = useDraftJS();
```
ryanjduffy, line 91: You might move this to the top next to `useAuth0` so the hooks stay together which helps reduce the chances of accidentally bailing on render before a hook is called which makes React sad.

ryanjduffy (approved):

src/ui/components/Comments/TranscriptComments/CommentCard.tsx
jaril: fixed!

PR #2612, opened 2021-05-14
Optimize library checkboxes

### LINKED ISSUE #2523: Library: selecting a checkbox can take awhile
**Description:**

The checkbox when toggling replays for actions has an animation that can take awhile

Screen Shot 2021-05-05 at 10 50 26 AM

jaril:
Fix #2523.

We used to pass down an array of checked rows to every row. This triggered a re-render of all the rows every time a row was toggled.

This PR changes it such that when a checkbox is toggled, only the relevant row is re-rendered.


jasonLaster (approved):
nice

PR #2613, opened 2021-05-14
Add limits to how many Replays can be displayed

### LINKED ISSUE #2596: Library: Search input can be slow
**Description:**

It looks like the search input can kick off some expensive react / apollo work. One frame took 1.66 seconds for instance

Screen Shot 2021-05-13 at 8 23 09 AM

jaril:
Fix #2596.

We can't render 100+ rows of Replays reasonably snappy-ly. At some point we'll introduce proper pagination, but this is just for plugging the hole until we get there.

For now, this just adds a dropdown for limiting the number of Replays displayed at one time. Definitely not the most elegant solution since (outside of the time/association filters) there's no affordances for cases where you can see 1-20 but would like to see 20-40.

### DIFF
--- a/src/ui/components/Dashboard/DashboardViewer.tsx
+++ b/src/ui/components/Dashboard/DashboardViewer.tsx
@@ -15,6 +15,7 @@ const TIME_IN_MS = {
 
 type TimeFilter = "all" | "month" | "week" | "day";
 type AssociationFilter = "all" | "collaborator" | "comment" | "author";
+type Limit = "all" | "100" | "50" | "20";
 
 const subStringInString = (subString: string, string: string | null) => {
   if (!string) {
@@ -64,6 +65,7 @@ export default function DashboardViewer({ recordings }: { recordings: Recording[
   const [searchString, setSearchString] = useState<string>("");
   const [timeFilter, setTimeFilter] = useState<TimeFilter>("all");
   const [associationFilter, setAssociationFilter] = useState<AssociationFilter>("all");
+  const [limit, setLimit] = useState<Limit>("20");
 
   const toggleEditing = () => {
     if (editing) {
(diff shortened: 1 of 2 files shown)


jaril:
@jonbell-lot23 Any thoughts on how we can limit the number of replays we display on the screen without going straight to a full pagination UI?

src/ui/components/Dashboard/DashboardViewer.tsx
```diff
@@ -117,11 +119,23 @@ export default function DashboardViewer({ recordings }: { recordings: Recording[
...
+              selected={limit}
+              setSelected={value => setLimit(value as Limit)}
+              options={[
+                { id: "all", name: "Show all" },
+                { id: "20", name: "20" },
+                { id: "50", name: "50" },
+                { id: "100", name: "100" },
+              ]}
+              className="w-32"
+            />
```
jasonLaster, line 132: I'd probably leave this out for now because it shouldn't be as primary as the other dropdowns

src/ui/components/Dashboard/DashboardViewerContent.js
```diff
@@ -8,18 +8,22 @@ export default function DashboardViewerContent({
...
+  limit,
 }) {
   const [ascOrder, setAscOrder] = useState(false);
   let sortedRecordings = sortBy(recordings, recording => {
     const order = ascOrder ? 1 : -1;
     return order * new Date(recording.date);
   });
 
+  const displayedRecordings =
+    limit === "all" ? sortedRecordings : sortedRecordings.slice(0, +limit);
```
jasonLaster, line 20: I'd start with just this slice :) and perhaps a show more link.

jonbell-lot23:
Following up -- we chatted about this on the call. I'm fine with this for now, but the average user isn't really hitting this issue. I think when it's time to paginate, we should paginate. Until then, this is lower priority.

PR #2629, opened 2021-05-17
Add upload screen unload prompt

jaril:


jasonLaster:
I think i know why this is important, but can you clarify the thinking?

PR #2687, opened 2021-05-21
Updating guide

jonbell-lot23:
For now we're making the guide a single screen rather than five. We'll add more to it later, but this is better for now.

New copy:
image

Old copy:
(I can't easily grab it, sorry!)

### DIFF
--- a/src/ui/components/shared/OnboardingModal/OnboardingModal.tsx
+++ b/src/ui/components/shared/OnboardingModal/OnboardingModal.tsx
@@ -8,65 +8,13 @@ import Modal from "../NewModal";
 
 const slides = [
   {
-    header: "Welcome to your Library 👋",
-    content: (
-      <div className="text-xl">{`There's nothing here yet, so let's record your first replay.`}</div>
-    ),
-  },
-  {
-    header: "Open the Replay browser",
+    header: "Welcome to Replay! 👋",
     content: (
       <>
-        <div className="text-xl space-y-4">
-          <div>
-            {`First, download and install the browser by following the instructions `}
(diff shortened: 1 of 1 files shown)


jonbell-lot23:
Addresses https://github.com/RecordReplay/devtools/issues/2542

jasonLaster:
It feels a bit funny to me to teach people how to record a replay when they visit the library.

I feel like it's intuitive to welcome folks to replay and describe how great it is to 

1. share replays with others
2. annotate replays 
3. debug replay

jonbell-lot23:
I don’t have the link to motion handy but I described the thinking there.

This is the screen you land on when you install Replay. This is when people are in the “yeah yeah yeah whatever let me just use it” mindset. It’s a very easy and common mistake to think it’s the appropriate place to “celebrate the feature” as Microsoft says … but it isn’t. 

Best practice is to have a more JIT approach, when you talk about debugging in devtools. And you talk about annotation on first load of Viewer. And Share has a giant button in the nav, so talking about it before a person even had anything to share is not the place for it.

jonbell-lot23:
I’m going to play a “disagree and commit” card here :)

This is better than the prior version and since it’s just text we can always edit later with no short term drawback. Conversely, I am concerned with overthinking these sorts of changes even after we have discussed them.

jonbell-lot23:
Whoops! Accidentally closed

jonbell-lot23:
I'd like to get this live. It's not a great screen, but it addresses some feedback I keep hearing in calls. Can we push it live and keep iterating from there?

PR #2712, opened 2021-05-25
Query the new GraphQL API for user info.

loganfsmyth:
Depends on https://github.com/RecordReplay/backend/pull/2091

### DIFF
--- a/src/ui/components/Account/Library.tsx
+++ b/src/ui/components/Account/Library.tsx
@@ -57,12 +57,7 @@ function Library({ setWorkspaceId, setModal, currentWorkspaceId }: PropsFromRedu
   }, [workspaces, loading]);
 
   useEffect(() => {
-    if (
-      userInfo &&
-      userInfo.authoredRecordingCount === 0 &&
-      userInfo.nags &&
-      !userInfo.nags.includes(Nag.FIRST_REPLAY)
-    ) {
+    if (!loading && userInfo?.nags && !userInfo.nags.includes(Nag.FIRST_REPLAY)) {
       setModal("onboarding");
     }
   }, [userInfo]);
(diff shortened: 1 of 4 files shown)


src/ui/hooks/users.ts
```diff
@@ -31,44 +29,32 @@ export enum Nag {
...
-          }
-          recordings_aggregate {
-            aggregate {
-              count
-            }
-          }
           internal
           nags
+          recordings {
+            totalCount
```
loganfsmyth, line 40: This doesn't exist in the API right now, but hopefully I can remove our need for `authoredRecordingCount` entirely, then I'll just drop this part of the query.

src/ui/components/Account/Library.tsx
```diff
@@ -57,12 +57,7 @@ function Library({ setWorkspaceId, setModal, currentWorkspaceId }: PropsFromRedu
   }, [workspaces, loading]);
 
   useEffect(() => {
-    if (
-      userInfo &&
-      userInfo.authoredRecordingCount === 0 &&
-      userInfo.nags &&
-      !userInfo.nags.includes(Nag.FIRST_REPLAY)
-    ) {
+    if (userInfo?.nags && !userInfo.nags.includes(Nag.FIRST_REPLAY)) {
```
loganfsmyth, line 60: I ran some SQL on the database to set FIRST_REPLAY for all users who have recordings, so the only users without `FIRST_REPLAY` in their nags at this point should be users who have no recordings.

loganfsmyth: For history's sake, this is what I ran:
```
UPDATE users SET nags = nags||'["first_replay"]'::jsonb WHERE EXISTS (SELECT * FROM recordings WHERE users.id = recordings.user_id) AND NOT (nags ?| array['first_replay']);
```

jaril (approved):
LGTM!

Issue #2644, opened 2021-05-18
Use proper routes + code splitting


jasonLaster:
**Description:**

We should be able to use proper routes to separate out the library, upload screen, devtools, and viewer so that the pages load faster.

This will also lay the foundation for other pages like the welcome and new tab

jasonLaster:
This is kinda what i have in mind from a code perspective
https://github.com/RecordReplay/devtools/compare/master...jasonLaster:bundles?expand=1

ryanjduffy:
I started looking into this a couple months back but found it difficult at the time to cleanly break the various views up. Part of the challenge was the combined redux-related code (actions, reducers) which meant that triggering an action in the viewer will require dependencies from devtools (and vice versa).

As a result, I expect we'll need to take an incremental approach to move towards a "split-able" app. I think @jasonLaster's branch is a good first step by getting the router set up to divide the main views of the app without touching the organization much.

From there, instituting some structural conventions might help start to align the code better. For example, creating top level source directories for `utilities`, `components`, `features`, and `views` that contain shared assets at increasing levels of composition and then start to move files to align with these structure. The idea here is that each level should often reuse at lower levels, occasionally at the same level, and never at higher levels. So `component` will use `utilities` or other `components` but never a `feature` (which I think of as a large, composite component that implements a capability for multiple views). As you fill these folders out, it becomes easier to see the inter-dependencies and allows you to start bundling more effectively.

A few other general notes:
* It's great to decompose potentially shared assets into separate files but I tend to favor keeping those files in the higher levels (e.g. views and features) until they are actually reused. This helps maintain focus on the immediate problem being solved until the requirements for reuse are more clear
* I try to avoid combining combining code from unrelated business features. This usually manifests as utility files that cut across business domains or "export all" index files for components or features/hooks. They can make consumption a bit easier but they tend to get bloated and cause unrelated dependencies to be pulled into bundles.

jasonLaster:
Thanks @ryanjduffy That's a great writeup. Could you file a tracking issue for this? I can easily see us breaking this down into lots of sub-tasks.

- [ ] protocol
- [ ] ui/actions
- [ ] ui/utils


Also, are there easy ways to visualize the dependency graph? This is one of those things where the more we can visualize and assert the better. Perhaps also show bundle sizes in PRs

PR #2739, opened 2021-05-31
Use code splitting to make the library load faster (#2644)

hbenl:
- to switch between library and devtools, the app currently needs to be reloaded - it would be nicer (and probably more performant) to switch without reloading
- the size of the main bundle is now <2MB but it can be made even smaller (because it still contains lots of code that is unnecessary for the library)
- ~~the library is currently missing lots of css~~


src/main.js
```diff
@@ -22,85 +22,55 @@ if (test) {
...
-    window.mouseClientY = e.clientY;
-  });
-  window.elementIsHovered = elem => {
-    if (!elem) {
-      return false;
+  useEffect(() => {
+    async function importAndInitialize() {
+      const imported = await (recordingId ? import("./app") : import("./library"));
+      const pageWithStore = await imported.initialize();
+      setPageWithStore(pageWithStore);
```
jaril, line 48: Nice

jaril (approved):
Looks good to me

PR #2741, opened 2021-05-31
Add left-nav (Note: this code is probably not good)

jonbell-lot23:
This code should probably not go live without running it by Jaril and others. I was trying to learn some React techniques that are new to me, so the likelihood that I messed something up is high.

Old:
image

New:
image

On the offchance that this code is actually ok, great! The next thing I want to tackle is making the comments section toggle when you click the icon. I tried for a while but couldn't get it working.


jaril:
This is a duplicate of #2761 so closing.

PR #2761, opened 2021-06-02
Adding left nav on the Viewer

jonbell-lot23:
Hurray! I think this works.

Later I'd like to add a toggle so clicking the comments icon tucks comments away. But this is good for now!

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
@@ -11,7 +11,6 @@
   background: var(--theme-toolbar-background);
   border-radius: 4px;
   border: 1px solid var(--theme-splitter-color);
-  
   line-height: 16px;
   color: var(--theme-toolbar-color);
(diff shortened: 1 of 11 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
```diff
@@ -11,7 +11,7 @@
   background: var(--theme-toolbar-background);
   border-radius: 4px;
   border: 1px solid var(--theme-splitter-color);
-  font-family: sans-serif;
+  font-family: system-ui, Arial, sans-serif;
```
jaril, line 14: Let's go ahead and just delete the `font-family` styling altogether for these selectors. That way it conveniently defaults to the system-ui.

jaril (approved):

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
jonbell-lot23: True!

PR #2762, opened 2021-06-03
Add error page for browser errors

ryanjduffy:
Adds a generic error page which gecko can use to message unusuable recordings (or other future error conditions).

Notes:
* The title text is hard-coded in the view and the subtitle text comes from the browser.
* This PR adds new routing to isolate the browser-specific pages (currently just this one but likely new-tab and welcome-to-replay from the website later). I expect this will have impacts on the web server config for production that will also need to be resolved.
* This PR adds a new folder structure to house this. I think this will help migrate to better bundling in the future.

image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -26,17 +26,20 @@ const React = require("react");
 const { useEffect, useState } = React;
 const ReactDOM = require("react-dom");
 const { Provider } = require("react-redux");
-const { BrowserRouter: Router } = require("react-router-dom");
+const { BrowserRouter: Router, Route, Switch } = require("react-router-dom");
 const tokenManager = require("ui/utils/tokenManager").default;
 const { setupTelemetry } = require("ui/utils/telemetry");
 const { ApolloWrapper } = require("ui/utils/apolloClient");
 const App = require("ui/components/App").default;
+const BlankLoadingScreen = require("ui/components/shared/BlankScreen").BlankLoadingScreen;
 
 require("image/image.css");
 
 document.body.addEventListener("contextmenu", e => e.preventDefault());
 
 setupTelemetry({ recordingId });
+
+const BrowserError = React.lazy(() => import("views/browser/error"));
 
 function PageSwitch() {
   const [pageWithStore, setPageWithStore] = useState(null);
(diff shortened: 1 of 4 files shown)


jonbell-lot23 (approved):
From a text standpoint, I can modify this later. Thumbs up from me for now.

ryanjduffy:
@hbenl - Added you as a reviewer so you can weigh in on how I've integrated with the lazy loading.

src/main.js
```diff
@@ -26,18 +26,21 @@ const React = require("react");
...
 const App = require("ui/components/App").default;
+const BlankLoadingScreen = require("ui/components/shared/BlankScreen").BlankLoadingScreen;
 
 require("image/image.css");
 
 document.body.addEventListener("contextmenu", e => e.preventDefault());
 
 setupTelemetry({ recordingId });
 
+const BrowserError = React.lazy(() => import("views/browser/error"));
```
hbenl, line 42: This defeats the purpose of lazy loading: you're moving the BrowserError component into its own bundle and then import that bundle no matter if it's needed or not (and ensure that library, viewer and devtools aren't shown until that bundle is loaded).
I think making the BrowserError component lazy loadable is rather pointless since it's very small. It would be much more important to make all of the library/viewer/devtools code load lazily (so that it doesn't need to be loaded to show a simple browser error page) - but that is outside of the scope of this PR.

hbenl (changes requested):

src/main.js
ryanjduffy: That's not really what is happening here. `React.lazy()` defers the import until the component is rendered so that bundle (11.js in my last run) is only included when you hit the `/browser/error` route. The bundles for the sign in and library (0.js, 6.js, 1.js, and 9.js) are not included for that route and only delivered the root or `/view`. main.js is included in both paths. Should note as well that for the library/devtools paths, since this new bundle is not requested, it doesn't block rendering them.

Granted, the value of splitting this into a bundle is really small now but will increase as we are able to split the shared components / design system code out so that this really small page can be just the bootstrap, the shared components, and its view.

Hope that helps clarify what I'm seeing. Happy to review it together tomorrow if that'd be helpful.

hbenl: OK, I had misunderstood how `React.lazy()` / `React.Suspense` work together, thanks for clearing that up!

jaril (approved):

PR #2769, opened 2021-06-04
Moving icons to Material style

jonbell-lot23:
We have an annoying blend of line weights and positions in our current UI. This is our first step towards standardising them to the Material iconset. Then we'll keep iterating from here.

Old:
image

New:
image
--

Old:
image

New:
image


jonbell-lot23:
Icon swapped out after an unprecedented 4/4 unanimous agreement on the new icon:

image

PR #2778, opened 2021-06-05
Partial loading

jasonLaster:
When the recording is unloaded from the beginning we should factor that in.

such as it is here
http://localhost:8080/index.html?id=8d73591c-7e7c-486c-83fc-5662c85e66d6&point=649037107345178429090913237074067&time=707.7966951903217&hasFrames=true


jasonLaster:
http://localhost:8080/index.html?id=8d73591c-7e7c-486c-83fc-5662c85e66d6&point=649037107345178429090913237074067&time=707.7966951903217&hasFrames=true

PR #2779, opened 2021-06-05
Hide done circle

jasonLaster:


ryanjduffy:
I just noticed this as well. A little animation here would be a great addition for some user delight.

jonbell-lot23:
See also: https://github.com/RecordReplay/devtools/issues/2745#event-4845680960

Agree @ryanjduffy , I was imagining a fade for ease of dev. But would be super delightful to have a little animation of it blipping away.

Issue #2789, opened 2021-06-07
Getting breakpoint positions is slow on typescript playground


kannanvijayan:
Copied from https://github.com/RecordReplay/backend/issues/2205

**Description:**

It usually takes several seconds for hits to load, to the point where the UI feels like it has stalled.  See also this loom video https://www.loom.com/share/da5010eec37041e3a01f4c6d9eb9eba2 from @Andarist around 5:15, where things are so slow it is confusing and interrupting his UX.

**Steps to Reproduce:**

1. Visit https://replay.io/view?id=6d3dde62-b010-41d3-a51a-f4f37a17da7d
2. Wait for execution indexing to finish
3. Click on lines in typescriptServices.js and wait for hits to load

**Additional Context:**

Here is the session ID from my own experiments:
6b91bd2a-545b-4678-9028-200b6f131726/652806c8-ea95-4d9a-8a1f-566d1acf41c0

* Video (Loom):

https://www.loom.com/share/da5010eec37041e3a01f4c6d9eb9eba2

* Replay:

https://replay.io/view?id=6d3dde62-b010-41d3-a51a-f4f37a17da7d

PR #2792, opened 2021-06-08
Cache breakpoint positions (#2789)

hbenl:


src/protocol/thread/thread.ts
```diff
@@ -362,12 +366,22 @@ class _ThreadFront {
...
+
     const begin = range ? range.start : undefined;
     const end = range ? range.end : undefined;
     const { lineLocations } = await client.Debugger.getPossibleBreakpoints(
       { sourceId, begin, end },
       this.sessionId
     );
+
+    if (!range) {
+      this.breakpointPositions.set(sourceId, lineLocations);
```
bhackett1024, line 382: Can we store a promise in the map so that if there are multiple requests for the breakpoint positions before the first request finishes, we still only ask the server for the positions once?

bhackett1024 (approved):

bhackett1024 (changes requested):
I don't think this will behave correctly if a range is supplied the first time we ask for the breakpoint positions, as then the limited result will be used for all future queries.  I think the simplest thing for now is to remove the range arguments from these functions, but there is some cleanup / efficiency improvements that could be made to how these functions are used by commands.js and source-actors.js, either now or later.

hbenl:
That's why I only cache the promise if no `range` argument was supplied. The only potential issue would be that if we ask for breakpoint positions with a `range` and had requested them without a `range` before, we'll get the full set of breakpoint positions - but that should be OK (especially since the backend also sometimes ignores the `range`, so the callers of this function shouldn't assume anyway that the `range` is respected).

bhackett1024 (approved):
Oops, I missed that, thanks!

PR #2796, opened 2021-06-08
Handle invalid/unauth recordings with error modal

### LINKED ISSUE #2750: Replay stalls when I don't have access
Go here and it'll stall out rather than saying you don't have access:
https://replay.io/view?id=938323f7-b056-44fb-8431-365896437633&point=28233114176857742571739553770506004&time=12178&hasFrames=false

ryanjduffy:
Handles `BadRecordingId` errors with the error modal.

Fixes #2750


jaril:
Continued in #2805.

Issue #2811, opened 2021-06-10
Multiple identical sources


hbenl:
A recording can contain multiple identical sources for the same URL (usually because the user navigated while recording) and that leads to some issues:
- Multiple sources are shown #2570
- The breakpoint panel in the editor disappears #2600
- Pending breakpoints show up multiple times #2810

Issue #2810, opened 2021-06-10
Pending breakpoints show up multiple times


hbenl:
If a recording with multiple identical sources for one URL is opened and a pending breakpoint was set for that URL, the breakpoint shows up multiple times (once for each identical source) in the side panel.

hbenl:
Fixed in #2812

Issue #2600, opened 2021-05-13
The breakpoint panel in the editor disappears


hbenl:
I just navigated to a breakpoint hit and then the breakpoint panel disappeared...
https://replay.io/view?id=8586ea6f-5b0c-4b52-a3f6-dca00b4dc40a

jasonLaster:
Left some comments - the widget is definitely cleared

hbenl:
I think this happens when the source has been loaded multiple times, in this case there are multiple `sourceId`s of preferred sources with the same URL. The editor identifies sources by URL, so it will always show only one of them, but the breakpoint panel identifies sources by `sourceId`, so it will only be shown in the source where the breakpoint was initially set.

hbenl:
Fixed in #2812

Issue #2570, opened 2021-05-11
Multiple sources are shown


jasonLaster:
**Description:**

Multiple versions of the same file appear

Example 1:

**Steps to Reproduce:**
1. go to https://replay.io/view?id=905288c2-f610-471a-98c2-9c544104a601
2. go to DevTools 
3. press cmd+p

**Expected**
1 version of `application-9e586eab0f44cc309d13` 

**Actual**
4 versions of `application-9e586eab0f44cc309d13`

Example 2:

**Steps to reproduce**
1. go to https://replay.io/view?id=c44db129-50b8-4735-8331-cf4b4cf57343
2. go to DevTools
3. press cmd+p

**Expected**
2 version of `SectionContainer` 

**Actual**
2 versions of `SectionContainer`

---

Replay: https://replay.io/view?id=c2a09d0f-a801-4cdf-900e-696b62a84a27&point=103845937239313203955335534296958800&time=63409&hasFrames=false

hbenl:
Fixed in #2812

PR #2812, opened 2021-06-10
Fix issues with multiple identical sources (#2811)

hbenl:


src/main.js
```diff
@@ -9,7 +9,10 @@ const test = url.searchParams.get("test");
 
 // During testing, make sure we clear local storage before importing
 // any code that might instantiate preferences from local storage.
-if (test) {
+// If the url contains a "navigated" parameter, we assume this is the
+// second part of a test that contains a navigation and don't clear
+// local storage in that case.
+if (test && !url.searchParams.get("navigated")) {
```
loganfsmyth, line 15: Would it be feasible to have the test harness script itself do this before each test? It seems like that would make things simpler.

loganfsmyth (commented):
Since I'm not super familiar with the state of things, this is a bit tough to review. Could you describe what overall architecture you're aiming for with this patch? 

Say a recording has 4 copies of a source, and they all have the same URL, and the first 2 have one set of contents, and the second 2 have another set of contents, what is the expected output? Is that something you're trying to address, or is this still assuming that all sources with the same URL are identical, for instance? Is the goal the present each source as a separate entry in the source-tree, or are we still presenting a single entry in the source tree to represent many sources?

Separately, are there changes to the protocol that we could make that would make any of this easier? It seems like all of the code around this is quite complex.

hbenl:
Due to sourcemapping we can't assume that all sources with the same URL are identical, of course. I'm relying entirely on the existing method `ThreadFront._chooseSourceIdList()` to group sources that "belong together" (i.e. they are original/generated/pretty-printed variants of the same "real" source from a single source load event) and pick a preferred and optionally an alternate source from each group.
I assume that the preferred (or alternate) sources for the same URL have the same content (although I don't verify this anywhere), so I pick the first preferred/alternate source for each URL and try to use that in the source tree and the editor (both of them never had a way to properly deal with multiple sources for the same URL anyway).
When running an analysis, I add the locations for all corresponding sources (the corresponding sources of a preferred/alternate source are the preferred/alternate sources with the same URL) to get all hits.
When adding breakpoints from a previous session, I only add them for one of the corresponding sources (but when the analysis is run, the other corresponding sources will be added to the locations).
In the QuickOpenModal I only add one source per URL.
See also #2570, #2600 and #2810.

src/main.js
hbenl: Yeah, that would be a lot cleaner. We should refactor how we're initialising the frontend anyway, so I'll have another look at this then.

loganfsmyth (approved):
Alright, gonna approve, though I admit I'm not super confident about it because I don't feel like I have a good sense of the frontend codebase at this point.

PR #2817, opened 2021-06-11
Add docs to the menu

jonbell-lot23:
Now it looks like this:

image

### DIFF
--- a/src/ui/components/Header/UserOptions.js
+++ b/src/ui/components/Header/UserOptions.js
@@ -36,6 +36,14 @@ function UserOptions({ recordingId, setModal }) {
     }
     window.location = dashboardUrl;
   };
+  const onDocsClick = () => {
+    const docsUrl = `https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02`;
+
+    if (event.metaKey) {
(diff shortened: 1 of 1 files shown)


src/ui/components/Header/UserOptions.js
```diff
@@ -36,6 +36,14 @@ function UserOptions({ recordingId, setModal }) {
     }
     window.location = dashboardUrl;
   };
+  const onDocsClick = () => {
+    const docsUrl = `https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02`;
+
+    if (event.metaKey) {
+      return window.open(docsUrl, "_blank");
+    }
+    window.open("https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02", "_blank");
```
jasonLaster, line 45: why not re-use the same variable?

Also, shouldn't this one use something other than blank?

jonbell-lot23: Apologies! This is an old branch. See here: https://github.com/RecordReplay/devtools/pull/2818/files

jonbell-lot23:
See here instead: https://github.com/RecordReplay/devtools/pull/2818/files

PR #2818, opened 2021-06-11
Adding documentation link to the menu

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/2813

### DIFF
--- a/src/ui/components/Header/UserOptions.js
+++ b/src/ui/components/Header/UserOptions.js
@@ -36,6 +36,14 @@ function UserOptions({ recordingId, setModal }) {
     }
     window.location = dashboardUrl;
   };
+  const onDocsClick = () => {
+    const docsUrl = `https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02`;
+
+    if (event.metaKey) {
+      return window.open(docsUrl, "replaydocs");
+    }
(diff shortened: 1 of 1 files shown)


src/ui/components/Header/UserOptions.js
```diff
@@ -36,6 +36,14 @@ function UserOptions({ recordingId, setModal }) {
     }
     window.location = dashboardUrl;
   };
+  const onDocsClick = () => {
+    const docsUrl = `https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02`;
+
+    if (event.metaKey) {
+      return window.open(docsUrl, "replaydocs");
+    }
+    window.open("https://www.notion.so/Docs-56758667f53a4d51b7c6fc7a641adb02", "replaydocs");
```
jasonLaster, line 45: isn't this the same: `window.open(docsUrl, "replaydocs");`

jasonLaster: what does `"replaydocs"` do? Do you need "_blank"

jonbell-lot23: It's totally the same, I'm not sure why I did that. Will change.

_blank causes trouble in this scenario because it means each time you click you're going to load a new tab. By naming it, it means each time you click docs it re-uses the tab. _blank is better for general web, named windows are better for software UI like this.

Issue #2707, opened 2021-05-24
Comments are appearing twice


jonbell-lot23:
Here's a Replay of it happening:
https://replay.io/view?id=d6f3ab9a-e239-4327-af9f-d3c3c4507a62&point=63281117965750637592966732904924625&time=24230&hasFrames=false

Optimistic comments seem to have a bug

jaril:
Yes, this is related to optimistic comments. Specifically, it occurs when you have both optimistic comments and comment polling (every X seconds) in place. 

Based on my limited digging, it looks to be some problem with Apollo not being aware that the optimistic comment and the new comment appearing after a poll is the same. Will need to debug it more.

jonbell-lot23:
Thanks for the info! From a UX standpoint, here are the best and second best approaches:

1. Have it appear immediately and never double up
2. Throw a spinner until we're 100% sure it's posted correctly

ryanjduffy:
@jaril little more data for you ... I added a `setTimeout` to defer landing the committed data which helped repro the duplicate comment from polling.

https://replay.io/view?id=304ea995-fcf9-49dd-8fab-135c7ec432fa&point=48028745945354581787087575348937557&time=22175.681922547068&hasFrames=true

Will be curious if apollo has any feedback but I expect that we'll need to send an extra data element to the backend to facilitate merging on polling. I don't see how else the front end (either our code or apollo) can know to merge the optimistic data with the real data. for example, add a method for generating semi-unique id and set that as the `id` for the optimistic data and the `clientId` for both the optimistic and data sent to the server so that you can reliably merge on `clientId`.

The other option would be to pause polling (or at least enqueue polled responses) while waiting on an optimistic response. We could do some version of this in our code but likely could be handled more elegantly in apollo code since it could index polling queries on the `__typename` and manage them internally.

PR #2819, opened 2021-06-11
Fix comments appearing twice (#2707)

hbenl:
The problem and its fix are different for comments and replies: in both cases, the `update` function is called twice (first with the `optimisticResponse` containing a temporary ID and the second time with the real response containing the final ID).
- for comments the issue is that in the second call, the list of comments may already contain the comment with the final ID if the `GET_COMMENTS` query was polled in the meantime (after the comment was added on the server but before the second call to `update`). Note that we never get the comment with the temporary ID back from the cache in the second call.
- for replies the issue is that in the second call, the list of replies contains the reply with the temporary ID that we added in the first call

### DIFF
--- a/src/ui/hooks/comments/useAddComment.tsx
+++ b/src/ui/hooks/comments/useAddComment.tsx
@@ -31,13 +31,14 @@ export default function useAddComment() {
   }
 
   return (comment: NewCommentVariable, recordingId: RecordingId) => {
+    const temporaryId = new Date().toISOString();
     addComment({
       variables: { input: comment },
       optimisticResponse: {
         addComment: {
           success: true,
           comment: {
-            id: new Date().toISOString(),
+            id: temporaryId,
             __typename: "Comment",
           },
           __typename: "AddComment",
@@ -76,7 +77,12 @@ export default function useAddComment() {
           ...data,
           recording: {
             ...data.recording,
-            comments: [...data.recording.comments, newComment],
+            comments: [
+              ...data.recording.comments.filter(
+                (c: any) => c.id !== temporaryId && c.id !== commentId
+              ),
+              newComment,
+            ],
           },
         };
 
--- a/src/ui/hooks/comments/useAddCommentReply.tsx
+++ b/src/ui/hooks/comments/useAddCommentReply.tsx
@@ -30,13 +30,14 @@ export default function useAddCommentReply() {
   }
 
   return (reply: NewReplyVariable, recordingId: RecordingId) => {
+    const temporaryId = new Date().toISOString();
     addCommentReply({
       variables: { input: reply },
       optimisticResponse: {
         addCommentReply: {
           success: true,
           commentReply: {
-            id: new Date().toISOString(),
+            id: temporaryId,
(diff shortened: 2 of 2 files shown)


src/ui/hooks/comments/useAddComment.tsx
```diff
@@ -59,6 +59,12 @@ export default function useAddComment() {
           query: GET_USER_ID,
         });
 
+        // check if the comment has already been added to the cache
+        // due to comments getting polled in parallel
+        if (data.recording.comments.some((c: any) => c.id === commentId)) {
```
jaril, line 64: My understanding was that this would still lead to duplicated comments. Order of events being:

1) Optimistic update data for `newComment1` is added to the cached query.
2) UI updates with the optimistic data.
3) Real server response data for `newComment1` comes back.
4) The optimistic update data is still in the cached query (due to polling being in flight).
5) This check compares the IDs for the optimistic and real `newComment1`, but the optimistic data has a different ID (`new Date().toISOString()`) compared to the real data (UUID generated by the server).
6) The condition doesn't pass and the function doesn't return early.
7) The real server response data is added to the cached query.
8) UI updates ends up with a duplicate comment, with both the optimistic and real data.

Am I missing something?

src/ui/hooks/comments/useAddCommentReply.tsx
```diff
@@ -77,7 +78,7 @@ export default function useAddCommentReply() {
 
         const newParentComment = {
           ...parentComment,
-          replies: [...parentComment.replies, newReply],
+          replies: [...parentComment.replies.filter((r: any) => r.id !== temporaryId), newReply],
```
jaril, line 81: Same problem as the reply implementation above. The ID generated by the optimistic update data and the real server data are different, so this won't filter the optimistic data out.

src/ui/hooks/comments/useAddComment.tsx
hbenl: > My understanding was that this would still lead to duplicated comments.

That's what I would have expected as well, but when I reproduced the issue I saw that Apollo's cache works differently than I expected:

> The optimistic update data is still in the cached query

Surprisingly it isn't when I reproduce this. I don't know why.

src/ui/hooks/comments/useAddCommentReply.tsx
hbenl: Same as above: this works in my experiments because Apollo doesn't work the way one would expect it to.

hbenl:
I have changed the logic to be more straightforward and robust: now the `update` function will always remove any comments with the `temporaryId` or the ID from its current `data` parameter (which may or may not be the same as the `temporaryId`). Then it adds a comment using the ID from its current `data` parameter. After that, the cache should always contain the new comment exactly once.
To reproduce the issue, I used the following trick: open the browser's devtools and set a breakpoint on the first breakable line of the `update` function. Every time that breakpoint is hit, wait a couple of seconds before resuming. That reproduces the issue for me most of the time (though not all the time).

Issue #2624, opened 2021-05-15
Upload screen should never time out


jonbell-lot23:
<img width="583" alt="image" src="https://user-images.githubusercontent.com/9154902/118343757-dc368780-b57e-11eb-8052-8e5ec6703ce7.png">

This is happening but we shouldn't end up in this state

jasonLaster:
(i thought this was done, but now i'm not sure)

ryanjduffy:
Would this be avoided if the session wasn't created until after the user saved?

jasonLaster:
@ryanjduffy yep

jonbell-lot23:
@ryanjduffy Which is a general issue overall: See https://github.com/RecordReplay/devtools/issues/2737 where for a while things could go live on the public internet before you even decided to save it, let alone make it public! 😱

We addressed one component of that, but we have some more holes in this vein. Do veins have holes? You get the idea.

jaril:
@hbenl Any thoughts on this? Right now, we initialize the session whenever we're looking at a Replay. Ideally, we should wait to see if the replay is not initialized yet (i.e. just finished recording, needs to show the upload screen). If it's initialized, create the session. If it's not initialized, don't create a session immediately but do it immediately after the user hits the save button.

The deferred session creation goes against the grain with how you implemented the `PageSwitch` component, so wanted to run it by you first!

hbenl:
@jaril I've created a draft PR to show how this could be done in principle: #2624
Eventually we need to overhaul how the app is initialized, currently it is always initialized for a particular purpose (library vs. devtools vs. UploadScreen) and always needs to be reloaded to switch.

PR #2830, opened 2021-06-14
Don't create a session if the recording is uninitialized (#2624)

hbenl:
This PR still has some issues:
- after clicking "Save & Upload" on the UploadScreen, the app needs to be reloaded so that it is reinitialized (currently it gets stuck on a BlankLoadingScreen)
- there are errors in the browser console because some code that tries to access the session is still running

### DIFF
--- a/src/upload.tsx
+++ b/src/upload.tsx
@@ -1,0 +1,57 @@
+import React, { useEffect } from "react";
+import { combineReducers, applyMiddleware } from "redux";
+import { connect, ConnectedProps } from "react-redux";
+const LogRocket = require("ui/utils/logrocket").default;
+import { isDevelopment, skipTelemetry } from "ui/utils/environment";
+import { sanityCheckMiddleware } from "ui/utils/sanitize";
+const configureStore = require("devtools/client/debugger/src/actions/utils/create-store").default;
+import reducer from "ui/reducers/app";
+import { setExpectedError } from "ui/actions/session";
+import { useGetRecording, useHasExpectedError } from "ui/hooks/recordings";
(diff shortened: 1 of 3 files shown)


hbenl:
I've changed the approach of this PR slightly: detecting an uninitialized recording now happens even earlier (in `src/main.js`) and I've created a separate initialization script `src/upload.tsx` for the `UploadScreen`.

src/upload.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  useEffect(() => {
+    if (recording?.isInitialized) {
+      window.onbeforeunload = null;
+      document.location.reload();
+    }
+  });
+
+  if (expectedError) {
+    return null;
```
jaril, line 35: Small nit - let's render <BlankLoadingScreen /> here. Otherwise, it goes `white screen → blank loading screen → white screen (null) → error`. Instead, it should be `white screen → blank loading screen → error`.

src/upload.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    if (expectedError) {
+      setExpectedError(expectedError);
+    }
+  }, [expectedError]);
+
+  useEffect(() => {
+    if (recording?.isInitialized) {
+      window.onbeforeunload = null;
+      document.location.reload();
```
jaril, line 30: It would be nice if we didn't have to refresh, but I can see how that's tricky to pull off with the current pageswitch implementation. We can just do it as a follow-up, I guess!

jaril (approved):

src/upload.tsx
hbenl: Yeah, I'll need to refactor how the app is initialized for that.

Issue #2335, opened 2021-04-13
Our download link should be dynamic


jonbell-lot23:
1. If you're using the Replay browser already, the link should go away
2. The link should say "Launch Replay" and launch the app if it's installed, otherwise take you to the install page

We should put a conditional in the Library dropdown so it's not in the library but still exists in the /view/ part of the product.

PR #2834, opened 2021-06-14
Add launch replay button and page

### LINKED ISSUE #2688: The library is missing a "Launch Replay" button
Now that we can determine whether or not you're in the browser, we should do this:

In Replay:
(Do nothing, since you're already in Replay, right?)

Not in Replay:
Blue button that says "Launch Replay"
Clicking it launches Replay (do we have the technical ability to do this?)

ryanjduffy:
Adding this behind `features.launchBrowser`

* Adds a "Launch Replay" button to the library which will open a modal and launch `replay:library`to trigger opening the browser

image

image

* Adds `/browser/launch` endpoint which presents a similar dialog over the blank screen so we can direct link to a launcher.

image

Fixes #2688

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -40,6 +40,7 @@ document.body.addEventListener("contextmenu", e => e.preventDefault());
 setupTelemetry({ recordingId });
 
 const BrowserError = React.lazy(() => import("views/browser/error"));
+const BrowserLaunch = React.lazy(() => import("views/browser/launch"));
 
 function PageSwitch() {
   const [pageWithStore, setPageWithStore] = useState(null);
@@ -72,6 +73,7 @@ ReactDOM.render(
     <Router>
       <Switch>
         <Route exact path="/browser/error" component={BrowserError} />
+        <Route exact path="/browser/launch" component={BrowserLaunch} />
         <Route>
           <tokenManager.Auth0Provider>
             <ApolloWrapper recordingId={recordingId}>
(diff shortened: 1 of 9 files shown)


ryanjduffy:
@jonbell-lot23 - This conflicts a bit with #2335. Do we want to keep this button or should it fold into the menu per the other issue?

jaril:
I can't get this to work. I expected a native browser pop-up that says Open Replay or something, and instead nothing happens.

Also the download link currently links to `localhost:8080`.

https://user-images.githubusercontent.com/15959269/121949711-c73c4800-cd26-11eb-9161-72e1720bda7d.mov

ryanjduffy:
@jaril Well that's odd! I just uninstalled replay and observed the following (on Mac Catalina fwiw):

1. Clicking "launch replay" without replay installed shows the modal but no native browser popup. I think this is okay since nothing else could handle the request anyway.
2. Dowload and install replay
3. Refresh the page and click "launch replay". Modal is shown and native popup to launch replay opens.
4. Click "open Replay" and replay is launched with `replay.io/view` loaded

jonbell-lot23:
>@jonbell-lot23 - This conflicts a bit with #2335. Do we want to keep this button or should it fold into the menu per the other issue?

Good catch! We should put a conditional in the Library dropdown so it's not in the library but still exists in the /view/ part of the product.

ryanjduffy:
Cool. I'll note that in #2335.

jaril:
> 1. Clicking "launch replay" without replay installed shows the modal but no native browser popup. I think this is okay since nothing else could handle the request anyway.

Yeah, strange. It wasn't opening the popup before. I updated my Replay browser and it works as expected now. Thanks! 👍

Hopefully it's just because I haven't updated the browser in a ridiculously long time. Although if there were problems with the link working for out-of-date browser versions, that might be a problem as well.

hbenl (approved):

PR #2859, opened 2021-06-15
Use the defaultworkspaceid as the library landing view

jaril:


jaril:
#2861 does this, but simpler.

Issue #2809, opened 2021-06-09
The indexing loader should use the executionIndexed


jasonLaster:
The first is when we are done with "executionIndexed". The other is when we have finished loading the regions. 

https://replay.io/view?id=57062ea4-45f7-47d4-97ba-8a4fceef7903&point=4218741198518267395787319693803933&time=2540.5173140620295&hasFrames=true

jasonLaster:
I started looking at this last week and noticed we are not tracking our execution indexing progress 

```diff
diff --git a/src/ui/actions/app.ts b/src/ui/actions/app.ts
index d46a8a1f..186a7ea2 100644
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -121,8 +121,9 @@ export function setupApp(recordingId: RecordingId, store: UIStore) {
     store.dispatch({ type: "loading", loading: 100 });
   });

-  ThreadFront.ensureProcessed("executionIndexed").then(() => {
-    console.log("indexed");
+  ThreadFront.ensureProcessed("executionIndexed", undefined, progress => {
+    console.log(progress);
+  }).then(() => {
     store.dispatch({ type: "indexed" });
   });
```

PR #2883, opened 2021-06-17
Improve loading/indexing progress indicator accuracy (#2809)

hbenl:
- we use both `loadedRegions` and `unprocessedRegions` events to calculate the recording's duration - in my experiments, the `unprocessedRegions` events contained the correct duration sooner than the `loadedRegions` events
- in the `onUnprocessedRegions()` callback we distinguish between the two processing levels
- the `IndexingLoader` component now uses the progress calculated from the `unprocessedRegions` events for the `executionIndexed` level

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -26,8 +26,9 @@ import { RecordingTarget } from "protocol/thread/thread";
 import { Workspace } from "ui/types";
 
 export type SetupAppAction = Action<"setup_app"> & { recordingId: RecordingId };
+export type SetRecordingDurationAction = Action<"set_recording_duration"> & { duration: number };
 export type LoadingAction = Action<"loading"> & { loading: number };
-export type IndexedAction = Action<"indexed">;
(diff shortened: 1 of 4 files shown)


src/ui/actions/app.ts
```diff
@@ -118,38 +120,59 @@ export function setupApp(recordingId: RecordingId, store: UIStore) {
...
+function onUnprocessedRegions({ level, regions }: unprocessedRegions): UIThunkAction {
   return ({ dispatch, getState }) => {
-    const loading = selectors.getLoading(getState());
-    const endPoint = Math.max(...regions.map(r => r.end), 0);
+    let endPoint = Math.max(...regions.map(r => r.end), 0);
     if (endPoint == 0) {
       return;
     }
+    const state = getState();
+    if (endPoint > selectors.getRecordingDuration(state)) {
```
jasonLaster, line 143: Another option would be to store the loaded regions in redux which we'll want to use in the future to show what is accessible in the timeline and console.

I'll go ahead and land this PR and then we can add regions as a follow up

PR #2886, opened 2021-06-18
Remove duration

jasonLaster:
I think this simplifies the logic a bit without introducing any regressions:

Here's a replay with the change
https://replay.io/view?id=006fbe06-b0bb-4f51-92ae-65c53eecc3e2

Here's a replay where i saw the value of the `endPoint == 0` check :)
https://replay.io/view?id=6a30b9d8-2e8f-48e7-a4b9-a5ff0c30775a&point=27259558510959546365715416566727139&time=38402.68472280999&hasFrames=true

worst case scenario, i think these callbacks could be out of order, in which case it's possible the endPoint might be slightly lower in one case, but I don't think that should have a significant impact.

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -246,7 +246,6 @@ class _ThreadFront {
     const sessionId = await this.waitForSession();
 
     client.Session.addLoadedRegionsListener((parameters: loadedRegions) => {
-      console.log("LoadedRegions", parameters);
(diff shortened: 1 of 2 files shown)


jasonLaster:
This doesn't really matter, just messing around a bit.

jonbell-lot23:
Curious to know if we can safely squash and merge, @jasonLaster ?

PR #2920, opened 2021-06-22
Improve the breakpoints display while we're indexing

### LINKED ISSUE #2703: The breakpoint hits tooltip should show we're indexing
**Description:**

At the moment, the breakpoint hits tooltip shows `...` when we're indexing the replay which can make it seem like fetching the number of hits can be slow when in fact we're just waiting for the recording to finish processing.

Tasks
- [ ] Update the tooltip to say "Indexing"
- [ ] Show the panel when we're registering breakpoints
- [ ] Replace the light blue breakpoint color with the normal blue

hbenl:
Fixes #2703

### DIFF
--- a/src/devtools/client/debugger/src/actions/sources/newSources.js
+++ b/src/devtools/client/debugger/src/actions/sources/newSources.js
@@ -86,6 +86,11 @@ function checkPendingBreakpoints(cx, sourceId) {
       return;
     }
 
+    for (const bp of pendingBreakpoints) {
+      const line = bp.location.line;
+      dispatch({ type: "SET_REQUESTED_BREAKPOINT", location: { sourceId, line } });
+    }
+
     // load the source text if there is a pending breakpoint for it
     await dispatch(loadSourceText({ cx, source }));
 
--- a/src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
+++ b/src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
@@ -110,6 +110,8 @@ export function getColumnBreakpoints(breakpoints, requestedBreakpoints, viewport
   // - there is atleast one other breakpoint on that line
   // - there is a breakpoint on that line
   const allBreakpoints = [...breakpoints, ...requestedBreakpoints];
+  // collect all breakpoint positions but make sure that we don't have 2 positions
+  // for the same line in the same source
   let positions = uniqBy(
     allBreakpoints.map(bp => bp.location),
     ({ sourceId, line }) => `${sourceId}:${line}`
--- a/src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
+++ b/src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
@@ -4,6 +4,7 @@
 //
 
 import groupBy from "lodash/groupBy";
+import uniqBy from "lodash/uniqBy";
 import { createSelector } from "reselect";
 
 import {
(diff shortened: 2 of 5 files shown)


src/devtools/client/debugger/src/actions/sources/newSources.js
```diff
@@ -86,6 +86,11 @@ function checkPendingBreakpoints(cx, sourceId) {
       return;
     }
 
+    for (const bp of pendingBreakpoints) {
+      const line = bp.location.line;
+      dispatch({ type: "SET_REQUESTED_BREAKPOINT", location: { sourceId, line } });
```
jasonLaster, line 91: why not include the column as well?

src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
```diff
@@ -108,7 +109,12 @@ export function getColumnBreakpoints(positions, breakpoints, viewport, selectedS
   // - the position is in the current viewport
   // - there is atleast one other breakpoint on that line
   // - there is a breakpoint on that line
-  const breakpointMap = groupBreakpoints(breakpoints, selectedSource);
+  const allBreakpoints = [...breakpoints, ...requestedBreakpoints];
+  let positions = uniqBy(
+    allBreakpoints.map(bp => bp.location),
+    ({ sourceId, line }) => `${sourceId}:${line}`
+  );
```
jasonLaster, line 116: should we update the comment above to describe what is happening here?

src/devtools/client/debugger/src/actions/sources/newSources.js
hbenl: I basically copied this from the only other place where a `SET_REQUESTED_BREAKPOINT` action is dispatched (in `modify.js`)

src/devtools/client/debugger/src/selectors/visibleColumnBreakpoints.js
hbenl: What about:
```
// collect all breakpoint positions but make sure that we don't have 2 positions for the same line in the same source
```

jasonLaster: works for me

src/devtools/client/debugger/src/actions/sources/newSources.js
jasonLaster: weird - feels like an omission, but tough to say...

hbenl: It does, but I found this is actually enforced [here](https://github.com/RecordReplay/devtools/blob/b4a2ce306178b3f1eb5b4d420b9dc18bfdf6abcb/src/devtools/client/debugger/src/reducers/breakpoints.js#L77)

hbenl:
I had a hard time creating recordings that show the effect of this PR - it depends very much on the timings between backend and frontend and the frontend performance is quite different when it's being recorded. So the following recordings may not differ much visually, but the comments show what's going on under the hood:
Without this PR: https://replay.io/view?id=243df849-ac85-4216-9663-65fcc62d9317
With this PR: https://replay.io/view?id=9628363e-be9b-4a26-8fbb-61cfd0ad29ac

src/devtools/client/debugger/src/actions/sources/newSources.js
jasonLaster: weird...

jasonLaster:
Looks good. Feel free to land when the tests are green

jasonLaster (approved):

hbenl:
All test are green locally

PR #2921, opened 2021-06-22
Fix video playback when switching modes

ryanjduffy:
## Issue

The video was being set up within graphics which cached the node. When switching between viewer and devtools, the node would be dropped and remounted breaking the cached node reference.

## Resolution

* Adds `videoNode` and `videoUrl` to redux so they can be updated across views
* Updates the `Video` class to pull the latest node from the store when executing a command

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -128,29 +128,22 @@ function onMouseEvents(events: MouseEvent[], store: UIStore) {
 }
 
 class VideoPlayer {
+  store: UIStore | null = null;
   video: HTMLVideoElement | null = null;
   all = new Uint8Array();
   blob?: Blob;
   videoReadyCallback?: (video: HTMLVideoElement) => any;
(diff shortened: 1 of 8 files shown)


kannanvijayan:
I tested this locally and it's basically working.  The only remaining issue seems to be if the user presses the video play button _before_ the full blob has been received and set as the video source.  If you hit play after that, things are fine.

I suspect that we can, at this point, try using media-stream APIs again to incrementally feed the data to the video element.

jasonLaster:
It's probably simpler to just disable playing the video until everything has loaded

jasonLaster (approved):

ryanjduffy:
Pushed an update that should retain video playback when swapping between views.

ryanjduffy:
No clue why TS is failing on this PR. I didn't touch that file, I can't repro it locally, and I don't see where `userSettings` is required as a prop for `DevTools`. 🤷‍♂️

Issue #2378, opened 2021-04-16
Reveal in Tree should work when the sources panel is not open


jasonLaster:
**Description:**
1. go to the editor
2. select transcript
3. right click select "reveal in tree"

ER: sources explorer is opened and file is revealed

jasonLaster:
https://replay.io/view?id=404641c6-6ad1-434b-823b-1158360d6d73&point=11358149379801870212437795548432456&time=5088.115462983926&hasFrames=true

PR #2943, opened 2021-06-24
Fix "Reveal in tree" when the sources tree is not open (#2378)

hbenl:


hbenl:
Recordings with comments explaining this PR:
Without this PR: https://replay.io/view?id=8b5e8a72-6a67-47f9-b496-6fa45950bfb3
With this PR: https://replay.io/view?id=1fcfcc39-4537-48ae-aeff-29e4c179459f

jaril (approved):
LGTM!

PR #2953, opened 2021-06-25
Ask the user to accept our Terms of Service

hbenl:
...before allowing him to initialize a recording


jasonLaster:
I made a couple of tweaks to the setup logic. if it looks good, feel free to merge tomorrow

hbenl:
LGTM, I just made the check in the setup logic more explicit to make it more obvious what happens if `userInfo` is `undefined` or `userInfo.acceptedTOSVersion` is `null`.
All checks passed locally.

PR #3006, opened 2021-07-01
Change the build for Vercel

hbenl:


jasonLaster:
### Background

We've been seeing several issues over the past month where webpack assets have not been loading. Earlier this week we saw that some of the issues were with Vercel fetching assets from cloudfront.

### Solution

This PR changes how we deploy the frontend so that instead of deploying to S3 we deploy to Vercel. This will simplify the process and avoid the additional hops that were likely causing the caching issues.


### Changes

* Vercel will host our frontend assets
* The frontend will be available on `app.replay.io` instead of `replay.io/view`

### Follow ups

* Sentry + LogRocket might need to learn about the new subdomain
* Vercel will automatically deploy merges. We will need to use the vercel cli to only deploy merges that pass CI
* We will be able to add the contenthash for cachebusting
* We will teach auth0 to support preview branches

Issue #2985, opened 2021-06-29
Browser mock tests


bhackett1024:
We pretty regularly have regressions where functionality in the devtools stops working.  https://github.com/RecordReplay/devtools/issues/2898 is a recent example.

Before launch we're going to have a QA department in place so that we can watch for these regressions more easily, but in addition to that we should have more automated tests so that we don't *need* QA to tell us if something which we have a test for has broken.

Our current test suite is able to test for a variety of issues that show up when debugging recordings, but it is not enough.  Our test suite is only able to deal with cases where we successfully created a recording in one specific way, and then testing the debugger's behavior after it is loaded.  There are many other aspects of the devtools we aren't able to test at all this way, such as connection problems like the one above and features like comments.

I think the best way to test these aspects of the devtools is with a set of unit tests for the browser.  We could have unit tests for individual components, which would be nice, but if we have unit tests for the browser itself where we mock the communication with the servers (either through the dispatch connection or through graphQL) then it should be a lot easier to test that the viewer behaves as expected in these different use cases, because we can inject any behavior we want into the interfaces we are mocking.

This runs the risk that we're increasing the development burden by having to maintain the mocked interfaces as things in the devtools shift around.  I think this can be mostly avoided with good choices for the interfaces we mock, though some friction will be unavoidable.

bhackett1024:
I'm going to start on getting a test harness in place for these sorts of tests, with an initial target test that if we are loading a recording and the backend sends back a sessionError message, we show a "session died" modal to the user.

bhackett1024:
I think calling these "mock tests" rather than "unit tests" would be a more appropriate use of this terminology.

bhackett1024:
We have a small mock test suite now which runs as part of CI on PRs and pushes to main.  We'll continue build out these tests over time but the infrastructure is in place.

PR #3008, opened 2021-07-01
Add mock test suite and a couple of simple tests, for #2985

bhackett1024:
This adds a test harness and a couple of tests for the start of a new mock test suite (see https://github.com/RecordReplay/devtools/issues/2985 for the rationale here).  When running a mock test, we set the "mock" URL param which causes the devtools to look for a "mockEnvironment" property on the window for use when performing database accesses and protocol commands.  The mock test installs this environment on the window and can customize its behavior as needed for fine grained control over how the devtools interact with the database and dispatch server.

The test itself is written using playwright, and uses our snazzy new playwright fork (https://www.npmjs.com/package/@recordreplay/playwright) so that we can record the tests with minimal configuration.

Once this lands, the remaining steps are to add these tests to CI, and to build out the test suite to improve its coverage.  We don't have to go crazy with test coverage, the goals are to have basic tests for most of the devtools' functionality, and to have a way to add new tests when things regress.

My main question here is about the right level of abstraction for mocking database accesses.  I went with the wrappers in src/ui/hooks because this is less brittle / crufty than trying to mock the graphQL operations themselves.  It does look like Apollo has a MockedProvider component (https://www.apollographql.com/docs/react/api/react/testing/) which we could use instead, though I think it would be harder to build things out and maintain them that way.


src/ui/hooks/recordings.ts
```diff
@@ -174,6 +183,11 @@ export async function getRecordingOwnerUserId(
 export function useGetRecording(
   recordingId: RecordingId | null
 ): { recording: Recording | undefined; isAuthorized: boolean; loading: boolean } {
+  if (isMock()) {
+    const env = mockEnvironment();
+    return env ? env.database.useGetRecording(recordingId) : { loading: true };
```
jaril, line 188: Was this meant to return `{loading: true}` temporarily if the mock environment isn't ready, and then return the actual mock recording afterwards? 

I'm only about 80% sure, but this looks like it will only ever return `{ loading: true }` if it happens to be called while the environment isn't ready. It won't update its returned value once the environment is ready.

hbenl: > It won't update its returned value once the environment is ready.

To be more precise: it doesn't need to update the returned value, but it needs to tell React that the value has changed so that all components using this hook are rerendered. You need something like this:
```typescript
  if (isMock()) {
    const [result, setResult] = useState<any>({
      recording: undefined,
      isAuthorized: false,
      loading: true,
    });
    useEffect(() => {
      async function waitForRecording() {
        const env = await waitForMockEnvironment();
        setResult(env.database.useGetRecording(recordingId));
      }
      waitForRecording();
    });
    return result;
  }
```
(React hooks are the devil in disguise)

hbenl:
I think the `MockedProvider` could be a good solution, mainly because it would remove the need to write mocks for React hooks (which is far from obvious - see above).
Furthermore, I'd like to have the mock request/response pairs to be part of the tests themselves instead of having all mocks for all tests placed in `mockEnvironment.ts` and selecting them based on the `options` argument. Those that are shared between tests could still be put in a shared file, but the code for selecting the desired mocks for a particular test would be part of that test.
I'll create a version of this PR that uses `MockedProvider`, then we can compare the two approaches.

src/ui/hooks/recordings.ts
jaril: Yeah, that looks good!

In general, adding hooks (useState/useEffect) inside an if statement is not ideal since React will throw an error if the order that the hooks are called in changes. But in this case, isMock() should always return the same value, and the order will always be the same, so it should be okay. The proper way to do it is a fair bit messier.

Last nit, let's pass in an empty array in the useEffect so that it only gets called once:

```js
    useEffect(() => {
      async function waitForRecording() {
        const env = await waitForMockEnvironment();
        setResult(env.database.useGetRecording(recordingId));
      }
      waitForRecording();
    }, []);
```

hbenl:
I've added a commit that uses Apollo's `MockedProvider` to mock all GraphQL requests.
- this should allow us to also test code that interacts with the Apollo cache
- I have used typescript in some files, so running a test now looks like `./node_modules/.bin/ts-node -r tsconfig-paths/register ./test/mock/scripts/session-error.js`.
- I have added a line to `runTest.js` that outputs the messages from the browser's console (useful for debugging if a test doesn't work).

bhackett1024:
> I've added a commit that uses Apollo's `MockedProvider` to mock all GraphQL requests.
> 
>     * this should allow us to also test code that interacts with the Apollo cache
> 
>     * I have used typescript in some files, so running a test now looks like `./node_modules/.bin/ts-node -r tsconfig-paths/register ./test/mock/scripts/session-error.js`.
> 
>     * I have added a line to `runTest.js` that outputs the messages from the browser's console (useful for debugging if a test doesn't work).

Awesome, this approach to mocking the graphQL accesses is much better.

jaril:
Looks good.

Is it supposed to be working locally? Ran `./node_modules/.bin/ts-node -r tsconfig-paths/register ./test/mock/scripts/session-error.js` but getting the error below:

```
Puter:devtools jarilvalenciano$ ./node_modules/.bin/ts-node -r tsconfig-paths/register ./test/mock/scripts/session-error.js
> sessionError
  Browser launched
  > Running example
Test Step: Browser launched
page.goto: NS_ERROR_CONNECTION_REFUSED
=========================== logs ===========================
navigating to "http://localhost:8080/view?mock=1&id=8c28fbe6-d2c2-4a6e-84b6-045d57741081", waiting until "load"
============================================================
Note: use DEBUG=pw:api environment variable to capture Playwright logs.
Error: 
    at Object.captureStackTrace (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/utils/stackTrace.js:48:19)
    at Connection.sendMessageToServer (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/connection.js:69:48)
    at Proxy.<anonymous> (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/channelOwner.js:64:61)
    at /Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/frame.js:102:65
    at Frame._wrapApiCall (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/channelOwner.js:77:34)
    at Frame.goto (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/frame.js:100:21)
    at /Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/page.js:296:60
    at Page._attributeToPage (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/page.js:231:20)
    at Page.goto (/Users/jarilvalenciano/Local/replay/devtools/node_modules/@recordreplay/playwright/lib/client/page.js:296:21)
    at /Users/jarilvalenciano/Local/replay/devtools/test/mock/scripts/session-error.js:22:14
    Test failed
  < Running example
< sessionError
```

hbenl:
@jaril Did you have the webpack dev-server running when you tried?

Issue #2927, opened 2021-06-23
Debug Line is occassionaly missing


jasonLaster:
https://replay.io/view?id=391ad889-4fbb-42ed-864b-ea7e7935c102

PR #3012, opened 2021-07-02
Always set the DebugLine on the source picked for display (#2927)

hbenl:
- uses `ThreadFront.pickCorrespondingSourceId` in the `DebugLine` connector to ensure the line is set on the correct source
- adds `sourceUrl` to frame and pause preview locations because that is needed for `ThreadFront.pickCorrespondingSourceId`

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/DebugLine.js
+++ b/src/devtools/client/debugger/src/components/Editor/DebugLine.js
@@ -4,6 +4,7 @@
 
 //
 import { PureComponent } from "react";
+import { ThreadFront } from "protocol/thread";
 import {
   toEditorLine,
   toEditorColumn,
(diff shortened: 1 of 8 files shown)


src/devtools/client/debugger/src/components/Editor/DebugLine.js
```diff
@@ -114,6 +115,12 @@ const mapStateToProps = state => {
   const frame = getVisibleSelectedFrame(state);
   const previewLocation = getPausePreviewLocation(state);
   const location = previewLocation || (frame && frame.location);
+  if (location) {
+    location.sourceId = ThreadFront.pickCorrespondingSourceId(
+      location.sourceId,
+      location.sourceUrl
+    );
+  }
```
jasonLaster, line 123: We should not be using `ThreadFront` in our React components. What is this trying to do?

jasonLaster (commented):
Could you add an end to end test for this case too?

src/devtools/client/debugger/src/components/Editor/DebugLine.js
hbenl: Good point, I moved the logic from the connector to a selector.

> What is this trying to do?

If there are multiple identical sources, it switches to the one that is [shown by the editor](https://github.com/RecordReplay/devtools/blob/e61fc29444acebff066ea3b24419d8f36ea4d2ec/src/devtools/client/debugger/src/actions/sources/select.js#L107-L109).

PR #3040, opened 2021-07-12
Show 2minute soft limit warning on upload screen

### LINKED ISSUE #2994: UploadScreen: Add duration soft limits
**Description:**

The upload screen should nudge people to make replays less than 2 minutes

jaril:
Fix #2994.

image

### DIFF
--- a/src/ui/components/UploadScreen/UploadScreen.tsx
+++ b/src/ui/components/UploadScreen/UploadScreen.tsx
@@ -10,6 +10,7 @@ import classNames from "classnames";
 import Modal from "ui/components/shared/NewModal";
 import { Recording, UserSettings, Workspace } from "ui/types";
 import { BlankLoadingScreen } from "../shared/BlankScreen";
+import MaterialIcon from "../shared/MaterialIcon";
 
 type UploadScreenProps = PropsFromRedux & { recording: Recording; userSettings: UserSettings };
 type Status = "saving" | "deleting" | "deleted" | null;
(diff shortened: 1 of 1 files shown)


jonbell-lot23:
Following up on @jasonLaster 's comment in Discord, we agree that we need a non-subtle message for the first load, and a more subtle approach for other loads.

I think we should push this live for now because it addresses the non-subtle use-case. Then we'll design a more subtle use case for the other times it loads as a follow-on ticket.

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -56,6 +57,23 @@ function DeletedScreen({ url }: { url: string }) {
   );
 }
 
+function SoftLimitWarning() {
+  return (
+    <div className="flex flex-row space-x-2 border rounded-lg p-4 border-yellow-400 bg-yellow-50">
+      <div className="flex flex-col space-y-2">
+        <div className="flex flex-row text-yellow-800 font-medium text-xl items-center">
+          <MaterialIcon className="text-yellow-800">warning_amber</MaterialIcon>
+          <div>This replay is over 2 minutes</div>
```
jasonLaster, line 66: **This replay** feels redundant, every warning would presumably apply.

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -56,6 +57,23 @@ function DeletedScreen({ url }: { url: string }) {
...
+        </div>
+        <div>
+          For the best debugging experience, we recommend keeping your replays brief. In the future,
+          please try to keep your Replays under 2 minutes.
+        </div>
```
jasonLaster, line 71: Perhaps something simpler like this? 

---

**2+ minutes**

We recommend keeping replays below 2 minutes for the best debugging experience. _Learn More_

PR #3057, opened 2021-07-13
Add mock test for unloaded regions

bhackett1024:


bhackett1024:
This PR builds out our mock test harness in several ways:

* Message handlers are specified by the mock test itself, similar to what we do with mock graphQL responses.
* Fix existing tests to avoid errors with missing message handlers or mocked graphQL responses.
* Add a new mock test that fully loads a recording, unloads the first half of it, and checks that the timeline is updated appropriately.

src/ui/hooks/recordings.ts
```diff
@@ -354,21 +354,9 @@ export function useUpdateIsPrivate() {
...
-        }
-      }
-    `,
-    {
-      variables: { recordingId },
-    }
-  );
+  const { data, loading, error } = useQuery(GET_RECORDING_USER_ID, {
+    variables: { recordingId },
+  });
```
bhackett1024, line 359: I replaced this query with GET_RECORDING_USER_ID because the contents of these two queries are the same except for the query name itself, but I'm not 100% sure this change is correct.  Do these query names actually matter?

jaril: This should be okay. The name matters for refetching after a query, since the query name's used to figure out what to refetch. For example, we refetch `GetRecording` here after the query to initialize a recording runs: https://github.com/RecordReplay/devtools/blob/master/src/ui/hooks/recordings.ts#L712

Both `GetOwnerUserId` and `GetRecordingUserId` aren't being refetched so this should be good as is.

jaril (approved):

hbenl (approved):
Very nice!

Issue #3037, opened 2021-07-12
Don't seek to points in unloaded regions


hbenl:
see #2832

PR #3068, opened 2021-07-14
Don't seek to points in unloaded regions (#3037)

hbenl:


ryanjduffy (changes requested):
I think I'd benefit from a walkthrough of this code to follow it better. Hoping we can catch up on Monday to talk through it.

Nearly everything works. Clicking on a loading region of the timeline doesn't seek but i noticed that dragging is still allowed. I tried to manufacture a repro here. Hopefully it's representative.

https://app.replay.io/?id=84effddc-0cb2-400f-be39-cf7198a33a12

PR #3069, opened 2021-07-14
Don't expand objects from unloaded regions in the ObjectInspector

hbenl:


ryanjduffy (approved):
This seems to work as advertised though I'm curious when it will come into effect once #3067 lands.

hbenl:
> I'm curious when it will come into effect once #3067 lands.

It shouldn't make any difference for the console, but it's still possible (at least in theory) that the region containing the current pause gets unloaded, in that case it would have an effect on the scopes panel. Also, I see it as a safety net in case an object from an unloaded region accidentally "leaks" into the UI.

ryanjduffy:
:+1: Thanks for the explanation

PR #3082, opened 2021-07-15
Add doctype to index.html

ryanjduffy:
* Adds `<!DOCTYPE>` which fixes the CSS Grid on `<table>` rendering on Safari
* Resolves a number of issues that arose after leaving quirks mode


jaril (approved):
Tested this on Chrome/Replay/Safari and checked the library/upload screen/replay. All looks good!

PR #3107, opened 2021-07-19
Hide unloaded regions while loading

### LINKED ISSUE #3070: Show unloaded regions after we're done loading
We're currently using the dotted line treatment when we're loading the replay which gives the impression that dotted lines are related to loading. I think we should only show them when we're done loading so it's clearer.

jaril:
Fix #3070.

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -258,9 +258,10 @@ class Timeline extends Component<PropsFromRedux> {
   }
 
   renderUnloadedRegions() {
-    const { loadedRegions, zoomRegion } = this.props;
+    const { loadedRegions, isFinishedLoadingRegions, zoomRegion } = this.props;
 
-    if (!loadedRegions || loadedRegions.length === 0) {
+    // Check loadedRegions to keep typescript happy.
+    if (!loadedRegions || !isFinishedLoadingRegions) {
       return null;
     }
 
@@ -348,6 +349,7 @@ class Timeline extends Component<PropsFromRedux> {
 const connector = connect(
   (state: UIState) => ({
     loadedRegions: selectors.getLoadedRegions(state)?.loaded,
+    isFinishedLoadingRegions: selectors.isFinishedLoadingRegions(state),
     zoomRegion: selectors.getZoomRegion(state),
     currentTime: selectors.getCurrentTime(state),
     hoverTime: selectors.getHoverTime(state),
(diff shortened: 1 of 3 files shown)


src/ui/components/Timeline/index.tsx
```diff
@@ -260,7 +282,7 @@ class Timeline extends Component<PropsFromRedux> {
   renderUnloadedRegions() {
     const { loadedRegions, zoomRegion } = this.props;
 
-    if (!loadedRegions || loadedRegions.length === 0) {
+    if (!loadedRegions || loadedRegions.length === 0 || !this.isFinishedLoadingRegions()) {
```
ryanjduffy, line 285: These first two conditions can be dropped since they are covered by the new function.

ryanjduffy (approved):
One note above but otherwise 👍 👍

src/ui/components/Timeline/index.tsx
```diff
@@ -127,6 +131,24 @@ class Timeline extends Component<PropsFromRedux> {
...
+      loadedRegions.length === 0
+    ) {
+      return false;
+    }
+
+    const loading = loadingRegions[0];
+    const loaded = loadedRegions[0];
+
+    return isSameTimeStampedPointRange(loading, loaded);
+  }
```
jasonLaster, line 150: I think this should move to a selector. Also as a cute hack it can bet

```js
    const loading = loadingRegions?.[0];
    const loaded = loadedRegions?.[0];
    return loading && loaded && isSameTimeStampedPointRange(loading, loaded);
```

src/ui/components/Timeline/index.tsx
```diff
@@ -258,9 +280,10 @@ class Timeline extends Component<PropsFromRedux> {
   }
 
   renderUnloadedRegions() {
-    const { loadedRegions, zoomRegion } = this.props;
+    const { loadedRegions, isFinishedLoadingRegions, zoomRegion } = this.props;
 
-    if (!loadedRegions || loadedRegions.length === 0) {
+    // Check loadedRegions to keep typescript happy.
```
ryanjduffy, line 285: Ah, Typescript ...

src/ui/components/Timeline/index.tsx
```diff
@@ -127,6 +131,24 @@ class Timeline extends Component<PropsFromRedux> {
     }
   };
 
+  isFinishedLoadingRegions() {
```
ryanjduffy, line 134: This can be dropped now, right?

jaril: Yep!

ryanjduffy (approved):
:shipit:

Issue #3088, opened 2021-07-16
Stack traces for logged errors do not map their source locations


loganfsmyth:
https://github.com/RecordReplay/devtools/pull/3085 just added stack trace output to our error object rendering, but it looks like the source locations are not rewritten to their mapped location. This shows up in two ways:
* If the script has sourcemaps, the source location points to the executed source not the original files
* If the script was pretty-printed, the source location uses the non-pretty line number, but opens the pretty file, so the opened location is wrong.

PR #3119, opened 2021-07-20
Fix source mapping in the SmartTrace component (#3088)

hbenl:
The `SmartTrace` component (which renders stacktraces in the console) already tried to sourcemap the locations, but failed because it used `ThreadFront.getSourceIdsForURL()`, which favors original sources. I created `ThreadFront.getGeneratedSourceIdsForURL()` to fix that.
Then I noticed that the stacktraces for `console.error()` calls are already sourcemapped, only the stacktraces in `Error` objects aren't sourcemapped yet, hence the new prop `mapSources` which specifies if `SmartTrace` should attempt to sourcemap the locations itself.
But perhaps we want to do this in the backend instead, so that we handle stacktraces consistently?

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -359,6 +359,15 @@ class _ThreadFront {
     });
   }
 
+  getGeneratedSourceIdsForURL(url: string) {
+    // Ignore IDs which are original versions of another ID with the same URL.
+    const ids = this.urlSources.map.get(url) || [];
+    return ids.filter(id => {
+      const generatedIds = this.sources.get(id)?.generatedSourceIds;
+      return (generatedIds || []).every(generatedId => !ids.includes(generatedId));
+    });
+  }
+
   async getSourceContents(sourceId: SourceId) {
     assert(this.sessionId);
     const { contents, contentType } = await client.Debugger.getSourceContents(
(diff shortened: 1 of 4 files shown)


bhackett1024:
Doing this in the backend instead would be nice, but the backend doesn't know specifically about Error object stack traces so there has to be some involvement from the target about the stack trace formatting needs.

It would be great if we had a test for this so that we can revisit this later and make changes without breaking things.

src/protocol/thread/thread.ts
```diff
@@ -749,6 +758,14 @@ class _ThreadFront {
...
           v => new ValueFront(pause, v)
         );
+        if (message.level === "error") {
+          (async () => {
+            const argsJson = await Promise.all(
+              (message as WiredMessage).argumentValues!.map(v => v.getJSON())
+            );
+            console.log(message.level, message.source, argsJson);
+          })();
+        }
```
loganfsmyth, line 768: Is this old debug logic?

loganfsmyth (approved):
I agree that doing this in the backend would ideal eventually.

src/protocol/thread/thread.ts
hbenl: Ooops, yeah, removed it.

PR #3130, opened 2021-07-23
Demo example

jasonLaster:
This is a simple flag that disables some unnecessary chrome when we're showing the devtools on the website

https://www.loom.com/share/3c72fd2f30134153bf9baca716655ae6


src/devtools/client/debugger/src/utils/telemetry.js
```diff
@@ -2,70 +2,6 @@
...
-  if (window.parent.frameElement) {
-    sessionId = window.parent.frameElement.getAttribute("session_id");
-  }
-
-  /* eslint-disable camelcase */
-  telemetry.recordEvent(eventName, "debugger", null, {
-    session_id: sessionId,
-    ...fields,
-  });
+  return;
```
jasonLaster, line 6: I don't think `recordEvent` was being used and it is certainly pretty ugly. We might want to use another tool in the future though so keeping recordEvent around

PR #3135, opened 2021-07-23
Initialize parts of the app dynamically (#3098)

hbenl:


hbenl:
This is a complete refactoring of the app's setup logic.
- the initial setup is handled by `src/ui/setup/*.ts` and setup for dynamically loaded parts of the app is handled by `src/ui/setup/dynamic/*.ts`
- a dynamic setup file should be included by the root component of the corresponding feature (e.g. [here](https://github.com/hbenl/webreplay-devtools/blob/76b0ede6d49611fab00748c109419afacc4c40ed/src/devtools/client/inspector/components/App.tsx#L14)). That component should then be wrapped with `React.lazy()` and `React.Suspense` (e.g. [here](https://github.com/hbenl/webreplay-devtools/blob/76b0ede6d49611fab00748c109419afacc4c40ed/src/ui/components/SecondaryToolbox/index.tsx#L18) and [here](https://github.com/hbenl/webreplay-devtools/blob/76b0ede6d49611fab00748c109419afacc4c40ed/src/ui/components/SecondaryToolbox/index.tsx#L106-L108))
- the app's routing was moved to `src/views/routing.tsx`

jasonLaster:
At a high level, this looks good, but I think Jaril or Ryan should do a review as well.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -0,0 +1,126 @@
...
+import { getUserSettings } from "ui/hooks/settings";
+const { LocalizationHelper } = require("devtools/shared/l10n");
+const { setupDemo } = require("ui/utils/demo");
+
+declare global {
+  interface Window {
+    gToolbox: DevToolsToolbox;
+    L10N: any;
+  }
+  interface AppHelpers {
```
ryanjduffy, line 46: This interface overlaps with the one defined in helpers.ts as well. Perhaps that's okay right now if the goal is to incrementally sunset these globals but if there are multiple modules that set up / interact with these globals, it might be worth consolidating them. I don't have specific advice in mind yet but we can discuss if you think it'd be useful.

src/ui/setup/dynamic/inspector.ts
```diff
@@ -0,0 +1,16 @@
+import { extendStore } from "../store";
```
ryanjduffy, line 1: I'm curious why this was refactored up to the shared ui space if it's only used by the inspector. Maybe the real question is why the inspector needs dedicated space in the store if it's the only one using the data. /shrug

src/views/routing.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useEffect } from "react";
```
ryanjduffy, line 1: I have a small preference to move this up a level so that `views` only contains the actual views and not the routing logic but I wouldn't want to block this on that. We can also revisit once the rest of the URL routing changes go in which would remove a chunk of this code as well.

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -15,6 +14,8 @@ import { UIState } from "ui/state";
 import { PanelName } from "ui/state/app";
 import { isDemo } from "ui/utils/environment";
 
+const InspectorApp = React.lazy(() => import("devtools/client/inspector/components/App"));
```
ryanjduffy, line 17: Is the inspector particularly heavyweight to lazy load it?

src/ui/components/WaitForReduxSlice.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ReactNode } from "react";
```
ryanjduffy, line 1: This might be a bit more abstraction than is needed for an early bail out but curious to hear more from you about it. If we want to keep it, I'd suggest converting it to a higher-order component so the usage is a bit more straightforward.

For example:

```jsx
const ConnectedDevTools = connector(waitForSlice('messages', _DevTools));
```

ryanjduffy (approved):
I think this is a good improvement overall. I like the reorganization of the setup logic; I think it's easier to find and follow. Some assorted comments below for consideration or discussion.

Thanks and nice work, @hbenl!

src/ui/setup/dynamic/devtools.ts
hbenl: I kept the declarations for global variables in the files that set those variables, but I'm open to organizing it differently. Perhaps we should discuss and write up some guidelines about the folder structure of our code (and refactor the code accordingly) - currently I'm often unsure what the "right" place for new code would be.

src/ui/setup/dynamic/inspector.ts
hbenl: I kept all the setup logic in one folder, but I'm open to organizing it differently.

> Maybe the real question is why the inspector needs dedicated space in the store if it's the only one using the data

Where else would you put that data?

src/ui/components/SecondaryToolbox/index.tsx
hbenl: No, but it's currently disabled by default so I thought it makes sense not to ship it to every user. But more importantly I used this to verify that it's relatively simple with the new setup organization to make a panel lazy-loadable.

src/ui/components/WaitForReduxSlice.tsx
hbenl: The issue I'm trying to solve here is that the setup logic for the devtools is asynchronous, so we need a bail out mechanism for the DevTools component while this setup is running. That mechanism needs to be in a separate component wrapping the DevTools component because the DevTools component itself tries to read data from the redux store that is only available after the asynchronous setup logic has finished. In the end this was the simplest solution I could think of.
I haven't worked with higher-order components yet, I'd love to try that out.

src/ui/setup/dynamic/inspector.ts
ryanjduffy: My inclination is to keep local state at the component and move it to redux when it needs to be used broadly across the app. That does add some mental complexity though since we have to then keep track of what is "shared app state" vs "local component state" but can ultimately lead to a smaller app state tree which is easier to grok than "every piece of state used anywhere in the app ever". :)

src/ui/setup/dynamic/devtools.ts
ryanjduffy: > I kept the declarations for global variables in the files that set those variables

This seems like a good practice when you need to use globals but my preference would be to avoid globals and to import the shared methods directly where they are needed.

In cases where we _want_ a global (like the `app` var that helps expose some stuff in user-land. This approach seems reasonable to me.

My point in this specific case is that we have `AppHelpers` defined twice which TS happily merges into one and I'd favor either bringing them into the same module or breaking them up more clearly so it's evident that `global.app` is fulfilled by two different modules.

src/ui/components/WaitForReduxSlice.tsx
ryanjduffy: Here's a quick branch to show what I think it could look like:

https://github.com/RecordReplay/devtools/compare/master...ryanjduffy:ryan/slice-hoc?expand=1

PR #3153, opened 2021-07-27
Test region loading.

kannanvijayan:


test/scripts/region_loading-01.js
```diff
@@ -0,0 +1,38 @@
...
+
+    await Test.warpToMessage("Number 5");
+
+    await Test.checkEvaluateInTopFrame("number", 5);
+    await Test.reverseStepOverToLine(19);
+
+    // The logpoint acts like a breakpoint when resuming.
+    await Test.resumeToLine(20);
+  }
+);
```
jasonLaster, line 38: Test looks good, but it's not clear how this is testing region loading?

kannanvijayan: It's not.  It's just a copy of the test files.  I didn't add reviewers to this yet, but mostly wanted to see if I could make an new test file and have it pass.

src/protocol/thread/thread.ts
```diff
@@ -245,13 +247,23 @@ class _ThreadFront {
...
-      listenerCallback(parameters);
+      assert(this.loadChangeListeners);
+      for (const cb of this.loadChangeListeners) {
+        cb(parameters);
+      }
     });
 
     await client.Session.listenForLoadChanges({}, sessionId);
   }
+  addLoadChangeListener(cb: (params: loadedRegions) => void) {
```
hbenl, line 263: I don't like having 2 methods for adding a listener for loaded regions

test/scripts/region_loading-01.js
```diff
@@ -0,0 +1,45 @@
...
+    await Test.removeAllBreakpoints();
+
+    // Unload a region of the recording 1 second in, 2 seconds in duration.
+    await Test.unloadRegion(...unloadTimerange);
+
+    await Test.addBreakpoint("doc_rr_region_loading.html", 20, undefined, {
+      logValue: `"Part2Logpoint Number " + number`,
+    });
+
+    Test.waitUntil(() => Test.findMessages("Loading...").length === 0);
```
hbenl, line 30: I think you forgot the `await` here. Furthermore, if the "Loading..." messages aren't added synchronously, this may finish immediately (before any breakpoint messages are added to the console instead of after all messages are updated).

hbenl (commented):
It would be nice if we could get rid of the logic for detecting when loading/unloading finishes - is there any chance that the backend could track this and send the reply to the load/unload request when loading/unloading is finished?

test/scripts/region_loading-01.js
```diff
@@ -0,0 +1,45 @@
+// Test basic logpoint functionality. When logpoints are added,
```
hbenl, line 1: Remove or update this comment

src/protocol/thread/thread.ts
kannanvijayan: Should I just modify `listenForLoadChanges` to allow for more than one listener?

hbenl: > make listenForLoadChanges be the method that is allowed to be called more than once per program, and eliminate addLoadChangeListener

Either that, or we can change the backend (as described below) so that we can simply do `await Test.loadRegion(...)` and know that the loading has finished. In that case, we could remove this altogether (and also some code in the test harness that tries to figure out when loading a region has finished).

kannanvijayan:
> It would be nice if we could get rid of the logic for detecting when loading/unloading finishes - is there any chance that the backend could track this and send the reply to the load/unload request when loading/unloading is finished?

Agreed, but that's a bit more work, and requires a further round of backend and protocol changes.  Right now "loadedRegions" events are sent out on an "on-changed" basis (but behind a timer throttle to prevent frequent updates).  I don't think we should block the test for that.

kannanvijayan:
@hbenl I'm running into an issue with calling `Test.findMessages` from within the test script.  It's return value is not an array.

I noticed that the functions exposed via the `Test` global are instrumented to print something out at the beginning ("Starting X" and "Finished X at").  Is there something that goes through `Test` and wraps all the bound functions with a wrapper that does those prints?  I suspect there may be an issue in that code that changes the returned values in some way.

src/protocol/thread/thread.ts
kannanvijayan: That would require doing the protocol and backend changes to detect end-of-loading as we discussed below.  The alternative is somehow hooking into the model that's being updated, and get a callback whenever `app.loadedRegions` is updated.

It seems like the capability to do that would be required in general for efficient diffing of models to figure out which re-renders to issue for the UI.  Can we attach callbacks at the model level rather than listening at the protocol socket level?  If so how do I do that?

kannanvijayan: After discussion with @hbenl we figured we could remove the `listenForLoadChanges` hooks entirely and rely purely on watching the UI to run through the test.  Tried that and it worked.

hbenl (approved):

kannanvijayan:
Other tests are failing due to intermittent issues but the relevant test in stripe 4 passes.

PR #3155, opened 2021-07-27
Add email address validation when inviting an email to Replay

jaril:
Fixes https://github.com/RecordReplay/backend/issues/2696

This adds some email address validation before inviting a user to Replay. Specifically this checks for a valid email address and that the e-mail address hasn't already been invited by the same user before.

This also sorts the invitations reverse-chronologically so that the newest invitations are displayed up top instead of being buried in the overflow.

### DIFF
--- a/src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
+++ b/src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
@@ -2,11 +2,14 @@ import React, { useState } from "react";
 import hooks from "ui/hooks";
 import "./ReplayInvitations.css";
 import { TextInput } from "ui/components/shared/Forms";
+import { validateEmail } from "ui/utils/helpers";
+import { Invitation } from "ui/hooks/invitations";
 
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/SettingsModal/ReplayInvitations.tsx
```diff
@@ -16,6 +19,20 @@ export default function ReplayInvitations() {
 
   const onSubmit = (e: React.FormEvent) => {
     e.preventDefault();
+
+    const isInvalidEmail = !validateEmail(inputValue);
+    const isAlreadyInvited =
+      invitations && invitations.map(i => i.invitedEmail).includes(inputValue);
```
jasonLaster, line 25: you might be able to use `invitations.some(() => {})`

jasonLaster (approved):
Looks good. 

Could you an include a replay of the new flow so I can see what it looks like. Also, not necessary now, but I think it would be great to use a light weight testing tool like react-testing for some of these components.

hbenl (approved):

PR #3160, opened 2021-07-28
Check the size of the main bundle in Github CI

hbenl:


jasonLaster (approved):

hbenl:
The bundling has recently changed so that the main bundle is significantly smaller, but it doesn't contain the code necessary for the library anymore. So if we want to verify how much code needs to be downloaded before the library can be displayed, we need a different approach. Perhaps we could write e2e tests that verify how much code is downloaded for several routes (i.e. the browser pages, the library and the devtools).

jasonLaster:
Perhaps we look at the sizes for all the chunks and have a maximum that makes sense across the board?

jasonLaster:
Thanks for doing this btw

Issue #2987, opened 2021-06-29
Does SOURCE 27 have a filename?


jasonLaster:
@andarist is pretty sure source 27 should have a filename. Let's see what's going on here

> This doesn't show a filename or anything useful. I know it might even be coming from eval or some other esoteric stuff so that might be expected - but I have a feeling that it's not. Can't be sure though. In addition to that - this source doesn't get prettified which is a problem becasue I can't setup breakpoints in it.

https://replay.io/view?id=a486a6d4-c30d-4391-b140-5f583cdf3a08&point=74963785911711832975260339536921895&time=28224&hasFrames=false

Andarist:
Well, im not 100% sure but its a production site so i suspect that this should have a filename. And even if it doesnt - this could be displayed/labeled better, right now SOURCE X is pretty enigmatic and confusing for users as they are not familiar with this scheme.

The second weird issue is that this doesnt get pretty printed.

loganfsmyth:
This is caused by https://github.com/RecordReplay/devtools/blob/f1c5ed464a97dc8abb41abde95a71e36c532234f/src/devtools/client/debugger/src/actions/sources/newSources.js#L216 because the file being viewed _has_ a pretty version, but @Andarist was viewing the non-pretty version.

Seems like maybe that code needs to move somewhere else.

hbenl:
This source should never have appeared in the search results since a pretty printed version of it is available. The linked PR fixes the detection of minified sources in global text searches.

PR #3169, opened 2021-07-30
Fix the detection of minified sources during a global text search (#2987)

hbenl:


src/devtools/client/debugger/src/actions/project-text-search.js
```diff
@@ -77,7 +78,7 @@ export function searchSources(cx, query) {
     await dispatch(addSearchQuery(cx, query));
     dispatch(updateSearchStatus(cx, statusType.fetching));
     const validSources = getSourceList(getState()).filter(
-      source => !hasPrettySource(getState(), source.id) && !isThirdParty(source)
+      source => !ThreadFront.isMinifiedSource(source.id) && !isThirdParty(source)
```
Andarist, line 81: when prettifying happens? do you prettify upfront everything so for every minified source there is always a pretty variant available? or do u prettify on demand? Im a little bit worried that if the latter is the case then you might miss valid search results here 🤔 

could you also clarify for me, a curious reader, what a third-party source is, in this context?

hbenl: Sources are prettified by the backend and I don't know the logic for when this happens. But `ThreadFront.isMinifiedSource()` actually checks if there is a prettified variant of the given source (the name of this method is a bit misleading), so I'm sure we won't miss any search results here.
`isThirdParty()` checks if the source's URL contains `node_modules` or `bower_components`.

PR #3136, opened 2021-07-23
Add team API Key UI

ryanjduffy:
## Feature

Adds UI to team settings to create workspace API keys

https://app.replay.io/?id=b9e95bdd-b1b2-4663-9b40-91ceced9d12d

## Changes

* Refactors the `APIKeys` component out of the user settings view and into a shared component reused by both team and user settings
* Adds support for selecting `scopes` if more than one scope is supported. For user keys, this is suppressed. For team keys, at least one scope must be selected (and both are selected by default).
* Adds simple link to the bottom of team settings to switch to the API keys view

image

PR #3171, opened 2021-07-30
Fix React warning message

hbenl:
The `connect()` method of `react-redux` adds the `dispatch()` function to the component's props, which `<div>` doesn't expect, so in development mode, React warns about this prop.


jaril (approved):

jaril:
Re-ran tests but I think this should be good to go!

ryanjduffy:
I've also fixed this (albeit differently) in #3136

PR #3181, opened 2021-08-03
Split out app bundle from browser pages

ryanjduffy:
## Summary

This continues the work started in #3098 to further split the app bundle from the browser pages to reduce the load on these "standalone" pages. In particular, this lightens the work required for `/browser/new-tab` so it can launch as quickly as possible from devtools.

## Changes

* Move the routing code back up to `main.js` and lazy load each of the views from there
* Refactors out the common view styling from `app.css` into `base.css` and import that from `main.js` so each view has the same base font size and html/body sizing
* Moves the remaining styles from `/styles.css` to `ui/components/App.css` so it can be alongside the component that imports it.

### DIFF
--- a/src/base.css
+++ b/src/base.css
@@ -1,0 +1,32 @@
+/* NOTE: this is added to protect against React DevTools */
+html {
+  font-size: 12px !important;
+}
+
+html, body {
+  height: 100vh;
+  width: 100vw;
+  margin: 0;
+  padding: 0;
+  overflow: hidden;
+}
+
+* {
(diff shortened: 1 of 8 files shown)


hbenl (approved):

ryanjduffy:
Added a few `preventDefault()` calls to links that would navigate to `#` (e.g. Team Settings, Add a new team) and cause unnecessary full page reloads. This will be improved as we migrate to more URL navigation but this is sufficient for now.

PR #3184, opened 2021-08-04
Add an add comment button to console log messages

### LINKED ISSUE #3094: No easy way to add comment on a console message
We have an "add comment" button when a breakpoint is set on a line, but when paused at a console message there probably isn't a breakpoint.  In this case going to the comments pane asks to click on the video to add a comment.  Mousing over the video shows a nonsense tooltip "Add comment to this line of code".

The tooltip needs to be fixed, but this UX is pretty awkward regardless.  It should be easy to add a comment when in the comments pane, without having to click on the video.

jaril:
Fix #3094

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -312,6 +312,8 @@
 }
 
 .webconsole-app .message-location {
+  display: flex;
+  flex-direction: row;
   float: right;
   color: var(--frame-link-source);
   /* Makes the file name truncated (and ellipsis shown) on the left side */
@@ -326,18 +328,6 @@
 
 .message-location a {
   color: var(--frame-link-source);
-}
-
-.message-jump-button {
-  display: none;
-}
-
-.message:hover .frame-link-source {
-  display: none;
-}
-
-.message:hover .message-jump-button {
-  display: block;
 }
 
 .webconsole-app .message-location:empty {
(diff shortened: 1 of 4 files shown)


jonbell-lot23:
Found a bug @jaril. If you're logged out, you can still add a comment:

image

(Obviously the comment isn't successfully added)

jonbell-lot23:
Pinging @jaril -- this didn't go live because of the conflicting files. Can you take a look and get it live?

jaril:
> Found a bug @jaril. If you're logged out, you can still add a comment:
> 
> image
> 
> (Obviously the comment isn't successfully added)

Split this out here: #3388

jaril:
Tests are looking flaky for unrelated reasons — waiting for this to go green then merging.

jasonLaster:
3 comments

1. We should use the primary blue gradient
2. I think it's better to continue using the correct rewind / replay button orientation
3. I wonder if we should shift the location N pixels to the left to make space for the button. Seeing the button remove the location is distracting

src/devtools/client/themes/webconsole.css
```diff
@@ -1071,7 +1061,7 @@ a.learn-more-link.webconsole-learn-more-link {
 }
 
 .webconsole-output .overlay-container:not(.debug):hover .button {
-  background: linear-gradient(152.2deg, #67daff 6.75%, #4382ff 93.14%);
+  background: linear-gradient(to bottom right, #68DCFC, #4689F8);
```
jonbell-lot23, line 1064: I think we settled on a solid colour rather than a gradient:

image

If so, it should be our primary accent colour.

Issue #3174, opened 2021-07-30
Chunk load errors when loading/navigating older versions of devtools


loganfsmyth:
I was hoping the Webpack 5 update would fix this, and having the hash in the URL does help avoid some errors, but there's still an issue here.

The primary thing is that it appears Vercel only hosts the current build's assets, so if you open devtools, and then we deploy, and then you perform an action that needs to load a bundle, it will fail because the bundle for that build no longer exists in the new deploy's assets. Now with the hash in the URL, instead of loading a file with the same name that has the wrong contents, we fail to load entirely, which at least makes it clearer that something is going wrong. This is what leads to `ChunkLoadError`s we're seeing in Sentry.

Secondly, our current routes mean that if you try to load a script and it doesn't exist, instead of getting a 404, we serve our `index.html`. So for instance if you load https://app.replay.io/82.615bf96308cfcc931130.js, you get the library view. This is what's leading to the `Unexpected token '<'` errors in Sentry.

For example, currently https://app.replay.io/82.6fe265211f12860bb99c.js returns JS for the current build, and https://app.replay.io/82.615bf96308cfcc931130.js returns the HTML for the index.

I would not be surprised if most of the errors in Sentry are my own sessions, because I use
<img width="417" alt="Screen Shot 2021-07-30 at 10 39 20 AM" src="https://user-images.githubusercontent.com/132260/127691294-71e08f7e-60f4-43ff-ab48-ee0979a3a360.png">, and it appears that with our current caching setup, it pulls up my most recent Replay page HTML from cache, as well as `main.*.js`, but when that scripts tries to run, it pretty much immediately tries to load more bundles and I guess those aren't in the cache and thus if a deploy has happened between the time that I closed the browser and opened it again, the requests all get HTML back instead of 404ing, and I end up with this:
<img width="1731" alt="Screen Shot 2021-07-30 at 10 43 21 AM" src="https://user-images.githubusercontent.com/132260/127691704-ce3a7ad3-fbbe-4f90-88c0-3ec82107764c.png">

loganfsmyth:
@hbenl

PR #3189, opened 2021-08-05
Ask the user to refresh the page when an unexpected error occurs (#3174)

hbenl:


jasonLaster (approved):

hbenl:
This would not be necessary if we use the alternative approach for prod deployments described in #3194

jasonLaster (approved):

PR #3191, opened 2021-08-05
Move recording id into URL path

ryanjduffy:


ryanjduffy:
@hbenl - This doesn't work yet and probably need a fair amount of work but sharing the current state in case you have insight on how to approach. I think we'll need to maintain full page reloads for now because so much of the app seems to depend upon initializing everything on load. We'll also probably want to add a redirect from `?id=:id` to `/:id` for backwards compatibility. 

Might be that I'm biting off too much for now and we could simplify this by adding the routing path and updating the code pulls the query string parameter to pull from the path instead.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -56,10 +56,12 @@ declare global {
 }
 
 const url = new URL(window.location.href);
-const recordingId = url.searchParams.get("id")!;
+const recordingId = url.pathname.match(/\/([a-z0-9-]+)/i)?.pop();
```
loganfsmyth, line 59: Seems like we should explicitly do `/recording/:uuid` since we would well also want `/workspace/:uuid` or something at some point. This regex also needs to start with `^`.

src/views/recording.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useIsRecordingInitialized, useGetRecordingOwnerUserId } from "ui/hooks/recordings";
+import BlankScreen from "ui/components/shared/BlankScreen";
+import { bootstrapApp } from "ui/setup";
+import { registerRecording } from "ui/utils/telemetry";
+import Upload from "views/upload";
+import DevTools from "ui/components/DevTools";
+
+import "ui/setup/dynamic/devtools";
+
+bootstrapApp();
```
hbenl, line 16: This is already called in `app.tsx` (and should be called there because it's also needed by the `Account` component)

src/views/app.tsx
```diff
@@ -1,106 +1,39 @@
...
-          <App>
+const AppRouting = () => {
+  return (
+    <tokenManager.Auth0Provider>
+      <ApolloWrapper>
+        <IntercomProvider appId={"k7f741xx"}>
+          <Provider store={window.store}>
             <React.Suspense fallback={<div>Loading</div>}>
-              <ConnectedPageSwitch />
+              <App>
```
hbenl, line 25: I kept `<App>` outside of `<React.Suspense>` because I didn't want the setup work done in the `App` component to be repeated. Any particular reason why you changed how these components are nested?

src/ui/components/DevTools.tsx
```diff
@@ -112,9 +110,9 @@ type DevToolsProps = ConnectedProps<typeof connector>;
 const ConnectedDevTools = connector(_DevTools);
 
 const DevTools = () => (
-  <WaitForReduxSlice slice="messages">
-    <ConnectedDevTools />
-  </WaitForReduxSlice>
+  // <WaitForReduxSlice slice="messages">
```
hbenl, line 113: Without `WaitForReduxSlice`, the `DevTools` can be rendered before their setup code has finished, which may lead to a crash.

hbenl:
> I think we'll need to maintain full page reloads for now because so much of the app seems to depend upon initializing everything on load

Theoretically, switching from the library to devtools should already be possible without reloading, but I never tested this. Anyway I think we should focus on the route changes for now and worry about doing less reloads after that.

src/ui/components/DevTools.tsx
ryanjduffy: This was a debug change. DevTools wasn't rendering so I removed this temporarily.

src/views/app.tsx
ryanjduffy: Just experimenting. Agree it should be outside the suspense.

src/views/recording.tsx
ryanjduffy: I think this was a copy/paste fail. It should be removed.

src/ui/setup/dynamic/devtools.ts
ryanjduffy: That's a good point. I hadn't considered teams as an endpoint but agree we should plan for it. That regex will hopefully go away and was just a quick hack to get the value for testing things out.

jasonLaster:
is this still needed?

ryanjduffy:
Nope! Closing

PR #3192, opened 2021-08-06
Keep devtools loading screens from flickering

### LINKED ISSUE #3190: Devtools should not show a loading label
The loading text in the top left is pretty jarring

https://app.replay.io/?id=6a7734a2-fc1f-44f1-b0f6-a91bc43a584d&point=4218741198514836301389300479558384&time=1621.9387935609411&hasFrames=true

jaril:
Fix #3190.

Right now when you open a Replay, there's a mix of white screens/blue screens/spinners/loading bars that show up before DevTools finally loads up. 

This makes it so that when we open a recording, it goes:
```
white screen -> spinner -> loading bar (scanning) -> devtools
```

There's a cleaner way to do this but this cleans this up right now and we can do a follow up.

Before:
https://app.replay.io/?id=a4cbbc7d-4434-48bf-a442-659d5fa3ff71
After:
https://app.replay.io/?id=fee081e4-19c8-4d2b-8152-b32067149e21

### DIFF
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -11,8 +11,10 @@ import { UIState } from "ui/state";
 import { BlankLoadingScreen, BlankProgressScreen } from "./shared/BlankScreen";
 import NonDevView from "./Views/NonDevView";
 import WaitForReduxSlice from "./WaitForReduxSlice";
+import LogRocket from "ui/utils/logrocket";
 
 import "ui/setup/dynamic/devtools";
+import { useAuth0 } from "@auth0/auth0-react";
 
 const DevView = React.lazy(() => import("./Views/DevView"));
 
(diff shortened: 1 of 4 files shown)


src/ui/components/DevTools.tsx
```diff
@@ -60,6 +64,11 @@ function _DevTools({
       setExpectedError(expectedError);
     }
   }, [recordingId, expectedError]);
+  useEffect(() => {
+    if (!recordingQueryLoading && !userInfo.loading && recording) {
+      LogRocket.createSession(recording, userInfo, auth);
```
jaril, line 69: We only need to create a session in DevTools, so I moved this code here.

jazzdan: Is that because we don't want LogRocket sessions for the library view?

jazzdan (approved):
LGTM!

src/ui/components/DevTools.tsx
jaril: Yeah. I think we do want to track other flows like the library eventually, but the current implementation keeps that from happening anyway. So logrocket behaves the same before and after this PR.

PR #3198, opened 2021-08-06
Fix debugger panels overflowing

### LINKED ISSUE #3187: Sidebar scrolling and layout can break
I noticed two issues recently with the devtools sidebar:

1. The vertical scrolling looks broken
2. Sometimes the layout can be shifted to the right. I think this happens if a source is particularly wide or there are lots of editor tabs

Screen Shot 2021-08-04 at 1 53 37 PM

Screen Shot 2021-08-03 at 8 51 45 AM

ryanjduffy:
Fixes #3187 by fixing the flex box sizing (because it's notoriously onery)

Repro for part 2:
* Open a replay with multiple sources
* Open several sources from the explorer such that one of the tabs is clipped on the right
* Select the overflowing source from the explorer

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/SecondaryPanes.css
@@ -27,6 +27,7 @@
 
 .secondary-panes-wrapper {
   width: 100%;
+  height: 100%;
   display: flex;
   flex-direction: column;
 }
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/SecondaryPanes.css
```diff
@@ -27,13 +27,23 @@
...
   margin-bottom: 0;
+  display: flex;
+  flex-direction: column;
+}
+
+.secondary-panes .accordion li {
+  flex: 1 1 auto;
+  height: 0;
+  display: flex;
+  flex-direction: column;
```
jasonLaster, line 46: I bet we can remove these styles

PR #3212, opened 2021-08-09
Load the screenshots in the library lazily

hbenl:


jasonLaster (approved):

jasonLaster:
5 seems like a no brainer for now. We'll probably still want to revisit storing images in the DB, but it can certainly wait too and the perf impact is huge

PR #3228, opened 2021-08-11
Allow seeking to unloaded regions

### LINKED ISSUE #3162: Uploading screen has white text
When viewing a replay a few minutes ago I saw a white box in the center of the screen with no text.  Selecting the box with my mouse revealed some text around uploading the replay.  The replay then finished loading so I didn't get to take a screenshot or create another replay of the problem.

jasonLaster:
Fixes #3162 

This allows users to seek to unloaded locations

before: https://app.replay.io/?id=37b65c20-6a3a-46ce-b8ee-bbbbbc3ca46d&point=25961484298513924218237035376279794&time=11843.80976560729&hasFrames=true
after: https://app.replay.io/?id=b35659de-86e0-40d7-8f3c-aeaa6d7a8a56

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -253,11 +253,7 @@ export function seek(
   hasFrames: boolean,
   pauseId?: PauseId
 ): UIThunkAction<boolean> {
-  return ({ getState, dispatch }) => {
-    if (!selectors.isRegionLoaded(getState(), time)) {
-      return false;
-    }
-
+  return ({ dispatch }) => {
     const pause = pauseId !== undefined ? Pause.getById(pauseId) : undefined;
 
     updateUrl({ point, time, hasFrames });


jasonLaster:
These two test failures seem to be more about sporatic backend issues 
* https://app.replay.io/?id=bfc9afb6-af82-497b-8355-d3703211adf6&point=175889056086474988289020730393234846&time=71515&hasFrames=false
* https://app.replay.io/?id=09800e59-8d16-4f85-b16c-1f53e627ec17&point=131754532788213488407162154387506472&time=83476&hasFrames=false

Is it clear what issues these are? How often they occur? ...

src/ui/actions/timeline.ts
```diff
@@ -253,11 +253,7 @@ export function seek(
   hasFrames: boolean,
   pauseId?: PauseId
 ): UIThunkAction<boolean> {
-  return ({ getState, dispatch }) => {
-    if (!selectors.isRegionLoaded(getState(), time)) {
-      return false;
-    }
-
```
jasonLaster, line 260: Thinking about this more, I think the logic should read

```js
  return ({ getState, dispatch }) => {
    if (!selectors.isRegionLoaded(getState(), time)) {
      hasFrames = false;
      pauseId = null;
    }
```

That way we always seek to the correct point in time, but we dont try to perform actions that will fail

hbenl: I think we should try removing this altogether first. If we set `pauseId` to `null` here, a new Pause will be created for the given point, which may then be reused later when the region has been loaded again. In this case the frontend may falsely keep assuming that there are no frames for this point.

hbenl (approved):

PR #3229, opened 2021-08-11
telemetry: set user before sending

jazzdan:


src/ui/utils/telemetry.ts
```diff
@@ -66,12 +67,13 @@ export async function sendTelemetryEvent(event: string, tags: any = {}) {
...
       return;
     }
+    tags["user"] = telemetryUser;
     const response = await fetch("https://telemetry.replay.io/", {
       method: "POST",
       headers: {
         "Content-Type": "application/json",
       },
-      body: JSON.stringify({ event, ...tags, user: telemetryUser }),
+      body: JSON.stringify({ event, ...tags }),
```
jasonLaster, line 76: hmm, why do we need to change this? In general, i think it is a bit cleaner to set the user in the new object as opposed to mutating the tags

src/ui/utils/telemetry.ts
```diff
@@ -43,6 +43,7 @@ type TelemetryUser = {
 let telemetryUser: TelemetryUser | undefined;
 
 export function setTelemetryContext({ id, email, internal }: TelemetryUser) {
+  telemetryUser = { id, email, internal };
```
jasonLaster, line 46: 👍

jasonLaster (approved):

src/ui/utils/telemetry.ts
jazzdan: Yeah I'll revert this change.

PR #3230, opened 2021-08-12
Add the protocol client to the app helper

jasonLaster:


src/ui/setup/dynamic/devtools.ts
```diff
@@ -1,6 +1,6 @@
 import { bindActionCreators } from "redux";
 
-import { initSocket } from "protocol/socket";
+import { initSocket, client } from "protocol/socket";
```
ryanjduffy, line 3: I don't think this needs to be added here as well. The setup in helpers.ts is run for all views so adding `client` there is sufficient. Plus dropping these changes will eliminate the TS errors (which could be fixed but are an artifact of these two places both settings up different parts of the app helper).

PR #3235, opened 2021-08-13
Refactor the routing

hbenl:
@ryanjduffy This is essentially your PR (#3191) with a few fixes and the following changes:
- I changed the routes for recordings to be `/recording/:recordingId` (as suggested [here](https://github.com/RecordReplay/devtools/pull/3191#discussion_r683796893))
- when clicking on a recording in the library, the app no longer needs to reload (see the changes in `RecordingItem.tsx`)

### DIFF
--- a/src/ui/reducers/app.ts
+++ b/src/ui/reducers/app.ts
@@ -6,11 +6,9 @@ import { prefs } from "../utils/prefs";
 import { Location } from "@recordreplay/protocol";
 import { getLocationAndConditionKey } from "devtools/client/debugger/src/utils/breakpoint";
 import { isSameTimeStampedPointRange } from "ui/utils/timeline";
-import { isDemo } from "ui/utils/environment";
 
 function initialAppState(): AppState {
   return {
-    recordingId: null,
     expectedError: null,
     unexpectedError: null,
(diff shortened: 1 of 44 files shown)


src/ui/reducers/app.ts
```diff
@@ -49,10 +47,6 @@ export default function update(
   action: AppActions | SessionActions
 ): AppState {
   switch (action.type) {
-    case "setup_app": {
```
ryanjduffy, line 52: Let's clean up the definition of this action in `ui/actions/app.ts` as well

src/views/app.tsx
```diff
@@ -106,4 +37,21 @@ const AppRouting = () => (
   </tokenManager.Auth0Provider>
 );
 
+function ViewRedirect() {
```
ryanjduffy, line 40: This is great! I think we should redirect on `/?id={recordingId}` too since the current version on `app.replay.io` produces that URL.

ryanjduffy (approved):
This looks great, @hbenl. I think it makes the app feel more responsive and sets the stage for better URLs and navigation throughout the app.

One comment below on redirects but otherwise, :shipit:!

PR #3256, opened 2021-08-17
Mixpanel events

jasonLaster:
This PR adds a new `trackEvent` API that uses mixpanel to setup our user funnel 

1. new replay
2. visit devtools
3. seek in the console to a point in time
4. add a breakpoint
5. add a comment
6. edit breakpoint

Here's a [replay](https://app.replay.io/recording/a3a376d5-881f-4559-afbf-230025ba0fb3?point=10060075164575555887329213690151644&time=5337.298172607201&hasFrames=true
) where I walk through the funnel

The nice thing is we can confirm that the mixpanel context is set as well

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/modify.js
+++ b/src/devtools/client/debugger/src/actions/breakpoints/modify.js
@@ -2,10 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-//
-
-import mixpanel from "mixpanel-browser";
-
 import { getLocationKey, getASTLocation } from "../../utils/breakpoint";
 
 import {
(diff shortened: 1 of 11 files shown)


src/devtools/client/debugger/src/actions/breakpoints/modify.js
```diff
@@ -89,12 +76,6 @@ export function addBreakpoint(
   shouldCancel = () => false
 ) {
   return async ({ dispatch, getState, client }) => {
-    recordEvent("add_breakpoint");
-    const workspace = getRecordingWorkspace(getState());
-    if (shouldTrack) {
-      trackBreakpoint(workspace);
-    }
-
```
jasonLaster, line 97: this is no longer needed

src/devtools/client/debugger/src/actions/breakpoints/modify.js
```diff
@@ -163,8 +144,6 @@ export function addBreakpoint(
 
 export function runAnalysis(cx, initialLocation, options) {
   return async ({ getState, client }) => {
-    recordEvent("run_analysis");
```
jasonLaster, line 166: recordEvent was a no-op so i just deleted it

PR #3257, opened 2021-08-17
Auto accept the TOS when a user logs in

jaril:


src/ui/utils/tokenManager.tsx
```diff
@@ -47,21 +47,19 @@ class TokenManager {
...
 
-  Auth0Provider = ({ children }: { children: ReactNode }) => {
+  Auth0Provider = ({ children, onRedirect }: { children: ReactNode; onRedirect: () => void }) => {
     const history = useHistory();
 
     const onRedirectCallback = (appState: AppState) => {
+      onRedirect();
+
       if (appState?.returnTo) {
         window.location.href = appState.returnTo;
```
loganfsmyth, line 57: Did you end up verifying if our approach here works? I realize this is going to be a problem because this will immediately refresh the page and thus won't allow the callback to do much useful. I _think_ we can probably use `history.replace` here to work around this issue, but I'm not super familiar with the router `history` API and not sure why this was written this way in the first place. Probably a good question for @hbenl ?

hbenl: Yeah, this isn't working. But using `history.replace()` instead of `window.location.href =` makes sense anyway (to avoid an unnecessary reload) and with that change it would work. And we can also remove the `else` branch here, pushing the current url again makes no sense.
To use `history.replace()`, the callers of `loginWithRedirect()` should to set `returnTo` to `window.location.pathname + window.location.search` instead of `window.location.href`.

PR #3258, opened 2021-08-18
Add subscription to add team flow behind a feature flag

ryanjduffy:
A greatly simplified implementation of adding subscription support for teams behind the `teamSubscription` feature flag.

* Adds `plan` variable to `createWorkspace` mutation which is conditionally passed based on the feature flag. When passed, the backend will create the subscription. Further, once the subscription is created, adding and removing members will trigger subscription updates from the backend.
* Adds a placeholder pane for the team settings (the link for it is also behind the feature flag) that displays the subscription key if the team has a subscription.
* Adds a new hook to query the subscription state for the settings pane.

image

### DIFF
--- a/src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
+++ b/src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
@@ -13,6 +13,7 @@ import hooks from "ui/hooks";
 import { Workspace, WorkspaceUser } from "ui/types";
 import { removeUrlParameters } from "ui/utils/environment";
 import { validateEmail } from "ui/utils/helpers";
+import { features } from "ui/utils/prefs";
 import { TextInput } from "../Forms";
 import Modal from "../NewModal";
 import { WorkspaceMembers } from "../WorkspaceSettingsModal/WorkspaceSettingsModal";
@@ -145,7 +146,13 @@ function SlideBody1({ hideModal, setNewWorkspace, setCurrent, total, current }: 
     removeUrlParameters();
     hideModal();
   };
-  const handleSave = () => createNewWorkspace({ variables: { name: inputValue, userId } });
+  const handleSave = () =>
+    createNewWorkspace({
+      variables: {
+        name: inputValue,
+        planKey: features.teamSubscription ? "test-beta-v1" : undefined,
+      },
+    });
 
   return (
     <>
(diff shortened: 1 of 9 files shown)


src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
```diff
@@ -271,6 +278,50 @@ function OnboardingModal(props: PropsFromRedux) {
...
+        },
+      });
+
+      setCurrent(3);
+    } catch (e) {
+      console.error("failed to create workspace and subscription");
+      console.log(e);
+    }
+  };
+
```
ryanjduffy, line 324: I lifted these up from the individual screens because I didn't want to create workspaces and invite members if the user bailed before submitting. The result is a lot of service orchestration that is probably better done on the backend.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSubscription.tsx
```diff
@@ -0,0 +1,43 @@
...
+                variables: {
+                  workspaceId,
+                  seatCount: 1,
+                  subscriptionKey: "team-1",
+                },
+              });
+            }}
+          >
+            Start with 1 Seat
+          </button>
```
ryanjduffy, line 38: This isn't relevant right now. In the future, we'll need something like this to help transition beta teams onto paid plans but it's cruft that we don't need right now.

src/ui/utils/apolloClient.tsx
```diff
@@ -144,7 +144,7 @@ function createHttpLink(token: string | undefined) {
   }
 
   return new HttpLink({
-    uri: "https://api.replay.io/v1/graphql",
+    uri: "http://localhost:8090/v1/graphql",
```
ryanjduffy, line 147: Oops

jasonLaster:
Big picture, I think this PR makes a couple of workflow changes that I think we can punt on for now. However, I could also be naive so i'm curious what you think.

1. We can keep the members UI the same (we will show the billing summary in the billing ui)
2. We can create the team and add seats the same as we do now. There does not need to be a confirm stp
3. Teams can be implicitly added to the beta plan today. New teams created after launching will be added to paid plan with the free-trial.

src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
ryanjduffy: After discussion with @jasonLaster, we decided to pull these back apart and trigger subscription creation with team creation and updating that subscription as members are added and removed.

ryanjduffy:
Updated this to merge in the new the team settings UI. Noting that it's still behind the `features.teamSubscription` feature flag.

cc: @jonbell-lot23

jaril (approved):

Issue #3252, opened 2021-08-17
Only show the React DevTools tab for React apps


hbenl:
for #2389

PR #3260, opened 2021-08-18
Refactor ReactDevToolsPanel and only show it for React apps (#3252)

hbenl:


src/ui/reducers/reactDevTools.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    default: {
+      return state;
+    }
+  }
+}
+
+export const getAnnotations = (state: UIState) => state.reactDevTools.annotations;
+export const hasReactComponents = (state: UIState) =>
+  getAnnotations(state).some(annotation => annotation.message.event === "operations");
```
jasonLaster, line 29: we can memoize this with `reselect`

hbenl: That shouldn't be necessary because this selector is returning a primitive value.

PR #3262, opened 2021-08-18
Redesign the replay invite onboarding flow

jaril:


src/ui/reducers/app.ts
```diff
@@ -22,7 +22,8 @@ function initialAppState(): AppState {
     uploading: null,
     awaitingSourcemaps: false,
     sessionId: null,
-    modal: null,
+    modal: "team-leader-onboarding",
```
jasonLaster, line 25: Is this intentional?

jaril: Nope - for testing

PR #3280, opened 2021-08-19
Fix the console line gap when a message is paused

### LINKED ISSUE #3278: Console Pause line should overlay the top of the console message
Notice how there is a gap between the console message and the pause line

Screen Shot 2021-08-19 at 2 08 55 PM

jaril:
https://user-images.githubusercontent.com/15959269/130156791-0cbaaccd-a55e-4f77-9f13-2cc0cf7c7db4.mov

This fixes #3278.

It also adds back the bottom border on the last console message. We removed the border before because we had the JS terminal right underneath the last message, which would supply that border visually. With the JS terminal pinned to the bottom, we no longer need that compensation.

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -129,14 +129,6 @@
   margin-top: 0px;
 }
 
-/* We already paint a top border on jsterm-input-container (and we need to keep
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):

jasonLaster:
wow - just removing cruft

Issue #3263, opened 2021-08-18
Auth0 redirects back to the sign in page


jasonLaster:
**Description:**

When I sign in to google, auth0 redirects me back to the sign in page until the sign in is complete. we should take users to a loading screen while we wait for auth0.

hbenl:
I assume that you meant this behaviour (which is actually not caused by a redirect): https://app.replay.io/recording/fe030ef7-61e6-4075-8b81-a1b50b1cb879

PR #3282, opened 2021-08-20
Don't create a new ApolloClient every time the ApolloWrapper is rendered (#3263)

hbenl:


jasonLaster (approved):

jasonLaster:
I believe it's also after you sign in to google and it hasn't heard back from auth0, but this is a nice step forward

ryanjduffy:
Nice! :shipit:

Issue #3270, opened 2021-08-19
Screen flickers when switching from Viewer to DevTools


hbenl:
When the user switches from the Viewer to the DevTools for the first time in his current session, the screen turns white while the necessary chunks are loaded:
https://app.replay.io/recording/ab700e3f-f6e7-4a7e-948f-08b8aa9d8af5

PR #3283, opened 2021-08-20
Preload the DevView component before switching to it (#3270)

hbenl:


ryanjduffy:
This is a clever use of the animation time to improve the responsiveness of the app. When I turn on 3G throttling, the app actually seems _less_ responsive because the load time for the JS is longer than 300ms so the UI looks like it hangs. We might need to add some sort of spinner / interstitial to swap out the viewer UI after 300ms if the devtools hasn't loaded yet.

https://app.replay.io/recording/4fc863a6-abf2-43a3-a0c3-20f5caa0fd59

jasonLaster (approved):

jasonLaster:
👍 at some point it would be nice to have some unit tests where for instance we could confirm that the transition occurs after we've imported the DevView

PR #3285, opened 2021-08-20
Add role selection to team settings ui

ryanjduffy:
## Summary

* Adds role ("User", "Developer", "Admin") selection to team members UI
* Defaults new members to "Developer" for now
* Removes "Danger Zone" and invite domain checkbox for non-admin users

https://user-images.githubusercontent.com/788456/130270281-0d905556-7a94-4d77-af32-b61604b271ac.mp4

## Follow Up Work

* We need to add a dropdown to the invite button to allow the user to select the role at invite time. They can immediately change it below but this would be a good usability improvement (and is already included in the wireframes)
* The dropdown in the member list is left aligned with the text and, as a result, jumps a bit when the text changes when changing roles. It should be right aligned and stay anchored below the arrow.
* Maybe API Keys should be restricted to the "Developer" role or greater?


jaril (approved):
LGTM!

PR #3309, opened 2021-08-24
Change the upload screen privacy text

### LINKED ISSUE #3288: Upload Screen: Default privacy should be more obvious
**Description:**

It should be more clear that the replay will be private

Screen Shot 2021-08-22 at 7 07 20 PM

jaril:
Fix #3288.

This changes the privacy summary text that the user sees after recording a replay and they are looking at the upload screen. The text depends on their default workspace:
- Has no default team - `Only you can view this`
- Has a default team - `Shared privately with yourTeamName`

It's worth nothing that while we have some logic for handling the case where the replay is shared publicly, that case is never entered. It's not possible to enter the case where the upload screen summary is shown for a public replay. This is because replays are private by default, and if a user does change their replay to public, they can't go back to displaying the replay summary.

### DIFF
--- a/src/ui/components/UploadScreen/UploadScreen.tsx
+++ b/src/ui/components/UploadScreen/UploadScreen.tsx
@@ -210,14 +210,18 @@ function SharingPreview({
   const workspaceName = [...workspaces, personalWorkspace].find(w => w.id === selectedWorkspaceId)!
     .name;
   let icon;
+  let text;
 
   if (!isPublic) {
     if (selectedWorkspaceId) {
+      text = `Shared privately with ${workspaceName}`;
       icon = "groups";
     } else {
+      text = `Only you can view this`;
       icon = "person";
     }
   } else {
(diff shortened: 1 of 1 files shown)


src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -210,14 +210,18 @@ function SharingPreview({
...
   if (!isPublic) {
     if (selectedWorkspaceId) {
+      text = `Shared privately with ${workspaceName}`;
       icon = "groups";
     } else {
+      text = `Only you can view this`;
       icon = "person";
     }
   } else {
+    text = "This replay can be viewed by anyone with the link";
```
ryanjduffy, line 224: This `text` isn't used anwhere yet. :)

jaril: Whoops!

PR #3311, opened 2021-08-24
Update team settings UI to match user settings

ryanjduffy:
* Generalizes `SettingsModal` so it can be used for both user and team settings
* Migrates team settings to use the new generalized view

image

**Notes**

I tried to preseve the generalization that already existed for settings but it turned out more complex than is warranted for the feature. Since I was already done, I've moved it ahead but if there are major concerns I'm content to redo this with a simpler refactoring. The advantage to this approach is that it will allow some growth like adding DB-managed team preferences or expanding to new settings (per-recording settings, organization settings).


src/ui/components/shared/SettingsModal/SettingsBody.tsx
```diff
@@ -145,61 +37,36 @@ export function SettingsBodyHeader({ children }: { children: React.ReactChild })
...
 
-export default function SettingsBody({ selectedSetting, userSettings }: SettingsBodyProps) {
-  const { title, items } = selectedSetting;
+export default function SettingsBody<
+  T extends string,
+  V extends Record<string, unknown>,
+  P extends Record<string, unknown>
+>({ onChange, panelProps, selectedSetting, values }: SettingsBodyProps<T, V, P>) {
+  const { title } = selectedSetting;
   const [showRefresh, setShowRefresh] = useState(false);
```
jaril, line 46: Is the `showRefresh` mechanism being used anywhere? If not, we can probably take it out.

jaril (approved):
LGTM!

src/ui/components/shared/SettingsModal/SettingsBody.tsx
ryanjduffy: Doesn't look like it. I've cleaned it up!

ryanjduffy:
Mock tests passed locally so going to go ahead with this.

PR #3336, opened 2021-08-26
Make a minimal share screen

### LINKED ISSUE #2863: [UX] Simplify share screen
*The problem statement*

Our share screen is complicated because we show everything at once. We should use concepts of progressive disclosure to show the most important information (copy URL) easily and tuck the other stuff away.

jaril:
Fix #2863.

image
image


jaril:
> Couple minor nits:
> 
> * There are a couple outstanding references in .css files to the `.private-settings` class that are probably cruft and could be removed
> * The email icon displayed when inviting a collaborator is not centered in the circle
> * When you invite a collaborator, it isn't reflected in the counts of "shared with X other person"

Fixed these before merging. Thanks Ryan!

PR #3377, opened 2021-08-29
mark not-yet-ready experiments as coming soon

### LINKED ISSUE #3300: Settings: Update experiments UI
**Description:**

At the moment anyone can sign up for any experiment. We should update it so that folks can toggle react devtools which is close, but the others' checkboxes are replaced with a coming soon label unless you're internal.

jasonLaster:
Fixes #3300

Screen Shot 2021-08-29 at 1 30 22 PM


jaril (approved):

jaril:
Yeah looks good: 
image

PR #3393, opened 2021-08-30
Add a loader for the events sidepanel

### LINKED ISSUE #3321: Handle loading of events in the events sidepanel
Follow up from #3307.

The events sidepanel doesn't currently handle any loading states for any of the events. 

It's possible to open the events sidepanel and only see a partially-loaded list of events. We should convey that a particular group of events (e.g. mouse/keyboard/navigation) is loading with a loading bar, and make sure that the list is updated properly.

jaril:
Fix #3321.

image

### DIFF
--- a/src/ui/reducers/app.ts
+++ b/src/ui/reducers/app.ts
@@ -1,4 +1,4 @@
-import { AppState, EventKind, PanelName, ReplayEvent, ViewMode } from "ui/state/app";
+import { AppState, EventCategory, EventKind, PanelName, ReplayEvent, ViewMode } from "ui/state/app";
 import { AppActions } from "ui/actions/app";
 import { UIState } from "ui/state";
 import { SessionActions } from "ui/actions/session";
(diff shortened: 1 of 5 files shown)


src/ui/reducers/app.ts
```diff
@@ -252,6 +252,30 @@ export const getPointsForHoveredLineNumber = (state: UIState) => {
 const NO_EVENTS: MouseEvent[] = [];
 export const getEventsForType = (state: UIState, type: string) =>
   state.app.events[type] || NO_EVENTS;
+
+// This will return an object with each category (mouse/keyboard/navigation)
+// as the key and a boolean value indicating whether it's still loading
+export const getEventCategoriesLoading = (state: UIState) => {
```
jasonLaster, line 258: This should probably be memoized with reselect

jasonLaster (approved):
This looks good, but in the back on my mind I'm wondering if it would be simpler to just show `Loading Events...` until we have results for all 3 categories

jaril:
> This looks good, but in the back on my mind I'm wondering if it would be simpler to just show `Loading Events...` until we have results for all 3 categories

It is and it isn't. I went with the more detailed approach since the blanket loader didn't look to hot and wasn't very helpful with showing what else might show up once the loading was finished. Planning to see how this feels — can easily adjust afterwards if it's too heavy.

src/ui/reducers/app.ts
jaril: There's not much to optimize here — as far as I've seen, this only gets computed <3 times. As for complexity, the category (e.g. `mouse`) matching to the event kinds (`mousedown`)  doesn't get bad at all, since there's a limited amount of permutations (max `3 * 7 = 21`) every time the selector runs..

Issue #3173, opened 2021-07-30
Consolidate loading screens


jaril:
As Holger mentioned in #3063, we have a lot of loading screens and it's getting harder to track what they do, leading to non-intentional flashes. Let's audit all of the big loading screens and make sure we consolidate them where we can to avoid flashing.

```
I've been thinking about our loading screens because the flashing gets worse with #3135 . I think we should have one loading screen component that we use everywhere and that component should figure out what to show (e.g. if there should be a progress indicator or not) based on the state in redux.
```

jasonLaster:
Adding two other issues here:

Description:
The Loading label should either be hidden or replaced with the indeterminate spinner.

https://app.replay.io/?id=7fff2504-36c1-4a01-b5d6-7843dc30ff61&point=3569704091223727585021232993534151&time=1969&hasFrames=true


This is another loading issue we might be able to fix at the same time

https://app.replay.io/recording/ac04fc44-4e1e-4a07-bb79-132770c9aaa8?point=1947111322887612301547710892212271&time=1468&hasFrames=false

PR #3394, opened 2021-08-30
Refactor app initialization (#3173)

hbenl:
- move some initialization logic out of React effect hooks
- move code that should only be run once out of `createSession()` (preparation for showing another recording without having to reload the app)
- consolidate the loading screens


src/ui/actions/session.ts
```diff
@@ -25,77 +33,176 @@ declare global {
...
-
-  try {
-    ThreadFront.setTest(getTest() || undefined);
-    ThreadFront.recordingId = recordingId;
-
-    if (!isTest()) {
-      tokenManager.addListener(({ token }) => {
-        if (token) {
-          sendMessage("Authentication.setAccessToken", { accessToken: token });
+export function checkRecordingAccessible(
```
ryanjduffy, line 36: Nit: Maybe `getAccessibleRecording` instead since it returns either the recording or null. "check" implies a boolean to me

src/ui/actions/session.ts
```diff
@@ -25,77 +33,176 @@ declare global {
...
+          message: "Invalid ID",
+          content: `"${recordingId}" is not a valid recording ID`,
+        })
+      );
+      return null;
+    }
+
+    try {
+      const recording = await getRecording(recordingId);
+      const userId = await getUserId();
```
ryanjduffy, line 52: Optimization: These could be parallelized.

src/ui/utils/logrocket.ts
```diff
@@ -4,22 +4,21 @@ import * as Sentry from "@sentry/react";
...
     recording,
     userInfo,
-    auth,
+    auth0User,
   }: {
     recording?: Recording;
-    userInfo: UserInfo;
-    auth: AuthContext;
+    userInfo?: Omit<UserInfo, "loading">;
+    auth0User: any;
```
ryanjduffy, line 18: This could use a type but I see that auth0 doesn't type it by default which is kinda lame.

ryanjduffy (approved):
Everything seems to work as expected. I shared a few minor comments for consideration but nothing blocking.

My more general feedback is that I think we should strive for more focused PRs whenever possible. The number of line changes isn't terribly high but the changes span many files and involve moving some things about so it's hard to track it. The risk here is that since it's difficult to review it's also difficult to find bugs. :)

I'm not sure what the right breakdown could have been in this particular case but something to keep in mind.

hbenl:
> I think we should strive for more focused PRs whenever possible

I agree, I just couldn't find a good way to split it up in this case.

Issue #3339, opened 2021-08-26
Add a comment in a bp fails


jasonLaster:
The Add comment button does not show a comment editor

https://app.replay.io/recording/e10e2e14-3f7e-4094-9ad7-58ac6a72dec4?point=141814607952702021779324188728756692&time=47864&hasFrames=false

hbenl:
`getFrames()` doesn't fail, it waits for `createPause()` to return, which never happens. Seek to the end of the recording and evaluate `outstanding()` in the replay console and you'll see that 10 `createPause()` requests never return. The backend doesn't send errors for these requests, so the frontend can't tell if these requests will still return later.
When I open the underlying recording (https://app.replay.io/recording/acec9389-27bc-45fb-9537-f617e6a68915) and try to reproduce the issue, I see a lot of hanging requests and eventually get an "Unexpected session error".

PR #3422, opened 2021-09-02
Comments: use the breakpoint's location if available and add a timeout for fetching the current location (#3339)

hbenl:


jasonLaster:
@hbenl mind including a replay so i can see what's going on

jaril (approved):

PR #3436, opened 2021-09-02
Add logo to loading screen

jonbell-lot23:
image

Ultimately we want to use Rive to animate this screen. But in the meantime this is a nice step forward and low hanging fruit.

More visual improvements to come!


jasonLaster:
- I like dropping the labels
- I'd vote for getting rid of the spinner which is pretty ugly in my opinion and just showing the logo

So in my flow it would be, logo followed by logo + progress bar. And eventually logo followed by animated logo :)

jasonLaster:
Here's where i landed
https://github.com/RecordReplay/devtools/pull/3444

Note, the labels are only used for uploading

Issue #3379, opened 2021-08-29
Graphics do not update on comment click


jasonLaster:
**Description:**
https://app.replay.io/recording/e18c5ee9-002e-4b87-9dff-a28592929c2d?point=52572005696168575725602854615909919&time=18601&hasFrames=false

jaril:
![image](https://user-images.githubusercontent.com/15959269/131332595-f84e5f83-2dec-43ef-8f66-b41fdaa5f91c.png)

hbenl:
Here's what happens in that recording:
- the user clicks on the first comment
- devtools starts to download the source for that comment's location, which is more than 2.5MB large
- the user clicks on the second comment
- devtools requests the screenshot, but the backend is still sending the message containing the source requested earlier, so the WebSocket downstream is blocked
- after some 5 seconds, the source download is finished, unblocking the WebSocket
- shortly thereafter devtools receives the screenshot, but by now the user has given up and seeked somewhere else

On slow connections, this (a large source download blocking the WebSocket) can happen a lot. One idea to solve it would be to download the sources in chunks, so that other requests can be served in between the chunks.

loganfsmyth:
Can you clarify what you see in the recording Jason linked that shows the 5 second delay? I'm not disagreeing but I'm also not sure what to look for. I threw some breakpoints in around getPaintContents and getSourceContents doesn't seem to block the response?

<img width="1424" alt="Screen Shot 2021-09-02 at 10 39 37 AM" src="https://user-images.githubusercontent.com/132260/131897305-c31178b6-6233-4d70-8c2d-ef5ba271e39d.png">

hbenl:
I used the following logpoints in `ui/actions/timeline.ts` to show when screenshots are requested and received and also turned on the timestamps in the console:

![Screenshot_2021-09-03_17-02-57](https://user-images.githubusercontent.com/16060807/132026889-d91f3a89-d856-4f6f-ae8f-dee8c5bece4d.png)

- at 0:09 the source is requested after the user clicked on the first comment marker in the timeline
- then 2 screenshots that were requested previously are received
- at 0:12 2 more screenshots are requested after the user clicked on the second comment marker in the timeline (this should be only one screenshot - there is a bug where clicking on a comment marker in the timeline seeks to 2 execution points)
- at 0:15 and 0:16  three more screenshots are requested
- at 0:16 the source is received
- at 0:18 the five screenshots that were requested while the source was loading are received

[...]

PR #3443, opened 2021-09-03
Fix double seek when clicking on a comment marker in the timeline

hbenl:
While looking at #3379 I noticed that a click on a comment marker in the timeline generates two calls to `seek()` (one from the `Timeline` component and one from the `CommentMarker` component). The reason is that the check if the mouse is hovering over a marker was outdated, so I fixed it.


jasonLaster:
mind including a replay so i can see what you mean?

jaril (approved):
Thanks!

PR #3447, opened 2021-09-03
Radically simplify the breakpoint editor

jasonLaster:

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
@@ -18,10 +18,6 @@
   overflow: hidden;
 }
 
-.breakpoint-panel.editing {
-  padding: 8px;
-}
-
 .breakpoint-panel button {
   height: 100%;
   border-radius: 4px;
@@ -56,6 +52,7 @@
   display: flex;
   flex-direction: row;
   overflow: hidden;
+  flex-grow: 1;
 }
 
 .breakpoint-panel .summary .action {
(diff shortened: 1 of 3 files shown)


jasonLaster:
Landing for a couple of reasons

1. this dramatically simplifies editing print statements 
2. the simplified version is much better for the landing page animations
3. we can always revert / tweak over the next couple of days

jaril:
Product discussion:
```
Jaril — Today at 4:23 PM
Always cautious when we're doing something we've done before, when it was intentionally changed away from that. This simple input UI (bare input with no submit button) is similar to the logpoints v1
That was rehauled because of two main concerns: users 1) didn't know when their change took effect, and 2) didn't know how they should format the logs inside the input (e.g. not realizing the input value is being passed into console.log())
So we added context cues (log(yourExpressionHere)) and an explicit save button. As best as I could tell, those resulted in less questions around the widget UI
I dig omitting conditional expressions, but this pulls both of those changes back. Are those two concerns not valid anymore? Or am I missing something?
(Worth nothing that those concerns didn't completely disappear — I still occasionally see somebody not understand what to put in the text input. But I'm not sure that the simplified direction makes those cases any better)

Jon Bell (he/they) — Today at 5:27 PM
This is awesome, thank you! Here are some thoughts:

1. If there were some objectively true way to know what was confusing to people, it'd be super easy. A/B test two things, run tests, do the one that won. In this case we are indeed light on research and overly biased by our own opinions. That's of course common at our size. So could this change be a step backwards? 100%. We don't think so, but it's absolutely possible.

2. A lot of design is like a walk through the woods. You set off in a direction with the best information you have, but then as you go on your journey, new insights come to light that cause adjustments. Like  "Oh, the bridge on the map is out. I guess we'll hike over here to this other bridge."

In our case, we definitely thought the conditional stuff was more important six months ago than we do now. (We are intending to bring it back, just not necessarily a P1 task before launch.)

3. When we added the paywall "cliff," that changed the calculus. We needed some way to say "You can't edit this, but usually you can." The previous design was way too subtle for something editable and novel. It needed a much clearer signal to say THIS IS A FORM ELEMENT. So it got a border, and then the paywall versus got the dimmed look. (With a padlock)

4. We made a calculated gamble that de-emphasising usability (removing save/cancel buttons) would be ok because a) you're not writing a novel in this text field, you're writing a quick note or expression. b) Hitting enter after typing "Test test" is a pretty natural thing, so we think people will be fine with it. Especially developers. c) We get a huge help from the red outline for invalid expressions, too. 

5. Another big shift is the website story. We now have a website shining a spotlight on this very feature, and comms supporting it as well. This is huge, because it will mean people coming to the feature won't be starting from scratch.
I didn't even get to all the issues, if you can believe it. But that's a good summary :slight_smile:
```

PR #3458, opened 2021-09-06
Don't reload the app after login

hbenl:
Reloading the app after login caused the loading screen to flicker and added an unnecessary delay.


jasonLaster (approved):

jasonLaster:
This makes me nervous before launch, but we can always revert if we see weird behavior

PR #3459, opened 2021-09-06
Play button revision

jonbell-lot23:
Current state:
image
image
image

Proposed:
image
image
image

Note that the refresh button is just the play button. I prefer this, actually. No reason to have an icon that says "You're at the end, start over" when the time and the progress bar "you're at the end" and the icon just needs to express "play."

This was a happy accident -- I did a lot of explorations with a refresh button and they just don't feel as good. We need to fit the arrow into a very small space, then maintain the width of two lines. By the time you do that, the icon feels cramped.


jasonLaster:
👍 Looks great

Issue #3448, opened 2021-09-03
Cmd+F shouldn't open editor search when editor is not focused


loganfsmyth:
Twice in the last 2 days I've tried to use Cmd+F to find a particular console message, because I had a breakpoint on before an action was performed and a breakpoint after, and wanted to find the start message corresponding to the end message so. Unfortunately the editor captures the keypress and opens the editor's search box even through I wasn't focused on the editor, I was focused on the console messages.

PR #3464, opened 2021-09-07
Only use debugger shortcuts when the debugger is focused (#3448)

hbenl:
The `KeyShortcuts` implementation used in the debugger expects the DOM node that it should attach to as a constructor argument. At the same time, the `KeyShortcuts` instance is shared with the child components, so it needs to exist before the debugger is rendered. To solve this chicken-and-egg problem, I pulled the outermost DOM node from the `Debugger` component to the `DebuggerLoader` component and pass it to `Debugger` as a prop.

### DIFF
--- a/src/devtools/client/debugger/src/components/App.js
+++ b/src/devtools/client/debugger/src/components/App.js
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-import React, { Component, useEffect, useState } from "react";
+import React, { Component, useEffect, useState, useRef } from "react";
 import PropTypes from "prop-types";
 
 import { connect } from "../utils/connect";
@@ -23,7 +23,8 @@ import { getSelectedPanel } from "ui/reducers/app.ts";
 
 import { KeyShortcuts } from "devtools-modules";
 import Services from "devtools-services";
-const shortcuts = new KeyShortcuts({ window, target: document });
+
+const globalShortcuts = new KeyShortcuts({ window, target: document });
 
 const { appinfo } = Services;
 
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/App.js
```diff
@@ -64,30 +63,36 @@ class Debugger extends Component {
...
 
+  constructor(props) {
+    super(props);
+    this.shortcuts = new KeyShortcuts({ window, target: props.wrapper });
+  }
+
   componentDidMount() {
-    shortcuts.on("CmdOrCtrl+Shift+P", this.toggleSourceQuickOpenModal);
-    shortcuts.on("CmdOrCtrl+Shift+O", this.toggleFunctionQuickOpenModal);
-    shortcuts.on("CmdOrCtrl+P", this.toggleSourceQuickOpenModal);
```
jasonLaster, line 73: It would be nice if `cmd+p`, `cmd+o`, `cmd+shift+p` and `cmd+shift+o` remained global. perhaps others as well

src/devtools/client/debugger/src/components/App.js
```diff
@@ -64,30 +63,36 @@ class Debugger extends Component {
...
+    this.shortcuts.on("CmdOrCtrl+Shift+P", this.toggleSourceQuickOpenModal);
+    this.shortcuts.on("CmdOrCtrl+Shift+O", this.toggleFunctionQuickOpenModal);
+    this.shortcuts.on("CmdOrCtrl+P", this.toggleSourceQuickOpenModal);
+    this.shortcuts.on("CmdOrCtrl+O", this.toggleProjectFunctionQuickOpenModal);
+    this.shortcuts.on("Ctrl+G", this.toggleLineQuickOpenModal);
 
-    shortcuts.on("Escape", this.onEscape);
-    shortcuts.on("Cmd+/", this.onCommandSlash);
+    this.shortcuts.on("Escape", this.onEscape);
+    this.shortcuts.on("Cmd+/", this.onCommandSlash);
```
jasonLaster, line 82: cmd+/ should remain global

src/devtools/client/debugger/src/components/App.js
```diff
@@ -64,30 +63,36 @@ class Debugger extends Component {
   };
 
   getChildContext = () => {
-    return { shortcuts, l10n: L10N };
+    return { shortcuts: this.shortcuts, l10n: L10N };
   };
 
+  constructor(props) {
+    super(props);
+    this.shortcuts = new KeyShortcuts({ window, target: props.wrapper });
```
jasonLaster, line 71: guess we need two shortcuts helpers. One for global, one for the wrapper

jasonLaster (approved):

hbenl:
@jasonLaster I updated the PR to make some shortcuts global again

jasonLaster (approved):
Looks good!

Issue #3370, opened 2021-08-29
Reenable limited markdown support for comments


jasonLaster:
**Description:**

We should support the simplest form of markdown where atleast formatting is preserved

### editing
<img width="236" alt="Screen Shot 2021-08-29 at 12 34 30 PM" src="https://user-images.githubusercontent.com/254562/131258113-89950e18-4756-425b-b11e-26509180d9e1.png">

### rendered
<img width="234" alt="Screen Shot 2021-08-29 at 12 35 15 PM" src="https://user-images.githubusercontent.com/254562/131258112-70ac9249-9bed-4cab-8033-a2b99c6c89d8.png">

hbenl:
Did we ever have markdown support in comments? Otherwise I'd suggest simply wrapping the content in [`<ReactMarkdown>`](https://www.npmjs.com/package/react-markdown).

PR #3465, opened 2021-09-07
Enable limited markdown support for comments (#3370)

hbenl:
This PR only enables the markdown-to-html conversion, but most markdown features don't work yet because of the css reset by tailwind. For example, `# Header` is converted to `<h1>Header</h1>`, but tailwind resets the default styling for `<h1>`, so this looks like regular text.


jasonLaster (approved):
Looks good. I'm glad that we strip all the stying for H1s

PR #3485, opened 2021-09-08
Redact recording view for logrocket

ryanjduffy:
## Summary

I tried to disable LogRocket entirely for the recording view but any changes I made (e.g. disabling DOM) mid-session removed all events for the entire session (even those that already happened). Instead, I took a broad stroke approach to redacting most of the recording app views so we're not capturing anything other than pure UI elements (e.g. the toolbar).

This includes an update to the `Redacted` component to support allowing capturing data for the element via a new optional `allowOptIn` prop. Also, `Redacted` now pulls a user setting, `enableLogRocket`, to work with `allowOptIn`

Examples:
* `<Redacted>some text</Redacted>` - content is never included in log rocket regardless of the user setting
* `<input data-private value={somePrivateValue} />` - content is never included in log rocket regardless of the user setting
* `<Redacted allowOptIn>some text</Redacted>` - content is included in log rocket if the user has enabled `enableLogRocket`

Since we do not have the rest of the infrastructure in place to support the new user setting, this PR attempts to get the value from graphql but will fall back to a feature flag if the setting doesn't exist.


src/ui/components/Video.tsx
```diff
@@ -58,7 +59,7 @@ function Video({
   const showCommentTool = isPaused && !isNodeTarget && isAuthenticated && !isNodePickerActive;
 
   return (
-    <div id="video">
+    <Redacted allowOptIn id="video">
```
jasonLaster, line 62: this is already redacted by virtue of the fact that LogRocket does not pick up canvas or video. And it would be nice to include the comment markers.

jasonLaster (approved):
One small tweak, but otherwise ready to go. Thanks!

Issue #3452, opened 2021-09-04
Replay did not change teams


jasonLaster:
From @jazzdan 

I changed the workspace of a recording from My Library to Replay, but no one from Replay can view it. Here's a recording of me changing the workspace of the Replay: https://app.replay.io/recording/23fafa1d-03a7-4a09-9cb5-0c4cfb5b651f

Here's the original Replay: https://app.replay.io/recording/fa7f279e-20e9-4dff-90cd-b5ec23b82519

PR #3493, opened 2021-09-08
Fix changing a recording's teams using the sharing modal (#3452)

hbenl:
This fixes 2 issues in the sharing modal:
- it didn't update a recording's workspace
- it assumed that the current recording's workspace was the user's default workspace, but that isn't always the case


jaril (approved):

jaril:
The behavior here is such that the team change is only applied once the user hits the `Done` button. 

That's probably the right thing to do. Let's ship and revise in the off-chance we want to go off in another direction.

PR #3497, opened 2021-09-08
Add UI feature guards for team users

ryanjduffy:
## Summary

* Adds `roles` to the `Recording` type to represent the possible roles a user might have for the given recording
* Adds a wrapper around the console input to show the locked message for "team-user" roles
* Prevents editing breakpoints for "team-user" roles
* Updates the styling of the breakpoint editor to match https://github.com/RecordReplay/devtools/issues/3408#issuecomment-909914767

image

### DIFF
--- a/src/ui/hooks/recordings.ts
+++ b/src/ui/hooks/recordings.ts
@@ -27,6 +27,7 @@ const GET_WORKSPACE_RECORDINGS = gql`
               createdAt
               private
               isInitialized
+              userRole
               comments {
                 user {
                   id
@@ -63,6 +64,7 @@ const GET_MY_RECORDINGS = gql`
             createdAt
             private
             isInitialized
+            userRole
             owner {
               id
               name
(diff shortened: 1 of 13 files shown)


ryanjduffy:
V2: Show the eval input but display the lock message when the user submits the eval.

lock

ryanjduffy:
v3: show the lock message as the "result" of console evals

https://user-images.githubusercontent.com/788456/132898485-49559f81-d01a-4f5c-8957-ddf474ceefcc.mp4

src/ui/hooks/recordings.ts
```diff
@@ -127,7 +129,7 @@ export async function getRecording(recordingId: RecordingId) {
 }
 
 export function useGetRecording(
-  recordingId: RecordingId | null
+  recordingId: RecordingId | null | undefined
```
jaril, line 132: Why is the typing relaxed here?

jaril (approved):

src/ui/hooks/recordings.ts
ryanjduffy: `getRecordingId` returns a `string` or `undefined` so I changed this to support either `null` or `undefined` since, afaict, there isn't a material difference for the GQL to accept either null-ish value.

ryanjduffy: Alternatively, we could change `getRecordingId` to return `null` instead of `undefined`. /shrug

PR #3499, opened 2021-09-08
Use the animated replay logo spinner in place of static logo

jaril:
image


jasonLaster:
@jaril I think your fun spinner is more entertaining!

PR #3502, opened 2021-09-08
Polish billing UI

ryanjduffy:
Cleans up the billing UI and removes the feature flag!

image
image
image


jonbell-lot23:
Looking groovy, thanks!

From a design standpoint, this is good to go. There are a few polish things that I'd like to add but they're better as a follow-up with me dinking around in the UI in my own branch. Design POV: ship it

jasonLaster (approved):

Issue #3470, opened 2021-09-07
Add /teams/<team name>/settings/billing route


jasonLaster:
**Description:**
There should be a URL that opens the team billing modal

PR #3510, opened 2021-09-09
Add team and settings routes (#3470)

hbenl:
Adds routes `/team/:teamID` and `/team/:teamID/settings/:tab`.
TODO: 
- the `tab` in the settings route now uses the tab's title, but that should be an ID instead because the title may contain characters that are not safe to put in a URL
- the displayed modal is still stored in redux (and the settings route triggers a corresponding update to the redux store) - it would be nice to take this out of redux and have all modals triggered by routes


jasonLaster:
stale

PR #3511, opened 2021-09-09
Add billing route (#3470)

hbenl:
Adds a `/team/:workspaceId/settings/billing` route. `workspaceId` needs to be the GraphQL ID of a workspace.

### DIFF
--- a/src/ui/setup/index.ts
+++ b/src/ui/setup/index.ts
@@ -1,10 +1,11 @@
+import { matchPath } from "react-router-dom";
 import { bootstrapStore } from "./store";
 import { registerStoreObserver, updatePrefs } from "./prefs";
 import { setupAppHelper } from "./helpers";
 import { setupDOMHelpers } from "./dom";
 import { setTelemetryContext, setupTelemetry } from "ui/utils/telemetry";
 import { UIStore } from "ui/actions";
-import { getTheme } from "ui/reducers/app";
+import { getTheme, getWorkspaceId } from "ui/reducers/app";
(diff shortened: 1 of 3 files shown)


hbenl:
To test this:
- find the ID of a workspace that you have access to (e.g. by setting a logpoint `node.id, node.name` [here](https://github.com/RecordReplay/devtools/blob/16ce139ea7a461615e4dff44f60ea77d37010545/src/ui/hooks/workspaces.ts#L131)
- go to `https://devtools-git-fork-hbenl-issue-3470-recordreplay.vercel.app/` and login
- go to `https://devtools-git-fork-hbenl-issue-3470-recordreplay.vercel.app/team/<workspaceID>/settings/billing`

src/ui/setup/index.ts
```diff
@@ -59,8 +67,10 @@ export function bootstrapApp() {
       setTelemetryContext(userInfo);
     }
 
-    const userSettings = await getUserSettings();
-    store.dispatch(setWorkspaceId(userSettings.defaultWorkspaceId));
+    if (!getWorkspaceId(store.getState())) {
```
jaril, line 70: Is this just so we don't bother querying+updating the workspaceId if they're going through the billing path?

jaril (approved):
LGTM!

src/ui/setup/index.ts
hbenl: No, it's because in this case we already set the `workspaceId` (to the one from the route) and don't want to change that to the user's default `workspaceId`.

PR #3528, opened 2021-09-10
Switch to static.replay.io

jasonLaster:
Test Plan:

1. tests should pass
2. try downloading a browser from the library

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -31,7 +31,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Download
-        run: wget https://replay.io/downloads/replay.dmg
+        run: wget https://static.replay.io/downloads/replay.dmg
       - name: Create artifact
         uses: actions/upload-artifact@v2
         with:
@@ -43,7 +43,7 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: Download
-        run: wget https://replay.io/downloads/macOS-replay-node
+        run: wget https://static.replay.io/downloads/macOS-replay-node
       - name: Create artifact
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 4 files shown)


jasonLaster:
hmm - @jazzdan seeing a bunch of failures (https://github.com/RecordReplay/devtools/actions/runs/1219577009)

jazzdan:
Yeah it's weird. I compared a failing test run and a passing test run's Test Stripe 0. The failing run produces a recording ID that a) [can be viewed ](https://app.replay.io/recording/265da87a-2fcb-48ae-a368-5ee71bdde837) and b) [looks identical to a passing one](https://app.replay.io/recording/9a62670a-39d9-4ac9-b1d9-2d662aae2ebe).

One difference that does jump out is a passing run has [SavedRecording](https://github.com/RecordReplay/devtools/runs/3565748633#step:4:289) but a failing run only has the subsequent [FinishedRecording](https://github.com/RecordReplay/devtools/pull/3528/checks?check_run_id=3562876214#step:4:279).

I would have expected the "Download replay" steps to fail, if anything. But those all succeeded.

jazzdan:
I think this means that the replay browser is firing the "finished" event, but not the "saved" event?

https://github.com/RecordReplay/gecko-dev/blob/d3c242bb7df5e41ad3c778aa0bae0bd32ff30165/devtools/startup/DevToolsStartup.jsm#L1896-L1917

jazzdan:
Or maybe it's hanging somewhere while processing the saved event

jazzdan:
I think that's as far as I can go in a reasonable amount of time. @bhackett1024 @kannanvijayan any ideas?

bhackett1024:
It looks to me like the webpack build broke, though I have no idea why:

```
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /dist/main.js</pre>
</body>
</html>
```

jazzdan:
@bhackett1024 hmm that line is present in a passing test too though https://github.com/RecordReplay/devtools/runs/3565748633#step:4:243

jasonLaster:
just confirmed that the download page works

jasonLaster:
okay - the tests just passed without the change

Issue #3428, opened 2021-09-02
Disable global symbol search


jasonLaster:
This search uses a ton of backend resources to fetch all files.

We should wrap `cmd+o` global function search in a db-feature flag similar to the elements panel and show it in the Experiments modal.

PR #3535, opened 2021-09-10
Add user setting for enabling global symbol search (#3428)

hbenl:
To be tested after RecordReplay/backend#2963 lands.


jasonLaster (approved):
Thanks!

PR #3537, opened 2021-09-10
Update the batch edit dropdown button in the library

### LINKED ISSUE #3477: Update the Batch Action dropdown in the library
This is functional for now, but TailwindUI-ify this and make it less ugly.

jaril:
Fixes #3477.

This updates the button for making bulk edits to the user's replay. It largely follows the same implementation as the recording dropdown in #3514.

## Demo:
https://user-images.githubusercontent.com/15959269/132892986-508065f7-4fe4-441e-8e31-fe84df80a7cf.mov

## Screenshots:
image
image

## Follow up
- [x] https://github.com/RecordReplay/devtools/issues/3538

### DIFF
--- a/src/ui/components/shared/MaterialIcon.tsx
+++ b/src/ui/components/shared/MaterialIcon.tsx
@@ -10,6 +10,8 @@ type MaterialIconProps = PropsFromRedux &
     children: string;
     highlighted?: boolean;
     outlined?: boolean;
+    // tailwind text color style, e.g. text-white, text-blue-200
+    color?: string;
   };
 
 function MaterialIcon({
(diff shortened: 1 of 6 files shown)


src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +20,7 @@ function MaterialIcon({
   highlighted,
   className,
   outlined,
+  color,
```
jaril, line 23: As a follow up, we should refactor this component to be less opinionated about color so it can just inherit the parent's.

jaril: #3538.

PR #3540, opened 2021-09-10
Fix the sign in button size

### LINKED ISSUE #3529: Sign in button is the wrong size
image

jaril:
Fix #3529.

This uses the standardized button as the sign in button.

@jonbell-lot23 heads up — even after the change, the header elements (specifically viewtoggle and sign in button) aren't aligned with each others borders. Not sure if you'd prefer making the toggle larger, or button smaller. Let me know.

image


jonbell-lot23:
@jaril Ideally we'd make the button smaller.

But I just checked it out in the browser and I don't mind it mismatched like this. It brings a bit of prominence to the CTA. We'll address later.

PR #3546, opened 2021-09-10
Update the launch replay modal

### LINKED ISSUE #3530: Update the Download modal
Couple of small tweaks

- proper close button
- Use the same secondary font size
- Should include the Mac and Download links instead of linking to /welcome

Screen Shot 2021-09-09 at 7 39 22 PM

jaril:
Fix #3530.

image

### DIFF
--- a/src/ui/components/shared/LaunchBrowserModal.tsx
+++ b/src/ui/components/shared/LaunchBrowserModal.tsx
@@ -1,8 +1,9 @@
+import classNames from "classnames";
 import React, { useEffect } from "react";
 import { connect } from "react-redux";
 import { actions } from "ui/actions";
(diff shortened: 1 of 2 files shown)


jaril:
@jasonLaster is this what you had in mind?

jasonLaster:
It's arguably better, I'd see if you could borrow the same component used for onboarding which has windows greyed out.

some small tweaks for this design
- alignment/padding could be better more space after divider, less below buttons
- Download for can be dropped with the right design

With that said, we might not want buttons at all if we assume most folks have Replay installed. In that case simple links for mac, linux and text for windows (coming soon) could be better since we are de-emphasizing the install step.

PR #3590, opened 2021-09-13
Add feature flag for new user invitations

ryanjduffy:
## Summary

* Adds LaunchDarkly sdk
* Adds `new-user-invitations` feature flag to:
  * Guarding viewing a recording from an uninvited author which, when the flag is active, will treat the author as always invited
  * Displaying the Invitations tab (hidden when the flag is active)


jasonLaster (approved):
Looks good

PR #3591, opened 2021-09-14
Add feature flag to disable creating teams

ryanjduffy:
## Summary

* Adds support for `create-workspace-enabled` flag to disable creating a team from:
  * the library
  * the team leader onboarding flow

image
image

### DIFF
--- a/src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
+++ b/src/ui/components/shared/NewWorkspaceModal/NewWorkspaceModal.tsx
@@ -5,6 +5,7 @@ import React, {
   MouseEventHandler,
   SetStateAction,
   useEffect,
+  useMemo,
   useRef,
   useState,
 } from "react";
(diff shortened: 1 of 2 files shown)


jonbell-lot23:
Hey @ryanjduffy, one addition: instead of "Ok, sounds good" let's change the button to "Join the waitlist" and target our waitlist. Thanks!

ryanjduffy:
Do you want to have an opt-out button as well? I think on the onboarding modal, we need some way to continue the flow. In the create a team modal, we could possibly get away without another button because the user can click out to close the modal.

jonbell-lot23:
Ah, good point. For the onboarding flow it should be

[Join Waiting List] [Skip]

Skip is easy.
Join waiting list should should load the form, which will be an overlay from the right side. Then when you're done with that, it'd be nice if it returned you to the next step in the onboarding. Is that doable?

ryanjduffy:
We'll see! At a minimum, we can open a new tab with the form and then the user could return to continue.

PR #3575, opened 2021-09-13
Make the browser user flows consistent with the onboarding flows

jaril:
Fix #3383. Fix #3564. Fix #3567.

image
image
image

PR #3592, opened 2021-09-14
Fix CSS variable file import

ryanjduffy:
Regression from #3575 which moved an import but grabbed the wrong file


jaril (approved):
Thanks!

PR #3608, opened 2021-09-15
Demo is hosted at static.replay.io now

jazzdan:


jaril:
Fixed with https://github.com/RecordReplay/devtools/commit/00113f62ba8f585c4565865fe5158e56245268f1

PR #3621, opened 2021-09-15
Router missing JS files to ensure they report a 404

ryanjduffy:
Redirects requests for JS files that do not exist to a path we know to not exist so it reports as a 404 rather than our root index.html


loganfsmyth (approved):
Definitely curious to see if this changes anything and worried it won't since the 404 page is still HTML, but it can't hurt to try!

PR #3627, opened 2021-09-16
Add missing Babel plugins to satisfy more target environments

Andarist:
fixes https://github.com/RecordReplay/devtools/issues/3619

Note that I have not tested this, but conceptually this was a straightforward thing to do so hopefully I've not missed anything about this situation.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -81,6 +81,8 @@ module.exports = {
               "@babel/plugin-transform-flow-strip-types",
               "@babel/plugin-transform-react-display-name",
               "@babel/plugin-proposal-class-properties",
+              "@babel/plugin-proposal-logical-assignment-operators",
+              "@babel/plugin-proposal-nullish-coalescing-operator",
               "@babel/plugin-proposal-optional-chaining",
             ],
           },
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -81,6 +81,8 @@ module.exports = {
               "@babel/plugin-transform-flow-strip-types",
               "@babel/plugin-transform-react-display-name",
               "@babel/plugin-proposal-class-properties",
+              "@babel/plugin-proposal-logical-assignment-operators",
+              "@babel/plugin-proposal-nullish-coalescing-operator",
```
Andarist, line 85: as a bonus, I've also added this plugin since it's like a sibling feature to the optional chaining (were introduced at the same time) and you had that one already here

hbenl:
Thanks!

webpack.config.js
hbenl: Yeah, that sounds like a good idea.

Qix-: Yes, the nullish coalescing operator would have been the next syntax error :) good catch!

PR #3643, opened 2021-09-17
Library fixes

jonbell-lot23:
Old version with too much whitespace:
image

New version:
image


jaril:
Tried my best to reduce whitespace (horizontal).

Before | After
-- | --
image | image

One thing to note is I tried to make the first and last columns narrower (checkbox, dropdown), and it looked too tight for my taste with ~4px of horizontal padding instead of the original 16px. I opted to keep the 16px horizontal padding for both columns.
x | 4px padding | 16px padding
-- | -- | --
checkbox | 
image | 
image
dropdown | 
image | 
image

PR #3646, opened 2021-09-17
Re-enable browser context menu support

ryanjduffy:
This was added by https://github.com/recordReplay/devtools/commit/452ea7fd but I don't think it's necessary so I'm proposing we remove it so native right-click behavior works.

### DIFF
--- a/src/ui/setup/dom.ts
+++ b/src/ui/setup/dom.ts
@@ -7,8 +7,6 @@ declare global {
 }
 
 export function setupDOMHelpers() {
-  document.body.addEventListener("contextmenu", e => e.preventDefault());
-
   // Set the current mouse position on the window. This is used in places where
   // testing element.matches(":hover") does not work right for some reason.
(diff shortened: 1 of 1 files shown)


ryanjduffy:
cc: @bhackett1024 on this since he was the author of the original commit referenced above.

jasonLaster (approved):
my hunch is that we have some other right click contextmenus and we dont want to show too...

jaril:
@ryanjduffy just bumping this up to keep it from going stale. Are there still any blockers on this one?

ryanjduffy:
Just confidence that it won't break anything :) I'm not sure of all the right click actions in the app off hand to test it thoroughly. We can give it a go and see what happens 😬

jaril:
I only know of 4: source files, source tabs, the debugger editor and outline functions. They're all behaving similarly with the patch.

There's an unrelated lint error, but I think we're good to.merge 👍

PR #3649, opened 2021-09-17
Add a nag to the tooltip when a user hovers on hits

### LINKED ISSUE #3585: Add Breakpoint Gutter Tooltip
**Description:**

We should add a nag for a tooltip that shows up on hover in the gutter. 

*only on the first run

Screen Shot 2021-09-13 at 2 01 27 PM

jaril:
Fix #3585.

This nag stays on the line number tooltip until the user clicks on a line number, and opens up the breakpoint widget.

image

Replay: https://app.replay.io/recording/4c455dbc-aed2-416c-a135-5a95154e9117
Demo: https://share.descript.com/view/heS5O4Pe6n5


jasonLaster:
@jaril could you add a replay for this?

ryanjduffy (approved):
:+1:

Suggestions for future work:
* Alignment of the nag seems a bit off. It's not vertically centered with the line nor is it aligned with the top or bottom edge of the breakpoint.
* The hit count is a bit lost for me. Would be good to give it some more visual emphasis.

ryanjduffy:
Forgot to mention I also noticed a flash of the tooltip if you get the nag and then set the breakpoint. Would be nice to eliminate that.

PR #3654, opened 2021-09-18
Library column polish

jonbell-lot23:
A follow-up on this: https://github.com/RecordReplay/devtools/pull/3643/commits/ebb55155db34271ef1306dfc0cb0f39406658af8

Previously:
image

Now:
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,7 +7,7 @@
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet" />
     <link rel="preconnect" href="https://fonts.googleapis.com" />
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&display=swap" rel="stylesheet">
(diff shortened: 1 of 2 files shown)


jasonLaster:
Went back to tailwind for another reason and compared their stacked list with ours.

- wondering about using color as opposed to bold for the title
- wondering about the different color for the url
- do we want more vertical padding?
- could we use inter light in the footer?

Screen Shot 2021-09-18 at 5 47 37 AM

jonbell-lot23:
**Color vs bold**
Our accent colour is blue, which looks like a link. We don't use accent colour anywhere in our visual language, so if we moved to that it would need to be part of a broader rethink of our design language.

**Different colour for URL**  
Not necessary, removed.

**More vertical padding?**  
A worthy exploration, but recommend another thread for that. This is a more short term fix.

**Inter light for second row**  
Added it to get the contrast we're looking for, and removed the bold header. We don't use light anywhere else in our UI, but it might be appropriate. To explore more.

**Avatars**  
They're pretty, but it doesn't solve enough of a need to offset the for the usability and IA issues it introduces.

Here's how it looks without bold, no color in header, and Inter light added to second row:
image

Looks good.

jasonLaster:
Yep - light is good. 

Thanks for the explanation on color. With you on padding, avatars, ...

jasonLaster (approved):

PR #3656, opened 2021-09-18
Polishing Events dropdown

jonbell-lot23:
Old:
image

New:
image

Doesn't that look nicer? Yay!

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.css
@@ -34,6 +34,7 @@
 .event-listener-header {
   display: flex;
   align-items: center;
+  font-size: 12px;
 }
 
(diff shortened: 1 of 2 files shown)


jasonLaster:
Here's where I stand
- [ ] Spacing
- [ ] Typography
- [ ] Badges

I feel like the badges feel a big heavy, i'm wondering what they're adding

jonbell-lot23:
Numbers right-aligned look like they're floating in space, so creating a vertical line for the eye to track gives it a better sense of place. Here's how Apple does this treatment in their standard controls:

image

And Microsoft attempts the "floating in space" effect, which I don't think feels as solid:

image

I dialled ours back so we have our visual alignment without being as heavy:

image

jasonLaster:
Thanks. That's a great explanation. Looks like there are some prettier lint issues but otherwise great.

jasonLaster (approved):

PR #3689, opened 2021-09-21
Handle stripe loading failure more gracefully

ryanjduffy:
## Summary

Displays an error message when unable to add a new payment method because the stripe library fails to load

## Notes

I tried to create a replay of this but it seems gecko doesn't respect resource blocking when recording! Creating an issue for that too.

### DIFF
--- a/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSubscription.tsx
+++ b/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSubscription.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Elements, CardElement, useStripe, useElements } from "@stripe/react-stripe-js";
 import { loadStripe } from "@stripe/stripe-js";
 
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSubscription.tsx
```diff
@@ -187,7 +192,9 @@ const getValue = (form: HTMLFormElement, field: string) => {
 
 function AddPaymentMethod({ onDone, workspaceId }: { onDone: () => void; workspaceId: string }) {
   const [saving, setSaving] = useState(false);
-  const [error, setError] = useState<string>();
+  const [error, setError] = useState<string | undefined>(
+    stripeLoadFailed ? "Unable to add a payment method at this time." : undefined
```
hbenl, line 196: If the component is rendered before the `stripePromise` is rejected, `error` won't get updated, so no error will be shown in the UI (only the console will show "Failed to load stripe").

hbenl (changes requested):

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSubscription.tsx
ryanjduffy: :+1: I've pushed the promise handling into an effect which I think should address this but I wasn't able to test this specific scenario well.

hbenl (approved):

PR #3708, opened 2021-09-22
Show "Loading..." in the console while evaluating (#2698)

hbenl:


ryanjduffy:
fyi - @jonbell-lot23 - This relates to user feedback when there is some sort of slower-than-expected performance.

ryanjduffy (approved):
I think this is a good improvement. It would be nice to defer the loading placeholder for a little bit to avoid the flash for short evaluations but I think that could come later (along with any other UX magic that Jon might offer).

jonbell-lot23:
Agree @ryanjduffy and @hbenl . I use UX powers of ten for this: https://www.nngroup.com/articles/powers-of-10-time-scales-in-ux/

I'd set a delay of one second before it appears. This means 2+2 won't show a flash, but more complex interactions will. We might find that one second is too long in practice, but it's a good starting point.

ryanjduffy:
I'd be curious what our current p90 is for evaluation times. Could inform our delay timing

Issue #3466, opened 2021-09-07
ReactDevTools: highlight and pick components


hbenl:
- when the user hovers over a component, the ReactDevTools frontend sends a `highlightNativeElement` event with a `rendererID` and an `id`
- when the user clicks the component picker, the frontend sends `startInspectingNative`/`stopInspectingNative` events
- to get the DOM nodes for a component, we can evaluate `__REACT_DEVTOOLS_GLOBAL_HOOK__.rendererInterfaces.get(rendererID).findNativeNodesForFiberID(id)`
- to get the entire component tree:
  - get the IDs of the root components using `store._roots`
  - get the child component IDs for a component with `id` using `store._idToElement.get(id).children`

hbenl:
#3709 only highlights components on hover, picking components is TBD.

PR #3709, opened 2021-09-22
Highlight react components on hover (#3466)

hbenl:


src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -34,28 +41,70 @@ class ReplayWall implements Wall {
...
+        const nodeFront = response.returned.getNodeFront();
+        if (!nodeFront) {
+          return;
+        }
+
+        await nodeFront.ensureLoaded();
+        if (!this.isHighlightedElementID(rendererID, id)) {
+          return;
+        }
+
```
ryanjduffy, line 98: Why three separate early returns along with the duplicate `isHighlightedElementId()` call? Maybe simplify to

```ts
        const nodeFront = response.returned?.getNodeFront();
        await nodeFront?.ensureLoaded();
        if (!nodeFront || !this.isHighlightedElementID(rendererID, id)) {
          return;
        }
```

ryanjduffy (approved):
Looks good, @hbenl! One simplification idea below but otherwise :shipit:

jasonLaster:
@hbenl can you include a replay? It'd be fun to see the new code here.

Also, this feels like something that would be good to cover with an end-to-end test. Even if we can't assess the canvas we can still listen for something that triggers a test pass/failure. For instance tests could listen for Canvas commands or something like that...

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
hbenl: `isHighlightedElement()` is used to check if the user still hovers over the same component in ReactDevTools. If he doesn't, we want to avoid unnecessary work (`nodeFront.ensureLoaded()` sends 5 requests to the backend). Imagine if the user quickly moves the mouse across the component tree, hitting 20 components: without the first `isHighlightedElement()` check, we'd be sending 100 unnecessary requests to the backend, with the first check most of that work will be avoided.
The `if (!nodeFront)` check isn't necessary though, so I simplified it as suggested.

hbenl:
Here's a recording: https://app.replay.io/recording/eccbaf62-fbdb-4fc5-a1ce-05ce601a004f
It also shows the first `isHighlightedElement()` check at work.

hbenl:
> Also, this feels like something that would be good to cover with an end-to-end test.

@jasonLaster see #3718

PR #3734, opened 2021-09-25
Simplify login screen

jasonLaster:
This PR simplifies the login screen with the principle that we should not be giving people too much to read / think about when there is an important CTA. Curious what others think

### before
Screen Shot 2021-09-25 at 7 22 36 AM

### after
Screen Shot 2021-09-25 at 7 33 10 AM

### DIFF
--- a/src/ui/components/Account/index.tsx
+++ b/src/ui/components/Account/index.tsx
@@ -65,21 +65,18 @@ function WelcomePage() {
             <p>In order to join your team, we first need you to sign in.</p>
           ) : (
             <>
-              <p>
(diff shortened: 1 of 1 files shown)


ryanjduffy:
No strong opinions. I'm curious if people are delayed by the copy on this page. Do we see any evidence in LogRocket or elsewhere? I like the link on its own line and left-aligned text is easier to read but I have no data to back up any of those opinions. :)

jasonLaster:
I feel like this is a don't make me think moment. Most folks will Login, if others are confused they will click learn more and see what we're about. The benefit for having the link on the same line is it is less distracting in the 95% case.

PR #3761, opened 2021-09-28
Treat socket errors like a closed socket after more than 5 minutes since the last message (ops#191)

hbenl:


hbenl:
Fixes RecordReplay/ops#191

jasonLaster (approved):

Issue #3745, opened 2021-09-26
Elements sidebar default width is too small


jasonLaster:
<img width="733" alt="Screen Shot 2021-09-26 at 7 57 57 AM" src="https://user-images.githubusercontent.com/254562/134813219-e05aa16a-1f4f-4df9-906e-38fdfcfec34c.png">

PR #3762, opened 2021-09-28
Fix the layout of the elements panel sidebar (#3745)

hbenl:
Fixes 4 issues:
- the minimum sidebar width was too small
- the sidebar width set by the user was not restored
- the panels in the sidebar showed no scrollbar when they overflowed
- the "Pseudo-elements" accordion was pushing other rules out of view


hbenl:
Before: http://localhost:8080/recording/fc0604da-666d-47ca-90a6-d19171d6791e
After: https://app.replay.io/recording/4cb4b8d4-d7a4-428e-92da-d3fe1f9d3444

jasonLaster (approved):

jasonLaster:
This is awesome. And yeah, looking at the After replay really helped see what was going on in App.tsx

PR #3779, opened 2021-09-29
Use reactjs-popup for the tooltips in the breakpoint panel (#3553)

hbenl:
Using `react-tooltip` in the breakpoint panel created several issues:
- #3553
- the tooltip had rendering issues in Firefox
- recently it stopped working altogether, even in Chrome

I replaced it with `reactjs-popup` and removed the duplicated "Editing logpoints is available for Developers in the Team plan" tooltip because it was also shown if the breakpoint was not editable because it was hit too many times. If the user is not a developer in the team plan, there is still a lock icon with that tooltip message.


jaril (approved):
LGTM!

PR #3805, opened 2021-10-01
Remember the user's view mode on a per-replay basis

### LINKED ISSUE #3655: Tabs should remember last state
**Problem**
When you're in DevTools and the page reloads, you're brought back to Viewer, causing you to navigate back to where you were. This happens after a timeout, a socket error, or anything else that reloads the page.

**Background**
This is contrary to core UX best practices. Hitting reload on a GitHub task does not take you to the top of the repo. Going to a Replay URL doesn't take you to the Library. Launching Xcode, VS Code, Nova, BBEdit, any Adobe Creative Suite app, Slack, Discord, Chrome, or any other app I can think of always maintains window state. You have to go out of your way to find counter-examples. Some apps shift views on you, but they are poorly designed.

**Fix**  
There are three options:

1. Always take you to Viewer (what we do today)
2. Always take you to DevTools
3. Default to Viewer, but remember your last tab. If you're in DevTools, you should be returned there the next time you load or reload the page.

We should do number 3 because it's UX best practice.

jaril:
Fix #3655.

### DIFF
--- a/src/ui/setup/index.ts
+++ b/src/ui/setup/index.ts
@@ -1,20 +1,22 @@
 import { matchPath } from "react-router-dom";
 import { bootstrapStore } from "./store";
-import { registerStoreObserver, updatePrefs } from "./prefs";
+import { registerStoreObserver, ReplaySessions, updatePrefs } from "./prefs";
 import { setupAppHelper } from "./helpers";
 import { setupDOMHelpers } from "./dom";
 import { setTelemetryContext, setupTelemetry } from "ui/utils/telemetry";
 import { UIStore } from "ui/actions";
 import { getTheme, getWorkspaceId } from "ui/reducers/app";
-import { setFontLoading, setModal, setWorkspaceId } from "ui/actions/app";
+import { setFontLoading, setModal, setViewMode, setWorkspaceId } from "ui/actions/app";
 import tokenManager from "ui/utils/tokenManager";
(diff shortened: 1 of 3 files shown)


src/ui/setup/index.ts
```diff
@@ -50,6 +52,18 @@ export function bootstrapApp() {
     store.dispatch(setModal("workspace-settings"));
   }
 
+  // Check if we have any preferences saved for this replay and apply
+  // them to the store
+  asyncStore.replaySessions.then((sessions: ReplaySessions) => {
```
jasonLaster, line 57: this would probably be better done before we create the store. that way we can pass it in as initial state. look at how we do it with pending breakpoints.

src/ui/setup/prefs.ts
```diff
@@ -31,4 +42,21 @@ export function updatePrefs(state: UIState, oldState: UIState) {
...
+}
+
+function updateReplaySessions(
+  state: UIState,
+  currentReplaySessions: ReplaySessions
+): ReplaySessions {
+  const recordingId = getRecordingId();
+  const currentReplaySession = { viewMode: getViewMode(state) };
+
+  return { ...currentReplaySessions, [recordingId!]: currentReplaySession };
```
jasonLaster, line 61: 👍

jasonLaster (approved):

PR #3832, opened 2021-10-04
Add opt-out redactions

jasonLaster:
1. Adds a settings modal options
2. Explicitly disables logrocket setup
3. Removes non-sensitive user redactions

Follow up: switch to a proper user setting which will be tied to the user account and not local storage


ryanjduffy (approved):
Functionally, it seems fine. There are some instances of the `data-private` attribute that should be cleaned up since they'll continue to be redacted.

Visually, the option is sort of just floating there but that could be addressed in a follow up ticket.

PR #3860, opened 2021-10-05
Reduce unmounts when toggling the visibility of devtool panes

jcmorrow:
### Motivation

This started as an investigation into https://github.com/RecordReplay/devtools/issues/3851. Toggling the visibility of the Editor was triggering an unmount and remount of basically every component in the DevView. DevView had 4 different branches, each with *slightly* different rendering trees, based on which components (editor, video, both, neither) were open. This was a lot of complexity *and* it actually didn't capture all of the options (because the user can also toggle whether or not the sidebar was open). As a matter of fact, that was causing it's own issue: https://github.com/RecordReplay/devtools/issues/3746 which should *also* be addressed by these changes.

I cleaned it up by simplifying that render tree so that it's always the same child components inside of the `SplitBox`. There is still one place where changing a view causes an unmount and remount which is inside of `Viewer`. This is because in the vertical view the video is the `startComponent` (because vertical SplitBoxes flow top-to-bottom), but in the horizontal view the video is supposed to be the *endComponent* (because horizontal SplitBoxes flow left-to-right). I thought I would be clever and just add `flex-direction: row-reverse` in the horizontal case but that caused a really horrible UI bug where trying to slide the component to make it bigger actually made it smaller and vice versa. I was far enough down the rabbit-hole by this point that I didn't want to spend any more time going that route.

### A Note on How SplitBox is Used

I cannot tell if this is fully intended, but it works quite well in my testing. Consider the following case, where the container is a SplitBox flowing left-to-right, with component A being the controlled component (which means component B has flex-grow on it and will take up whatever space A leaves open):

image

Let's also say that component A might not be visible right now, in which case we want component B to take up its space, ignoring the partition. Rather than doing this:
```
if (!componentA) {
 return componentB;
} else {
  return <SplitBox left={componentA} right={componentB}/>
}
```
You can do something like this:
```
return <SplitBox left={componentA} right={componentB} maxWidth={componentA ? "auto" : "0px"}/>
```
This says "give me a SplitBox that contains these two components, but if this one doesn't exist, then make sure it doesn't take up any space". Without code like this, it's actually pretty difficult to avoid having many different `if` checks in the render statements or having accidental gaps in the UI where the SplitBox is making room for an element that doesn't exist.

This gives us the following:

https://user-images.githubusercontent.com/5903784/135941252-b0be4582-ef85-4e20-912e-0171a9a5f397.mp4

### Long-term Considerations

I think this is in a pretty nice state for the user interface, but I don't love how it feels to code UI's like this with the nested SplitBoxes. I think a future we should move towards, particularly for layouts like DevTools, is a CSS grid, where the drag events just change the CSS, rather than the react state, and things can be moved from panel to panel by just changing their grid number. That's a much larger undertaking.


jcmorrow:
This also makes the left bar resizable as requested in https://github.com/RecordReplay/devtools/issues/3794, although I still need to double-check that it is properly remembered in settings.

src/devtools/client/debugger/src/components/Editor/index.js
```diff
@@ -333,16 +332,6 @@ class Editor extends PureComponent {
     }
   }
 
-  setSize(nextProps, editor) {
-    if (!editor) {
-      return;
-    }
-
-    if (nextProps.startPanelSize !== this.props.startPanelSize) {
-      editor.codeMirror.setSize();
```
jcmorrow, line 342: The [docs on this](https://codemirror.net/doc/manual.html#setSize) indicate that passing null effectively means "don't change anything", so this should be a noop. I've checked it in the browser and indeed I cannot tell any difference between leaving this code in and removing it. Lmk if I'm missing something here!

src/ui/components/SidePanel.tsx
```diff
@@ -31,7 +29,7 @@ function SidePanel({ selectedPrimaryPanel, resizable }: SidePanelProps) {
     <Redacted
       allowOptIn
       style={{
-        width: resizable ? "100%" : `${SIDEPANEL_WIDTH}px`,
```
jcmorrow, line 34: The resizing is now done in `SplitBox`

src/devtools/client/debugger/src/components/Editor/index.js
jaril: That looks to be right 👍

jaril (commented):
Looks good on the devtools side but it looks like this broke the viewer view: https://share.descript.com/view/6KrUlkwajd8

jaril (approved):
Looking good! One last nit: for the initial size of the sidebar, let's keep the behavior the same as before. So `240px` instead of `284px`.

Feel free to merge as soon as that's changed!

PR #3874, opened 2021-10-05
Sharing modal redesign

### LINKED ISSUE #3739: Loom Style Sharing
Loom does some really great things to keep sharing simple and less stressful

Frame 1

I think we should learn from a lot of the human psychology they have here

Frame 17

1. The copy link is a no brainer, nobody needs to see the whole thing
2. The Add People form is simple and anyone should be able to add a collaborator
3. The privacy settings description is easy to understand. Here only the author and team member should be able to edit. Also tucking it away at the bottom is really nice.

jaril:
First part of #3739.

image
image


jasonLaster (approved):
Code looks good.

PR #3891, opened 2021-10-06
comments v5

jaril:


jasonLaster:
Awesome!

I wonder if we'd benefit from the slightest background e.g. #fdfdfd in the sidebar which would let us use backgrounds as depth...

Screen Shot 2021-10-06 at 8 00 14 AM

jaril:
Continued in #3914.

PR #3899, opened 2021-10-06
Persist console history across refreshes

jcmorrow:
### Motivation

This is just an extension of #3896, making it a little easier to pick up where you left off in case you have to refresh the page.

### Storing and Retrieving from Async Storage

I'm just writing this down because it took me a while to figure out how to thread all of this together:

`UIState` is the store, but because it's a mish-mash of old reducers and new ones it's type definition is only *some* of what it has. If you're adding some functionality in the new typescript files and referencing old reducers you might see tsc complain. In that case, you can back-fill type definitions using `.d.ts` files to help. So in this case I was referencing the `messages` reducer, which is defined in 

```
src/devtools/client/webconsole/reducers/messages.js
```

So I had to add a set of type annotations in

```
src/devtools/client/webconsole/reducers/messages.d.ts
```

I only added the bare minimum that I needed to implement this feature, since a lot of that reducer will probably get re-worked soon anyways.

### Populating Redux Store from Async Storage on Page Load

Because asyncStorage is... well... async, it's not as simple as writing 

```
initialState = {
   localStorage.getKey("store")
}
```

Instead, you should populate the store from inside of `src/ui/setup/dynamic/devtools.ts`, which is responsible for injecting new `initialState` pieces that come from async functions.

### Updating Async Storage from Redux Store on Event

`src/ui/setup/prefs.ts` has an `updatePrefs` function, which is subscribed to the redux store. `updateAsyncPref` takes a key and a selector, and updates that key in `asyncStorage` whenever the value of that selector changes.

### DIFF
--- a/src/devtools/client/webconsole/actions/input.js
+++ b/src/devtools/client/webconsole/actions/input.js
@@ -4,12 +4,7 @@
 
 "use strict";
 
-const {
-  EVALUATE_EXPRESSION,
-  SET_TERMINAL_INPUT,
-  SET_TERMINAL_EAGER_RESULT,
-} = require("devtools/client/webconsole/constants");
-const { getAllPrefs } = require("devtools/client/webconsole/selectors/prefs");
+const { EVALUATE_EXPRESSION } = require("devtools/client/webconsole/constants");
 const { ThreadFront, createPrimitiveValueFront } = require("protocol/thread");
 const { assert } = require("protocol/utils");
 
--- a/src/ui/utils/commandHistory.ts
+++ b/src/ui/utils/commandHistory.ts
@@ -1,0 +1,9 @@
+import uniq from "lodash/uniq";
+
+export type Command = string;
+
+const MAX_HISTORY_LENGTH = 1000;
+
+export const appendToHistory = (command: Command, history: Command[]): Command[] => {
+  return uniq([command, ...history].slice(0, MAX_HISTORY_LENGTH));
+};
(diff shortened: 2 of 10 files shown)


src/devtools/client/webconsole/actions/input.js
```diff
@@ -79,6 +75,8 @@ function evaluateExpression(expression) {
       return null;
     }
 
+    appendToStoredHistory(expression);
```
jasonLaster, line 78: okay - i think there's a nicer way to do this. We have a pattern in atleast the debugger, maybe the console already where we subscribe to store events and then update prefs there

https://github.com/RecordReplay/devtools/blob/master/src/ui/setup/prefs.ts#L29-L53

jasonLaster: Thanks. All for moving towards `sagas` or `XState`.

I'm not currently feeling the pain of the subscribe callback. I haven't seen the perf impact when measured and find it pretty simple in this one use case. Other side-effects (API calls, other thunks would be very bad though).

src/ui/utils/commandHistory.ts
```diff
@@ -0,0 +1,9 @@
+import uniq from "lodash/uniq";
+
+export type Command = string;
+
+const MAX_HISTORY_LENGTH = 1000;
+
+export const appendToHistory = (command: Command, history: Command[]): Command[] => {
+  return uniq([command, ...history].slice(0, MAX_HISTORY_LENGTH));
+};
```
jasonLaster, line 9: nice 👍

PR #3940, opened 2021-10-08
Add a BorderBridge component to keep the pause line uninterrupted

### LINKED ISSUE #3932: Our "red stripe" clashes with our comment divider
image

image

On the left, notice how the gray line overlaps the red line and causes a strange appearance. It should look like the example on the right, where the line is tinted darker where the line overlaps.

I am aware that this is a total pain in the ass because it means we can't just use a border. But this is the sort of polish that people expect from a professional-grade app!

jaril:
Fix #3932.

image

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
@@ -55,6 +55,27 @@ function formatRelativeTime(date: Date) {
   return `${minutes}m`;
 }
 
+function BorderBridge({
+  comments,
+  comment,
+  isPaused,
+}: {
+  comments: (Comment | PendingNewComment)[];
+  comment: Comment | PendingNewComment;
+  isPaused: boolean;
+}) {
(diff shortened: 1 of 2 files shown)


src/ui/components/Comments/TranscriptComments/CommentCard.tsx
```diff
@@ -139,13 +162,14 @@ function CommentCard({
...
       className={classNames(
-        `mx-auto w-full border-b border-gray-300 cursor-pointer transition`,
+        `mx-auto relative w-full border-b border-gray-300 cursor-pointer transition`,
         isPaused || hoveredComment === comment.id ? "bg-gray-50" : "bg-white"
       )}
       onClick={() => seekToComment(comment)}
       onMouseEnter={() => setHoveredComment(comment.id)}
       onMouseLeave={() => setHoveredComment(null)}
     >
+      <BorderBridge {...{ comments, comment, isPaused }} />
```
jasonLaster, line 172: The other approach would be to use a `::before` pseudo element ... I think

PR #3944, opened 2021-10-08
Adjust UI of Breakpoints

### LINKED ISSUE #3340: Breakpoint widget loads funny
This replay shows a couple of funny loading states for breakpoints

1. the log is missing 
2. the next/prev probably isn't needed
3. loading... could be centered
4. do we want to show the add comment button

https://app.replay.io/recording/e10e2e14-3f7e-4094-9ad7-58ac6a72dec4?point=141814607952702021779324188728756692&time=47864&hasFrames=false

### LINKED ISSUE #3825: Should be able to leave a comment when a print statement has been called > 200 times
Screen Shot 2021-10-03 at 10 35 32 AM

jcmorrow:
Fixes #3825. Fixes #3340.

### Motivation

This addresses both #3825 and #3340. The thinking here is that we should have an absolute minimum of difference between loading, error, warning, and normal states.

### Changes

- Break the various PanelSummary components out into their own files. This component is surprisingly complex, and has been a bit of a hot-point for changes. I've tried to give the logic that is there a little more room to breathe by pulling the components rendered out into nearby files, to remove their behavior from the cognitive load completely when thinking about how breakpoints behave.
- The comment button is now always visible, as is the timeline, and `leading-none` has been added to the Material Icon that is shown on warning, so that it doesn't increase the size a couple of pixels when it gets rendered (this creates a small jump and makes the heights inconsistent).
- Disabled navigation buttons will no longer turn the cursor into a pointer

https://user-images.githubusercontent.com/5903784/136997361-37ce0cdd-c397-4fac-a32b-997c94500240.mp4


jasonLaster:
Looks good! 

Can we also add the default expression?

PR #3945, opened 2021-10-08
Make sure that we initialize the replay sessions

### LINKED ISSUE #3857: Remember I have hidden video even after a refresh
image

STR:

1. Load Replay, go to DevTools
2. Hide Video
3. Reload

Expected:

* The video should still be hidden after reloading the page

Actual:

* The video comes back every time, making me have to re-hide it

jaril:
Fix #3857.


src/ui/setup/index.ts
```diff
@@ -34,9 +40,26 @@ export function bootstrapApp() {
...
+      console.log({ session, session });
 
-  setupAppHelper(store);
+      if (!session) {
+        return;
+      }
+
+      store.dispatch(setViewMode(session.viewMode));
+      store.dispatch(setShowVideoPanel(session.showVideoPanel));
+      store.dispatch(setShowEditor(session.showEditor));
```
jcmorrow, line 57: What do you think about moving this into `src/ui/setup/dynamic/devtools.ts`, kind of similar to what I did over at: https://github.com/RecordReplay/devtools/pull/3899/files#diff-54af09a258c2f6698b5fd88700d5ce17b3c7c77ffc5e2431420d5c9e39532942R77-R79?

jcmorrow: Also, not necessary for this PR, but it seems like it could be worth it to have a `setReplaySession` action which can update all of these at the same time for when we switch recordings.

jcmorrow (commented):
Couple of small comments, looks good! 💿

src/ui/setup/index.ts
```diff
@@ -34,9 +40,26 @@ export function bootstrapApp() {
...
   window.store = store;
 
-  registerStoreObserver(store, updatePrefs);
+  // Check if we have any preferences saved for this replay and apply
+  // them to the store. Todo: Do this step before we create the store.
+  asyncStore.replaySessions
+    .then((sessions: ReplaySessions) => {
+      const session = sessions[getRecordingId()!];
+
+      console.log({ session, session });
```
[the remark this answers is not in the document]
jaril: 👌

src/ui/setup/prefs.ts
```diff
@@ -43,20 +49,17 @@ export function updatePrefs(state: UIState, oldState: UIState) {
   updatePref("selectedPanel", getSelectedPanel);
   updateAsyncPref("eventListenerBreakpoints", (state: UIState) => state.eventListenerBreakpoints);
 
-  const previousReplaySessions = currentReplaySessions;
-  currentReplaySessions = updateReplaySessions(state, currentReplaySessions);
-
-  if (previousReplaySessions && previousReplaySessions !== currentReplaySessions) {
-    asyncStore.replaySessions = currentReplaySessions;
-  }
+  await updateReplaySessions(state);
```
jasonLaster, line 52: this might be a moderate perf slowdown. Kinda how @jcmorrow predicted :) 

I'm thinking it might be reasonable to remove the `await` here as long as we can guarantee ordering.

jaril: Yeah there's nothing really waiting for this function so we can just axe it

PR #3948, opened 2021-10-09
Handle failed repaintGraphics

### LINKED ISSUE #3900: Clicking on timeline does not update video until Play is pressed
**Steps to Reproduce:**

1. Open a recording
2. Press play to the point where you can see an image
3. Click on the timeline somewhere later in the recording

**Expected Result:**

Image should update to reflect the state at the point in time you clicked. It appears to continue to display whatever image was there before until you press Play again.

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:

jasonLaster:
This adds support for catching failed repaintGraphics commands painting the closest graphics. fixes #3900 

before: [replay](https://app.replay.io/recording/d3d01584-4b52-4a3b-8eb6-9c34e40319aa?point=8762000950135628567926910023107222&time=3924.6592465753424&hasFrames=true#)
after: [replay](https://app.replay.io/recording/25166723-69c5-43ca-8287-b277ae505b02?point=13954297809385237553354531699427275&time=8415.865964625295&hasFrames=true)

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -239,19 +239,26 @@ export function setupGraphics(store: UIStore) {
     assert(pause);
 
     const rv = await pause.repaintGraphics();
-    if (pause === ThreadFront.currentPause && rv) {
-      const { mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
-      let { description, screenShot } = rv;
-      if (screenShot) {
-        repaintedScreenshots.set(description.hash, screenShot);
-      } else {
-        screenShot = repaintedScreenshots.get(description.hash);
-        if (!screenShot) {
-          console.error("Missing repainted screenshot", description);
-          return;
+    if (pause === ThreadFront.currentPause) {
+      // First try to use the new graphics,
+      // then fallback to the most recent graphics.
+      if (rv) {
+        const { mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
+        let { description, screenShot } = rv;
+        if (screenShot) {
+          repaintedScreenshots.set(description.hash, screenShot);
+        } else {
+          screenShot = repaintedScreenshots.get(description.hash);
+          if (!screenShot) {
+            console.error("Missing repainted screenshot", description);
+            return;
+          }
         }
+        paintGraphics(screenShot, mouse);
+      } else {
+        const { screen, mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
+        paintGraphics(screen, mouse);
       }
-      paintGraphics(screenShot, mouse);
     }
   });
 }


src/protocol/graphics.ts
```diff
@@ -238,20 +238,28 @@ export function setupGraphics(store: UIStore) {
...
+          screenShot = repaintedScreenshots.get(description.hash);
+          if (!screenShot) {
+            console.error("Missing repainted screenshot", description);
+            return;
+          }
         }
+        paintGraphics(screenShot, mouse);
+      }
+    } catch (e) {
+      if (e.code == 33) {
```
jasonLaster, line 258: I'm not sure if we're exporting protocol error constants

bhackett1024: Using a constant 33 is fine, but this needs a comment.

src/protocol/graphics.ts
```diff
@@ -238,20 +238,28 @@ export function setupGraphics(store: UIStore) {
...
+        const { screen, mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
+        paintGraphics(screen, mouse);
+        Video.seek(ThreadFront.currentTime);
```
bhackett1024, line 261: Why is this call needed?  It's not called in the non-exceptional case and seems like it shouldn't be needed here.

src/protocol/thread/pause.ts
```diff
@@ -484,6 +484,7 @@ export class Pause {
       rv = await this.sendMessage(client.DOM.repaintGraphics, {});
     } catch (e) {
       console.error("DOM.repaintGraphics failed", e);
+      throw e;
```
bhackett1024, line 487: This change isn't necessary, as the caller can test for null.  This change is also incorrect, because the repaintGraphicsWaiter won't get resolved and later calls to repaint the graphics will get stuck waiting forever.

src/protocol/graphics.ts
```diff
@@ -238,20 +238,28 @@ export function setupGraphics(store: UIStore) {
...
+            console.error("Missing repainted screenshot", description);
+            return;
+          }
         }
+        paintGraphics(screenShot, mouse);
+      }
+    } catch (e) {
+      if (e.code == 33) {
+        const { screen, mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
+        paintGraphics(screen, mouse);
```
bhackett1024, line 260: It isn't always valid to paint these graphics at this point.  If we await on the repaint, the user navigates somewhere else, and then the repaint throws, we can't update the graphics.  This is what the "if (pause === ...)" block above is for.

src/protocol/graphics.ts
```diff
@@ -238,20 +238,28 @@ export function setupGraphics(store: UIStore) {
...
-      if (screenShot) {
-        repaintedScreenshots.set(description.hash, screenShot);
-      } else {
-        screenShot = repaintedScreenshots.get(description.hash);
-        if (!screenShot) {
-          console.error("Missing repainted screenshot", description);
-          return;
+    try {
+      const rv = await pause.repaintGraphics();
+      if (pause === ThreadFront.currentPause && rv) {
```
bhackett1024, line 243: A null rv means that repainting failed.  So, instead of throwing an exception we can just add logic that runs in the !rv case.

bhackett1024 (changes requested):

src/protocol/graphics.ts
```diff
@@ -239,19 +239,25 @@ export function setupGraphics(store: UIStore) {
...
+          repaintedScreenshots.set(description.hash, screenShot);
+        } else {
+          screenShot = repaintedScreenshots.get(description.hash);
+          if (!screenShot) {
+            console.error("Missing repainted screenshot", description);
+            return;
+          }
         }
+        paintGraphics(screenShot, mouse);
+      } else if (rv === null) {
```
bhackett1024, line 256: This can just be an "else" branch, there aren't other meaningful falsey values and when repainting we always want to paint *something* if we can.

src/protocol/graphics.ts
```diff
@@ -239,19 +239,25 @@ export function setupGraphics(store: UIStore) {
...
-      let { description, screenShot } = rv;
-      if (screenShot) {
-        repaintedScreenshots.set(description.hash, screenShot);
-      } else {
-        screenShot = repaintedScreenshots.get(description.hash);
-        if (!screenShot) {
-          console.error("Missing repainted screenshot", description);
-          return;
+    if (pause === ThreadFront.currentPause) {
+      if (rv) {
```
bhackett1024, line 243: This needs a brief comment describing what's going on.

src/protocol/graphics.ts
bhackett1024: This call needs to be removed.  It isn't used in the other branch so it doesn't belong here.

PR #3956, opened 2021-10-09
Restyle share btn

jasonLaster:
I think eventually we want the material icon, but not thrilled w/ how it looks at the moment

Screen Shot 2021-10-09 at 7 56 44 AM
Screen Shot 2021-10-09 at 7 56 28 AM


jasonLaster:
closing since i think we'll do a better job with the chrome ui work

Issue #3946, opened 2021-10-08
Many DOM elements can have their full contents loaded unnecessarily


bhackett1024:
STR: See comments in https://app.replay.io/recording/a9906416-1501-4604-95a5-bbe006417c7e

In this facebook recording we are very sluggish updating the elements panel to show a node which was picked that is several levels deep.  The main problem seems to be that we're fetching the full contents of more than 450 different nodes, which slows things down quite a lot.  Getting the data about the selected node and its transitive parents in the backend took 4 seconds, but it took 7 seconds for the client to receive the response to that command and 19 seconds to actually finish updating the UI.  The four second time isn't great either, but I think it is also being slowed down by all the commands to load node contents.

It looks like the reason we're loading so many nodes is because `NodeFront.childNodes()` loads the full contents of all of those child nodes.  This is (well, shouldn't be) necessary --- we should only need these full contents when a user has actually selected on a node and we need to show its rules/style/etc. information.  I think we're loading the child node contents here as a form of preloading, so that if the user selects a node (expanding its children) then they can keep clicking on other child nodes and get the information immediately without a round trip to the server.

This strategy is backfiring here though, because we're preloading so much stuff that it takes an extremely long time to update the UI and show the user what they want to see now.  We need to be more careful about how node contents get loaded.  Some ideas for improving this are below.  We probably need to do all of these to make this work well on complex sites.

1. Don't preload node contents until after everything needed to update the UI finishes loading.
2. Limit the number of nodes we're preloading at a time.  Sending 1500 commands at once (300 nodes x 5 commands to load the node contents) to operate on the same pause is never a good idea.
3. Limit the maximum number of child nodes we can preload content for.  If there is a super complex DOM node with a zillion children we don't want to preload all of their contents.

PR #3957, opened 2021-10-09
Lazily load node properties

jasonLaster:
We were previously fetching node properties when we loaded the children nodes. Now we load them on demand. improves #3946

This PR has three primary changes:
1. We no longer load all of the element's data upfront
2. The 5 node protocol methods now have their own async request wrappers
3. The redux code now knows that they're async.

### DIFF
--- a/src/protocol/thread/node.ts
+++ b/src/protocol/thread/node.ts
@@ -35,6 +35,14 @@ export class NodeFront {
   private _quads: BoxModel | null;
   private _bounds: Rect | null;
 
+  private _waiters: {
+    computedStyle: Deferred<any> | null;
+    rules: Deferred<any> | null;
+    listeners: Deferred<any> | null;
+    quads: Deferred<any> | null;
+    bounds: Deferred<any> | null;
+  };
+
   constructor(pause: Pause, data: WiredObject) {
     this._pause = pause;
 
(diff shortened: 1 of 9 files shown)


bhackett1024:
I like the direction this is going, but there are CI failures on the inspector tests which need to be addressed before this can land:

Failure inspector-02.js https://app.replay.io/recording/f27b7ba8-143f-4acd-ab20-1248118ac063
Failure highlighter.js https://app.replay.io/recording/1ea100c1-ac3b-41de-96e7-729ba86c6411

jasonLaster:
* inspector-02 was a test harness issue
* highlighter caught a legitimate issue where we needed to fetch the boxModel before highlighting
* I left `hasEventListeners` as a todo since it is relatively minor and we can follow up later

src/protocol/thread/node.ts
```diff
@@ -198,67 +214,35 @@ export class NodeFront {
...
-        .then(
-          ({ rect }) => {
-            this._bounds = rect;
-          },
-          () => {
-            this._bounds = null;
-          }
-        ),
-    ]);
+    await this.getComputedStyle();
```
bhackett1024, line 245: This method seems kind of strange now, and at least needs a comment saying why having a computed style == the node is loaded.  It seems best to remove this method, can we just do that?

PR #3961, opened 2021-10-09
More resliliant react node picker

jasonLaster:
The ReactDevTools Node Picker currently fails when we cannot fetch the boundingClientRects. For the user, this looks like the node picker is selected and should work, but there aren't any components on the page. 

This PR changes two things
1. it wraps `getBoundingClientRects` similar to how we wrap `repaintGraphics` 
2. it adds some error handling for ReactDevtools

### DIFF
--- a/src/protocol/thread/pause.ts
+++ b/src/protocol/thread/pause.ts
@@ -13,6 +13,7 @@ import {
   NodeBounds,
   PauseData,
   repaintGraphicsResult,
+  getAllBoundingClientRectsResult,
 } from "@recordreplay/protocol";
 import { client } from "../socket";
 import { defer, assert, Deferred } from "../utils";
@@ -97,7 +98,7 @@ export class Pause {
   documentNode: NodeFront | undefined;
   domFronts: Map<string, DOMFront>;
   stack: WiredFrame[] | undefined;
-  loadMouseTargetsWaiter: Deferred<void> | undefined;
+  loadMouseTargetsWaiter: Deferred<getAllBoundingClientRectsResult | null> | undefined;
   repaintGraphicsWaiter: Deferred<repaintGraphicsResult | null> | undefined;
   mouseTargets: NodeBounds[] | undefined;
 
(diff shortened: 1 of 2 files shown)


src/protocol/thread/pause.ts
```diff
@@ -451,9 +451,18 @@ export class Pause {
...
-    this.mouseTargets = elements;
-    this.loadMouseTargetsWaiter.resolve();
+    let rv;
+    try {
+      rv = await this.sendMessage(client.DOM.getAllBoundingClientRects, {});
+      this.mouseTargets = rv.elements;
+    } catch (e) {
+      console.error("DOM.getAllBoundingClientRects failed", e);
+    } finally {
+      this.mouseTargets = [];
```
jasonLaster, line 461: setting `mouseTargets` to an empty array prevents `getMouseTarget` and other consumers from failing. Another option would be to set it to `null` and add a guard in the consumers

bhackett1024: This code is broken, this will overwrite the mouseTargets assignment that happens in the successful case.  This mouseTargets assignment belongs in the catch block.

Also note that inspector-02 (which tests the node picker) is failing in CI:

Failure inspector-02.js https://app.replay.io/recording/233304ca-6cff-4db7-8b1e-bd7ae55328ea

src/protocol/thread/pause.ts
```diff
@@ -451,9 +451,18 @@ export class Pause {
...
+      this.loadMouseTargetsWaiter.resolve();
+    }
+
+    return rv;
```
bhackett1024, line 465: rv here is basically some random object.  This function should return a boolean indicating if it failed or not.

src/protocol/thread/pause.ts
```diff
@@ -451,9 +451,18 @@ export class Pause {
...
-    this.loadMouseTargetsWaiter.resolve();
+    let rv;
+    try {
+      rv = await this.sendMessage(client.DOM.getAllBoundingClientRects, {});
+      this.mouseTargets = rv.elements;
+    } catch (e) {
+      console.error("DOM.getAllBoundingClientRects failed", e);
+    } finally {
+      this.mouseTargets = [];
+      this.loadMouseTargetsWaiter.resolve();
```
bhackett1024, line 462: This needs to resolve() with the return value because that is what will be returned when this.loadMouseTargetsWaiter is set on entry.  This also doesn't need to belong in the finally clause because the catch block ensures we won't be throwing here.

bhackett1024 (changes requested):

src/protocol/thread/pause.ts
bhackett1024: It would be nice if this returned !!rv so the return value is a boolean.

bhackett1024 (approved):

PR #3973, opened 2021-10-11
Keep the comment editor from losing its input when the user toggles views

### LINKED ISSUE #3871: Comment content lost when switching between panes
If I start writing a comment and then switch to another pane on the left, like the paused state, and then switch back, the content that I added to the comment is gone and I have to start over writing it.

jaril:
Fix #3871.

Demo: https://app.replay.io/recording/caabcc67-9ed4-4c19-9d0a-8ed8fc2f5858

### DIFF
--- a/src/ui/actions/comments.ts
+++ b/src/ui/actions/comments.ts
@@ -20,9 +20,14 @@ const {
 
 type SetPendingComment = Action<"set_pending_comment"> & { comment: PendingComment | null };
 type SetHoveredComment = Action<"set_hovered_comment"> & { comment: any };
+type UpdatePendingCommentContent = Action<"update_pending_comment_content"> & { content: string };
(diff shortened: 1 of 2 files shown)


src/ui/actions/comments.ts
```diff
@@ -35,6 +36,21 @@ export function clearPendingComment(): SetPendingComment {
...
+export function updatePendingCommentContent(content: string): UIThunkAction {
+  return async ({ getState, dispatch }) => {
+    const pendingComment = getPendingComment(getState());
+
+    if (!pendingComment) {
+      return;
+    }
+
+    const newPendingComment = { ...pendingComment };
+    newPendingComment.comment.content = content;
```
jasonLaster, line 48: this is still a shallow copy.

you probably need something like this
```js
{ ...pendingComment, comment: {...pending.comment, comment: content  } `;
```

jasonLaster (approved):
Looks good!

PR #3974, opened 2021-10-11
(WIP) Add Storybook

jasonLaster:
This PR gets us close wrt to Storybook support and certainly teaches us some things as well.

1. Storybook has the potential to be awesome. Hot Reloading and fixture support will be a big win
2. Adopting Storybook will force us to remove some of the old Firefox cruft which is keeping Replay non-standard and preventing Jest support and likely Next as well. This was a pre-req for example (https://github.com/RecordReplay/devtools/pull/3970)
3. Using the MockApolloProvider like we do for our mock tests will help with stubbing the graphql hooks
4. This is currently failing to include tailwindcss for some unknown reason. This likely fairly straightforward
5. Architecturally, I went with wrapping one of the stories with a bunch of divs because that was simpler and importing a CSS module... This is probably one reason why folks adopt more css-in-js approaches and separate behavior from presentation.
6. When we land storybook we should add some automated testing because otherwise these will quickly go stale
7. We should share stuff from our original webpack config so we dont dupe stuff unecessarily


jasonLaster:
@jcmorrow would you like to run with this this week?

jaril:
Closing this since @jcmorrow is working on a separate PR.

jasonLaster:
opening a new issue https://github.com/RecordReplay/devtools/issues/4349

PR #3989, opened 2021-10-12
Maintain a sidebar min-width of 240px

jaril:
Fix #3960.


jcmorrow (approved):
I mentioned that I thought this might break something but actually it's all good! The thing that is tricky that I was trying to remember is the issue I ran into [here](https://github.com/RecordReplay/devtools/pull/3877) where using the old SplitBox component some units get `parseInt` run on them, and so values that look reasonable end up getting interpreted in the wrong units, but that is *not* happening here. 🌤️

Issue #3320, opened 2021-08-25
Collapse consecutive keypress events in the events sidepanel


jaril:
Follow up from #3307.

Add heuristics for collapsing keyboard events that should be together. Instead of having `Key Press H` and `Key Press I`, there should be some logic that realizes that perhaps the user was typing something and it should be presented as `Typed HI`.

jasonLaster:
Nice I think some form of tree view makes sense in the future. Today, I think we just need a subset of events we show now. For instance, I think it's fine to pluck the first of n events so the user seeks `Key Down "H"` and not Down, Press, Up. This should both be simpler and still literal.

PR #3994, opened 2021-10-12
Show the current time in the timeline

### LINKED ISSUE #3822: Events timeline should show the current pause
**Description:**

the events timeline should behave similar to the console with a secondary pause line in between the events or the current paused event highlighted 

Screen Shot 2021-10-02 at 9 28 56 PM

jcmorrow:
Closes #3822 , #3320 is coming up next.

Changes
---

- Split out some of the components in the Events folder, to try and give them some room to grow, since I'm about to add a bunch of logic.
- Add `lightGrey` to the colors that our tailwind knows about
- Add a secondary accent (the *now* color) line that shows where in the event timeline the current time is, and fade events that have not yet happened. This is quite pleasing to watch as the video plays.
- Fix what I *think* is a bug. Events on the timeline are represented in html as `button` elements, so that they behave like they are clickable (which makes sense, because they are). When you click on one of them, it gets focus. If you then press space bar two things happen at once:

1. The viewer attempts to start playing
2. The viewer immediately jumps to that point in time and pauses

It manifests as like a small UI hiccup every time you press the space bar while having focus on a timeline event. I could actually see a good argument for allowing timeline events to work this way, if people wanted to use the keyboard to navigate the timeline, but I think most of the time this behavior is just confusing, and if they *really* want to tab through and select they can always hit "Enter", which has the same effect as space, but doesn't try to start the player first.

https://user-images.githubusercontent.com/5903784/137043297-bf7d02b6-a69d-49af-907d-cabe9ccacee3.mp4

### DIFF
--- a/src/ui/components/Events/Event.tsx
+++ b/src/ui/components/Events/Event.tsx
@@ -1,0 +1,86 @@
+import React from "react";
+import { KeyboardEvent, NavigationEvent } from "@recordreplay/protocol";
+import classNames from "classnames";
+import { ReplayEvent } from "ui/state/app";
+import { getFormattedTime } from "ui/utils/timeline";
+import MaterialIcon from "../shared/MaterialIcon";
+
+function Event({
+  onSeek,
+  currentTime,
+  executionPoint,
+  event,
+}: {
+  event: ReplayEvent;
+  currentTime: any;
+  executionPoint: any;
+  onSeek: (point: string, time: number) => void;
+}) {
+  const isPaused = event.time === currentTime && executionPoint === event.point;
+
+  let text: string;
+  let title: string;
+  let icon: string;
+
+  if (event.kind?.includes("mouse")) {
+    text = "Mouse Click";
+    title = "Mouse Click Event";
+    icon = "ads_click";
+  } else if (event.kind?.includes("navigation")) {
+    const ev = event as NavigationEvent;
+    text = `${ev.url}`;
+    title = "Navigation Event";
+    icon = "place";
+  } else {
+    icon = "keyboard";
+    let ev = event as KeyboardEvent;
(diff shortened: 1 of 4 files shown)


src/ui/components/Events/Events.css
```diff
@@ -0,0 +1,6 @@
+.current-time {
```
jcmorrow, line 1: I went off-tailwind here. This is like a weird decorative div that has very specific display properties. Would this be better inside of a component with inline styles?

src/ui/components/Events/Event.tsx
```diff
@@ -0,0 +1,91 @@
...
+        }
+      )}
+    >
+      <div className="flex flex-row space-x-2 items-center overflow-hidden">
+        <MaterialIcon className="group-hover:text-primaryAccent">{icon}</MaterialIcon>
+        <div
+          className={classNames(
+            "group-hover:text-primaryAccent overflow-ellipsis overflow-hidden whitespace-pre",
+            {
+              "font-semibold": isPaused,
```
jcmorrow, line 75: I see a lot of a`classNames(conditional ? "class-name" : "")` scattered throughout here. `classNames` is specifically made to make that phrasing into the more elegant:

```{ "class-name": conditional }```

So I've switched to that usage where possible.

jcmorrow:
> We've got accessibility issues because the blue is just too light to be used for text. Let's keep this in for now, but we're likely to adjust more later on.

Thanks, this was what I was looking for! Quick follow-up, should we change the paused state to match that? (so, only the icon highlighted), especially now that we have the red line as well? Also, shouldn't the blue icon when paused actually be a red icon? Like the comment button on paused comments? Maybe I'm just confusing concepts now.

PR #4009, opened 2021-10-13
Create a preferences tab in the settings modal

### LINKED ISSUE #4008: Add a Preferences panel to the settings modal

jaril:
Fixes #4008.

https://user-images.githubusercontent.com/15959269/137385746-d98c120d-f640-42c2-b56c-3a4e185db43e.mov

### DIFF
--- a/src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
+++ b/src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
@@ -2,52 +2,55 @@ import React, { useState } from "react";
 import hooks from "ui/hooks";
 import { prefs } from "ui/utils/prefs";
 import Checkbox from "../Forms/Checkbox";
-import { EmailPreference } from "ui/hooks/users";
+import { EmailSubscription } from "ui/hooks/users";
 import Radio from "../Forms/Radio";
 
 const EMAIL_NOTIFICATIONS = {
-  [EmailPreference.COLLABORATOR_REQUEST]: "When somebody invites you to collaborate on a replay",
-  [EmailPreference.REPLAY_COMMENT]: "When somebody comments on your replay",
-  [EmailPreference.NEW_TEAM_INVITE]: "When you're invited to a new team",
+  [EmailSubscription.COLLABORATOR_REQUEST]: "When somebody invites you to collaborate on a replay",
+  [EmailSubscription.REPLAY_COMMENT]: "When somebody comments on your replay",
+  [EmailSubscription.NEW_TEAM_INVITE]: "When you're invited to a new team",
 };
 
 function Notification({
   content,
-  preference,
-  emailsOptedOut,
+  emailType,
+  unsubscribedEmailTypes,
 }: {
   content: string;
-  preference: EmailPreference;
-  emailsOptedOut: EmailPreference[];
+  emailType: EmailSubscription;
+  unsubscribedEmailTypes: EmailSubscription[];
 }) {
-  const enableEmailPreference = hooks.useEnableEmailPreference();
-  const disableEmailPreference = hooks.useDisableEmailPreference();
-
-  const checked = !emailsOptedOut.includes(preference);
+  const subscribeToEmailType = hooks.useSubscribeToEmailType();
+  const unsubscribeToEmailType = hooks.useUnsubscribeToEmailType();
+  const checked = !unsubscribedEmailTypes.includes(emailType);
 
   const handleChange = () => {
     if (checked) {
-      disableEmailPreference(preference, emailsOptedOut);
+      unsubscribeToEmailType(emailType);
     } else {
-      enableEmailPreference(preference, emailsOptedOut);
+      subscribeToEmailType(emailType);
     }
   };
 
   return (
-    <label className="flex items-center space-x-2 cursor-pointer" data-private htmlFor={preference}>
-      <Checkbox id={preference} checked={checked} onChange={handleChange} />
+    <label className="flex items-center space-x-2 cursor-pointer" data-private htmlFor={emailType}>
+      <Checkbox id={emailType} checked={checked} onChange={handleChange} />
       <div>{content}</div>
     </label>
   );
 }
 
-function MarketingPreferences({ emailsOptedOut }: { emailsOptedOut: EmailPreference[] }) {
-  const enableEmailPreference = hooks.useEnableEmailPreference();
-  const disableEmailPreference = hooks.useDisableEmailPreference();
-  const isEnabled = !emailsOptedOut.includes(EmailPreference.MARKETING);
+function MarketingPreferences({
+  unsubscribedEmailTypes,
+}: {
+  unsubscribedEmailTypes: EmailSubscription[];
+}) {
+  const subscribeToEmailType = hooks.useSubscribeToEmailType();
+  const unsubscribeToEmailType = hooks.useUnsubscribeToEmailType();
+  const isEnabled = !unsubscribedEmailTypes.includes(EmailSubscription.MARKETING);
 
-  const onEnable = () => enableEmailPreference(EmailPreference.MARKETING, emailsOptedOut);
-  const onDisable = () => disableEmailPreference(EmailPreference.MARKETING, emailsOptedOut);
+  const onEnable = () => subscribeToEmailType(EmailSubscription.MARKETING);
+  const onDisable = () => unsubscribeToEmailType(EmailSubscription.MARKETING);
 
   return (
     <div className="space-y-4">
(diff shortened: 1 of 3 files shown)


src/ui/components/shared/Forms/Radio.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export default function Radio({ id, name, checked, onChange }: React.HTMLProps<HTMLInputElement>) {
```
jcmorrow, line 3: Not sure about this suggestion, so feel free to ignore, but maybe we could do `overrides` and `{...overrides}` on line 7?

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+import { getFeatureFlag } from "ui/utils/launchdarkly";
+import { AvatarImage } from "ui/components/Avatar";
+import { prefs } from "ui/utils/prefs";
+import Checkbox from "../Forms/Checkbox";
+import { disableEmailPreference, EmailPreference, enableEmailPreference } from "ui/hooks/users";
+import Radio from "../Forms/Radio";
+
+const EMAIL_NOTIFICATIONS = {
+  [EmailPreference.COLLABORATOR_REQUESTS]: "When somebody invites you to collaborate on a replay",
```
jcmorrow, line 27: Big fan of using the enum here and getting away from strings 👍

jcmorrow: This is a super nit-pick, but these might read more consistently if they were all singular, e.g. `REPLAY_COMMENT` and `COLLABORATOR_REQUEST`.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+  emailPreferences,
+}: {
+  content: string;
+  preference: EmailPreference;
+  emailPreferences: EmailPreference[];
+}) {
+  const updateUserEmailPreferences = hooks.useUpdateUserEmailPreferences();
+
+  // The preferences list is a blacklist of preferences that are disabled.
+  const checked = !emailPreferences.includes(preference);
```
jcmorrow, line 44: This double-negative logic is very confusing, `if it doesn't exist in the deny-list then allow it`. If I were somewhere else in the code and I saw 
```
const preferences: EmailPreference[] {
  marketing: true,
  collaboratorRequest: true
}
```

I would 100% assume that that user had opted in to those emails, but currently that would actually mean that those are the only options this user has opted *out* of? Either we should call those `EmailOptOut` or `BlockedEmail`, or we should keep the name `EmailPreference` and make it such that `true === subscribed`. This should also make it more clear how the checkboxes line up, so `true` equates to  `checked`.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const handleChange = () => {
+    let newPrefs;
```
jcmorrow, line 47: I think personally I would do away with the `let` business here and just allow the duplication:

```
    if (checked) {
      update(disable(preference, emailPreferences));
    } else {
      update(enable(preference, emailPreferences));
    }
```

I know it violates DRY, but I think it's easier to read. I suspect that there's an even cleaner way to manage this, but I don't see it right now.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    updateUserEmailPreferences({ variables: { emailPreferences: newPrefs } });
+  };
+
+  return (
+    <div className="flex items-center space-x-2 cursor-pointer" data-private>
+      <Checkbox id={preference} checked={checked} onChange={handleChange} />
+      <label className="cursor-pointer" htmlFor={preference}>
+        <div>{content}</div>
```
jcmorrow, line 61: Is the `div` doing something here?

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+    } else {
+      newPrefs = enableEmailPreference(preference, emailPreferences);
+    }
+
+    updateUserEmailPreferences({ variables: { emailPreferences: newPrefs } });
+  };
+
+  return (
+    <div className="flex items-center space-x-2 cursor-pointer" data-private>
+      <Checkbox id={preference} checked={checked} onChange={handleChange} />
```
jcmorrow, line 59: I'm used to sticking checkboxes inside the label just so there's no area between the two where clicking doesn't trigger the box. Either way works, just something to watch out for!

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+function Notification({
+  content,
+  preference,
+  emailPreferences,
+}: {
+  content: string;
+  preference: EmailPreference;
+  emailPreferences: EmailPreference[];
+}) {
+  const updateUserEmailPreferences = hooks.useUpdateUserEmailPreferences();
```
jcmorrow, line 41: maybe call these `update`? It makes them a lot easier to read and locally it's the only thing that `update` could possible mean I think.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const handleChange = () => {
+    let newPrefs;
+    if (checked) {
+      newPrefs = disableEmailPreference(preference, emailPreferences);
+    } else {
+      newPrefs = enableEmailPreference(preference, emailPreferences);
+    }
+
+    updateUserEmailPreferences({ variables: { emailPreferences: newPrefs } });
```
jcmorrow, line 54: Maybe we could change the signature of this method so it just takes `newPrefs`? It seems like an API detail that it has to be in this nested JSON structure, and the part that is managing the actual HTTP request is probably where that logic should be stored.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const updateUserEmailPreferences = hooks.useUpdateUserEmailPreferences();
+  const isEnabled = !emailPreferences.includes(EmailPreference.MARKETING_EMAILS);
+
+  const onEnable = () => {
+    const newPrefs = enableEmailPreference(EmailPreference.MARKETING_EMAILS, emailPreferences);
+    updateUserEmailPreferences({ variables: { emailPreferences: newPrefs } });
+  };
+  const onDisable = () => {
+    const newPrefs = disableEmailPreference(EmailPreference.MARKETING_EMAILS, emailPreferences);
+    updateUserEmailPreferences({ variables: { emailPreferences: newPrefs } });
```
jcmorrow, line 77: And if you make the signature changes and naming changes this line and `onEnable` get much more concise:

```
updateUserEmailPreferences(enable(EmailPreference.MARKETING_EMAILS, emailPreferences)
```

src/ui/hooks/users.ts
```diff
@@ -32,6 +32,13 @@ export enum Nag {
   FIRST_BREAKPOINT_ADD = "first_breakpoint_add",
 }
 
+export enum EmailPreference {
+  MARKETING_EMAILS = "marketing_emails",
```
jcmorrow, line 36: I would maybe change this to `MARKETING`? Adding `EMAILS` seems redundant.

src/ui/components/shared/Forms/Radio.tsx
jaril: 👍

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
jaril: 👍

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
jaril: Nope!

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,153 @@
...
+  emailsOptedOut: EmailPreference[];
+}) {
+  const enableEmailPreference = hooks.useEnableEmailPreference();
+  const disableEmailPreference = hooks.useDisableEmailPreference();
+
+  const checked = !emailsOptedOut.includes(preference);
+
+  const handleChange = () => {
+    if (checked) {
+      disableEmailPreference(preference, emailsOptedOut);
```
jcmorrow, line 30: beautiful 🖌️

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -0,0 +1,160 @@
...
+  return (
+    <div className="space-y-4">
+      <div className="text-lg">Email Preferences</div>
+      <div className="flex flex-col space-y-2 p-1">
+        <label className="flex space-x-2 cursor-pointer" htmlFor="enable_marketing" data-private>
+          <Radio
+            id="enable_marketing"
+            name="marketing_email"
+            checked={isEnabled}
+            onChange={onEnable}
```
jcmorrow, line 64: It seems a little funky to *always* be calling `onEnable` when `onChange` is called, rather than switching on whether or not we are `checked` after change?

PR #4020, opened 2021-10-14
Add routing to open the user settings modal

### LINKED ISSUE #4019: Add routing that opens up the user's settings on the preferences tab

jaril:
Fix #4019.


src/ui/setup/index.ts
```diff
@@ -90,3 +78,21 @@ export async function bootstrapApp() {
 
   return store;
 }
+
+async function maybeAutoOpenModal(store: UIStore) {
```
ryanjduffy, line 82: This doesn't need to be `async`

PR #4027, opened 2021-10-15
Fetch preferences for a replay when the user navigates from the library

### LINKED ISSUE #4014: Recordings are opened with video+editor hidden
This is a bug concerning how we fetch replay-specific preferences (viewMode, selectedPrimaryPanel, showVideoPanel, showEditor) and routing. 

The expected behavior is that those preferences should persist within a replay. This is the case when you hit refresh, your preferences are loaded back and you're right back where you were. 

However, this doesn't happen when you first open the library then click on a replay. Instead, your preferences are initialized to falsy while you're in the library (because there's no recording ID), and isn't updated when you get routed to the replay.

jaril:
Fix #4014.

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -9,6 +9,7 @@ import {
   loadedRegions,
   KeyboardEvent,
   NavigationEvent,
+  RecordingId,
 } from "@recordreplay/protocol";
 import { ThreadFront, RecordingTarget } from "protocol/thread/thread";
 import * as selectors from "ui/reducers/app";
@@ -30,6 +31,7 @@ import groupBy from "lodash/groupBy";
 import { compareBigInt } from "ui/utils/helpers";
 import { isTest } from "ui/utils/environment";
 import tokenManager from "ui/utils/tokenManager";
+import { asyncStore } from "ui/utils/prefs";
 
 export type SetRecordingDurationAction = Action<"set_recording_duration"> & { duration: number };
 export type LoadingAction = Action<"loading"> & { loading: number };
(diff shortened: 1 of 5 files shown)


src/ui/actions/app.ts
```diff
@@ -375,3 +377,21 @@ export function setShowVideoPanel(showVideoPanel: boolean): SetShowVideoPanelAct
 export function setShowEditor(showEditor: boolean): SetShowEditorAction {
   return { type: "set_show_editor", showEditor };
 }
+
+export function loadReplayPrefs(recordingId: RecordingId): UIThunkAction {
+  return async ({ dispatch }) => {
+    const replaySessions = await asyncStore.replaySessions;
+    const session = replaySessions[recordingId];
+
+    if (!session) {
```
jcmorrow, line 386: I assume this represents the case where the user has never viewed that recording before. In that case, should we load the default settings? Having a `DEFAULT_SETTINGS` const would also be a nice way of centralizing those values. I know they also live in prefs, sort of, but pulling out a well-named constant might be more self-documenting and central.

src/ui/reducers/app.ts
```diff
@@ -48,6 +48,12 @@ const syncInitialAppState: AppState = {
 
 export async function getInitialAppState(): Promise<AppState> {
   const recordingId = getRecordingId();
+
+  // If we're in the library, there are no preferences to fetch.
+  if (!recordingId) {
```
jcmorrow, line 53: What do you think about wrapping the call below in an `if(recordingId)` rather than returning early `if(!recordingId)`? I could see someone adding some more initialization below this and getting very confused when it doesn't occur on non-recording pages.

src/ui/actions/app.ts
jaril: #4047.

jcmorrow (approved):
This works for me, we could probably include more things in this, but let's get this out now and we can think about other things to add later ⚡

PR #4029, opened 2021-10-15
Show a tooltip over a log when it's not editable

### LINKED ISSUE #3727: Non editable breakpoint expression tooltip not showing
When a breakpoint isn't editable (we can't edit the expression), we have a lock icon and we gray out the background of the expression. We also include a tooltip for when the user hovers over it.

That tooltip shows up for the lock button, but it doesn't for the expression.

https://user-images.githubusercontent.com/15959269/134748190-251b8dd7-8914-427f-bbde-90e5f70ddce3.mov

Diff for forcing the uneditable breakpoint treatment:
```
Puter:devtools jarilvalenciano$ git diff
diff --git a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary.tsx b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary.tsx
index 0ac00458..de146117 100644
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary.tsx
@@ -60,7 +60,7 @@ function PanelSummary({
   const isHot = analysisPoints && analysisPoints.length > prefs.maxHitsDisplayed;
   const didExceedMaxHitsEditable = analysisPoints && analysisPoints.length < prefs.maxHitsEditable;
   const isTeamDeveloper = recording ? recording.userRole !== "team-user" : false;
-  const isEditable = didExceedMaxHitsEditable && isTeamDeveloper;
+  const isEditable = didExceedMaxHitsEditable && isTeamDeveloper && false;
 
   const handleClick = (event: React.MouseEvent, input: Input) => {
     if (!isEditable) {
@@ -186,7 +186,7 @@ function PanelSummary({
           ) : null}
         </button>
       </div>
-      {!isTeamDeveloper ? (
+      {!isTeamDeveloper || true ? (
         <span
           className="material-icons cursor-default text-gray-400"
           data-tip="Editing logpoints is available for Developers in the Team plan"
```

jaril:
Fix #3727.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/SummaryRow.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/SummaryRow.tsx
@@ -1,16 +1,16 @@
 import React from "react";
 
 export default function SummaryRow({
-  expression,
+  children,
   label,
 }: {
-  expression: React.ReactElement;
+  children: React.ReactElement;
   label: string | null;
 }) {
   return (
     <div className="flex flex-row space-x-1 items-center">
       {label ? <div className="w-6 flex-shrink-0">{label}</div> : null}
-      {expression}
+      {children}
     </div>
   );
 }
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/index.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/index.tsx
@@ -35,15 +35,15 @@ function PanelSummary({
   currentTime,
   analysisPoints,
 }: PanelSummaryProps) {
-  const isTeamDeveloper = hooks.useIsTeamDeveloper();
+  const { isTeamDeveloper } = hooks.useIsTeamDeveloper();
   const conditionValue = breakpoint.options.condition;
   const logValue = breakpoint.options.logValue;
 
   const isHot = analysisPoints && analysisPoints.length > prefs.maxHitsDisplayed;
-  const didExceedMaxHitsEditable = !!(
+  const isUnderMaxHitsEditable = !!(
     analysisPoints && analysisPoints.length < prefs.maxHitsEditable
   );
-  const isEditable = didExceedMaxHitsEditable && isTeamDeveloper;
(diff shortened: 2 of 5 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/SummaryRow.tsx
```diff
@@ -1,65 +1,16 @@
...
 
-import MaterialIcon from "ui/components/shared/MaterialIcon";
-import escapeHtml from "escape-html";
-const { getCodeMirror } = require("devtools/client/debugger/src/utils/editor/create-editor");
-
-interface SummaryRowProps {
-  isEditable: boolean;
-  handleClick: (event: React.MouseEvent) => void;
+export default function SummaryRow({
+  expression,
```
ryanjduffy, line 4: If there's only one prop that accepts a `ReactNode`, we should consider making it be `children` since it lends itself to more readable code.

`<MyThing coolProp={<span>Hi!</span>}>` vs `<MyThing><span>Hi!</span></MyThing>`

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/index.tsx
```diff
@@ -106,8 +102,23 @@ function PanelSummary({
   return (
     <div className="summary space-x-2" onClick={e => handleClick(e, "logValue")}>
       <div className="options items-center flex-col flex-grow">
-        <Condition {...{ isEditable, handleClick, conditionValue }} />
-        <Log {...{ isEditable, handleClick, logValue }} hasCondition={!!conditionValue} />
+        {conditionValue ? (
+          <Condition
+            {...{
```
ryanjduffy, line 107: We should either user an object spread or JSX attributes. This is a code style thing but we should probably land on a default. _My preference_ is to use JSX attributes unless we're spreading an existing object.

```
  // uses both but doesn't construct an object just to spread it
  <Condition {...rest} a={a} b={b} />
```

PR #4036, opened 2021-10-16
Fix focus and click targets for comments

jcmorrow:
This is more complicated than I expected! It's complicated for these reasons:
- There are multiple editors per CommentCard (at least one, that is disabled by default, for the top-level comment, and at least one for the enabled reply box [assuming we're paused there, otherwise none of this really matters])
- Every part of the comment serves as a click target to focus the related reply box
- Reply boxes can appear and disappear (sometimes several at once!), so you have to think about both competing for focus as well as competing for autofocus.

I went with a context here. This is one of the scenarios where I've seen them work pretty well, and it makes it much easier that we don't have to thread focus information through like 4 layers of components.

Screencaps
---
Easy situation - clicking on a comment card focuses its related input, clicking away blurs it, clicking on it multiple times doesn't do anything special, it just stays focused

https://user-images.githubusercontent.com/5903784/137574505-659e74cf-6417-4b86-b206-89675af18839.mp4

---

Slightly harder - focus switches between cards smoothly

https://user-images.githubusercontent.com/5903784/137574574-7e587a9b-c4df-4000-a429-06dc00ff4703.mp4

---

Harder still - focus and autofocus is managed while editors are appearing and disappearing

https://user-images.githubusercontent.com/5903784/137574590-97845350-523b-4577-80be-ea51ba6a74f4.mp4

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
@@ -129,17 +129,25 @@ type PropsFromParent = {
 };
 type CommentCardProps = PropsFromRedux & PropsFromParent;
 
+export const FocusContext = React.createContext({
+  autofocus: false,
+  isFocused: false,
+  blur: () => {},
+});
+
 function CommentCard({
   comment,
   comments,
   currentTime,
   executionPoint,
-  seekToComment,
   hoveredComment,
-  setHoveredComment,
   pendingComment,
+  seekToComment,
+  setHoveredComment,
 }: CommentCardProps) {
   const isPaused = currentTime === comment.time && executionPoint === comment.point;
+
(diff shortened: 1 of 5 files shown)


src/ui/components/Comments/TranscriptComments/CommentCard.tsx
```diff
@@ -129,29 +129,42 @@ type PropsFromParent = {
 };
 type CommentCardProps = PropsFromRedux & PropsFromParent;
 
+export const FocusContext = React.createContext({
+  autofocus: false,
+  isFocused: false,
+  blur: () => {},
```
jonbell-lot23, line 135: Test -- ignore this comment

jcmorrow:
Merging this as it's a definite improvement, and we can make follow-on issues for any future changes!

PR #4052, opened 2021-10-19
Don't allow empty comments to be submitted

jcmorrow:
Previously we were storing comments as markdown, so checking `content.length` would tell us if the comment was empty. Now that we store the comments in json an empty comment is like `{content: blahblahblah}` so we can't just qualify all strings that are longer than 0 as valid comments. Instead, we check at the editor level what our character count is (which only counts characters that actually get rendered), and submit if *that* is greater than 0.

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
@@ -23,12 +23,6 @@ function NewCommentEditor({ clearPendingComment, comment, setModal, type }: NewC
       return;
     }
 
-    // For now we can simply bail if the input happens to be empty. We should fix
-    // this in the next pass to handle and show an error prompt.
-    if (inputValue == "") {
-      return;
-    }
-
     if (type == "new_reply") {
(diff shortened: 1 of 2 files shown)


jasonLaster:
This is an interesting one :)

I dont know if the current UI currently would look good, but many times i do want to leave empty comments as bookmarks for interesting places in the code

jaril:
Empty comments are kinda nice, and Jason's fond of em :P

jcmorrow:
Ah, interesting. The reason I assumed we should *not* allow them is because of this code:

```
    // For now we can simply bail if the input happens to be empty. We should fix
    // this in the next pass to handle and show an error prompt.
    if (inputValue == "") {
      return;
    }
```

It was originally added here: https://github.com/RecordReplay/devtools/commit/4441cef48efcfb86618c0a1c7af47ccb4f0d3c85#diff-6ef3e26d34c4402d0fb7015f8f15dac2c0c8c4a2ede849f1e762c3c1b19bbbfbR32-R62

But maybe it just never worked?

jcmorrow:
I'll close this and open a new PR that just removes the dead code.

PR #4054, opened 2021-10-19
Add trimming UI under a pref

jaril:
**Demo: https://app.replay.io/recording/140349be-1e92-4d8a-9ddd-33f152083d72**

This adds a Trimming UI (see the issue). 

Currently, the user has no influence on the way we load/unload regions on the backend. It's possible for us to (unintentionally) unload something they're interested in. The originally-discussed trimming UI's purpose is to allow the user to influence the way we load/unload regions on the backend.

I don't believe we currently have the backend capability to do that yet fully. Instead, this is an exploration for the front-end piece of that behavior. It allows the user to enter trimming mode where they can specify a span of times that they're interested in. 

This action won't affect any backend behavior. For now, this action will simply limit things that they see in the devtools. By trimming, we'll limit items like breakpoint hits/messages that we display to the time subset that they've specified.

This will be merged under a pref so we (_cough cough_ @jonbell-lot23) can play with it and get a good idea for how we should proceed with it. I limited the scope just to get the MVP out, but there are a few obvious suspects for follow-up fixes to this:
1) Dragging the whole span left<>right should retain its original width until the user mouses up. [Demo](https://user-images.githubusercontent.com/15959269/137936532-0c3351d1-6013-413d-b5ec-634a9ab99596.mov)
2) Make the trim span blue background fill the entire height of the timeline, as specified in the notion docs.
3) The video should update while dragging the start/end dragger elements to show the user what part they're on.
4) While dragging the start/end dragger elements, the time tooltip should appear.
5) Persist the `trimRegion` locally on a per-replay basis.
6) Hovering on the timeline while trimming trips the timeline's hover state. Should it?
6) The outside-of-trim elements are still navigable with stepping. It shouldn't be.

There's also some general discussions that would be good to get through before shipping this (non-pref). Off the top of my head:
- Right now we use the dotted line visual language to indicate unloaded regions. This coopts that exactly and uses it to indicate the areas outside the trimmed span. Should we/how should we differentiate between the two?
- Should the `trimRegion` continue to be persisted on a local, per-replay basis? Or something else?
- What would it take to get the originally-intended backend-influencing version of this?

To enable this, open up the console and do `app.features.trimming = true`.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -24,7 +24,14 @@ import { getPendingComment } from "ui/reducers/comments";
 import { UIStore, UIThunkAction } from ".";
 import { Action } from "redux";
 import { PauseEventArgs } from "protocol/thread/thread";
-import { TimelineState, Tooltip, ZoomRegion, HoveredItem, TrimRegion } from "ui/state/timeline";
+import {
+  TimelineState,
+  Tooltip,
+  ZoomRegion,
+  HoveredItem,
+  TrimRegion,
+  TrimOperation,
+} from "ui/state/timeline";
 import { getPausePointParams, getTest } from "ui/utils/environment";
 import { assert, waitForTime } from "protocol/utils";
 import { features } from "ui/utils/prefs";
@@ -33,7 +40,6 @@ import { getFirstComment } from "ui/hooks/comments/comments";
 import { isRepaintEnabled } from "protocol/enable-repaint";
 import { getModal } from "ui/reducers/app";
 import clamp from "lodash/clamp";
-import { TrimOperation } from "ui/components/Timeline/Trimmer";
 
 export type SetTimelineStateAction = Action<"set_timeline_state"> & {
   state: Partial<TimelineState>;
--- a/src/ui/components/Timeline/Trimmer.tsx
+++ b/src/ui/components/Timeline/Trimmer.tsx
@@ -5,15 +5,11 @@ import { actions } from "ui/actions";
 import { selectors } from "ui/reducers";
 import { UIState } from "ui/state";
 import clamp from "lodash/clamp";
-import { TrimRegion, ZoomRegion } from "ui/state/timeline";
+import { TrimOperation, TrimRegion, ZoomRegion } from "ui/state/timeline";
 import { getVisiblePosition } from "ui/utils/timeline";
 import classNames from "classnames";
 
-export enum TrimOperation {
-  resizeStart = "resizeStart",
-  resizeEnd = "resizeEnd",
-  moveSpan = "moveSpan",
-}
+
 type TrimmerProps = PropsFromRedux & { width: number };
 
 const getPosition = (time: number, zoom: ZoomRegion) => {
--- a/src/ui/state/timeline.ts
+++ b/src/ui/state/timeline.ts
@@ -50,4 +50,8 @@ interface HoveredLocation extends Location {
   column: number;
 }
 
-export type DragTarget = "startTime" | "endTime" | "span";
+export enum TrimOperation {
+  resizeStart = "resizeStart",
+  resizeEnd = "resizeEnd",
+  moveSpan = "moveSpan",
+}


src/devtools/client/webconsole/selectors/messages.js
```diff
@@ -31,12 +33,22 @@ export const getVisibleMessages = createSelector(
   state => state.messages.visibleMessages,
   state => state.consoleUI.zoomStartTime,
   state => state.consoleUI.zoomEndTime,
-  (messages, visibleMessages, zoomStartTime, zoomEndTime) => {
-    return visibleMessages.filter(id => {
+  getTrimRegion,
+  (messages, visibleMessages, zoomStartTime, zoomEndTime, trimRegion) => {
```
jcmorrow, line 37: Will this run on every update? If so, it's probably better to use `useMemo` in the components that care about filtering.

src/ui/actions/timeline.ts
```diff
@@ -202,8 +217,9 @@ export function setTimelineState(state: Partial<TimelineState>): SetTimelineStat
 export function setTimelineToTime(time: number | null, updateGraphics = true): UIThunkAction {
   return async ({ dispatch, getState }) => {
     dispatch(setTimelineState({ hoverTime: time }));
+    const isTrimming = getModal(getState()) === "trimming";
```
jcmorrow, line 220: Maybe pull out `state` here as a local so we don't call `getState` potentially three times?

src/ui/actions/timeline.ts
```diff
@@ -500,3 +516,40 @@ export function clearHoveredItem(): UIThunkAction {
...
+  type: "startTime" | "endTime" | "span",
+  midpointDifference: number
+): UIThunkAction {
+  return ({ dispatch, getState }) => {
+    const state = getState();
+    const hoverTime = getHoverTime(state);
+    const trimRegion = getTrimRegion(state);
+    const zoomRegion = getZoomRegion(state);
+
+    if (!hoverTime || !trimRegion) return;
```
jcmorrow, line 537: I would maybe word this as `if !(hoverTime && trimRegion) return;`
Just to reduce the number of negatives, but either way it's a little hard to parse. Maybe give this a meaningful name? What do both of these values being true indicate? What does it mean if one or both of them is missing?

src/ui/actions/timeline.ts
```diff
@@ -500,3 +516,40 @@ export function clearHoveredItem(): UIThunkAction {
...
+
+export function setTrimRegion(trimRegion: {
+  startTime: number;
+  endTime: number;
+}): SetTrimRegionAction {
+  return { type: "set_trim_region", trimRegion };
+}
+
+export function updateTrimRegion(
+  type: "startTime" | "endTime" | "span",
```
jcmorrow, line 528: These sound like a good typescript `Enum`

src/ui/actions/timeline.ts
```diff
@@ -500,3 +516,40 @@ export function clearHoveredItem(): UIThunkAction {
...
+    const state = getState();
+    const hoverTime = getHoverTime(state);
+    const trimRegion = getTrimRegion(state);
+    const zoomRegion = getZoomRegion(state);
+
+    if (!hoverTime || !trimRegion) return;
+
+    if (type === "span") {
+      const { startTime, endTime } = trimRegion;
+      const oldSpanMidpoint = (endTime + startTime) / 2;
```
jcmorrow, line 541: I would maybe pull this and the clamp function out into well-named utils like `timelineMidPoint(start, end)` and `clampToZoomRegion(zoomRegion, time)` but YMMV!

src/ui/actions/timeline.ts
```diff
@@ -500,3 +516,40 @@ export function clearHoveredItem(): UIThunkAction {
...
+      const newMidpoint = hoverTime;
+      const midpointShift = newMidpoint - oldSpanMidpoint;
+
+      const newTrimRegion = {
+        startTime: clamp(startTime + midpointShift + midpointDifference, 0, zoomRegion.endTime),
+        endTime: clamp(endTime + midpointShift + midpointDifference, 0, zoomRegion.endTime),
```
jcmorrow, line 547: It's interesting that we specify the max but the min is always 0?

src/ui/actions/timeline.ts
```diff
@@ -500,3 +516,40 @@ export function clearHoveredItem(): UIThunkAction {
...
+      };
+
+      dispatch(setTrimRegion(newTrimRegion));
+    } else {
+      dispatch(setTrimRegion({ ...trimRegion, [type]: hoverTime! }));
```
jcmorrow, line 552: Stop. 🛑 Hover time! 🔨

src/ui/components/Timeline/Timeline.css
```diff
@@ -50,6 +50,8 @@
 }
 
 .timeline .progress-bar {
+  display: flex;
+  align-items: center;
```
jcmorrow, line 54: Just gonna suuuuper nit here and advocate for sorting styles when they are all of equal importance.

src/ui/components/Timeline/Trimmer.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+import { actions } from "ui/actions";
+import { selectors } from "ui/reducers";
+import { UIState } from "ui/state";
+import clamp from "lodash/clamp";
+import { TrimRegion, ZoomRegion } from "ui/state/timeline";
+import { getVisiblePosition } from "ui/utils/timeline";
+import classNames from "classnames";
+
+export type DragTarget = "startTime" | "endTime" | "span";
```
jcmorrow, line 12: `Enum`?

src/ui/actions/timeline.ts
jcmorrow: Also, it looks like we also use this check elsewhere on line 126 of trimmer and maybe also on line 10 of tooltip. Maybe we should centralize that logic somewhere?

src/ui/actions/timeline.ts
jcmorrow: Looks like a good place could be `utils/timeline`

jcmorrow (approved):
Most of these are just style comments. I don't have a ton to say about the architecture of this. One concern I have is that we have a lot of places where we're checking things like:

```
if (isTrimming) {
  return
}
```

And that makes me concerned that if we ever end up with an edge-case where `isTrimming` is incorrectly set a lot of things will suddenly stop working. I don't see that edge case yet, but just something to look out for!

src/ui/actions/timeline.ts
jcmorrow: Nope, `startTime` makes total sense.

src/ui/components/Timeline/Timeline.css
jcmorrow: well in this case just putting `align-items` first since it comes earlier in the alphabet. I usually either have a css linter do this for me or run `:sort` in vim.

src/devtools/client/webconsole/selectors/messages.js
jcmorrow: Yep, I missed that this was happening inside of `createSelector` because of the way the diff was cut off. All good to go 💯

PR #4055, opened 2021-10-19
Show windows early access badge

jasonLaster:
This adds a windows early access badge while loading the replay. It's missing the platform check though which is surprising to me, since it should be in the buildId `macOS-gecko-20211018105919`, but the buildId only returns `gecko` 

Screen Shot 2021-10-19 at 8 25 31 AM

### DIFF
--- a/src/image/image.css
+++ b/src/image/image.css
@@ -1,3 +1,8 @@
+.img.windows {
+  mask: url(~image/images/icon-windows.svg);
+  width: 24px;
+  height: 24px;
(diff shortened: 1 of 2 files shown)


jasonLaster:
@jazzdan here's a replay https://app.replay.io/recording/6f07bbe8-edb5-4ad3-a15c-18fc71b5a49c?point=10709112272013099469841295082523161&time=6101.476947535771&hasFrames=true

PR #4061, opened 2021-10-19
Add telemetry to E2E tests

jcmorrow:
I have verified that we can see these events in Honeycomb!

Closes https://github.com/RecordReplay/devtools/issues/3984

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -2,6 +2,7 @@
 
 const fs = require("fs");
 const os = require("os");
+const https = require("https");
 const { spawnSync } = require("child_process");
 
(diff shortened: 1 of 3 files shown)


test/utils.js
```diff
@@ -85,4 +86,34 @@ function defer() {
   return { promise, resolve, reject };
 }
 
-module.exports = { findGeckoPath, createTestScript, tmpFile, spawnChecked, defer };
+function sendTelemetryEvent(telemetryEvent, tags) {
+  const headers = {
```
jasonLaster, line 90: perhaps move the guard check here...

PR #4062, opened 2021-10-20
Migrate LogRocket pref to user settings

ryanjduffy:
Blocked by https://github.com/RecordReplay/backend/pull/3293

* Migrates existing pref to settings
* Updates logrocket init to use setting
* Updates settings modal to update setting instead of pref

### DIFF
--- a/src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
+++ b/src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
@@ -80,6 +80,8 @@ function PrivacyPreferences() {
     setDisableLogRocket(userSettings.disableLogRocket);
   }, [loading]);
 
+  if (loading) return null;
+
   return (
     <div className="space-y-4">
       <div className="text-lg">Privacy Preferences</div>


src/ui/utils/apolloClient.tsx
```diff
@@ -146,7 +146,7 @@ function createHttpLink(token: string | undefined) {
   }
 
   return new HttpLink({
-    uri: "https://api.replay.io/v1/graphql",
+    uri: "http://localhost:8090/v1/graphql",
```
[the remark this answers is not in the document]
ryanjduffy: I've been meaning to make this configurable for a while too because i've committed it multiple times in a PR. :/

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -63,11 +63,22 @@ function NotificationPreferences({
...
+    updateLogRocket({
+      variables: {
+        newValue,
+      },
+    });
   };
-  const [disableLogRocket, setDisableLogRocket] = useState(prefs.disableLogRocket);
+
+  useEffect(() => {
+    setDisableLogRocket(userSettings.disableLogRocket);
```
jaril, line 80: Looks like it's possible for `useGetUserSettings` to return `loading: true` here since it's not always being called in the parent. It is if you open the debugger, but if you stay in the viewer you'll hit loading here.

This currently allows this UI to default to give the impression that LogRocket is disabled while loading. What do you think about not showing `PrivacyPreferences` until after loading, so at least we're not giving the wrong impression?

src/ui/actions/session.ts
```diff
@@ -120,6 +122,7 @@ export function createSession(recordingId: string): UIThunkAction {
         recording,
         userInfo,
         auth0User: tokenManager.auth0Client?.user,
+        userSettings,
```
jaril, line 125: Maybe just pass in `disabled: userSettings` here?

jaril: On second thought we pass in `recording` and `userInfo` just to check its properties and see if we need to bail. Might be good to pull those out and just do the if statement here.

src/ui/components/Library/Library.tsx
```diff
@@ -80,10 +81,10 @@ function LibraryLoader(props: PropsFromRedux) {
   }
 
   useEffect(() => {
-    if (!userInfo.loading) {
-      LogRocket.createSession({ userInfo, auth0User: auth.user });
+    if (!userInfo.loading && !userSettingsLoading) {
+      LogRocket.createSession({ userInfo, auth0User: auth.user, userSettings });
     }
-  }, [auth, userInfo]);
+  }, [auth, userInfo, userSettings, userSettingsLoading]);
```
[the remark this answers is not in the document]
jaril: Yes. Yes we do. :P

jaril (approved):
Looks good to me. Left some comments, but I don't feel too strongly about any of them so i'll leave it up to you!

PR #4064, opened 2021-10-20
First draft of new copy

jonbell-lot23:
This addresses https://github.com/RecordReplay/devtools/issues/3906. @colbyr I was just working on this when it scooted over, so take it away!

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/FirstEditNag.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/FirstEditNag.tsx
@@ -3,9 +3,9 @@ import hooks from "ui/hooks";
 
 import { Nag } from "ui/hooks/users";
 const TEXT = {
-  clickPrompt: "Click below to edit the message",
-  enterPrompt: "Type something and hit Enter",
-  success: "The console has been refreshed",
+  clickPrompt: "Want to see something cool? Click below to edit...",
+  enterPrompt: "Now type something here...",
+  success: "Check the console! We went back in time to add those print statements!",
 };
 
 function FirstEditNag({ editing, nags }: { editing: boolean; nags: Nag[] }) {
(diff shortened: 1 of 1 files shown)


colbyr:
Thanks @jonbell-lot23!

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/FirstEditNag.tsx
```diff
@@ -1,15 +1,18 @@
...
-  success: "The console has been refreshed",
+  clickPrompt: "Want to see something cool? Click below to edit...",
+  enterPrompt: "Now type something here...",
+  success: "Check the console! We went back in time to add those print statements!",
 };
 
+type Step = keyof typeof TEXT;
+
 function FirstEditNag({ editing, nags }: { editing: boolean; nags: Nag[] }) {
-  const [step, setStep] = useState<"clickPrompt" | "enterPrompt" | "success">("clickPrompt");
```
colbyr, line 12: Took the liberty of generating this type from the `TEXT` definition.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/FirstEditNag.css
```diff
@@ -0,0 +1,4 @@
+.first-edit-nag {
+  background: var(--theme-popup-color);
+  color: var(--theme-popup-background);
```
colbyr, line 3: I think this'll match [LineNumberTooltip](https://github.com/RecordReplay/devtools/blob/81d9a94fbbcdeeafeb1b13b4a895f5d8c5072641/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.css#L9-L10).

@jonbell-lot23 do you get a little bit of an optical illusion where it looks like there's not border here or is that just me? Maybe I can try to make the top border match the nag background or something.

Screen Shot 2021-10-20 at 3 54 27 PM

jasonLaster: @colbyr I wonder if we should go in the other direction and make both nags blue which I think might be more playful and more clearly a call out.

It could also be fun to do other hover effects, like hovering in the panel makes the stars glisten yellow, even sparkle... 

This is probably a separate issue, but we should also look at the hover area for the expression because even though the entire box is clickable, only a portion changes on hover and we can make the hover effect more obvious. All of this is coming from the place that half of the users who add a breakpoint actually edit it. We should get that # as high as possible.

PR #4066, opened 2021-10-20
Add back stack trace expanding/collapsing

jcmorrow:
I was deleting a lot of message reducer code for https://github.com/RecordReplay/devtools/pull/3896 and it turns out that
some of the code that I deleted was actually responsible for expanding/collapsing stacktraces in the console.

Fixes https://github.com/RecordReplay/devtools/issues/4013


jasonLaster:
@jcmorrow I bet this will fix this issue https://github.com/RecordReplay/devtools/issues/3981

PR #4070, opened 2021-10-20
Show DevTools Sidebar

jasonLaster:
improves #3854

https://app.replay.io/recording/1b28841b-03dc-41e9-8461-61bb2bbc6de6?point=41213856322882733085590641679794375&time=17122.858465853278&hasFrames=true

### DIFF
--- a/src/ui/components/Toolbox.tsx
+++ b/src/ui/components/Toolbox.tsx
@@ -2,16 +2,15 @@ import React from "react";
 
 import DebuggerApp from "devtools/client/debugger/src/components/App";
 import "./Toolbox.css";
-import { Redacted } from "./Redacted";
 
 export default function Toolbox() {
   return (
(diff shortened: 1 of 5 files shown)


src/ui/components/Toolbox.tsx
```diff
@@ -6,12 +6,10 @@ import { Redacted } from "./Redacted";
 
 export default function Toolbox() {
   return (
-    <Redacted id="toolbox">
-      <div className="toolbox-top-panels">
```
jaril, line 10: We're still using the `#toolbox` selector in a few CSS spots. We should leave this in, or just consolidate `toolbox`/`toolbox-top-panels`.

jaril (approved):
Selector CSS nit. I'd say minor, but it's currently breaking the splitter between editor and video+console.

Good to ship once that's fixed!

PR #4078, opened 2021-10-20
Only repaint graphics in loaded regions

hbenl:
When repainting is enabled, it can take some time until the first graphics is shown. It seems to me that `DOM.repaintGraphics` needs to wait until the corresponding region has been loaded (@bhackett1024  please confirm), so I created a fallback that will use a recorded screenshot when the current pause's region hasn't been loaded yet.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -38,7 +38,7 @@ import { features } from "ui/utils/prefs";
 import KeyShortcuts, { isEditableElement } from "ui/utils/key-shortcuts";
 import { getFirstComment } from "ui/hooks/comments/comments";
 import { isRepaintEnabled } from "protocol/enable-repaint";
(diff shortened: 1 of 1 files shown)


src/ui/actions/timeline.ts
```diff
@@ -177,7 +177,7 @@ function onPaused({ time }: PauseEventArgs): UIThunkAction {
     dispatch(setTimelineState({ currentTime: time, playback: null }));
 
     try {
-      if (!isRepaintEnabled()) {
+      if (!isRepaintEnabled() || !isRegionLoaded(store.getState(), time)) {
```
bhackett1024, line 180: Having a comment here would be nice, since we're trying to strike a balance between showing graphics quickly vs. avoiding jitter due to showing the most recent paint and then replacing them with the repainted graphics soon afterwards.

PR #4084, opened 2021-10-20
Add a single test to the frontend

jcmorrow:
One of the action items that came out of last week's retro was "start unit testing the front-end". And the smallest iteration of that was "get one test running on the front-end, with jest, that also runs on CI".

I was surprised at how close I got (I think) to getting all of the old tests to run and pass. One option would be to simply mark the rest as pending (or just get rid of them if they are failing but everything is fine). I ended up with around ~50 failures total. It's enough that I don't want to spend time getting all of those to run, because in many cases that would involve understanding what was actually going wrong, and that sounds like a big time commitment.

Closes #4084


jasonLaster:
# Just one Jest Test

PR #4090, opened 2021-10-21
New upload screen

### LINKED ISSUE #4041: Upload Screen v4
This ticket talks about showing our privacy details more prominently: https://github.com/RecordReplay/backend/issues/2164

One of the places we'll show this is in an updated upload screen.

jaril:
Replay demo: https://app.replay.io/recording/a3aeae6c-dc65-489f-802a-0ebe2f189374?point=66201784950761379877771993315018196&time=33727&hasFrames=false#

Demo:
https://user-images.githubusercontent.com/15959269/138306336-6cf27d74-0b13-4e46-ac0f-74da8575c110.mov

This fixes #4041.

This is the first part of implementing the new "jellyfish" design that we worked on with Zypsy. Functionally, the upload screen still largely works the same. 

Summary of noticeable changes are:
- Removed the replay target URL, and the replay's duration from the upload screen.
- Removed the badge that indicates that the trial is expiring.
- Removed the warning when a user records something that is over 2 minutes.

### DIFF
--- a/src/ui/components/UploadScreen/SettingsPreview.tsx
+++ b/src/ui/components/UploadScreen/SettingsPreview.tsx
@@ -1,0 +1,64 @@
+import React from "react";
+import { Workspace } from "ui/types";
+
+const getIconAndText = (
+  isPublic: boolean,
+  selectedWorkspaceId: string | null,
+  workspaceName: string
+) => {
+  let text, icon;
+
+  if (!isPublic) {
+    if (selectedWorkspaceId) {
+      text = `Shared privately with ${workspaceName}`;
+      icon = "groups";
+    } else {
+      text = `Only you can view this`;
+      icon = "person";
+    }
+  } else {
+    text = "This replay can be viewed by anyone with the link";
+    icon = "public";
+  }
+
+  return { text, icon };
+};
+
+export default function SettingsPreview({
+  onClick,
+  isPublic,
+  workspaces,
+  selectedWorkspaceId,
+}: {
+  onClick: () => void;
+  isPublic: boolean;
+  workspaces: Workspace[];
+  selectedWorkspaceId: string | null;
+}) {
+  const personalWorkspace = { id: null, name: "My Library" };
+  const workspaceName = [...workspaces, personalWorkspace].find(w => w.id === selectedWorkspaceId)!
+    .name;
+
+  const { icon, text } = getIconAndText(isPublic, selectedWorkspaceId, workspaceName);
+
+  return (
+    <button
+      className="w-full flex flex-row justify-between items-center focus:outline-none"
+      onClick={onClick}
+      style={{ minHeight: "38px" }}
+    >
+      <div className="space-x-2.5 flex flex-row items-center">
+        <span className="material-icons" style={{ fontSize: "24px" }}>
+          {icon}
+        </span>
+        <div className="font-medium">{text}</div>
+      </div>
+      <div className="space-x-2 flex flex-row items-center text-primaryAccent">
+        <div className="font-medium">Edit</div>
+        <span className="material-icons" style={{ fontSize: "24px" }}>
+          edit
+        </span>
+      </div>
+    </button>
+  );
+}
--- a/src/ui/components/UploadScreen/SettingsPreview.tsx
+++ b/src/ui/components/UploadScreen/SettingsPreview.tsx
@@ -1,20 +1,24 @@
 import React from "react";
 import { Workspace } from "ui/types";
+import { MY_LIBRARY, personalWorkspace } from "./Sharing";
 
 const getIconAndText = (
   isPublic: boolean,
-  selectedWorkspaceId: string | null,
+  selectedWorkspaceId: string,
   workspaceName: string
-) => {
+): {
+  text: string;
+  icon: string;
+} => {
   let text, icon;
 
   if (!isPublic) {
-    if (selectedWorkspaceId) {
-      text = `Shared privately with ${workspaceName}`;
-      icon = "groups";
-    } else {
+    if ((selectedWorkspaceId = MY_LIBRARY)) {
       text = `Only you can view this`;
       icon = "person";
+    } else {
+      text = `Shared privately with ${workspaceName}`;
+      icon = "groups";
     }
   } else {
     text = "This replay can be viewed by anyone with the link";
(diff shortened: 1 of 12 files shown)


src/ui/components/UploadScreen/SettingsPreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (selectedWorkspaceId) {
+      text = `Shared privately with ${workspaceName}`;
+      icon = "groups";
+    } else {
+      text = `Only you can view this`;
+      icon = "person";
+    }
+  } else {
+    text = "This replay can be viewed by anyone with the link";
+    icon = "public";
```
jcmorrow, line 21: What do you think about introducing a type like `PrivacyMode`.
```
interface PrivacyMode {
  text: string,
  icon: string,
}

const PUBLIC = {
  text: "This replay can be viewed by anyone with the link"
  icon: "public"
}

const ONLY_YOU = {
  ...
}

const privacyMode(isPublic: boolean, workSpaceId: string | null, workspaceName: string): PrivacyMode => {
  if (isPublic) {
    return PUBLIC
  } else if (workSpaceId) {
    return {
    text: `Shared privately with ${workspaceName}`
    icon: "groups",
    }
  } 
​return PRIVATE
}
```

GitHub is hard to write code on, but hopefully you get what I mean?

src/ui/components/UploadScreen/SettingsPreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+export default function SettingsPreview({
+  onClick,
+  isPublic,
+  workspaces,
+  selectedWorkspaceId,
+}: {
+  onClick: () => void;
+  isPublic: boolean;
+  workspaces: Workspace[];
+  selectedWorkspaceId: string | null;
```
jcmorrow, line 36: Is this enough props that it feels worth pulling out into a separate type just to clean up this signature?

src/ui/components/UploadScreen/SettingsPreview.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { icon, text } = getIconAndText(isPublic, selectedWorkspaceId, workspaceName);
+
+  return (
+    <button
+      className="w-full flex flex-row justify-between items-center focus:outline-none"
+      onClick={onClick}
+      style={{ minHeight: "38px" }}
+    >
+      <div className="space-x-2.5 flex flex-row items-center">
+        <span className="material-icons" style={{ fontSize: "24px" }}>
```
jcmorrow, line 51: If we're going to use the `fontSize: 24px` rule a lot maybe we configure that in tailwind?

src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { Workspace } from "ui/types";
+import { Toggle } from "../shared/Forms";
+import SettingsPreview from "./SettingsPreview";
+
+type SharingProps = {
+  workspaces: Workspace[];
+  selectedWorkspaceId: string | null;
+  setSelectedWorkspaceId: Dispatch<SetStateAction<string | null>>;
+  isPublic: boolean;
+  setIsPublic: Dispatch<SetStateAction<boolean>>;
```
jcmorrow, line 13: I would `sort` these, but I know they are in logical order, so you do you!

src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+function EditableSettings({
+  workspaces,
+  selectedWorkspaceId,
+  setSelectedWorkspaceId,
+  isPublic,
+  setIsPublic,
+}: Omit<SharingProps, "showSharingSettings">) {
+  const updateDefaultWorkspace = hooks.useUpdateDefaultWorkspace();
+  const handleWorkspaceSelect = (id: string | null) => {
```
jcmorrow, line 24: Is this null only because of the `My Library` case? In that case, a null-value object like `DEFAULT_WORKSPACE` might be better. Maybe watch https://www.youtube.com/watch?v=OMPfEXIlTVE&t=1526s if you've never seen it? Sandi Metz is 💯

src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+    </div>
+  );
+}
+
+export default function Sharing(props: SharingProps) {
+  const [isEditing, setIsEditing] = useState(false);
+  const toggleIsEditing = () => setIsEditing(!isEditing);
+
+  let content;
```
jcmorrow, line 51: I know we've joked about ternaries recently, but this is one place where personally I kind of like them. If in a render function the ternary is literally `isEditing "? <A/> : <B/>` it feels very readable still (to me at least).

jcmorrow: In general, I think

```
<divs and stuff>
{renderOtherThing()}
</divs and stuff>
```
or, similarly, pre-calculating a ReactNode as a local and interpolating it in, feels like an anti-pattern to me. If that thing is complex enough that it's worth wrapping in its own method, make that method a component or inline it. Same thing with local variables.

src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <div>Public Access</div>
+        <Toggle enabled={isPublic} setEnabled={setIsPublic} />
+      </div>
+    </div>
+  );
+}
+
+export default function Sharing(props: SharingProps) {
+  const [isEditing, setIsEditing] = useState(false);
+  const toggleIsEditing = () => setIsEditing(!isEditing);
```
jcmorrow, line 49: I would also inline this since we're only using it once and the method name is as long as the method.

src/ui/components/UploadScreen/TeamSelect.tsx
```diff
@@ -11,13 +24,36 @@ export default function TeamSelect({
...
+      buttonContent={button}
+      setExpanded={setExpanded}
+      expanded={expanded}
+      buttonStyle=""
+      distance={0}
+      position="bottom-right"
+    >
+      <Dropdown widthClass="w-64">
+        <div className="max-h-48 overflow-auto">
+          {displayedWorkspaces.map(w => (
```
jcmorrow, line 50: what do you think about calling this `workspace`? `w` works fine when it's this simple, but if it ever grows or gets more variables in it it would be confusing. I usually only use one letter names when that letter is *super* well understood and the function is a fat-arrow, no brackets function. For example:

`onClick={(e) => e.preventDefault()}`

or

`people.map(p => p.name)`

I realize that this is a border-line case so ITYJ.

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -55,199 +53,41 @@ function DeletedScreen({ url }: { url: string }) {
...
-    } else {
-      text = `Only you can view this`;
-      icon = "person";
-    }
-  } else {
-    text = "This replay can be viewed by anyone with the link";
-    icon = "public";
-  }
-
+function ReplayScreenshot({ screenData }: { screenData: string }) {
```
jcmorrow, line 81: Huh... this got a lot simpler! I like that you've pushed this code off into discrete, meaningful places 👍

src/ui/components/UploadScreen/SettingsPreview.tsx
jaril: Added the typing just to make sure that we avoid situations where the functions ends up returning something we don't expect. Apart from that, not really a fan of extracting the return values into constants since it's not being reused.

PR #4091, opened 2021-10-21
More resilient highlighter and node picker (#3971)

hbenl:
The container node for the highlighter markup is removed when the user clicks "Hide Video" and recreated when he clicks "Show Video". This PR ensures that no errors are thrown when the video is hidden and node picker and highlighter always work when the video is shown.


src/devtools/client/inspector/inspector.ts
```diff
@@ -81,7 +81,7 @@ export class Inspector {
   panelDoc: Document | null;
   panelWin: Window | null;
   store: UIStore | null;
-  highlighter: any;
+  highlighter: typeof Highlighter;
```
jcmorrow, line 84: Nice, I like this direction for shimming in the older stuff much better than `any`.

src/highlighter/highlighter.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.boxModelHighlighter?.show(node);
+  }
+
+  unhighlight() {
+    this.maybeDestroyBrokenHighlighter();
+    this.boxModelHighlighter?.hide();
+    this.currentNode = null;
+  }
+
+  private maybeDestroyBrokenHighlighter() {
```
jcmorrow, line 39: This looks like it's wrapping some tricky edge-cases that you have run into. Personally, if I were to run into this code in the codebase without tests or a PR description about why it's here and how it works, I would be very scared to touch it for fear that I don't know the particular edge cases it's addressing. It seems like as a team Replay isn't super interested in documenting things via commit messages, but I wonder if there is a place where we could put this kind of knowledge/documentation?

src/highlighter/highlighter.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (!this.boxModelHighlighter) {
+      return;
+    }
+    if (!document.getElementById("box-model-root")) {
+      this.boxModelHighlighter.destroy();
+      this.boxModelHighlighter = undefined;
+    }
```
jcmorrow, line 46: both this and `maybeCreate` have two `if`s with no else clause. Is there a case where both return `false`? Is that expected?

src/highlighter/highlighter.ts
jcmorrow: Also, sorry if that is a stupid question, I don't know anything about how the node picker works 😅

PR #4092, opened 2021-10-21
Properly display breakpoints when they are not editable

jcmorrow:
Somewhere in this quagmire of conditionals (I'm still not sure exactly where) we had a bug where the breakpoint panel would say "This breakpoint is not editable because it was hit 700+ times" while also saying "This breakpoint was hit 350 times".

# Before

https://user-images.githubusercontent.com/5903784/138303272-c88d8a68-5df2-4298-ba54-9c751bf500cb.mp4

# After

https://user-images.githubusercontent.com/5903784/138303532-8540bacd-3d52-4a25-bba3-897b2485acff.mp4


jcmorrow:
Hey @jonbell-lot23 & @jaril can I get confirmation that this is the correct thing to show when a breakpoint is uneditable because of too many hits?

PR #4094, opened 2021-10-21
Track time-to-view-replay

jasonLaster:
This is close. The problem is that submitting kicks off a new navigation which clears the timings obj

https://app.replay.io/recording/1c185559-efb1-43a5-a706-7d91c3b608b4?point=15576890577250974698006545581474272&time=7782&hasFrames=false#


ryanjduffy:
@jasonLaster pushed a commit to this to cache timings in localstorage so we can pick it back up on the next page. probably some edge cases here to clean this up but should be a good start

ryanjduffy:
@jasonLaster I reverted the local storage silliness with #4238 eliminating the reload

ryanjduffy (approved):

PR #4200, opened 2021-10-27
Highlight tokens while the preview loads

colbyr:
Fixes https://github.com/RecordReplay/devtools/issues/3696

Highlight tokens in the source explore immediately, which the preview loads in the background. Before the token wouldn't be highlighted until the preview loaded.

Kapture 2021-10-27 at 15 30 55

The easy part was splitting up the highlight rendering and the popover rendering. Unfortunately that created some unfortunate race-conditions with the existing redux flow, e.g. where things would stay highlighted in some cases after you moved the mouse away or the highlight would get cleared out from under the popover.


colbyr:
@jcmorrow helped me looked through the test failures compared to other PRs and it seems like they aren't unique to these changes. Gonna give this a shot.

PR #4235, opened 2021-10-29
Track comment changes in Redux

colbyr:
In https://github.com/RecordReplay/devtools/issues/4042, we want to show a confirmation dialog if the user takes an action that would delete a comment they started writing. However, we don't want to show a prompt if they haven't typed anything yet.

The problem right now is that only the CommentEditor has the in progress comment's state, but the comment can be clear from many different places (e.g. [in the Timeline](https://github.com/RecordReplay/devtools/blob/386715395a9caf8d07ac62ae622d67d09bb17fdd/src/ui/components/Timeline/index.tsx#L166-L171)).

This change updates `pendingComment.comment.content` in the redux store when the comment editor loses focus. That way my next PR will be able to check if a comment has any content before prompting the user about deleting it.

In my thinking and testing, it seems to work ok, but obviously my knowledge is still pretty limited thoughts are very much welcome.

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/CommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/CommentEditor.tsx
@@ -4,7 +4,7 @@ import { selectors } from "ui/reducers";
 import { actions } from "ui/actions";
 import { UIState } from "ui/state";
 import hooks from "ui/hooks";
-import { Comment, isReply, PendingComment, Remark, Reply } from "ui/state/comments";
+import { Comment, PendingComment, Reply } from "ui/state/comments";
 
 import "./CommentEditor.css";
 import { User } from "ui/types";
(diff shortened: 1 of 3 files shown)


jcmorrow:
@colbyr The weird stuff is resolved by #4234 right? Or is there something else there?

jcmorrow:
I'm actually not convinced that this is a good idea. Here's my thinking:

- We currently only allow for one pending comment
- Because of that we sometimes do actions that have to clear that pending comment, and this introduces action-at-a-distance type problems.
- One way to solve that is to put more and more state into the global store, so that components everywhere can do destructive things more accurately, but IMO this is just adding complexity to make up for some previously added complexity.
- What if, instead, we just allowed people to have multiple pending comments? Then we could have *one* place where you discard a comment and it could live near that comment's component, and keep the state local.

Thoughts?

colbyr:
Closing this since we're going to skip the nags for now as we look forward to a future with multiple-pending comments!
https://discord.com/channels/779097926135054346/904840751127601265/904840753505771571

PR #4238, opened 2021-10-29
Avoid reload after uploading recording

ryanjduffy:
## Issue

The forced reload after upload was hampering tracking the user-facing time to upload and was extra work that generally made the app less responsive.

## Resolution

Use some callback plumbing to let the recording view know that upload has completed and the user can be redirected to devtools. Also refetches the recording query so we have the most recent data in memory.

### DIFF
--- a/src/views/recording.tsx
+++ b/src/views/recording.tsx
@@ -11,6 +11,7 @@ import DevTools from "ui/components/DevTools";
 function Recording({ getAccessibleRecording }: PropsFromRedux) {
   const recordingId = useGetRecordingId();
   const [recording, setRecording] = useState<RecordingInfo | null>();
+  const [uploadComplete, setUploadComplete] = useState(false);
   useEffect(() => {
     async function getRecording() {
       setRecording(await getAccessibleRecording(recordingId));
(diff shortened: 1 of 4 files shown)


src/views/recording.tsx
```diff
@@ -25,8 +26,8 @@ function Recording({ getAccessibleRecording }: PropsFromRedux) {
   // Add a check to make sure the recording has an associated user ID.
   // We skip the upload step if there's no associated user ID, which
   // is the case for CI test recordings.
-  if (recording.isInitialized === false && !isTest() && recording.userId) {
-    return <Upload />;
+  if (!uploadComplete && recording.isInitialized === false && !isTest() && recording.userId) {
```
jaril, line 29: This if statement is hairy but can't think of how to unpack it either

src/views/upload.tsx
```diff
@@ -4,24 +4,28 @@ import { useGetUserSettings } from "ui/hooks/settings";
...
   const recordingId = useGetRecordingId();
-  const { recording } = useGetRecording(recordingId);
+  const { recording, refetch } = useGetRecording(recordingId);
   // Make sure to get the user's settings before showing the upload screen.
   const { userSettings, loading } = useGetUserSettings();
 
   useEffect(() => {
     if (recording?.isInitialized) {
       window.onbeforeunload = null;
-      document.location.reload();
```
jaril, line 16: I'm not 100% sure, but will `recording.isInitialized` ever be `true` before we call `refetch()`?

If it is true, that means that we've updated the recording in the cache. Do we need to still call refetch then?

jaril: Just tested it and seems to work. 😳

jaril (approved):
This appears to work, just not 100% sure about the refetch logic.

jaril (approved):
Yeah let's drop the `refetch`, and feel free to merge afterwards. 👍

PR #4240, opened 2021-10-30
Polish paused and playing UI states and transitions

jcmorrow:
### Summary

- Always show the breakpoint navigation status, even when playing the video. This reduces how much the UI jumps. I don't think it's *too* jarring that the point count doesn't go up as you play. If it is we can always sub in a `-` there or something.
- Always keep the max number of characters worth of space for the navigation status. If this thing changes its width (because we go from point `9 -> 10` or `99 -> 100`), then it ends up changing the width of the timeline as well, which makes the whole UI subtly shift.

https://user-images.githubusercontent.com/5903784/139517858-92452e8a-f41b-4e67-97df-55a63f25a3c7.mp4

- Same thing for the overall timeline time. This one is a lot subtler because the number of digits is constant, but since it's not a fixed-width font the change can result in subtle motion in the corner. This tries to reduce that motion as much as possible:

https://user-images.githubusercontent.com/5903784/139518313-6f927422-7e8a-4093-ac34-5172302b93d0.mp4

- Don't let codemirror interpret left or right arrows as horizontal scrolling. The left and right arrows should default to being timeline nudges. This is a super useful feature, and it's often quite subtle since you are going through one paint at a time, so you can pretty easily hit this key like ten times without realizing it's not doing anything, which is a frustrating experience.
- Don't let the browser interpret `spacebar` as vertical scrolling - similar to the last one, space bar universally means *scroll down* in documents. However, in media players and similar apps, space bar typically means play/pause. That's a much more natural meaning for Replay, so we short-circuit the `body`s normal handler, which will try and scroll things down vertically (I think it scrolls the either the active element or the element under cursor, which often results in the codemirror editor getting scrolled).
- Don't consider `readonly` text inputs to be editable. We have some code that implements the "toggle playing on spacebar logic". It relies on a helper to not activate when the user is typing in, for instance, a comment field. However, the way that codemirror works is that it focuses an invisible `textarea` tag, so that it can manage focus like a normal textfield. When you put it in `readonly` mode it is... well... read-only, and in that case, we want that `play/pause` behavior, even if codemirror is the currently focused element, so I've adjusted that helper function to only consider inputs and text areas to be editable when they are not read-only.
- I've also added a default to `lastExecutionPoint` of `0`. I'm somewhat neutral on this. My general reason for this is that it's nice to have type consistency across all states, so if the null-ish object for that type makes sense, initialize with that, rather than `null` and then have to check for `null` everywhere. This doesn't always work but it seems like it makes sense here.

Fixes https://github.com/RecordReplay/devtools/issues/3773
Fixes https://github.com/RecordReplay/devtools/issues/4137

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -347,6 +347,7 @@ class Timeline extends Component<PropsFromRedux> {
     const percent = getVisiblePosition({ time: currentTime, zoom: zoomRegion }) * 100;
     const hoverPercent = getVisiblePosition({ time: hoverTime, zoom: zoomRegion }) * 100;
     const precachedPercent = getVisiblePosition({ time: precachedTime, zoom: zoomRegion }) * 100;
+    const formattedTime = getFormattedTime(currentTime);
 
     return (
       <div className="timeline">
@@ -380,8 +381,11 @@ class Timeline extends Component<PropsFromRedux> {
           </div>
           <Tooltip timelineWidth={this.overlayWidth} />
         </div>
-        <div className="timeline-time">
-          <span className="time-current">{getFormattedTime(currentTime)}</span>
+        <div
+          className="timeline-time text-right"
+          style={{ minWidth: `${formattedTime.length * 2 + 2}ch` }}
+        >
+          <span className="time-current">{formattedTime}</span>
           <span className="time-divider">/</span>
           <span className="time-total">{getFormattedTime(recordingDuration || 0)}</span>
         </div>
--- a/src/ui/utils/key-shortcuts.ts
+++ b/src/ui/utils/key-shortcuts.ts
@@ -73,13 +73,15 @@ export function cancelBubbling(listener: KeyboardEventListener): KeyboardEventLi
   };
 }
 
+function isEditableTag(target: HTMLElement) {
+  return ["INPUT", "TEXTAREA"].includes(target.tagName) && target.getAttribute("readonly") === null;
+}
+
 export function isEditableElement(target: EventTarget) {
-  if (target instanceof HTMLElement) {
-    if (target.tagName === "INPUT" || target.tagName === "TEXTAREA" || target.isContentEditable) {
-      return true;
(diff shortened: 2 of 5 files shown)


src/ui/components/Timeline/index.tsx
```diff
@@ -380,8 +381,11 @@ class Timeline extends Component<PropsFromRedux> {
           </div>
           <Tooltip timelineWidth={this.overlayWidth} />
         </div>
-        <div className="timeline-time">
-          <span className="time-current">{getFormattedTime(currentTime)}</span>
+        <div
+          className="timeline-time text-right"
+          style={{ minWidth: `${formattedTime.length * 2 + 2}ch` }}
```
jasonLaster, line 386: Nice, i had no idea you could do this :P

src/ui/utils/key-shortcuts.ts
```diff
@@ -73,13 +73,17 @@ export function cancelBubbling(listener: KeyboardEventListener): KeyboardEventLi
   };
 }
 
+function isEditableTag(target: HTMLElement) {
+  return (
+    ["INPUT", "TEXTAREA"].includes(target.tagName) && !(target.getAttribute("readonly") !== null)
+  );
```
jasonLaster, line 79: wow - this is super smart. I wouldn't have thought of going into `key-shortcuts`, but it's obvious in hindsight

jasonLaster (commented):
Love the writeup, it makes it much easier to follow. 

Also, really impressed that you're making the code better as you go. It's great to see the debt go down.

src/ui/components/Timeline/index.tsx
jcmorrow: I also had forgotten about this unit of measure 😄

src/ui/utils/key-shortcuts.ts
jcmorrow: Also I'm just realizing this double negation is completely unreadable, what was I thinking? 😆

PR #4241, opened 2021-10-30
Rework breakpoint panel

jcmorrow:
There was a lot of unused CSS here, and also a lot of CSS that was sort of circuitous. I found the best approach was to rip most of it out and then rebuild. When you have click targets that extend outside of their elements and are also close to each other the way that you mark up space becomes *very* important. For instance, the difference between:

CleanShot 2021-10-29 at 23 01 51

Where the whole background is a click target for A is *very* different structurally from:

CleanShot 2021-10-29 at 23 02 47

Where the top half is a click target for B and the bottom half is a click target for A. But once you get that structure right, some things become easier. For instance, B is no longer a child of A's click target, so you can remove some `preventDefault` style code, which is a sign that you're moving with the grain rather than against it.

But getting that design right with a minimum of changes as you snap back and forth between editor mode is *hard*. I still haven't eliminated all movement, but I think this is as close to pixel-perfect as it is worth pushing this particular iteration. I think the long term vision for this component should actually be to remove CodeMirror from it. Don't get me wrong, CodeMirror is awesome, but every time we render a conditional break-point we are instantiating two full-blown code editors in JS, when what we *really* want is just a text input tag with some syntax highlighting. Even on a fast computer, you can *feel* the DOM rebuilding itself as it transitions in and out of editor mode. It sort of feels like trying to drive a nail in with a jackhammer. It's not that it won't work, it just doesn't feel like the right tool for the job.

### Before

https://user-images.githubusercontent.com/5903784/139524082-e6fb541b-2aa0-4b11-909d-88554c92d8e3.mp4

### After

https://user-images.githubusercontent.com/5903784/139522602-112a2f67-718b-4504-bdd2-cf9b07554954.mp4

Fixes https://github.com/RecordReplay/devtools/issues/4119 except for the part about keeping console statements around, which was a "nice to have". Let's make a follow-on issue for that that we can pick up as part of console work!

### DIFF
--- a/src/devtools/client/debugger/src/actions/types/ASTAction.js
+++ b/src/devtools/client/debugger/src/actions/types/ASTAction.js
@@ -1,5 +1,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
-
-//
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Condition.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Condition.tsx
@@ -1,14 +1,15 @@
 import React from "react";
-import { Input } from ".";
 import { SummaryExpression, SummaryExpressionProps } from "./SummaryExpression";
 import SummaryRow from "./SummaryRow";
 
 type ConditionProps = SummaryExpressionProps & {
-  handleClick: (event: React.MouseEvent, input: Input) => void;
+  onClick: () => void;
 };
 
-export default function Condition({ handleClick, ...rest }: ConditionProps) {
-  const expression = <SummaryExpression {...rest} handleClick={e => handleClick(e, "condition")} />;
-
-  return <SummaryRow label={"if"}>{expression}</SummaryRow>;
+export default function Condition({ onClick, ...rest }: ConditionProps) {
+  return (
+    <SummaryRow onClick={onClick} label={"if"}>
+      <SummaryExpression {...rest} />
+    </SummaryRow>
+  );
 }
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Popup.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Popup.tsx
@@ -9,7 +9,7 @@ interface PopupProps {
 export default function Popup({ trigger, children }: PopupProps) {
   return (
     <ReactPopup
-      trigger={trigger}
+      trigger={<>{trigger}</>}
       on="hover"
       position="right center"
       mouseEnterDelay={200}
--- a/package-lock.json
+++ b/package-lock.json
@@ -86,6 +86,7 @@
         "@recordreplay/playwright": "^1.10.3",
         "@recordreplay/recordings-cli": "^0.3.1",
         "@types/classnames": "^2.2.11",
+        "@types/codemirror": "^5.60.5",
         "@types/escape-html": "^1.0.1",
         "@types/lodash": "^4.14.168",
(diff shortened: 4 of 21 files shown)


jonbell-lot23:
This is some top notch documentation/rationale, thank you!

I notice the "after" movie seems to have more delay than the "before" movie. Are they flipped around?

jcmorrow:
> This is some top notch documentation/rationale, thank you!
> 
> I notice the "after" movie seems to have more delay than the "before" movie. Are they flipped around?

Actually, the "Before" seemed faster because I had already run with that condition before, so the results were coming back faster. I've swapped that out for what it looks like to try a new condition on prod right now (which involves the `if` field temporarily disappearing and then popping back in later).

src/devtools/client/debugger/src/actions/types/ASTAction.js
```diff
@@ -1,5 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
-
-//
```
jasonLaster, line 5: Any reason why you're deleting this? I think we should keep the license for code that comes from moz. New files do not need it though.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Condition.tsx
```diff
@@ -1,14 +1,15 @@
...
-export default function Condition({ handleClick, ...rest }: ConditionProps) {
-  const expression = <SummaryExpression {...rest} handleClick={e => handleClick(e, "condition")} />;
-
-  return <SummaryRow label={"if"}>{expression}</SummaryRow>;
+export default function Condition({ onClick, ...rest }: ConditionProps) {
+  return (
+    <SummaryRow onClick={onClick} label={"if"}>
+      <SummaryExpression {...rest} />
+    </SummaryRow>
+  );
```
jasonLaster, line 14: Smart

jasonLaster (commented):
Looks great!

src/devtools/client/debugger/src/actions/types/ASTAction.js
jcmorrow: Oh yeah I meant to comment on this. All of these files were empty except for the license, so I deleted them, but lmk if they are actually serving a purpose and I can restore them!

jasonLaster:
RM

jcmorrow:
> Looks great! The only thing I'd call out is the padding for the form in the new UI.

Yeah, I also don't love it. I opted for less-jumping, but that means that either the breakpoint panel when in non-form mode has to have a lot of padding, or when in form mode it hast to have less. Perhaps I should try and find a happier medium?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/Popup.tsx
```diff
@@ -9,7 +9,7 @@ interface PopupProps {
 export default function Popup({ trigger, children }: PopupProps) {
   return (
     <ReactPopup
-      trigger={trigger}
+      trigger={<>{trigger}</>}
```
jcmorrow, line 12: This might look weird, but it's to deal with the `function components cannot be given refs` error. I'm not 100% clear on why this solves it, but it does. Perhaps somebody else can better explain this?

jcmorrow:
Oh and on the "long-term vision" of moving away from something as heavy as codemirror for this:[ TipTap has a syntax highlighting extension](https://tiptap.dev/examples/syntax-highlighting) 😄

jasonLaster:
Noted - it's all trade-offs 😉

jonbell-lot23:
Just adding a historical note that I agree the spacing improvements would be nice.

jcmorrow:
I have added a *touch* more spacing, just to make the form feel a little less cramped. It's definitely not perfect, but feels a bit better, and there is no "jump" in spacing whether you have an if condition or not anymore 😄

PR #4254, opened 2021-11-01
Fix the react_devtools e2e test

hbenl:
The test measures a highlighted element, but that element's width depended on the width of the browser window while recording. Now I set the width explicitly.
The `react_devtools` test uses a CRA app for its example recording. To update the CRA app, change `test/examples/cra/src/App.js` and then run `npm run build-cra-e2e-test`.
I have also added some bits to the documentation for running our tests.


jasonLaster (approved):

test/scripts/react_devtools.js
```diff
@@ -26,7 +26,7 @@ Test.describe(`Test React DevTools.`, async () => {
 
   rootComponent.dispatchEvent(new MouseEvent("mouseover", { bubbles: true }));
   await Test.checkHighlighterVisible(true);
-  await Test.checkHighlighterShape("M40,16 L1280,16 L1280,35 L40,35");
+  await Test.checkHighlighterShape("M40,16 L140,16 L140,35 L40,35");
```
jcmorrow, line 29: I might be mistaken here, because I know there are sometimes multiple failures, but if this is the failure that I'm *thinking* of it usually fails without giving any indication that it's a numerical expectation that was failing. Would it be possible for us to move towards a world where failing tests give more of an explicit output like Jest expectations do? Something like:

`expecting 140, got 1280 on line 29 of react_devtools.js` would go a *long* way.

jcmorrow (approved):
I left a long term suggestion that I'd love to get your thoughts on, looks great, thanks for fixing!

PR #4265, opened 2021-11-01
Add a reply button

jcmorrow:
This is the latest iteration in the "when to show/focus reply editors" saga. I like this, it feels like a nice compromise, at least after kicking the tires locally for a few minutes. I think the change should all read pretty straight-forwardly, with one exception: I've made the reply button 50% width. I did this because `auto` made the click target pretty small, but 100% still felt like you could easily open and focus the editor when you were actually intending to navigate to that comments point in the timeline instead. One other nice way to make this clear to the user might be to add a ⏩  or ⏪  icon on the left side of the comment when it is hovered and clicking would cause you to navigate on the timeline, but I will leave that for another time.

### Preview

https://user-images.githubusercontent.com/5903784/139725302-f567ec06-2e7d-4536-a327-c0a4c24c078b.mp4

Fixes https://github.com/RecordReplay/devtools/issues/4242

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
@@ -164,11 +164,7 @@ function CommentCard({
         hoveredComment === comment.id ? "bg-toolbarBackground" : "bg-white"
       )}
       onMouseDown={e => {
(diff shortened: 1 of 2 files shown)


jasonLaster:
lets remove the focus state

PR #4266, opened 2021-11-01
Make the timeline full width

### LINKED ISSUE #4250: Make the global timeline full width
**Description:**
We would like the global timeline to always be full width. We can do that now as a prep-step for https://github.com/RecordReplay/devtools/issues/4159.

Jon exploration DevTools

jaril:
Fix #4250.

image
image


colbyr (approved):
Just tried the vercel preview. Looks awesome 👏 👏

PR #4268, opened 2021-11-02
Allow embedded links to be copied and pasted

### LINKED ISSUE #4258: copy and pasting a smart link should work replay

jcmorrow:
There were a couple of weird issues with the regexes we were using when they were pasted in the middle of content. This is probably a rare case but it's nice to handle it correctly. We will now not match a malformed Replay URL, which should help users realize that their link isn't fully copied in the case that they try and paste an incomplete URL. We also render text correctly, so you can paste *from* a replay comment and get the link back. For example:

https://user-images.githubusercontent.com/5903784/139773469-4d3ae80a-b5e1-4d5a-92fc-c124257a180e.mp4

Copying a smart link from one comment and pasting it into another comment now *also* works. The problem here was that I had set up our `parseHTML` matcher to `a[href]`, which means that if you pasted *any* HTML that had a valid link in it, TipTap would try and interpret that as a GitHub link. Now, we use `a[data-github-link]` and when we `renderHTML` we make sure that all github links have that data attribute (and Replay links get a similar `data-replay-link` attribute). Again, the mention extension was  indispensable in figuring out why this was broken. I find that reading other extension code almost always answers my TipTap questions.

Fixes #4258

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/replayLink.ts
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/replayLink.ts
@@ -5,7 +5,7 @@ export interface ReplayLinkOptions {
   HTMLAttributes: Record<string, any>;
 }
 
-export const inputRegex = /(https:\/\/app.replay.io\/recording\/(.*)\?.*)/g;
+export const inputRegex = /(https:\/\/app.replay.io\/recording\/([a-zA-z0-9\-]{36})[^ ]*)/g;
 
 const getAttributes = (match: string[]) => {
   const [href, , recordingId] = match;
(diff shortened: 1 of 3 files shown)


src/ui/components/Comments/TranscriptComments/CommentEditor/replayLink.ts
```diff
@@ -5,7 +5,7 @@ export interface ReplayLinkOptions {
   HTMLAttributes: Record<string, any>;
 }
 
-export const inputRegex = /(https:\/\/app.replay.io\/recording\/(.*)\?.*)/g;
+export const inputRegex = /(https:\/\/app.replay.io\/recording\/([a-zA-z0-9\-]{36})[^ ]*)/g;
```
jasonLaster, line 8: wow - such regexes.

jasonLaster (approved):
Initially i thought we should consolidate these, but they seem different enough to warrant their own components

colbyr (approved):
Nice!

PR #4270, opened 2021-11-02
Fix the detection of comment marker clicks (#4269)

hbenl:
The logic for detecting whether a Timeline click was on a comment marker wasn't working reliably. It used the `hoveredComment` from the redux store, but that mechanism is currently broken (sometimes `hoveredComment` is set even though we're not hovering over a comment) and seems brittle anyway, so I opted for a more robust mechanism.

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -125,7 +125,10 @@ class Timeline extends Component<PropsFromRedux> {
       setTimelineState,
       isTrimming,
     } = this.props;
-    const hoveringOverMarker = !!hoveredComment;
+    // if the user clicked on a comment marker, we already seek to the comment's
+    // execution point, so we don't want to seek a second time here
+    const clickedOnCommentMarker =
+      e.target instanceof Element && [...e.target.classList].includes("comment-marker");
     const mouseTime = this.getMouseTime(e);
 
     trackEvent("timeline.progress_select");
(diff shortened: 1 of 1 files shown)


src/ui/components/Timeline/index.tsx
```diff
@@ -125,15 +125,18 @@ class Timeline extends Component<PropsFromRedux> {
       setTimelineState,
       isTrimming,
     } = this.props;
-    const hoveringOverMarker = !!hoveredComment;
+    // if the user clicked on a comment marker, we already seek to the comment's
+    // execution point, so we don't want to seek a second time here
+    const clickedOnCommentMarker =
```
jcmorrow, line 130: I'm very glad that we're not using the redux state anymore, because that *does* seem super brittle, but I'm a little confused as to why this click handler isn't *on* the element that we're checking for? Wouldn't a more elegant way of doing this same check be to set the onClickHandler on the comment markers themselves?

jcmorrow (approved):
I left a related comment, but if that's a lot of work to implement I support documenting the better option as a future change to make in the commit/PR message and then just merging this for now so we can get something better out! 👍

src/ui/components/Timeline/index.tsx
hbenl: We have an onClick handler on the comment markers themselves (which does almost the same thing), but we also need this handler when the user doesn't click on a comment marker and we want to prevent that both handlers run for the same click.

PR #4277, opened 2021-11-02
fix property highlighting in some cases

colbyr:
Fixes https://github.com/RecordReplay/devtools/issues/3913

### Before
Screen Shot 2021-11-02 at 2 58 08 PM

### After
Screen Shot 2021-11-02 at 2 57 46 PM

### DIFF
--- a/src/devtools/client/debugger/src/utils/editor/get-expression.js
+++ b/src/devtools/client/debugger/src/utils/editor/get-expression.js
@@ -32,7 +32,7 @@ export function getExpressionFromCoords(cm, coord) {
   while (startHighlight > 1 && line.charAt(startHighlight - 1) === ".") {
     const tokenBefore = tokenAtTextPosition(cm, {
       line: coord.line,
-      column: startHighlight - 2,
+      column: startHighlight - 1,
     });
 
     if (!tokenBefore || !tokenBefore.type) {


src/devtools/client/debugger/src/utils/editor/get-expression.js
```diff
@@ -32,7 +32,7 @@ export function getExpressionFromCoords(cm, coord) {
   while (startHighlight > 1 && line.charAt(startHighlight - 1) === ".") {
     const tokenBefore = tokenAtTextPosition(cm, {
       line: coord.line,
-      column: startHighlight - 2,
```
colbyr, line 35: At first glance this seemed right to me, but I guess it's an off-by-one error. It only shows up if the first variable in a property chain is one character long.

```js
// this works fine
one.two.three

// but this wouldn't highlight
o.two.three
```

☝️  the off-by-one would skip the `o` and hit the `" "` and cause us to bail out further down the loop.

https://github.com/RecordReplay/devtools/blob/dbdf738bdb3509165820b4dc5bd39145091c875b/src/devtools/client/debugger/src/utils/editor/get-expression.js#L38-L40

jcmorrow: Is it possible for this to end up with a negative index? Or, asked another way, can `startHighlight` ever be `1`?

bhackett1024: > Is it possible for this to end up with a negative index? Or, asked another way, can `startHighlight` ever be `1`?

The clause of the `while` loop ensures `startHighlight > 1` I think.

bhackett1024 (approved):
This LGTM, though I'm not familiar with this code.

colbyr:
@bhackett1024 ah these lines blame to you but upon closer inspection it looks like it was some mass file moving. Thank you for taking a look!

src/devtools/client/debugger/src/utils/editor/get-expression.js
colbyr: ☝️  yep what Brian said

jasonLaster: @colbyr I wonder if we can re-enable these unit tests https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/utils/editor/tests/get-expression.spec.js

cc @jcmorrow who might know how. Perhaps change it to `.test.js`

jcmorrow: > The clause of the while loop ensures startHighlight > 1 I think.

🤦  yep, good call!

jcmorrow: And yeah, it might be possible to enable those tests, but you'll have to manually lift them out of `src/devtools`, as `jest` is currently set to ignore that whole folder (and they should be renamed to match the naming scheme as Jason mentioned).

jaril:
My only nit here is that I would expect the "." to be highlighted as well.

src/devtools/client/debugger/src/utils/editor/get-expression.js
jasonLaster: perhaps we can allow `src/devtools` but it has to use the new naming scheme...

colbyr: sure I can do that! I didn't even notice there was a test in there. thanks @jasonLaster

colbyr: https://github.com/RecordReplay/devtools/pull/4290 this will let jest run `.test.js` files in src/devtools.

Getting the [get-expression tests](https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/utils/editor/tests/get-expression.spec.js) might be a different story though... CodeMirror doesn't seem to work very well in the jsdom environment 🙈

jasonLaster: @colbyr weird - this is what we had before... jest + jsdom, but yeah - feel free to punt if this is a rabbit hole

colbyr: Yeah I think this is beyond me right now. I spent a few minutes trying to get them running, but for whatever reason the outputs seem to be different than they are in the browser environment and they're failing with or without my code changes in this PR.

One of the failing get-expression tests

Idk how long its been since these tests were running, but maybe something in jsdom/codemirror/jest has changed since then. If anyone is hankering to get into it, I put up a branch where I was at: https://github.com/RecordReplay/devtools/compare/master...colbyr:bring-back-get-expression-tests

jasonLaster: cool - lets skip it.

colbyr:
Following up on my discord convo with Jaril, we're definitely in agreement that what gets highlighted is kind of weird, but actual highlighting happens [elsewhere](https://github.com/RecordReplay/devtools/blob/5aba7e50fe75c463aa50b0224fb7de1aeea72488/src/devtools/client/debugger/src/components/Editor/Preview/PreviewHighlight.tsx#L11-L47). It's probably a problem for a different day as well.

PR #4290, opened 2021-11-03
allow tests in src/devtools

colbyr:
As requested in https://github.com/RecordReplay/devtools/pull/4277#discussion_r741474380

After this change jest will run tests in `src/devtools` as long as they end in `*.test.js`.

Jest still won't run the old tests which end in `.spec.js`.


jest.config.js
```diff
@@ -2,6 +2,11 @@ module.exports = {
   moduleNameMapper: {
     "ui/(.*)": "<rootDir>/src/ui/$1",
   },
-  // If you want to run all of the old tests then remove this ignore pattern
-  modulePathIgnorePatterns: ["src/devtools"],
+  testEnvironment: "jsdom",
+  testMatch: [
+    "**/__tests__/**/*.[jt]s?(x)",
+    "**/?(*.)+(test).[jt]s?(x)",
```
colbyr, line 8: ☝️  this is jest's [default test matcher](https://jestjs.io/docs/configuration#testmatch-arraystring), but without `.spec.js`

jasonLaster (approved):

jest.config.js
jcmorrow: Might be worth removing my comment below and adding a comment pointing to this PR to explain the reasoning.

colbyr:
@jcmorrow yeah I was just gonna ask if it was cool to blow away those snapshots 👍

Issue #4289, opened 2021-11-03
Timeline inconsistency


hbenl:
https://app.replay.io/recording/ea5b079c-9ca8-4ff8-9087-a22a2302d15c

In this recording, seeking from the first to the second comment moves the pause marker in the console to an _earlier_ message. The second comment has a time that is larger than the first comment's time (that's why it's shown _after_ the first comment), but its executionPoint is smaller than the first comment's (that's why the pause marker moves to an _earlier_ message).

hbenl:
Whilte #4291 was technically correct, it makes Replay look broken: in many recordings there's a big gap between the first two paints and clicking in the timeline in this gap always jumps to the beginning of the recording.
We need a way to get an execution point for a given time (or very close to that time) without relying on paints and mouse events only.

hbenl:
See also #4350

jasonLaster:
re-upping Holger's point

> We need a way to get an execution point for a given time (or very close to that time) without relying on paints and mouse events only.

jaril:
@bvaughn With the new timeline, mind checking if this still applies??

bvaughn:
Don't think my timeline work would have impacted this?

jasonLaster:
Yep - this is different

@hbenl added this to the next sprint. Feel free to punt, but feels like an important "broken glass" issue that we should fix on the backend.

jcmorrow:
>  We need a way to get an execution point for a given time (or very close to that time) without relying on paints and mouse events only.

💯 

This would be useful in general as well, not just for this problem.

gideonred:
This might be fixed in https://github.com/RecordReplay/backend/pull/5565.

jcmorrow:
@hbenl I've put my face on this because it is closely tied to #6820, I hope that's ok!

PR #4291, opened 2021-11-03
Fix timeline inconsistency (#4289)

hbenl:
When the user clicks on the Timeline, we look for the last paint or mouse event before that point in time and seek to the execution point of that event. But we only used the execution point from that event, not the time. This led to an inconsistent time/executionPoint pair used in the frontend (and also saved in comments).

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -139,7 +139,7 @@ class Timeline extends Component<PropsFromRedux> {
     if (hoverTime != null && !clickedOnCommentMarker) {
       const event = mostRecentPaintOrMouseEvent(mouseTime);
       if (event && event.point) {
-        if (!seek(event.point, mouseTime, false)) {
+        if (!seek(event.point, event.time, false)) {
           // if seeking to the new point failed because it is in an unloaded region,
           // we reset the timeline to the current time
           setTimelineToTime(ThreadFront.currentTime, true);


jasonLaster:
Can you share a Replay of the fix so we can see the differeence

jasonLaster:
fixes #4289

hbenl:
> Can you share a Replay of the fix so we can see the differeence

Go to https://devtools-bodv02ru4-recordreplay.vercel.app/recording/f57eaa99-e541-43a3-bca4-9b61a3478771 and click anywhere in the Timeline between the markers for the first and second comment - it will send you to a time before the first comment.
If you try the same thing in https://app.replay.io/recording/f57eaa99-e541-43a3-bca4-9b61a3478771, it will seemingly send you to the time you selected, but in reality it uses an execution point before the first comment (this is also how I placed the second comment, so the inconsistent time/point combo is saved in that comment, causing the weird behaviour that you can see in https://app.replay.io/recording/ea5b079c-9ca8-4ff8-9087-a22a2302d15c - watch how the pause marker in the console moves when selecting comments).
This is an extreme example because in that recording, the browser couldn't paint for several seconds, so going back to the last paint/mouse event makes a big jump in time.

bhackett1024 (approved):

bhackett1024:
Thanks!  It's great that the execution point and timeline time can be consistent now.

PR #4292, opened 2021-11-03
consolidate css

jasonLaster:
This simplifies the Next migration considerably


src/ui/components/Header/ViewToggle.tsx
```diff
@@ -114,7 +114,7 @@ function ViewToggle({
 
   return (
     <AnimateSharedLayout type="crossfade">
-      <button className="view-toggle" type="button">
+      <div className="view-toggle" role="button">
```
jcmorrow, line 117: This change is because of line 93 in `common.css`:

```
/* Devtools rely on the old Gecko/HTML4 button hit testing */
html|button > * {
  pointer-events: none;
}
```

I don't understand that line, Chrome doesn't even seem to think it's a valid selector in prod, but for some reason it *does* pick it up in this branch locally? I'm not a huge fan of nesting content inside of buttons like this anyways, but I think maybe we should still delete that rule unless it's playing some important role that I don't understand.

jasonLaster: weird. such a strange line

Issue #4224, opened 2021-10-29
Elements panel performance


hbenl:
When selecting an element in the Elements panel, it can take up to several seconds for the panel to update. It seems that (unlike in the past) this performance problem has nothing to do with backend performance, but is a frontend issue.

hbenl:
I created a branch that only generates the redux state for the computed style panel only when necessary: https://github.com/hbenl/webreplay-devtools/tree/issue-4224.
But the performance improvement was much smaller than I expected, so I dug deeper and found that the DOM for the markup view is updated almost immediately but the browser doesn't render the updated DOM until the computation of the redux state for the side panels is finished. This is because the computation is almost synchronous: there are some `await`s in there, but they wait for promises that are already resolved and waiting for a resolved promise doesn't unblock browser rendering.

hbenl:
Fixed by #4295

PR #4295, opened 2021-11-04
Improve Elements panel performance (#4224)

hbenl:
This PR contains 4 independent changes to improve the performance of the Elements panel:
- adds a tiny delay before updating the style panels to give the browser an opportunity to render the updated markup view
- removes the `_hasUpdatedCSSVariable` check in the `ElementStyle` class that had no effect
- adds a cache to the `shortSource` function
- adds 2 redux actions with a large payload to the list of actions that should not be sanitized

### DIFF
--- a/src/devtools/client/inspector/rules/models/element-style.ts
+++ b/src/devtools/client/inspector/rules/models/element-style.ts
@@ -399,12 +399,13 @@ export default class ElementStyle {
       // overridden state has changed for the text property.
       // _hasUpdatedCSSVariable will return true if the declaration contains any
       // of the updated CSS variable names.
-      if (
-        this._updatePropertyOverridden(textProp) ||
-        this._hasUpdatedCSSVariable(textProp, changedVariableNamesSet)
-      ) {
-        textProp.updateEditor();
-      }
+      // if (
+      //   this._updatePropertyOverridden(textProp) ||
+      //   this._hasUpdatedCSSVariable(textProp, changedVariableNamesSet)
+      // ) {
+      //   textProp.updateEditor();
+      // }
+      this._updatePropertyOverridden(textProp);
 
       // For each editor show or hide the inactive CSS icon as needed.
       // if (textProp.editor && this.unusedCssEnabled) {
--- a/src/devtools/client/inspector/rules/rules.js
+++ b/src/devtools/client/inspector/rules/rules.js
@@ -578,6 +578,11 @@ class RulesView {
    * to be refresh (e.g. when print media simulation is toggled).
    */
   async updateElementStyle() {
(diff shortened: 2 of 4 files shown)


src/devtools/client/inspector/rules/models/element-style.ts
```diff
@@ -399,12 +399,13 @@ export default class ElementStyle {
...
-        this._hasUpdatedCSSVariable(textProp, changedVariableNamesSet)
-      ) {
-        textProp.updateEditor();
-      }
+      // if (
+      //   this._updatePropertyOverridden(textProp) ||
+      //   this._hasUpdatedCSSVariable(textProp, changedVariableNamesSet)
+      // ) {
+      //   textProp.updateEditor();
+      // }
```
jcmorrow, line 407: What do you think about just deleting this code? Do you think it's very likely that we'll uncomment it soon?

src/devtools/client/inspector/rules/rules.js
```diff
@@ -578,6 +578,11 @@ class RulesView {
    * to be refresh (e.g. when print media simulation is toggled).
    */
   async updateElementStyle() {
+    // Updating the style panels can take more than 100ms and blocks the browser from
```
jcmorrow, line 581: 100ms! That's a pretty long time! What's so heavy about the code this function calls?

jcmorrow: Also, and I am no expert on how this works, but what if we instead did:

```
this.elementStyle.populate().then(() => this.updateRules())
```

without the `await`. Would that have a similar effect or am I missing something?

jcmorrow (approved):
Couple of questions just for my own edification, but looks good! 🎨

src/devtools/client/inspector/rules/models/element-style.ts
hbenl: You're right, this was just a bad old habit.

src/devtools/client/inspector/rules/rules.js
hbenl: No, we need the `setTimeout` in there because awaiting a resolved promise doesn't unblock the rendering. And (AFAIK) it doesn't make a difference if you use `await` or `.then()`.

hbenl: > 100ms! That's a pretty long time! What's so heavy about the code this function calls?

Yeah, that's a mystery to me too. I found and fixed (in this PR) 2 places that slowed the computation down a lot, but it's still slower than I would expect it to be. I'm pretty sure if I spent more time on this, I could find other places to speed up.

Issue #3852, opened 2021-10-04
Improve full text search


jasonLaster:
### Discussed in https://github.com/RecordReplay/devtools/discussions/3766

<div type='discussions-op-text'>

<sup>Originally posted by **jasonLaster** September 25, 2021</sup>
**Description:**

Full project search is currently pretty slow because we have to first fetch all the files and then search them in a worker. We could speed this up by doing the search on the backend with ripgrep and streaming the results.</div>

### Todos
- [x] #4427
- [x] #4428
- [x] #4429
- [x] #4430
- [x] https://github.com/RecordReplay/devtools/issues/4573
- [x] https://github.com/RecordReplay/devtools/issues/4585
- [x] https://github.com/RecordReplay/devtools/issues/4597
- [x] Remember search history
- [x] select result with arrow keys
- [x] left arrow key should collapse section
- [x] down arrow from input should go to results
- [x] https://github.com/RecordReplay/backend/issues/3668
- [ ] Add include / exclude path query parameter

### Nice to have
- [ ] option to search node_modules
- [ ] matchCase, regex, word break

jasonLaster:
Small, mostly marketing change. I think we should refer to this as Full Text Search

PR #4297, opened 2021-11-04
Faster full-project search

colbyr:
Related to #3852 

I was poking through the project search code to get familiar with how it works today, and I noticed that the redux action loads each source in sequence before we search it. For kicks, I tried loading the sources in parallel instead.

On a cold-cache, that made a _huge_ difference in the time to search a Replay-Replay. I think there would still be benefits to moving search to the backend, but maybe this could be a helpful very-short-term change?

I made a replay, but unfortunately both versions take about the same amount of time when the replay browser is recording a replay (I guess there's probably some overhead to recording?). You can better observe the difference regular screen captures.

### This branch
https://user-images.githubusercontent.com/478109/140329460-8f4a7b83-b7d5-4c49-aa88-f89ca346ea17.mov

### master
https://user-images.githubusercontent.com/478109/140329928-fed9995f-8ca2-4f9e-b066-6771fa2ef2e9.mov

### DIFF
--- a/src/devtools/client/debugger/src/actions/project-text-search.js
+++ b/src/devtools/client/debugger/src/actions/project-text-search.js
@@ -80,13 +80,18 @@ export function searchSources(cx, query) {
     const validSources = getSourceList(getState()).filter(
       source => !ThreadFront.isMinifiedSource(source.id) && !isThirdParty(source)
     );
-    for (const source of validSources) {
-      if (cancelled) {
-        return;
-      }
-      await dispatch(loadSourceText({ cx, source }));
-      await dispatch(searchSource(cx, source.id, query));
-    }
+    await Promise.all(
+      validSources.map(async source => {
+        if (cancelled) {
+          return;
+        }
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/actions/project-text-search.js
```diff
@@ -80,13 +80,18 @@ export function searchSources(cx, query) {
...
+        if (cancelled) {
+          return;
+        }
+        await dispatch(loadSourceText({ cx, source }));
+        if (cancelled) {
+          return;
+        }
+        await dispatch(searchSource(cx, source.id, query));
+      })
+    );
```
colbyr, line 94: I imagine this could be quite a large number of files on any arbitrary project. In practice, maybe it makes sense to request in batches of 100 or 1000 rather than EVERYTHING at once.

jasonLaster: yeah - chunking is definitely advisable here

PR #4299, opened 2021-11-04
Fix highlighter and node picker when screenshots are scaled

hbenl:
https://app.replay.io/recording/ac28b258-800b-42f0-be73-b0bbb04c6011 is a recording with scaled screenshots because it was recorded on a retina display (I think). See how highlighting elements and using the node picker is way off.
But in https://devtools-oa8akoob9-recordreplay.vercel.app/recording/ac28b258-800b-42f0-be73-b0bbb04c6011 it works.

### DIFF
--- a/src/devtools/server/actors/highlighters/box-model.js
+++ b/src/devtools/server/actors/highlighters/box-model.js
@@ -24,7 +24,7 @@ const nodeConstants = require("devtools/shared/dom-node-constants");
 const { LocalizationHelper } = require("devtools/shared/l10n");
 const STRINGS_URI = "devtools/shared/locales/highlighters.properties";
 const L10N = new LocalizationHelper(STRINGS_URI);
-const { refreshGraphics } = require("protocol/graphics");
+const { refreshGraphics, getDevicePixelRatio } = require("protocol/graphics");
 
 // Note that the order of items in this array is important because it is used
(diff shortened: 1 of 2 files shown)


src/devtools/server/actors/highlighters/box-model.js
```diff
@@ -522,84 +522,85 @@ class BoxModelHighlighter extends AutoRefreshHighlighter {
 
   _getBoxPathCoordinates(boxQuad, nextBoxQuad) {
     const { p1, p2, p3, p4 } = boxQuad;
+    const scale = getDevicePixelRatio();
```
ryanjduffy, line 525: Might be worth adding a `scaleQuad` method that does the calculations and returns a new quad. Would reduce the diff here and be easier to follow I think.

src/devtools/server/actors/highlighters/box-model.js
```diff
@@ -705,6 +706,7 @@ class BoxModelHighlighter extends AutoRefreshHighlighter {
    */
   _updateGuide(side, point = -1) {
     const guide = this.getElement("guide-" + side);
+    const scale = getDevicePixelRatio();
```
ryanjduffy, line 709: Similarly, a `scalePoint` could be useful but inlining the calc after this line would be okay too

ryanjduffy (approved):

PR #4305, opened 2021-11-04
Expose protocol client for use while testing things.

loganfsmyth:
I always want this when I'm prototyping new API stuff so figured I might as well add it.


jaril (approved):

jasonLaster (approved):
I think `app.client` should also work, but seems reasonable

PR #4312, opened 2021-11-04
Add recordingId to redux so that console messages can use it

jasonLaster:


jasonLaster:
@ryanjduffy is this still needed?

ryanjduffy:
I don't think so

PR #4314, opened 2021-11-05
Borderless UI follow up fixes

### LINKED ISSUE #4309: Breathe UI tweaks
1. I think these widths should be more consistent. (Jaril and Jon talked about this and I have some new ideas about it) Let's just go ahead and add the 8px thing like we see in devtools.
image

2. The share button is too large
image

3. The editor is showing a blank window
image

4. One of the gaps is 10px but should be 8px
image

5. We added a single pixel stroke at the top. The design calls to remove it. (And if we do keep it, we need to work on the spacing)
image

6. Small issue: scopes should have a rounded bottom when it's the last item.
image

jaril:
Fix #4309.

1) 8px vertical border between the video and the sidepanels
image
2) Made the share button height ~32px. This isn't quite the same as 34px as the view toggle, but because of the gray border the view toggle looks a tad smaller than it actually does and this ends up looking 👌. Thoughts welcome.
image
3) Can't reproduce this one.
4) Made the padding around the toolbar 8px.
image
5) Made an executive decision on the border underneath the header and made it thinner so it's more consistent with the other gaps. Otherwise it ended up looking too thick. 
image

@jonbell-lot23 I'll leave the decision re: #2 and #5 to you. Let me know what you think!

### DIFF
--- a/src/devtools/client/debugger/src/components/App.js
+++ b/src/devtools/client/debugger/src/components/App.js
@@ -162,7 +162,7 @@ class Debugger extends Component {
     const horizontal = this.isHorizontal();
 
     return (
-      <div className="editor-pane pt-2 overflow-hidden rounded-lg">
+      <div className="editor-pane overflow-hidden rounded-lg">
         <div className="editor-container relative">
           {!isDemo() && <EditorTabs horizontal={horizontal} />}
           <Redacted>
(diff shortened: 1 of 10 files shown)


jonbell-lot23:
Love 'em, thanks!

jasonLaster:
I spy a border 😉 

Screen Shot 2021-11-04 at 10 56 47 PM

jasonLaster (approved):
Looking great! Love these tweaks

PR #4315, opened 2021-11-05
Make the handles bigger

jcmorrow:
Our splitters are almost all 1px, and it makes it *really* hard to drag things! Especially now that the UI has these nice big gaps between the panes, it *really* feels like you should be able to grab easily, but it ends up being difficult and frustrating.

Other changes
- We had two copies of SplitBox.css that, AFAICT, are the same. I deleted what seemed like the oldest one.
- There were a *lot* of styles on the splitter, but AFAICT, none of those are needed? It's just a transparent div that sits between the other two divs. I'm trying to keep this as simple as possible!

https://user-images.githubusercontent.com/5903784/140438630-75343eca-2988-4a4e-968c-c9a400403f2b.mp4


jasonLaster (approved):

src/devtools/client/shared/components/splitter/SplitBox.js
```diff
@@ -279,9 +279,10 @@ class SplitBox extends Component {
       };
     }
 
+    console.log(splitterSize, this.props);
```
jasonLaster, line 282: 🧂

jcmorrow: thaaaaankyou!

PR #4339, opened 2021-11-05
Build & export for e2e tests

ryanjduffy:


ryanjduffy:
build + export doesn't work without more hackery because the routes aren't handled by the root any longer.

PR #4346, opened 2021-11-06
Fix end-to-end tests post Next.js migration

jcmorrow:
### Changes

- In order to run either the mock tests or the end-to-end tests, we need a dev server running in CI. Previously, we would spawn that manually from inside of a script. Now, we start that from inside of the action that does the testing. This means that if you want to run the tests locally, you'll also need your local dev server running (which was already true).
- Add the `test/examples` and `test/scripts` folder to public. This means that we will be able to run against preview branches as well.
- Add a GitHub action for setting up Next.js with a cache, so we only have to copy one line everywhere, not an entire step. It would be nice if we could further refactor out the test stripes, because a lot of things are repeated 9 times.
- Remove the `dev-server.js` file. If we're going to use next.js, then we should try and keep our config as out-of-the-box as possible, in order to allow us to stay up-to-date. The fact that the `test` folder has to be available on the development server for the tests to run at all seems like the real problem here. A good follow-on task would be to sort out the best way to handle those files in the test/dev environment.
- Go back to node 14 for lock file. I didn't realize when I changed this that we were stuck on such an outdated node version. We should fix this ASAP, but until then, let's stick with what was working, just to minimize the factors that have gone into resolving all of the failing tests.
- Log more from playwright tests. This could be annoying, maybe we should tune it, but it's nice to know what's going on in these tests, and we're only going to look at the output when something is failing anyways, so it may as well be verbose.
- Disable the worst of the failing tests.

### But the tests are still failing...

That's true! We still have flaky, intermittent tests. The goal of this PR was to get things to *run*, getting them to pass is a future us problem! I have been unable to get the mock tests to work to any extent on CI, although they do pass intermittently locally.

There are also tests (it doesn't seem to be consistent which ones) where the browser starts up and just never seems to navigate to the page properly. I haven't been able to figure out where it's getting stuck, but when tests time out now, it's usually for this reason.

### Notes on GitHub Actions

I'm going to put these into a notion doc as well, but documenting here couldn't hurt. Three things that tripped me up on this PR:

- *any* merge conflicts mean actions *will not run at all*. If you have a branch and it's unclear why actions aren't kicking off, make sure you check for and resolve merge conflicts.
- When using actions that are type `javascript` your `inputs` get automatically turned into environment variables, which is convenient. However **when using `composite` actions this won't happen**. You have to manually pass the environment variables in with an [expression](https://docs.github.com/en/actions/learn-github-actions/expressions).
- The `checkout` action that we use all the time performs a strict `clean` of the directory that it's about to clone into! So, if you are trying to do some sort of caching or otherwise unloading assets into a dir that you also need to do `git checkout XXXX` in (with the checkout action), you either need to checkout *first*, or you can use `[clean: false](https://github.com/actions/checkout#usage)` I think, though I haven't actually tried this.

Fixes #4313


.github/actions/mockTest/action.yml
```diff
@@ -1,6 +1,32 @@
...
+    - name: "Setup NextJS Cache"
+      uses: actions/cache@v2
+      with:
+        path: ${{ github.workspace }}/.next/cache
+        # Generate a new cache whenever packages or source files change.
+        key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**.[jt]s', '**.[jt]sx') }}
+        # If source files changed but packages didn't, rebuild from a prior cache.
+        restore-keys: |
+          ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-
+    - run: ./node_modules/.bin/next dev -p 8080 &
```
ryanjduffy, line 26: Nice using `&` here!

jcmorrow: Thanks! It felt a *little* hacky, but better than trying to do the `start a shell process from inside of the test runner` thing 🤷

jasonLaster: oh gosh - i thought we might need to background a process 🙍‍♂️

test/mock/scripts/session-error.ts
```diff
@@ -41,5 +41,7 @@ const bindings = basicBindings();
 runTest("sessionError", async (page: Page) => {
   await page.goto(devtoolsURL({ id: recordingId }));
   await installMockEnvironment(page, { graphqlMocks, messageHandlers, bindings });
-  await page.textContent("text=Unexpected session error");
+  await page.textContent(
+    "text=Something went wrong while replaying, we'll look into it as soon as possible"
+  );
```
jasonLaster, line 46: should this be it's own tiny PR fix to get this suite green?

src/ui/components/App.tsx
```diff
@@ -89,7 +89,7 @@ function App({ theme, modal, children }: AppProps) {
   }, [theme]);
 
   useEffect(() => {
-    if (auth.isAuthenticated) {
+    if (auth.isAuthenticated && !isTest()) {
```
jasonLaster, line 92: why is this needed?

jasonLaster (commented):
there's a lot of great stuff in here

src/ui/components/App.tsx
jcmorrow: Yeah, gonna be honest, I don't have a 100% understanding of this one. Something was interacting with updating or getting preferences, and that was failing in tests. I'm not sure if there's something special about how we auth in test such that auth.isAuthenticated returns true but user is not completely present or something? IDK!

test/mock/scripts/session-error.ts
jcmorrow: These tests are still super flaky sadly. They pass maybe 25% of the time locally.

jasonLaster: :/ that's sad. All the more reason to ship lots of small fixes and chip away at it.

src/ui/components/App.tsx
jasonLaster: haha - fair enough, i could see why updating user settings as a test would be a no bueno...

test/harness.js
```diff
@@ -1,58 +0,0 @@
-/* Copyright 2020 Record Replay Inc. */
-
-// Harness used for local testing.
```
jasonLaster, line 3: Weird, i thought this file was used :)

jcmorrow: `/src/test/harness.js` *is* used, a lot. The thing that gave away this file not being used was the header include in it 😄 . I've never seen that in a JS file before!

jcmorrow: Wait... I'm confused, I think you're correct, the tests don't run without this file,  and also I do *not* understand how 

```
# include "header.js"
```

is valid JS? What am I missing here?

ryanjduffy: Iirc, it's compiled to run tests in gecko

.github/actions/mockTest/action.yml
```diff
@@ -1,6 +1,14 @@
...
 runs:
-  using: "node12"
-  main: "./main.js"
+  using: "composite"
+  steps:
+    - run: ./node_modules/.bin/next dev -p 8080 &
+      shell: bash
+      env:
+        NEXT_PUBLIC_API_URL: https://api.replay.io/v1/graphql
+        NEXT_PUBLIC_DISPATCH_URL: wss://dispatch.replay.io
```
jcmorrow, line 10: Let's move these to a `.env.sample` and then just `cp` that in this job. This should look as close as possible to a new dev setup following the README.

src/protocol/utils.ts
```diff
@@ -68,20 +66,14 @@ export function binarySearch(start: number, end: number, callback: (mid: number)
 }
 
 function NotAllowed() {
+  console.log(new Error().stack);
```
jcmorrow, line 69: remove.

public/test/examples/example-recordings.json
```diff
@@ -0,0 +1,15 @@
...
+  "doc_navigate.html": "3ff31137-8782-4da3-a377-50adcfd5f2b4",
+  "doc_prod_bundle.html": "4260d4e2-a987-4c53-8255-203423d174f5",
+  "doc_rr_basic.html": "78354612-139c-4e27-9ec1-5168c95c1ac9",
+  "doc_rr_error.html": "92ec8227-af85-4713-9409-23dab44c0532",
+  "doc_rr_objects.html": "0e00cdbd-2d56-4cfc-96fe-4daf4d7f295b",
+  "doc_rr_preview.html": "6cde0088-44ce-4ca5-8f4a-4cccb28203be",
+  "doc_rr_region_loading.html": "50ccfe8d-7eba-4847-ab1d-311d4daf2b5d",
+  "doc_rr_worker.html": "42c6c475-dcfe-4807-96d8-393e4d802624",
+  "doc_stacking.html": "53860427-adef-46ed-b59d-c4fb4a309a6c",
+}
```
jasonLaster, line 15: This should probably not ship

test/run.js
```diff
@@ -16,6 +16,9 @@ const {
 } = require("./utils");
 const { listAllRecordings } = require("@recordreplay/recordings-cli");
 
+const EXAMPLE_SERVER = "http://localhost:8080/test/examples";
+const REPLAY_SERVER = "http://localhost:8080";
```
jasonLaster, line 20: Oh snap! Nice job extracting this.

.github/actions/setup-next/action.yml
```diff
@@ -0,0 +1,14 @@
+name: "Setup Next.js"
+description: "Installs node dependencies, with cache, and gets Next up and running, also with a cache."
```
ryanjduffy, line 2: Nit: this only does the nextjs setup + cache since the node stuff is in the other action

.github/workflows/test.yml
```diff
@@ -61,114 +54,146 @@ jobs:
...
-    needs: [build, download-browser, download-node, download-driver]
-    timeout-minutes: 15
+    needs: [build-server, download-browser, download-node, download-driver]
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v2
       - uses: actions/download-artifact@v2
+      - run: mv node_modules/node_modules.tgz ./
+      - run: tar -xzf node_modules.tgz
+      - uses: ./.github/actions/setup-next
```
ryanjduffy, line 66: I don't propose this change for here but i'm curious if we should bundle all of these into our test custom action since it's already a composite action now and it's not really reusable outside this repo anyway.

.gitignore
```diff
@@ -7,7 +7,8 @@ cypress/videos/
 test/example-recordings.json
 .idea/
 cypress.env.json
-.env*
```
ryanjduffy, line 10: I think we still want to exclude all `.env*` files with `.env.sample` being the exception so we (read: me because i `git add .` freely) don't accidentally add `.env.local` to git.

README.md
```diff
@@ -16,7 +16,8 @@ Replay's DevTools is a React app built on top of the Replay [protocol](https://w
 git clone git@github.com:RecordReplay/devtools.git
 cd devtools
 npm install
-npm start
+cp .env.sample .env
+npm run dev
```
ryanjduffy, line 20: Thanks for catching this!

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
```diff
@@ -125,7 +125,7 @@ function BreakpointNavigationStatus({ executionPoint, analysisPoints, indexed })
   let maxStatusLength = 0;
   if (!indexed) {
     status = "Indexing";
-  } else if (!analysisPoints | !executionPoint) {
+  } else if (!analysisPoints || !executionPoint) {
```
ryanjduffy, line 128: Oops :)

src/protocol/thread/bounds.ts
```diff
@@ -26,7 +26,10 @@ export class NodeBoundsFront {
   }
 }
 
-Object.setPrototypeOf(NodeBoundsFront.prototype, new Proxy({}, DisallowEverythingProxyHandler));
+Object.setPrototypeOf(
+  NodeBoundsFront.prototype,
+  new Proxy(Object.create(null), DisallowEverythingProxyHandler)
```
ryanjduffy, line 31: cc: @loganfsmyth - this fixes the issue I mentioned in a call regarding hot reloading checking exports to see if they were React components which caused this to log `Not Alllowed` errors in the console. I think this is a safe fix but I don't have much background on why this Proxy exists to be sure.

src/protocol/utils.ts
```diff
@@ -72,16 +70,9 @@ function NotAllowed() {
...
-  //   return null;
-  // },
-  //has() {
-  //  NotAllowed();
-  //  return false;
-  //},
   get() {
-    NotAllowed();
+    return null;
+    // NotAllowed();
```
ryanjduffy, line 75: I'm not sure about this one but I don't know the consequence of it.

cc: @loganfsmyth  (I think you were already mentioned on this so apologies if this is duplicate).

src/ui/setup/dynamic/devtools.ts
```diff
@@ -62,8 +62,7 @@ declare global {
 }
 
 export default async (store: Store) => {
-  const url = new URL(window.location.href);
-  const dispatchUrl = url.searchParams.get("dispatch") || process.env.NEXT_PUBLIC_DISPATCH_URL;
+  const dispatchUrl = process.env.NEXT_PUBLIC_DISPATCH_URL;
```
ryanjduffy, line 65: I think we need to keep the QS parameter support around for now unless @jazzdan has already resolved this. Either way, this PR probably isn't the right place to make the change unless Nextjs broke it.

ryanjduffy (approved):
This is a huge fix and I really appreciate your help on it. I'm not sure about copying the test files into public vs leaving them where they are until they can be moved out of devtools but I wouldn't block on it. Few notes/comments for consideration. Well done, @jcmorrow

.github/actions/setup-next/action.yml
jcmorrow: Yep, thanks, originally this did both and I forgot to change the description.

.github/workflows/test.yml
jcmorrow: Yeah, I actually did that, and then I realized that you can't call another locally defined action from a locally defined action. More specifically you can't use uses inside of an action (although it seems like you can as long as it's not a local uses? Idk.) More details here: https://github.com/actions/runner/issues/646. But I do think the untarring can be consolidated, so I'll move that.

src/ui/hooks/recordings.ts
```diff
@@ -170,7 +170,7 @@ export function useIsTeamDeveloper() {
     console.error("Apollo error while getting the user's role", error);
   }
 
-  return { isTeamDeveloper: data.recording.userRole !== "team-user", loading };
+  return { isTeamDeveloper: data?.recording.userRole !== "team-user", loading };
```
jasonLaster, line 173: gosh - you're deep in the guts of devtools here

test/run.js
```diff
@@ -475,6 +492,7 @@ async function createExampleBrowserRecording(url, target) {
   }
 
   if (failures.length) {
+    console.log({ timings });
```
jasonLaster, line 495: Nice

jasonLaster (commented):
Lots of good stuff!

jasonLaster:
Even if this is one PR, it should be several commits for version control reasons

A simple rule of thumb would be
1. Test example files
2. src fixes
3. helper changes
4. harness changes

src/protocol/utils.ts
loganfsmyth: I'd `return undefined;` instead of `null`, but otherwise I think this change is enough that changing the the `Object.create(null)` calls back to `{}` will be fine.

src/protocol/utils.ts
```diff
@@ -72,16 +70,9 @@ function NotAllowed() {
...
 
 export const DisallowEverythingProxyHandler: ProxyHandler<object> = {
-  // getPrototypeOf() {
-  //   NotAllowed();
-  //   return null;
-  // },
-  //has() {
-  //  NotAllowed();
-  //  return false;
-  //},
```
loganfsmyth, line 82: As much as I hate to leave random commented-out code, I'd probably keep this as
```diff
-  // getPrototypeOf() {
-  //   NotAllowed();
-  //   return null;
-  // },
-  //has() {
-   //  NotAllowed();
-  //  return false;
-   //},
+  // getPrototypeOf() { },
+  // has() { },
```
Proxy handlers are a specific list of handlers and it's really nice to at a glance see which ones _aren't_ implemented.

test/manifest.js
```diff
@@ -240,11 +240,12 @@ module.exports = [
...
-    example: "doc_inspector_styles.html",
-    script: "inspector-03.js",
-    targets: ["gecko"],
-  },
+  // {
+  //   This  test just times out, it doesn't even seem to run
+  //   example: "doc_inspector_styles.html",
+  //   script: "inspector-03.js",
+  //   targets: ["gecko"],
+  // },
```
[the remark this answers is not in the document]
bhackett1024: agreed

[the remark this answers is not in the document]
jcmorrow: https://github.com/RecordReplay/devtools/issues/4438 is tracking 🙂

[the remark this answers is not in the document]
bhackett1024: Thanks!

PR #4348, opened 2021-11-06
Make editor background white

jcmorrow:
It was really disappearing in the negative space.

Before

CleanShot 2021-11-05 at 20 53 58@2x

After

CleanShot 2021-11-05 at 21 01 05


jasonLaster:
Looks great!

PR #4367, opened 2021-11-08
Nags should use a helper

jasonLaster:
switches to a helper to make the access safe and the logic clearer

### DIFF
--- a/src/ui/utils/user.ts
+++ b/src/ui/utils/user.ts
@@ -1,7 +1,13 @@
+import { Nag } from "ui/hooks/users";
+
 export function getAvatarColor(avatarID?: number) {
   if (!avatarID) {
     return "#696969";
   }
   const avatarColors = ["#2D4551", "#509A8F", "#E4C478", "#E9A56C", "#D97559"];
   return avatarColors[avatarID % avatarColors.length];
+}
+
+export function shouldShowNag(nags: Nag[], key: Nag) {
+  return nags && !nags.includes(key);
 }
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -19,6 +19,7 @@ import FirstEditNag from "./FirstEditNag";
 import hooks from "ui/hooks";
 import { Nag } from "ui/hooks/users";
 import { prefs } from "ui/utils/prefs";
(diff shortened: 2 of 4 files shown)


src/ui/hooks/users.ts
```diff
@@ -72,7 +72,7 @@ export function useGetUserInfo() {
   const picture: string = data?.viewer?.user.picture;
   const email: string = data?.viewer?.email;
   const internal: boolean = data?.viewer?.internal;
-  const nags: Nag[] = data?.viewer?.nags;
+  const nags: Nag[] = data?.viewer?.nags || [];
```
jaril, line 75: This may be problematic. The nags is a list of things that tell us what not to display. An empty nags array would default to showing all the nags.

If that's the intention, then this works great. Just wanted to make sure that's the case!

jasonLaster: gosh - good catch.

The problem I'm trying to solve is this line
`!nags.includes(Nag.FIRST_REPLAY_2)` where we don't have an array. The better solution would be to wrap this stuff in a helper.

src/ui/utils/user.ts
```diff
@@ -1,7 +1,13 @@
...
+
 export function getAvatarColor(avatarID?: number) {
   if (!avatarID) {
     return "#696969";
   }
   const avatarColors = ["#2D4551", "#509A8F", "#E4C478", "#E9A56C", "#D97559"];
   return avatarColors[avatarID % avatarColors.length];
 }
+
+export function shouldShowNag(nags: Nag[], key: Nag) {
```
jcmorrow, line 11: What if we explicitly called the `nags` param `nagsToHide` or `dontShow`, or `hiddenNags`, just something to make it clear what the significance of `nags` is?

jcmorrow (approved):
I think this makes sense to me now? It feels like it might be clearer overall to have an object that is like `DismissedNags` and actually contains the `Nag` enum values as keys, with `booleans` on the other side, so we would also have protection from things like typos, and it would also deal with the default problem, but this is also ok for now 👍

src/ui/utils/user.ts
jasonLaster: `shouldShowNag(hiddenNags:` hurts my brain even more lol... 

Never sure how about OO in these contexts, but yeah an object is what i thought of first too

jaril (approved):
Yeah switching to the helper's much clearer!

PR #4381, opened 2021-11-08
Don't try to dismiss nags if we're in a test

jaril:


src/ui/hooks/users.ts
```diff
@@ -118,10 +118,15 @@ export function useDismissNag() {
     console.error("Apollo error while updating the user's nags:", error);
   }
 
-  return (nag: Nag) =>
+  return (nag: Nag) => {
+    if (isTest()) {
+      return;
+    }
+
```
jasonLaster, line 125: I'd tighten it a bit further. We should not be calling this mutation if we already have the nag

```diff
 export async function dismissNag(nag: Nag) {
+  const { nags } = await useGetUserInfo();
+  if (!nags || nags.includes(nag)) {
+    return;
+  }
+
   await mutate({
     mutation: DISMISS_NAG,
     variables: { nag },
```

PR #4388, opened 2021-11-08
Going back to inaccessible bright blue colour

jonbell-lot23:
Putting the blue back to the super bright blue from before. For now. Even though it's not accessible.


jasonLaster:
It's so darn whimsical

PR #4392, opened 2021-11-09
Keep pause info panels from expanding to full height when expanded

### LINKED ISSUE #3200: Pause panel should use a stacked sidebar accordion
I just landed Ryan's PR which fixes the layout issues for the pause panel, but unintentionally switched the pause panel to the sources accordion ux where the panels are evenly spaced.

Lets bring back the stacked accordions which avoids unnecessary empty space for the top panels

jaril:
Fix #3200.

image

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/SecondaryPanes.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/SecondaryPanes.css
@@ -39,10 +39,6 @@
   flex-direction: column;
 }
 
-.secondary-panes .accordion li {
-  display: contents;
-}
-
 .secondary-panes-wrapper .accordion li:last-child ._content {
   border-bottom: 0;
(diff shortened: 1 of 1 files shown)


jcmorrow (changes requested):
Sadly I don't think this works. Now you can't scroll when the panes go beyond their bounds

https://user-images.githubusercontent.com/5903784/140843196-d6e0eccc-1f56-41e7-bdd3-72accd9324c4.mp4

You can't tell from the video but I'm trying to scroll 😆

jaril:
> Sadly I don't think this works. Now you can't scroll when the panes go beyond their bounds
> 
>  CleanShot.2021-11-08.at.17.09.44.mp4 
> You can't tell from the video but I'm trying to scroll 😆

Yeah noticed that. Gonna play around with it a little bit later!

PR #4393, opened 2021-11-09
Force fetching recording from server

ryanjduffy:
There are a few errors that seem to stem from failing to clean up some module-scoped data (thread event listeners, screenshot cache). This likely resolves these but does make the UI a bit clunky (e.g. flash of white screen while fetching the HTML).

https://user-images.githubusercontent.com/788456/140842897-056bf3d4-c37c-4bac-8721-9d9f5146d018.mp4

### DIFF
--- a/src/ui/components/Library/RecordingRow.tsx
+++ b/src/ui/components/Library/RecordingRow.tsx
@@ -75,7 +75,7 @@ function RecordingRow({
     }
 
     loadReplayPrefs(recording.id);
-    router.push(url);
+    window.location.href = url;
   };
   const toggleChecked = () => {
     if (selected) {


src/ui/components/Library/RecordingRow.tsx
```diff
@@ -75,7 +75,7 @@ function RecordingRow({
     }
 
     loadReplayPrefs(recording.id);
-    router.push(url);
+    window.location.href = url;
```
jasonLaster, line 78: I wonder if it is any better to use a Next Link element https://nextjs.org/docs/api-reference/next/link

jasonLaster (approved):
I think it is better to start here and make it faster with SSR and other next features that Link offers.

Trust you and @hbenl to weigh the trade-offs

src/ui/components/Library/RecordingRow.tsx
ryanjduffy: I don't think it'll produce the same result. I believe it does client side route change by default. We can investigate more in the follow up.

PR #4396, opened 2021-11-09
Refactor the code for painting screenshots (#4373) (#4395)

### LINKED ISSUE #4373: Video should not block on repainting
**Description:**

When repainting is enabled, the viewer updates significantly slower because it can take time to repaint and then send the screen. We should always use the default screenshot first and then update it with the repainted one when it is ready. 

Note, This is not ideal, but I think it is better than adding a significant delay.

### LINKED ISSUE #4395: Screenshot not updated while dragging timeline
If the user has repainting enabled and drags the time marker over the timeline, the screenshot is not updated.
Repainting enabled: https://app.replay.io/recording/88782d05-6d88-4826-bbcf-714ce15e0516
Repainting disabled: https://app.replay.io/recording/780130b4-99b9-4db8-a867-2db4bcf375ca

hbenl:
Painting screenshots after seeking to a new pause was handled in 2 places: `ui/actions/timeline.ts` for recorded screenshots and `protocol/graphics.ts` for repainted screenshots. This PR consolidates the logic in `protocol/graphics.ts` and
- fixes #4373
- fixes #4395 
- removes some properties from redux that were not used anymore

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -2,7 +2,7 @@
 
 import { ThreadFront } from "./thread";
 import { assert, binarySearch } from "./utils";
-import { ScreenshotCache } from "./screenshot-cache";
+import { DownloadCancelledError, ScreenshotCache } from "./screenshot-cache";
 import ResizeObserverPolyfill from "resize-observer-polyfill";
 import {
   TimeStampedPoint,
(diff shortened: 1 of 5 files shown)


src/protocol/graphics.ts
```diff
@@ -228,22 +228,31 @@ export function setupGraphics(store: UIStore) {
     });
   });
 
-  ThreadFront.on("paused", async () => {
-    if (!isRepaintEnabled()) {
-      return;
-    }
-
+  ThreadFront.on("paused", async ({ time }) => {
     await ThreadFront.ensureAllSources();
```
jcmorrow, line 232: Should this block the graphics stuff as well, or would it be worth doing like a `Promises.all` sort of thing to kick these off at the same time?

src/protocol/graphics.ts
```diff
@@ -255,11 +264,8 @@ export function setupGraphics(store: UIStore) {
           }
         }
         paintGraphics(screenShot, mouse);
-      } else {
-        const { screen, mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
-        paintGraphics(screen, mouse);
       }
-    }
+    } catch {}
```
jcmorrow, line 268: What error are we expecting here? If it's just a general purpose catch-all, can we still call Sentry even if we stop the error from propogating?

src/ui/state/timeline.ts
```diff
@@ -25,15 +20,12 @@ export interface TimelineState {
...
   playbackPrecachedTime: number;
-  screenShot: ScreenShot | null;
-  mouse: MouseAndClickPosition | null | undefined;
   recordingDuration: number | null;
   zoomRegion: ZoomRegion;
   timelineDimensions: { width: number; left: number; top: number };
   hoverTime: number | null;
   unprocessedRegions: TimeRange[];
   shouldAnimate: boolean;
-  tooltip: Tooltip | null;
```
jcmorrow, line 36: Nice, awesome to see this! I feel like CSS and redux are the two places where it is easiest to just leave unused code forever.

jcmorrow (approved):
Couple of questions, and I won't pretend to understand how all of this works, but it seems reasonable to me! 🖌️

src/protocol/graphics.ts
hbenl: Good point, I didn't think much about this part (because I didn't change it and because it's irrelevant once the recording is fully loaded) but it actually blocks the graphics stuff unnecessarily when the sources haven't been loaded yet (and that can take some time). I'll improve this.

src/protocol/graphics.ts
hbenl: We're expecting a [DownloadCancelledError](https://github.com/RecordReplay/devtools/blob/73b26477108a666885b40150c3ae7b62fb80e5b8/src/protocol/screenshot-cache.ts#L6) because screenshot downloads are cancellable (they are cancelled when too many screenshot downloads are requested at the same time, although this was more relevant in the past when the timeline tooltip showed a thumbnail of the screenshot from the hovered time - that could queue a lot of screenshot downloads, but it would only ever use the last one, so it made sense to cancel the others).
I'll check the error and rethrow if it's not a `DownloadCancelledError`.

PR #4399, opened 2021-11-09
Make the handles big in viewer as well

jcmorrow:
I made this change in DevTools last week but apparently I missed one.

### Before (Zoomed-in and with a red background for visibility)

CleanShot 2021-11-09 at 08 57 27

### After (Zoomed-in and with a red background for visibility)

CleanShot 2021-11-09 at 08 56 18


jaril (approved):
LGTM!

PR #4411, opened 2021-11-09
Reduce the number of warnings/errors

jcmorrow:
These are all small fixes that came from #4346 that either eliminate
possible console errors or eliminate linting/tsc errors.

### DIFF
--- a/src/protocol/utils.ts
+++ b/src/protocol/utils.ts
@@ -81,7 +81,7 @@ export const DisallowEverythingProxyHandler: ProxyHandler<object> = {
   //  return false;
   //},
   get() {
-    NotAllowed();
+    return undefined;
   },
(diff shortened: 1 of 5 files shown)


src/protocol/utils.ts
```diff
@@ -81,7 +81,7 @@ export const DisallowEverythingProxyHandler: ProxyHandler<object> = {
   //  return false;
   //},
   get() {
-    NotAllowed();
```
colbyr, line 84: Aha I'm often mystified by that "NotAllowed" error.

colbyr (approved):

src/protocol/utils.ts
jcmorrow: It's just so helpful. I'll miss it.

PR #4416, opened 2021-11-10
Remove TeamLeaderOnboardingModal

jaril:
This PR removes the TeamLeaderOnboarding flow, which essentially helped the user create a team, invite members, then download the replay browser.

This case almost never happens anymore since we did away with its triggers after we opened public access to Replay. This flow was specifically targeting folks who were either invited directly via the Replay invites in the settings pane, or were sent an email invite link as we opened up cohorts. Both of those are no longer used so we can get rid of this.


src/ui/utils/onboarding.ts
```diff
@@ -0,0 +1,23 @@
...
+export const isTeamReferral = () => isTeamMemberInvite() || hasTeamInvitationCode();
+
+// This is for the user onboarding flow where the user signs up for Replay using
+// a Replay team invite that they received in their email.
+export const singleInvitation = (invitations: number, workspaces: number) =>
+  invitations === 1 && workspaces === 0;
+
+// This is for the user onboarding flow where the user opens the Replay
+// browser for the first time. It teaches them how to create their first replay.
+export function firstReplay(nags: Nag[]) {
```
jcmorrow, line 20: These things don't desperately *need* unit tests, but what do you think about adding some, just to get into the habit of doing it? Plus, tests can serve as a kind of terse documentation, much like the comments here are doing.

jcmorrow (approved):
Left one comment about tests, but it's just a suggestion, looks good! Love the color! 🔴

src/ui/utils/onboarding.ts
jaril: Let's do it!

PR #4419, opened 2021-11-10
Fix double equals

jazzdan:


jcmorrow (approved):
Was this a semgrep catch? 🤖

jasonLaster:
okay - that's odd... but shouldnt cause any logical issus

jazzdan:
> Was this a semgrep catch? 🤖

Yep!

PR #4444, opened 2021-11-11
Don't init the store and websocket on hot reload

jcmorrow:
This is.... not the most defensive code I have ever written. Should we use a more obscure key, or is that some other reliable way of knowing that this has already been called?

### DIFF
--- a/src/ui/setup/dynamic/devtools.ts
+++ b/src/ui/setup/dynamic/devtools.ts
@@ -14,7 +14,7 @@ import comments from "ui/reducers/comments";
 import contextMenus from "ui/reducers/contextMenus";
 import reactDevTools from "ui/reducers/reactDevTools";
 import { selectors } from "ui/reducers";
-import { actions } from "ui/actions";
+import { actions, UIStore } from "ui/actions";
 const { setupApp, setupTimeline, setupReactDevTools, createSession } = actions;
 
 import * as dbgClient from "devtools/client/debugger/src/client";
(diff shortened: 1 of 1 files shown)


src/ui/setup/dynamic/devtools.ts
```diff
@@ -106,39 +107,42 @@ export default async (store: Store) => {
...
-  setupTimeline(store);
-  setupEventListeners(store);
-  setupGraphics(store);
-  initOutputSyntaxHighlighting();
-  setupMessages(store);
-  setupLogpoints(store);
-  setupExceptions(store);
-  setupReactDevTools(store);
+  if (!window.loaded) {
+    window.loaded = true;
```
jasonLaster, line 111: I think I'd do two things here:

1. extract to a function
2. more specific variable

```js
function bootstrapApp() {
  if (window.hasAlreadyBootstrapped) {
    return
  } 
    extendStore(store, initialState, reducers, thunkArgs);

    dbgClient.bootstrap(store);

    // Initialize the socket so we can communicate with the server
    initSocket(store, dispatchUrl);
}
```

jasonLaster: also we might need to include this as well

```js
  const settings = await getUserSettings();
  updateEnableRepaint(settings.enableRepaint);
  setupDemo();
```

jasonLaster (approved):

src/ui/setup/dynamic/devtools.ts
jcmorrow: Yeah, actually, this whole function is supposed to be idempotent.

jcmorrow: I think

jasonLaster (approved):
love it

PR #4445, opened 2021-11-11
Support rendering proxy objects.

loganfsmyth:
Refs https://github.com/RecordReplay/backend/issues/3516

### DIFF
--- a/src/protocol/thread/value.ts
+++ b/src/protocol/thread/value.ts
@@ -152,6 +152,11 @@ export class ValueFront {
     return (this.hasPreview() && this._object!.preview!.promiseState) || null;
   }
 
+  previewProxyState() {
+    // Older recordings did not set proxyState so this could be null.
+    return (this.hasPreview() && this._object!.preview!.proxyState) || null;
+  }
+
   className() {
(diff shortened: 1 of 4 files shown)


src/protocol/thread/value.ts
```diff
@@ -152,6 +152,11 @@ export class ValueFront {
     return (this.hasPreview() && this._object!.preview!.promiseState) || null;
   }
 
+  previewProxyState() {
+    // Older recordings did not set proxyState so this could be null.
+    return (this.hasPreview() && this._object!.preview!.proxyState) || null;
```
jaril, line 157: Small nit, but what do you think about doing: `return this._object?.preview?.proxyState || null`.

jaril: Meh could go either way, tbh this is more readable!

jaril (approved):
LGTM!

src/protocol/thread/value.ts
loganfsmyth: This was essentially a copy-paste from the function above this, so probably in favor of leaving it.

PR #4448, opened 2021-11-11
Simplify the outline view

jasonLaster:
* Remove footer for alphabetizing
* Remove unnecessary context menu
* Clean up styling so the form is on top and the scrolled region is below the header


jasonLaster:
I sat down to add `react-window` https://github.com/RecordReplay/devtools/issues/4442 and found that we should first clean up the outline view ...

colbyr (approved):
Looks pretty good to me 👍

PR #4450, opened 2021-11-11
Keep the red pause indicator in the events panel from shifting the layout when it changes

### LINKED ISSUE #4362: Red line shouldn't have jumping effect in Events
https://www.loom.com/share/9186855ea0a749619fe7d8ff63314f62

Right now things are jumping around in events because of the red line moving. The line shouldn't push things around. Notice how in console things don't jump.

jaril:
Fix #4362.

https://user-images.githubusercontent.com/15959269/141238834-6fc1e348-abf8-4070-8ad2-852f9b5c2a59.mov

### DIFF
--- a/src/ui/components/Events/Events.tsx
+++ b/src/ui/components/Events/Events.tsx
@@ -9,9 +9,24 @@ const { getExecutionPoint } = require("devtools/client/debugger/src/reducers/pau
 
 import Event from "./Event";
 import { trackEvent } from "ui/utils/telemetry";
+import classNames from "classnames";
 
-function CurrentTimeLine() {
-  return <div className="bg-secondaryAccent w-full m-0" style={{ height: "3px" }} />;
+function CurrentTimeLine({
+  currentEventIndex,
+  index,
+}: {
+  currentEventIndex: number;
+  index: number;
+}) {
+  return (
+    <div
+      className={classNames("w-full m-0", {
+        "bg-secondaryAccent": currentEventIndex === index,
+        "bg-transparent": currentEventIndex === index,
+      })}
+      style={{ height: "3px" }}
+    />
+  );
 }
 
 const FILTERED_EVENT_TYPES = ["keydown", "keyup"];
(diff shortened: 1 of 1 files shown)


src/ui/components/Events/Events.tsx
```diff
@@ -9,9 +9,24 @@ const { getExecutionPoint } = require("devtools/client/debugger/src/reducers/pau
...
+function CurrentTimeLine({
+  currentEventIndex,
+  index,
+}: {
+  currentEventIndex: number;
+  index: number;
+}) {
+  return (
+    <div
+      className={classNames("w-full m-0", {
```
jcmorrow, line 23: Do you need the `w-full` here? I think the div should be full-width by default, as it's a block-level element.

src/ui/components/Events/Events.tsx
```diff
@@ -9,9 +9,24 @@ const { getExecutionPoint } = require("devtools/client/debugger/src/reducers/pau
...
+        "bg-secondaryAccent": currentEventIndex === index,
```
jcmorrow, line 24: What do you think about combining these two into one line?

src/ui/components/Events/Events.tsx
```diff
@@ -30,20 +45,18 @@ function Events({ currentTime, events, executionPoint, seek }: PropsFromRedux) {
...
   return (
-
-    <div className="flex flex-col py-1.5 self-stretch space-y-1.5 w-full text-xs bg-white">
-    
+    <div className="flex flex-col py-1.5 self-stretch w-full text-xs bg-white">
       {events.map((e, i) => {
         return (
           <div key={e.point}>
-            {i === currentEventIndex ? <CurrentTimeLine /> : null}
+            <CurrentTimeLine currentEventIndex={currentEventIndex} index={i} />
```
jcmorrow, line 52: I could go either way on this, but it seems like maybe a lot to pass around these two variables just to do an equality check on them later. What do you think about having an `active` state, and this component can render `<CurrentTimeLine active={i === currentEventIndex}/>`?

src/ui/components/Events/Events.tsx
```diff
@@ -30,20 +45,18 @@ function Events({ currentTime, events, executionPoint, seek }: PropsFromRedux) {
   );
 
   return (
-
-    <div className="flex flex-col py-1.5 self-stretch space-y-1.5 w-full text-xs bg-white">
-    
+    <div className="flex flex-col py-1.5 self-stretch w-full text-xs bg-white">
```
jcmorrow, line 48: I think the only classes that are doing anything here are: `py-1.5 text-xs bg-white`?

jcmorrow (commented):
I have some suggestions, but I don't feel super strongly about them.

src/ui/components/Events/Events.tsx
jaril: 👍

src/ui/components/Events/Events.tsx
jaril: 👍

jaril:
Went ahead and made the changes 👍

https://user-images.githubusercontent.com/15959269/141353874-72a58036-b6ac-42c2-be38-9f778cdbb373.mov

jasonLaster (approved):

PR #4453, opened 2021-11-11
Don't error when workspaces is not yet populated

jcmorrow:
Also, in general I find that `sortBy` is a very worthwhile import. JS sorting is a silly thing, and very easy to spend time wrestling with or get subtly wrong.


jaril (approved):
_Way_ cleaner.

PR #4456, opened 2021-11-11
Add support for API key auth

ryanjduffy:
## Issue

Google auth uses captcha to protect against bots and automated UI tests struggle with captchas because ... ya know ... they _are_ bots.

## Resolution

* Captures the API key via the request header and passes it to the App in props using the Next.js API: `getInitialProps`.
* Adds plumbing to the auth provider component to use the API key as the token when set and skip any token refreshes
* Fixes up a few places that asserted we have a `user` via auth0 and now may need to wait a tick for the graphql query to resolve

## Notes
We could support other ways to input the api key (e.g. a query string) but it's not designed to be a user-facing feature so the extra friction seems like a feature not a bug. Open to discussion

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -2,7 +2,8 @@ import "../src/test-prep";
 import "ui/utils/whatwg-url-fix";
 
 import Head from "next/head";
-import type { AppProps } from "next/app";
+import type { AppContext, AppProps } from "next/app";
+import NextApp from "next/app";
 import React, { useEffect, useState } from "react";
 import { Provider } from "react-redux";
 import { IntercomProvider } from "react-use-intercom";
@@ -135,7 +136,7 @@ import { InstallRouteListener } from "ui/utils/routeListener";
 // _ONLY_ set this flag if you want to disable the frontend entirely
 const maintenanceMode = false;
 
-const AppRouting = ({ Component, pageProps }: AppProps) => {
+const AppRouting = ({ apiKey, Component, pageProps }: AppProps & AuthProps) => {
   const [store, setStore] = useState<Store | null>(null);
   useEffect(() => {
     bootstrapApp().then((store: Store) => setStore(store));
@@ -151,7 +152,7 @@ const AppRouting = ({ Component, pageProps }: AppProps) => {
 
   return (
     <Provider store={store}>
-      <tokenManager.Auth0Provider>
+      <tokenManager.Auth0Provider apiKey={apiKey}>
         <ApolloWrapper>
           <IntercomProvider appId={"k7f741xx"} autoBoot>
             <ConfirmProvider>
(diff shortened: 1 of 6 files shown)


ryanjduffy:
This is a bit difficult to test because the `apiKey` is only accepted via a HTTP Header. You can try it out more easily with https://github.com/RecordReplay/playwright-tests/pull/104 if you set `RECORD_REPLAY_API_KEY` to a valid api key and `RECORD_REPLAY_APP_URL=https://devtools-git-fork-ryanjduffy-ryan-support-a-335c95-recordreplay.vercel.app`

src/ui/components/Library/SidebarFooter.tsx
```diff
@@ -6,7 +6,7 @@ import { AvatarImage } from "../Avatar";
 
 function SidebarFooter({ setModal }: PropsFromRedux) {
   const { user } = useAuth0();
-  const { name, picture } = user!;
+  const { name, picture } = user || { name: "", picture: "" };
```
jasonLaster, line 9: this feels like `useAuth0` should have some better defaults

src/ui/utils/useAuth0.ts
```diff
@@ -22,6 +24,25 @@ export type AuthContext = Auth0ContextInterface | typeof TEST_AUTH;
...
   const auth = useOrigAuth0();
+  const { loading, name, email, picture } = useGetUserInfo();
+  const { token, apiKey } = useToken();
+
+  if (apiKey && token) {
+    return {
+      isLoading: loading,
+      isAuthenticated: true,
+      user: loading
+        ? undefined
```
jasonLaster, line 35: maybe return an empty user object instead

pages/_app.tsx
```diff
@@ -186,4 +187,24 @@ const AppRouting = ({ Component, pageProps }: AppProps) => {
   );
 };
 
+interface AuthProps {
+  apiKey?: string;
+}
+AppRouting.getInitialProps = (appContext: AppContext) => {
+  const props = NextApp.getInitialProps(appContext);
+  const authHeader = appContext.ctx.req?.headers.authorization;
```
jasonLaster, line 195: wait, how is this passed in?

src/ui/components/Library/SidebarFooter.tsx
ryanjduffy: I think we're being optimistic here since the auth0 does mark `user` as possibly undefined. I think it's reasonable to handle this `undefined` here rather than receiving an empty or partial object and inferring the loading state from that.

pages/_app.tsx
ryanjduffy: It's coming from the server when you pass in the `Authorization` HTTP header. Check out https://github.com/RecordReplay/playwright-tests/pull/104/files#diff-4a2dc5597012da2f007b9fe6f3b92dc7efc11c4be73cd04f34a52b08523c62dfR57-R59 for how we could pass it in for playwright.

jasonLaster: wow! looks great

hbenl (approved):
LGTM

PR #4459, opened 2021-11-11
semgrep: fix noopener/noreferrer errors

jazzdan:


jaril (approved):
Looks good. We can probably clean it up by creating a separate Link component that's just `target="_blank" rel="noreferrer noopener"`.

Don't feel too strongly about that as a blocker since semgrep will tell us if we miss it. Can also be a quick follow-up.
(_edit: filed it [here](#4460) so we're covered._)

jazzdan:
Thanks for the issue @jaril!

PR #4462, opened 2021-11-11
Add Storybook

### LINKED ISSUE #4349: Add Storybook
https://storybook.js.org/ will help us 

* document our design system and important ui components
* be a nice development environment for new components
* help us think through how to use replay to look into visual regressions in CI

jcmorrow:
### Changes

- Add Storybook and dependencies. This was not that bad! I wrestled with PostCSS for a while, but ultimately I just had to get everything on the right version number. I wish there had been more of a "gotchya" moment in this struggle, but it was really just the nth time I ran `npm install` and it worked.
- move some css variables up to `:root`. This makes sure that those variables are also present in Storybook.
- Add stories for network monitor. Notice that I export by default the connected request table, but I also export the non-connected RequestTable component, because that is easier to plug into Storybook.
- In the future, we'll probably want to *slightly* consolidate the number of elements that have to be connected, because that makes prototyping very fast, but the long-term plan should be to properly setup a Storybook Redux store (probably just a partial one) and get our Apollo mocks into a place where they can be used in Storybook as well as our tests.

Fixes #4349


postcss.config.js
```diff
@@ -2,20 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-var mapUrl = require("postcss-url-mapper");
```
jasonLaster, line 5: love seeing this cruft deleted

jasonLaster (approved):
Looking good! It's much easier to see this stuff in its own PR. Especially the postcss stuff

postcss.config.js
jcmorrow: You know you can always count on me to find the unused stuff 🔨

jasonLaster: there are stories in that code! 😉

colbyr (commented):
Just tried this out! looks awesome 👏

Issue #4423, opened 2021-11-10
Arrow keys cause the elements panel to scroll


jasonLaster:
**Description:**
https://app.replay.io/recording/72fac77a-1ba5-4ce3-afa7-977a26286492?point=46081634643218528162481821055977377&time=18744&hasFrames=false

PR #4468, opened 2021-11-12
Call preventDefault on cursor key events in the markup panel (#4423)

hbenl:
There are 2 reasons why the markup panel scrolls when using the cursor keys: the browser's default action for those keyboard events in a scrollable panel and the Node component [calling](https://github.com/RecordReplay/devtools/blob/15c45a443ca05914d1af5ab4ed8374940158e77b/src/devtools/client/inspector/markup/components/Node.tsx#L70) `scrollIntoView()` on its DOM node.
This PR only disables the first mechanism but also prepares everything to disable the second mechanism: all that is needed is to change the condition [here](https://github.com/RecordReplay/devtools/blob/15c45a443ca05914d1af5ab4ed8374940158e77b/src/devtools/client/inspector/markup/markup.js#L213). But the scrolling introduced by the second mechanism feels right to me.

### DIFF
--- a/src/devtools/client/inspector/markup/components/Nodes.tsx
+++ b/src/devtools/client/inspector/markup/components/Nodes.tsx
@@ -11,7 +11,7 @@ import {
   onUpKey,
 } from "../actions/markup";
 import useKeyShortcuts from "ui/utils/use-key-shortcuts";
-import { cancelBubbling } from "ui/utils/key-shortcuts";
+import { cancelBubbling, preventDefault } from "ui/utils/key-shortcuts";
 import Node from "./Node";
 import { MarkupProps } from "./MarkupApp";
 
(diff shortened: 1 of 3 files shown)


jcmorrow (approved):
👍

src/devtools/client/inspector/markup/components/Nodes.tsx
```diff
@@ -21,12 +21,12 @@ function Nodes(props: MarkupProps & PropsFromRedux) {
...
-      Left: cancelBubbling(onLeftKey),
-      Right: cancelBubbling(onRightKey),
-      PageUp: cancelBubbling(onPageUpKey),
-      PageDown: cancelBubbling(onPageDownKey),
+      Up: cancelBubbling(preventDefault(onUpKey)),
+      Down: cancelBubbling(preventDefault(onDownKey)),
+      Left: cancelBubbling(preventDefault(onLeftKey)),
+      Right: cancelBubbling(preventDefault(onRightKey)),
+      PageUp: cancelBubbling(preventDefault(onPageUpKey)),
+      PageDown: cancelBubbling(preventDefault(onPageDownKey)),
```
jasonLaster, line 29: Nice

PR #4470, opened 2021-11-12
Create a shared Anchor component

### LINKED ISSUE #4460: Create a link component
This should be a quick one. We have a number of links across the app that are always `target="_blank" rel="noreferrer noopener"`. Would be pretty handy to just a `<Link>` component that comes with those out of the box, instead of possibly forgetting it.

Referencing #4459.

jaril:
Fix #4460.

### DIFF
--- a/src/ui/components/shared/UserSettingsModal/UserSettingsModal.tsx
+++ b/src/ui/components/shared/UserSettingsModal/UserSettingsModal.tsx
@@ -19,6 +19,7 @@ import { SettingsBodyHeader } from "../SettingsModal/SettingsBody";
 import { getFeatureFlag } from "ui/utils/launchdarkly";
 import { AvatarImage } from "ui/components/Avatar";
 import PreferencesSettings from "./PreferencesSettings";
+import ExternalLink from "../ExternalLink";
 
 function Support() {
   return (
@@ -28,9 +29,7 @@ function Support() {
           <SettingsBodyHeader>Join us on Discord</SettingsBodyHeader>
           <div className="description">
             Come chat with us on our{" "}
-            <a href="https://discord.gg/n2dTK6kcRX" target="_blank" rel="noopener noreferrer">
-              Discord server.
-            </a>
+            <ExternalLink href="https://discord.gg/n2dTK6kcRX">Discord server.</ExternalLink>
           </div>
         </label>
       </li>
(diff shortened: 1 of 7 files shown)


src/ui/components/shared/UserSettingsModal/UserSettingsModal.tsx
```diff
@@ -28,9 +29,9 @@ function Support() {
           <SettingsBodyHeader>Join us on Discord</SettingsBodyHeader>
           <div className="description">
             Come chat with us on our{" "}
-            <a href="https://discord.gg/n2dTK6kcRX" target="_blank" rel="noopener noreferrer">
+            <Anchor href="https://discord.gg/n2dTK6kcRX" target="_blank" rel="noopener noreferrer">
```
jcmorrow, line 32: Maybe you forgot to remove the `target` and `rel` here?

src/ui/components/shared/WorkspaceSettingsModal/TeamPricingPage.tsx
```diff
@@ -35,14 +36,14 @@ export function TeamPricingPage({
       >
         Add Payment Method
       </Button>
-      <a
+      <Anchor
```
jcmorrow, line 39: Same here?

src/ui/components/shared/WorkspaceSettingsModal/TeamPricingPage.tsx
```diff
@@ -15,15 +16,15 @@ export function TeamPricingPage({
       <p className="mb-4">
         With a Replay Team Plan, you can expand your debugging superpowers with powerful
         collaboration features that make it easy to work together to fix bugs and understand your
-        software better.{" "}
-        <a
+        software better.
+        <Anchor
```
jcmorrow, line 20: Same here

src/ui/components/shared/Anchor.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { AnchorHTMLAttributes, ReactNode } from "react";
+
+export default function Anchor({
```
jcmorrow, line 3: I dunno about you, but `Anchor` doesn't mean much to me in a vacuum. I know `Link` is easily confused with `Next/Link`, but what if we called this like `ExternalLink`, since that's really what it's doing, making it safe to link to external resource?

jcmorrow (approved):
Couple of little things, but I think this is a good change overall 👍

src/ui/components/shared/UserSettingsModal/UserSettingsModal.tsx
jaril: Fixed the type to catch this!

src/ui/components/shared/Anchor.tsx
jaril: Switched to ExternalLink

PR #4472, opened 2021-11-12
Get recording ID from the URL

jcmorrow:
This seems to work just fine locally, @ryanjduffy maybe you have the best context on whether this is a bad idea or not? 

For context, here is the issue we are dealing with in tests:

https://app.replay.io/recording/e96100a3-dcb1-4427-9817-5f2dbe627918


ryanjduffy:
I think it's generally fine but I'd be surprised that the result is any different since they are nearly the same. The only advantage to the hook is that it abstracts away the silly `string | string[]` typing which is failing typechecking on this PR.

jcmorrow:
Yep, nvm, you are correct, https://app.replay.io/recording/988e092d-aee0-42f7-a451-1ca0bcee6ea7 came through on this branch. Argh.

PR #4473, opened 2021-11-12
Don't start loading the recording page until our URL params have hydrated

jcmorrow:
Follow-up to my misguided attempt in https://github.com/RecordReplay/devtools/pull/4472. This still uses `getRouter` under the hood as @ryanjduffy pointed out, but now we shouldn't get stuck in a 404 state ever.


pages/recording/[id].tsx
```diff
@@ -15,14 +15,14 @@ function Recording({ getAccessibleRecording }: PropsFromRedux) {
   const [recording, setRecording] = useState<RecordingInfo | null>();
   const [uploadComplete, setUploadComplete] = useState(false);
   useEffect(() => {
-    if (!store) return;
+    if (!store || !recordingId) return;
```
Comment (ryanjduffy) [inline pages/recording/[id].tsx:18]: 💯

PR #4483, opened 2021-11-13
Fix Apollo Mocks

jcmorrow:
This gets the final CI check passing. The primary changes are:
- Update Apollo mocks to look like our updated API responses
- Take playwright out of headless mode (not sure why is necessary)
- Tell semgrep to ignore the `test` folder
- Do lots of miscellaneous sorting in files that I touched during this change


src/ui/components/App.tsx
```diff
@@ -98,7 +98,7 @@ function App({ children, fontLoading, modal, theme }: AppProps) {
   }, [theme]);
 
   useEffect(() => {
-    if (!isTest() && auth.isAuthenticated) {
+    if (!(isMock() || isTest()) && auth.isAuthenticated) {
```
jasonLaster, line 101: wonder if isMock can be part of isTest... Not sure

src/ui/utils/apolloClient.tsx
```diff
@@ -119,8 +119,15 @@ export function extractGraphQLError(error: ApolloError | undefined): string | un
...
+        fields: {
+          viewer: {
+            merge: true,
+          },
+        },
+      },
       AuthenticatedUser: {
-        keyFields: [],
+        merge: true,
       },
```
jasonLaster, line 131: what's this trying to do?

jasonLaster (commented):
I can see how these tests could quickly go stale...

jcmorrow:
> I can see how these tests could quickly go stale...

I think so, but one thing that will help is adding types to these Mocks. That way if we add or remove a field it's not just that the mocktests console.log but also that the typescript lint fails.

src/ui/utils/apolloClient.tsx
jcmorrow: This is interesting! So, Apollo's cache is a flattened, normalized store. This is so that when highly nested objects come in not just the high-level resource is cached, but so are the nested objects. It's a good idea, but it's also an ambitious thing to do, and goes wrong in interesting ways.

By default, the Apollo cache will look for an `id` field in nested resources, and if it finds that it will apply some merging logic to make sure that item gets updated in the store. If it doesn't find that ID key, it blows up. But there are actually many reasonable ways to handle that situation, if you have context on what the objects purpose is. In this case, we know that `AuthenticatedUser` doesn't have an id because there is only ever one of them. This means that anytime we get an AuthenticatedUser in a response, it should overwrite the cache. Adding `merge: true` specifies that behavior. There are lots of other things you can do as well, see the docs for more info: https://www.apollographql.com/docs/react/caching/cache-field-behavior/#the-merge-function

src/ui/components/App.tsx
jcmorrow: Yeah, that's probably reasonable. Let me make that change.

PR #4488, opened 2021-11-13
Begin virtualizing the outline view

jasonLaster:
This is a work in progress

* react-window is mostly working
* refactoring the internal state is 75% there


jasonLaster:
After thinking about this a bit more I think this warrants a new approach

1. start with a new functional component in storybook. the old outline view is pretty messy
2. move some of the state to redux. We should be able to pass in the functions and classes. The component doesn't need that

PR #4489, opened 2021-11-13
Fix right most padding in the UI

### LINKED ISSUE #4486: Chrome background not drawn correctly
image

Notice the bottom right side. The gray should extend all the way to the side.

jasonLaster:
fixes #4486


jcmorrow (approved):
Seems good to me! I think if we're going to keep this much padding it might be good to consider adding some kind of panel border so that it fits with the other white-background panels, but this looks much better than the lop-sided padding we have now! ⚡

jasonLaster:
Yep - i'm also in favor of a border. 

CC @jaril for help on the highlighter

jaril:
Closed by #4516.

PR #4503, opened 2021-11-15
Reenable returning to a route after login

hbenl:
@ryanjduffy You had disabled this for the Next.js migration, but when I just tested it, it seemed to work fine. Did you see any issues that I missed?


ryanjduffy:
This was probably a mistake. In don't remember changing it on purpose. If it's working locally, let's give it a go.

Issue #4505, opened 2021-11-15
NodePicker assertion fails


hbenl:
Selecting a node with the NodePicker before the elements panel is initialized fails an assertion:
https://app.replay.io/recording/a27c9bb1-db17-4e19-9cce-a8e33ab2293c

PR #4506, opened 2021-11-15
Wait for the root node to be added to the inspector tree (#4505)

hbenl:
Before:
https://app.replay.io/recording/a27c9bb1-db17-4e19-9cce-a8e33ab2293c
After:
https://app.replay.io/recording/b7ffd3ab-9485-437b-85a2-e33116941663

### DIFF
--- a/src/devtools/client/inspector/markup/actions/markup.ts
+++ b/src/devtools/client/inspector/markup/actions/markup.ts
@@ -1,5 +1,5 @@
 import { Action } from "redux";
-import { assert } from "protocol/utils";
+import { assert, defer, Deferred } from "protocol/utils";
 import { ThreadFront } from "protocol/thread";
 import { NodeFront } from "protocol/thread/node";
 import Selection from "devtools/client/framework/selection";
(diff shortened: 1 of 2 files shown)


src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -50,12 +56,26 @@ export function reset(): ResetAction {
 /**
  * Clears the tree and adds the new root node.
  */
-export function newRoot(rootNodeFront: NodeFront): UIThunkAction {
+export function newRoot(): UIThunkAction {
   return async ({ dispatch }) => {
+    const pause = ThreadFront.currentPause;
```
jasonLaster, line 61: forgot if we do this now, but passing threadFront through as an option would be nice


```js
return async ({ dispatch, threadFront }) => {
}

```

jasonLaster (approved):

src/devtools/client/inspector/markup/actions/markup.ts
hbenl: No, this currently isn't added to the thunk args.

PR #4511, opened 2021-11-15
Fix events overflow

jaril:
There's an overflow issue in the events panel that's causing the last part to be pushed down. Likely related to `height: 100%`.

Replay with a lot of events that's nice to test this with: https://app.replay.io/recording/fc8dade7-d55c-41aa-8f34-8476f809f5d4

Before: 
image

After: 
image

Demo:

https://user-images.githubusercontent.com/15959269/141844629-751e03d2-c712-4629-9866-05de6bdb2716.mov

### DIFF
--- a/src/devtools/client/debugger/src/components/shared/Accordion.css
+++ b/src/devtools/client/debugger/src/components/shared/Accordion.css
@@ -13,6 +13,11 @@
   height: 100%;
 }
 
+.accordion > li {
+  display: flex;
+  flex-direction: column;
+}
+
 .accordion ._header {
   background-color: #fff;
   display: flex;
--- a/src/devtools/client/debugger/src/components/shared/Accordion.css
+++ b/src/devtools/client/debugger/src/components/shared/Accordion.css
@@ -73,6 +73,5 @@
 .accordion ._content {
   font-size: var(--theme-body-font-size);
   overflow-y: auto;
-  height: 100%;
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/components/shared/Accordion.css
```diff
@@ -13,6 +13,11 @@
   height: 100%;
 }
 
+.accordion > li {
+  display: flex;
+  flex-direction: column;
+}
```
jcmorrow, line 19: Is this doing anything?

jcmorrow (approved):
Removing the 100% height is great. Not sure about the other changes.

src/devtools/client/debugger/src/components/shared/Accordion.css
jaril: Yeah — without this, the entire Events panel (header+content/events) overflows the sidebar container.

jcmorrow: Hmmmmmmm maybe we should look at this together. I'm not seeing that.

jcmorrow: Ah, I see the problem. It's because we have overflow-hidden elsewhere. Well, let's say this is fine for now, I'll fix it up as part of my accordion CSS overhaul.

Issue #4496, opened 2021-11-13
What gets rendered in the Viewer changes based upon my mouse leaving the breakpoint "box"


gideonred:
Here's a recording with comments explaining the unexpected behavior: https://app.replay.io/recording/49d2725b-af09-4f1d-ae1c-7828bf2c3fcd

jasonLaster:
cc @jonbell-lot23 curious what you think of this loading state

jonbell-lot23:
It's a tough one. We have two extreme options, and the answer will be somewhere in the middle:

1. Never show loading state, meaning surprising jumps like this (current approach)
2. Show loading state, which can feel heavy

Here's a "loading states powers of ten" video I made for situations like this:
https://www.loom.com/share/6a73dcc7aba643709e8366f60fa57232

I don't think it's tenable to never show loading states. I understand a reluctance to put a spinner up there every time we seek in the video, but the alternative is this sort of disjointed and jumpy experience. So I agree with @jasonLaster 's comment in the Replay:

"We should show probably show a spinner in the bottom right of the video to make it clear that we're still painting..."

gideonred:
FYI, I don't think this was loading in the background and catching up. There's a clear change happening as my mouse is leaving the breakpoint dialogue. If I left my mouse in the breakpoint dialogue the repaint never happens.

jonbell-lot23:
@gideonred Never mind, then! My loading stuff is relevant overall but maybe not for this particular issue.

PR #4518, opened 2021-11-16
Don't redraw the current screenshot if it hasn't been replaced earlier (#4496)

hbenl:
`clearHoveredItem()` triggers redrawing the current screenshot because it may have been replaced earlier in `setHoveredItem()`. Since `setTimelineToTime()` only draws _recorded_ screenshots, this will replace a repainted screenshot.
The redrawing logic is only necessary if the user has enabled one of the hover features (which are disabled by default).
This PR will only disable the redrawing in case the hover features are disabled. If one of the hover features is enabled, the user will still notice a bug: after hovering, the screenshot will be replaced with the current _recorded_ screenshot instead of the current _repainted_ screenshot. I didn't want to spend time fixing this because I think we should probably remove the hover features instead.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -15,6 +15,7 @@ import {
 } from "protocol/graphics";
 import {
   getCurrentTime,
+  getHoveredItem,
   getHoverTime,
   getPlayback,
   getTrimRegion,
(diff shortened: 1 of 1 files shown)


gideonred:
Thanks for looking into this Holger!

jaril (approved):
LGTM!

PR #4519, opened 2021-11-16
Standardize on 8px between horizontal panels

jcmorrow:
Fixes https://github.com/RecordReplay/devtools/issues/4360

After:

CleanShot 2021-11-16 at 08 06 39

CleanShot 2021-11-16 at 08 06 27

### DIFF
--- a/src/ui/components/App.css
+++ b/src/ui/components/App.css
@@ -1,8 +1,8 @@
 #video {
-  height: 100%;
-  position: relative;
-  overflow: hidden;
   flex-grow: 1;
+  height: 100%;
+  overflow: hidden;
+  position: relative;
 }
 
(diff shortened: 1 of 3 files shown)


jasonLaster:
4 feels a bit narrow, but tough to say

jcmorrow:
Yeah, I sort of agree, that's why I initially had the handles as 8px. @jonbell-lot23 what do you think?

jonbell-lot23:
@jcmorrow : I don't think 4px is too tight, but I am totally fine with 8px. The key is that they are consistent because we were juggling a lot of inconsistency before.

jcmorrow:
I'm going to make it consistently 8px for now. If we want to change it to 4 in the future, that is easy to do!

PR #4522, opened 2021-11-16
Add a toggle for network settings in user settings

jaril:
This allows us to toggle the network panel feature from localstorage using the user settings modal.

There's a larger follow up for us to do which would wrap the features in a hook so that when it's toggled, everything updates automatically, i.e. While in the network panel, if you toggle it off, you should be yanked into the console. For now, we've added a `restart required` notice on it to communicate the UI's limitations.

### DIFF
--- a/src/ui/components/SecondaryToolbox/index.tsx
+++ b/src/ui/components/SecondaryToolbox/index.tsx
@@ -138,9 +138,7 @@ function SecondaryToolbox({
   const showReact = userSettings.showReact;
   const isNode = recordingTarget === "node";
 
-  const allowedPanels = getAllowedPanels(userSettings);
-
-  if (!allowedPanels.includes(selectedPanel)) {
+  if (!getAllowedPanels(userSettings).includes(selectedPanel)) {
     setSelectedPanel("console");
   }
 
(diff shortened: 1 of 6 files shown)


jaril:
Sorting out a ts error...

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -121,6 +138,12 @@ function SecondaryToolbox({
   const showReact = userSettings.showReact;
   const isNode = recordingTarget === "node";
 
+  const allowedPanels = getAllowedPanels(userSettings);
+
+  if (!allowedPanels.includes(selectedPanel)) {
```
jcmorrow, line 143: maybe remove the local here, so `getAllowedPanels(userSettings).includes...`

src/ui/hooks/settings.ts
```diff
@@ -6,6 +6,8 @@ import { SettingItemKey } from "ui/components/shared/SettingsModal/types";
 import useAuth0 from "ui/utils/useAuth0";
 import type { UserSettings } from "../types";
 import { ADD_USER_API_KEY, DELETE_USER_API_KEY, GET_USER_SETTINGS } from "ui/graphql/settings";
+import { features } from "ui/utils/prefs";
+import { useEffect, useState } from "react";
```
jcmorrow, line 10: Probably remove these?

jcmorrow (approved):
Tiny suggestions, looks great! 👍

PR #4524, opened 2021-11-16
Make the sourced accordion into a stacked accordion

### LINKED ISSUE #3200: Pause panel should use a stacked sidebar accordion

jcmorrow:
This adds a new Accordion class, which shares an interface with the old one, but is IMO considerably simpler. Right now it only has a stacked mode, but it's quite simple to add a non-stacked mode by setting `flex-basis: 0` and `flex-grow: 1` on open children. One thing that I don't love about this design is that the heights use a content basis, which means that if there is any overflow, the shrunken space will be shared evenly by the children, but their final heights will reflect differences in their contents. This means in situations where there are contents with very different heights you will have very uneven heights after shrinking as well. There are a lot of examples in storybook to play around with.

CleanShot 2021-11-16 at 11 44 54

CleanShot 2021-11-16 at 11 44 42

Fixes #3200 (finally)

### DIFF
--- a/src/devtools/client/debugger/src/components/shared/Accordion.css
+++ b/src/devtools/client/debugger/src/components/shared/Accordion.css
@@ -75,3 +75,15 @@
   overflow-y: auto;
   background-color: white;
 }
+
+.accordion li {
+  flex: 1;
+  flex-grow: 0;
+  flex-shrink: 1;
+  flex-basis: auto;
+  min-height: fit-content;
+}
+
+.accordion li.expanded {
+  min-height: 20%;
+}
--- a/src/ui/components/Accordion.module.css
+++ b/src/ui/components/Accordion.module.css
@@ -1,0 +1,46 @@
+.accordion {
+  border-radius: 8px;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
+
+.accordionItemContainer:first-child {
+  border-top-left-radius: 8px;
+  border-top-right-radius: 8px;
+  border-top: none;
+}
+
+.accordionItemContainer.open {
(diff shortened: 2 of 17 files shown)


jasonLaster:
Nice use case for story book

src/devtools/client/debugger/src/components/shared/Accordion.css
```diff
@@ -68,6 +68,20 @@
...
+  flex-basis: auto;
+  min-height: fit-content;
+}
+
+.accordion li.expanded {
+  min-height: 20%;
+}
+
+.accordion li.lastOpen {
+}
```
jaril, line 87: Unused style.

src/ui/components/Accordion.module.css
```diff
@@ -0,0 +1,46 @@
...
+
+.accordionItemContainer:not(.open) {
+  flex-shrink: 0;
+}
+
+.accordionItemContainer .header {
+  background: white;
+  cursor: pointer;
+  font-size: 15px;
+  line-height: calc(16 / 12);
```
jaril, line 32: Small nit but we can probably get away with just `1.25`.

jaril (approved):

src/ui/components/Accordion.module.css
jcmorrow: Yeah, this was pre-existing, but I agree with you, I'll change it 👍

PR #4525, opened 2021-11-16
Make secondary buttons in the library have white text

### LINKED ISSUE #4361: Library button colours are off (also fix the cancel button in our comments confirmation)
The text here should be blue:
image

And the background here should be white:
image

jaril:
Fix #4361.

image
image


jasonLaster (approved):
Looks good!

Maybe add to SB?

PR #4526, opened 2021-11-16
Circular loader

jonbell-lot23:
We're trying to make a global loader like this:
image

([Figma link](https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/Replay-Design-Doc?node-id=8344%3A58248) and [GitHub link](https://github.com/RecordReplay/devtools/issues/4337)

Right now the code sort of works ... but now the play button isn't clickable because the loader is overlapping. We need to fix it with some z-index magic.


jaril:
Moved to #4536.

PR #4527, opened 2021-11-16
Add support for SAML auth callbacks

ryanjduffy:
## Issue

Auth0's SAML integration forwards a `code` but doesn't include the `state` parameter which seems to be required for the SDK to validate the authentication

## Resolution

Adds a new page, `/callback` which forces a second authentication loop with the identity provider to complete login. We'll need to fix this later but this will get us by for now.


src/ui/utils/tokenManager.tsx
```diff
@@ -13,8 +13,8 @@ const tokenRefreshSecondsBeforeExpiry = 60;
 
 // Auth0 E2E tests require a different clientId
 const clientId: string = usesWindow(win => {
-  const isTesting = win ? new URL(win.location.href).searchParams.get("e2etest") : false;
-  return isTesting ? "uL3A8jxVqotF5Q6bmn5QTV46hU97MPQm" : "4FvFnJJW4XlnUyrXQF8zOLw6vNAH1MAo";
+  // const isTesting = win ? new URL(win.location.href).searchParams.get("e2etest") : false;
+  return "hIAHByYBMYKvKUh06Sro6vAgrMcolImb"; // isTesting ? "uL3A8jxVqotF5Q6bmn5QTV46hU97MPQm" : "4FvFnJJW4XlnUyrXQF8zOLw6vNAH1MAo";
```
ryanjduffy, line 17: This is a temporary change to test the SAML integration and will be removed for integration

ryanjduffy:
I've updated the Auth0 config to allow this new path as a callback url and enabled the relevant SAML config for our app as well.

jaril (approved):

PR #4531, opened 2021-11-17
Add a `useFeature` hook

jcmorrow:
This sets an observer on our preferences store and allows for a component to refresh when that preference changes. This is used for `features`, but it could easily be expanded to other `prefs`. This allows us to update the secondary toolbox to add or remove the network monitor without refreshing the page.

### DIFF
--- a/src/ui/hooks/settings.ts
+++ b/src/ui/hooks/settings.ts
@@ -6,6 +6,9 @@ import { SettingItemKey } from "ui/components/shared/SettingsModal/types";
 import useAuth0 from "ui/utils/useAuth0";
 import type { UserSettings } from "../types";
 import { ADD_USER_API_KEY, DELETE_USER_API_KEY, GET_USER_SETTINGS } from "ui/graphql/settings";
+import { features } from "ui/utils/prefs";
+import { prefs as prefsService } from "devtools-services";
+import { useEffect, useState } from "react";
 
 const emptySettings: UserSettings = {
   apiKeys: [],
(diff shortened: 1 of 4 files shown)


src/ui/hooks/settings.ts
```diff
@@ -63,6 +66,20 @@ export function useGetUserSettings() {
...
+export const useFeature = (prefKey: keyof typeof features) => {
+  const [pref, setPref] = useState(features[prefKey]);
+
+  const onUpdate = (prefs: any) => {
+    setPref(prefs.getBoolPref(`devtools.features.${prefKey}`));
+  };
+
+  useEffect(() => {
+    prefsService.addObserver(`devtools.features.${prefKey}`, onUpdate, false);
+  }, []);
```
loganfsmyth, line 78: Random review:

I think useEffect in this way is at risk of continuously adding observers as the structure of the page changes, right? At least if the component using this hook were consistently rendered and then removed.

Seems like we'd want more like:
```
export const useFeature = (prefKey: keyof typeof features) => {
  const [pref, setPref] = useState(features[prefKey]);

  const prefName = `devtools.features.${prefKey}`;

  useEffect(() => {
    // Callback inside the hook since it isn't used outside anywhere
    const onUpdate = (prefs: any) => {
      setPref(prefs.getBoolPref(prefName));
    };

    prefsService.addObserver(prefName, onUpdate, false);
    // Include cleanup hook
    return () =>  prefsService.removeObserver(prefName, onUpdate, false);
  }, [
    // Include effect pref name
    prefName,
  ]);

  return pref;
};
```

jcmorrow: Definitely agree on the remove observer call. I remember thinking I should do that and then clearly I forgot 😆

jcmorrow: I kind of thought that the cleanup function would get run at the end of the component lifetime, but just looked it up and found out that react will do the cleanup before processing the next call 👍  Thanks for the suggestions @loganfsmyth !

src/ui/hooks/settings.ts
```diff
@@ -63,6 +66,20 @@ export function useGetUserSettings() {
...
+export const useFeature = (prefKey: keyof typeof features) => {
+  const [pref, setPref] = useState(Boolean(features[prefKey]));
+
+  useEffect(() => {
+    const onUpdate = (prefs: any) => {
+      setPref(prefs.getBoolPref(`devtools.features.${prefKey}`));
+    };
+
+    prefsService.addObserver(`devtools.features.${prefKey}`, onUpdate, false);
+  }, [prefKey]);
```
loganfsmyth, line 78: Looks like the remove hook never got in here?

PR #4533, opened 2021-11-17
Add a button for opening quick open

jaril:
This adds an icon right by the sources panel that when clicked, opens quick open.

https://user-images.githubusercontent.com/15959269/142239701-b4fa0170-131f-4f44-9dc5-090441bf531e.mov


jasonLaster:
Love it!

PR #4536, opened 2021-11-17
Move the circular loader to the play button

jaril:
image


src/ui/components/Toolbar.tsx
```diff
@@ -149,25 +108,10 @@ function Toolbar({
...
-      <div className="flex flex-col space-y-1 items-center">
-        {progressPercentage !== 100 && viewMode === "dev" ? (
-          <IndexingLoader {...{ progressPercentage, viewMode }} />
-        ) : null}
-      </div>
     </div>
   );
 }
 
-const Help = () => (
```
jaril, line 161: This was unused

PR #4537, opened 2021-11-17
Add a request detail view to the network tab

jcmorrow:
### Changes
- Add a `RequestDetail` component that will show a more detailed view of the selected request
- Put that detail view into a SplitBox in the Network tab
- Change all tabs in the SecondaryToolbox to toggle their visibility, rather than their presence in the DOM, so that it's more likely to keep state as people are switching tabs.
- Add headers to the storybook mock data
- Try and get storybook to work with request table and request detail components. One frustrating thing about Storybook is that when it doesn't work it just... doesn't work. The components don't appear in the sidebar, but there are no errors, or any signal about what's wrong. The only thing I've realized here is that calling `connect` in a file can break Storybook *even if the component you are important from that file *isn't* the connected component. I'm not digging further into this right now, it's just not worth the rabbit hole 🐰 .

### Video

https://user-images.githubusercontent.com/5903784/142294108-3b2a6e46-80ee-46e2-b020-d53c82d95ea6.mp4

### DIFF
--- a/src/ui/components/SecondaryToolbox/index.tsx
+++ b/src/ui/components/SecondaryToolbox/index.tsx
@@ -16,8 +16,9 @@ import ToolboxOptions from "./ToolboxOptions";
 import { trackEvent } from "ui/utils/telemetry";
 
 import "ui/setup/dynamic/inspector";
-import { ConnectedRequestTable } from "../NetworkMonitor/RequestTable";
 import { UserSettings } from "ui/types";
+import NetworkMonitor from "../NetworkMonitor";
+import classNames from "classnames";
 
 const InspectorApp = React.lazy(() => import("devtools/client/inspector/components/App"));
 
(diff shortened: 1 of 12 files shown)


jasonLaster:
> I'm not digging further into this right now, it's just not worth the rabbit hole 🐰 .

Yep - but if you record a replay, you can file an issue in Storybook with it and I'll CC gert 😉

jcmorrow:
> Yep - but if you record a replay, you can file an issue in Storybook with it and I'll CC Gert. Every bug is a sales opportunity! 😉

Ha! Yeah the problem is that I'm pretty sure it's a silent webpack failure. Nothing is wrong on the client-side, it's just like the client isn't even aware that there's a story to display ☹️

jasonLaster:
heh - might still be a nice opportunity to confirm the hypothesis... but yeah, shrug

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -157,12 +161,22 @@ function SecondaryToolbox({
...
+        </div>
+        {/* {selectedPanel === "inspector" ? <InspectorPanel /> : null} */}
+        <div className={classNames("h-full", { hidden: selectedPanel !== "inspector" })}>
+          <InspectorPanel />
+        </div>
+        {/* {selectedPanel === "react-components" ? <ReactDevtoolsPanel /> : null} */}
+        <div className={classNames("h-full", { hidden: selectedPanel !== "react-components" })}>
           <ReactDevtoolsPanel />
-        ) : null}
+        </div>
```
jasonLaster, line 179: this code is less dark, but probably one of those things you just have to try and see how it feels

jonbell-lot23:
Aaaaaa this is rad

PR #4538, opened 2021-11-17
implement searchSources in ThreadFront

colbyr:
I was hoping to get some feedback on implementing the [search protocol](https://github.com/RecordReplay/backend/pull/3499) here in `ThreadFront`.

This is the protocol for reference:
```pdl
  type SearchSourceContentsMatch extends object
    properties
      Location location
      # Some snippet of text from around the match
      string context
      # The beginning of the match within the context snippet
      SourceLocation contextStart
      # The end of the match within the context snippet
      SourceLocation contextEnd


  command searchSourceContents
    parameters
      string searchId
      string query


  event searchSourceContentsMatches
    parameters
      string searchId
      array of SearchSourceContentsMatch matches
```

The search protocol is really similar to [findSources](https://github.com/RecordReplay/devtools/blob/a7584b763bb9e7c95e54878b7624305085c775d5/src/protocol/thread/thread.ts#L308-L337), but it differs in that `searchSources` will be called 0 or more times during a session, while `findSources` is called exactly 1 time.

The `searchSources` method takes a callback that receives batches of search results until the search is complete.

```ts
await ThreadFront.searchSources({query: 'some search string'}, (matches) => {
  // this callback will run 1 or more times with batches of search results
})
// once the searchSources promise resolves the search is complete
console.log('search is done')
```

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -31,12 +31,15 @@ import {
   SameLineSourceLocations,
   RequestEventInfo,
   RequestInfo,
+  searchSourceContentsMatches,
+  SearchSourceContentsMatch,
 } from "@recordreplay/protocol";
 import { client, log } from "../socket";
 import { defer, assert, EventEmitter, ArrayMap } from "../utils";
 import { MappedLocationCache } from "../mapped-location-cache";
 import { ValueFront } from "./value";
 import { Pause } from "./pause";
+import { uniqueId } from "lodash";
 
 export interface RecordingDescription {
   duration: TimeStamp;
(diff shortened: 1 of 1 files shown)


src/protocol/thread/thread.ts
```diff
@@ -336,6 +341,32 @@ class _ThreadFront {
...
+      this.searchWaiters = new Map();
+      client.Debugger.addSearchSourceContentsMatchesListener(
+        ({ searchId, matches }: { searchId: string; matches: SearchSourceContentsMatch[] }) => {
+          const searchWaiter = this.searchWaiters?.get(searchId);
+          if (searchWaiter) {
+            searchWaiter(matches);
+          }
+        }
+      );
+    }
```
colbyr, line 358: I put this lazy-initialization code inline to keep the search logic co-located. I've since noticed that that's a bit of a departure from the way other ThreadFront things are initialized though, so I can totally change that.

jasonLaster: yeah - i'd probably do it at the top to be consistent. there isn't a perf hit

jasonLaster (approved):
Looks solid

jasonLaster (approved):

PR #4553, opened 2021-11-18
Add file search to the left sidebar

jaril:
https://user-images.githubusercontent.com/15959269/142335985-5aa146d2-3585-4430-b810-4ca2c206d3da.mov


colbyr (approved):
Looks great to me! Would be cool if the search box was focused when the tab first appears. Not sure how hard that is though.

PR #4561, opened 2021-11-18
New search results

colbyr:
Fixes https://github.com/RecordReplay/devtools/issues/4429

v2 of https://github.com/RecordReplay/devtools/pull/4551 which, as @jaril suggested, loads search results in the react component rather than redux.

There's probably a bunch of redux code we can remove once this is in, but I'd rather do that separately from these changes.

Kapture 2021-11-18 at 10 29 53

https://app.replay.io/recording/e42ffda4-fa62-4c1f-ab96-22741d227a53?point=74639267355417640915662185020001708&time=38451&hasFrames=false

### DIFF
--- a/src/devtools/client/debugger/src/components/ProjectSearch.js
+++ b/src/devtools/client/debugger/src/components/ProjectSearch.js
@@ -13,7 +13,6 @@ import actions from "../actions";
 import { getEditor } from "../utils/editor";
 import { highlightMatches } from "../utils/project-search";
 
-import { statusType } from "../reducers/project-text-search";
 import { getRelativePath } from "../utils/sources-tree";
 import {
   getActiveSearch,
@@ -21,6 +20,7 @@ import {
   getTextSearchStatus,
   getTextSearchQuery,
   getContext,
+  getSources,
 } from "../selectors";
 
 import ManagedTree from "./shared/ManagedTree";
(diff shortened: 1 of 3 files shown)


jaril (approved):
LGTM!

src/devtools/client/debugger/src/components/ProjectSearch.js
```diff
@@ -29,6 +29,37 @@ import AccessibleImage from "./shared/AccessibleImage";
...
+      sourceId,
+      filepath: sourcesById[sourceId]?.url,
+      matches: matches.map(match => {
+        // We have to do this array dance to navigate the string in unicode "code points"
+        // because `colunm` is calculated using "code points" as opposed to JS strings
+        // which use "code units". It makes a difference in string with fun unicode characters.
+        const matchStr = [...match.context]
+          .slice(match.contextStart.column, match.contextEnd.column)
+          .join("");
+        return {
```
jasonLaster, line 50: haha I see.

Maybe we want to create a new helper that will make this more obvious and also let us write some unit tests down the line.

`unicodeSlice(match.context, match.contextStart.column, match.contextEnd.column)`

src/devtools/client/debugger/src/components/ProjectSearch.js
```diff
@@ -120,7 +179,7 @@ export class ProjectSearch extends Component {
     }
     this.setState({ focusedItem: null });
     const query = sanitizeQuery(this.state.inputValue);
-    if (query) {
+    if (query && query !== this.state.query && query.length >= 3) {
```
jasonLaster, line 182: nice

jasonLaster (approved):
Nice. One comment, but definitely shippable

src/devtools/client/debugger/src/components/ProjectSearch.js
colbyr: @jasonLaster yeah I like that idea. I'm going merge to unblock @jaril and then follow up with a helper.

PR #4567, opened 2021-11-18
Codepoint helpers

colbyr:
RE: https://github.com/RecordReplay/devtools/pull/4561#discussion_r752427573

Adds the code point string helpers @jasonLaster suggested.


jasonLaster (approved):
looks good. Also - would make for a great post when we look for 👻

PR #4571, opened 2021-11-18
Outline perf improvements

colbyr:
Related to https://github.com/RecordReplay/devtools/issues/4442

I don't know if I'm going to have time re-write the outline view with `react-window` before EOD tomorrow. I'm gonna try but in the meantime, this fixes the most egregious problems with the one we have now.

Right now every component in the outline scrolls any time you scroll or click in the source panel. For a big module, that update feels real slow (It's extra slow in the video because I turned on highlight component updates). ~Also focusing doesn't work in the panel at all.~

https://user-images.githubusercontent.com/478109/142503738-3cba405c-1ee5-47b9-be14-2bfa1301cc72.mov

This PR memoizes the components that make up the list of functions, and removes props that would've triggered needless rerenders.

(Again the video is artificially janky from highlight component updates)

https://user-images.githubusercontent.com/478109/142504191-3953bcfe-734e-48a5-98a3-a840f5cfa1b6.mov

Replay: https://app.replay.io/recording/1e307651-ae2d-4ddb-a105-9011addf2b58#

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/Outline.js
@@ -68,12 +68,49 @@ const OutlineFunction = React.memo(function OutlineFunction({ isFocused, func, o
     <li
       className={classnames("outline-list__element", { focused: isFocused })}
       ref={itemRef}
-      onClick={() => onSelect(func)}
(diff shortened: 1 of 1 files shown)


colbyr:
Replay incoming: _[insert replay here]_

jasonLaster (approved):
Solid and I like that you started with a simple optimization.

PR #4572, opened 2021-11-18
Move the result count to a separate row

jaril:
https://user-images.githubusercontent.com/15959269/142504936-a35588b6-bf68-45eb-b0f9-6b1a86af566c.mov


colbyr (approved):
I know this was motivated by an annoying flex thing, but i kind like this better 😅

jcmorrow:
I think if we're gonna have it on the next row it might make more sense to add some left-padding to line it up with the expand/collapse arrows?

PR #4582, opened 2021-11-19
Fix search result counts

colbyr:
Fixes https://github.com/RecordReplay/devtools/issues/4577
Also let's the result container scroll.

### After
Screen Shot 2021-11-19 at 11 06 22 AM

Replay: https://app.replay.io/recording/c7bb4069-bed4-4646-ad3b-f161cf99a913

### DIFF
--- a/src/devtools/client/debugger/src/components/ProjectSearch.js
+++ b/src/devtools/client/debugger/src/components/ProjectSearch.js
@@ -274,7 +274,7 @@ export class ProjectSearch extends Component {
       <div className="px-2">
         {status === "DONE" ? (
           <div className="whitespace-pre pl-2">
-            {totalMatches} result{totalMatches === 1 ? "" : "s"}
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/ProjectSearch.js
```diff
@@ -327,6 +334,7 @@ export class ProjectSearch extends Component {
   }
 
   render() {
+    console.info(this.state);
```
jasonLaster, line 337: 🧂 not needed

jasonLaster (approved):
Looks good. I like the clean up!

src/devtools/client/debugger/src/components/ProjectSearch.js
colbyr: lol whoops good catch

PR #4583, opened 2021-11-19
Fix the keyboard shortcut for opening file search

jaril:
This gets us a natural Cmd+Shift+F keyboard shortcut to pop open full text search, regardless of which view (devtools/viewer) you're in.

### DIFF
--- a/src/devtools/client/debugger/src/components/ProjectSearch.js
+++ b/src/devtools/client/debugger/src/components/ProjectSearch.js
@@ -92,15 +92,15 @@ export class ProjectSearch extends Component {
   }
 
   componentDidMount() {
-    const { globalShortcuts, shortcuts } = this.context;
-    globalShortcuts.on("CmdOrCtrl+Shift+F", this.toggleProjectTextSearch);
-    shortcuts.on("Enter", this.onEnterPress);
+    const { shortcuts } = this.context;
+    // Todo: Fix the way we handle keyboard events here. -jvv
+    shortcuts?.on("Enter", this.onEnterPress);
   }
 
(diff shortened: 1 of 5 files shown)


jasonLaster:
One quick bug, but otherwise good.

CMD+Shift+F in non-dev should open devtools

jaril:
> One quick bug, but otherwise good.

Good catch — fixing.

src/ui/components/KeyboardShortcuts.tsx
```diff
@@ -15,11 +15,19 @@ function setupShortcuts() {
 
 const globalShortcuts = setupShortcuts();
 
-function KeyboardShortcuts({ setSelectedPrimaryPanel }: PropsFromRedux) {
+function KeyboardShortcuts({ viewMode, setSelectedPrimaryPanel, setViewMode }: PropsFromRedux) {
+  const openFileSearch = () => {
```
jasonLaster, line 19: i think we should refer to this as `FullTextSearch` or ProjectSearch

src/devtools/client/debugger/src/components/ProjectSearch.js
```diff
@@ -199,7 +199,7 @@ export class ProjectSearch extends Component {
   };
 
   onEnterPress = () => {
-    if (!this.isProjectSearchEnabled() || !this.state.focusedItem || this.state.inputFocused) {
+    if (!this.state.focusedItem || this.state.inputFocused) {
```
jasonLaster, line 202: nice

jasonLaster (approved):

PR #4584, opened 2021-11-19
Append a narrow classname to the editor pane when the width is <240px

### LINKED ISSUE #4574: Default editor state revisions
Issue one:
image

This shouldn't crop awkwardly when the window gets smaller. Jaril is going to help with a classname for this.

Issue two:
image

Oops, also this doesn't look right in Safari!

jaril:
Fix #4574.

This adds a classname that we can use as a breakpoint for when the editor pane gets too narrow, so that the suggested actions aren't cut off.

### DIFF
--- a/src/ui/utils/useWidthObserver.ts
+++ b/src/ui/utils/useWidthObserver.ts
@@ -1,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+var resizeObserver;
+
+// Set the element to be observed in the component's state, and pass it into this
+// hook so that it returns the width of the element whenever it's resized.
+export default function useWidthObserver(node: HTMLElement) {
+  const [width, setWidth] = useState<number | null>(null);
+
+  useEffect(() => {
+    if (!node) return;
+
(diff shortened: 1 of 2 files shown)


jaril:
It still requires the narrow version of the design. @jonbell-lot23

src/ui/utils/useWidthObserver.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+var resizeObserver;
+
+// Set the element to be observed in the component's state, and pass it into this
+// hook so that it returns the width of the element whenever it's resized.
+export default function useWidthObserver(node: HTMLElement) {
+  const [width, setWidth] = useState<number | null>(null);
```
jasonLaster, line 8: probably want to use a `useRef` hook so we can have more of these

jasonLaster (approved):
Feel free to land with the fix

jonbell-lot23:
Squashing and merging so I can more easily deal with the narrow design (later)

PR #4593, opened 2021-11-20
Outline: Only scroll when not visible

jasonLaster:
1. fix scrolling with isVisible
2. fix scrolling with css so that the input is on top and list is correct
3. style the look in feel

Tests
1. clicking an item in the list should not jump
2. clicking a function in the editor should scroll
3. filtering in the filter

https://app.replay.io/recording/f2e6a591-c555-42dc-b17d-4e26526eb768

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/OutlineFilter.js
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/OutlineFilter.js
@@ -35,7 +35,7 @@ export default class OutlineFilter extends Component {
       <div className="outline-filter pt-1 px-3">
         <form>
           <input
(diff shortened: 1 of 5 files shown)


src/devtools/client/debugger/src/components/PrimaryPanes/OutlineFilter.js
```diff
@@ -35,7 +35,9 @@ export default class OutlineFilter extends Component {
       <div className="outline-filter pt-1 px-3">
         <form>
           <input
-            className={"w-full h-full text-xs px-2 py-1 rounded-md"}
+            className={
```
jaril, line 38: We can remove the brackets here

jaril (approved):

PR #4612, opened 2021-11-21
Add text explaining the error when we need the user to sign in

### LINKED ISSUE #4611: Add some context for when a user needs to log in before looking at a private recording
When an anonymous user opens a private Replay, we ask them to sign in so we know if they're allowed to look at that Replay (or not). 

The current error screen doesn't really explain that. There's a sign in button so that's a good enough prompt but I'm pretty sure we had some text that used to be there before. Let's make sure that's back.

image

jaril:
Fix #4611.

image

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -97,8 +97,8 @@ function getRecordingNotAccessibleError(
   }
 
   return {
-    message: "",
-    content: "",
+    message: "Almost there",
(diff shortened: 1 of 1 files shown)


jonbell-lot23 (approved):
This works. I think I'd go with:

Almost There

This is a private replay. Please sign in.

[Sign in to Replay]

PR #4634, opened 2021-11-22
Accordion: dont render component when collapsed

jasonLaster:
This will prevent us from rendering the accordion content when collapsed. This is important because the outline component can be expensive

### DIFF
--- a/src/ui/components/Accordion.tsx
+++ b/src/ui/components/Accordion.tsx
@@ -51,7 +51,7 @@ const Accordion = ({ items }: { items: AccordionItem[] }) => {
               </h2>
             </div>
             <div className={classNames(styles.content, { [styles.open]: !collapsed })}>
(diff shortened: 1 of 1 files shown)


src/ui/components/Accordion.tsx
```diff
@@ -51,7 +51,7 @@ const Accordion = ({ items }: { items: AccordionItem[] }) => {
               </h2>
             </div>
             <div className={classNames(styles.content, { [styles.open]: !collapsed })}>
-              {component}
+              {!collapsed && component}
```
jcmorrow, line 54: One bummer about this is that state won't be maintained across component minimizing/maximizing. Maybe when we've improved the performance of that component we can turn this back on.

jcmorrow (approved):

jaril (approved):

PR #4639, opened 2021-11-22
Host our own fonts

### LINKED ISSUE #3814: Host our fonts
https://twitter.com/rsms/status/1444330242353479682

The official inter font is two years old. We probably will have better perf if we host it ourselves too.

some additional resources
* https://github.com/vercel/next.js/issues/512
* https://github.com/vercel/next.js/discussions/14581
* https://blog.cloudflare.com/fast-google-fonts-with-cloudflare-workers/
* https://google-webfonts-helper.herokuapp.com/fonts/

jcmorrow:
### Changes

- Add the font assets for Inter and the Material Icon set to `/public`
- Load the font-face declarations for those in `_app.tsx`
- Delete everything relating to font-loading, fontface observer, etc. All of that should "just work" now. I've looked, and fonts still aren't loaded in E2E tests, but the tests pass anyways. I'm not digging any further in there.

Closes #3814


jasonLaster:
Looks great on the preview branch. Ship it!

jasonLaster (approved):

jcmorrow:
I left this in the commit, I'll also add it here for posterity:

>  I'm experiencing a brief flash of unstyled sans-serif before Inter loads on the Library page. I have *no freaking clue* why this is happening. I must have read the first 3 pages of search results about next.js and font-loading, and I still can't figure it out. It also *never* happens on production, which is currently using Google Fonts, and again I have no idea why. It makes no sense that loading from Google Fonts would be any different than loading the exact same thing from our own server. If anything, it should be less-prone to timing issues like this (I would think). Next.js says that they do some js-build-time-optimization of fonts loaded from Google Fonts, but I don't see any evidence of that happening, so I don't think that's it. Anyways, this solves the "fonts not loading sometimes" issue, so that's good enough for now.

PR #4642, opened 2021-11-23
Host our own fonts

### LINKED ISSUE #3814: Host our fonts

jcmorrow:
- Add the font assets for Inter and the Material Icon set to `/public`
- Load the font-face declarations for those in `_app.tsx`
- Delete everything relating to font-loading, fontface observer, etc.
  All of that should "just work" now. I've looked, and fonts still
  aren't loaded in E2E tests, but the tests pass anyways. I'm not
  digging any further in there.

I'm experiencing a brief flash of unstyled sans-serif before Inter loads
on the Library page. I have *no freaking clue* why this is happening. I
must have read the first 3 pages of search results about next.js and
font-loading, and I still can't figure it out. It also *never* happens
on production, which is currently using Google Fonts, and again I have
no idea why. It makes no sense that loading from Google Fonts would be
any different than loading the exact same thing from our own server. If
anything, it should be less-prone to timing issues like this (I would
think). Next.js says that they do some js-build-time-optimization of
fonts loaded from Google Fonts, but I don't see any evidence of that
happening, so I don't think that's it. Anyways, this solves the "fonts
not loading sometimes" issue, so that's good enough for now.

Closes #3814


public/fonts/material_icons/material_icons.css
```diff
@@ -0,0 +1,66 @@
...
+  text-rendering: optimizeLegibility;
+
+  /* Support for Firefox. */
+  -moz-osx-font-smoothing: grayscale;
+
+  /* Support for IE. */
+  font-feature-settings: "liga";
+}
+
+.material-icons-outlined {
```
jcmorrow, line 43: This class was missing last time.

public/fonts/material_icons/material_icons.css
```diff
@@ -0,0 +1,49 @@
...
+@font-face {
+  font-family: "Material Icons Outlined";
+  font-style: normal;
+  font-weight: 400;
+  src: url(/fonts/material_icons/MaterialIconsOutlined-Regular.woff2) format("woff2"),
+    url(/fonts/material_icons/MaterialIconsOutlined-Regular.otf) format("opentype");
+}
+
+.material-icons,
+.material-icons-outlined {
```
jcmorrow, line 19: Was missing this class last time!

abstractalgo:
> I'm experiencing a brief flash of unstyled sans-serif before Inter loads on the Library page. I have _no freaking clue_ why this is happening. I must have read the first 3 pages of search results about next.js and font-loading, and I still can't figure it out. It also _never_ happens on production, which is currently using Google Fonts, and again I have no idea why. It makes no sense that loading from Google Fonts would be any different than loading the exact same thing from our own server.

It was due to `font-display: swap` that we added, which doesn't exist in styles loaded from Google directly (which default to `font-display: block`, so there is no ugly font flashing. (explained [here](https://github.com/RecordReplay/devtools/pull/5484#issue-1150498157))

PR #4651, opened 2021-11-23
Add message of the day to new tab page

ryanjduffy:
* Adds message of the day from the current user's organization (if they are part of one) to the new tab page

image

I placed it above the logo so it wouldn't be lost with the existing copy. Some proper visual design here is welcome.


src/ui/hooks/users.ts
```diff
@@ -34,6 +34,7 @@ export function useUserIsAuthor() {
 }
 
 export type UserInfo = {
+  motd: string | null;
```
jasonLaster, line 37: surprised this is a user property and not a workspace... 

Is the idea that if they are a member of an org they get this?

jasonLaster: I would expect the client to be smart enough to fetch workspaces and check for an org and motd

ryanjduffy: This may be part of the broader discussion we need to have. Without asserting that a given user will always only be part of one org (meaning in this case, one team with a message of the day), we can't know which workspace they're targeting when starting a process (opening a new tab, starting a recording, etc).

My naive approach here was to just pick one on the backend (it could be the most recent, the current default workspace's, the first we find ...) and return that for the user rather than trying to derive it on the frontend.

That said, we could also expose it for the workspace so we can include it in the UI for that workspace somewhere too but for the purposes of the new tab page, I didn't want to implement all that logic on the frontend.

ryanjduffy: We've updated the backend to complain when a user is part of more than one org so there isn't any "choosing" to be done. This is still being exposed as part of the user API for now but it might move in the future.

PR #4652, opened 2021-11-23
Make sure we only fire events when hovering in the editor

### LINKED ISSUE #4637: TypeError: Cannot read properties of null (reading 'querySelector')
Sentry Issue: [DEVTOOLS-160](https://sentry.io/organizations/replay/issues/2808037129/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'querySelector')
  at getLineNumberNode (./src/devtools/client/debugger/src/utils/editor/gutter-events.js:25:37)
  at isValidTarget (./src/devtools/client/debugger/src/utils/editor/gutter-events.js:31:26)
  at apply (./src/devtools/client/debugger/src/utils/editor/gutter-events.js:45:10)
```

jaril:
Demo: https://app.replay.io/recording/a306fc77-4ba4-4cae-be6d-3054af78c3b6

`gutter-events` could still use a refactor but this fixes #4637

### DIFF
--- a/src/devtools/client/debugger/src/utils/editor/gutter-events.js
+++ b/src/devtools/client/debugger/src/utils/editor/gutter-events.js
@@ -20,22 +20,18 @@ function getLineNumberNode(target) {
     target.closest(".CodeMirror-gutter-wrapper")?.parentElement ||
     target.closest(".CodeMirror-line")?.parentElement;
 
-  if (!row) return null;
-
-  const wrapper = row.querySelector(".CodeMirror-gutter-wrapper");
-
-  return wrapper.querySelector(".CodeMirror-linenumber");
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/utils/editor/gutter-events.js
```diff
@@ -20,22 +20,18 @@ function getLineNumberNode(target) {
     target.closest(".CodeMirror-gutter-wrapper")?.parentElement ||
     target.closest(".CodeMirror-line")?.parentElement;
 
-  if (!row) return null;
```
jaril, line 23: We don't need this check anymore since we do the same check in `isValidTarget`

PR #4658, opened 2021-11-23
Make the experimental user settings pane consistent with the notification preferences

### LINKED ISSUE #4539: Scrollbar is overlapping checkboxes
image

jaril:
Fix #4539.

image

This keeps two tabs (experiments, preferences) with checkboxes to toggle look consistent. 

Note that this decouples our checkbox rows from `SettingsModal` component. Instead of passing in a list of items, this makes it so that we're forced pass in a component instead. This way, we have the flexibility to compose that component in whichever way we want.


jasonLaster:
@jaril Looks good, but we should probably continue including the secondary text as well which is nice for folks to know what they're enabling. Happy to design it first in figma if the spacing / secondary text grey is tricky.

jaril:
> @jaril Looks good, but we should probably continue including the secondary text as well which is nice for folks to know what they're enabling. Happy to design it first in figma if the spacing / secondary text grey is tricky.

Not too tricky. Took a stab at it:

image

ryanjduffy (approved):
Good idea. The abstraction was okay for a while but it has run its course. :)

I noticed some alignment issues with the checkbox that weren't in your screenshot. Worth checking out.

image

jaril:
@ryanjduffy Good catch — fixed! image

jasonLaster (approved):
Looks good

ryanjduffy:
This can be a follow up if you'd like but the layout is still a bit off on the preview branch in chrome. I updated the first row to use grid layout and that seems to work well.

image

image

jaril:
Ah thanks for checking the preview branch — looks fine on local but seems to be showing differently on the preview. I'll dig into it later but for now let's ship with the grid implementation!

jaril:
image

PR #4667, opened 2021-11-24
Add a General settings panel which includes a lot of org options

jcmorrow:
This is decently on the way to lining up with the designs in https://github.com/RecordReplay/devtools/issues/4575. It's not actually hooked up to the backend yet though 😅 

CleanShot 2021-11-23 at 20 21 18

### DIFF
--- a/src/ui/components/shared/SettingsModal/SettingsBody.tsx
+++ b/src/ui/components/shared/SettingsModal/SettingsBody.tsx
@@ -19,7 +19,11 @@ function SettingsBodyWrapper({ children }: { children: (React.ReactChild | null)
 }
 
 export function SettingsHeader({ children }: { children: React.ReactChild }) {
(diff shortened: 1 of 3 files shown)


jasonLaster (commented):
It might take a bit of finagling to conditionally add the menu item, but then we can land this darkly with `app.features.generalSetting`

jcmorrow:
> It might take a bit of finagling to conditionally add the menu item, but then we can land this darkly with `app.features.generalSetting`

True! Maybe a good plan would be to implement the Name and Logo fields (which can stay on for everyone) first, and feature-flag the org stuff until the rest of the backend changes are in?

jasonLaster:
Yep! I like it

jasonLaster:
Is this still relevant. It looks like it should be.

jcmorrow:
Done in https://github.com/RecordReplay/devtools/pull/4704

PR #4668, opened 2021-11-24
Allow users to jump from the event transcript to a line of code

jaril:


jaril:
https://github.com/RecordReplay/devtools/pull/4713

PR #4672, opened 2021-11-24
Removed jellyfish and reduced font size

jonbell-lot23:
Old:
image

New:
image

Notes:

1. We shouldn't show broken images for favicons. Can we confirm that these aren't returning a 404 before showing them? We can do this in another ticket

2. This screen is heavy on tailwind, whereas we should have aligned more closely to our other modals.

### DIFF
--- a/src/ui/components/UploadScreen/Privacy.tsx
+++ b/src/ui/components/UploadScreen/Privacy.tsx
@@ -86,7 +86,7 @@ export function Privacy() {
   const { operations } = recording!;
 
   return (
-    <div className="flex w-full h-full rounded-xl shadow-xl text-lg relative p-8 bg-jellyfish overflow-hidden">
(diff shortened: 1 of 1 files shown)


src/ui/components/UploadScreen/Privacy.tsx
```diff
@@ -86,7 +86,7 @@ export function Privacy() {
   const { operations } = recording!;
 
   return (
-    <div className="flex w-full h-full rounded-xl shadow-xl text-lg relative p-8 bg-jellyfish overflow-hidden">
+    <div className="flex w-full h-full rounded-xl shadow-xl text-md relative p-8 bg-white overflow-hidden">
```
jasonLaster, line 89: this should be a conditional so it is a jelly fish in the upload screen and white in devtools

jonbell-lot23: Aw crap. Thanks.

PR #4673, opened 2021-11-24
Outline: refactor outline view

jasonLaster:
* move outline view to its own folder
* extract functional components
* simplify class component


jaril:
Few things to close before we merge this:

- [ ] https://github.com/RecordReplay/devtools/issues/4700
- [ ] Classes aren't being displayed at all

PR #4678, opened 2021-11-24
Defensively check local nags

jcmorrow:
I can't tell if this is happening in prod, but it happened in this E2E
test: https://app.replay.io/recording/9febd9f5-11b4-477a-8af8-24250b22c259?point=29855706938421465504100931554249690&time=61856&hasFrames=false


jcmorrow:
@jaril I can't tell if this case should return `true` or `false`. I'll leave this PR in your hands!

jaril:
Continued in #4695

PR #4697, opened 2021-11-29
New breakpoint visual

jaril:


jasonLaster:
I'd love to pick this up again and see if we can get it across the finish line

PR #4699, opened 2021-11-29
Fix the breakpoint panel's empty state

### LINKED ISSUE #4614: Breakpoint empty state touch-up
[Figma link](https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/?node-id=9051%3A79412)

We don't want to put a hat here, because it's doubling up with our other Mad Hatter work. But we can do something subtle like this.

**If editor is open**
image

**If editor isn't open**
image

jaril:
Fix #4614.

image
image

Changed this up a bit from the original issue:
- Copied the styling from comments.
- Changed the copy so that it's more concrete and accurate.


jonbell-lot23 (approved):
We'll probably want to change the icon, but I think this is good to go. The text is much better the way you did it, thanks.

PR #4709, opened 2021-11-30
Fix react framework events

jasonLaster:
This gets react framework events to work at the expense of older versions of React. We should try to support both versions of react and add an end to end test for them.


jaril:
Demo for react 16: https://app.replay.io/recording/921c27fa-529b-42fd-9c0d-eb4ef11a0ceb

jaril:
Simplified in #4724.

PR #4716, opened 2021-11-30
Add more trackers for mixpanel

jaril:
This adds the following trackers:
- login.first_login
- session.devtools_start


src/ui/hooks/recordings.ts
```diff
@@ -108,27 +108,6 @@ export function useGetRecordingId() {
   return Array.isArray(id) ? id[0] : id!;
 }
 
-export async function getRecordingOwnerUserId(
```
jaril, line 111: These weren't being used.

PR #4704, opened 2021-11-29
Add Org-specific settings to Team Settings Modal

ryanjduffy:

PR #4718, opened 2021-12-01
Disable making recordings public when disabled by org

ryanjduffy:
Blocked by #4704 

## Issue

Orgs can disable making their replays public and we should suppress the UI in these cases.

## Resolution

I've removed the UI from the sharing dialog and the library recording menu but have disabled it in the upload screen. The latter seemed awkward to remove since it is only relevant for workspaces that are orgs with that setting enabled so we'd be hiding and showing it based on the selected org.


src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -36,11 +51,17 @@ function EditableSettings({
         <TeamSelect {...{ workspaces, handleWorkspaceSelect, selectedWorkspaceId }} />
       ) : null}
       <div
-        className="space-x-2 select-none flex flex-row items-center justify-between w-full border border-textFieldBorder rounded-md shadow-sm px-2.5 py-1.5 text-left cursor-default focus:outline-none focus:ring-1 focus:ring-primaryAccent focus:border-primaryAccentHover bg-jellyfish"
-        onClick={() => setIsPublic(!isPublic)}
+        className={classNames(
+          publicDisabled ? "opacity-60" : undefined,
```
jasonLaster, line 55: i like your logic with disabling if the user has multiple teams. if the has one org, then it's preferred to hide it.

src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    ref.current = setTimeout(() => {
+      updateWorkspaceSettings({
+        variables: {
+          workspaceId,
+          motd: message,
+        },
+      });
+    }, 500);
+  }, [message]);
```
jasonLaster, line 41: why the timeout?

jasonLaster (commented):
Will test and review tomorrow

src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
ryanjduffy: Debouncing the update since it fires on every value change

jasonLaster: got it.

ryanjduffy:
I added a few storybook stories for the sharing block from the upload screen. It's only sort of useful but trying to keep moving storybook forward a bit.

src/ui/components/UploadScreen/Sharing.tsx
ryanjduffy: I took a slightly different approach for this. We could totally do precisely this but it requires a bit more clean up (e.g. if you remove the public toggle, the team select should go full width but the dropdown is configured for a fixed width). 

Instead, I figure the 80% case for orgs is that a user is only part of 1 workspace (the org) and both the library and public recordings are disabled so we can skip the edit button altogether and just report that the recording will land in the org and be private.

src/ui/components/UploadScreen/Sharing.tsx
```diff
@@ -30,17 +31,23 @@ function EditableSettings({
     updateDefaultWorkspace({ variables: { workspaceId: dbWorkspaceId } });
   };
 
+  const workspace = workspaces.find(w => w.id === selectedWorkspaceId);
+  const disabled = workspace?.settings.features.recording.public === false;
+
```
jasonLaster: id also hide this if its the only workspace... but that can be a follow up

jasonLaster (approved):

PR #4722, opened 2021-12-01
Show the unwrapped react events in the events filter/dropdown count

jaril:
_super_ rough.

### DIFF
--- a/src/devtools/client/debugger/src/actions/event-listeners.js
+++ b/src/devtools/client/debugger/src/actions/event-listeners.js
@@ -9,6 +9,7 @@ import remove from "lodash/remove";
 import difference from "lodash/difference";
 
 const { getAvailableEventBreakpoints } = require("devtools/server/actors/utils/event-breakpoints");
+import { unwrapFrameworkPoints } from "protocol/event-listeners";
 import * as selectors from "../selectors";
 
 import { features } from "ui/utils/prefs";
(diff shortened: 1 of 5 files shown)


src/devtools/client/debugger/src/actions/event-listeners.js
```diff
@@ -103,8 +104,22 @@ export function getEventListenerBreakpointTypes() {
     dispatch({ type: "RECEIVE_EVENT_LISTENER_TYPES", categories });
 
     if (features.eventCount) {
-      const eventTypePoints = await client.fetchEventTypePoints(INITIAL_EVENT_BREAKPOINTS);
+      const { eventTypePoints, eventTypeResults } = await client.fetchEventTypePoints(
+        INITIAL_EVENT_BREAKPOINTS
+      );
+      console.log({ eventTypePoints, eventTypeResults });
       dispatch({ type: "RECEIVE_EVENT_LISTENER_POINTS", eventTypePoints });
```
jaril, line 111: This is where we stopped before—we just added the `eventTypePoints` to Redux. That point data was then used in two places: 1) the hit count beside an event in the event dropdown, and 2) to match an event in the event panel (sidebar) to a point with a source.

src/devtools/client/debugger/src/actions/event-listeners.js
```diff
@@ -103,8 +104,22 @@ export function getEventListenerBreakpointTypes() {
...
+
+      const unwrappedFrameworkPoints = await unwrapFrameworkPoints(eventTypeResults);
+
+      if (unwrappedFrameworkPoints) {
```
jaril, line 115: Implementation can definitely be improved here, but essentially, this checks if there's any react events to be "unwrapped" in the points that we just got. If so, it unwraps them to get the real event from the react event, then adds it to the other `eventTypePoints` from before. That means you end end up with both the react event and its corresponding unwrapped react event in the store under `eventTypePoints[eventNameHere]`.

src/devtools/client/debugger/src/reducers/event-listeners.js
```diff
@@ -34,6 +34,18 @@ function update(state = initialEventListenerState(), action) {
         eventTypePoints: { ...state.eventTypePoints, ...action.eventTypePoints },
       };
 
+    case "RECEIVE_FRAMEWORK_POINTS":
+      const newTypePoints = { ...state.eventTypePoints };
+
+      Object.entries(action.eventTypeFrameworkPoints).forEach(([key, events]) => {
```
jaril, line 40: Yikes.

src/protocol/logpoint.ts
```diff
@@ -352,31 +382,52 @@ export function setLogpointByURL(
...
+          },
+          onAnalysisResult: result => {
+            console.log("onAnalysisResult", { result });
+
+            const logGroupId = newLogGroupId();
+            eventTypeLogGroupId[eventType] = logGroupId;
+
+            console.log(">>>start waiting");
+            collectedResults.push(...result);
+            // const rv = await getUnwrappedFrameworkPoint(logGroupId, result);
```
jaril, line 417: If there is a way to simplify this PR, it would probably have something to do with changing this `onAnalysisResult`. We could just immediately unwrap the result ("event") here and do something with it, instead of having to handle that separately within the action (`getEventListenerBreakpointTypes`)

jasonLaster:
lets drop this... it feels a bit too unwieldy

jasonLaster:
closing as this got unnecessarily complex

PR #4725, opened 2021-12-01
Add special behavior for the default sidepanel open when a replay is opened

### LINKED ISSUE #4687: Default to Replay Info unless there are comments
**Description**
Today we default to the comments pane, which is usually blank. We're going to default to Replay Info ... unless there's a comment.

**Scenarios**
Author of Replay loads Replay for first time: default to Replay info
Author of Replay loads Replay every other time: default to last place they were

Someone loads Replay for the first time and there's no comment: default to Replay info
Someone loads Replay for the first and there is a comment: default to Comment
Someone loads Replay every other time: default to the last place they were

jaril:
Fix #4687.

Putting this on hold, it's trickier than I first thought it would be.

### DIFF
--- a/src/ui/graphql/recordings.ts
+++ b/src/ui/graphql/recordings.ts
@@ -24,6 +24,9 @@ export const GET_RECORDING = gql`
       ownerNeedsInvite
       userRole
       operations
+      comments {
+        id
+      }
       owner {
         id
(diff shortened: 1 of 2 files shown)


jaril:
Continued in #4886.

PR #4732, opened 2021-12-02
Add a console loading bar whenever there's a logpoint analysis in flight

### LINKED ISSUE #4365: Console loading state
**Description:**

We do not currently show when the console is loading. This happens at the beginning while the logs are being fetched and then streaming in.

jaril:
Fix #4365.

Replay: https://app.replay.io/recording/cc9f03b4-fc7b-4da7-8e2c-0da29191777d

Details:
- This skips doing anything for primitive logpoints, because we consider those to return immediately.
- Currently only works for adding logpoints, we should consider event breakpoints separately as a follow up.
- There's no short circuiting mechanism for a logpoint analysis, so if a user removes the breakpoint while the analysis is running, this continues showing the progress bar. Should probably be handled before shipping. _(edit: fixed)_

https://user-images.githubusercontent.com/15959269/144466062-81ffe713-6954-460f-854c-3a98e48d7b42.mov

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/modify.js
+++ b/src/devtools/client/debugger/src/actions/breakpoints/modify.js
@@ -26,6 +26,7 @@ import { setSkipPausing } from "../pause/skipPausing";
 import { comparePosition } from "../../utils/location";
 import { getTextAtPosition } from "../../utils/source";
 import { trackEvent } from "ui/utils/telemetry";
+import { removeLogGroupId } from "devtools/client/webconsole/actions/messages";
 
 // This file has the primitive operations used to modify individual breakpoints
 // and keep them in sync with the breakpoints installed on server threads. These
@@ -137,9 +138,21 @@ export function addBreakpoint(
     if (disabled) {
       // If we just clobbered an enabled breakpoint with a disabled one, we need
       // to remove any installed breakpoint in the server.
-      await client.removeBreakpoint(location);
-    } else {
-      await client.setBreakpoint(breakpoint.location, breakpoint.options);
+      return await client.removeBreakpoint(location);
+    }
+
+    const [oldLogGroupId, logGroupId] = await client.setBreakpoint(
+      breakpoint.location,
+      breakpoint.options
+    );
+
+    // Keep a reference to the analysis's log group id in redux. That way, we can reference
+    // and clear it if we happen to remove the breakpoint before the analysis is finished.
+    dispatch(setBreakpointOptions(cx, location, { ...options, logGroupId }));
+
+    // This is defensive just to make sure we always clear unneeded logGroupIds.
+    if (oldLogGroupId !== logGroupId) {
+      dispatch(removeLogGroupId(oldLogGroupId));
     }
   };
 }
@@ -188,6 +201,10 @@ export function removeBreakpoint(cx, initialBreakpoint) {
     // If the breakpoint is disabled then it is not installed in the server.
     if (!breakpoint.disabled) {
       await client.removeBreakpoint(breakpoint.location);
+    }
+
+    if (breakpoint.options.logGroupId) {
+      dispatch(removeLogGroupId);
     }
   };
 }
--- a/src/devtools/client/webconsole/actions/messages.js
+++ b/src/devtools/client/webconsole/actions/messages.js
@@ -34,7 +34,11 @@ export function setupMessages(store) {
     store.dispatch(onLogpointLoading(logGroupId, point, time, location));
   LogpointHandlers.onResult = (logGroupId, point, time, location, pause, values) =>
     store.dispatch(onLogpointResult(logGroupId, point, time, location, pause, values));
-  LogpointHandlers.clearLogpoint = logGroupId => store.dispatch(messagesClearLogpoint(logGroupId));
+  LogpointHandlers.clearLogpoint = logGroupId => {
+    store.dispatch(removeLogGroupId(logGroupId));
+    store.dispatch(messagesClearLogpoint(logGroupId));
+  };
+  LogpointHandlers.onError = logGroupId => store.dispatch(removeLogGroupId(logGroupId));
 
   ThreadFront.findConsoleMessages((_, msg) => store.dispatch(onConsoleMessage(msg)));
 }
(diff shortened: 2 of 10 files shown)


jaril:
> * There's no short circuiting mechanism for a logpoint analysis, so if a user removes the breakpoint while the analysis is running, this continues showing the progress bar. Should probably be handled before shipping.

Fixed this by keeping a record of the `logGroupId` in the  breakpoint object in store so that we can clear it if the user removes a breakpoint.

src/devtools/client/webconsole/components/Output/ConsoleLoadingBar.tsx
```diff
@@ -0,0 +1,20 @@
...
+import LoadingProgressBar from "ui/components/shared/LoadingProgressBar";
+import { UIState } from "ui/state";
+import { hasPendingLogGroupIds } from "../../selectors/messages";
+
+function ConsoleLoadingBar({ hasPendingLogGroupIds }: PropsFromRedux) {
+  if (!hasPendingLogGroupIds) {
+    return null;
+  }
+
+  return <LoadingProgressBar />;
```
jasonLaster, line 12: Nice - does this push the UI down? Can we make it absolute?

src/devtools/client/webconsole/reducers/messages.js
```diff
@@ -172,6 +175,23 @@ function messages(state = MessageState(), action) {
...
+      const newPendingLogGroupIds = new Set(state.pendingLogGroupIds);
+      newPendingLogGroupIds.delete(action.logGroupId);
+
+      return {
+        ...state,
+        pendingLogGroupIds: newPendingLogGroupIds,
+      };
+    }
+
+    case "ADD_LOG_GROUP_ID": {
```
jasonLaster, line 188: perhaps `ADD_PENDING_LOG_GROUP_ID`

src/devtools/client/webconsole/actions/messages.js
```diff
@@ -34,7 +34,11 @@ export function setupMessages(store) {
     store.dispatch(onLogpointLoading(logGroupId, point, time, location));
   LogpointHandlers.onResult = (logGroupId, point, time, location, pause, values) =>
     store.dispatch(onLogpointResult(logGroupId, point, time, location, pause, values));
-  LogpointHandlers.clearLogpoint = logGroupId => store.dispatch(messagesClearLogpoint(logGroupId));
+  LogpointHandlers.clearLogpoint = logGroupId => {
+    store.dispatch(removeLogGroupId(logGroupId));
+    store.dispatch(messagesClearLogpoint(logGroupId));
```
jasonLaster, line 39: minor point, but it's okay to also listen for this action in the reducer and do the same thing. you might not need: removeLogGroupId

src/devtools/client/debugger/src/actions/breakpoints/modify.js
```diff
@@ -137,9 +138,21 @@ export function addBreakpoint(
...
+      breakpoint.options
+    );
+
+    // Keep a reference to the analysis's log group id in redux. That way, we can reference
+    // and clear it if we happen to remove the breakpoint before the analysis is finished.
+    dispatch(setBreakpointOptions(cx, location, { ...options, logGroupId }));
+
+    // This is defensive just to make sure we always clear unneeded logGroupIds.
+    if (oldLogGroupId !== logGroupId) {
+      dispatch(removeLogGroupId(oldLogGroupId));
```
jasonLaster, line 155: can we do this all in the reducer? feels like this could be simpler. For instance, the console reducer could listen for the `setBreakpointOptions` action and see ooh, we're trying to change an already existing analysis... also - another option is to use a `location-key` which might be simpler than log-group-id.

jasonLaster (commented):
This is complicated stuff :)

jasonLaster:
By the way, there's another use case where we also want to listen for when the initial console messages are loaded

I think this should be sufficient

```diff
diff --git a/src/devtools/client/webconsole/actions/messages.js b/src/devtools/client/webconsole/actions/messages.js
index fa3705e3..c8ef5682 100644
--- a/src/devtools/client/webconsole/actions/messages.js
+++ b/src/devtools/client/webconsole/actions/messages.js
@@ -29,14 +29,17 @@ const defaultIdGenerator = new IdGenerator();
 let queuedMessages = [];
 let throttledDispatchPromise = null;

-export function setupMessages(store) {
+export async function setupMessages(store) {
   LogpointHandlers.onPointLoading = (logGroupId, point, time, location) =>
     store.dispatch(onLogpointLoading(logGroupId, point, time, location));
   LogpointHandlers.onResult = (logGroupId, point, time, location, pause, values) =>
     store.dispatch(onLogpointResult(logGroupId, point, time, location, pause, values));
   LogpointHandlers.clearLogpoint = logGroupId => store.dispatch(messagesClearLogpoint(logGroupId));

-  ThreadFront.findConsoleMessages((_, msg) => store.dispatch(onConsoleMessage(msg)));
+  await ThreadFront.findConsoleMessages((_, msg) => store.dispatch(onConsoleMessage(msg)));
+  dispatch({
+    type: "console-messages-loaded",
+  });
 }

 function convertStack(stack, { frames }) {
diff --git a/src/protocol/thread/thread.ts b/src/protocol/thread/thread.ts
index 6e8c998b..fed9ac7f 100644
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -817,11 +817,6 @@ class _ThreadFront {
   async findConsoleMessages(onConsoleMessage: (pause: Pause, message: Message) => void) {
     const sessionId = await this.waitForSession();

-    client.Console.findMessages({}, sessionId).then(({ overflow }) => {
-      if (overflow) {
-        console.warn("Too many console messages, not all will be shown");
-      }
-    });
     client.Console.addNewMessageListener(async ({ message }) => {
       await this.ensureAllSources();
       const pause = new Pause(sessionId);
@@ -843,6 +838,12 @@ class _ThreadFront {
       }
       onConsoleMessage(pause, message);
     });
+
+    return client.Console.findMessages({}, sessionId).then(({ overflow }) => {
+      if (overflow) {
+        console.warn("Too many console messages, not all will be shown");
+      }
+    });
   }
```

PR #4733, opened 2021-12-02
Fix UserAPIKeys and remove values prop from settings

ryanjduffy:
## Issue

API Keys pane in the user settings doesn't render anything

## Analysis

This was the only pane that was relying on the settings being passed down to it via the `values` prop. This prop was dropped in #4658.

## Resolution

* Fetch the settings in UserAPIKeys instead of relying on it being passed down
* Remove the `values` stuff from `SettingsModal` since it wasn't used by anything else

## Notes

I don't love all of these views fetching their own data. It makes it harder to test them in storybook without stubbing out the APIs. But, we've already done it in a bunch of places and the generalization of `SettingsModal` was adding more complexity than value so here we are :)

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/index.tsx
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/index.tsx
@@ -43,7 +43,7 @@ class PrimaryPanes extends Component<PropsFromRedux> {
       onToggle: () => {
         this.props.toggleSourcesCollapse();
       },
-      collapsed: prefs.sourcesCollapsed,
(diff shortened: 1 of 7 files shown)


jaril (approved):

jaril:
Lint problem 👆

ryanjduffy:
There were some type failures I missed + another random one in `PrimaryPanes`

jaril:
> There were some type failures I missed + another random one in `PrimaryPanes`

Yep just merged a fix. Should be clean now!

PR #4743, opened 2021-12-03
Add content security policy headers

ryanjduffy:


ryanjduffy:
We could try turning this on in report-only mode for now as well and monitor for a week first. There's definitely some risk rolling this out as is even though I've tried to exercise the critical flows on the preview branch and in local development.

next.config.js
```diff
@@ -1,5 +1,16 @@
 const { RetryChunkLoadPlugin } = require("webpack-retry-chunk-load-plugin");
 
+const self = "'self' https://*.replay.io wss://*.replay.io";
```
loganfsmyth, line 3: Do we expect to run scripts from other subdomains? If not, I'd code to be specific with `https://app.replay.io` wherever possible.

next.config.js
```diff
@@ -1,5 +1,16 @@
...
+const csp = `
+  frame-ancestors ${self};
+  default-src ${self};
+  style-src ${self} 'unsafe-inline';
```
loganfsmyth, line 7: I think it'd be worth some comments about why we need `unsafe-inline` and `unsafe-eval`.

next.config.js
```diff
@@ -1,5 +1,16 @@
...
+  script-src ${self} 'unsafe-eval' data: blob: https://*.stripe.com https://*.lr-ingest.io https://*.intercom.io https://*.intercomcdn.com https://*.launchdarkly.com;
```
loganfsmyth, line 8: Same here for why `blob:` is needed.

gideonred:
Adding @ejcx to review

next.config.js
```diff
@@ -1,5 +1,16 @@
...
+  connect-src ${self} https://webreplay.us.auth0.com https://*.launchdarkly.com/ https://*.stripe.com https://*.sentry.io https://*.intercom.io wss://*.intercom.io;
```
loganfsmyth, line 9: Have we looked at the docs for these services, or are these based on observations of network traffic? For example https://www.intercom.com/help/en/articles/3894-using-intercom-with-content-security-policy include `https://uploads.intercomusercontent.com`, among others.

I'd guess Sentry and other services have official docs for what they expect.

ryanjduffy: I'll check the docs. There were all based on observations.

ryanjduffy:
@loganfsmyth - I reworked this PR based on the following docs (recording here for posterity) and observations. I've narrowed a few of the rules but the `connect-src` is still pretty broad for our 3rd party services. Some of these are as documented and some are a simplification (in particular the `connect-src`). Thoughts?

https://www.intercom.com/help/en/articles/3894-using-intercom-with-content-security-policy
https://docs.logrocket.com/docs/troubleshooting-sessions#due-to-content-security-policy
https://stripe.com/docs/security/guide#content-security-policy
https://docs.sentry.io/product/security-policy-reporting/
https://rapidsec.com/csp-packages/mixpanel
https://github.com/vercel/next.js/blob/master/examples/with-strict-csp/pages/_document.js

ryanjduffy:
@jcmorrow - would you be up for reviewing this?

jcmorrow:
Yep, I started and then got distracted!

pages/_document.tsx
```diff
@@ -0,0 +1,58 @@
...
+      isDev ? "http://localhost:* http://*.replay.local ws://localhost:* " : ""
+    }`,
+    `frame-src https://js.stripe.com https://hooks.stripe.com`,
+    `script-src 'self' https://widget.intercom.io https://js.intercomcdn.com/ https://cdn.logrocket.io https://cdn.lr-ingest.io https://cdn.lr-in.com https://js.stripe.com ${hash} ${
+      // Required by local dev server
+      isDev ? "'unsafe-eval'" : ""
+    }`,
+
+    // From vercel's CSP config
+    isDev ? `font-src 'self' data:` : "",
```
jcmorrow, line 27: since we're doing a filter boolean how would you feel about doing `isDev && "font-src 'self' data"` for these? I recall maybe you said you prefer ternaries to `&&`s at some point?

jcmorrow (approved):
I'm sure I haven't exercised every possible external provider, but this approach seems good 👍

pages/_document.tsx
ryanjduffy: Yeah, I do. I like that the return value from the expression is explicit and you don't get whatever falsey value happened to be in `isDev`

PR #4746, opened 2021-12-04
Memoize outline view

jasonLaster:
* Memoizes the utils that fetch the function lists
* also limits the list that is shown to 500 global functions and 50 class functions. Using a virtualized list will help

### DIFF
--- a/src/ui/components/Accordion.module.css
+++ b/src/ui/components/Accordion.module.css
@@ -43,4 +43,7 @@
 
 .content.open {
   height: auto;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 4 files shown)


src/ui/components/Accordion.module.css
```diff
@@ -43,4 +43,7 @@
 
 .content.open {
   height: auto;
+  display: flex;
+  flex-direction: column;
+  overflow-y: hidden;
```
jasonLaster, line 48: this is probably the riskiest part, but it helps keep the outline filter input at the top

PR #4750, opened 2021-12-05
Virtualized outline

jasonLaster:
This new virtualized outline view is significantly faster than before and definitely worth finishing.

The three things that would be nice to do before landing are:

1. Figure out how to fix storybook which failed when I connected the component
2. Figure out how to get a ref to the virtualized list for scrolling
3. Remove the old outline view


src/devtools/client/debugger/src/types.ts
```diff
@@ -0,0 +1,30 @@
...
+export type SourceSymbols = {
+  classes: ClassSymbol[];
+  functions: FunctionSymbol[];
+  loading: boolean;
+};
+
+export type Source = {
+  id: string;
+  url: string;
+};
```
jasonLaster, line 30: this type feels especially weak, which makes me sad because we used to have great types for the source objects...

jasonLaster:
This is looking good. I'm ready to ship this with or without storybook

https://www.loom.com/share/4e9634cc3d5e42069d88877f0ca8a36e

jasonLaster:
Okay! Storybook is now working, so we should be good

src/ui/components/shared/LoadingTip.tsx
```diff
@@ -1,7 +1,7 @@
 import classNames from "classnames";
 import React, { useEffect, useRef } from "react";
 import { connect, ConnectedProps } from "react-redux";
-import { actions } from "ui/actions";
+import { setLoadingPageTipIndex } from "ui/actions/app";
```
jasonLaster, line 4: this fixed storybook 🤦  In the future, we could look in webpack's bootstrap function to find where webpack failed

PR #4753, opened 2021-12-05
Network Monitor Improvements

jasonLaster:
* Create a Table HoC
* Move filter bar to the top
* clean up request details
* simplify selecting a row
* extract some components
* Add resource headers
* Add type filtering

https://www.loom.com/share/74c8738cec944b63b74a513ce5bdedcd

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestRow.tsx
+++ b/src/ui/components/NetworkMonitor/RequestRow.tsx
@@ -1,0 +1,84 @@
+import React from "react";
+import styles from "./RequestTable.module.css";
+import classNames from "classnames";
+import { RequestSummary } from "./utils";
+import { Row } from "react-table";
+
+export const RequestRow = ({
+  currentTime,
+  isFirstInFuture,
+  isInPast,
+  onClick,
+  onSeek,
+  prepareRow,
+  row,
+}: {
+  currentTime: number;
+  isFirstInFuture: boolean;
+  isInPast: boolean;
+  onClick: (row: RequestSummary) => void;
+  onSeek: (row: RequestSummary) => void;
+  prepareRow: (row: Row<RequestSummary>) => void;
+  row: Row<RequestSummary>;
+}) => {
+  let currentRow = false;
+  // Did we just pass the row boundary that contains the current time?
+  // If so, let's render this row as the "current" row and all rows
+  // after it as future rows.
+  // if (
+  //   selectedRequest?.id === row.original.id ||
+  //   (!selectedRequest && renderingRequestsInThePast && currentTime <= row.original.point.time)
+  // ) {
+  //   renderingRequestsInThePast = false;
+  //   currentRow = true;
+  // }
+  prepareRow(row);
+  return (
+    <div
+      className={classNames(styles.row, {
+        [styles.current]: isFirstInFuture,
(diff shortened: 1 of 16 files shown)


src/ui/components/NetworkMonitor/RequestRow.tsx
```diff
@@ -0,0 +1,66 @@
...
+  // Did we just pass the row boundary that contains the current time?
+  // If so, let's render this row as the "current" row and all rows
+  // after it as future rows.
+  // if (
+  //   selectedRequest?.id === row.original.id ||
+  //   (!selectedRequest && renderingRequestsInThePast && currentTime <= row.original.point.time)
+  // ) {
+  //   renderingRequestsInThePast = false;
+  //   currentRow = true;
+  // }
```
jasonLaster, line 16: We'll need to refactor this to work in the new world... I think we could probably pass in a currentRowIndex which will be stable and an easy check

jasonLaster:
https://app.replay.io/recording/c6850e7b-a8df-48a2-986b-9862a8a62f45?point=134999718341537425369985633627414178&time=80291&hasFrames=false

src/ui/components/NetworkMonitor/RequestRow.tsx
jcmorrow: I think the struggle with currentRowIndex is when things get filtered.

src/ui/components/NetworkMonitor/TypesDropdown.tsx
```diff
@@ -0,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
jcmorrow, line 3: Are we supposed to be adding this to every new file, or is there some particular criteria for when we do or don't add it?

src/ui/components/NetworkMonitor/utils.ts
```diff
@@ -96,7 +115,27 @@ export const partialRequestsToCompleteSummaries = (
         triggerPoint: r.triggerPoint,
         url: request.event.requestUrl,
       };
+    })
+    .filter(row => {
+      console.log(types);
```
jcmorrow, line 120: ✂️

jcmorrow (approved):
I think this looks ready to go out. The only follow-ups I would do are bringing back the timeline and using more specific material icons, as we discussed.

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -7,31 +7,53 @@ import { getCurrentTime } from "ui/reducers/timeline";
...
+              initialSize="50%"
+              minSize={selectedRequest ? "20%" : "100%"}
+              maxSize={selectedRequest ? "80%" : "100%"}
+              endPanel={selectedRequest ? <RequestDetails request={selectedRequest} /> : <div />}
+              startPanel={
+                <RequestTable
+                  table={table}
+                  data={data}
+                  currentTime={currentTime}
+                  onRowSelect={setSelectedRequest}
```
jcmorrow, line 45: Is there actually a change to the row selection here? This seems like the same to me but I couldn't quite tell.

Issue #4756, opened 2021-12-06
Elements - arrow key navigation should be faster


jasonLaster:
**Description:**

The up / down arrow keys should not scroll unless the next element is out of view. Also, they should be a lot faster. Right now it feels like we're doing something heavy like re-rendering the whole tree

https://app.replay.io/recording/b9353076-ba95-4062-ad46-5b0ae4ce628c

hbenl:
The scrolling is fixed by #4759 but the performance still needs to be improved.

PR #4759, opened 2021-12-06
Elements panel: only scroll when selected node is out of view (#4756)

hbenl:


hbenl:
No, this is just for scrolling.

PR #4761, opened 2021-12-06
Add a toggle and a warning for sourcemaps

### LINKED ISSUE #3916: DevTools should prompt users to setup sourcemaps
[Figma link](https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/?node-id=8783%3A71242)

This replay would be 10x better if the user published sourcemaps with Next.js. We should have a nag that points them in the right direction.

https://app.replay.io/recording/7d36814c-62eb-40a9-a791-f052d7f87879

jaril:
Fix #3916.

Testing targets:
- With sourcemaps: http://localhost:8080/recording/860c85e9-6aa7-4df1-aa91-07f1ddac6a8a
- Without sourcemaps: http://localhost:8080/recording/7d36814c-62eb-40a9-a791-f052d7f87879


src/devtools/client/debugger/src/reducers/sources.js
```diff
@@ -664,4 +665,9 @@ export function getTextAtLocation(state, id, location) {
   return text;
 }
 
+export function getIsSourceMappedSource(state) {
+  const sourceId = getSelectedSourceId(state);
+  return ThreadFront.isSourceMappedSource(sourceId);
```
jasonLaster, line 670: It feels a little dirty to be calling into ThreadFront from a reducer like this. It would be cleaner to have the source have an `isOriginal` field that we could check. e.g. `return source.isOriginal`

src/devtools/client/debugger/src/types.ts
```diff
@@ -23,8 +23,3 @@ export type SourceSymbols = {
   functions: FunctionSymbol[];
   loading: boolean;
 };
-
-export type Source = {
-  id: string;
-  url: string;
-};
```
jasonLaster, line 30: nice. this was dirty

src/ui/components/shared/Modals/SourcemapSetupModal.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <Modal options={{ maskTransparency: "translucent" }} onMaskClick={hideModal}>
+      <Dialog
+        className={classNames("flex flex-col items-center")}
+        style={{ animation: "dropdownFadeIn ease 200ms", width: 400 }}
+      >
+        <DialogLogo />
+        <DialogTitle>Replay is better with sourcemaps</DialogTitle>
+        <DialogDescription>
+          {`We noticed that you have sourcemaps, but they haven't been uploaded to our servers yet.
+          Ready to get started?`}
```
jasonLaster, line 24: ### MSG 1, if they have NextJS

We noticed that you're using NextJS which makes adding sourcemaps super easy. 

https://docs.replay.io/docs/1923e679c1e4411db1bda29536eb1e31#6e444abdd19642af9ddc34766ff84bf2

### MSG 2, if they have a sourceURL

We noticed that you have sourcemaps, but they are not publicly available. Would you like to upload them to Replay?

https://docs.replay.io/docs/1923e679c1e4411db1bda29536eb1e31#912af13ef09a41f4ae774a90796ebbc1

jasonLaster (approved):
couple of comments

src/devtools/client/debugger/src/reducers/sources.js
jaril: 👍

Issue #4773, opened 2021-12-07
Trying to open Network monitor leads to "Unexpected error" and I can no longer open any session with network calls


gideonred:
Replay of the behaviour: https://app.replay.io/recording/1ef6252c-1c6d-4e66-bb23-877b9ab63efc

cc @jcmorrow @loganfsmyth

gideonred:
The only way around this is to open a recording with no network calls and select the console or elements panel. I can then open other Replays

gideonred:
Original Replay: https://app.replay.io/recording/77b285c5-d1e1-4dc0-aeb6-1192f69c7708

gideonred:
![Screen Shot 2021-12-07 at 5 44 57 PM](https://user-images.githubusercontent.com/28845582/145117495-672d768f-b176-4379-94fa-1f544c290e9d.png)


Confirmed it's fixed. Closing

PR #4775, opened 2021-12-07
Revert "Add trackers to the network panel"

loganfsmyth:
Reverts RecordReplay/devtools#4772


loganfsmyth:
This was causing #4773

PR #4776, opened 2021-12-07
Add rudimentary auto-join option

ryanjduffy:
It's not attractive and needs some copy but at least adds the setting so our initial orgs can "set and forget".

https://app.replay.io/recording/80d809cd-7323-486f-8771-9761a790296a#

image

### DIFF
--- a/src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
+++ b/src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
@@ -1,7 +1,7 @@
 import classNames from "classnames";
 import React, { useEffect, useRef, useState } from "react";
 import hooks from "ui/hooks";
-import { WorkspaceSettings } from "ui/types";
+import { PartialWorkspaceSettingsFeatures, WorkspaceSettings } from "ui/types";
 import useDebounceState from "./useDebounceState";
 
 const Label = ({ className, children }: { className?: string; children: React.ReactNode }) => {
@@ -32,7 +32,7 @@ const OrganizationSettings = ({ workspaceId }: { workspaceId: string }) => {
     })
   );
 
-  const updateFeature = (features: Partial<WorkspaceSettings["features"]>) => {
+  const updateFeature = (features: PartialWorkspaceSettingsFeatures) => {
     updateWorkspaceSettings({
       variables: {
         workspaceId,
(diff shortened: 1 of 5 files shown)


src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
```diff
@@ -123,6 +123,35 @@ const OrganizationSettings = ({ workspaceId }: { workspaceId: string }) => {
...
+                  user: {
+                    autoJoin: Number(e.currentTarget.selectedOptions.item(0)?.value) || null,
+                  },
+                })
+              }
+            >
+              <option value="0">None</option>
+              <option value="1">Viewer</option>
+              <option value="3">Developer</option>
+              <option value="131">Admin</option>
```
jcmorrow, line 150: what are these magical numbers?

src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
```diff
@@ -123,6 +123,35 @@ const OrganizationSettings = ({ workspaceId }: { workspaceId: string }) => {
...
+                "border-gray-300": disabled,
+              })}
+              disabled={disabled}
+              id="auto_add_users"
+              name="auto_add_users"
+              value={workspace.settings.features.user.autoJoin || 0}
+              onChange={e =>
+                updateFeature({
+                  user: {
+                    autoJoin: Number(e.currentTarget.selectedOptions.item(0)?.value) || null,
```
jcmorrow, line 142: This is a doozy of a line.

src/ui/types/index.ts
```diff
@@ -163,6 +164,11 @@ export interface WorkspaceSettings {
   motd: string | null;
 }
 
+// https://typeofnan.dev/creating-your-own-deeppartial-type-in-typescript/
+type DeepPartial<T> = Partial<{ [P in keyof T]: DeepPartial<T[P]> }>;
```
jcmorrow, line 168: 🪄

jcmorrow (commented):
I'm not sure I understand 100% of this, but it looks mostly reasonable?

src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
ryanjduffy: They map to a bitmask on the backend. Normally we use strings on the frontend and map them to the bitmask but we don't have that hook in place for `features` atm. It's rudimentary, after all ...

jasonLaster (approved):

src/ui/components/shared/WorkspaceSettingsModal/OrganizationSettings.tsx
ryanjduffy: @bvaughn I think the graphql API expects strings (e.g. `viewer`, `admin`)

bvaughn: It does. It's unclear to me if/how the bitmask strings factor in on the frontend anymore. I guess... maybe they don't?

ryanjduffy: I don't think we should be using the bitmask on the frontend at this point

PR #4779, opened 2021-12-08
Add additional auth0 CSP entries

ryanjduffy:
https://rapidsec.com/csp-packages/auth0


jcmorrow (approved):
I think these are all good additions, it looks like based on [this message](https://discord.com/channels/779097926135054346/917858569720856656/918080623757910017) we might also need `telemetry.replay.io`?

ryanjduffy:
> I think these are all good additions, it looks like based on [this message](https://discord.com/channels/779097926135054346/917858569720856656/918080623757910017) we might also need `telemetry.replay.io`?

Nice catch. I'll add that as well.

PR #4789, opened 2021-12-08
Remove fontfaceobserver

jasonLaster:


jcmorrow:
Let's not be too quick to merge this, if the problem persists is shows up as a much worse visual bug than just the icons missing

jcmorrow (approved):
I've tried to break this a bunch, and it seems quite solid. I can't believe I didn't think of the "what if we deleted all of the code?" solution, that's usually my signature move. 🔴

PR #4799, opened 2021-12-09
Network monitor polish

jcmorrow:
- Make sure that headers are not getting vertically clipped (it seems like sometimes with everything expanded the last header isn't visible even when you scroll all the way down)
- Make the cursor pointer over request rows
- Highlight the currently selected request row
- Bold the term in request and response header sections
- Make text wrap in detail tables

https://user-images.githubusercontent.com/5903784/145447992-52fc296a-2d5e-49e1-a0e3-bbae3a49ca9e.mp4

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestRow.tsx
+++ b/src/ui/components/NetworkMonitor/RequestRow.tsx
@@ -8,6 +8,7 @@ export const RequestRow = ({
   currentTime,
   isFirstInFuture,
   isInPast,
+  isSelected,
   onClick,
   onSeek,
   row,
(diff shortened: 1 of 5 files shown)


src/ui/components/NetworkMonitor/RequestRow.tsx
```diff
@@ -8,32 +8,24 @@ export const RequestRow = ({
...
-  // Did we just pass the row boundary that contains the current time?
-  // If so, let's render this row as the "current" row and all rows
-  // after it as future rows.
-  // if (
-  //   selectedRequest?.id === row.original.id ||
-  //   (!selectedRequest && renderingRequestsInThePast && currentTime <= row.original.point.time)
-  // ) {
-  //   renderingRequestsInThePast = false;
-  //   currentRow = true;
-  // }
```
jasonLaster, line 32: should we delete?

src/ui/components/NetworkMonitor/RequestTable.module.css
```diff
@@ -4,12 +4,21 @@ table.requests {
...
+.row:hover {
+  cursor: pointer;
 }
 
 .row:not(:first-child) {
   border-top: 2px solid var(--theme-toolbar-background);
 }
 
+.row.selected {
+  border: 2px solid var(--primary-accent);
```
jasonLaster, line 19: we should use the same row selected treatment as other places: blue background, white text which is also what chrome does too.

jasonLaster (approved):
One tweak, otherwise great

src/ui/components/NetworkMonitor/RequestRow.tsx
jcmorrow: Well, we already have this feature back in, so I figure we can safely delete the old implementation.

PR #4823, opened 2021-12-10
Only show the first line of folded console errors

### LINKED ISSUE #4630: Console error stack appears to be shown even when collapsed
Demo: https://app.replay.io/recording/5946bfb7-ed4b-4cb2-9cae-80cd42def087

jcmorrow:
Fixes #4630.

If this is an issue for things other than errors we might have to expand it. I'm a little nervous about console styling because it's pretty intricate and I don't have a lot of experience with it, so I kept the targeting pretty specific.

### Before

CleanShot 2021-12-10 at 15 06 41

### After

CleanShot 2021-12-10 at 15 07 06

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -145,6 +145,14 @@
   background-color: var(--console-error-background);
 }
 
+.webconsole-app .message.error .message-body {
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):
Smart

PR #4824, opened 2021-12-10
Fix `breakpoints-05` test

jcmorrow:
We removed the `mapped-source` class from the toggle in  7003f6d784328bc2b45c33e8d2ccbdff3edb60f8, which was breaking the test. Also, there's a problem where if a user who is not logged in tries to dismiss a nag, but that's actually not what was failing the test, so 🤷.

### DIFF
--- a/src/test/harness.js
+++ b/src/test/harness.js
@@ -554,7 +554,7 @@ async function toggleExceptionLogging() {
 }
 
 async function toggleMappedSources() {
-  return clickElement(".mapped-source");
+  return clickElement(".mapped-source button");
 }
 
 async function playbackRecording() {
--- a/src/ui/hooks/users.ts
+++ b/src/ui/hooks/users.ts
@@ -127,17 +127,17 @@ export function useGetUserInfo(): UserInfo & { loading: boolean } {
 }
 
 export function useDismissNag() {
+  const { id, nags } = useGetUserInfo();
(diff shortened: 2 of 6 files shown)


src/ui/hooks/users.ts
```diff
@@ -127,17 +127,19 @@ export function useGetUserInfo(): UserInfo & { loading: boolean } {
...
   if (error) {
     console.error("Apollo error while updating the user's nags:", error);
   }
 
   return (nag: Nag) => {
-    if (!nags || nags.includes(nag)) {
+    console.log({ id });
+    if (!nags || nags.includes(nag) || !id) {
+      console.log("RETURNING");
       return;
```
jasonLaster, line 143: 🧂

src/test/harness.js
```diff
@@ -554,7 +554,7 @@ async function toggleExceptionLogging() {
 }
 
 async function toggleMappedSources() {
-  return clickElement(".mapped-source");
+  return clickElement(".mapped-source button");
```
jasonLaster, line 557: gosh - it is so easy to break these tests and miss it...

jasonLaster (approved):

src/test/harness.js
jcmorrow: Don't worry, we're going to be at a place soon where test failures will be meaningful, I have hope 😄

PR #4832, opened 2021-12-13
Node picker loading

jasonLaster:
- [x] Node picker should have a selected state
- [x] Comment tool should show a tooltip when loading targets

WIP:
- [x] we should move this action to the correct inspector component
- [x] the action should be typed


jasonLaster:
This is now working. The key is that the NodePicker is actually not part of the inspector

PR #4833, opened 2021-12-13
Show a spinner when the graphics are loading

### LINKED ISSUE #4108: Handle different loading states in the video
This is to continue the discussion from #4093, #4096, #4513

We have a few outstanding loading handling items:

- Initial screenshot: When you open up a replay and you're met with a blank screen on the video.
- Video loading: When the video is loading slower than it's playing, and it stutters or doesn't play at all.
- Screenshot repainting: When after navigating, the takes a while to update.

The first one was fixed by #4096. However, we still have the rest. Talked with Jason about this and our conclusions were:
- Video loading: Stuttering is not ideal. We should add some buffering logic that lets the Video indicate that it's buffering, get n (5?) seconds worth of loaded playback, then resume playing until it hits an unloaded screenshot. In which case, it just buffers again.
- Screenshot repainting: We should find a way to be really fast at this so we never have to deal with the 1-2s screenshot update delay between navigations.

jasonLaster:
fixes #4108

Shows a spinner when we're waiting for graphics to load.

Also deletes the Hud


src/ui/reducers/timeline.ts
```diff
@@ -15,6 +15,7 @@ function initialTimelineState(): TimelineState {
     timelineDimensions: { left: 1, top: 1, width: 1 },
     hoveredItem: null,
     trimRegion: null,
+    stalled: false,
```
jcmorrow, line 18: shall we sort these while we're in here? (I know, I'm the worst).

src/ui/state/timeline.ts
```diff
@@ -17,8 +17,8 @@ export interface TimelineState {
     startTime: number;
     startDate: number;
     time: number;
-    stalled?: boolean;
   } | null;
+  stalled?: boolean;
```
jcmorrow, line 21: same on the `sort`ing 😅

jcmorrow (approved):
Look at all those passing tests 😍

src/ui/state/timeline.ts
jasonLaster: yes, but that could be it's own PR.

PR #4838, opened 2021-12-13
Only show material icons after they have loaded fixes #4835

### LINKED ISSUE #4835: MaterialIcon alt tags are appearing before loading
**Description:**
https://app.replay.io/recording/310026b0-6ddb-4e18-b559-ef616a43cc08?point=20444668885348964560410743337386057&time=20799.532295557095&hasFrames=true#

Screen Shot 2021-12-12 at 7 47 49 PM

jasonLaster:
We used to use a font-face loader, but when the font would load via the cache it would fail. This gets around it by checking until the font is loaded

### DIFF
--- a/public/fonts/material_icons/material_icons.css
+++ b/public/fonts/material_icons/material_icons.css
@@ -17,6 +17,11 @@
     url(/fonts/material_icons/MaterialIconsOutlined-Regular.otf) format("opentype");
 }
 
+.material-icon-loaded .material-icons,
+.material-icon-loaded .material-icons-outlined {
+  opacity: 1;
+}
+
 .material-icons,
 .material-icons-outlined {
   font-weight: normal;
@@ -29,6 +34,8 @@
   word-wrap: normal;
   white-space: nowrap;
   direction: ltr;
+  opacity: 0;
+  width: 1rem;
 
   /* Support for all WebKit browsers. */
   -webkit-font-smoothing: antialiased;
--- a/src/ui/components/shared/MaterialIcon.tsx
+++ b/src/ui/components/shared/MaterialIcon.tsx
@@ -1,5 +1,5 @@
 import classnames from "classnames";
-import React from "react";
+import React, { useEffect } from "react";
 
 const SIZE_STYLES = {
   xs: "text-xs",
@@ -18,6 +18,22 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
   iconSize?: keyof typeof SIZE_STYLES;
 };
 
+let isChecking = false;
+function useMaterialIconCheck() {
+  useEffect(() => {
+    if (isChecking) {
+      return;
+    }
+    isChecking = true;
+    let id = setInterval(() => {
+      if (typeof document === "object" && (document as any).fonts.check("12px Material Icons")) {
+        document.body.classList.add("material-icon-loaded");
+        clearInterval(id);
+      }
+    }, 100);
+  }, []);
+}
+
 export default function MaterialIcon({
   children,
   className,
@@ -26,6 +42,8 @@ export default function MaterialIcon({
   iconSize = "base",
   ...rest
 }: MaterialIconProps) {
+  useMaterialIconCheck();
+
   return (
     <div
       {...rest}


src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -26,6 +43,8 @@ export default function MaterialIcon({
   iconSize = "base",
   ...rest
 }: MaterialIconProps) {
+  useMaterialIconCheck();
```
jcmorrow, line 46: I would probably put this at the `App.tsx` level, so it doesn't start setting timeouts from every single material icon. Also, what do you think about passing in the function that does the timeout setting, but leaving `useEffect` in the component?

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,23 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
...
+let isChecking = false;
+function useMaterialIconCheck() {
+  useEffect(() => {
+    if (isChecking) {
+      return;
+    }
+    isChecking = true;
+    let id = setInterval(() => {
+      if (typeof document === "object" && document.fonts.check("12px Material Icons")) {
+        console.log("int", id);
```
jcmorrow, line 30: 🔬

public/fonts/material_icons/material_icons.css
```diff
@@ -17,6 +17,11 @@
     url(/fonts/material_icons/MaterialIconsOutlined-Regular.otf) format("opentype");
 }
 
+.material-icon-loaded .material-icons,
+.material-icon-loaded .material-icons-outlined {
+  opacity: 1;
```
jcmorrow, line 22: I'm not sure if `opacity` or `visibility` is better for accessibility, but either way, we should probably add a `width: 1rem` so that there is less UI jumping.

jcmorrow (commented):
Couple of questions!

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,23 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
...
 
+let isChecking = false;
+function useMaterialIconCheck() {
+  useEffect(() => {
+    if (isChecking) {
+      return;
+    }
+    isChecking = true;
+    let id = setInterval(() => {
+      if (typeof document === "object" && document.fonts.check("12px Material Icons")) {
```
jcmorrow, line 29: I would probably do `(document as any).fonts` here to appease the typescript gods. You could also declare fonts on window, but it doesn't seem particularly useful.

public/fonts/material_icons/material_icons.css
jasonLaster: oh interesting, I assumed opacity would not affect layout

src/ui/components/shared/MaterialIcon.tsx
jasonLaster: This only does one check which is why i like co-locating it

jasonLaster:
https://app.replay.io/recording/e1f35e6f-5edf-4742-aa86-80cfe8a36512?point=11033630826315986259269616490515389&time=15053&hasFrames=false

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,22 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
   iconSize?: keyof typeof SIZE_STYLES;
 };
 
+let isChecking = false;
```
abstractalgo, line 21: Can inline inside the effect by creating a ref object (`useRef`). The effect could then be reused in multiple places (not that we want, but still better encapsulation).

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,22 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
...
+function useMaterialIconCheck() {
+  useEffect(() => {
+    if (isChecking) {
+      return;
+    }
+    isChecking = true;
+    let id = setInterval(() => {
```
abstractalgo, line 28: `isChecking` and a handle for interval always go in pair, so `isChecking` is redundant. We can do something like `const intervalHandleId: number = useRef(-1)` and just use that to check things.

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,22 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
...
+      if (typeof document === "object" && (document as any).fonts.check("12px Material Icons")) {
+        document.body.classList.add("material-icon-loaded");
```
abstractalgo, line 30: Honestly, I'm terrified by this, but if there's no better way... 😬 
Have we considered `.ready()` API (https://stackoverflow.com/a/32292880/1895436), that resolves a Promise, instead of us checking on interval?

Maybe at least not inlining strings, but moving them to some `const`s would be better. (this suggestion is not a blocker)

src/ui/components/shared/MaterialIcon.tsx
```diff
@@ -18,6 +18,22 @@ type MaterialIconProps = React.HTMLProps<HTMLDivElement> & {
...
+        clearInterval(id);
```
abstractalgo, line 31: We also want to clear the interval on the effect cleanup (returned function from the effect).

PR #4839, opened 2021-12-13
Set `workspaceId` in Mixpanel call

### LINKED ISSUE #4822: Track the Organization/Team when we send record analytic events
When we report things to MixPanel (and other analytics) we should include team/org details. This will help with WAT measurement, but it's also just a good thing to have.

jcmorrow:
This will set the `workspaceId` for all events that get sent to MixPanel, and we can use that to determine which teams (or workspaces) have been active in a given time period.

Fixes #4822


ryanjduffy:
Do we need to catch when they select a workspace?

jcmorrow:
> Do we need to catch when they select a workspace from the sidebar?

I think that's a good idea. So far this would work as long as a user views a replay after changing their workspace, but setting it at the time of switch is also good 👍

PR #4843, opened 2021-12-14
Sources loading

jasonLaster:
Sources loading
https://app.replay.io/recording/593520cf-8f09-4faf-b59e-1884d0160320#

### DIFF
--- a/src/devtools/client/debugger/src/client/index.js
+++ b/src/devtools/client/debugger/src/client/index.js
@@ -44,10 +44,10 @@ export async function loadInitialState() {
 let boundActions;
 let store;
 
-function setupDebugger() {
+async function setupDebugger() {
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/client/index.js
```diff
@@ -56,6 +56,7 @@ function setupDebugger() {
       },
     })
   );
+  store.dispatch({ type: "SOURCES_LOADED" });
```
jcmorrow, line 59: Nice change 👍

jcmorrow (approved):
I'm having trouble finding a replay with sources that are slow enough to load to actually try this out on the preview branch, but the replay looks good, and the code makes sense 🌈

PR #4844, opened 2021-12-14
Launcher

jasonLaster:
Screen Shot 2021-12-13 at 11 08 09 PM


jaril:
https://github.com/RecordReplay/devtools/issues/4682

abstractalgo:
The changes are pretty lightweight, so might be a good candidate to take care of some other things that are being touched by these changes - a more typesafe Redux store and opting for `useSelector` instead of `mapStateToProps`, and moving to `.tsx` file with `React.FC` as component's type with props.

Gradual changes *will* accumulate over time. :)

jasonLaster:
Good call. Will switch to tsx. Will try `useSelector` when I have time to experiment.

PR #4846, opened 2021-12-14
Update @recordreplay/protocol

jaril:
This also removes `labelTestSession` since that's been [removed from the protocol](https://github.com/RecordReplay/backend/commit/a2d68898edd7dad959f2a98dfebc45cdc5eebafd) since.


hbenl:
You should also include the updated `package-lock.json`

hbenl (approved):

PR #4852, opened 2021-12-15
Add Stack Trace to Network Monitor Request Detail

### LINKED ISSUE #4808: Show the initiator (call stack at trigger point) in the request detail
We have this information theoretically, but right now it involves another round-trip to the server per-network-request. Maybe in an ideal future we send some of this info down with the requests during the initial network event load?

jcmorrow:
This adds a `Frames` component to the `Stack Trace` tab inside of the network monitor's request detail pane.

CleanShot 2021-12-14 at 17 16 29

It also makes a change so that if you are on a tab in the Network Request Detail panel and the active tabs change (because the active request changes), any invisible tabs will be "redirected" to the headers tab.

Replay: https://app.replay.io/recording/823241b9-54fd-4cc4-b315-3937694c0b0c

Fixes #4808

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.js
@@ -4,7 +4,6 @@
 
 //
 import React, { Component } from "react";
-import PropTypes from "prop-types";
 import classNames from "classnames";
 
 import { getLibraryFromUrl } from "../../../utils/pause/frames";
@@ -125,15 +124,13 @@ export default class Group extends Component {
   }
 
   renderDescription() {
-    const { l10n } = this.context;
     const { group } = this.props;
 
     const frame = group[0];
     const expanded = this.state.expanded;
-    const l10NEntry = this.state.expanded
-      ? "callStack.group.collapseTooltip"
-      : "callStack.group.expandTooltip";
-    const title = l10n.getFormatStr(l10NEntry, frame.library);
+    const title = this.state.expanded
+      ? `Collapse ${frame.library} frames`
+      : `Expand ${frame.library} frames`;
 
     return (
       <div
@@ -169,4 +166,3 @@ export default class Group extends Component {
 }
 
 Group.displayName = "Group";
-Group.contextTypes = { l10n: PropTypes.object };
--- a/src/ui/actions/network.ts
+++ b/src/ui/actions/network.ts
@@ -1,4 +1,7 @@
-import { RequestInfo, RequestEventInfo } from "@recordreplay/protocol";
+import { RequestInfo, RequestEventInfo, TimeStampedPoint } from "@recordreplay/protocol";
+import { ThreadFront } from "protocol/thread";
+import { AppDispatch } from "ui/setup";
+import { createFrame } from "devtools/client/debugger/src/client/create";
 
 export const NEW_NETWORK_REQUESTS = "NEW_NETWORK_REQUESTS";
 
@@ -7,9 +10,14 @@ type NewNetworkRequestsAction = {
   payload: { requests: RequestInfo[]; events: RequestEventInfo[] };
 };
 
-export type NetworkAction = NewNetworkRequestsAction;
+type SetFramesAction = {
+  type: "SET_FRAMES";
+  payload: { frames: any[]; point: string };
+};
 
-export const newNetworkRequestsAction = ({
+export type NetworkAction = NewNetworkRequestsAction | SetFramesAction;
+
+export const newNetworkRequests = ({
   requests,
   events,
 }: {
@@ -19,3 +27,15 @@ export const newNetworkRequestsAction = ({
   type: "NEW_NETWORK_REQUESTS",
   payload: { requests, events },
 });
+
+export function fetchFrames(tsPoint: TimeStampedPoint) {
+  return async ({ dispatch }: { dispatch: AppDispatch }) => {
+    const pause = ThreadFront.ensurePause(tsPoint.point, tsPoint.time);
+    const frames = (await pause.getFrames())?.filter(Boolean) || [];
+    const formattedFrames = await Promise.all(frames?.map((frame, i) => createFrame(frame, i)));
+    dispatch({
+      type: "SET_FRAMES",
+      payload: { frames: formattedFrames, point: tsPoint.point },
+    });
+  };
+}
--- a/src/ui/components/NetworkMonitor/RequestDetails.tsx
+++ b/src/ui/components/NetworkMonitor/RequestDetails.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { RequestSummary } from "./utils";
 import styles from "./RequestDetails.module.css";
 import classNames from "classnames";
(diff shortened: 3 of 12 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.js
```diff
@@ -133,7 +133,7 @@ export default class Group extends Component {
     const l10NEntry = this.state.expanded
       ? "callStack.group.collapseTooltip"
       : "callStack.group.expandTooltip";
-    const title = l10n.getFormatStr(l10NEntry, frame.library);
+    const title = l10n?.getFormatStr(l10NEntry, frame.library);
```
jcmorrow, line 136: I have no idea why this is `undefined` here but not in the `SecondaryPanel`, but not having a title doesn't seem to render strangely 🤷

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -63,6 +64,15 @@ const Cookies = ({ request }: { request: RequestSummary }) => {
   );
 };
 
+const StackTrace = () => {
+  return (
+    <div>
+      <h1 className="p-3 font-bold">Stack Trace</h1>
+      <Frames panel="debugger" />
```
jcmorrow, line 71: I also tried `webconsole` instead of `debugger` here because we check for that in a file, but changing it doesn't seem to do anything, so I'm just leaving as is for now.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.js
jasonLaster: It'd be nice to delete all the references to `l10n?.getFormatStr` we can inline the strings

src/ui/components/NetworkMonitor/RequestDetails.tsx
jasonLaster: that's weird

jasonLaster:
Can you share a replay of it working?

src/ui/components/NetworkMonitor/RequestDetails.tsx
jasonLaster: Lets wrap this in a div so we can add some padding lol. Amazing that this just works though. Also, where do you pass the frames in? If you select different requests you should have different frames?

In hindsight - i think that's the interesting thing. At this point we probably have the network requests in redux and we probably should have a dispatch that fetches the frames and saves them for the request in redux as well so we do not re-fetch them.

jcmorrow: Well, when we jump to a network request we are jumping to a pause with frames, so we are already fetching the frames (because that's what is displayed in the Pause panel). If the fetch is expensive, I agree that we should be caching it, but this change shouldn't be adding any additional load to what was already there, it's just displaying the information in another place. You can see this by opening the pause panel right now and jumping forward and backward to requests in the network monitor.

jcmorrow:
This is working pretty well for me, but I'm getting a console error on navigating via the stack trace that seems troubling: https://app.replay.io/recording/8df35348-cac8-4657-a7ae-ce6538c7a69f

jcmorrow:
Alright, this is finally working. I'm still not 100% sure where the `thread.js` error was coming from, but I did some cleanup and it's no longer happening 🤷

src/ui/actions/network.ts
```diff
@@ -19,3 +27,15 @@ export const newNetworkRequestsAction = ({
...
+
+export function fetchFrames(point: string, time: number) {
+  return async ({ dispatch }: { dispatch: AppDispatch }) => {
+    const pause = ThreadFront.ensurePause(point, time);
+    const frames = await pause
+      .getFrames()
+      .then(frames => (frames || []).filter(Boolean))
+      .then(frames => Promise.all(frames.map((frame, i) => createFrame(frame, i))))
+      .then(frames => frames.filter(Boolean));
+    dispatch({ type: "SET_FRAMES", payload: { frames, point } });
```
jasonLaster, line 39: this is a nit, but it's nicer to use awaits...

```js
const pause = await ThreadFront.ensurePause(point, time);
const fetchedFrames = await pause.getFrames()
const  frames = fetchedFrames?.filter(Boolean).map(createFrame)?.filter(Boolean)
dispatch({ type: "SET_FRAMES", payload: { frames, point } });
```

There are some other implicit things here. 
* Not everything is async
* I dont think we need both Boolean filters

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -49,7 +66,10 @@ export const NetworkMonitor = ({ seek, requests, events, currentTime }: PropsFro
                   table={table}
                   data={data}
                   currentTime={currentTime}
-                  onRowSelect={setSelectedRequest}
+                  onRowSelect={row => {
+                    fetchFrames(row.point.point, row.point.time);
```
jasonLaster, line 70: maybe `fetchFrames` should take `row.point`?

src/ui/reducers/network.ts
```diff
@@ -1,31 +1,52 @@
...
+        ...state,
         events: [...action.payload.events, ...state.events],
         requests: [...action.payload.requests, ...state.requests],
       };
+    case "SET_FRAMES":
+      return {
+        ...state,
+        frames: {
+          ...state.frames,
+          [action.payload.point]: action.payload.frames,
```
jasonLaster, line 34: no longer sure if we need to cache here... probably fine if it is just 

```
return { ...state, frames: action.frames }
```

The one caveat is we'd have to clear it when a new row is selected. Dont feel strongly

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -176,7 +213,9 @@ const RequestDetails = ({
         {activeTab == "cookies" && <Cookies request={request} />}
         {activeTab == "response" && <ComingSoon />}
         {activeTab == "request" && <ComingSoon />}
-        {activeTab == "initiator" && <ComingSoon />}
+        {activeTab == "stackTrace" && (
+          <StackTrace cx={cx} frames={frames} selectFrame={selectFrame} />
```
jasonLaster, line 217: at a ui level, it would be nice to add some hover interactions... but that should probably be a follow up

jasonLaster (approved):
Some small things, but otherwise great

src/ui/reducers/network.ts
jcmorrow: It can work either way, but I'm not a big fan of implicit IDs. When storing state in redux, I think a structure like:

```
{ currentFrames: [] }
```

is *far* inferior to:

{
  currentRequestId: "123",
  framesByRequestId: {
    123: []
  }
}
```

For me, it's not about caching, it's about clarity. Looking at the problems we're dealing with on the billing side right now, it's just another version of this exact problem. I am of the opinion that local data stores should imitate the structure of the source that they are pulling from, unless there is a very good reason not to.

OK, end of rant 😄

src/ui/actions/network.ts
jcmorrow: I think "nicer" is kind of in the eye of the beholder 😄. Personally, I think all of the extra local variables make the await version super noisy, but I'm happy to switch to it if it's closer to the rest of the codebase. 💯 for consistency!

jasonLaster: fair enough

PR #4855, opened 2021-12-15
Fonts loading improvement

abstractalgo:
Using Promise provided by the [fonts API](https://developer.mozilla.org/en-US/docs/Web/API/Document/fonts#examples) (originally discovered at https://stackoverflow.com/a/32292880/1895436).


jasonLaster:
Much cleaner

PR #4858, opened 2021-12-15
Use custom auth0 hook everywhere

### LINKED ISSUE #4821: Can't comment when authenticated via API key
This is affecting QA Wolf and E2E testing

ryanjduffy:
## Issue

Fixes #4821 

## Resolution

There were a few places where we were using the stock `useAuth0` hook rather than our custom wrapper which means that some of our special handling for auth via API keys or the e2e test identity do not work.

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/CommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/CommentEditor.tsx
@@ -10,6 +10,7 @@ import { User } from "ui/types";
 import TipTapEditor from "./TipTapEditor";
 import { FocusContext } from "../CommentCard";
 import classNames from "classnames";
+import useAuth0 from "ui/utils/useAuth0";
 
 export function getCommentEditorDOMId(comment: Comment | Reply) {
   return `comment-editor-${comment.id}`;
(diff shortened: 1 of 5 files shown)


jcmorrow (approved):
Oh my gosh, thank you! I was kicking myself trying to figure out why auth was failing in some tests! Great find!

jasonLaster:
That's so tricky!

ryanjduffy:
But wait ... there's more ... Seems like the tiptap editor doesn't pick up the new auth state because the extensions are created before the auth state is known and aren't updated when the auth state lands.

ryanjduffy:
@jcmorrow - Mind giving this another look? Any concerns with bailing on rendering the editor?

jcmorrow:
> @jcmorrow - Mind giving this another look? Any concerns with bailing on rendering the editor?

Yeah, if we don't absolutely have to `return null` there it would be great not to. I kind of like the current flow which is click to reply -> get asked to login. Right now this happens on the preview branch:

https://user-images.githubusercontent.com/5903784/146281005-8c068a29-8804-4cd3-af70-922c98f8f528.mp4

ryanjduffy:
:+1: I missed that it would drop the "Reply". I'll check it out again.

jcmorrow (approved):
🚢

PR #4865, opened 2021-12-17
Improving `Nodes` perf

abstractalgo:
### problems and culprits

- expanding Nodes in the hierarchy (deep nesting triggered by Node Picker e.g.)
  - [ ] layout recalculation when inserting `Nodes`
    - potential fix: do one final re-layout, instead of one per nesting
  - [ ] communication with backend takes a while
    - potential fix: batch information retrieval (perhaps using `Analysis` module, as suggested by Holger)
- up/down navigation is janky
  - [ ] either because of the ~~calculation of styles~~ or rules rendering. this happens only in some cases where there are a lot rules
    - potential fix: ~~optimize rules calc~~, **compare with regular browser tools what gets shown and if we have a bug** (done) optimize rendering of rules (virtualize rendering e.g.) (**we memo'd things here**), or be smarter about events handling (debounce e.g.)
    
Perf capture (for Node-picking a nested Node):

image

(zooming in a bit)

image

---

- [x] adding more types
- [ ] (batch layout recalc at the end of Node expansion)
- [ ] (batch protocol)
- [ ] improve Rules rendering and perf
- [ ] do a final sweep for cleanup (remove console logs etc)


src/devtools/client/inspector/rules/components/RulesApp.tsx
```diff
@@ -38,15 +43,17 @@ export const RulesApp: FC<RulesAppProps> = ({
...
+    () => ({
+      onToggleDeclaration,
+      onToggleSelectorHighlighter,
+      showDeclarationNameEditor,
+      showDeclarationValueEditor,
+      showNewDeclarationEditor,
+      showSelectorEditor,
+    }),
+    []
+  );
```
jasonLaster, line 56: smart

src/devtools/client/inspector/rules/rules.js
```diff
@@ -346,8 +346,8 @@ class RulesView {
    *         The pseudo class to toggle on or off.
    */
   onTogglePseudoClass(value) {
-    this.store.dispatch(togglePseudoClass(value));
-    this.inspector.togglePseudoClass(value);
+    // this.store.dispatch(togglePseudoClass(value));
+    // this.inspector.togglePseudoClass(value);
   }
```
jasonLaster, line 351: lets delete more code paths like this that will never be used. Everytime we touch the inspector we should deleting lots of code :)

jasonLaster (commented):
Looking good!

what's the perf impact?

src/devtools/client/inspector/rules/components/RulesApp.tsx
```diff
@@ -63,120 +70,123 @@ export const RulesApp: FC<RulesAppProps> = ({
...
     if (!rulesQuery && rules.length === 0) {
       return <div className="devtools-sidepanel-no-result">{getStr("rule.empty")}</div>;
     }
 
-    const inheritedRules = [];
-    const pseudoElementRules = [];
-    const styleRules = [];
+    const nonEmptyRules = rules.filter(rule =>
+      rule.declarations.some(declaration => !declaration.isInvisible)
+    );
```
abstractalgo, line 151: The rest of the changes are more on the side of "maintenance", but this is the actual change that saves a huge number of rules from being unnecessarily rendered. This change seems to be on par with what browsers render in their own DevTools panels, so this is not "a hack".

src/devtools/client/inspector/rules/components/Declarations.tsx
```diff
@@ -14,11 +14,9 @@ export const Declarations: FC<DeclarationsProps> = ({ declarations, query }) =>
 
   return (
     <ul className="ruleview-propertylist">
-      {declarations
-        .filter(declaration => !declaration.isInvisible)
```
abstractalgo, line 18: Because we render `Rule`s slightly different between the three categories (pseudo, style, inherited), this wasn't filtering sufficiently all of the rules, so that's why the filtering was moved to be a responsibility of the `RulesApp`.

abstractalgo:
Want to merge these changes and also some things from https://github.com/RecordReplay/devtools/pull/4882/ because it rewrites some of the more interesting parts of how panels are rendered.

src/devtools/client/inspector/rules/rules.js
abstractalgo: I meant to save it for later as it may someone some time, but sure, I can remove code. :)

src/devtools/client/inspector/markup/markup.ts
```diff
@@ -162,8 +192,12 @@ class MarkupView {
    * @param  {String} nodeId
    *         The NodeFront object id to select.
    */
-  onSelectNode(nodeId) {
-    this.selection.setNodeFront(ThreadFront.currentPause.getNodeFront(nodeId), {
+  onSelectNode(nodeId: string) {
+    assert(ThreadFront.currentPause);
+
+    console.log(`select ${nodeId}`);
```
hbenl, line 198: I think you forgot this

src/devtools/client/inspector/markup/markup.ts
```diff
@@ -127,7 +156,8 @@ class MarkupView {
 
   get eventTooltip() {
     if (!this._eventTooltip) {
-      this._eventTooltip = new HTMLTooltip(this.toolbox.doc, {
+      this._eventTooltip = new HTMLTooltip({
```
hbenl, line 159: You accidentally removed the first argument here

src/devtools/client/inspector/rules/components/RulesApp.tsx
```diff
@@ -9,9 +9,14 @@ import { Toolbar } from "devtools/client/inspector/rules/components/Toolbar";
 import { getStr } from "devtools/client/inspector/rules/utils/l10n";
 import { useSelector } from "react-redux";
 import { RulesState, RuleState } from "../state/rules";
+import { RuleInheritance } from "../models/rule";
+import _ from "lodash";
```
hbenl, line 13: You don't seem to be using lodash in this file but even if you did you shouldn't import it like this, otherwise all of lodash will be included in the bundle. We always import single functions from lodash, e.g. `import throttle from "lodash/throttle";`

src/devtools/client/inspector/rules/components/RulesApp.tsx
```diff
@@ -63,120 +70,123 @@ export const RulesApp: FC<RulesAppProps> = ({
...
-        component: Rules,
-        componentProps,
-        header: getStr("rule.pseudoElement"),
-        id: "rules-section-pseudoelement",
-        opened: Services.prefs.getBoolPref(SHOW_PSEUDO_ELEMENTS_PREF),
-        onToggle: (opened: boolean) => {
-          Services.prefs.setBoolPref(SHOW_PSEUDO_ELEMENTS_PREF, opened);
+  const renderPseudoElementRules = useCallback(
+    (rules: RuleState[]) => {
+      type FCProps<C> = C extends FC<infer P> ? P : never;
```
hbenl, line 114: That is such a useful type, perhaps we should move it to some utilities file

hbenl (approved):

src/devtools/client/inspector/rules/rules.js
jasonLaster: Understood. We can always look at how Firefox works. We should prioritize what we're currently supporting

src/devtools/client/inspector/rules/components/RulesApp.tsx
jasonLaster: Smart

src/devtools/client/inspector/rules/components/RulesApp.tsx
abstractalgo: My keyboard is acting up and while doing auto-complete it accidentally does a selection and IDE then auto-imports what it can, but I should be more careful when doing a cleanup later. Good catch, thanks 👍

src/devtools/client/inspector/rules/components/RulesApp.tsx
abstractalgo: Because our codebase rarely uses `FC` components and we still have a lot of untyped JS, it's unlikely that many of these cases will come up. Eventually we'll migrate everything to TS and we won't have to do inference ourselves but TS typecheker will be able to warn us (i.e. when we migrate the `Accordion` component). I'll refrain from complicating it for now.

src/devtools/client/inspector/markup/markup.ts
abstractalgo: But `toolbox` doesn't have `doc` property. And in other cases where `HTMLTooltip` is being instantiated, it makes the same mistake of passing an undefined prop from some inspector's toolbox object. It's also very silly because the first argument is not being used at all in the `HTMLTooltip` constructor. Either way, good catch.

I'll just pass a global `Document` object fr now, but maybe removing the first argument for the constructor of `HTMLTooltip` would be an even better fix.

src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -203,8 +204,8 @@ export function selectionChanged(
   };
 }
 
-export function selectNode(nodeId: string, reason?: string): UIThunkAction {
-  return ({ toolbox }) => {
+export function selectNode(nodeId: string, reason?: NodeSelectionReason): UIThunkAction {
+  return ({ toolbox }: { toolbox: DevToolsToolbox }) => {
```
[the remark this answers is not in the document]
abstractalgo: I'm not sure what you meant 😕

PR #4869, opened 2021-12-17
Fix broken inspector test

jcmorrow:
I think this works?

### DIFF
--- a/src/devtools/client/debugger/src/components/WelcomeBox.tsx
+++ b/src/devtools/client/debugger/src/components/WelcomeBox.tsx
@@ -91,7 +91,7 @@ function WelcomeBox({
       </div>
       <div className="launcher justify-center z-10 flex mt-4 px-8 w-full ">
         {actions.map(action => (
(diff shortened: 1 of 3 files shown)


jasonLaster (approved):
This is super strange

jasonLaster:
seeing a TS error for DeclarationValue

jcmorrow:
> This is super strange

Yeah. Also, if you want an example of something that was *really* difficult to debug, even with Replay, this is a good example. This was manifesting as a `Objects are not valid as a React child` error!

jasonLaster (approved):

jasonLaster:
Yes! We should keep a document of hard to debug bugs!

jasonLaster:
And yeah - i think i hit it locally too

https://app.replay.io/recording/fbb1bb91-149f-420f-8efa-9c63de32e816?point=19795631794575832568756173673595881&time=21189.11028610629&hasFrames=true

PR #4874, opened 2021-12-18
Disable hot event listeners fixes #4861

### LINKED ISSUE #4861: Disable event listeners with max points
We currently allow users to view console logs for event listeners with more than the max points. We should disable that in the same way that we disable breakpoints with more than the max number of hits.

Screen Shot 2021-12-17 at 8 07 49 PM

jasonLaster:


src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
```diff
@@ -295,16 +295,21 @@ class EventListeners extends Component {
       return null;
     }
 
+    const isHot = points.length > prefs.maxHitsEditable;
+    const title = isHot ? `Cannot view ${event.name} events` : `View ${event.name} events`;
+    console.log(isHot, points.length);
```
jcmorrow, line 300: ✂️

src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
```diff
@@ -295,16 +295,21 @@ class EventListeners extends Component {
...
+
     return (
       <li
-        className="px-2 pl-5 mb-1 mr-1 flex flex-row items-center hover:bg-gray-200 rounded-md"
+        className="px-2 pl-5 mb-1 mr-1 flex flex-row items-center hover:bg-gray-100 rounded-md"
         key={event.id}
       >
         <div className="flex flex-row justify-between w-full">
-          <label className="w-full flex flex-row items-center">
+          <label title={title} className="w-full flex flex-row items-center">
```
jcmorrow, line 308: Is this just for the hover text?

src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
```diff
@@ -313,11 +318,9 @@ class EventListeners extends Component {
               {searchText ? this.renderCategory(category) : null}
               {event.name}
             </span>
-            {features.eventCount ? (
```
jcmorrow, line 316: what is eventCount supposed to be doing here? It looks like now in the case where `features.eventCount` is false but there are `points` we will still render all of this, is that right?

jcmorrow (commented):
Just a couple questions and a `console.log`

src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
jasonLaster: yeah - i think it's safe to remove this

src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
jasonLaster: yep

PR #4876, opened 2021-12-18
Dont let the editor be hidden by the split panes

jasonLaster:
Previously it was possible to slide the devview splitpane so far that it would hide the editor which was especially problematic on refresh.

* also removes some calls to updateTimelineDimensions which are no longer needed.
* also renames toolboxHeight to toolboxSize

There is still an issue resizes the sidebar when the editor is hidden

### DIFF
--- a/src/ui/components/Views/DevView.tsx
+++ b/src/ui/components/Views/DevView.tsx
@@ -8,13 +8,35 @@ import Toolbar from "../Toolbar";
 import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
 
 import { installObserver } from "../../../protocol/graphics";
-import { updateTimelineDimensions } from "../../actions/timeline";
 import { prefs } from "../../utils/prefs";
 import { selectors } from "../../reducers";
 import { UIState } from "ui/state";
 
+function minSize(sidePanelCollapsed: boolean, showEditor: boolean): string {
+  if (!sidePanelCollapsed && showEditor) {
+    return "300px";
+  }
+
+  if (!sidePanelCollapsed || showEditor) {
+    return "200px";
+  }
+
+  return "0px";
+}
+
+function maxSize(sidePanelCollapsed: boolean, showEditor: boolean) {
+  if (showEditor) {
+    return "80%";
+  }
+
+  if (sidePanelCollapsed) {
+    return "0";
+  }
+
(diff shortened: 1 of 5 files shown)


src/ui/components/Views/DevView.tsx
```diff
@@ -47,8 +44,14 @@ function DevView({
         <SplitBox
           style={{ width: "100%", overflow: "hidden" }}
           splitterSize={8}
-          initialSize={prefs.toolboxHeight as string}
-          minSize={showEditor || !sidePanelCollapsed ? "240px" : "0%"}
+          initialSize={prefs.toolboxSize as string}
+          minSize={
+            !sidePanelCollapsed && showEditor
```
jaril, line 49: Nested ternary is 😬. Maybe a helper similar to `maxSize()` that looks something like this?

```js
function minSize(...) {
  if (!sidePanelCollapsed) {
    if (showEditor) {
      return "260px"
    } else {
      return "200px"
    }
  } else {
    return "0%"
  };
}
```

src/ui/components/Views/DevView.tsx
```diff
@@ -71,9 +74,7 @@ const connector = connect(
     showEditor: selectors.getShowEditor(state),
     sidePanelCollapsed: selectors.getPaneCollapse(state),
   }),
-  {
-    updateTimelineDimensions,
-  }
+  {}
```
jaril, line 77: Can just remove this.

src/ui/components/Views/NonDevView.tsx
```diff
@@ -71,9 +69,7 @@ const connector = connect(
     selectedPrimaryPanel: selectors.getSelectedPrimaryPanel(state),
     sidePanelCollapsed: selectors.getPaneCollapse(state),
   }),
-  {
-    updateTimelineDimensions,
-  }
+  {}
```
jaril, line 72: Same here — can just delete the `{}`

src/ui/components/Views/NonDevView.tsx
```diff
@@ -33,15 +32,14 @@ export function EventsFilter() {
   );
 }
 
-function NonDevView({ updateTimelineDimensions, sidePanelCollapsed }: PropsFromRedux) {
```
jaril, line 36: Nice.

PR #4877, opened 2021-12-18
Show ReplayLogo when loading DevView

jasonLaster:
I've found that it's possible in chrome to wait 10-30 seconds to view the DevView which is purely waiting on the import.

This moves that logic to DevTools and uses suspense to show a loader


jaril (approved):

jcmorrow:
> 10-30 seconds

Holy crap

PR #4882, opened 2021-12-20
Improvements to the Inspector and a rewrite

abstractalgo:
Improvements to the Inspector:
- [x] removed a 3-pane mode completely (it was unused)
- [x] complete rewrite of the `InspectorApp` (i.e. main Elements panel) to a nice, memoized FC component with hooks
- [x] more type safety around `Selection`, `DevToolsToolbox` and `Inspector`


src/devtools/client/inspector/components/App.tsx
```diff
@@ -1,254 +1,142 @@
...
-  private onSplitboxResize = (width: number) => {
-    const { is3PaneModeEnabled } = this.props;
-    if (is3PaneModeEnabled) {
-      prefs.sidebarSize = width;
-    } else {
-      prefs.splitSidebarSize = width;
-    }
-  };
+  const sidebarContainerRef = useRef<HTMLDivElement>(null); // nosemgrep
+  const splitBoxRef = useRef<SplitBox>(null); // nosemgrep
```
hbenl, line 39: I think these refs aren't needed anymore (they are set below but never used).

src/devtools/client/inspector/components/App.tsx
```diff
@@ -1,254 +1,142 @@
...
-                      </div>
-                    </div>
-                  </div>
-                }
-                endPanel={
-                  <div className="devtools-inspector-tab-panel">
-                    <div id="inspector-sidebar-container">
-                      <div id="inspector-sidebar" hidden={false}>
-                        {this.renderSidebar()}
+  }, [inspector, inspectorInited, activeTab]);
```
hbenl, line 74: This array should include all the props that you pass to the components in the memoized function. Right now these props never change their value (I think), but there's no guarantee that will always be the case.

src/devtools/client/shared/components/splitter/SplitBox.css
```diff
@@ -60,9 +60,13 @@
...
+/* This disables splitter bar dragging when the user is
 in the console. Dragging/resizing should only work in the debugger
 or the inspector where the split console can be displayed.  */
 #toolbox.console .split-box.horz > .splitter {
   pointer-events: none;
 }
+
+.inspector-sidebar-splitter > .splitter {
+  background: var(--theme-splitter-color);
+}
```
hbenl, line 72: I recently added https://github.com/RecordReplay/devtools/blob/c825925daa9294c21239f076c06896996c51b440/src/devtools/client/themes/inspector.css#L46-L48, which does the same thing. One of them could be removed.

hbenl (approved):
Found some details that could be improved, otherwise LGTM.

src/devtools/client/inspector/components/App.tsx
abstractalgo: Yeah, this is all, and better safe than sorry.

src/devtools/client/shared/components/splitter/SplitBox.css
abstractalgo: I even remember it and yet mistakenly thought it _was_ the same file; will remove.

PR #4885, opened 2021-12-20
Repaint after evaluations

hbenl:


jasonLaster (approved):
It'd help to have a before/after replay here, but i see what you're doing and it looks good.

PR #4886, opened 2021-12-20
Add special behavior for the default sidepanel open when a replay is opened

jaril:


src/ui/graphql/recordings.ts
```diff
@@ -24,6 +24,9 @@ export const GET_RECORDING = gql`
       ownerNeedsInvite
       userRole
       operations
+      comments {
+        id
+      }
```
jasonLaster, line 29: haha - we should probably move the last commit or commenty things to its own PR

src/ui/reducers/layout.ts
```diff
@@ -1,26 +1,147 @@
...
+  }
+
+  if (comments.length) {
+    return "comments";
+  } else {
+    return syncInitialLayoutState.selectedPrimaryPanel;
+  }
+};
+
+export async function getInitialLayoutState(): Promise<LayoutState> {
```
jasonLaster, line 37: nice - this is a lot cleaner

jasonLaster (commented):
Looks great!

Also, because the components dont change too much, the risk seems smaller than i'd expect

PR #4887, opened 2021-12-20
Remove splitConsole references

jaril:
This removes a bunch of splitConsole related stuff that we used to care about while devtools was part of the browser, but is irrelevant now.


abstractalgo (approved):
lgtm

PR #4889, opened 2021-12-22
Show the target URL in the replay info

### LINKED ISSUE #4398: Replay Info Pane should include the last URL
It's important for users to be able to quickly get the recorded URL. Today the best way to do that is to go to the library or run location.href in the console.

In the future, i think the URL should be available above the video because it can change but that's not needed now.

NOTE: it is fine to just show the domain and on click copy the full url.

jaril:
Fix #4398.

image

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -2,16 +2,13 @@ import React, { ReactNode } from "react";
 import hooks from "ui/hooks";
 import { OperationsData } from "ui/types";
 import { formatRelativeTime } from "ui/utils/comments";
-import { getRecordingId } from "ui/utils/environment";
+import { getDisplayedUrl, getRecordingId } from "ui/utils/environment";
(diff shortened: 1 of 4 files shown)


jaril:
Mostly checking if the test will pass

src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -53,6 +53,12 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
           <MaterialIcon iconSize="xl">{icon}</MaterialIcon>
           <div>{summary}</div>
         </Row>
+        <Row>
+          <MaterialIcon iconSize="xl">language</MaterialIcon>
+          <div className="overflow-hidden overflow-ellipsis whitespace-pre" title={recording.url}>
+            {recording.url}
```
jasonLaster, line 59: we'll probably want to format it so it only shows the query params on hover, but this is a good first step

jaril: Fixed!

PR #4892, opened 2021-12-22
Make the network monitor's orientation responsive

jcmorrow:
If it's wider than 700px, it will render the panes side-by-side. If it's
narrower, it will render as stacked panes.

https://user-images.githubusercontent.com/5903784/147041986-4db01250-0a3e-4aef-b8a6-4cbfa0cd9e62.mp4

### DIFF
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -1,5 +1,5 @@
-import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
-import React, { useState } from "react";
+import SplitBox from "devtools-splitter";
+import React, { useEffect, useRef, useState } from "react";
 import { connect, ConnectedProps } from "react-redux";
 import { actions } from "ui/actions";
 import { getEvents, getFormattedFrames, getRequests } from "ui/reducers/network";
@@ -25,6 +25,9 @@ export const NetworkMonitor = ({
 }: PropsFromRedux) => {
   const [selectedRequest, setSelectedRequest] = useState<RequestSummary>();
   const [types, setTypes] = useState<Set<RequestType>>(new Set(["xhr"]));
+  const [vert, setVert] = useState<boolean>(false);
+
+  const container = useRef<HTMLDivElement>(null);
 
   const closePanel = () => setSelectedRequest(undefined);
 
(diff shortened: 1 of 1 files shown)


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -37,8 +40,22 @@ export const NetworkMonitor = ({
     setTypes(new Set(types));
   };
 
+  let resizeObserver = new ResizeObserver(() => {
+    if (container.current!.offsetWidth > 700) {
+      setVert(true);
+    } else {
+      setVert(false);
+    }
+  });
```
jasonLaster, line 49: smallest nit in the world, but one of my things is unecessary booleans

```js
  let resizeObserver = new ResizeObserver(() => setVert(container.current!.offsetWidth > 700))
```

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -1,5 +1,5 @@
-import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
-import React, { useState } from "react";
+import SplitBox from "devtools-splitter";
```
jasonLaster, line 1: yeesh... we should really do something about two splitter components and this vendors file...

jasonLaster (approved):

src/ui/components/NetworkMonitor/index.tsx
abstractalgo: `resizeObserver` should be a ref probably.

jcmorrow: @abstractalgo can you elaborate on that? I'm not sure what you mean?

src/ui/components/NetworkMonitor/index.tsx
jcmorrow: For record-keeping: we did https://github.com/RecordReplay/devtools/pull/4906

src/ui/components/NetworkMonitor/index.tsx
abstractalgo: To create `resizeObserver` via `useRef`, because currently, we're recreating it on every rerender. @jcmorrow

jcmorrow: ah, thanks, got it!

PR #4902, opened 2021-12-23
Console messages loading

jasonLaster:


abstractalgo (changes requested):
In `src/devtools/client/webconsole/reducers/messages.d.ts` we're missing a bunch of things from `src/devtools/client/webconsole/reducers/messages.js`'s full interface (and among them `messagesLoaded: boolean;` for this PR).

The rest looks simple enough.

PR #4905, opened 2021-12-23
Set the comments panel as the default if there is a comment

### LINKED ISSUE #4687: Default to Replay Info unless there are comments

jaril:
Fix #4687.

This is the second part for https://github.com/RecordReplay/devtools/pull/4904.

This changes how we render the app so that we have access to things provided by `AppUtilities` (e.g. apollo client). This means that by the time we can use queries as we bootstrap the store.

Behavior:
1) User's first time viewing replay A and there's no comment: Show Events/Replay Info
2) User's first time viewing replay A and there's a comment: Show Comments
3) User's n-th time viewing replay A: Show last panel they were on

https://user-images.githubusercontent.com/15959269/147260709-1f70c9c7-5b30-45c4-9bfd-cdf3fe0e2eb4.mov

### DIFF
--- a/src/ui/setup/index.ts
+++ b/src/ui/setup/index.ts
@@ -36,7 +36,6 @@ export async function bootstrapApp() {
   if (typeof window === "undefined") return store;
 
   setupTelemetry();
-
   setupDOMHelpers();
 
   window.store = store;
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -4,14 +4,14 @@ import "ui/utils/whatwg-url-fix";
 import Head from "next/head";
 import type { AppContext, AppProps } from "next/app";
 import NextApp from "next/app";
-import React, { useEffect, useState } from "react";
+import React, { ReactNode, useEffect, useState } from "react";
 import { Provider } from "react-redux";
 import { IntercomProvider } from "react-use-intercom";
 import tokenManager from "ui/utils/tokenManager";
 import { ApolloWrapper } from "ui/utils/apolloClient";
-import LoadingScreen, { LoadingScreenTemplate } from "ui/components/shared/LoadingScreen";
+import LoadingScreen, { StaticLoadingScreen } from "ui/components/shared/LoadingScreen";
 import ErrorBoundary from "ui/components/ErrorBoundary";
-import App from "ui/components/App";
+import _App from "ui/components/App";
 import { bootstrapApp } from "ui/setup";
 import "image/image.css";
 import { Store } from "redux";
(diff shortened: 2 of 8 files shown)


jasonLaster:
This might be a good util to tack onto app

```
app.clearSessions = () => indexedDB.databases().then(db => indexedDB.deleteDatabase(db.name))
```

src/ui/setup/index.ts
```diff
@@ -76,5 +76,14 @@ export async function bootstrapApp() {
...
   });
 
+  // Only initialize the layout state once the apollo client is ready.
+  clientWaiter.promise.then(async () => {
+    const layoutState = await getInitialLayoutState();
+    store.dispatch(setLayout(layoutState));
+    // We don't start setting things into local preferences until we've
+    // finished fully initializing the state.
+    registerStoreObserver(store, updatePrefs);
+  });
```
jasonLaster, line 86: this feels like a smell on many levels. 

why can't we set the state before apollo is setup?

Is there a way to kick off an action from a react component once apollo is setup?

jasonLaster: cc @jcmorrow

jaril: Yeah

The [state's still set up before apollo is set up](https://github.com/RecordReplay/devtools/blob/master/pages/_app.tsx#L137-L157). If you're asking about `registerStoreObserver`, it's pushed to here because otherwise it would create a local session immediately. If it does, then by the time we run getInitialLayoutState, there's already a local session for that replay, even if it's the user's first time opening it.

This code is late night spaghetti though so @jcmorrow if you've got ideas let me know!

src/ui/setup/index.ts
```diff
@@ -32,16 +32,15 @@ export async function bootstrapApp() {
...
 
   if (typeof window === "undefined") return store;
 
   setupTelemetry();
-
   setupDOMHelpers();
 
   window.store = store;
 
-  registerStoreObserver(store, updatePrefs);
```
jcmorrow, line 44: Is it important that we moved this a couple of lines, or was it just noise from us shuffling things around? If it's just noise, maybe let's revert this change.

jcmorrow (approved):
Pairing on this reminded me of the old (I think Martin Fowler) quote:

> First make the change easy (this might be hard), then make the easy change

src/ui/setup/index.ts
```diff
@@ -1,4 +1,4 @@
-import { bootstrapStore } from "./store";
+import { bootstrapStore, extendStore } from "./store";
```
jasonLaster, line 1: is `extendStore` used

jasonLaster (commented):
Trust you two

PR #4915, opened 2021-12-24
WIP Record E2E tests with Playwright

### LINKED ISSUE #4501: E2E V2
**Description:**
Now that we've migrated to Next.js and stabilized our e2e tests we can begin proactively improving the tooling

- [ ] Run Tests against Next.js preview branches
- [ ] Use Playwright to record examples
- [ ] Use Playwright to test devtools

### Test Next.js preview branches

Using preview branches will simplify our testing infra. Speed up builds. Allow the tests to run anywhere...

### Use Playwright to record examples + devtools

Using playwright to record the examples will simplify `run.js`. It will also let us exercise the playwright tools.

jasonLaster:
Fixes #4501

Summary:

This PR does a couple of things, but the primary goal is to replace the browser specific logic with Playwright. 

* Extracts a shared `state` object which makes it easier to isolate some of the logic
* Extracts the examples, test, and node logic into their own files so it is easier to see what they are doing
* Removes some secondary options which I wasn't sure if we needed: `--separate`, `--record-viewer`, 
* I was not able to test Chromium primarily because i'm working on a mac
* I was not sure how to test Node because it did not write the recording id to a file.  

Requirements:
- [x] Record Gecko examples with playwright
- [x] Extract recording logic out of `run.js` so it is easier to work incrementally
- [ ] Record Chromium examples with playwright
- [ ] Record Node with replay-node 
- [ ] Record test run with playwright
- [ ] add mac12 to registry https://github.com/RecordReplay/playwright/pull/5


public/test/examples/bundle.js
```diff
@@ -116,7 +116,7 @@ function update(obj, arr) {
   arr.push("new");
 }
 function recordingFinished() {
-  dump(`RecReplaySendAsyncMessage Example__Finished`);
+  window.Example__Finished = true; dump(`RecReplaySendAsyncMessage Example__Finished`);
```
jcmorrow, line 119: This seems like it should be centralized someplace to avoid the shotgun surgery.

test/e2e/recordPlaywright.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  const browser = await playwright[browserName].launch({
+    headless: false,
+  });
+
+  const context = await browser.newContext();
+  const page = await context.newPage();
+  await page.goto(example);
+  await waitUntil(() => page.evaluate(() => window.Example__Finished));
```
jcmorrow, line 22: Does this put a 10 second timer on all tests by default? Usually `waitUntil` only waits for 10 seconds I think.

test/e2e/recordPlaywright.js
```diff
@@ -0,0 +1,64 @@
...
+  await page.close();
+  await context.close();
+  await browser.close();
+}
+
+async function upload(state, example) {
+  const list = cli.listAllRecordings();
+  const recording = _.findLast(list, rec => rec.metadata.uri == example);
```
jcmorrow, line 30: This works for our test suite because we put the test name in the URL, but it's probably worth flagging as something that won't work on other people's tests.

test/e2e/state.js
```diff
@@ -0,0 +1,141 @@
...
+
+  // CONSTANTS
+
+  // Creating an example returns this constant if the test should be skipped.
+  SkipTest: "skipped",
+
+  // Creating an example returns this constant if the test should be recorded
+  // as part of the viewer process. The test will be recorded and then stopped to
+  // switch to the viewer, as when using replay normally.
+  RecordExampleInViewer: "record-example-in-viewer",
```
jcmorrow, line 49: I'm not sure I understand this one. Why do we have a constant for this rather than doing something like `clickOn(viewerButton)` in the tests that need it?

test/e2e/state.js
```diff
@@ -0,0 +1,141 @@
...
+        break;
+      case "--separate":
+        state.shouldRecordExamples = true;
+        state.recordExamplesSeparately = true;
+        break;
+      case "--timeout":
+        state.testTimeout = +argv[++i];
+        break;
+      case "--target":
+        state.onlyTarget = argv[++i];
```
jcmorrow, line 93: Total side-note: I would love to get rid of the `onlyTarget` variable at some point and make it so that passing `--target` just changes targets to `[onlyTarget]`.

jcmorrow (commented):
Super interesting, looking forward to hearing how it's been working on this so far!

test/e2e/recordPlaywright.js
jasonLaster: this is just for the examples, but might be something to bump

test/e2e/recordPlaywright.js
jasonLaster: that's fair. this code isn't trying to be generic.

test/e2e/state.js
jasonLaster: yeah - i didn't follow this either. Lets see if we still need it.

test/e2e/state.js
jasonLaster: hmm, where is targets defined? I think i isolated this variable to one spot which makes it a bit easier to modify going forward.

jcmorrow: `targets` comes in from `manifest.js`

test/e2e/getExample.js
```diff
@@ -0,0 +1,93 @@
...
+  return recordingId;
+}
+
+function updateExampleFile(state, example, recordingId) {
+  const newExampleRecordings = { ...state.ExampleRecordings, [example]: recordingId };
+  fs.writeFileSync("./test/example-recordings.json", JSON.stringify(newExampleRecordings, null, 2));
+}
+
+async function recordExample(state, example, target) {
+  const browser = target == "gecko" ? "firefox" : "chromium";
```
jcmorrow, line 52: This check doesn't make any sense to me. Why do we designate `gecko` in the manifest but then translate it into `firefox` here? Would it be better to just update the manifest to say `firefox` and remove this ternary?

jasonLaster: Agreed.

I could make the argument for gecko being consistent with chromium, but in a playwright world it is simpler to use playwright conventions instead of mapping between the two.

public/test/examples/cra/src/App.js
```diff
@@ -28,7 +28,7 @@ function App() {
       console.error(new Error("Baz"));
 
       // eslint-disable-next-line no-undef
-      dump(`RecReplaySendAsyncMessage Example__Finished`);
+      console.log("Example__Finished");
```
bhackett1024, line 31: It seems like this weird string can be changed to something more appropriate now, or is there a better way to signal to playwright that the test has finished?

src/test/harness.js
```diff
@@ -900,9 +901,13 @@ const commands = mapValues(testCommands, (command, name) => {
...
   console.log(`# Test ${description}`);
-  await start();
-  await cbk();
-  finish();
+  try {
+    await start();
+    await cbk();
+    finish();
+  } catch (e) {
+    window.TestFinished = e.message;
```
bhackett1024, line 909: What is this for?  This needs a comment at least.

bhackett1024 (approved):

test/e2e/runTest.js
```diff
@@ -47,11 +47,20 @@ async function recordBrowser(state, test, testPath, browserName) {
   return { success, why };
 }
 
+const pathname = url => {
+  try {
+    new URL(url).pathname;
+  } catch {
+    return "";
+  }
+};
```
jasonLaster, line 56: I bet we can simplify this by simply splitting on the `/` and getting the last part :)

public/test/examples/cra/src/App.js
jasonLaster: there aren't any great mechanisms for the webpage to talk to playwright and it's probably something that should be discouraged. Our tests and examples are a bit weird in this regard.

We probably could re-write our examples so that playwright decides when they are done.
At somepoint, we might might want to re-write these tests so that they are also run through playwright as opposed to being loaded in content. This doesn't seem like a priority for the time being though.

src/test/harness.js
jasonLaster: Definitely. This is a _clever_ way to pass in the reason why the test failed, but I think there's a simpler way to do it.

jasonLaster:
See this example which works by waiting for deployment_status and then re-running the checks: https://github.com/jasonLaster/oyvey/blob/main/.github/workflows/test.yml

test/header.js
```diff
@@ -1,166 +0,0 @@
-/* Copyright 2020 Record Replay Inc. */
```
bhackett1024, line 1: I forgot to add in my review, but this file is actually used by the backend for static live tests.  It would be difficult to rework those tests to use playwright currently for a couple reasons:

1. We don't release a playwright windows build currently.
2. Playwright builds are generated from a branch that is merged from gecko's webreplay-release branch, so generating playwright builds for gecko changes that haven't merged to that branch isn't possible yet.  Our CI needs to run the static live tests on gecko branches.

I don't see an easy fix for 2.  I think the best fix would be to add this to the backend (cutting the dependency) before removing it from devtools.

test/utils.js
```diff
@@ -1,128 +0,0 @@
...
-  "Contents/MacOS/firefox", // macOS unbranded
-  "replay", // linux, using distribution.
-  "firefox", // linux unbranded
-  "dist/bin/replay", // linux, testing with local build.
-  "dist/bin/firefox", // linux local unbranded
-  "replay.exe", // windows
-];
-
-// Get the executable to use when starting gecko.
-function findGeckoPath() {
```
bhackett1024, line 25: The backend uses this function, see test/runlive/run.ts.  I think this logic should be copied to the backend before removing it here, so the dependency on the devtools can be removed.

bhackett1024:
One other dependency is that the standalone tests in the backend use the devtools test runner, including during backend CI.  Can you check that running standalone tests still works with these changes?

test/mock/src/runTest.ts
```diff
@@ -5,7 +5,9 @@ dotenv.config();
 
 type BrowserName = "chromium" | "firefox";
 let browserName: BrowserName = process.env.PLAYWRIGHT_CHROMIUM ? "chromium" : "firefox";
+console.log({ RECORD_REPLAY_PATH: process.env.RECORD_REPLAY_PATH });
```
jasonLaster, line 8: Is this log needed?

test/mock/src/graphql/users.ts
```diff
@@ -32,5 +41,9 @@ export function createGetUserMock(opts: { user?: { id: string; uuid: string } })
       },
     },
   };
-  return [...cloneResponse(getUser, 8), ...cloneResponse(getUserId, 8)];
+  return [
+    ...cloneResponse(getUser, 8),
+    ...cloneResponse(getUserId, 8),
+    ...cloneResponse(dismissNag, 8),
+  ];
```
jasonLaster, line 48: Nice catch. Gosh there is a lot of good stuff in this PR

test/manifest.js
```diff
@@ -242,47 +242,55 @@ module.exports = [
...
   },
   {
     example: "doc_inspector_basic.html",
+    pending: true,
     script: "highlighter.js",
     targets: ["gecko"],
   },
   {
     example: "doc_stacking.html",
+    pending: true,
```
jasonLaster, line 293: This is a lot of pending tests, but as long as we revisit before shipping the inspector that is okay

src/ui/reducers/layout.ts
```diff
@@ -27,35 +27,39 @@ const getDefaultSelectedPrimaryPanel = (session: any, recording?: Recording) =>
...
     return {
       ...syncInitialLayoutState,
+      viewMode: initialViewMode,
       selectedPrimaryPanel: getDefaultSelectedPrimaryPanel(session, recording),
+      showVideoPanel: "showVideoPanel" in session ? session.showVideoPanel : showVideoPanel,
+      showEditor: "showEditor" in session ? session.showEditor : showEditor,
     };
+  } catch (e) {
+    return syncInitialLayoutState;
   }
```
jasonLaster, line 62: Why is this try/catch needed? It's scary that this could occasionally fail and we wouldnt know why

jasonLaster (commented):
One important question. Otherwise looks shippable!

test/mock/src/runTest.ts
jcmorrow: Nope, good catch!

.github/workflows/test.yml
```diff
@@ -53,129 +38,106 @@ jobs:
...
+    name: Wait for Vercel Preview Branch
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/download-artifact@v2
-      - uses: ./.github/actions/setup-next
-      - uses: ./.github/actions/test
+      - name: Waiting for 200 from the Vercel Preview
+        uses: patrickedqvist/wait-for-vercel-preview@v1.1.1
+        id: waitFor200
```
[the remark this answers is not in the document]
jcmorrow: Yeah, so they are slightly different takes on the problem. I never actually saw `deployment_status` work, but my understanding of that approach is:

```
-> PR gets opened
-> Build gets initiated
-> All of the initial steps are performed
-> All of the playwright tests are skipped (because the deployment isn't done yet)
-> The build finishes as passing because most of the things that would cause it to fail are skipped
... Some minutes go by...
-> Vercel finished the deployment
-> All of the jobs run again
```

This flow is:

```
-> PR gets opened
-> Build gets initiated
-> All of the initial steps are performed
-> All of the playwright tests wait until the deployment returns 200
-> All tests get run
-> The overall build is passed/failed depending on tests
```

The first approach seems kind of wasteful? Like, it basically does a bunch of stuff and then says "ya know what, this thing isn't actually ready, let's cancel and do this later. Whereas the second approach runs everything once, and waits at the correct times. I would be ok with the first process if we amended it so that it *failed* the job because the vercel deployment wasn't ready yet, and that was the first check that got run, so that we didn't waste additional CI minutes, but it also seems like that would generate a lot of noisy failures. What do you think?

src/ui/reducers/layout.ts
jcmorrow: You're right, I can make the `try/catch` much smaller.

test/manifest.js
jcmorrow: Yeah. This one is *so* weird. More details here: https://www.notion.so/replayio/Elements-Panel-Doesn-t-Load-in-Test-1a7e7f06335b421c87421005eecb96ef

PR #4917, opened 2021-12-27
Always show the frame timeline fixes #4359

### LINKED ISSUE #4359: Frame progress bar is blinking out of view when used
Here's a Loom demonstrating the effect:
https://www.loom.com/share/eb8318f28eed4d61a37d7518126e016e

Expected:
The slider control should be there at all times.

Actual:
When you're done dragging the gripper, the entire div disappears for a second while it reloads.

jasonLaster:
* Always show the frame timeline
* Fix react issue with <p>


jaril (approved):
LGTM — just re-running tests to make sure it's all gucci.

PR #4918, opened 2021-12-28
Use google's favicon api for fetching favicons

jaril:
We used to use https://faviconkit.com/ for displaying favicons in the privacy panel where we have a list of domains. That API's no longer maintained, so this switches us to using Google's favicon api instead in the meantime.

### DIFF
--- a/src/ui/components/UploadScreen/Privacy.tsx
+++ b/src/ui/components/UploadScreen/Privacy.tsx
@@ -44,13 +44,13 @@ export function ToggleShowPrivacyButton({
 
 function FavIcon({ url }: { url: string }) {
   return (
-    <div className="h-5 w-5 relative">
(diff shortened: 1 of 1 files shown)


jcmorrow:
Something doesn't look right to me

CleanShot 2021-12-28 at 11 35 14@2x

`stripe.com` definitely has a favicon. When I inspect that logo it says it's loading `https://www.google.com/s2/favicons?domain=url12.com` which seems wrong?

jaril:
> Something doesn't look right to me
> 
> CleanShot 2021-12-28 at 11 35 14@2x
> 
> `stripe.com` definitely has a favicon. When I inspect that logo it says it's loading `https://www.google.com/s2/favicons?domain=url12.com` which seems wrong?

Added the wrong branch — it's good now!

jasonLaster (approved):

PR #4920, opened 2021-12-28
Add console drawer

### LINKED ISSUE #4364: Console filter drawer
**Description:**

We don't currently have an empty state which is more interesting than it sounds for a couple of reasons. First, it can be an opportunity to show the user that they are filtering the messages either with the filter query or by hiding logs, errors, warnings. Also, it's an opportunity to teach users to add print statements.

jaril:
Fix #4364.

https://user-images.githubusercontent.com/15959269/147586891-ae49116b-4948-49d9-b855-af2d74b0c011.mov

This PR:
- Moves most of the filter bar (event listeners, console filters, console settings) to a collapsible filter drawer.
- Simplifies the console text search and the filter bar.


jasonLaster (approved):
Looks great. Also appreciate a bunch of the code cleanup as well.

PR #4921, opened 2021-12-28
Use graphql api for user data instead of auth0

ryanjduffy:
## Issue

We're currently pulling email, name, and picture from auth0 to display in the footer and in the user settings modal but this data isn't available for API key based auth or the upcoming Replay browser-injected auth.

## Resolution

* Add `name` and `picture` to the user info graphql query
* Use the user info query instead of auth0 for those data elements


jasonLaster (approved):
It's nice that we're simplifying this stuff

PR #4941, opened 2022-01-05
Make sure the document title is updated when renaming a replay

### LINKED ISSUE #4481: Editing the title of a recording should update the window title
Here for example I edited the title to add on

>  with some new title text added on

after the existing title, but the FF tab itself still reflects the old title

Screen Shot 2021-11-12 at 4 11 47 PM

jaril:
Fix #4481.


jcmorrow (approved):
Nice 👍

PR #4946, opened 2022-01-05
Remove old jest tests

jasonLaster:
These jest tests have never been run. They are also using some of the redux helpers we'd like to modernize, so it is safer to remove them and then use the firefox test helpers as a reference when we want to re-create a unit test suite


jcmorrow (approved):
🔥 🔥 🔥

PR #4949, opened 2022-01-05
Hide the splitter if the pane is not split

jcmorrow:
The draggable splitter is only useful if both panels are non-null and open.

Fixes https://github.com/RecordReplay/devtools/issues/4936

### DIFF
--- a/src/devtools/packages/devtools-splitter/src/SplitBox.js
+++ b/src/devtools/packages/devtools-splitter/src/SplitBox.js
@@ -162,13 +162,8 @@ class SplitBox extends Component {
   // Rendering
   preparePanelStyles() {
     const vert = this.state.vert;
-    const {
-      minSize,
-      maxSize,
-      startPanelCollapsed,
-      endPanelControl,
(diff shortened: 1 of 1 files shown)


src/devtools/packages/devtools-splitter/src/SplitBox.js
```diff
@@ -246,13 +241,15 @@ class SplitBox extends Component {
...
         : null,
-      Draggable({
-        className: "splitter",
-        style: splitterStyle,
-        onStart: this.onStartMove,
-        onStop: this.onStopMove,
-        onMove: this.onMove,
-      }),
+      !startPanelCollapsed && !endPanelCollapsed && startPanel && endPanel
+        ? Draggable({
```
jasonLaster, line 245: this is a style thing, but I'd probably extract to a variable

`shouldShowDraggable ? Draggable`

jcmorrow: I think I made it better 😅

PR #4950, opened 2022-01-06
Adding custom icon component

jonbell-lot23:
**Background**
We have a component for loading Material icons, but it's not extensible to custom icons. We sometimes need custom icons, so I asked Jaril to make an Icon component and now it works!

**What next?**
In the short term, we just need this standardised component. Later on we can let the component work at different sizes or pull from SVG files, but we don't need that yet.

**Testing out some new icons**
image

I updated these icons to test the component. As a bonus, I think they look much better. Material icons are very thick and chunky, whereas these are more elegant and light.


jonbell-lot23:
Update: we should upgrade to a clearer approach (CSS) as described by Josh here: https://discord.com/channels/779097926135054346/795737681589174312/928820388744540263

jonbell-lot23:
Update: we should roll this back in favour of a clearer approach as described by Josh here: https://discord.com/channels/779097926135054346/795737681589174312/928820388744540263

PR #4955, opened 2022-01-06
Aligning avatar to other icons

jonbell-lot23:
Our profile avatar was too large, which meant it was misaligned with our other icons.

Old:
image

New:
image


jcmorrow (approved):
Nice find ✏️

PR #4957, opened 2022-01-06
Add basic social cards for public replays

ryanjduffy:
## Issue

We'd like to have more engaging embeds when sharing replay links

## Resolution

Adds the necessary `<meta>` tags to support richer embeds for the various social platforms

## Notes

This is a hacky implementation because the proper implementation was taking a long time. So instead of properly server-side rendering the recording page in full, I've added a special `headOnly` prop to the SSR recording page which will return only the `<Head>` element on the server. Further, this hack matches on the route within the `_app` and injects it early because there are several components that all hijack the page rendering if they don't have everything necessary for them to render.

We'll need to address this collection of loading states (and ideally consolidate them) but that is more intrusive so I've elected to use a low impact hack to get the value out the door and we can return to the proper arch later.

PS - In the course of trying to unwind the loading state issues, I noticed some assumptions in other layers that will need to be addressed as well. couple examples:
* the apollo client assumes that it won't be rendered until the auth token has been received. both of these (token manage and apollo provider) are layers around the app and page component so decoupling this relationship will be necessary.
* we're blocking rendering of the app until the redux store is boostrapped but that bootstrapping process is more than "just make the initial store" and instead includes a bunch of logic that could probably be deferred and come in separately.

In short, we need to look through all the places that block rendering the page and clean them up so that the page is making the decision if it can render itself or a consolidated loading page rather than various app-level components blocking rendering as the control passes through them.

## Media

image

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -4,7 +4,7 @@ import "ui/utils/whatwg-url-fix";
 import Head from "next/head";
 import type { AppContext, AppProps } from "next/app";
 import NextApp from "next/app";
-import React, { ReactNode, useEffect, useState } from "react";
+import React, { ReactNode, useEffect, useMemo, useState } from "react";
 import { Provider } from "react-redux";
 import { IntercomProvider } from "react-use-intercom";
 import tokenManager from "ui/utils/tokenManager";
@@ -12,7 +12,7 @@ import { ApolloWrapper } from "ui/utils/apolloClient";
 import LoadingScreen, { StaticLoadingScreen } from "ui/components/shared/LoadingScreen";
 import ErrorBoundary from "ui/components/ErrorBoundary";
 import _App from "ui/components/App";
-import { bootstrapApp, bootstrapDefaultStore } from "ui/setup";
+import { bootstrapApp, initStore } from "ui/setup";
 import "image/image.css";
 import { Store } from "redux";
 import { ConfirmProvider } from "ui/components/shared/Confirm";
@@ -149,16 +149,16 @@ function AppUtilities({ children, apiKey }: { children: ReactNode } & AuthProps)
   );
 }
 function Routing({ Component, pageProps }: AppProps) {
-  const [store, setStore] = useState<Store>(bootstrapDefaultStore());
-  useEffect(() => {
-    bootstrapApp().then((store: Store) => setStore(store));
+  const store = useMemo(() => {
+    const s = initStore();
+    bootstrapApp(s);
+
+    return s;
   }, []);
 
   if (maintenanceMode) {
     return <MaintenanceModeScreen />;
   }
-
-  console.log(">>> app");
 
   return (
     <Provider store={store}>
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -4,12 +4,12 @@ import "ui/utils/whatwg-url-fix";
 import Head from "next/head";
 import type { AppContext, AppProps } from "next/app";
 import NextApp from "next/app";
-import React, { ReactNode, useEffect, useMemo, useState } from "react";
+import React, { ReactNode, useEffect, useRef } from "react";
 import { Provider } from "react-redux";
 import { IntercomProvider } from "react-use-intercom";
 import tokenManager from "ui/utils/tokenManager";
 import { ApolloWrapper } from "ui/utils/apolloClient";
-import LoadingScreen, { StaticLoadingScreen } from "ui/components/shared/LoadingScreen";
+import LoadingScreen from "ui/components/shared/LoadingScreen";
 import ErrorBoundary from "ui/components/ErrorBoundary";
 import _App from "ui/components/App";
 import { bootstrapApp, initStore } from "ui/setup";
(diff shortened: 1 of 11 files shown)


jasonLaster (approved):
Oh snap. Thank you nextjs

pages/recording/[id].tsx
```diff
@@ -127,14 +110,49 @@ export async function getStaticProps({ params }: { params: { id: string } }) {
...
+    title?: string;
+    url?: string;
+    duration?: number;
+    image?: string;
+  };
+}
 
-export default connector(RecordingPage);
+export default function SSRRecordingPage({ metadata, ssrLoading }: SSRProps) {
+  console.log(">>>> metadata");
```
Comment (jasonLaster) [inline pages/recording/[id].tsx]: 🧂

pages/recording/[id].tsx
```diff
@@ -127,14 +110,49 @@ export async function getStaticProps({ params }: { params: { id: string } }) {
...
+      <meta name="twitter:site" content="@replayio" />
+      <meta
+        property="twitter:description"
+        content={`${Math.round((metadata.duration || 0) / 1000)} second replay`}
+      />
+    </Head>
+  ) : null;
+
+  if (ssrLoading) {
+    console.log("ssr head", metadata.title);
```
Comment (jasonLaster) [inline pages/recording/[id].tsx]: 🧂

src/ui/utils/apolloClient.tsx
```diff
@@ -53,9 +53,9 @@ export function ApolloWrapper({ children }: { children: ReactNode }) {
     );
   }
 
-  if (!isTest() && loading) {
-    return <StaticLoadingScreen />;
-  }
+  // if (!isTest() && loading) {
+  //   return <StaticLoadingScreen />;
+  // }
```
jasonLaster, line 58: ?

ryanjduffy: We have several places that hijack the flow to show loading screens that prevent the recording page SSR page from running. I'm commenting them out for now while this PR is in draft to get something running.

pages/recording/[id].tsx
```diff
@@ -34,11 +98,41 @@ function Recording({ getAccessibleRecording }: PropsFromRedux) {
...
   }
 }
 
 const connector = connect(null, { getAccessibleRecording });
 type PropsFromRedux = ConnectedProps<typeof connector>;
 
-export default connector(Recording);
+const ConnectedRecordingPage = connector(RecordingPage);
+
+export async function getStaticProps({ params }: { params: { id: string } }) {
```
Comment (abstractalgo) [inline pages/recording/[id].tsx:115]: Might be better to use the correct type https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation for context and not raw type.

pages/recording/[id].tsx
```diff
@@ -1,15 +1,74 @@
...
+
+  return (
+    <Head>
+      {/* nosemgrep typescript.react.security.audit.reac-http-leak.react-http-leak */}
+      <meta property="og:title" content={title} />
+      {/* nosemgrep typescript.react.security.audit.reac-http-leak.react-http-leak */}
+      <meta property="og:description" content={description} />
+      {/* nosemgrep typescript.react.security.audit.reac-http-leak.react-http-leak */}
+      <meta property="og:image" content={image} />
+      <meta name="twitter:card" content={"summary_large_image"} />
```
Comment (abstractalgo) [inline pages/recording/[id].tsx:56]: Can remove curly brackets (unless some formatter adds it).

pages/recording/[id].tsx
```diff
@@ -1,15 +1,74 @@
...
+      {/* nosemgrep typescript.react.security.audit.reac-http-leak.react-http-leak */}
+      <meta property="twitter:image" content={image} />
+      {/* nosemgrep typescript.react.security.audit.reac-http-leak.react-http-leak */}
+      <meta property="twitter:title" content={title} />
+      <meta name="twitter:site" content="@replayio" />
```
Comment (abstractalgo) [inline pages/recording/[id].tsx:61]: Might be good to extract Twitter handle to somewhere if reused.

pages/_app.tsx
```diff
@@ -186,7 +192,20 @@ function Routing({ Component, pageProps }: AppProps) {
...
 
 const App = ({ apiKey, ...props }: AppProps & AuthProps) => {
-  return <AppUtilities apiKey={apiKey}>{<Routing {...props} />}</AppUtilities>;
+  const router = useRouter();
+  let head: React.ReactNode;
+
+  // HACK: Coordinates with the recording page to render its <head> contents for
+  // social meta tags. This can be removed once we are able to handle SSR
+  // properly all the way to the pages.
+  if (props.__N_SSG && router.pathname.match(/^\/recording\//)) {
```
abstractalgo, line 201: Could use adding info in the comment about what is `__N_SSG`.

src/ui/hooks/recordings.ts
```diff
@@ -734,3 +734,48 @@ export function useUpdateRecordingTitle() {
...
+      variables: {
+        recordingId: id,
+      },
+    }),
+  });
+
+  const json: {
+    data: { recording: Recording & { owner?: { name: string } } };
+    error: any;
+  } = await resp.json();
```
jcmorrow, line 772: I think we need to gracefully handle errors here. Right now the preview branch crashes on private recordings: https://devtools-ewr79cidf-recordreplay.vercel.app/recording/823241b9-54fd-4cc4-b315-3937694c0b0c

abstractalgo (approved):
I'm pretty terrified of this solution - way too many things to keep in mind, and to later redo.
But considering the obvious value that we get from having this in place, and the obvious challenges that we cannot overcome currently, I guess it's fine. 🙃

jcmorrow (commented):
One problem with private recordings, otherwise it's really cool that this is working! I can't believe how many hoops there are to jump through with doing SSR open graph stuff, well done! 🍰

src/ui/hooks/recordings.ts
ryanjduffy: I missed a console error on this. Oddly, returning `null` a couple lines down fixes the SSR handling of private replays.

ryanjduffy: Sharing the working link now to close the loop: https://devtools-git-fork-ryanjduffy-ryan-social-cards-recordreplay.vercel.app/recording/823241b9-54fd-4cc4-b315-3937694c0b0c

PR #4962, opened 2022-01-07
Pull tip index out of redux

ryanjduffy:
Saw a comment about not showing tips in one of the loading states because the store was initialized yet. There doesn't seem to be a compelling reason to use the store for this state data since it's localized to a given instance of the tips so this removes that requirement so tips can be shown on that loading screen as well. Seems like a good simplification and decoupling but open to discuss!

I didn't enable the tips on that block yet but we can tack that on to this PR if it is desired.

https://github.com/RecordReplay/devtools/blob/0bab84257e3e20fb7452fbc558f722e3aeb4edf6/pages/_app.tsx#L157-L161

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -93,9 +93,6 @@ export type SetRecordingWorkspaceAction = Action<"set_recording_workspace"> & {
 export type SetLoadedRegions = Action<"set_loaded_regions"> & {
   parameters: loadedRegions;
 };
-export type SetLoadingPageTipIndexAction = Action<"set_loading_page_tip_index"> & {
-  index: number;
-};
 
 export type SetMouseTargetsLoading = Action<"mouse_targets_loading"> & {
   loading: boolean;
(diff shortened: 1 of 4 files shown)


jaril:
The original reason for keeping it in Redux was because the loading screen can be rendered multiple times from different components, and so keeping the tipIndex in the component's state would cause it to keep jumping back to 0.

jcmorrow:
> The original reason for keeping it in Redux was because the loading screen can be rendered multiple times from different components, and so keeping the tipIndex in the component's state would cause it to keep jumping back to 0.

Would it be worth randomizing the starting index to fix this instead? I feel like that's how most video game loading screens do tips and it's like 🤷

ryanjduffy:
That'd could be fine. I do worry that there will still be some wonkiness that because there are several loading screens the tips will show for a seemingly random amount of time before switching.

ryanjduffy:
Probably better to bail on this and fix the loading screen story but that'll be messy. I was hoping for a quick win to unblock the visual changes Jon wanted to do.

jcmorrow:
> That'd could be fine. I do worry that there will still be some wonkiness that because there are several loading screens the tips will show for a seemingly random amount of time before switching.

Honestly I'm totally fine with this, but I leave it up to you!

Issue #4952, opened 2022-01-06
Command Palette 1.2


jonbell-lot23:
**Background**
We have a vision for how we want Command Palette to work, and we've been breaking the work into chunks. This is our next chunk.

<img width="758" alt="image" src="https://user-images.githubusercontent.com/9154902/148313792-b758f6ba-e984-49cf-afc8-fc5509605f05.png">

**Changes for 1.2**

(We discussed these changes over several weeks and had hours of conversation about it, so I hope this doesn't take anyone by surprise now that we're back from the holiday break!)

- [x] Replace the dashboard with a deployed command palette.
- [x] Add a fixed tab icon so you can always get back to the command palette
- [x] Default the command palette to these four options
- [x] Add the keyboard shortcut visual as shown in this screenshot
- [x] Add all the "add it" features [in this table](https://www.notion.so/replayio/ea1e85a8ad6d46b88bb315a1cf337243?v=8ea1fd7ed2bf4e0dbbd7cd1c4bbecc37)

**Next steps after these go live**
- Keyboard shortcuts
- Icons
- More actions

PR #4963, opened 2022-01-07
Command Palette follow-up items

jaril:
Takes care of the first few items of #4952:

- Default the command palette to these four options (Open file, search functions, search text, open sources)
- Add the keyboard shortcut visual as shown in this screenshot
- Add all the "add it" features in this table


src/devtools/client/debugger/src/actions/ast.js
```diff
@@ -0,0 +1,39 @@
+import { ThreadFront } from "protocol/thread";
+import { getSourceIDsToSearch } from "devtools/client/debugger/src/utils/source";
+import { formatProjectFunctions } from "../utils/quick-open";
+import { getSources } from "../reducers/sources";
+import {
+  getGlobalFunctions,
+  isGlobalFunctionsLoading,
+} from "devtools/client/debugger/src/reducers/ast";
+
+export function loadGlobalFunctions() {
```
jaril, line 10: Separated this out since otherwise there's a circular dependency.

src/image/image.css
```diff
@@ -490,6 +490,12 @@ button.open-inspector {
 .img.loom {
   -webkit-mask-image: url(/images/Loom.svg);
 }
+.img.k-icon {
+  -webkit-mask-image: url(/images/k-icon.svg);
+}
+.img.cmd-icon {
+  -webkit-mask-image: url(/images/cmd-icon.svg);
+}
```
jaril, line 498: This is the original way that we used to import images to devtools. Probably should based our new Icon implementation off of it.

src/image/image.css
```diff
@@ -490,6 +490,12 @@ button.open-inspector {
 .img.loom {
   -webkit-mask-image: url(/images/Loom.svg);
 }
+.img.k-icon {
+  -webkit-mask-image: url(/images/k-icon.svg);
+}
+.img.cmd-icon {
+  -webkit-mask-image: url(/images/cmd-icon.svg);
```
jasonLaster, line 497: should we use `mask-image`

src/ui/actions/app.ts
```diff
@@ -369,29 +363,44 @@ export function setLoadingPageTipIndex(index: number): SetLoadingPageTipIndexAct
...
+    } else if (key === "open_file_search") {
+      dispatch(openQuickOpen());
     } else if (key === "open_full_text_search") {
       dispatch(setViewMode("dev"));
       dispatch(setSelectedPrimaryPanel("search"));
-    } else if (key === "open_sources" || key === "open_outline") {
-      dispatch(setViewMode("dev"));
-      dispatch(setSelectedPrimaryPanel("explorer"));
+    } else if (key === "open_function_search") {
+      dispatch(openQuickOpen("@", true));
```
jasonLaster, line 381: we should probably switch to dev here too and a bunch of others as well.

src/ui/components/CommandPalette/CommandPalette.tsx
```diff
@@ -1,38 +1,118 @@
...
   { key: "show_sharing", label: "Show Sharing Options" },
 ];
 
-function CommandPalette({ hideCommandPalette, executeCommand }: PropsFromRedux) {
+const DEFAULT_COMMANDS: CommandKey[] = [
+  "open_file_search",
+  "open_function_search",
+  "open_full_text_search",
+  "open_sources",
+];
```
jasonLaster, line 69: love bumping these to the top for the default, but I would still show the others below. Being able to scroll through the full list is a nice feature

jasonLaster (approved):
Works really well and looks great.

src/ui/components/CommandPalette/CommandPalette.tsx
jasonLaster: In fact, keeping these top results on the top is important so that there isn't as big of a jump when you start searching.

jonbell-lot23: I misgivings about putting the whole list there, because it's important that it feel curated and high importance in its default state. Otherwise you run into [Hick's Law](https://lawsofux.com/hicks-law/), where more options slows people down and overwhelms them.

That said, I think it's great to have the whole list appear anywhere else you hit command-K. That's the moment to go big on all our amazing options. We could even say when you press command-K here, the whole list appears.

jasonLaster: Are you referring to when it's open by default in the empty editor? I can see how that would be a bit different than the typical command palette experience where you want to optimize for discoverability. See [linear](https://blog.superhuman.com/how-to-build-a-remarkable-command-palette/)

I think a good way to have it both ways is to add some visual separation between the top 4 results and the others. This way you signal, we think these options are great, the others are nice too. This could work in both contexts and keep us consistent

jonbell-lot23: I could go for that -- but would the second list be *all* options or another group of cool things?

I think all would be a bit much. But we could try it.

jonbell-lot23: Actually, on second thought ...

I don't really mind since we're cutting things off after the first four. My concern was if the region stretched forever, but my concern goes away if we crop. Never mind!

PR #4964, opened 2022-01-08
Add response bodies to the Network Monitor

jcmorrow:
Changes
---
- Upgrade protocol to use the latest version which supports response
  bodies
- Add fetchers for requesting response bodies from the backend
- Store those, paying close attention to the ordering and deduping,
  because multiple parts might be received out of order
- Add an HttpBody component. Notice this is somewhat generalized,
  because we'll be reusing it for request bodies

- Add Storybook stories for the JSON, plaintext, and multipart scenarios

Replay: https://app.replay.io/recording/7771349a-993b-4faf-9f1b-6ff13386b3b2#


src/ui/components/NetworkMonitor/ResponseBody.tsx
```diff
@@ -0,0 +1,54 @@
...
+}: {
+  request?: RequestSummary;
+  responseBodyParts: ResponseBodyData[];
+}) => {
+  if (!request || !responseBodyParts?.length) {
+    return null;
+  }
+
+  const contentLength = request.requestHeaders.find(h => h.name.toLowerCase() === "content-length");
+  const [validJSON, content] = safeJSONParseBase64(responseBodyParts.map(p => p.value).join(""));
```
loganfsmyth, line 31: There's absolutely no guarantee that these parts will be in order so this is a bit dangerous. You also can't concatenate base64 values together like this. You'll need to `atob` each one separately. See my next comment for more info.

src/ui/components/NetworkMonitor/ResponseBody.tsx
```diff
@@ -0,0 +1,54 @@
...
+      ) : (
+        <>
+          <div
+            className="flex items-center py-1 cursor-pointer font-bold"
+            onClick={() => setExpanded(!expanded)}
+          >
+            <TriangleToggle open={expanded} />
+            Response body:
+          </div>
+          {expanded && <div>{validJSON ? <ReactJson src={content} /> : <pre>{content}</pre>}</div>}
```
loganfsmyth, line 47: This pre will be showing a bunch of individual base64 pieces, so we probably don't want to render it like this. You can decode each individual part value with https://stackoverflow.com/a/21797381/785065 to get an ArrayBuffer, Then I'd vote to check `Content-Type` to see if it start with `text/` then pass the data to https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder to get the actual text to display. 

If the data _isn't_ `text/*` then we should offer a download link for the URL `const url = URL.createObjectURL(new Blob([arraybuffer], { type: contentType }));`. Note that we should only create the URL temporarily then call `URL.revokeObjectURL` after or else this will leak.

src/ui/components/NetworkMonitor/ResponseBody.tsx
```diff
@@ -0,0 +1,54 @@
...
+  responseBodyParts,
+}: {
+  request?: RequestSummary;
+  responseBodyParts: ResponseBodyData[];
+}) => {
+  if (!request || !responseBodyParts?.length) {
+    return null;
+  }
+
+  const contentLength = request.requestHeaders.find(h => h.name.toLowerCase() === "content-length");
```
loganfsmyth, line 30: This header is the length of the request, not the response.

src/ui/components/NetworkMonitor/ResponseBody.tsx
```diff
@@ -0,0 +1,54 @@
...
+  try {
+    return [true, JSON.parse(atob(maybeJSON))];
+  } catch (e) {
+    console.log(maybeJSON);
+    console.log(e);
+    return [false, maybeJSON];
+  }
+};
+
+const ResponseBody = ({
```
loganfsmyth, line 19: We'll want to share this for request bodies soon so probably better to have a more generic name?

src/ui/reducers/network.ts
```diff
@@ -26,6 +28,15 @@ const update = (state: NetworkState = initialState(), action: NetworkAction): Ne
         events: [...action.payload.events, ...state.events],
         requests: [...action.payload.requests, ...state.requests],
       };
+    case "NEW_RESPONSE_BODY_PARTS":
+      console.log({ action, responsesValue: state.responses[action.payload.responseBodyParts.id] });
```
loganfsmyth, line 32: Leftover debug code.

src/ui/reducers/network.ts
```diff
@@ -26,6 +28,15 @@ const update = (state: NetworkState = initialState(), action: NetworkAction): Ne
...
+      return {
+        ...state,
+        responses: {
+          ...state.responses,
+          [action.payload.responseBodyParts.id]: action.payload.responseBodyParts.parts,
+        },
```
loganfsmyth, line 38: I'm somewhat worried we're going to end up crashing the user's tab by collecting all this data in the store, but maybe we can see what happens. Might be good to discard entries in here that haven't been used in a while.

src/stories/NetworkMonitor/ResponseBody.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+const request = requestSummary("1", "https://app.replay.io/graphql", 200);
+
+Basic.args = {
+  request,
+  responseBodyParts: [
+    {
+      length: 12,
+      offset: 0,
+      value: "Zm9vYmFyCg==",
+    },
```
loganfsmyth, line 24: Seems like a test with multiple parts is in order.

src/ui/components/NetworkMonitor/ResponseBody.tsx
jcmorrow: Doh!

src/ui/reducers/network.ts
jcmorrow: If we're only going to request at most 5MB with the change up above, we should be OK I think. We only load this when they go to the response tab, so they would have to go to a *lot* of response tabs to make a big memory dent.

src/ui/components/NetworkMonitor/ResponseBody.tsx
jcmorrow: This should be all taken care of now!

src/ui/components/NetworkMonitor/ResponseBody.tsx
jcmorrow: I haven't added the download link yet, but otherwise this is done. I think that should be a quick follow-up so we can focus on getting this and request bodies out.

src/stories/NetworkMonitor/ResponseBody.stories.tsx
jcmorrow: Done!

jasonLaster (approved):
Looks good!

PR #4966, opened 2022-01-08
Switch Quick Open to text-md

jasonLaster:
Using text-md is more consistent with the other text sizes in the app.

Screen Shot 2022-01-08 at 8 18 31 AM


jcmorrow:
It seems like there are a lot more changes than the text size here, is that all intended or was this based off of another branch?

jasonLaster:
this already landed

PR #4970, opened 2022-01-08
Small usability improvements to the net monitor

jasonLaster:
* query params are formatted like firefox
* details row hovered
* cursor default
* status is on the left so that the name isnt overlayed by the fast forward button
* js and html request types are now selected by default


jcmorrow:
LGTM 🥘

PR #4977, opened 2022-01-10
Add support for enterprise contracts in team billing page

ryanjduffy:
## Issue

We have an enterprise contract but need to do a touch of clean up in the team settings UI so we can display something

## Resolution

* Adds a new metadata entry for `ent-v1` plans
* Hides the "add payment method" link for `contract` plans since the payment will be managed outside of stripe.

## Media

https://app.replay.io/recording/ce9bc0e7-89d2-4ced-980e-79c52d248df2#

image


jcmorrow (approved):
LGTM 💳

PR #4984, opened 2022-01-11
Distinguish between hovered row and button

### LINKED ISSUE #4873: Net Monitor fast forward button
There are two visual issues with the fast forward button

1. the button overlaps the name of the request. Maybe we need to have something to the left of the name like status code that is less important
2. the fast forward button bleeds into the selected row

---
Screen Shot 2021-12-17 at 7 47 40 PM

---
Screen Shot 2021-12-17 at 7 47 51 PM
Screen Shot 2021-12-17 at 7 47 45 PM

jcmorrow:
We were sort of losing the distinction between the hovered network monitor row when it was selected and the fast-forward/rewind button that appears over top of it. Adding a slight color distinction and a box shadow helps a lot.

Fixes #4873

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestRow.tsx
+++ b/src/ui/components/NetworkMonitor/RequestRow.tsx
@@ -34,7 +34,7 @@ export const RequestRow = ({
       <div {...row.getRowProps()}>
         {row.original.triggerPoint && row.original.triggerPoint.time !== currentTime && (
           <div
-            className={classNames(styles.seekBadge)}
+            className={classNames(styles.seekBadge, "shadow-md")}
             onClick={() => {
               if (!row.original.triggerPoint) {
                 return;
--- a/src/ui/components/NetworkMonitor/RequestTable.module.css
+++ b/src/ui/components/NetworkMonitor/RequestTable.module.css
@@ -16,8 +16,8 @@ table.requests {
 }
 
 .row.selected {
-  background: var(--primary-accent);
-  border-color: var(--primary-accent);
+  background: var(--primary-accent-hover);
+  border-color: var(--primary-accent-hover);
   color: white;
 }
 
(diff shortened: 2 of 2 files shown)


src/ui/components/NetworkMonitor/RequestRow.tsx
```diff
@@ -43,7 +43,7 @@ export const RequestRow = ({
             }}
           >
             <div
-              className={classNames("img", {
+              className={classNames("img shadow-md", {
```
ryanjduffy, line 46: I think `shadow-md` needs to move up a level to show the shadow around the whole badge and not just the text.

ryanjduffy (changes requested):
The bg color change looks good. There's also a small gap to the left of the badge that'd be good to fix up.

image

jcmorrow:
> The bg color change looks good. There's also a small gap to the left of the badge that'd be good to fix up.
> 
> image

Good call, this comes from the border change a little while ago.

src/ui/components/NetworkMonitor/RequestRow.tsx
jcmorrow: I think I'm confused here. Did I accidentally add a text-shadow somewhere instead of a box-shadow?

ryanjduffy: No. this is a box shadow but it's on the icon and not the badge. My screenshot above shows it on the badge because i manually added it to verify but I didn't see the shadow in the original PR. Do you see the same?

jcmorrow: Ahhhhh, I see what you're saying. I thought I saw it but it might have just been an illusion of the lighter on darker color

jcmorrow: Yep, that's what it was. Thanks Ryan!

PR #4988, opened 2022-01-11
Add a canonical request type

jcmorrow:
I hope this makes sense. When we get requests, they can have a variety of types. But when we present filtering options to the user, we usually want to show rougher groupings. I have added a new type called `CanonicalRequestType`, which reflects what group we lump that kind of request into. We use those values to let the user filter requests. I think this simplifies the problem a little bit, because all of the logic goes into one place, rather than having to do one-off checks (like we were doing for `Fetch/XHR`) for categories that contain multiple smaller types.

### DIFF
--- a/src/ui/components/NetworkMonitor/FilterBar.tsx
+++ b/src/ui/components/NetworkMonitor/FilterBar.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import MaterialIcon from "../shared/MaterialIcon";
 import TypesDropdown from "./TypesDropdown";
-import { RequestType } from "./utils";
+import { CanonicalRequestType } from "./utils";
 
 export default function FilterBar({
   table,
@@ -9,8 +9,8 @@ export default function FilterBar({
   types,
 }: {
   table: any;
-  toggleType: (type: RequestType) => void;
-  types: Set<RequestType>;
+  toggleType: (type: CanonicalRequestType) => void;
+  types: Set<CanonicalRequestType>;
 }) {
   return (
     <div className="bg-white flex items-center px-1 py-1">
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -13,7 +13,7 @@ import { getCurrentTime } from "ui/reducers/timeline";
 import { UIState } from "ui/state";
 import RequestDetails from "./RequestDetails";
 import RequestTable from "./RequestTable";
-import { RequestSummary, RequestType } from "./utils";
+import { CanonicalRequestType, RequestSummary } from "./utils";
 import FilterBar from "./FilterBar";
 import Table from "./Table";
 import { fetchFrames, fetchResponseBody, fetchRequestBody } from "ui/actions/network";
@@ -32,20 +32,21 @@ export const NetworkMonitor = ({
   selectFrame,
 }: PropsFromRedux) => {
   const [selectedRequest, setSelectedRequest] = useState<RequestSummary>();
-  const [types, setTypes] = useState<Set<RequestType>>(new Set(["xhr", "javascript", "html"]));
+  const [types, setTypes] = useState<Set<CanonicalRequestType>>(new Set([]));
   const [vert, setVert] = useState<boolean>(false);
 
   const container = useRef<HTMLDivElement>(null);
 
   const closePanel = () => setSelectedRequest(undefined);
 
-  const toggleType = (type: RequestType) => {
-    if (types.has(type)) {
-      types.delete(type);
+  const toggleType = (type: CanonicalRequestType) => {
+    const newTypes = new Set(types);
+    if (newTypes.has(type)) {
+      newTypes.delete(type);
     } else {
-      types.add(type);
+      newTypes.add(type);
     }
-    setTypes(new Set(types));
+    setTypes(newTypes);
   };
 
   let resizeObserver = useRef(
--- a/src/ui/components/NetworkMonitor/utils.ts
+++ b/src/ui/components/NetworkMonitor/utils.ts
@@ -11,7 +11,19 @@ import {
 import keyBy from "lodash/keyBy";
 import { compareNumericStrings } from "protocol/utils";
 
-export type ContentType = "json" | "text" | "other";
+export enum CanonicalRequestType {
+  CSS,
+  FETCH_XHR,
+  FONT,
+  HTML,
+  IMAGE,
+  JAVASCRIPT,
+  MANIFEST,
+  MEDIA,
+  OTHER,
+  WASM,
+  WEBSOCKET,
+}
 
 export type RequestSummary = {
   domain: string;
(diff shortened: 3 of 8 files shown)


src/ui/components/NetworkMonitor/utils.ts
```diff
@@ -149,20 +167,14 @@ export const partialRequestsToCompleteSummaries = (
         return true;
       }
 
-      if (types.has(row.type)) {
-        return true;
-      }
-      if (types.has("xhr") && row.type.match(/json/)) {
-        return true;
+      if (!row.cause) {
+        return false;
```
loganfsmyth, line 171: Instead of this, I think we should do

```
if (types.has(REQUEST_TYPES[row.cause || "other"])) {
```

loganfsmyth: Actually better yet
```
if (types.has(REQUEST_TYPES[row.cause] || REQUEST_TYPES.other)) {

File: src/ui/components/NetworkMonitor/utils.ts [hunk 2]
```diff
@@ -121,6 +138,7 @@ export const partialRequestsToCompleteSummaries = (
       const documentType = getDocumentType(response.event.responseHeaders);
       const type: RequestType = (documentType?.split("/")?.[1] || documentType) as RequestType;
       return {
+        cause: request.event.requestCause as RequestType,
```
loganfsmyth, line 141: Since we're casting this, this could introduce unexpected keys. Maybe it would be better to have this be
```
cause: REQUEST_TYPES.hasOwnProperty(request.event.requestCause) ? request.event.requestCause : "other",
```
then we don't need the null check below at all.

src/ui/components/NetworkMonitor/utils.ts
jcmorrow: Oh, this wasn't to handle `Other`, that should be handled already, this is the case where for some reason the cause is `undefined`, but I'm realizing now that is impossible and I should just put that in the type system.

src/ui/components/NetworkMonitor/utils.ts
jcmorrow: Yeah, I think I can write this line better now!

jcmorrow: Actually, yeah, because this type comes from protocol let's do it the ternary way 👍

src/ui/components/NetworkMonitor/utils.ts
```diff
@@ -33,39 +34,55 @@ export type RequestSummary = {
   url: string;
 };
 
+// From https://github.com/RecordReplay/gecko-dev/blob/webreplay-release/devtools/server/actors/replay/network-helpers.jsm#L14
 export const REQUEST_TYPES = {
+  document: "HTML",
```
loganfsmyth, line 39: Might be worth defining
```
enum CanonicalRequestType {
  html,
  // ...
};
```
so `REQUEST_TYPES` is a `Record<string, CanonicalRequestType>` lookup, and we can avoid relying on the human-readable string as the key in `REQUEST_ICONS`?

jcmorrow: Yep, good idea.

src/ui/components/NetworkMonitor/utils.ts
loganfsmyth: `request.event.requestCause` can be `undefined` so that's the main thing to keep in mind. The ternary should cover it though.

loganfsmyth: now this can be removed seems like.

src/ui/components/NetworkMonitor/utils.ts
```diff
@@ -30,42 +30,74 @@ export type RequestSummary = {
...
+  webManifest: "Manifest",
   media: "Media",
-  other: "Other",
+  script: "Javascript",
+  stylesheet: "CSS",
   wasm: "WASM",
   websocket: "Websocket",
 };
 
+export type RequestType = keyof typeof REQUEST_TYPES;
```
loganfsmyth, line 82: I'm gonna guess this won't be used anywhere once the types are exported as an enum.

src/ui/components/NetworkMonitor/FilterBar.tsx
```diff
@@ -1,16 +1,16 @@
 import React from "react";
 import MaterialIcon from "../shared/MaterialIcon";
 import TypesDropdown from "./TypesDropdown";
-import { RequestType } from "./utils";
+import { CanonicalRequestType, RequestType } from "./utils";
```
loganfsmyth, line 4: `RequestType` unused

src/ui/components/NetworkMonitor/utils.ts
```diff
@@ -30,42 +30,74 @@ export type RequestSummary = {
...
+  "Image",
+  "Javascript",
+  "Manifest",
+  "Media",
+  "Other",
+  "WASM",
+  "Websocket",
+] as const;
+
+export type CanonicalRequestType = typeof CANONICAL_REQUEST_TYPES[number];
```
loganfsmyth, line 51: To elaborate, I usually try to avoid UI strings being primary keys in object. This is also a slightly better fit for an enum because you can reuse the enum keys. Something for instance like
```
export enum CanonicalRequestType {
  CSS,
  XHR,
  FONT,
  HTML,
  IMAGE,
  JS,
  MANIFEST,
  MEDIA,
  OTHER,
  WASM,
  WS,
}

// I'm not exporting this anymore because `CanonicalRequestType` is usable for the dropdown.
const REQUEST_TYPES: Record<string, CanonicalRequestType> = {
  document: CanonicalRequestType.HTML,
  subdocument: CanonicalRequestType.HTML,
  objectSubdoc: CanonicalRequestType.HTML,

  // and so on
};

export const REQUEST_TYPE_INFO: Record<CanonicalRequestType, { label: string, icon: string }> = {
  [CanonicalRequestType.HTML]: { icon: "description", string: "HTML" },
  [CanonicalRequestType.JS]: { icon: "code", string: "JavaScript" },

  // and so on
};
```

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -32,14 +32,16 @@ export const NetworkMonitor = ({
   selectFrame,
 }: PropsFromRedux) => {
   const [selectedRequest, setSelectedRequest] = useState<RequestSummary>();
-  const [types, setTypes] = useState<Set<RequestType>>(new Set(["xhr", "javascript", "html"]));
+  const [types, setTypes] = useState<Set<CanonicalRequestType>>(
```
loganfsmyth, line 35: Also still very in favor of this being empty by default.

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -32,14 +32,16 @@ export const NetworkMonitor = ({
...
   const [vert, setVert] = useState<boolean>(false);
 
   const container = useRef<HTMLDivElement>(null);
 
   const closePanel = () => setSelectedRequest(undefined);
 
-  const toggleType = (type: RequestType) => {
+  const toggleType = (type: CanonicalRequestType) => {
     if (types.has(type)) {
       types.delete(type);
```
loganfsmyth, line 46: Ohh just spotted this, also a bit questionable to mutate the existing set. Better to do
```
const toggleType = (type: CanonicalRequestType) => {
  const newTypes = new Set(types);
  if (newTypes.has(type)) {
    newTypes.delete(type);
  } else {
    newTypes.add(type);
  }
  setTypes(newType);
};
```

loganfsmyth (approved):

PR #4989, opened 2022-01-12
Add a tab indicator to the sidebar, and add a badge beside pause info when paused

### LINKED ISSUE #4911: Toolbar badging
**Description:**

Today we use a solid fill to denote selected panes and paused. This can be confusing for users. 

* VS Code uses a left border to reinforce selected state. we use this pattern in other places too.
* We could badge pause state. This is one approach but there are others too. I don't love this partial icon because the grey bleeds. It'd probably be better to have something that overlays it or start with the left border and see if that is enough

https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/Replay-Design-Doc?node-id=9774%3A123571

Frame 298

---

### Two attempts at badging the pause icon

Frame 299
Frame 300

jaril:
image

Fix #4911.

### DIFF
--- a/src/ui/components/CommandPalette/CommandPaletteModal.tsx
+++ b/src/ui/components/CommandPalette/CommandPaletteModal.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import Modal from "../shared/NewModal";
 import { actions } from "ui/actions";
 import { useDispatch } from "react-redux";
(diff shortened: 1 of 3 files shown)


jonbell-lot23:
- Let's move from red to Replay fuschia
- Bottom right is non-standard but I think it works

Looks good!

Issue #4998, opened 2022-01-12
e2e test failures are not detected


hbenl:
The playwright-based e2e test runner currently fails to detect most test failures. Only tests that run for more than 30 seconds are reported as failures.

PR #4999, opened 2022-01-12
Fix the detection of failed e2e tests (#4998)

hbenl:


jcmorrow (approved):
Well this would explain why things have been passing more consistently 😆

PR #5002, opened 2022-01-12
Stop ESLint warnings

### LINKED ISSUE #4965: Missing dependencies in useEffect
```

./pages/callback.tsx
26:6  Warning: React Hook useEffect has missing dependencies: 'auth0', 'connection', and 'router'. Either include them or remove the dependency array.  react-hooks/exhaustive-deps

./pages/profile/[...view].tsx
19:6  Warning: React Hook useEffect has missing dependencies: 'isAuthenticated', 'replace', and 'setModal'. Either include them or remove the dependency array. If 'setModal' changes too often, find the parent component that defines it and wrap that definition in useCallback.  react-hooks/exhaustive-deps

./pages/recording/[id].tsx
25:6  Warning: React Hook useEffect has a missing dependency: 'getAccessibleRecording'. Either include it or remove the dependency array. If 'getAccessibleRecording' changes too often, find the parent component that defines it and wrap that definition in useCallback.  react-hooks/exhaustive-deps

./pages/settings.tsx
10:6  Warning: React Hook useEffect has a missing dependency: 'replace'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps

./pages/team/[...id].tsx
27:6  Warning: React Hook useEffect has missing dependencies: 'replace', 'setWorkspaceId', and 'updateDefaultWorkspace'. Either include them or remove the dependency array. If 'setWorkspaceId' changes too often, find the parent component that defines it and wrap that definition in useCallback.  react-hooks/exhaustive-deps
33:6  Warning: React Hook useEffect has a missing dependency: 'setModal'. Either include it or remove the dependency array. If 'setModal' changes too often, find the parent component that defines it and wrap that definition in useCallback.  react-hooks/exhaustive-deps
```

jcmorrow:
Fixes #4965 by taking the ESLint warning literally. I am pretty sure this is fine, although it's a little dumb because we really don't expect these functions to change. The problem here usually starts when calling one of these functions inside of the effect causes the prop to change, causing another call and on and on into an infinite loop.

@abstractalgo Any thoughts on the *correct* thing to do here (probably using redux hooks I'm guessing 😄 )

### DIFF
--- a/pages/callback.tsx
+++ b/pages/callback.tsx
@@ -23,7 +23,7 @@ export default function Connection() {
         redirectUri: home,
       })
     );
-  }, []);
+  }, [auth0, connection, router]);
 
   return <LoadingScreen />;
 }
--- a/pages/profile/[...view].tsx
+++ b/pages/profile/[...view].tsx
@@ -16,7 +16,7 @@ function ProfilePage({ setModal }: PropsFromRedux) {
       setModal("settings", { view: "preferences" });
       replace("/");
     }
-  }, [modal]);
+  }, [isAuthenticated, modal, setModal, replace]);
 
   return <Account />;
 }
--- a/pages/recording/[id].tsx
+++ b/pages/recording/[id].tsx
@@ -84,7 +84,7 @@ function RecordingPage({
       setRecording(await getAccessibleRecording(recordingId));
     }
     getRecording();
-  }, [recordingId, store]);
+  }, [recordingId, store, getAccessibleRecording]);
 
   if (!recording || typeof window === "undefined") {
     return (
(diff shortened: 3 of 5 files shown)


pages/callback.tsx
```diff
@@ -23,7 +23,7 @@ export default function Connection() {
         redirectUri: home,
       })
     );
-  }, []);
+  });
```
abstractalgo, line 26: This will execute on every rerender, which is for sure not what we want.

Either do as:
```
useEffect(() => {
   ...
   // eslint-disable-next-line react-hooks/exhaustive-deps
}, []) 
```
or actually put all of the dependencies that it's complaining about - a perf hit for checking deps is usually negligible (shallow compare).

PR #5006, opened 2022-01-12
Fix the network monitor's borders

jcmorrow:
# Before
CleanShot 2022-01-12 at 14 37 17
- Border runs across only the left half of the filter bar when the details are open
- Border only runs along the bottom of the request detail tabs but doesn't extend to the end of the bar
- Scrollbar is big and combines with the splitter to make a crazy wide seam in pretty restricted real-estate

# After
CleanShot 2022-01-12 at 14 56 07

- Borders are the same width and color and won't suddenly stop running across anything
- Scroll bar is *much* narrower
- "But Josh!" you will say "how am I supposed to grab that tiny 1px splitter!?" And to that I say:

```
  background-clip: padding-box;
```

And by that I mean that the splitter can be grabbed across an 8px area, but will only present as 1px wide because of a little magic in the combination of `background-clip: padding-box`, transparent borders, and negative margin.

### DIFF
--- a/src/devtools/client/shared/components/splitter/SplitBox.css
+++ b/src/devtools/client/shared/components/splitter/SplitBox.css
@@ -6,8 +6,6 @@
   display: flex;
   flex: 1;
   min-width: 0;
-  height: 100%;
-  width: 100%;
 }
 
 .split-box.vert {
--- a/src/devtools/client/shared/components/tabs/Tabs.css
+++ b/src/devtools/client/shared/components/tabs/Tabs.css
@@ -12,7 +12,7 @@
 }
 
 /* Hides the tab strip in the TabBar */
-div[hidetabs=true] .tabs .tabs-navigation {
+div[hidetabs="true"] .tabs .tabs-navigation {
   display: none;
 }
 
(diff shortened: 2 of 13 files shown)


jasonLaster:
`background-clip: padding-box` 🎉

jasonLaster (approved):
Looks great!

The only weird thing i see is the vertical scrollbar in the table looks clipped now.

Also, unrelated to this PR, the last row in the table is also clipped.

jcmorrow:
> Looks great!
> 
> The only weird thing i see is the vertical scrollbar in the table looks clipped now.
> 
> Also, unrelated to this PR, the last row in the table is also clipped.

Alright, digging into why that clipping was happening caused all sorts of unwinding and rebuilding of the CSS of this thing. I'm not sure how we got into such a complicated combination of overflows! I think it's considerably simpler now, all of the information should be visible. I've hidden the scroll bar, though tbh I'm not sure that I love it? There is still this one weird strip along the right side of the requests table that the browser throws in whether or not the content is vertically scrollable (so it's not a scrollbar) and doesn't seem possible to get rid of? Maybe with another 4 hours of fiddling, but I think I'm good with this version for now 🙃

PR #5008, opened 2022-01-13
Conditionally enable/disable step debugging commands

jaril:
This PR should keep users from trying to navigate via step commands when it 1) doesn't make sense for them to, or 2) we can't handle the navigation for some reason.

Resume/Resume backwards:
- Disabled when the user has not added any breakpoints yet, since there's nothing to resume to.

Step over/over backwards/into/out:
- Disabled when the user is not paused on a frame/line of code.
- Even if they're paused on a frame, it's possible that we can't properly navigate from that frame because of [the current protocol's limitations](https://github.com/RecordReplay/backend/pull/3367) (e.g. more than ~3k steps in the current frame).


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -103,7 +103,14 @@ function LineNumberTooltip({
...
 
   if (analysisPoints === "error") {
-    return <StaticTooltip targetNode={targetNode}>Failed</StaticTooltip>;
+    return (
+      <StaticTooltip targetNode={targetNode} className="hot">
+        <div className="flex items-center">
+          <MaterialIcon className="mr-1">warning_amber</MaterialIcon>
+          <div>10k+ hits</div>
+        </div>
+      </StaticTooltip>
```
jasonLaster, line 112: Nice

src/devtools/client/debugger/src/components/SecondaryPanes/CommandBar.js
```diff
@@ -98,91 +104,131 @@ class CommandBar extends Component {
...
+        type="resume"
+      />
+    );
+  }
+
+  renderStepButtons() {
+    const { isPaused, hasFramePositions } = this.props;
+    const disabled = !isPaused || !hasFramePositions;
+    const disabledTooltip = !isPaused
+      ? "Stepping is disabled until you add a breakpoint"
```
jasonLaster, line 183: `Stepping is disabled until you pause`

jasonLaster (approved):
Nice

PR #5010, opened 2022-01-13
Fix e2e tests that use the elements panel

hbenl:
Fixes the failures caused by RecordReplay/gecko-dev#665 and a bug in the test harness.


jcmorrow (approved):
Nice!

PR #5015, opened 2022-01-14
Do not cache SSR for uninitialized recordings

ryanjduffy:
## Issue

The server-side rendering of new, uninitialized recordings doesn't include the social metadata because the recording is private by default so the first render of the page for the upload screen caches a rendering for 60 minutes that doesn't include the metadata.

## Resolution

Set the refresh time to be "as soon as possible" (1) if the recording is not initialized.

## Notes

This isn't quite ready yet because if the recording hasn't been initialized yet, it can't be public so the graphql query can't find it.


ryanjduffy:
Going to close it for now. We're not convinced the backend impl is a good idea and it doesn't seem to be impacting us too much yet.

PR #5016, opened 2022-01-14
Disable failing e2e tests

hbenl:
At least some of these tests are currently failing because they need sourcemaps in the example recordings:
- playwright currently doesn't include sourcemaps in the example recordings: RecordReplay/ops#209
- at least some of the bundles and sourcemaps in the examples were manually edited in RecordReplay/devtools#4915 and are now broken - we need to recreate them with the necessary change (replacing `dump("RecReplaySendAsyncMessage Example__Finished")` with `console.log("ExampleFinished")`)

### DIFF
--- a/test/manifest.js
+++ b/test/manifest.js
@@ -44,6 +44,7 @@ module.exports = [
     // Not supported on chromium, needs source maps.
     // https://github.com/RecordReplay/chromium/issues/5
     targets: ["gecko"],
+    pending: true,
   },
   {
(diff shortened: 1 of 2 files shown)


jcmorrow (commented):
Seems good! Are there instructions for how to fix the bundles/sourcemaps when they get edited?

hbenl:
@jcmorrow Not that I know of. We may have to extract the original sources from the sourcemaps and recreate the bundle from those.

PR #5018, opened 2022-01-14
Fix the SplitBox css

hbenl:
@jcmorrow You removed these 2 lines in #5006, but this affects other panels (I noticed it in the elements panel). Adding them back I couldn't see any difference in the Network panel, but maybe I'm overlooking something...


jcmorrow:
@hbenl could you also say what this is fixing so I know where to look next time?

jcmorrow (approved):

hbenl:
@jcmorrow I noticed it in the elements panel - it didn't fill its container

PR #5021, opened 2022-01-14
Add slug to recording URL path

ryanjduffy:
I was reading through some customer feedback and they noted that it would be nice to include a slug of the title in the URL of a recording so it was easier to see at a glance. This sort of follows up on the social cards work since that only works for public recordings but this improves sharing for private recordings.

The implementation does a URL replace for a recording if the URL doesn't include the slug to "convert" old URLs to the slugified format. It also updates the Library and the share modal to use the new format.

URL example:

```
http://localhost:8080/recording/create-next-app-with-basic-auth/2cc4d499-a18b-4b3a-adcf-317c8766fa43
```

https://app.replay.io/recording/597a46be-8ebf-4b8a-ab99-d45aa8e73873


src/ui/hooks/recordings.ts
```diff
@@ -105,7 +107,16 @@ const GET_MY_RECORDINGS = gql`
...
 export function useGetRecordingId() {
   const { id } = useRouter().query;
-  return Array.isArray(id) ? id[0] : id!;
+  return extractIdAndSlug(id).id!;
+}
+
+export function getRecordingURL(recording: Recording) {
+  let slug = "";
+  if (recording.title) {
+    slug = `${slugify(recording.title).toLowerCase()}/`;
```
jcmorrow, line 116: This is super pedantic but I think the `/` should on line 119 and this should be 

```
    slug = slugify(recording.title).toLowerCase()
```

src/ui/utils/helpers.ts
```diff
@@ -30,3 +31,23 @@ export function isValidTeamName(name: string) {
 
   return true;
 }
+
+export function extractIdAndSlug(params: string | string[] | undefined) {
```
jcmorrow, line 35: This is probably just a style thing, but to me the fact that next router has so many possible return types and the fact that we're looking at a variety of urls are problems that are separate and should be handled separately. I'm thinking something like:

```
export function toArray(params: string | string[] | undefined): string[] {
  if (Array.isArray(params)) {
    return params;
  } else if (params === undefined) {
    return [];
  } else {
    return Array(params);
  }
}

export function extractIdAndSlug(params: string | string[] | undefined) {
  let asArray = toArray(params);
  const id: string | undefined = toArray(asArray).find(validateUUID);
  const slug = asArray.length === 2 ? asArray[0] : undefined;

  return { id, slug };
}
```

jcmorrow (commented):
My inner backend engineer really wants the urls not to have the `/` between two identifiers for the same resource. I would expect `my-long-slug-123456456456` not `my-long-slug/123456456456`. Better yet would be a world where `/recordings/my-long-slug` works, obviously, but that involves a lot more work.

jcmorrow:
On the preview branch, `https://devtools-px8uc874d-recordreplay.vercel.app/recording/i-can-type-whatever-i-want-here/68b7c20d-0b0d-49c1-bb4a-622d2266fd1a` gets redirected to `https://devtools-px8uc874d-recordreplay.vercel.app/recording/comment-wrong-date/68b7c20d-0b0d-49c1-bb4a-622d2266fd1a?id=i-can-type-whatever-i-want-here&id=68b7c20d-0b0d-49c1-bb4a-622d2266fd1a` which is a bit odd (notice the multiple `id` query params).

src/ui/hooks/recordings.ts
ryanjduffy: That was my first impl but we can't guarantee a title (e.g. from node recordings) so the logic below gets messy if we have to conditionally add the slash below based on `slug` being valued. It's doable but not my preference.

src/ui/utils/helpers.ts
ryanjduffy: I considered something along this line but wanted to be more prescriptive about the URL format being either `/id` or `/slug/id` (and not `/slug/random/stuff/id`). There's probably a compromise though that converts to an Array as you've suggested but retains the prescriptive structure.

ryanjduffy:
Yeah. I just saw that as well. Must be another "fix" hiding somewhere from the old `/view?id=` paths.

ryanjduffy:
@jcmorrow - The extra `?id=` is fixed by 94e5aa3. I also added commits to replace the old `/view?id=` redirect with a next server-side redirect and a refactoring fix for `toArray`.

ryanjduffy:
> My inner backend engineer really wants the urls not to have the / between two identifiers for the same resource. I would expect my-long-slug-123456456456 not my-long-slug/123456456456.

That's fair. It's a bit odd to structure the URL that way. I like the readability of having the title and id split by something other than `-` because the ID uses that delimiter as well. I don't want to change the URL pattern more than once so let's get consensus before moving this forward.

> Better yet would be a world where /recordings/my-long-slug works, obviously, but that involves a lot more work.

Definitely more work. I don't think we need to get into the game of ensuring globally-unique recording names right now. Maybe if they were qualified by a team id but even then it'll get out of hand quickly if an org is using automated tests to make thousands of recordings.

jcmorrow:
@ryanjduffy maybe this is crazy but does `slug:id` work?

src/ui/hooks/recordings.ts
jcmorrow: What about `["recording",slug,recording.id].filter(Boolean).join("/")`? Maybe I'm just making things worse 😆

src/ui/utils/helpers.ts
jcmorrow: yeah I am 100% certain there is a clever, elegant way to do it, but this is readable enough, and that's what matters most 👍

src/ui/hooks/recordings.ts
jcmorrow: or `compact(["recording", slug, recording.id].join("/")` is probably clearer

ryanjduffy:
@jcmorrow what do you think of using a double dash (`--`) as a separator? I like it better than `:` since it adds a bit more space between the slug and the id. Also, `slugify` will already de-dupe multiple separators so we know that `--` will be our separator and not come from the title nor the id.

```
http://localhost:8080/recording/check-source-maps-with-replay-app--c7602b7a-3e75-41ab-8ce7-eb1b2506b3e1
```

src/ui/hooks/recordings.ts
```diff
@@ -105,7 +107,16 @@ const GET_MY_RECORDINGS = gql`
...
 export function useGetRecordingId() {
   const { id } = useRouter().query;
-  return Array.isArray(id) ? id[0] : id!;
+  return extractIdAndSlug(id).id!;
+}
+
+export function getRecordingURL(recording: Recording) {
+  let slug = "";
+  if (recording.title) {
+    slug = `${slugify(recording.title).toLowerCase()}--`;
```
abstractalgo, line 116: There's also a related part that parses URL (split by `--`) down below, that is in a different place, with no shared locality in any way, even though they work with the same assumption (namely, what's the expected URL format).

I'd recommend making getting and parsing URL closer to each other (in some same utils file e.g.) or ever part of some static methods on a helper class. Updating anything with URLs will require weird digging through the codebase and hope we didn't forget to update it somewhere.

(this is a non-blocking suggestion though)

abstractalgo: At some point, we will have to consolidate different pages, routes and URL patterns that exist, that are currently all hardcoded in different places, so a step in that general direction is a good thing either way.

jcmorrow: Yeah, it might feel silly to extract `--` to a named constant like `URL_SEPARATOR` but even something as simple as that and importing it from the same place would help towards making the link discoverable 👍 , but also making generator/parser functions that live next to each other works for me!

jcmorrow:
Thanks for getting buy-in on these changes @ryanjduffy! I could easily see us making a choice here that causes headaches in the future, I think getting multiple brains on it makes it a lot less likely that will happen 🙏 💯

ryanjduffy:
I did another round of refactoring. I discovered we have a `utils/recording.ts` so I moved both the `getRecordingId` and `getRecordingURL` there since neither are hooks nor do they have any dependency on the hooks. So I ended up touching a lot more files but I think it results in a better dependency tree.

cc: @jcmorrow or @abstractalgo for a 👍

Issue #4909, opened 2021-12-23
Elements should expand before loaded


jasonLaster:
**Description:**
Notice how it takes this element 1 second to expand. This can often be longer if there is a lot of content.

We should optimistically expand and show a simple `Loading...` span

https://app.replay.io/recording/3f56cff4-cd4f-4920-80dd-5fffe82f9937?point=46730671738432206504419732100226260&time=29132&hasFrames=true#

PR #5029, opened 2022-01-17
Elements panel: expand elements while their children are being loaded (#4909)

hbenl:
https://app.replay.io/recording/0ffed9ac-d6a9-47af-a4c4-4ca9a1dbe25d


abstractalgo (approved):
lgtm

Issue #4908, opened 2021-12-23
Elements: Differentiate between selected and focused nodes


jasonLaster:
It's really easy to confuse being focused on a node and able to use the arrow keys and just being selected.

We should differentiate between the two.

```css
.tag-line .tag-state.theme-selected {
  background-color: #3ac0ff;
}

.tag-line .tag-state.theme-selected:focus {
  background-color: var(--theme-selection-background);
}
```

The open question for me is how does the inspector select the next node? based on that we can add a `.focus()` in a useEffect

PR #5031, opened 2022-01-17
Show if the markup panel has focus (#4908)

hbenl:
When the markup panel has focus (and therefore the cursor keys navigate in the markup and not in the timeline), we draw a blue border around it - the same border that we use to highlight text inputs having focus.

### DIFF
--- a/src/devtools/client/themes/markup.css
+++ b/src/devtools/client/themes/markup.css
@@ -442,6 +442,11 @@ ul.children + .tag-line::before {
 
   overflow: auto;
   height: 100%;
+  border: 1px solid transparent;
+}
+
+#markup-box:focus-within {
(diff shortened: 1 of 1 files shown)


abstractalgo (approved):
As we discussed, this works fine, maybe just worth considering if we want to use blue/grey background for the selected node to those if there's an actual focus on the panel or now, which would make it a bit more consistent with the FF/Chrome and how they signal the focus of the panel.

hbenl:
> maybe just worth considering if we want to use blue/grey background for the selected node

Implemented in #5034

abstractalgo:
We can close this one then?

hbenl:
> We can close this one then?

Yeah, I just kept it open in case someone wants to compare the two.

PR #5034, opened 2022-01-18
Highlight the selected element in grey if the markup panel doesn't have focus (#4908)

hbenl:


jasonLaster (approved):

abstractalgo (approved):
Trying lighter colors and leaving node's text the same (without turning to white), or using borders,... as is looks the best (and lowest effort for the actual value).

Firefox itself doesn't do coloring differently based on panel being focused or not, always vivd blue bg + white text, so we don't have a reference.

Let's merge this and perhaps reconsider later once we have styles more unified. 👍

PR #5035, opened 2022-01-18
Fix the breadcrumbs click handler in dev

hbenl:


abstractalgo:
Tested it, but made absolutely no difference, still unresponsive and missing `...`.

abstractalgo (approved):

PR #5039, opened 2022-01-18
Spruce up the elements panel toolbar

jcmorrow:
- Vertically center search bar items
- Make sure that the left side of the search box isn't cut off (I think at some point we removed the accompanying padding but not the negative margin)
- Make the toolbars the same height across the two panels
- Add a little more horizontal space around the results count
- Make the search box flex-grow so the results are on the right
- Remove the blue border on the input when focused

### Before
CleanShot 2022-01-18 at 10 29 58

### After

CleanShot 2022-01-18 at 14 53 13

### DIFF
--- a/src/devtools/client/inspector/markup/components/MarkupApp.tsx
+++ b/src/devtools/client/inspector/markup/components/MarkupApp.tsx
@@ -39,8 +39,11 @@ function MarkupApp(props: PropsFromRedux & { inspector: Inspector }) {
             />
           </div>
           <div id="inspector-searchlabel-container" hidden={true}>
-            <div className="devtools-separator"></div>
-            <span id="inspector-searchlabel"></span>
+            <div
+              className="devtools-separator"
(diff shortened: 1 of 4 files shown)


jasonLaster:
One small ui change. should the search bar fill the horizontal space so that results are on the right?

jcmorrow:
> One small ui suggestion. I think the search bar fill the horizontal space so that results are on the right. Also not sure if we need to blue border

Yep. Agreed, I can fix that up in a jiffy 🚗

hbenl (commented):
- I already added the flex-grow in #5030
- if we want to remove the blue border, we should also do this for the filter boxes in the rules and computed side panels

jcmorrow:
Oh sorry @hbenl I didn't see that you were touching this at the same time, otherwise I would have just left a comment about the toolbar heights (because that was all I started out fixing and the rest just got tacked on). I'll rebase and leave the blue outline for now 👍

PR #5040, opened 2022-01-18
Update the UI to show if the entire recording has been unloaded

### LINKED ISSUE #4715: Show when no regions loaded
In this recording in this link https://github.com/RecordReplay/ops/issues/376#issuecomment-980378493 (not providing the recording directly because it was shared with us externally) we unloaded large parts of the recording (possibly the entire recording, not sure), but there is no indication of this in the UI.  We have a mock test checking that the UI updates to show unloaded regions.  Is this test not enabled, or not working?

Also, we used to have console logs when the loaded regions from the backend changed.  I found these pretty handy to see what was going on, irrespective of the UI, but they don't seem to be logged anymore.  Can these be added back?

jaril:
Fix #4715.

image

### DIFF
--- a/src/ui/components/Timeline/index.tsx
+++ b/src/ui/components/Timeline/index.tsx
@@ -290,6 +290,12 @@ class Timeline extends Component<PropsFromRedux> {
       return null;
     }
 
+    // An empty loadedRegions array after we've finished loading regions means that
+    // the whole recording has been unloaded.
+    if (loadedRegions.length === 0) {
+      return <div className="unloaded-regions w-full" />;
+    }
+
     const { begin, end } = loadedRegions[0];
     const { endTime } = zoomRegion;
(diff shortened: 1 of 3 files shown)


src/ui/components/Timeline/index.tsx
```diff
@@ -290,6 +290,12 @@ class Timeline extends Component<PropsFromRedux> {
       return null;
     }
 
+    // An empty loadedRegions array after we've finished loading regions means that
+    // the whole recording has been unloaded.
+    if (loadedRegions.length === 0) {
+      return <div className="unloaded-regions start" style={{ width: `100%` }} />;
```
abstractalgo, line 296: Can use Tailwind's `w-full`.

abstractalgo (approved):
lgtm (maybe just post a screenshot in descriptioon)

PR #5042, opened 2022-01-18
Make the modals stack properly

jcmorrow:
This solves the problem in two ways:

- Makes the error block out everything behind it
- Closes other modals when we set an error

Either one is enough to fix this particular bug, but both seems more
future-proof.

Fixes https://github.com/RecordReplay/devtools/issues/4967

### DIFF
--- a/src/ui/components/shared/Viewport.tsx
+++ b/src/ui/components/shared/Viewport.tsx
@@ -13,7 +13,7 @@ function FullViewportWrapper({
   return (
     <main
       className={classNames(
-        "w-full fixed h-full z-50 flex items-center justify-center",
+        "w-full fixed h-full z-50 flex items-center justify-center bg-chrome",
         classnames
       )}
     >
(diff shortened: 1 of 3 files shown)


src/ui/reducers/app.ts
```diff
@@ -76,11 +76,11 @@ export default function update(
     }
 
     case "set_expected_error": {
-      return { ...state, expectedError: action.error };
+      return { ...state, expectedError: action.error, modal: null };
```
jaril, line 79: There's also `modalOptions` that we probably should clear out just to keep everything tidy.

src/ui/components/shared/Viewport.tsx
```diff
@@ -13,7 +13,7 @@ function FullViewportWrapper({
   return (
     <main
       className={classNames(
-        "w-full fixed h-full z-50 flex items-center justify-center",
+        "w-full fixed h-full z-50 flex items-center justify-center bg-chrome",
```
jaril, line 16: What was the reason for adding `bg-chrome` here?

jcmorrow: Yeah, so it's no longer possible to get into this state (because the modal behind will close if you pop an error), but previously it was possible to have both the modal and the error open, and because the background to the error is translucent, and so is the background to the modal (but with the modal body sitting on top) you end up with this:

CleanShot 2022-01-18 at 14 32 23@2x

Issue #5032, opened 2022-01-17
Elements panel's Breadcrumbs improvements


abstractalgo:
**Description:**

Improvements to breadcrumbs part of the Elements panel regarding having left-most and right-most `...` buttons for scrolling through node's breadcrumbs (currently missing).

![image](https://user-images.githubusercontent.com/1355455/149817865-a982c7e3-8a18-49c0-8a0a-355f2300f902.png)

Clicking on the breadcrumb item already navigates to the item properly.

hbenl:
This works in Firefox but not in other browsers. The reason is that the `visibility` of the buttons is changed in `overflow`/`underflow` event handlers, but those events are only supported in Firefox.

PR #5043, opened 2022-01-19
Elements panel breadcrumbs fixes (#5032)

hbenl:
- fixes the scroll buttons in Chrome
- fixes the ellipsis that is used when a breadcrumb is shortened

### DIFF
--- a/src/devtools/client/inspector/breadcrumbs.js
+++ b/src/devtools/client/inspector/breadcrumbs.js
@@ -6,7 +6,7 @@
 
 const promise = Promise;
 const flags = require("devtools/shared/flags");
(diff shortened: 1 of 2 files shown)


abstractalgo (approved):
Tested in Chrome and FF, works very nicely!

One question though, under which circumstances should the ellipsis be shown?

hbenl:
> One question though, under which circumstances should the ellipsis be shown?

https://github.com/RecordReplay/devtools/blob/a8fcfb8afb3a6dbcb71e25bba93b19024ea60db6/src/devtools/client/inspector/breadcrumbs.js#L471-L479

Issue #5044, opened 2022-01-19
Markup panel: the selection should stretch to the right border of the panel


hbenl:
![Screenshot from 2022-01-19 16-11-10](https://user-images.githubusercontent.com/16060807/150158865-1c166c92-8695-4835-bdfd-aecae5affd5a.png)

PR #5045, opened 2022-01-19
Make the markup box fill its panel (#5044)

hbenl:
Before:
Screenshot from 2022-01-19 16-11-10

After:
Screenshot from 2022-01-19 16-17-05

### DIFF
--- a/src/devtools/client/themes/markup.css
+++ b/src/devtools/client/themes/markup.css
@@ -454,7 +454,7 @@ ul.children + .tag-line::before {
 }
 
 #markup-root-wrapper {
-  width: fit-content;
+  min-width: fit-content;
 }
 
 #markup-root ul {


abstractalgo:
Not a desired behavior (notice empty space on the right).

https://user-images.githubusercontent.com/1355455/150182942-c9c4a849-17fb-4aba-86ce-a96b18e95c1b.mp4

abstractalgo (changes requested):
Not a desired behavior (notice empty space on the right).

https://user-images.githubusercontent.com/1355455/150182942-c9c4a849-17fb-4aba-86ce-a96b18e95c1b.mp4

hbenl:
Thanks! So that's what the `width: fit-content` was doing there in the first place. Let's try with `min-width: fit-content` then.

abstractalgo (approved):
lgtm 👍

Issue #4757, opened 2021-12-06
Elements - left arrow should not go back to a prior frame


jasonLaster:
**Description:**

The default behavior works, but if you go quickly you can cause the elements panel to miss one and then seek backwards

https://app.replay.io/recording/b9353076-ba95-4062-ad46-5b0ae4ce628c

jasonLaster:
This is difficult to do in practice, but we could make it clearer by 

a) having a different select / focus state https://github.com/RecordReplay/devtools/issues/4908
b) ensuring that arrow keys always bring the focus to another node. This is an important invariant we should maintain and perhaps test for.

hbenl:
I was finally able to reproduce this:
- click on a child element
- press the left key twice:
  - the first keypress selects the parent element, but the child still has focus
  - the second keypress collapses the parent element, now the focused child has disappeared and hence the markup box loses focus
- press the left key again -> the recording jumps to the previous screenshot

PR #5048, opened 2022-01-19
Ensure that the markup panel doesn't lose focus on navigation (#4757)

hbenl:
repro steps: https://github.com/RecordReplay/devtools/issues/4757#issuecomment-1016575966


abstractalgo:
Seems good to me! 👍

abstractalgo (approved):

PR #5049, opened 2022-01-19
Remove custom Babel config to unblock SWC

abstractalgo:
Because we used a custom Babel config, SWC ([a new NextJS compiler from v12](https://nextjs.org/docs/advanced-features/compiler)) was disabled:
image

Babel config was only including `@babel/plugin-transform-react-display-name` plugin, which is either way included with `preset-react` that comes pre-setup with Next, and it seems that we aren't using it either way...so, all in all, it doesn't change anything for us, but we do get a SWC compiler that should be much faster.


jcmorrow (approved):
Seems like a big win to me!

PR #5050, opened 2022-01-19
Add configuration options to test runner

jcmorrow:
This adds explicit configuration options for `driverPath` (the linker to use when recording) and `browserPath` (the Replay Browser executable to use). Passing things by environment variables still totally works, but explicitly configuring this stuff makes doing things like overriding `replay-node` but only in certain cases (which is how we run the backend test suite) a *lot* easier.

### DIFF
--- a/test/src/state.js
+++ b/test/src/state.js
@@ -19,7 +19,8 @@ const defaultState = {
   headless: false,
   shouldRecordExamples: !!process.env.SHOULD_RECORD_EXAMPLES,
 
-  executablePath: process.env.RECORD_REPLAY_PATH,
+  browserPath: process.env.RECORD_REPLAY_PATH,
+  driverPath: process.env.RECORD_REPLAY_DRIVER,
   onlyTarget: process.env.onlyTarget,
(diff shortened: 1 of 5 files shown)


jasonLaster (approved):
Nice

test/src/state.js
```diff
@@ -51,6 +54,12 @@ function processArgs(state, argv) {
       case "--count":
         state.count = +argv[++i];
         break;
+      case "--driverPath":
+        state.driverPath = +argv[++i];
```
hbenl, line 58: `state.driverPath` is not a number, so you need to remove the `+` before `argv[++i]`

test/src/state.js
```diff
@@ -51,6 +54,12 @@ function processArgs(state, argv) {
...
+        break;
+      case "--browserPath":
+        state.browserPath = +argv[++i];
```
hbenl, line 61: same here

hbenl (changes requested):

test/src/state.js
jcmorrow: 🤦

PR #5052, opened 2022-01-20
Add support for annual subscriptions

ryanjduffy:
## Issue

We'd like to be able to support yearly subscriptions. At the moment, this is limited to subscriptions manually created but will likely eventually allow for self service selection between monthly and annual subscriptions.

## Resolution

This first iteration adds support for displaying the details of two new annual plans -- one for teams and one for orgs -- so that any teams manually configured for those plans will render correctly in the app.

image

### DIFF
--- a/src/ui/components/shared/WorkspaceSettingsModal/utils.tsx
+++ b/src/ui/components/shared/WorkspaceSettingsModal/utils.tsx
@@ -45,6 +45,7 @@ export const pricingDetailsForSubscription = (subscription: Subscription): PlanP
         billingSchedule: null,
         displayName: "Beta Tester Appreciation",
         seatPrice: 0,
+        discount: 0,
         trial: isTrial(subscription),
       };
     case "team-v1":
@@ -53,13 +54,15 @@ export const pricingDetailsForSubscription = (subscription: Subscription): PlanP
         billingSchedule: "monthly",
         displayName: "Team",
         seatPrice: 20,
+        discount: 0,
         trial: isTrial(subscription),
       };
     case "team-annual-v1":
       return {
         billingSchedule: "annual",
         displayName: "Team",
-        seatPrice: 220,
+        seatPrice: 20,
+        discount: 0.1,
         trial: isTrial(subscription),
       };
     case "org-v1":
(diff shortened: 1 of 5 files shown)


src/ui/components/shared/WorkspaceSettingsModal/utils.tsx
```diff
@@ -55,13 +55,27 @@ export const pricingDetailsForSubscription = (subscription: Subscription): PlanP
...
         billingSchedule: "monthly",
         displayName: "Organization",
         seatPrice: 75,
         trial: isTrial(subscription),
       };
+    case "org-annual-v1":
+      return {
+        billingSchedule: "annual",
+        displayName: "Organization",
+        seatPrice: 825,
```
jasonLaster, line 76: We should still display monthly price. We just need a new discount field which we can set to 10%.

src/ui/components/shared/WorkspaceSettingsModal/utils.tsx
```diff
@@ -81,9 +95,5 @@ export const cycleCharge = (planPricing: SubscriptionWithPricing): number => {
     return 0;
   }
 
-  let monthsPerCycle = 1;
-  if (planPricing.billingSchedule === "annual") {
-    monthsPerCycle = 12;
-  }
-  return monthsPerCycle * planPricing.seatPrice * planPricing.seatCount;
+  return planPricing.seatPrice * planPricing.seatCount;
```
jasonLaster, line 98: this should be `(planPricing.billingSchedule == "annual" ? 12 : 1) *  planPricing.seatPrice * planPricing.seatCount * (1 - planPricing.discount)`

jasonLaster (commented):
Left some thoughts here.

ryanjduffy:
Works for me. I'll patch that up tomorrow

src/ui/components/shared/WorkspaceSettingsModal/utils.tsx
ryanjduffy: Right now, we display "cost per seat" and "monthly charge" for monthly renewals. Do you want to show both monthly and annual cost for annual renewals along with the "cost per seat"? What about the discount?

Here's what I have now but I don't think the messaging is clear enough.

image

jasonLaster: I think it's better to show

```
Cost  per seat:        $20.00
Discount:                     10%
Savings:                 -$24.00
Annual charge.      $216.00
```

ryanjduffy: Maybe this?

image

jasonLaster (approved):

jasonLaster:
One TS issue, otherwise good.

PR #5053, opened 2022-01-20
Add back autocomplete

### LINKED ISSUE #5003: Bring back console autocomplete
We used to have autocomplete in the console but it was removed to keep the initial implementation simpler. We should bring it back.

jaril:
Fix #5003.

### V1 requirements:
- [x] it should autocomplete variable names
- [x] it should autocomplete property names for scope objects
- [x] it should include a scope object's prototype's properties in the autocomplete options
- [x] it should handle both bracket and dot property syntax
- [x] it should support autocomplete for the top frame
- [x] it should hide the autocomplete options when the user moves the cursor left/right, or presses escape
- [x] it should hide the autocomplete options when the user moves the cursor left/right using the mouse
- [x] it should not show autocomplete if the current token is also the only autocomplete match available
- [x] modifying the console terminal input should show autocomplete, if there are corresponding matches
- [x] it should use fuzzy matching to find autocomplete matches
- [x] it should have clickable autocomplete options
- [x] it should be offset from the left such that it lines up with the current token being autocompleted

### Follow up: 
Moved to https://github.com/RecordReplay/devtools/issues/5080

### DIFF
--- a/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
+++ b/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
@@ -3,32 +3,29 @@ import classNames from "classnames";
 
 function Match({
   label,
-  selectedIndex,
-  index,
+  isSelected,
   onClick,
 }: {
   label: string;
-  selectedIndex: number;
-  index: number;
-  onClick: (index: number) => void;
+  isSelected: boolean;
+  onClick: (match: string) => void;
 }) {
-  const buttonNode = useRef<HTMLButtonElement | null>(null);
-  const selected = selectedIndex === index;
+  const buttonNode = useRef<HTMLButtonElement>(null);
 
   useEffect(() => {
-    if (selected && buttonNode.current) {
+    if (isSelected && buttonNode.current) {
       buttonNode.current.scrollIntoView({ block: "nearest", inline: "end" });
     }
-  }, [selected]);
+  }, [isSelected]);
 
   return (
     <button
       className={classNames(
         "text-left px-1 cursor-default",
-        selected ? "bg-blue-700 text-white" : "hover:bg-blue-100"
+        isSelected ? "bg-blue-700 text-white" : "hover:bg-blue-100"
       )}
       ref={buttonNode}
-      onClick={() => onClick(index)}
+      onClick={() => onClick(label)}
     >
       {label}
     </button>
@@ -44,7 +41,7 @@ export default function Autocomplete({
   leftOffset: number;
   matches: string[];
   selectedIndex: number;
-  onMatchClick: (index: number) => void;
+  onMatchClick: (match: string) => void;
 }) {
   return (
     <div
@@ -59,13 +56,7 @@ export default function Autocomplete({
       }}
     >
       {matches.map((match, i) => (
-        <Match
-          label={match}
-          selectedIndex={selectedIndex}
-          index={i}
-          key={i}
-          onClick={onMatchClick}
-        />
+        <Match label={match} isSelected={i === selectedIndex} key={i} onClick={onMatchClick} />
       ))}
     </div>
   );
--- a/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
+++ b/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
@@ -1,0 +1,116 @@
+import classNames from "classnames";
+import { getFrameScope } from "devtools/client/debugger/src/reducers/pause";
+import { getSelectedFrame } from "devtools/client/debugger/src/selectors";
+import React, { useEffect, useRef } from "react";
+import { useSelector } from "react-redux";
+import { UIState } from "ui/state";
+import {
+  getBinding,
+  getBindingNames,
+  getGlobalVariables,
+  getPropertiesForObject,
+  shouldVariableAutocomplete,
+} from "../../utils/autocomplete";
+
+function useGetAutocompleteMatches(input: string) {
+  const selectedFrame = useSelector(getSelectedFrame);
+  // Is the selected frame id always 0:0? -jvv
+  const scopes = useSelector((state: UIState) => getFrameScope(state, "0:0"));
+
+  if (!scopes?.scope) {
+    return [];
+  }
+
+  const bindingNames = getBindingNames(scopes);
+
+  if (shouldVariableAutocomplete(input)) {
+    const variableNames = [...bindingNames, ...getGlobalVariables(scopes.scope)];
+    return variableNames.filter(name => name.toLowerCase().includes(input.toLowerCase()));
+  } else {
+    const expressions = input.split(".");
+    const lastExpression = expressions[expressions.length - 1];
+    const parentExpression = expressions.slice(0, expressions.length - 1).join(".");
+
+    const binding = getBinding(parentExpression, scopes);
+
+    if (!binding) {
+      return [];
+    }
+
+    const object = binding.value._object;
+
+    if (!object) {
+      return [];
+    }
+
+    const properties = getPropertiesForObject(object);
+
+    return properties.filter(p => p.toLowerCase().includes(lastExpression.toLowerCase()));
+  }
+}
+
+function Match({
+  label,
+  selectedIndex,
+  index,
+}: {
+  label: string;
+  selectedIndex: number;
+  index: number;
+}) {
+  const buttonNode = useRef<HTMLButtonElement | null>(null);
+  const selected = selectedIndex === index;
+
+  useEffect(() => {
+    if (selected && buttonNode.current) {
+      buttonNode.current.scrollIntoView({ block: "nearest" });
+    }
+  }, [selected]);
+
+  return (
+    <button
+      className={classNames(
+        "text-left px-1",
+        selected ? "bg-blue-700 text-white" : "hover:bg-blue-100"
+      )}
+      ref={buttonNode}
+    >
+      {label}
+    </button>
+  );
+}
+
+export default function Autocomplete({
+  input,
+  selectedIndex,
+}: {
+  input: string;
+  selectedIndex: number;
+}) {
+  const matches = useGetAutocompleteMatches(input);
+
+  if (!matches.length) {
+    return null;
+  }
+
+  return (
+    <div
+      className="flex flex-col border py-1 absolute left-7 bottom-8 font-mono overflow-auto"
+      style={{ minWidth: "240px", maxHeight: "240px", fontSize: "11px" }}
+    >
+      {matches.map((match, i) => (
+        <Match label={match} selectedIndex={selectedIndex} index={i} key={i} />
+      ))}
+    </div>
+  );
+}
+
+// V1
+// it should handle chained objects
+// it should handle computed property syntax
+// it should know what block it's in and suggest stuff for that block based on cursor position
+// it should replace the correct expression with the selected autocompleted expression
+// pressing escape should close autocomplete
+
+// V2
+// Todo: the dropdown should have a variable X position based on the cursor position
--- a/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
+++ b/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
@@ -44,18 +44,14 @@ export default function Autocomplete({
   selectedIndex: number;
   onMatchClick: (index: number) => void;
 }) {
-  if (!matches.length) {
-    return null;
-  }
-
   return (
     <div
       className="flex flex-col bg-white border py-1 absolute left-7 -mb-1 shadow-sm font-mono overflow-auto"
       style={{
         minWidth: "240px",
         maxHeight: "160px",
-        fontSize: "11px",
-        bottom: "calc(var(--editor-footer-height))",
+        fontSize: "var(--theme-code-font-size)",
+        bottom: "var(--editor-footer-height)",
       }}
     >
       {matches.map((match, i) => (
--- a/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
+++ b/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
@@ -13,8 +13,8 @@ function Match({
   const buttonNode = useRef<HTMLButtonElement>(null);
 
   useEffect(() => {
-    if (isSelected && buttonNode.current) {
-      buttonNode.current.scrollIntoView({ block: "nearest", inline: "end" });
+    if (isSelected) {
+      buttonNode.current?.scrollIntoView({ block: "nearest", inline: "end" });
     }
   }, [isSelected]);
 
@@ -47,12 +47,12 @@ export default function Autocomplete({
     <div
       className="flex flex-col bg-white border py-1 absolute left-7 -mb-1 shadow-sm font-mono overflow-x-hidden overflow-y-auto"
       style={{
+        bottom: "var(--editor-footer-height)",
+        fontSize: "var(--theme-code-font-size)",
+        marginLeft: `${leftOffset}px`,
+        maxHeight: "160px",
         maxWidth: "200px",
         minWidth: "160px",
-        maxHeight: "160px",
-        fontSize: "var(--theme-code-font-size)",
-        bottom: "var(--editor-footer-height)",
-        marginLeft: `${leftOffset}px`,
       }}
     >
       {matches.map((match, i) => (
--- a/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
+++ b/src/devtools/client/webconsole/components/Input/Autocomplete.tsx
@@ -48,9 +48,10 @@ export default function Autocomplete({
 }) {
   return (
     <div
-      className="flex flex-col bg-white border py-1 absolute left-7 -mb-1 shadow-sm font-mono overflow-auto"
+      className="flex flex-col bg-white border py-1 absolute left-7 -mb-1 shadow-sm font-mono overflow-x-hidden overflow-y-auto"
       style={{
-        minWidth: "240px",
+        maxWidth: "200px",
+        minWidth: "160px",
         maxHeight: "160px",
         fontSize: "var(--theme-code-font-size)",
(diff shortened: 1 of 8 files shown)


src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+
+type ObjectPreviewProperty = {
+  value: ValueFront;
+  writable: boolean;
+  configurable: boolean;
+  enumerable: boolean;
+  name: string;
+  get?: any;
+  set?: any;
+};
```
jcmorrow, line 13: What do you think about putting newlines around these just to give them some separation?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  scopeKind: string;
+  type: "block" | string;
+};
+type FrameScope = {
+  scope: Scope;
+  pending: boolean;
+  originalScopesUnavailable: boolean;
+};
+
+function getMatchString(str: string) {
```
jcmorrow, line 57: I find this a little confusing because the component above is called a `Match`, but this is saying `normalize this string for searching purposes` right? Maybe `normalizeString` would be a better name?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  return str.toLowerCase().replace(/['"`]+/g, "");
```
jcmorrow, line 58: so this strips all quote-like things?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  if (getMatchString(query) === "") {
+    return candidates;
+  }
+  return filter(candidates, query);
+}
+
+function isIdentifier(expression: string) {
+  try {
+    const parsed = babelParser.parseExpression(expression);
+    const isIdentifier = t.isIdentifier(parsed);
```
jcmorrow, line 70: What do you think about doing 
```
import {parseExpression} from "@babel/parser";
import { isIdentifier, isMemberExpression } from "@babel/types";
```
The one-letter global variable threw me.

I would also personally be ok with

```
try {
    return isIdentifier(parseExpression(expression))
} catch (e) {
    return false;
}
```

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    return isIdentifier;
+  } catch (e) {
+    return false;
+  }
+}
+function isMemberExpression(expression: string) {
+  try {
+    const parsed = babelParser.parseExpression(expression);
+    const rv = t.isMemberExpression(parsed);
```
jcmorrow, line 80: same here, I feel like more explicit imports and fewer locals would be easier to read but YMMV

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    return rv;
+  } catch (e) {
+    return false;
+  }
+}
+
+function isBracketNotation(input: string) {
+  const expression = input.slice(0, input.lastIndexOf("[") + 1);
+  return isMemberExpression(expression + `"a"]`);
```
jcmorrow, line 90: This is clever ✏️  

I would maybe make the "a" into like `"property"]`, just to make it clear that `"a"` has no special meaning.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    const rv = t.isMemberExpression(parsed);
+
+    return rv;
+  } catch (e) {
+    return false;
+  }
+}
+
+function isBracketNotation(input: string) {
+  const expression = input.slice(0, input.lastIndexOf("[") + 1);
```
jcmorrow, line 89: if we're extracting to a local for readability here, `expression` doesn't say very much. Maybe `upToLastOpenBracket`? Or `beforeLastOpenBracket`?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  return isMemberExpression(expression + `"a"]`);
+}
+function isDotNotation(input: string) {
+  return isMemberExpression(input + `a`);
```
jcmorrow, line 93: Ah, ok since we're using the `a` thing multiple times, could we pull it out to like

```
// This stands in for a single JS token in places where we don't have a complete expression yet
const TOKEN_PLACEHOLDER = "tokenPlaceholder"
```

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (["Object", "Array"].includes(object.className)) {
+      const prototypeProperties = Object.getOwnPropertyNames(Object.prototype);
+      properties.unshift(...prototypeProperties);
+    }
+
+    if (object.className === "Array") {
+      const prototypeProperties = Object.getOwnPropertyNames(Array.prototype);
+      properties.unshift(...prototypeProperties);
+    }
+  }
```
jcmorrow, line 118: This looks to me like it should be pulled out into a small function that returns from the branches rather than assigning to a local.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  const prototype = preview?.prototypeValue;
+
+  // Recursively gather the properties through the prototype chain.
+  if (prototype?._object) {
+    const prototypeProperties = getPropertiesForObject(prototype._object);
+    properties.unshift(...prototypeProperties);
```
jcmorrow, line 125: Oooooooooooo this is fun. OK, if what I said above gets way harder then maybe don't worry about it 😄 .

But it seems to me like 
```
return [...getLocalProperties(object), ...getLocalProperties(object.preview?.prototypeValue?._object)]```

might work?

File: src/devtools/client/webconsole/utils/autocomplete.ts [hunk 11]
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  return properties;
+}
+
+export function getParentExpression(input: string) {
+  if (isBracketNotation(input)) {
+    return input.slice(0, input.lastIndexOf("["));
```
jcmorrow, line 133: it seems like this and line 89 should be combined somehow.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  const parentExpression = getParentExpression(input);
+
+  if (!parentExpression) {
+    return false;
+  }
+
+  // Right now, this only works for direct property access (e.g., a.b)
+  // and doesn't support chained expressions (e.g., a.b.c.d). This is because
+  // we don't (yet) have the chained values in the fetched frame scope data.
+  return isIdentifier(parentExpression);
```
jcmorrow, line 153: 👍

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+function shouldVariableAutocomplete(input: string) {
+  return isIdentifier(input);
+}
+
+function getBinding(name: string, frameScope: FrameScope) {
+  return frameScope.scope.bindings.find(b => b.name === name);
+}
+function getBindingNames(scope: FrameScope): string[] {
+  if (!scope?.scope) {
```
jcmorrow, line 163: wat.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    return [];
+  }
+
+  return scope.scope.bindings.map(b => b.name);
```
jcmorrow, line 167: maybe `(scope?.scope.bindings || []).map(b => b.name)`

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+function getGlobalVariables(scopes: Scope) {
+  const variableNames = [];
+  const globalObject = scopes.parent.object;
+
+  if (globalObject) {
+    const properties = getPropertiesForObject(globalObject._object);
+    variableNames.push(...properties);
+  }
+
+  return variableNames;
```
jcmorrow, line 178: `return getPropertiesForObject(scopes.parent.object?._object)` works if you change to using `object?.preview` in `getPropertiesForObject` (which I think js a good change anyways).

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+}
+
+export function getLastToken(input: string) {
+  if (isBracketNotation(input)) {
+    return input.slice(input.lastIndexOf("[") + 1);
```
jcmorrow, line 183: Yeah this is a repetition of the logic above, feels like there is a clean abstraction in here somewhere?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  } else {
+    return input.slice(input.lastIndexOf(".") + 1);
+  }
+}
+
+export function getCursorIndex(value: string) {
```
jcmorrow, line 189: This is an interesting one.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    return getParentExpression(value).length + 1;
+  } else {
+    return 0;
+  }
+}
+
+export function insertAutocompleteMatch(value: string, match: string) {
+  const parentExpression = getParentExpression(value);
+
+  if (shouldPropertyAutocomplete(value)) {
```
jcmorrow, line 201: Oh, I was reading this as `if this returns false we don't autocomplete`, but actually it's `if this returns false we variable autocomplete`?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+      return `${parentExpression}.${match}`;
+    }
+  } else {
+    // For variable autocomplete.
+    return match;
+  }
+}
+
+export function getAutocompleteMatches(input: string, frameScope: FrameScope) {
+  if (!frameScope?.scope) {
```
jcmorrow, line 214: What does it mean when `frameScope?.scope` is falsey?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    return [];
+  }
+
+  const bindingNames = getBindingNames(frameScope);
```
jcmorrow, line 218: I think this should be down in the branch where it's used maybe?

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (!object) {
+      return [];
+    }
+
+    const properties = getPropertiesForObject(object);
+
+    const filteredProperties = fuzzyFilter(properties, getMatchString(lastToken));
+    return filteredProperties.map(p => (computedProperty ? `"${p}"` : p));
+  } else if (shouldVariableAutocomplete(input)) {
+    const variableNames = [...bindingNames, ...getGlobalVariables(frameScope.scope)];
```
jcmorrow, line 242: And honestly feels like it could just be `getBindingNames(frameScope)` here. No need for the local.

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+    const filteredNames = fuzzyFilter(variableNames, getMatchString(input));
+
+    return filteredNames;
```
jcmorrow, line 245: Idk, I would probably remove the local here, but it's up to you!

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  return [];
```
jcmorrow, line 248: This feels like it should be a guard.

jcmorrow (commented):
I left a lot of comments, but don't panic! I think this is really great and surprisingly understandable given how complicated it is! LMK if you want to talk through any of this!

src/devtools/client/webconsole/components/Input/Autocomplete.tsx
```diff
@@ -0,0 +1,72 @@
...
+  selectedIndex,
+  index,
+  onClick,
+}: {
+  label: string;
+  selectedIndex: number;
+  index: number;
+  onClick: (index: number) => void;
+}) {
+  const buttonNode = useRef<HTMLButtonElement | null>(null);
```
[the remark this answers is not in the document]
jaril: That was specifically added because TS was yelling at me for something that I can't quite remember. It's gone now so off it goes!

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: 👍 yep, and added comments to make it clear what it's for.

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: yep!

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: good call

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: lol almost too clever but it works really well 🤷‍♂️

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: Redundant guard I added early on. Doesn't look like it's needed though.

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: simplified this so we don't need the fallback!

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: Simplified so we can hand wave this away

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: What do you mean here?

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: Exactly. The variable autocomplete was implicit in the else so I added a comment. How would you do it?

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: Yeah. Can't think of a way to simplify this right now though, I'll sleep on it.

src/devtools/client/webconsole/utils/autocomplete.ts
jaril: Yeah, sleepin on it

src/devtools/client/webconsole/utils/autocomplete.ts
jcmorrow: It's probably a little late for this refactoring, but I think I would have three *types* of auto-complete: 

```
enum LineCompletion {
  DotProperty,
  BracketProperty,
  Variable
}
```

And then have each one handle its own matching/generating a new line. The idea here is that we might not know if this is *every* type of autocomplete or not. For instance, what if we added some built-in commands that didn't quite follow this pattern, now we have another set of nested `if/else`s to add to a bunch of places in this file. If we reframe the problem into "there are at least three types of auto-complete, and we're not 100% certain that there will never be any more, but we *are* certain that every auto-completion line will only ever be *one* type)" then this code gets easier to extend later. If you want to pair on making this change lmk, but also, if you would rather move on with life I will understand 😄

src/devtools/client/webconsole/utils/autocomplete.ts
jcmorrow: I would probably have a tiny data structure that looked like:

```
{
  left: "someObject",
  right: "somePartiallyWrittenPropertyName",
  separator: "."
}
```
Or you could do `parent/child`, because I think that's another metaphor we use in here somewhere.

jcmorrow (commented):
Couple of more thoughts, but again, feel free to take or leave!

src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+function isIdentifier(expression: string) {
+  try {
+    return _isIdentifier(parseExpression(expression));
+  } catch (e) {
+    return false;
+  }
+}
+function isMemberExpression(expression: string) {
```
jcmorrow, line 85: Actually, the more I stare at this, the more I realize we need to ping @loganfsmyth , because there's no way we're going to do a better job than babel here.

For instance,

`foo["some.long[property`

is a valid thing to try and autocomplete that I think our logic totally chokes on. I'm guessing that using something like `parseExpression` (and maybe chopping off characters from the end until it stops raising for invalid syntax) will tell you: the last valid character (which indicates the kind of access we're trying to do), and also the type of the left-term, which tells us whether we should be trying to autocomplete right now.

jcmorrow (approved):
🌟

PR #5054, opened 2022-01-20
Hide whitespace nodes in the markup panel

hbenl:
Firefox and Chrome devtools don't show whitespace text nodes by default (since they're rarely useful), so I think we shouldn't either.
This PR hides them and also makes sure that keyboard navigation in the markup panel doesn't select a hidden node.


abstractalgo:
Does it make sense to filter out whitespace nodes immediately after we load the children? That we don't have to be mindful about intentionally skipping certain types of nodes while traversing them, so might be cleaner/easier.

hbenl:
Ah, yes, that's much simpler.

abstractalgo (approved):

PR #5055, opened 2022-01-20
Fix the splitter styling in the elements panel

hbenl:
Before:
Screenshot from 2022-01-20 14-46-18

After:
Screenshot from 2022-01-20 14-46-01

### DIFF
--- a/src/devtools/client/inspector/components/App.tsx
+++ b/src/devtools/client/inspector/components/App.tsx
@@ -200,7 +200,7 @@ class InspectorApp extends Component<PropsFromRedux> {
             minSize="20%"
             maxSize="80%"
             onMove={this.onSplitboxResize}
-            splitterSize={2}
+            splitterSize={4}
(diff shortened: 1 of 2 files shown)


abstractalgo:
I went for a leaner splitter in #4882 to stay consistent with other panel dividers.

image

What do you think?

---

`SplitBox`'s resize handle should have a splitter that you can style and resize and an actual area that handles mouse events that's overlaid on top and is invisible, however, but that would require a lot more changes and checking that we didn't break something else.

hbenl:
I agree that overlaying an invisible area for mouse events would be ideal but nothing we have to fix right now.
My main concern with the current splitter is that it's so narrow that I find it frustrating to use. But it doesn't have to be `8px`, `4px` should be enough:
Screenshot from 2022-01-20 15-14-55

abstractalgo (approved):

PR #5057, opened 2022-01-20
Fix hovering on markup breadcrumbs

hbenl:
I broke this in #5035.


abstractalgo (approved):
both of us at fault for it then 😄

PR #5059, opened 2022-01-20
Fix exception when accessing a source from an evaluation

hbenl:


src/devtools/client/debugger/src/components/Editor/SourcemapToggle.tsx
```diff
@@ -9,7 +9,7 @@ import { Switch } from "@headlessui/react";
 import classNames from "classnames";
 import { setModal } from "ui/actions/app";
 
-export const isNextUrl = (url: string) => url.includes("/_next/");
+export const isNextUrl = (url: string | undefined) => url && url.includes("/_next/");
```
jaril, line 12: Can't we do `url?: string` here?

jaril (approved):
One nit but looks good otherwise!

src/devtools/client/debugger/src/components/Editor/SourcemapToggle.tsx
hbenl: Sure, that would be possible. There is a slight difference though: with `url?: string` we'd allow calling `isNextUrl()` without an argument, which makes little sense. With `url: string | undefined` you always have to provide an argument, even if that argument may be `undefined`. That's why I preferred this form.

PR #5060, opened 2022-01-20
Don't update the graphics when the mouse leaves the timeline

hbenl:
The graphics was updated unnecessarily when the mouse leaves the timeline. With repainting enabled, this meant that the repainted screenshot was replaced by the recorded screenshot.


jcmorrow (approved):
Seems reasonable to me 👍

PR #5070, opened 2022-01-21
Event Listeners panel

abstractalgo:


abstractalgo:
A few observations from this example and reusing the code that was previously there for event listeners tooltip:
image

The function name is not properly captured. Even though we could try to get the symbol name like we're doing in the `Outline` section, these "incorrect" names seem to be on par with how Chrome detects them as well.

We wrongly detect something as a "native code", i.e. we should be able to get to the function in the source code (why it's not working then?).

hbenl (approved):
LGTM!

PR #5074, opened 2022-01-21
Add a simple timings panel

jcmorrow:
We waited on this because when I first tried to verify timings against what Chrome and Firefox were reporting I was not seeing the same numbers, but hitting a server where I am manually controlling the time that the server takes to respond, everything looks good. That makes me think there were just legitimate time differences in the requests I was recording before.

And remove Coming Soon because all tabs are populated!
CleanShot 2022-01-21 at 13 50 35

Other Changes
---
- There was an errant `0` when the request had no query params
- Remove the "Coming soon!" component, because all tabs are populated 😄 
- Try and standardize the left-padding that all of the tabs use (this is hard because some have collapsed/expanded state to show and others don't)


jcmorrow:
@ceceliacreates I'm thinking a Docs change of something like:

> The timings panel allows you to see the recording time that the request was initiated, as well as when the request returned (note: this is when the response first came back from the server, and does not include time spent transferring the *body* of the request).

And a screenshot like the other sections (feel free to use the one above or one that matches the others better).

abstractalgo (approved):

PR #5075, opened 2022-01-21
Fix updating the slug when the recording title changes

ryanjduffy:
## Issue

Changing the recordings title does not immediately update the slug in the URL although returning to the library and then back to the recording does fix it.

## Resolution

* Refetch the `GetRecording` query when the title changes
* Refactor the slug generation into a new hook that uses `useGetRecording` to ensure the refresh.

## Notes

While there is a local `recording` object in the component, we're getting it via a redux action which isn't refreshed when the title changes. We could also refactor that code to refresh itself (any maybe that's the better solution) but I don't think the result is much different.

### DIFF
--- a/pages/recording/[id].tsx
+++ b/pages/recording/[id].tsx
@@ -5,7 +5,7 @@ import { connect, ConnectedProps, useStore } from "react-redux";
 import { isTest } from "ui/utils/environment";
 import { getAccessibleRecording } from "ui/actions/session";
 import { Recording as RecordingInfo } from "ui/types";
-import { getRecordingMetadata, useGetRecordingId } from "ui/hooks/recordings";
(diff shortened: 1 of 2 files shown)


pages/recording/[id].tsx
```diff
@@ -71,24 +71,9 @@ function RecordingHead({ metadata }: MetadataProps) {
...
-  useEffect(() => {
-    if (!store || !recordingId) return;
-
-    async function getRecording() {
-      await setup(store);
-      setRecording(await getAccessibleRecording(recordingId));
-    }
-    getRecording();
-  }, [recordingId, store, getAccessibleRecording]);
+  const { recording } = useGetRecording(recordingId);
```
Comment (jcmorrow) [inline pages/recording/[id].tsx:76]: This is a welcome abstraction 😄

jcmorrow (approved):
I think this makes sense 🏷️

PR #5076, opened 2022-01-21
Split out the namespace of MixPanel events

jcmorrow:
It's actually really hard to pull this information out in a useful way
via MixPanel reporting when it is literally part of the event name, but
by also adding it to the event as a discrete field it should be easier
to do things like break down a particular report by namespace.

### DIFF
--- a/src/ui/utils/mixpanel.ts
+++ b/src/ui/utils/mixpanel.ts
@@ -50,13 +50,19 @@ export const maybeTrackTeamChange = (newWorkspaceId: string | null) => {
   }
 };
 
+const NULL_NAMESPACE = "no_namespace";
+const namespaceFromEventName = (event: string): string => {
+  const namespace = event.slice(0, event.indexOf("."));
+  return namespace.length ? namespace : NULL_NAMESPACE;
+};
+
 export async function trackMixpanelEvent(event: string, properties?: Dict) {
   if (prefs.logTelemetryEvent) {
     console.log("🔴", event, properties);
   }
 
   if (!mixpanelDisabled) {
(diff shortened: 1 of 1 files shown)


src/ui/utils/mixpanel.ts
```diff
@@ -55,8 +55,14 @@ export async function trackMixpanelEvent(event: string, properties?: Dict) {
     console.log("🔴", event, properties);
   }
 
+  let namespace;
```
jaril, line 58: maybe simplify to something like this?

```js
const namespaces = event.split(".");

mixpanel.track(
  event, 
  {
    ...properties, 
    namespace: nameSpaces.length ? nameSpace : null
  }
}
```
```

jaril (approved):
Looks good! Also having the no-namespace namespace (whether that's undefined/null) should make it easy to identify which events are being hit without a namespace.

src/ui/utils/mixpanel.ts
jasonLaster: as long as we are code golfing

`namespace: event.indexOf(".") ? event.split(".")[0] : ""`

abstractalgo:
Having strong TS types for this would be really good as to prevent many silly mistakes that would then make it harder to parse the information on the Mixpanel side.

Using [template literals](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) might be recommended here, as it would allow correct matching of event types to their corresponding "namespaces", yet be a single string; but going for a separate, distinctive property might be even better.

jcmorrow:
Merging this and then following up with types as discussed in #5089

PR #5077, opened 2022-01-21
Track time spent in specific layouts

jcmorrow:
This is a stupid idea that I probably won't merge but wanted to share with @jaril

### DIFF
--- a/src/ui/utils/mixpanel.ts
+++ b/src/ui/utils/mixpanel.ts
@@ -5,6 +5,10 @@ import { ViewMode } from "ui/state/layout";
 import { isReplayBrowser, skipTelemetry } from "./environment";
 import { prefs } from "./prefs";
 import { TelemetryUser, trackTiming } from "./telemetry";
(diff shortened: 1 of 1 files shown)


abstractalgo (commented):
Mixpanel and similar tools queue events, so might as well send an event every 1s. Not sure if this reaches some limits for the Mixpanel pricing plan we're on, but if we want granularity... :D

This would also be fun to see as "a total usage" metric (e.g. per month).

jasonLaster:
haha - probably learn more by watching more logrocket sessions :)

jcmorrow:
> Mixpanel and similar tools queue events, so might as well send an event every 1s. Not sure if this reaches some limits for the Mixpanel pricing plan we're on, but if we want granularity... :D
> 
> This would also be fun to see as "a total usage" metric (e.g. per month).

At some point we had a bug where one user sent 400K events one day. So I think we're safe for now 😆

PR #5079, opened 2022-01-21
Add trackers to the network panel

### LINKED ISSUE #5078: Add mixpanel trackers to the network panel

jaril:
Fix #5078.

This ought to help us have more visibility in the [network panel usage dashboard](https://mixpanel.com/project/2349904/view/2894968/app/dashboards#id=2775967&focused-bookmark-id=27783645)

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestTable.tsx
+++ b/src/ui/components/NetworkMonitor/RequestTable.tsx
@@ -5,6 +5,7 @@ import { RequestSummary } from "./utils";
 import { HeaderGroups } from "./HeaderGroups";
 import { RequestRow } from "./RequestRow";
 import { Row, TableInstance } from "react-table";
+import { safeTrackEvent } from "ui/utils/mixpanel";
 
 const RequestTable = ({
   className,
@@ -26,6 +27,7 @@ const RequestTable = ({
   const { columns, getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = table;
 
   const onSeek = (request: RequestSummary) => {
+    safeTrackEvent("net_monitor.seek_to_request");
     seek(request.point.point, request.point.time, true);
     onRowSelect(request);
   };
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -19,6 +19,8 @@ import Table from "./Table";
 import { fetchFrames, fetchResponseBody, fetchRequestBody } from "ui/actions/network";
 import { getThreadContext } from "devtools/client/debugger/src/selectors";
 import LoadingProgressBar from "../shared/LoadingProgressBar";
+import mixpanel from "mixpanel-browser";
+import { safeTrackEvent } from "ui/utils/mixpanel";
 
 export const NetworkMonitor = ({
   currentTime,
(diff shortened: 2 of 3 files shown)


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -62,13 +66,16 @@ export const NetworkMonitor = ({
...
   if (loading) {
+    mixpanel.time_event("netMonitor.open_network_monitor");
     return (
       <div className="relative">
         <LoadingProgressBar />
       </div>
     );
   }
 
+  trackEvent("netMonitor.open_network_monitor");
```
jaril, line 77: This one's fun — we'll be able to inspect this event and find out how long people are staring at the progress bar.

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -26,6 +27,7 @@ const RequestTable = ({
   const { columns, getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = table;
 
   const onSeek = (request: RequestSummary) => {
+    trackEvent("netMonitor.seek_to_request");
```
jcmorrow, line 30: maybe standardize on `netMonitor` vs. `netmonitor`. Do you think this would be worth pulling out into constants?  (I think maybe I saw a PR from @abstractalgo  about this?)

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -62,13 +66,16 @@ export const NetworkMonitor = ({
   }, [container.current]);
 
   if (loading) {
+    mixpanel.time_event("netMonitor.open_network_monitor");
```
jcmorrow, line 69: Ooooo interesting - I haven't seen this before. Do we need to guard against situations where tracking is disabled?

src/ui/components/NetworkMonitor/index.tsx
jcmorrow: Ah, this kind of answers my question above!

jcmorrow: Wait I want this everywhere now.

jcmorrow (approved):
Couple of small things but this looks great! 📐

src/ui/components/NetworkMonitor/RequestTable.tsx
abstractalgo: I did mention in the Discord thread ([private thread](https://discord.com/channels/779097926135054346/934430769135362049)) that we most definitely should add type safety around tracking.

abstractalgo:
I don't want to be "that guy", but I would really prefer not to merge this before adding types. 😄

jaril:
> I don't want to be "that guy", but I would really prefer not to merge this before adding types. 😄

I understand why but I'll take the other side — what's important is getting the network monitor usage data. I'd rather get this in sooner than later if it means we won't miss 1-2 days worth of user data.

Maybe there's a compromise. We could implement the types now but limit it to these trackers, and then we can convert all of the other mixpanel trackers separately?

@abstractalgo

abstractalgo:
@jaril 

I agree completely to merge this asap, and as we have a separate ticket for improving tracking, then it's more than fine IMO. 👍

PR #5085, opened 2022-01-24
Allow all Elements panels to be scrollable

abstractalgo:
The change affects:

- Event Listeners panel
- Layout panel

Before:

https://user-images.githubusercontent.com/1355455/150826526-a8bc97e7-0ef7-4b4a-ba43-76d399707afa.mp4

After:

https://user-images.githubusercontent.com/1355455/150826180-2a9ec885-30b1-46e6-bb4d-5c5149d90e69.mp4


abstractalgo:
Tagged @jaril because of the Accordion stuff (might not be super relevant).

hbenl (approved):

src/devtools/client/debugger/src/components/shared/Accordion.css
```diff
@@ -10,7 +10,7 @@
   display: flex;
   flex-direction: column;
   overflow-y: hidden;
-  height: 100%;
+  min-height: 100%;
```
jcmorrow, line 13: This does break (or at least significantly change) the behavior of the pause panel accordion.

jcmorrow: Production is on the right, the preview for this branch is on the left.

https://user-images.githubusercontent.com/5903784/150836580-7b24f966-6eae-4a25-83cb-b17f3e480288.mp4

jcmorrow: I think that the preview branch version is a bit easier to use, but also feels broken if you are expecting a true accordion. We can always add a special case to either of these to keep them working.

abstractalgo: Great catch, thanks! 🙌 

I'll reconsider what might be the best approach (specific selector for Layout accordion only; or something else).

abstractalgo:
Closing in favor of #5128 .

PR #5093, opened 2022-01-24
Allow team members to change the privacy of a recording that belongs to their team

### LINKED ISSUE #4762: Team members should be able to change replay privacy
**Description:**

It should be possible for me as a replay team member to be able to make a replay that someone else created private or public. I should also be able to to change the title.

jaril:
Fix #4762.

Demo: https://share.descript.com/view/HKbG6afjkbC

This gets a fair bit messy because of how any dropdown option can potentially affect either the public/private, or the replay's team, or both.

This makes it so that if you're the owner of a replay, you're allowed to a) make the replay public/private, and b) move the replay to your library or any of the teams you belong to. If you're a team member who belongs to the same team that the replay does, you're allowed to make the replay public/private.

**Case 1: Replay owner**
```
- Anyone with the link (Toggle public)
- People with access (Toggle private **and move to My Library**)
- Team 1 (Toggle private **and move to Team 1**)
- Team 2
- Team 3
```

**Case 2: Replay team member**
```
- Anyone with the link (Toggle public)
- People with access (Toggle private)
```

**Case 3: No edit permissions, e.g. public team replay for a team the user doesn't belong to, or public replay that doesn't belong to a team**
```
No options, since the dropdown is disabled.
```

### DIFF
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { Dispatch, ReactNode, SetStateAction, useState } from "react";
 import PortalDropdown from "../PortalDropdown";
 import { Dropdown, DropdownItem, DropdownItemContent } from "ui/components/Library/LibraryDropdown";
 import { WorkspaceId } from "ui/state/app";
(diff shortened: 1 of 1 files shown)


jaril:
This still needs a fair bit of a clean up!

jaril:
@ceceliacreates 👋

jaril:
Tested and it all works. Added the [demo](https://share.descript.com/view/HKbG6afjkbC) in the description.

ceceliacreates:
Hey @jaril just to confirm this is only from the Share button within the Replay Browser, correct? I don't see the ability to share a Replay on the same team from the Replay Library.

Screen Shot 2022-01-26 at 10 53 32 AM

PR #5095, opened 2022-01-24
Use the correct recordingID for creating console comments

### LINKED ISSUE #5094: When submitting data to the GraphQL server, the recording ID should never include the slug
This is what was causing https://app.replay.io/recording/stubhub-expired-listing--b29bb7ab-a072-4f34-90fa-1b8d6104a501.

For some reason when adding comments via the console we get a URL that includes the slug:

CleanShot 2022-01-24 at 14 51 48@2x

jcmorrow:
After #5021 `router.query.id` now returns a slug as well as a recordingID. I've tried grepping through the code to see if this will bite us anywhere else but I didn't find anything.

This also fixes this bug:

https://user-images.githubusercontent.com/5903784/150882189-c85a62f0-a1b9-4aec-9ab0-718e1dd6f36b.mp4

Fixes #5094

### DIFF
--- a/src/ui/hooks/comments/useAddComment.tsx
+++ b/src/ui/hooks/comments/useAddComment.tsx
@@ -25,7 +25,6 @@ export default function useAddComment() {
   }
 
   return (comment: Comment) => {
-    const temporaryId = new Date().toISOString();
     trackEvent("comments.create");
 
     addComment({
@@ -39,7 +38,7 @@ export default function useAddComment() {
         addComment: {
           success: true,
           comment: {
-            id: temporaryId,
+            id: comment["id"],
             __typename: "Comment",
           },
           __typename: "AddComment",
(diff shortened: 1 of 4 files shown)


src/ui/hooks/comments/useAddComment.tsx
```diff
@@ -25,7 +25,6 @@ export default function useAddComment() {
   }
 
   return (comment: Comment) => {
-    const temporaryId = new Date().toISOString();
```
jcmorrow, line 28: @jaril can you remember why we do this? I couldn't remember and removing it seemed fine 🤷

jasonLaster (approved):
Nice. Will let @jaril do a proper review, but excited to see the UI be smoother too

ryanjduffy (approved):

src/ui/hooks/comments/useAddComment.tsx
jcmorrow: I'm going to assume this is fine because we already removed it for replies in previous commit.

PR #5099, opened 2022-01-25
Make analytics calls type-safe

jcmorrow:
This builds on the change @jaril started in #5079. Now, when adding calls to MixPanel either use an existing event, or add a new event type to the event list in `utils/mixpanel.ts`. This is nice for a few reasons:

- TSC will guarantee that there is no typo in the event names/details
- When adding a new event call, all *other* event calls are visible, so it's less likely that someone will add something like `open_dev_tools` when there is already an `open-dev-tools` event or something else very similar. (This is also a good example of the reason that I like to sort arrays and objects by default).

One thing I still have to figure out - we have a fair amount of events that did not conform to the format of things like underscores, common namespaces, etc. I think maybe I can make a mapping from the old events onto the new events and upload that to MixPanel, I'll figure that out tomorrow. Obviously it would be nice to have both clean data going forward and also historical data without having to account for differences across what events were called over time.

Fixes https://github.com/RecordReplay/devtools/issues/5089

### DIFF
--- a/src/ui/utils/mixpanel.ts
+++ b/src/ui/utils/mixpanel.ts
@@ -6,13 +6,82 @@ import { isReplayBrowser, skipTelemetry } from "./environment";
 import { prefs } from "./prefs";
 import { TelemetryUser, trackTiming } from "./telemetry";
 import { CanonicalRequestType } from "ui/components/NetworkMonitor/utils";
+import { PrimaryPanelName, SecondaryPanelName, WorkspaceId } from "ui/state/app";
 
 type MixpanelEvent =
+  | ["breakpoint.minus_click"]
+  | ["breakpoint.plus_click"]
+  | ["breakpoint.preview_has_hits"]
+  | ["breakpoint.preview_no_hits"]
+  | ["breakpoint.remove"]
+  | ["breakpoint.set_condition"]
+  | ["breakpoint.set_log"]
+  | ["comments.create"]
+  | ["comments.delete"]
+  | ["comments.focus"]
+  | ["comments.select_location"]
+  | ["comments.start_edit"]
+  | ["console.clear_messages"]
+  | ["events_timeline.select"]
+  | ["events_timeline.select_source"]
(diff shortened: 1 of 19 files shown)


src/ui/utils/mixpanel.ts
```diff
@@ -6,13 +6,89 @@ import { isReplayBrowser, skipTelemetry } from "./environment";
...
+  | ["toolbox.secondary.video_toggle"]
+  | ["toolbox.toggle_sidebar"]
+  | ["upload.complete", { sessionId: SessionId }]
+  | ["upload.create_replay", { isDemo: boolean }]
+  | ["upload.discard", { isDemo: boolean }]
+  | ["user_options.launch_replay"]
+  | ["user_options.select_docs"]
+  | ["user_options.select_settings"]
+  | [`toolbox.primary.${PrimaryPanelName}_select`]
+  | [`toolbox.secondary.${SecondaryPanelName}_select`];
```
abstractalgo, line 91: Can move all `toolbox.` up to be together.

abstractalgo (approved):
this is so nice 😌

src/ui/utils/mixpanel.ts
jcmorrow: yeah `sort` got me because of the different quotes

src/ui/utils/mixpanel.ts
```diff
@@ -6,13 +6,89 @@ import { isReplayBrowser, skipTelemetry } from "./environment";
...
+  | ["share_modal.set_private"]
+  | ["share_modal.set_public"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
+  | ["share_modal.set_team"]
```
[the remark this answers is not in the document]
jcmorrow: lol, thanks for the catch 👍

PR #5101, opened 2022-01-25
Keep prefs for the selected panel (console, network, etc) on a per replay basis

### LINKED ISSUE #5081: Persist selected secondary panel tab
**Description:**
Small pet peeve, but when you refresh and the elements tab is selected it is no longer remembered. I believe it should be in the saved replay state.

jaril:
Fix #5081.

### DIFF
--- a/src/devtools/client/debugger/panel.js
+++ b/src/devtools/client/debugger/panel.js
@@ -5,7 +5,7 @@
 import { assert } from "protocol/utils";
 import { openDocLink } from "devtools/client/shared/link";
 import { onConnect } from "devtools/client/debugger/src/client";
-import { setSelectedPanel } from "ui/actions/app";
+import { setSelectedPanel } from "ui/actions/layout";
 
 export class DebuggerPanel {
   constructor(toolbox) {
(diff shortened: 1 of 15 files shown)


src/devtools/client/debugger/panel.js
```diff
@@ -5,7 +5,7 @@
 import { assert } from "protocol/utils";
 import { openDocLink } from "devtools/client/shared/link";
 import { onConnect } from "devtools/client/debugger/src/client";
-import { setSelectedPanel } from "ui/actions/app";
+import { setSelectedPanel } from "ui/actions/layout";
```
jcmorrow, line 8: 👍

src/ui/reducers/layout.ts
```diff
@@ -1,7 +1,7 @@
 import { UIState } from "ui/state";
 import { LayoutState } from "ui/state/layout";
 import { LayoutAction } from "ui/actions/layout";
-import { asyncStore } from "../utils/prefs";
+import { asyncStore, prefs } from "../utils/prefs";
```
jcmorrow, line 4: It looks like maybe you don't need to add this import?

jcmorrow (approved):
I fully support moving those actions and selectors as well. I think if I'm being a little pedantic it might be nice to have the moving things around in one commit, and then the change of where things get stored in another commit, just to make it clear where the magic happens, but either way is fine, these are not huge changes.

jcmorrow:
> Fix #5101.

I think you linked the wrong thing FYI, *this* is 5101 😉

jcmorrow (approved):
:shipit:

PR #5103, opened 2022-01-25
Fix hovering in the tooltip smart gap

### LINKED ISSUE #4774: Symbol tooltips closed too eagerly
**Description:**

It happens for me quite often that a tooltip that gets displayed when I hover over a symbol~ gets closed when I try to move my pointer to it.

Consider this situation
Screenshot 2021-12-07 at 22 38 32

When I see this I intuitively move my cursor towards "entries" in the tooltip, so I can try to expand that and inspect the content of this data. However, that name is quite short and the tooltip is somewhat wide and my pointer trajectory is connecting those 2 points in a straight line - this makes my pointer leave the source symbol area for a short moment without yet entering the tooltip itself. 

You naturally must close the tooltip when I move my pointer out of the source element and the tooltip - but it seems that timings and stuff could be fine-tuned here because it really happens quite often for me (not every time though!)

**Expected Result:**

I should be able to move my cursor to the tooltip content without the tooltip being closed - if I do that reasonably fast.

jaril:
Fix #4774.

Turns out we had this fancy smart gap polygon thing for hovering between the gap of the token and the tooltip, but the hovering wasn't being handled correctly since it (`elementIsHovered()`) relied on the bounding box. This was fixed when we just checked if the gap had matched the `:hover` selector instead.

### DIFF
--- a/src/devtools/client/debugger/src/components/shared/Popover.js
+++ b/src/devtools/client/debugger/src/components/shared/Popover.js
@@ -57,7 +57,7 @@ class Popover extends Component {
   }
 
   onTimeout = () => {
-    const isHoveredOnGap = this.$gap && window.elementIsHovered(this.$gap);
+    const isHoveredOnGap = this.$gap && this.$gap.matches(":hover");
     const isHoveredOnPopover =
       this.$popoverPreview && window.elementIsHovered(this.$popoverPreview);
     const isHoveredOnTooltip = this.$tooltip && window.elementIsHovered(this.$tooltip);


src/devtools/client/debugger/src/components/shared/Popover.js
```diff
@@ -57,7 +57,8 @@ class Popover extends Component {
   }
 
   onTimeout = () => {
-    const isHoveredOnGap = this.$gap && window.elementIsHovered(this.$gap);
+    const isHoveredOnGap =
+      this.$gap && [...this.$gap.firstElementChild.classList].includes("hovered");
```
jcmorrow, line 61: I'm not a big fan of using these `onMouseEnter` `onMouseLeave` things. Is it possible to use `.matches(':hover')` instead?

src/devtools/client/debugger/src/components/shared/SmartGap.js
```diff
@@ -98,36 +101,41 @@ export default function SmartGap({ token, preview, type, gapHeight, coords, offs
...
-  const coordinates = getSmartGapCoordinates(
-    previewRect,
-    tokenRect,
-    -offset,
-    orientation,
-    gapHeight,
-    coords
-  );
+  React.useEffect(() => {
+    // $FlowIgnore
```
jcmorrow, line 105: What's up here?

jcmorrow (commented):
Couple of thoughts, may or may not be useful!

jasonLaster:
@jaril can you add some before/after replays

src/devtools/client/debugger/src/components/shared/Popover.js
jaril: oh boy i love it when this happens. yep that got it down to a single line.

jaril:
Now it's a one line fix. @jcmorrow you're a superstar.

Before: https://app.replay.io/recording/smart-gap-not-working--dadd5e2e-27a0-4366-90cb-48832a274d16
After: https://app.replay.io/recording/smart-gap-working--81767e25-6e18-4b7e-9321-6fed55a51656

PR #5115, opened 2022-01-25
Try and make timings more specific and flexible

jcmorrow:
The approach that I first took on the timings panel assumed that if the request was complete we would have a response event with a time. It turns out that some requests will have a request-done event without a response event, and others might have a response event without a request-done event. (See https://static.replay.io/protocol/tot/Network/#type-RequestEventInfo for more info on what all of these events mean).

Now, we will try and show timing information, but if we don't have a `firstByte` time or an `end` time we won't show those fields in the summary or timing tab.

Replay: https://app.replay.io/recording/network-timings-are-more-accurate--911f42cf-8f53-434b-a67d-2b3140edaf97

Fixes https://github.com/RecordReplay/devtools/issues/5114


jasonLaster (approved):

jasonLaster:
Could you share a replay of it working? Would be nice for posterity

PR #5118, opened 2022-01-26
Some polish on our replay info pane

jonbell-lot23:
Here's a 38 second Loom showing what I changed:
https://www.loom.com/share/6dabbabf92e44b2ea4da9d704a97c0dd

The links are now more consistent.

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -39,7 +39,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
 
   return (
     <div className="flex-grow overflow-auto overflow-x-hidden flex flex-column items-center bg-white border-splitter">
-      <div className="flex flex-col p-1.5 self-stretch space-y-1.5 w-full text-xs group">
+      <div className="flex flex-col p-1.5 self-stretch space-y-1.5 w-full text-xs">
         {recording.user ? (
           <Row>
             <AvatarImage className="h-5 w-5 rounded-full avatar" src={recording.user.picture} />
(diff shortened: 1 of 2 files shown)


src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -51,12 +51,20 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
...
         </Row>
-        <Row>
-          <MaterialIcon iconSize="xl">language</MaterialIcon>
-          <div className="overflow-hidden overflow-ellipsis whitespace-pre" title={recording.url}>
-            {getDisplayedUrl(recording.url)}
-          </div>
-        </Row>
+
+        <div className="group">
+          <Row>
```
abstractalgo, line 56: Suggestion: maybe add a `className` prop to `Row` components, and just pass `group` value to it instead of creating an additional `div.group`?

src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -51,12 +51,20 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
...
+            <MaterialIcon iconSize="xl" className="group-hover:text-primaryAccent">
+              language
+            </MaterialIcon>
+            <div className="overflow-hidden overflow-ellipsis whitespace-pre" title={recording.url}>
+              <a href={recording.url} target="_blank">
+                {getDisplayedUrl(recording.url)}
+              </a>
```
abstractalgo, line 63: Generally, it would be good to add `rel="noopener noreferrer"` to all `<a>` tags.

(related: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

jaril: On a non-blocking note, I could have sworn we had a <Link> component from a few months ago just so we didn't have to worry about this bit. Will make a point to check that out sometime today.

jasonLaster:
@jonbell-lot23 I like it! One other small win would be to make the share/permissions link open the share modal so you can change who can view the replay.

jonbell-lot23:
> @jonbell-lot23 I like it! One other small win would be to make the share/permissions link open the share modal so you can change who can view the replay. You should be able to do this with a redux action 

Ah great idea. I have nearly zero experience with redux so this will be fun to learn

src/ui/components/Events/ReplayInfo.tsx
jonbell-lot23: Fixed!

jonbell-lot23:
Moving the share link feature request here: https://github.com/RecordReplay/devtools/issues/5124

PR #5119, opened 2022-01-26
Polished events rows

jonbell-lot23:
Here's a 30 second demonstration of what I changed:
https://www.loom.com/share/27d4bdb715ed4c52a4cd8e6bb6a71022

The red line in our events was inelegantly touching the grey background, so I added some more space to look nicer.

### DIFF
--- a/src/ui/components/Events/Event.tsx
+++ b/src/ui/components/Events/Event.tsx
@@ -65,7 +65,7 @@ export default function Event({
       onKeyDown={onKeyDown}
       className={classNames(
         "event flex flex-row justify-between items-center mb-1 mt-1 user-select-none",
-        "group block p-1 pl-3 pr-2 w-full rounded-lg hover:bg-gray-100 focus:outline-none cursor-pointer",
+        "group block py-1 pl-3 pr-2 w-full rounded-lg hover:bg-gray-100 focus:outline-none cursor-pointer",
         {
           "text-lightGrey": currentTime < time,
           "text-primaryAccent font-semibold": isPaused,
--- a/src/ui/components/Events/Event.tsx
+++ b/src/ui/components/Events/Event.tsx
@@ -64,8 +64,8 @@ export default function Event({
       onClick={onClick}
       onKeyDown={onKeyDown}
       className={classNames(
-        "event flex flex-row justify-between items-center space-x-2 user-select-none",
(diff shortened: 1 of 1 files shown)


src/ui/components/Events/Event.tsx
```diff
@@ -64,8 +64,8 @@ export default function Event({
       onClick={onClick}
       onKeyDown={onKeyDown}
       className={classNames(
-        "event flex flex-row justify-between items-center space-x-2 user-select-none",
-        "group block p-3 w-full rounded-lg hover:bg-gray-100 focus:outline-none cursor-pointer",
+        "event flex flex-row justify-between items-center mb-1 mt-1 user-select-none",
+        "group block p-1 pl-3 pr-2 w-full rounded-lg hover:bg-gray-100 focus:outline-none cursor-pointer",
```
jcmorrow, line 68: Since `pl-3` handles left and `pr-2` handles right it's probably best to use `py-1` rather than `p-1`. I make this mistake all the time, but technically classes don't cascade depending on the order they are listed, they cascade depending on the order they are declared, so it's best not to have two conflicting tailwind classes, since it's kind of like leaving up to chance which one of them will actually take effect.

I haven't investigated this deeply, but I sort of suspect that tailwind is actually pretty smart about this and tries to order things so that it "just works", because I haven't noticed it being a problem even when I mess it up. Still, it seems like a good thing to be on the lookout for.

I really like tailwind. And, also, I am sort of afraid for the first time I meet a front-end dev who has only ever written tailwind and doesn't understand how CSS works and I have to say "back in my day we wrote stylesheets by hand and you had to craft your own classes and selectors!" And then I'll have to cut the conversation short to go clean my dentures or something.

jcmorrow (approved):
Tiny nit pick of a comment. Also, you're back! Welcome back! I missed you! I'll be at a dentist appointment during office hours tomorrow, sadly, but I look forward to joining next week!

jcmorrow:
Also, per a suggestion from @abstractalgo we're assigning PR's to ourselves and also to the Sprint, just to make sure we're making our best effort to track everything in one place, so I'm going to do those things.

src/ui/components/Events/Event.tsx
jonbell-lot23: lol to all of that

Totally agree! Thanks for the tip, will change to py-1!

PR #5120, opened 2022-01-26
ResponsiveTabs component

abstractalgo:
A component whose tabs that overflow are moved into a dropdown. In action:

https://user-images.githubusercontent.com/1355455/152902056-10376b65-3385-451f-92b1-08106a887681.mp4

Some special behaviors:

- [x] elements that overflow will go to a dropdown
- [x] active element is always visible (i.e. never in a dropdown) and at most as the last in the list of visible elements (if not in the order that it should be relative to all the other elements)
- [x] clicking outside closes the dropdown

---

I wrote the interface to be as minimal as possible (by using regular React `children`), and this will allow us to use this component with zero changes to existing tabs and toolbars (they render and style themselves and have all the click handlers that they need without `ResponsiveTabs` having to know literally anything about it or the logic that manages them).

---

Network panel uses the same/similar tabs, but it had a slightly different behavior where it shows "X" (close panel) button as well, so I didn't make those changes. Also, the Network panel has Splitter settings set in such way that prevent tabs from ever overflowing.


jaril:
@abstractalgo how's this going??

abstractalgo:
@jaril pretty good, left is to enforce the visibility of the selected element at all times (i.e. never to be selected in the dropdown) by reordering them a bit

src/devtools/client/inspector/ResponsiveTabs.tsx
```diff
@@ -0,0 +1,175 @@
...
+      if (idx === activeIdx) {
+        continue;
+      }
+
+      runningWidth += tabsRef.current[idx].clientWidth;
+
+      if (runningWidth <= containerWidth) {
+        if (idx < activeIdx) {
+          tabIndices.pop();
+          tabIndices = [...tabIndices, idx, activeIdx];
```
hbenl, line 78: I'd prefer `tabIndices.splice(tabIndices.length - 1, 0, idx)`, so that it's more obvious that you're just inserting `idx` at a specific position.

src/devtools/client/inspector/ResponsiveTabs.tsx
```diff
@@ -0,0 +1,175 @@
...
+    };
+  }, [recalcTabsVisibilityAndOrder]);
+
+  useEffect(() => {
+    recalcTabsVisibilityAndOrder();
+  }, [activeIdx, recalcTabsVisibilityAndOrder]);
+
+  useEffect(() => {
+    recalcTabsVisibilityAndOrder();
+  }, []);
```
hbenl, line 116: Remove this effect, as the previous one already does all the work.

src/devtools/client/inspector/components/App.tsx
```diff
@@ -102,7 +103,7 @@ const InspectorApp: FC = () => {
                   <div className="devtools-sidebar-tabs">
                     <div className="tabs">
                       <nav className="tabs-navigation">
-                        <ul className="tabs-menu" role="tablist">
+                        <ResponsiveTabs activeIdx={availableTabs.indexOf(activeTab)}>
```
hbenl, line 106: Since we're not using `<ul>` for the container anymore, the items should not be `<li>`s anymore either.

hbenl (approved):
Some small nits, otherwise LGTM.

Issue #5124, opened 2022-01-26
Add share link to Replay info


jonbell-lot23:
As mentioned [in this ticket](https://github.com/RecordReplay/devtools/pull/5118), it would be nice to have the ability to share from the Replay info pane

jaril:
Closed by #5126.

PR #5126, opened 2022-01-26
Adding share link to replay info

jonbell-lot23:
Addresses #5124

Old, without a link:
image

New, with a link:
image

One minute loom explaining design tradeoffs:
https://www.loom.com/share/b6443801e1204503a2eb301cb849f168

### DIFF
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -10,8 +10,7 @@ import { isPublicDisabled } from "ui/utils/org";
 
 const WorkspacePrivacySummary = ({ workspace: { name } }: { workspace: Workspace }) => (
   <span>
-    {`Members of `}
-    <span className="underline">{name}</span>
+    <span className="font-bold">{name}</span>
     {` can view`}
   </span>
 );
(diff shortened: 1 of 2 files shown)


src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -10,8 +10,7 @@ import { isPublicDisabled } from "ui/utils/org";
 
 const WorkspacePrivacySummary = ({ workspace: { name } }: { workspace: Workspace }) => (
   <span>
-    {`Members of `}
-    <span className="underline">{name}</span>
```
jaril, line 14: I prefer having `Members of` since I think that makes it the grouping consistent (Anyone..., Only people..., Members of...). I'll leave it to you!

jaril:
The dropdown's a shared component so it affected the share modal as well:

image

jaril:
Looks pretty good! There's just the question of whether we want all of the privacy dropdowns to be that small, or if we want it to be specific to the replay info pane dropdown.

jonbell-lot23:
Good question! Yup, I made them both small on purpose to align closer to our style guide. Currently documenting when to stick to defaults and when to go bigger.

jaril:
@ceceliacreates this might need a docs related 👋

PR #5131, opened 2022-01-27
Moving our default size from “small” to “default”

jonbell-lot23:
**Background**  
We have a lot of inline tailwind styles, which has worked pretty well so far. I've been auditing our CSS to find ways to standardise our design system further, and I noticed how "text-sm" is our default size. Which isn't great naming. It should be more relative.

**Fix**\
This PR moves from text-sm to text-default to give us a better reference point. This sets up some further work.


tailwind.config.js
```diff
@@ -4,6 +4,14 @@ module.exports = {
...
   darkMode: false, // or 'media' or 'class'
   theme: {
+    fontSize: {
+      xs: ".75rem",
+      sm: ".875rem",
+      default: ".875rem",
+      lg: "1.125rem",
+      xl: "1.25rem",
+      "2xl": "1.5rem",
+    },
```
jaril, line 14: We don't need to hardcore all the other font sizes except for `default` here, do we?

src/ui/components/App.css
```diff
@@ -1,3 +1,10 @@
+.theme-tab-font-size,
+.secondary-toolbox-header,
+.source-tab,
+h2 {
+  font-size: 14px;
+}
```
jaril, line 6: What's this change for?

tailwind.config.js
jonbell-lot23: We don't strictly need to yet, but this is setting up follow-up work I'm in the process of. Would like to keep it for now, and we can always strip it out later when I'm done.

src/ui/components/App.css
jonbell-lot23: I admit this looks ugly in this format, and will be cleaning it up further. Here's what's up:

We are using tailwind's type ramp, as shown by this middle column:

image

And I noticed that our nav tabs were using 15px, which breaks out of our system. So I'm forcing them back down to 14px. In the coming PRs I'm going to clean this up to use our system more cleanly, but for now I'm just forcing to 14px.

jonbell-lot23:
> LGTM. Only other (very particular) nit I'd have is the naming.
> 
> `default` sounds ironic, since if it's default then we shouldn't ever have to set it in the first place? I think I'd prefer some other generic like `body`. I'll leave it up to you!

Totally agree! I'd love to be able to drop these entirely, but we set the whole UI to 16px which makes that impossible.

What I really like is a new developer copying a component and seeing the word "default" and knowing exactly what that word means. And then seeing "small" and knowing "I bet that's for small text." 

Thanks!

PR #5134, opened 2022-01-27
Adding robustness to React DevTools

abstractalgo:
Still a WIP trying to figure out why it might not work, but adding types and docs along the way.

- [x] improve docs
- [x] change how React DevTools get initialised by showing the panel only for points on timeline that happened after the first annotation
- [x] better UI/UX handling for failures (if bridge protocol mismatch happens - show what to try to do)
  - [x] add tracking events for failures

What happens if the React DevTools work, but the user has picked a point on timeline before it was inited.

https://user-images.githubusercontent.com/1355455/154148166-4368bca1-68d4-4204-bebb-2cf1aa9ccbcf.mp4

What happens if the React DevTools can't init at all (bridge protocol version mismatch):

image

---

### How

When (non-React) `DevTools` init, it receives annotations for the React DevTools. If no annotation - no React DevTools. If some annotations - then React DevTools generally exists; so we store the time it is first available, and if the user picks any point in time before that one, we keep the tab for React visible and show the message inside that it's not inited. If some other point in time is picked when React DevTools is available, we simply show them.

In the case where the bridge protocol check has failed, either due to timeout or an actual protocol mismatch, we still keep the React tab visible and display a message telling the user what to try.

With this change, we basically never shutdown React DevTools, we always try to "recover" as much as possible from whatever the situation is.

---

Relevant:
- https://github.com/RecordReplay/gecko-dev/issues/322
- https://github.com/RecordReplay/gecko-dev/issues/318


abstractalgo:
Off to a good start 😬 

image

abstractalgo:
Yeah, can't just upgrade `react-devtools-inline/frontend` without it being matched on the backend (at least the protocol version, not necessarily exact npm package version).

abstractalgo:
The procotol changed at `4.22.0`: https://github.com/facebook/react/blob/main/packages/react-devtools/CHANGELOG.md#4220-december-13-2021

abstractalgo:
There's an unhandled situation in `ReactDevTools:149` where the eval from backend might have failed or returned an exception.

image

To figure out:
- why it even failed/threw (especially considering that this is a super basic op that it fails on `getBridgeProtocol`)
- even in the case of "unhappy path", what do we do about it? retry mechanic (would it help even), or some user-facing alert (lowers confidence that something once captured can be inspected/replayed),...

ceceliacreates:
If it's helpful, I recorded a clone of https://github.com/ahfarmer/calculator locally and the replay had this issue. https://app.replay.io/recording/react-calculator--aaa087dc-ea2a-4ea0-9e9f-f20efd3f49d1.

abstractalgo:
Ohhh, amazing, a perfect repro! Thanks so much! 🙌  @ceceliacreates

abstractalgo:
related https://github.com/RecordReplay/devtools/issues/5243

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -75,7 +79,8 @@ class ReplayWall implements Wall {
       case "getBridgeProtocol": {
         const response = await this.sendRequest(event, payload);
         if (response === undefined) {
-          this.onShutdown();
+          trackEvent("error.reactdevtools.get_protocol_fail");
+          setLastProtocolCheckFailed();
```
jasonLaster, line 83: probably change this to `setProtocolFailed` to be consistent

src/ui/setup/dynamic/devtools.ts
```diff
@@ -169,8 +166,6 @@ export default async function DevTools(store: Store) {
   setupExceptions(store);
   setupReactDevTools(store);
 
-  const settings = await getUserSettings();
```
jasonLaster, line 172: I like this change, but lets make it its own PR and just land it.

src/ui/reducers/reactDevTools.ts
```diff
@@ -2,27 +2,54 @@ import { compareNumericStrings } from "protocol/utils";
...
+      let firstOpAnnotationPoint: ExecutionPoint | null = null;
+      {
+        for (const { message, point } of annotations) {
+          if (message.event === "operations") {
+            firstOpAnnotationPoint = point;
+            break;
+          }
+        }
+      }
+      const hasReactComponents = firstOpAnnotationPoint !== null;
```
jasonLaster, line 31: couple of stylistic points:

1. lets rename `firstOpAnnotationPoint` to `initPoint` or something more intent based and less implementation based.
2. I think `_.find` would be better for finding the first operation.

```js
const firstOperations = find(annotations, annotation => annotation.message.event == "operations")
const initPoint = firstOperations?.point
const hasReactComponents = !!initPoint
...
```

src/ui/reducers/reactDevTools.ts
```diff
@@ -2,27 +2,54 @@ import { compareNumericStrings } from "protocol/utils";
...
+        ...state,
+        currentPoint: action.currentPoint,
+        lastProtocolCheckFailed: false,
+      };
+    }
+
+    case "set_protocol_fail": {
+      return {
+        ...state,
+        lastProtocolCheckFailed: true,
```
jasonLaster, line 51: I think `last` is redundant. Perhaps `protocolCheckFailed`

jasonLaster (commented):
Left some general comments

jasonLaster (approved):

PR #5141, opened 2022-01-27
Test events

jasonLaster:
We should also send E2EFinished events on test success


jcmorrow (approved):
Nice, one `if` statement down, 5,348,594 to go 🪓

PR #5142, opened 2022-01-27
Add some loading timing data into the session.devtools_start event

jaril:


src/ui/utils/mixpanel.ts
```diff
@@ -147,6 +147,11 @@ export async function trackMixpanelEvent(...[event, properties]: [...MixpanelEve
     mixpanel.track(event, { ...properties, namespace: namespaceFromEventName(event) });
   }
 }
+export async function timeMixpanelEvent(event: string) {
```
jaril, line 150: Only the event is relevant here. @abstractalgo got some TS-fu that can get us the right type?

abstractalgo: If I understand correctly, `event: MixpanelEvent[0]`?

jaril: yep!

jaril:
```
Also, we should mark every request for that recordingId+userId combo accepted since they could have requested multiple times (which I think is fine)
```

PR #5143, opened 2022-01-27
Switch from padding to margin in Replay Info

jcmorrow:
If we use padding here there's a flexbox oddity (not sure if it's
officially a bug or not, though there are definitely some flexbox
padding bugs) where the height gets incorrectly calculated.

### Before

https://user-images.githubusercontent.com/5903784/151436054-117ba3e5-2fd5-4f7b-8224-b9462ce7d50d.mp4

### After

https://user-images.githubusercontent.com/5903784/151436075-7877b9a5-e26d-4e20-a120-77f69a0ad820.mp4

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -40,7 +40,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
 
   return (
     <div className="flex overflow-hidden flex flex-column items-center bg-white border-splitter">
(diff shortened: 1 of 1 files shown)


abstractalgo (approved):

jcmorrow:
It's actually a bit weirder than this even, see https://www.loom.com/share/12874adc64aa4f15b4eaa447da3b3f7b

But this does help in some cases

abstractalgo:
Accordion really making it easy for us, eh? 😄 
I wouldn't dare to guess, but flex usually does make for some weird situations. 😕

PR #5153, opened 2022-01-27
Add a request access flow to the unauthorized error screen

jaril:
This adds the following:
- Show a request access button when a logged-in user is viewing a replay they're not authorized to.
- Show a list of collaborator requests in the sharing modal.

Let's add some docs (@ceceliacreates) here if we need to. Definitely get some qa wolf tests.

Demo: https://share.descript.com/view/r18vVVcnm4z

https://user-images.githubusercontent.com/15959269/151447826-a2d290c5-3504-4ffa-840f-c9114e9e0691.mov

### DIFF
--- a/src/ui/components/shared/Error.tsx
+++ b/src/ui/components/shared/Error.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import useAuth0 from "ui/utils/useAuth0";
 import Modal from "ui/components/shared/NewModal";
-import { connect, ConnectedProps } from "react-redux";
+import { connect, ConnectedProps, useDispatch } from "react-redux";
 import * as selectors from "ui/reducers/app";
 import { UIState } from "ui/state";
 import classNames from "classnames";
@@ -13,6 +13,8 @@ import { PrimaryButton } from "./Button";
 import { useRouter } from "next/dist/client/router";
 import { BubbleViewportWrapper } from "./Viewport";
 import { getRecordingId } from "ui/utils/recording";
+import { setExpectedError } from "ui/actions/session";
+import { useRequestRecordingAccess } from "ui/hooks/recordings";
 
 export function PopupBlockedError() {
   return (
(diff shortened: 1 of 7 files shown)


src/ui/components/shared/Error.tsx
```diff
@@ -103,6 +130,8 @@ function ActionButton({ action }: { action: string }) {
     button = <LibraryButton />;
   } else if (action == "team-billing") {
     button = <TeamBillingButton />;
+  } else if (action == "request-access") {
```
abstractalgo, line 133: `===`

jasonLaster (approved):

jcmorrow:
This is maybe a silly request - but do you think it would be worthwhile to put min heights/widths on the dialog so that when it changes from one message to another it does not change shape slightly?

jaril:
> This is maybe a silly request - but do you think it would be worthwhile to put min heights/widths on the dialog so that when it changes from one message to another it does not change shape slightly?

I'm down. I think @jonbell-lot23 is in favor of things not shifting around too so should be good.

src/ui/hooks/recordings.ts
```diff
@@ -114,6 +114,8 @@ export async function getRecording(recordingId: RecordingId) {
     query: GET_RECORDING,
     variables: { recordingId },
   });
+
+  console.log({ data: result.data });
```
jasonLaster, line 118: 🧂

src/ui/hooks/recordings.ts
```diff
@@ -802,3 +806,30 @@ export function useRequestRecordingAccess() {
...
+
+  if (data) {
+    console.log({ data });
+  }
+  if (loading) {
+    console.log({ loading });
+  }
+  if (error) {
+    console.log({ error });
+  }
```
jasonLaster, line 832: 🧂

src/ui/components/Library/ViewerRouter.tsx
```diff
@@ -83,6 +83,8 @@ function ViewerRouter(props: ViewerRouterProps) {
   const { features, loading } = hooks.useGetUserInfo();
   const { currentWorkspaceId } = props;
 
+  console.log({ currentWorkspaceId });
```
jasonLaster, line 86: 🧂

jasonLaster (approved):
Code looks legit.

ceceliacreates:
Yes. This would work well in the Library page and also the upcoming Replay overview/What is Replay section.

PR #5154, opened 2022-01-28
Remove whatwg-url module

ryanjduffy:
## Issue

When we load a lot of sources (like 56k sources for example), a lot of time is taken building the source tree. I found that a large amount of time was taken parsing the URLs to prep the the tree.

image

## Resolution

Swapping to the native `URL` from the `whatwg-url` polyfill greatly reduces the time spent.

image


jasonLaster:
This is one of those areas where I will defer to @loganfsmyth. We use whatwg-url to be consistent with source mapped URLs. Not sure if we still need it though.

If we're talking source tree perf, i think we can do a much better job now that we get all of the sources at once.

ryanjduffy:
@jasonLaster, @loganfsmyth - Good to know! I assumed it was just an old polyfill. If it is still needed, there are other ways to ensure it isn't blocking rendering but this does seem like an easy win if we can get away it it.

jaril:
@ryanjduffy is this waiting on a review?

loganfsmyth:
The stuff to check would mostly be the behavior for paths that have custom schemes like `webpack:` and such. For example: https://github.com/mozilla/source-map/blob/7a263408de1b60e60b6220ca3e36d9b3e4aa047d/lib/url-browser.js#L10 

For devtools it mostly depends on how much we actually rely on the parser to separate out `hostname` and `pathname` and such consistently. It may also be fine to hack together a regex, hard to say.

I'd say do some spot-checking for behavior with `webpack://` original sources and if things seem decent, it seems reasonable to remove it and then address anything that does pop up later. AFAIK we don't have to do any actual manipulation of the URL objects, so that also removes a lot of complexity.

ryanjduffy:
I can see the inconsistency in the tree:

Chrome
image

Replay:
image

Safari:
image

PR #5156, opened 2022-01-28
Upload screen no longer cropping buttons

### LINKED ISSUE #4363: The upload screen is cropping off the save button
image

When the upload screen window is too small vertically, we need to use responsive design to make sure the buttons aren't cropped off. Here's the design proposal

1. When we don't have enough space vertically, drop the screenshot.
2. Use an inline scroll so if the window is too small, at least someone can scroll down to it.

jonbell-lot23:
Loom explaining the change:
https://www.loom.com/share/ac4d35e91896496ca3d97f7e8fd37da3

Fixes #4363

### DIFF
--- a/src/ui/components/UploadScreen/Privacy.tsx
+++ b/src/ui/components/UploadScreen/Privacy.tsx
@@ -46,10 +46,10 @@ function FavIcon({ url }: { url: string }) {
   return (
     <div className="relative">
       <div className="flex">
-        <MaterialIcon>public</MaterialIcon>
+        <MaterialIcon className="invisible">public</MaterialIcon>
       </div>
       <img
-        className="h-4 w-4 absolute top-0 left-0 bg-white"
+        className="h-4 w-4 absolute top-1 left-1 bg-white"
         src={`https://www.google.com/s2/favicons?domain=${url}`}
       />
     </div>
--- a/src/ui/components/UploadScreen/Privacy.tsx
+++ b/src/ui/components/UploadScreen/Privacy.tsx
@@ -46,10 +46,10 @@ function FavIcon({ url }: { url: string }) {
   return (
     <div className="relative">
       <div className="flex">
-        <MaterialIcon className="invisible">public</MaterialIcon>
+        <MaterialIcon>public</MaterialIcon>
       </div>
       <img
-        className="h-4 w-4 absolute top-1 left-1 bg-white"
+        className="h-4 w-4 absolute top-0 left-0 bg-white"
         src={`https://www.google.com/s2/favicons?domain=${url}`}
       />
     </div>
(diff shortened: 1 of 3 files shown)


src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -112,8 +112,8 @@ function ReplayScreenshot({
 }) {
   return (
     <div
-      className="relative rounded-lg px-6 pt-6 shadow-xl h-64 bg-jellyfish"
-      style={{ height: "280px" }}
+      className="relative rounded-lg px-6 pt-6 shadow-xl h-64 bg-jellyfish bg-white short:hidden"
```
jaril, line 115: this has two `bg`s

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -112,8 +112,8 @@ function ReplayScreenshot({
...
+      style={{ height: "280px;" }}
```
jaril, line 116: can drop the `;`

src/ui/components/UploadScreen/Privacy.tsx
```diff
@@ -46,10 +46,10 @@ function FavIcon({ url }: { url: string }) {
   return (
     <div className="relative">
       <div className="flex">
-        <MaterialIcon>public</MaterialIcon>
+        <MaterialIcon className="invisible">public</MaterialIcon>
```
jaril, line 49: This is here as a backup while the actual favicon loads. Was there a reason for keeping this invisible instead of just removing it altogether?

src/ui/components/UploadScreen/Privacy.tsx
```diff
@@ -46,10 +46,10 @@ function FavIcon({ url }: { url: string }) {
...
       </div>
       <img
-        className="h-4 w-4 absolute top-0 left-0 bg-white"
+        className="h-4 w-4 absolute top-1 left-1 bg-white"
```
jaril, line 52: What's the reason for adding the adding top/left values? It looks like it ends up uncentering the images in the row.

image

src/ui/components/UploadScreen/Privacy.tsx
jonbell-lot23: Yeah, I had a lot of trouble with these @jaril. When I made them larger they looked blurry. Let's pair up and figure out how to fix this!

src/ui/components/UploadScreen/Privacy.tsx
jonbell-lot23: Yeah that looks bad. I was seeing the opposite problem, maybe it was a browser difference or something? To address.

jaril:
@jonbell-lot23 reminder to pull me in later to pair and get this shipped :)

jaril (approved):

PR #5159, opened 2022-01-28
Set a min height for the error screen text

jaril:
This gives us ~40px of minimum height to keep the error screen fairly standard.


jcmorrow (approved):
🙏  THANK YOU!

PR #5167, opened 2022-01-28
Pop open the share modal when the user navigates to a recording with /share

### LINKED ISSUE #5163: Add routing so that we have a way to automatically open the sharing modal upon opening a replay
Follow up to #5153.

This allows us to send the replay owner a notification when somebody requests access to their replay with a link that opens their replay, and pops open the sharing modal immediately after.

jaril:
Fix #5163.

### DIFF
--- a/pages/recording/[id]/share.tsx
+++ b/pages/recording/[id]/share.tsx
@@ -1,0 +1,13 @@
+import { useRouter } from "next/router";
+import { useDispatch } from "react-redux";
+import { setModal } from "ui/actions/app";
+import { useGetRecordingId } from "ui/hooks/recordings";
+
+export default function Share() {
+  const recordingId = useGetRecordingId();
+  const router = useRouter();
+  const dispatch = useDispatch();
+
+  router.replace({ pathname: "/recording/[id]", query: router.query });
+  dispatch(setModal("sharing", { recordingId }));
+}


ryanjduffy (commented):
I worry about overloading the `[id]` route with all possible future subroutes. Did you try adding ` [id]/share` route? That would be easier to extend if it's possible.

pages/recording/[id]/share.tsx
```diff
@@ -0,0 +1,15 @@
+import { useRouter } from "next/router";
+import { useDispatch } from "react-redux";
+import { setModal } from "ui/actions/app";
+import { useGetRecordingId } from "ui/hooks/recordings";
+
+export default function Share() {
+  const recordingId = useGetRecordingId();
```
Comment (ryanjduffy) [inline pages/recording/[id]/share.tsx:7]: Can probably drop this since you're pulling the recording id from the query too.

ryanjduffy (approved):

ryanjduffy:
I'll add that I think using `replace()` is appropriate until we start including modals in our routes otherwise. Since clicking the Share button doesn't update the route yet, it makes sense to me to drop `/share` for this case.

pages/recording/[id]/share.tsx
Comment (jaril) [reply pages/recording/[id]/share.tsx]: left it on just in case we have to handle slugs

gideonred:
FYI, if the replay-soc2 bot is annoying here what we could do is add logic to it to not approve draft PRs and you guys just need to flag the PR as draft.

gideonred:
or, I guess I could just have it check it has already been approved.

pages/recording/[id]/share.tsx
Comment (ryanjduffy) [reply pages/recording/[id]/share.tsx:7]: gotcha :+1:

ryanjduffy (approved):

PR #5191, opened 2022-01-31
Set the my library value to null

jaril:


ryanjduffy (approved):

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -157,7 +157,7 @@ export default function UploadScreen({ recording, userSettings, onUpload }: Uplo
     e.preventDefault();
 
     setStatus("saving");
-    const workspaceId = selectedWorkspaceId == "" ? null : selectedWorkspaceId;
+    const workspaceId = selectedWorkspaceId == MY_LIBRARY ? null : selectedWorkspaceId;
```
loganfsmyth, line 160: Can `selectedWorkspaceId` itself be changed to be `null` by default? Checks like this are super easy to break.

PR #5198, opened 2022-02-01
Replay checkin -- halfway done draft

jonbell-lot23:
Hey @jaril ! Here's my half done replay check-in work.

[Here's the original ticket](https://github.com/RecordReplay/devtools/issues/5186)

And here are the things that still need doing:

* Put this in the top header of the accordion
* Swap out the colour when you select a new item
* And of course connect it to the back-end

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "prop-types": "^15.7.2",
     "properties-parser": "^0.3.1",
     "react": "^0.0.0-experimental-edfe50510-20210823",
+    "react-canvas-confetti": "^1.2.1",
     "react-circular-progressbar": "^2.0.4",
     "react-devtools-inline": "^4.17.0",
     "react-dom": "^0.0.0-experimental-edfe50510-20210823",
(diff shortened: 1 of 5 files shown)


jaril:
Do we need three options? "Couldn't resolve" doesn't seem practical. If you're in a situation where you couldn't resolve a replay, you're gonna close it and run as far away from it as possible.

jonbell-lot23:
> Do we need three options? "Couldn't resolve" doesn't seem practical. If you're in a situation where you couldn't resolve a replay, you're gonna close it and run as far away from it as possible.

Great question. I totally agree that the vast majority of people aren't going to bother changing this status just to explain that it didn't work. 

That said, I do eventually want the ability for someone to mark a Replay as bad. It gives them a way to file a bug, ask for a feature, signal to Cecelia that we need to set up a call, etc.

So I'd probably keep it as a proof of concept and something for us to discuss/iterate on. But dropping it for now isn't so bad either. We'll be iterating.

jaril:
> So I'd probably keep it as a proof of concept. But dropping it for now isn't so bad either. We'll be iterating on what we learn.

Gotcha. Unless there's any strong objections, let's drop it. "This replay is bad, somebody do something" is important enough that it should be handled in its own flow once we're ready for it.

jonbell-lot23:
Oh sorry, I think we had a miscommunication here. @jaril I'm hoping you can address each of these three items.

PR #5206, opened 2022-02-01
Disable telemetry on preview branches

ryanjduffy:
* Clean up usage of `isDeployPreview()`
* Add `isDeployPreview()` to `skipTelemetry()` so we don't send prod errors from preview branches


jaril (approved):
🚀

PR #5209, opened 2022-02-02
Simplifying icon component strategy

jonbell-lot23:
**Background**

We have a lot of icons sprinkled through our app, but a few different ways to call them. Notably, it's not easy to use hover effects on most of the older ones. So this PR standardises an approach that allows us to use hover.

**Important note**

This kind of standardisation seems like a low priority, but from a polish/design standpoint, it's a P1. Our app appearance is currently a Frankenstein monster of different line widths and icon libraries, so this is our first step towards making it more professional-looking.

**More details**  

[Here's the raw icon audit from a few weeks ago](https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/?node-id=9820%3A94699)
And [here's a new document](https://www.figma.com/file/SSr1ljyzF0lCXTL7BX1Whh/?node-id=10149%3A96683) called Icon Library where we'll track which icons are using this simplified approach


jaril:
Rapid fire comments:
- The SVGs should be placed in some central folder instead (`src/image/images` looks like a good candidate).
- I know we all agreed that directly importing the SVGs would be the more straightforward approach. After having a think about it, in the case that we had to find all of the places where an SVG was being imported like this, it would be a pain in the butt to replace. I think my preference would be to have a boring `<Icon>docs</Icon>`, that would then go and reference `docs` to find the appropriate URL `src/image/images/docs,svg`.
  - edit: `/import.*svg"/g` would get you what i was referring to here, i suppose.
- This approach hard codes the icon color into the SVG. We have enough cases where we make icons different colors that I think it's reasonable to have our standard approach make it easy to change an icon's colors. The way the current icons do this is by using the svg as a mask, then using that to mask some color (see `.img` selectors in the codebase). My preference here would be to have a component where you could do `<Icon color="blue">docs</Icon>` or <div className="bg-blue-500"><Icon>docs</Icon></div>`.

@jcmorrow thoughts?

jonbell-lot23:
1. Colors: 100%. Not being able to set color or use hover effects is something we should address.
2. I kept the files in the component folder because philosophically I think that's a more solid approach. (Just like CSS files) But I forgot about the use-case of using the same icon in multiple places, which is clearly better to centralise.

Will move the files into a single place now.

jonbell-lot23:
Ok @jaril @jcmorrow and @ryanjduffy , check out this latest commit.

1. I'm centralising everything in /public/
2. I'm using a component as per Jaril's recommendations

But I'm still not clear how to get colors lighting up on hover. Any thoughts how I can do this?

src/ui/components/shared/Icon.tsx
```diff
@@ -1,26 +1,13 @@
...
-      style={{ display: "inline-block", verticalAlign: "middle" }}
-      viewBox={"0 0 20 20"}
-      width={`20px`}
-      height={`20px`}
-      xmlns="http://www.w3.org/2000/svg"
-      xmlnsXlink="http://www.w3.org/1999/xlink"
-    >
-      <path fill={"currentColor"} d={icons[iconName]} />
-    </svg>
+    <img
```
jaril, line 5: Not quite:
1) We already implement something similar to what we're trying to get here so you can copy a good chunk of code (pasting screenshot below for the command palette icon in the editor, but you can inspect that too).
2) In case it's cleaner to just use a proper CSS file here, we should. The way we import is a little bit roundabout. After creating `Icon.css`, make sure you import it in `pages/_app.tsx` along with the other stylesheets.

image

jonbell-lot23:
**The old way**

Import
```
import CommandPaletteButton from "./CommandPaletteButton";
```

Call
```
<CommandPaletteButton />
```

Update image.css
```
.img.palette {
  -webkit-mask-image: url(/images/palette.svg);
}
```

**My proposal**
I'm using the same three steps, but generalising the component to work with any image. Here's an example:
```
<Icon filename="help" width="20px" height="20px" />
```

We still need to update the CSS file with the image, but I like how this is more clear about is going to be placed, and what the filename is.






<Icon filename="help" width="20px" height="20px" />

src/ui/components/Header/UserOptions.tsx
```diff
@@ -74,21 +74,21 @@ function UserOptions({ setModal, noBrowserItem }: UserOptionsProps) {
...
         <button className="row" onClick={onSettingsClick}>
-          <Icon iconName="settings" />
+          <Icon filename="settings" width="20px" height="20px" />
           <span>Settings</span>
         </button>
         {features.launchBrowser ? (
           window.__IS_RECORD_REPLAY_RUNTIME__ || noBrowserItem ? null : (
             <button className="row" onClick={onLaunchClick}>
-              <Icon iconName="replay" />
+              <Icon filename="replay-logo" width="20px" height="20px" />
```
jaril, line 91: If all `<Icon />` are all supposed to be 20x20 for now, let's just hard code it into the component. 

Once we want to extend it to different sizes, it'd be nicer to just be able to pass some well-typed size (e.g. <Icon size="some-size" />, where it's `sm` | `md` | `lg` or whichever scale you prefer).

src/ui/components/shared/Icon.tsx
```diff
@@ -1,26 +1,10 @@
...
-      viewBox={"0 0 20 20"}
-      width={`20px`}
-      height={`20px`}
-      xmlns="http://www.w3.org/2000/svg"
-      xmlnsXlink="http://www.w3.org/1999/xlink"
-    >
-      <path fill={"currentColor"} d={icons[iconName]} />
-    </svg>
-  );
+  return <div className={`img ${props.filename}`} style={iconStyle} />;
```
jaril, line 9: Ha big ups on just re-using `.img` 👍

jaril: But yeah just setting style to `{height: "20px", width: "20px"}` should keep this nice and simple (for now).

jcmorrow: yeah agreed, I would not use the local variable here

jcmorrow:
I would normally expect colors to be done with tailwind, just like we do text colors, background colors, etc.

jaril:
> * I'm still not clear how to do hover states and colors.
- This one's simple because we're using the svg as a mask image. You essentially just need to apply a background color to the returned div in <Icon /> and that's what the color of the icon ends up being. Lazy way is to just add a prop `classNames`, so you can pass in tailwind colors, e.g. `<Icon className="bg-red-500 hover:bg-red-700" />` then pipe it into the div.

> * The Launch Replay icon isn't showing up right, and I'm not sure why
The Launch Replay icon isn't showing up because the alpha channel's a bit wonky. For an SVG to work well as a mask here, you want an image where the dark parts as 100 opacity, then everything else being 0 opacity. Right now, the logo has the triangles in the middle but I'm pretty sure they're a layer of white triangles on top of the rounded square. Instead, you want to subtract those triangles from the rounded square. That's in Figma.

jcmorrow:
For an example of how to style SVG's with tailwind, check the docs under `SVG`: https://tailwindcss.com/docs/fill

src/ui/components/shared/Icon.tsx
jonbell-lot23: Removed. But it'll probably come roaring back soon when I use this component elsewhere in the system.

jonbell-lot23:
Ok, ready to go! @jaril what do you think?

jaril (approved):

PR #5215, opened 2022-02-02
Refactor analysisPoints -> analysisPoint

jaril:
Fix https://github.com/RecordReplay/customer-support/issues/18

We used to have `analysisPoints` in Redux correspond to a collection that had the location as they key and an array of its "analysis points" as its value. That is, until we hit an error while doing an analysis. The two distinct cases are:
1) Analysis to get the points, e.g. we're trying to find the number of hits to show in the tooltip
2) Analysis to get the results, e.g. we're evaluating an expression to show its results in the console.

Right now, the code is written such that an error for either replaces the array (whether existing or not) value in `analysisPoints` with the string `"error"`. Kinda janky but it works. Well, actually, not so much since we don't make a distinction between a points error, and a results error. That's bad since ideally we'd like to treat those things separately in the UI.

Another bad thing about it is that it an error when getting the results can completely torpedo the points count that's already saved in Redux, which is what's happening in [this issue](https://github.com/RecordReplay/customer-support/issues/18).

This tries to introduce a wrapper object around the array of `analysisPoints`, and an `errors` array where we can store the error that occured (1, or 2, or 1 and 2).

Note: the interface is super awkward so I'm open to naming suggestions :)

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/modify.js
+++ b/src/devtools/client/debugger/src/actions/breakpoints/modify.js
@@ -152,14 +152,14 @@ export function runAnalysis(cx, initialLocation, options) {
       return;
     }
 
-    // Don't run the analysis if we already have the analysis points for that
+    // Don't run the analysis if we already have the analysis point for that
     // location.
-    const analysisPoints = selectors.getAnalysisPointsForLocation(
+    const analysisPoint = selectors.getAnalysisPointForLocation(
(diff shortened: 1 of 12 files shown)


jaril:
Wrote myself into a corner here. Got a better idea hold on to your butts...

jaril:
Continued in #5223.

Issue #5137, opened 2022-01-27
Improve e2e and mock tests


hbenl:
- fix the tests that are currently skipped
- replace our homegrown test runner with jest
- remove the distinction between e2e and mock tests (all tests should use the GraphQL mocks, the protocol mocks are optional)
- rewrite the tests to not run in the browser anymore

bhackett1024:
>     * rewrite the tests to not run in the browser anymore

Can you elaborate on this point?

hbenl:
@bhackett1024 Currently our test scripts run entirely in the browser and write their results to the browser console, which our test runner script then extracts using playwright. I would like to use playwright the way it's intended to be used: run the test script outside of the browser and control the browser using playwright's API. This is certainly the most controversial part of this issue since the rewrite would require some work without providing an immediate benefit, so it's just a proposal for now.

bhackett1024:
> @bhackett1024 Currently our test scripts run entirely in the browser and write their results to the browser console, which our test runner script then extracts using playwright. I would like to use playwright the way it's intended to be used: run the test script outside of the browser and control the browser using playwright's API. This is certainly the most controversial part of this issue since the rewrite would require some work without providing an immediate benefit, so it's just a proposal for now.

Cool, thanks for explaining!  I agree this would be a really nice architectural change, and I think it's worth doing.  I feel it *does* provide an immediate benefit, which is that it gives us a chance to more effectively dogfood the use of replay+playwright for building a test suite that's easy to debug.  We should make it work in as close a way as we can to what users will want to use playwright for, so that we can make sure this is a good experience for ourselves and so that we can show it to users as an example.

abstractalgo:
@hbenl with the PR merged, do we close this ticket or maybe leave it for some more work?

jasonLaster:
feel free to re-open or open a new issue

PR #5217, opened 2022-02-03
Fix the bundles used in e2e tests

hbenl:
- added original sources and webpack configs for our test bundles
- updated the bundles
- the test runner now uploads the example recordings directly when using firefox (instead of recording to a file and uploading that afterwards) so that they can contain sourcemaps (see RecordReplay/gecko-dev#435)
- chromium tests are currently broken, see RecordReplay/chromium#25
- the example recordings are now added correctly to `test/example-recordings.json`
- reenabled 9 tests
- the breakpoints-07 test contains a navigation, after that navigation playwright seems semi-broken and the test doesn't continue - perhaps this could be fixed by rewriting the tests to not run in the browser anymore (see #5137) so that the navigation is triggered using the playwright API
- the object_preview-05 test fails due to a scope mapping issue in the backend (RecordReplay/backend#4483)
- the react_devtools test doesn't work with an example created with playwright because that browser doesn't support react devtools yet - you can run that test by adding `"cra/dist/index.html": "acbf7e7e-9ad7-44f1-a5bd-70aa8b6c0fea"` to `test/example-recordings.json` (that recording was made manually with our standard gecko browser)


src/test/harness.js
```diff
@@ -138,14 +138,14 @@ async function selectSource(url) {
 }
 
 async function addLogpoint(url, line) {
-  const bpCount = dbgSelectors.getBreakpointCount();
+  const bpCount = dbgSelectors.getLogpointCount();
```
abstractalgo, line 141: `lpCount` then, not `bpCount` 😄

test/src/getExample.js
```diff
@@ -51,14 +87,25 @@ async function upload(state, example) {
...
+  state.exampleRecordings = { ...state.exampleRecordings, [example]: recordingId };
+  fs.writeFileSync(
+    "./test/example-recordings.json",
+    JSON.stringify(state.exampleRecordings, null, 2)
+  );
 }
 
 async function recordExample(state, example, target) {
+  const exampleUrl = `${state.testingServer}/test/examples/${example}`;
   const browser = target == "gecko" ? "firefox" : "chromium";
```
abstractalgo, line 99: `===`

package.json
```diff
@@ -149,7 +154,9 @@
     "style-loader": "^3.2.1",
     "ts-node": "^10.0.0",
     "tsconfig-paths": "^3.9.0",
-    "typescript": "^4.5.4"
+    "typescript": "^4.5.4",
+    "webpack": "^5.68.0",
+    "webpack-cli": "^4.9.2"
```
abstractalgo, line 159: We could create a `package.json` specifically for tests, instead of pollute a global one.

hbenl:
Superseded by #5275

Issue #5222, opened 2022-02-03
Enforce the backend's 200 hit limit on analysis results


jaril:
The backend limits full analysis results to 200, but we still allow users to perform actions that lead straight to that error.

This limit only applies to more complex expressions (e.g. `console.log(">>", window)`). This doesn't apply to an expression that only has primitives, like our default expression (e.g. `console.log(123, "runScript")`).

We should disable editing the default expression if there's more than 200 hits for that location. We had this check in place before, but it seems to have regressed.

AND write a QA wolf test for it :).

jaril:
@jonbell-lot23 sticking your head on this for now until we dig up the designs from figma!

jasonLaster:
We should also limit console events that are 200+

We could also implement paging to get the first 200

jaril:
events: 
- cap the hit count to `200+` instead of 201 or 1k
- clicking on `200+`, disable checkbox, then tooltip

PR #5223, opened 2022-02-03
Keep the hits count from disappearing if the analysis result throws

jaril:
Fixes https://github.com/RecordReplay/customer-support/issues/18.

We have a [hard-coded limit for a full analysis on the backend](https://github.com/RecordReplay/backend/blob/cbb48ebfd2de92d4463fcfb5a9691ccfd40da028/src/control/handlers/analysis.ts#L313). When that throws, it ends up rewriting the existing data about the corresponding location, which is why we have that weird "10k+ hits" behavior.

We could handle that specific full-analysis-for-200+  error separately, or even better, just not be in a place where that's the case in the first place (see follow-up item #5222).

### DIFF
--- a/src/protocol/logpoint.ts
+++ b/src/protocol/logpoint.ts
@@ -274,7 +274,8 @@ async function setMultiSourceLogpoint(
     }
   };
 
-  if (condition || (showInConsole && !primitives)) {
+  const shouldGetResults = condition || (showInConsole && !primitives);
+  if (shouldGetResults) {
     handler.onAnalysisResult = result => {
       results.push(...result);
       if (showInConsole && (condition || !primitives)) {
@@ -286,7 +287,12 @@ async function setMultiSourceLogpoint(
   try {
     await analysisManager.runAnalysis(params, handler);
   } catch {
-    saveAnalysisError(locations, condition);
+    // Only save the error if we're only grabbing the points for a location.
+    // This means that we're not handling cases where the full analysis
+    // throws. We should add that as a follow-up.
+    if (!shouldGetResults) {
+      saveAnalysisError(locations, condition);
+    }
     return;
   }
 


src/protocol/logpoint.ts
```diff
@@ -274,7 +274,8 @@ async function setMultiSourceLogpoint(
     }
   };
 
-  if (condition || (showInConsole && !primitives)) {
+  const shouldGetResults = condition || (showInConsole && !primitives);
```
jcmorrow, line 277: This is not your fault, but I'm struggling with these names. I assume that `condition` means the user-specified `if` statement, `showInConsole` means that we're expected to show the results of this in the console (when is this not true?) and I'm not sure what primitives is? And `shouldGetResults` is kind of confusing because I would expect that if `shouldGetResults` is `false` we wouldn't call `runAnalysis`. So I guess I'm not sure what getting results means here!  Anyways, the change looks fine I think, or at least, I believe you that it does what it says on the tin, but it might be a nice time to revisit some of this naming if you have all of the context in your head!

jcmorrow (approved):
I am confused by names, but that doesn't have to block this, I'm confused all the time anyways.

src/protocol/logpoint.ts
jaril: Yeah it takes a second:
- `showInConsole`: logpoints (or print statements) show up in the console. But we also happen to use it as a roundabout way to get the number of hits for a line. It's still an analysis, but we just care about the hits count. And the way we indicate that we're only trying to get the hits is with `showInConsole` set to `false`.
- `shouldGetResults`: Similar-ish to above. If we just need the hits, we don't need the backend to actually run the full analysis.
- `primitives` is a flag for checking if the entire expression is all primitive (`"1", 2, false`), in which case we don't need to run the full analysis since there's nothing specific to each pause point to evaluate.

jaril: Gonna merge it but I totally understand how this is all sorts of weird as it's currently written. I'm adding it to my try-cleaning-it-up-on-cleanup-friday-afternoon list!

Issue #5218, opened 2022-02-03
Mock tests do not (really) run in CI


hbenl:
Currently at least one of our mock tests is broken (invalid-recording-01 waits for the wrong error message) but CI always reports success...

PR #5229, opened 2022-02-04
Fix mock tests (#5218)

hbenl:
- CI: runs `npm install` in the root folder instead of the `test` subfolder (this was the main issue preventing the mock tests from running in CI)
- fixes a bug where trying to open a recording with a recordingId that is not a UUID (plus slug) didn't show an error message
- fixes a bug if the `MockedProvider` passes a nullish `mockRef` to its callback
- fixes uploading of recordings for failed tests by adding our test API key
- fixes one of the tests that looked for an outdated error message

### DIFF
--- a/pages/recording/[id].tsx
+++ b/pages/recording/[id].tsx
@@ -3,9 +3,14 @@ import { useRouter } from "next/router";
 import React, { useEffect, useState } from "react";
 import { connect, ConnectedProps, useStore } from "react-redux";
 import { isTest } from "ui/utils/environment";
-import { getAccessibleRecording } from "ui/actions/session";
+import { getAccessibleRecording, setExpectedError } from "ui/actions/session";
 import { Recording as RecordingInfo } from "ui/types";
-import { getRecordingMetadata, useGetRecording, useGetRecordingId } from "ui/hooks/recordings";
+import {
+  getRecordingMetadata,
+  useGetRecording,
+  useGetRecordingId,
+  useGetRawRecordingIdWithSlug,
+} from "ui/hooks/recordings";
 import { getRecordingURL } from "ui/utils/recording";
 import LoadingScreen from "ui/components/shared/LoadingScreen";
 import Upload from "./upload";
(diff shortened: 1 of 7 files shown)


pages/recording/[id].tsx
```diff
@@ -103,22 +108,31 @@ function useRecordingSlug(recordingId: string) {
 
 function RecordingPage({
   getAccessibleRecording,
+  setExpectedError,
   head,
 }: PropsFromRedux & { head?: React.ReactNode }) {
   const store = useStore();
   const recordingId = useGetRecordingId();
+  const rawRecordingId = useGetRawRecordingIdWithSlug();
```
Comment (abstractalgo) [inline pages/recording/[id].tsx:116]: This two are quite similar and confusing when used at the same time; we could probably unify them into something that returns a single object with e.g. `{ id: string, idAndSlug: string, slug: string }` etc.

Also, considering that a bunch of functions that live inside `hooks/recordings.ts` have `get` in their name and actually do a remote fetch, the naming of these functions is then more wrong, so I'd suggest renaming to something along the lines get `useQueryRecordingMeta` or `useRouterRecordingMeta` or similar, to better express the type of operation (sync or async) and where the data is coming from (router vs store vs remote gql).

test/mock/run.js
```diff
@@ -43,11 +43,14 @@ async function main() {
...
     );
-    if (rv.status) {
+    if (rv.status || rv.signal || rv.error) {
+      console.log(`Failed with status ${rv.status}, signal ${rv.signal} and error ${rv.error}`);
       const recordings = listAllRecordings();
       let url = "<recording unavailable";
       if (recordings.length) {
-        const recordingId = await uploadRecording(recordings[recordings.length - 1].id);
+        const recordingId = await uploadRecording(recordings[recordings.length - 1].id, {
+          apiKey: "rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ",
```
abstractalgo, line 52: Magic const.

pages/recording/[id].tsx
Comment (abstractalgo) [reply pages/recording/[id].tsx:116]: In general, we could probably do something more smart about organizing recording information and hooks that fetch required data.

We have a single recording, and parts of its data are either there, or require fetching. Wrapping everything inside a single hook along the lines of `useCurrentRecording` would handle a bunch of responsibility of calling the right functions, eliminate a ton of smaller unnecessary hooks and simplify the interface, centralising it in one place. But indeed out of scope for this PR particularly.

abstractalgo (approved):
In general, this looks fine to me 👍

PR #5238, opened 2022-02-07
Ensure that frames are only added if the pause hasn't changed (#4688, #5233)

hbenl:


abstractalgo (approved):
lgtm

Would it be too big of a lift to convert files to Typescript?

jasonLaster:
You might be able to do the same thing by reverting this so that we go back to throwing
https://github.com/RecordReplay/devtools/commit/b94093a767e463302282b73fe635bebead26b6ab

Another benefit is that you could also do something similar here and other places where we care about the pause context

```diff
diff --git a/src/devtools/client/inspector/markup/actions/markup.ts b/src/devtools/client/inspector/markup/actions/markup.ts
index 88f328b8..c5a90d20 100644
--- a/src/devtools/client/inspector/markup/actions/markup.ts
+++ b/src/devtools/client/inspector/markup/actions/markup.ts
@@ -13,6 +13,7 @@ import {
 } from "../selectors/markup";
 import { UIState } from "ui/state";
 import Highlighter from "highlighter/highlighter";
+import { getThreadContext } from "devtools/client/debugger/src/selectors";
 const { DOCUMENT_TYPE_NODE, TEXT_NODE } = require("devtools/shared/dom-node-constants");
 const { features } = require("devtools/client/inspector/prefs");

@@ -90,7 +91,8 @@ export function newRoot(): UIThunkAction {
  * Adds the children of a node to the tree and updates the parent's `children` property.
  */
 export function addChildren(parentFront: NodeFront, childFronts: NodeFront[]): UIThunkAction {
-  return async ({ dispatch }) => {
+  return async ({ dispatch, getState }) => {
+    const cx = getThreadContext(getState());
     if (!features.showWhitespaceNodes) {
       childFronts = childFronts.filter(
         node => node.nodeType !== TEXT_NODE || /[^\s]/.exec(node.getNodeValue()!)
@@ -106,6 +108,7 @@ export function addChildren(parentFront: NodeFront, childFronts: NodeFront[]): U
       type: "ADD_CHILDREN",
       parentNodeId: parentFront.objectId(),
       children,
+      cx,
     });
   };
 }
```

Back in November, i thought we didn't need it, but we probably do

jasonLaster:
another perhaps quieter way to handle it would be on the reducer side

```diff
diff --git a/src/devtools/client/debugger/src/reducers/pause.js b/src/devtools/client/debugger/src/reducers/pause.js
index 7b36ca96..7bfbc685 100644
--- a/src/devtools/client/debugger/src/reducers/pause.js
+++ b/src/devtools/client/debugger/src/reducers/pause.js
@@ -50,6 +50,10 @@ const resumedPauseState = {
 };

 function update(state = createPauseState(), action) {
+  if (action.cx?.pauseCounter != state.threadcx.pauseCounter) {
+    return;
+  }
+
```

the downside is that other reducers like the markup reducer would need to do something similar

PR #5248, opened 2022-02-07
Adding missing cursors

jonbell-lot23:
Super simple changes: we're missing cursors on some actionable icons, so this PR adds them.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Tabs.css
+++ b/src/devtools/client/debugger/src/components/Editor/Tabs.css
@@ -47,6 +47,7 @@
   font-size: var(--theme-tab-font-size);
   background-color: transparent;
   vertical-align: bottom;
+  cursor: pointer;
 }
 
 .source-header button.command-palette {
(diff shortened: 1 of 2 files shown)


abstractalgo:
I'd recommend we try stay more uniform with these sort of changes - if we're using Tailwind, let's use it.

Even better than plain editing of CSS properties inside class is also using `@apply` from Tailwind.

Or explicitly adding class names (e.g. `cursor-pointer` on the JSX object itself, like it was done for some objects in this same PR).

jonbell-lot23:
I hear you @abstractalgo, here are some details:

1. We have a Tailwind/CSS philosophy written here in our [Replay Interface Guidelines](https://www.notion.so/replayio/Replay-Interface-Guidelines-RIG-bf43ebd84369450ba8c868fb022d5e67#b545ba7bf56e477fa6d6a02d9d8bade0). The summary is that we love Tailwind, but it will always co-exist with CSS, and where possible we'd like to standardise components that don't actually need tailwind anymore. (Versus standardising on never needing custom CSS, which isn't possible)

2. If we know we'll never remove Tailwind or custom CSS, it just comes down to picking the right tool for the job. In this case, I used Tailwind inline where it was obvious, but used CSS where it's abstracted to a higher level. (For example, we might use these tabs in multiple places, so that belongs in a global CSS rather than a per-component inline approach)

PR #5251, opened 2022-02-07
Polishing timeline hit targets (Fitts’ Law of UX)

jonbell-lot23:
**Background:**
* Our loading screen has a thicker loader bar and this change aligns our timeline with it.
* It's important that these targets be larger for easier grabbing
* It's also good to change the cursor into a gripper on something grippable

Old:
image

New:
image


jasonLaster (approved):

jaril:
@ceceliacreates maybe docs to update screenshots? But I don't think that's really necessary so just a heads up!

PR #5256, opened 2022-02-08
Hide React DevTools behind a flag

### LINKED ISSUE #5241: Hide React DevTools behind a flag
It's not fully baked yet so let's tuck it behind a user-enable-able flag until we're more confident in it.

jaril:
Fix #5241. 

React DevTools isn't quite stable yet. This hides it back behind a flag so we're setting expectations correctly.


jaril:
@ceceliacreates This un-publics React DevTools so heads up!

PR #5259, opened 2022-02-08
Eager eval for autocomplete

jaril:
This uses autocomplete to come up with autocomplete properties for objects as the user is typing them.

Previously we were limited to the immediate children of objects on the window. This allows us to do a bit more nesting (e.g. `foo.bar.b` => `foo.bar.baz`).

### DIFF
--- a/src/devtools/client/debugger/panel.d.ts
+++ b/src/devtools/client/debugger/panel.d.ts
@@ -12,4 +12,5 @@ export class DebuggerPanel {
   getSourceByActorId(id?: string): any;
   getSourceByURL(url?: string): any;
   selectSource(sourceId: string, line?: number, column?: number): any;
+  getFrameId(): { asyncIndex: number; frameId?: string };
 }
--- a/src/devtools/client/webconsole/actions/input.js
+++ b/src/devtools/client/webconsole/actions/input.js
@@ -113,6 +113,31 @@ function evaluateExpression(expression) {
   };
 }
 
+function eagerEvalExpression(expression) {
+  return async ({ _, toolbox }) => {
+    if (!expression) {
+      return null;
+    }
+
+    const { asyncIndex, frameId } = toolbox.getPanel("debugger").getFrameId();
+
+    try {
+      const response = await evaluateJSAsync(expression, {
+        asyncIndex,
+        frameId,
+        pure: true,
+      });
+      console.log(">>", { response });
+    } catch (err) {
+      let msg = "Error: Eager Evaluation failed";
+      if (err.message) {
+        msg += ` - ${err.message}`;
+      }
+      console.error(msg);
+    }
+  };
+}
+
 /**
  * Evaluate a JavaScript expression asynchronously.
  *
@@ -121,8 +146,9 @@ function evaluateExpression(expression) {
  *                          devtools/shared/fronts/webconsole.js
  */
 async function evaluateJSAsync(expression, options = {}) {
-  const { asyncIndex, frameId } = options;
-  const rv = await ThreadFront.evaluate(asyncIndex, frameId, expression);
+  const { asyncIndex, frameId, pure } = options;
+  //reminder that there would be no results if the function were impure -logan
+  const rv = await ThreadFront.evaluate(asyncIndex, frameId, expression, pure);
   const { returned, exception, failed } = rv;
 
   let v;
@@ -167,4 +193,5 @@ function onExpressionEvaluated(response) {
 module.exports = {
   evaluateExpression,
   paywallExpression,
+  eagerEvalExpression,
 };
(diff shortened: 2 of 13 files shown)


src/devtools/client/webconsole/actions/input.js
```diff
@@ -113,6 +113,30 @@ function evaluateExpression(expression) {
   };
 }
 
+function eagerEvalExpression(expression) {
```
abstractalgo, line 116: Let's migrate this to .ts file, it won't take too long since it's only a handful of actions.

src/devtools/client/debugger/panel.d.ts
```diff
@@ -12,4 +12,5 @@ export class DebuggerPanel {
   getSourceByActorId(id?: string): any;
   getSourceByURL(url?: string): any;
   selectSource(sourceId: string, line?: number, column?: number): any;
+  getFrameId(): { asyncIndex: number; frameId?: string };
```
abstractalgo, line 15: We can migrate `src/devtools/client/debugger/panel.js` to .ts as well, we have large chunk of types already.

src/devtools/client/webconsole/actions/input.js
```diff
@@ -121,8 +145,9 @@ function evaluateExpression(expression) {
...
  */
 async function evaluateJSAsync(expression, options = {}) {
-  const { asyncIndex, frameId } = options;
-  const rv = await ThreadFront.evaluate(asyncIndex, frameId, expression);
+  const { asyncIndex, frameId, pure } = options;
+  //reminder that there would be no results if the function were impure -logan
+  const rv = await ThreadFront.evaluate(asyncIndex, frameId, expression, pure);
   const { returned, exception, failed } = rv;
 
   let v;
```
abstractalgo, line 153: Can immediately destructure.

Either way, let's please not leave one-letter var names in the code. 😄

src/devtools/client/webconsole/components/Input/JSTerm.tsx
```diff
@@ -132,7 +159,14 @@ export default function JSTerm() {
...
     if (
-      [Keys.ENTER, Keys.TAB, Keys.ESCAPE, Keys.ARROW_RIGHT, Keys.ARROW_LEFT].includes(e.key as Keys)
+      [
+        Keys.BACKSPACE,
+        Keys.ENTER,
+        Keys.TAB,
+        Keys.ESCAPE,
+        Keys.ARROW_RIGHT,
+        Keys.ARROW_LEFT,
+      ].includes(e.key as Keys)
```
abstractalgo, line 169: This is irrelevant functionally, but it's more correct to cast array to `string[]` (from narrower to wider type) than it is to cast `e.key` to `Keys` (wider to narrower type).

src/protocol/thread/thread.ts
```diff
@@ -620,11 +620,16 @@ class _ThreadFront {
     return await pause.getScopes(frameId);
   }
 
-  async evaluate(asyncIndex: number, frameId: FrameId | undefined, text: string) {
+  async evaluate(
+    asyncIndex: number,
+    frameId: FrameId | undefined,
+    text: string,
+    pure: boolean = false
+  ) {
```
abstractalgo, line 628: Better to have an object as a single param that may have optional fields than doing a weird mix of optional params as `...arguments`.

src/devtools/client/webconsole/utils/autocomplete-eager.ts
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    const { returned, exception, failed } = await ThreadFront.evaluate(
+      asyncIndex,
+      frameId,
+      expression,
+      true
+    );
+    if (returned && !(failed || exception)) {
+      const evaluatedProperties = getPropertiesForObject(returned.getObject());
+      properties.push(...evaluatedProperties);
```
abstractalgo, line 18: We can do `return [...evaluatedProperties]` and also lose the var declaration at the top.

src/devtools/client/webconsole/utils/autocomplete-eager.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+  } catch (err: any) {
+    let msg = "Error: Eager Evaluation failed";
+    if (err.message) {
+      msg += ` - ${err.message}`;
+    }
+    console.error(msg);
+  }
+
+  return properties;
```
abstractalgo, line 28: We can just `return []`.

abstractalgo (changes requested):

src/devtools/client/debugger/panel.d.ts
jaril: I'll skip this one. We should be moving away from the debugger panel object anyway.

abstractalgo (approved):
- use enums for all keyboard handlers
- tsx to ts file
- (optional in this PR: moving paywalled actions from JSTerm itself)

PR #5260, opened 2022-02-08
Redirect users to the new login screen from the error screen

### LINKED ISSUE #5245: When I try to sign in from a private replay, I should be directed to the new login screen
https://www.loom.com/share/4ba8e2930de24959807cc3292ce8a027

jaril:
Fix #5245.


ryanjduffy (approved):
Seems like we should lift up the "is auth" checks eventually but this lgtm!

PR #5264, opened 2022-02-09
Eager eval the current expression in JSTerm

jaril:


src/devtools/client/webconsole/components/Input/useHistory.tsx
```diff
@@ -0,0 +1,20 @@
+import { useSelector } from "react-redux";
+import { getCommandHistory } from "../../selectors/messages";
+import clamp from "lodash/clamp";
+import { useState } from "react";
+
+export default function useHistory(setValue: (newValue: string) => void) {
```
abstractalgo, line 6: Let's at least name to `useCommandHistory` or `useEvalCommandHistory`, again, simply to not confuse it with navigation's/router's history.

src/devtools/client/webconsole/components/Input/useAutocomplete.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return matches;
+}
+function useGetMatches(expression: string) {
+  const scopeMatches = useGetScopeMatches(expression);
+  const evalMatches = useGetEvalMatches(expression);
+
+  return uniq([...scopeMatches, ...evalMatches]);
+}
+function useShowAutocomplete(expression: string, hideAutocomplete: boolean, matches: string[]) {
```
abstractalgo, line 59: This can lose the `use` part, as it's not a hook but an ordinary function.

abstractalgo (approved):
Much cleaner imo, gj :)

One thing that I noticed that really hinders readability is lack of return types - one often then needs to parse entirety of function to look for return statement and see if types diverge sometimes or it's always the same one. Often this gets enforced with an eslint rule - to always specify a return type, and while that can sometime take a toll on the coding process because you may need to update types as you rewrite parts of the code, it improves the readability and stops from some silly mistakes. So until we get that in, let's also try as much as possible to specify return types, especially if it's some easy ones.

But yeah, overall, in a much better state, I think.

jaril:
@ceceliacreates this changes the console terminal a teensy bit. Might need to update console docs!

PR #5266, opened 2022-02-09
Animation polish

jonbell-lot23:
This issue is half done and I'm struggling with the next bits. Here's a Loom showing what's up:
https://www.loom.com/share/79153ce938664a13ac5bb1cef97c9a4e

The side panel is animating, but the other part of the console is snapping in and out, which looks odd.


jonbell-lot23:
@jaril I spent some time trying to get this working. Might be a good thing for us to discuss on a call because I can't figure it out myself.

PR #5275, opened 2022-02-10
Fix the bundles used in e2e tests

hbenl:
- added original sources and webpack configs for our test bundles
- updated the bundles
- the example recordings are now added correctly to `test/example-recordings.json`
- reenabled 8 tests
- the breakpoints-07 test contains a navigation, after that navigation playwright seems semi-broken and the test doesn't continue - perhaps this could be fixed by rewriting the tests to not run in the browser anymore (see #5137) so that the navigation is triggered using the playwright API
- the sourcemap_stacktrace test fails because we're unable to create the example recording in CI (RecordReplay/gecko-dev#726)
- the object_preview-05 test fails due to a scope mapping issue in the backend (RecordReplay/backend#4483)
- the react_devtools test doesn't work with an example created with playwright because that browser doesn't support react devtools yet - you can run that test by adding `"cra/dist/index.html": "acbf7e7e-9ad7-44f1-a5bd-70aa8b6c0fea"` to `test/example-recordings.json` (that recording was made manually with our standard gecko browser)


public/test/examples/bundles/README.md
```diff
@@ -0,0 +1,11 @@
+This folder contains the sources for the bundles used in E2E tests.
+To rebuild the bundles run
+* `npm run build-e2e-prod-bundle`
+* `npm run build-e2e-exceptions-bundle`
+* `npm run build-e2e-mapped-styles-bundle`
+* `npm run build-e2e-mapped-minified-bundle`
+After rebuilding you must upload their sourcemaps by running
+* `npm run upload-e2e-prod-bundle-sourcemap`
```
abstractalgo, line 8: Needs line breaks, otherwise Markdown gets formatted wrongly.

test/manifest.js
```diff
@@ -44,15 +44,13 @@ module.exports = [
...
   },
   {
-    // This test is pretty darn broken. The sources are different when recording
-    // the example in CI vs. locally
     example: "doc_navigate.html",
     script: "breakpoints-07.js",
-    pending: true,
     targets: ["gecko", "chromium"],
+    // Disabled because the test contains a navigation that playwright doesn't handle well
+    pending: true,
```
abstractalgo, line 53: I saw this being named `disabled` or `skip` in testing frameworks. Had to look it up what it was `pending` for. 🙃  (just a suggestion)

test/manifest.js
```diff
@@ -149,6 +145,8 @@ module.exports = [
     example: "cra/dist/index.html",
     script: "sourcemap_stacktrace.js",
     targets: ["gecko", "chromium"],
+    // Disabled because we can't record the example in CI
+    // https://github.com/RecordReplay/gecko-dev/issues/726
```
abstractalgo, line 149: Love the reference to the issue 😍

test/src/state.js
```diff
@@ -16,8 +18,9 @@ const defaultState = {
   exampleRecordings: fs.existsSync("./test/example-recordings.json")
     ? JSON.parse(fs.readFileSync("./test/example-recordings.json"))
     : {},
-  headless: false,
+  headless: !!process.env.RECORD_REPLAY_PLAYWRIGHT_HEADLESS,
```
abstractalgo, line 21: Do we add this to the `.env` file, for better visibility that this option even exists, since we're already exposing it so it can be parametrised?

test/src/getExample.js
```diff
@@ -51,14 +87,22 @@ async function upload(state, example) {
...
-  await record(state, browser, example);
-  const recordingId = await upload(state, example);
+  const exampleUrl = `${state.testingServer}/test/examples/${example}`;
+  const browser = target === "gecko" ? "firefox" : "chromium";
+  let recordingId;
+  await recordToFile(state, browser, exampleUrl);
+  recordingId = await upload(state, exampleUrl);
+  // await recordToCloud(state, browser, exampleUrl);
+  // recordingId = fs.readFileSync(state.exampleRecordingIdFile, "utf8").trim();
+  // fs.unlinkSync(state.exampleRecordingIdFile);
```
abstractalgo, line 105: `recordToCloud` isn't used.
What are the plans with this?

abstractalgo (commented):
Could you add brief descriptions to tests, just briefly explaining what each test is trying to achieve, and how/why you did it in that specific way. Literally a sentence or two, nothing much.

Generally, things make sense.

(disclaimer: I didn't do a run with this code and compare with the old one to see if exactly matched to the desired outcome of this change)

test/src/state.js
hbenl: It's set in `.github/actions/test/action.yml` and I sometimes set it in `.vscode/settings.json` if I want to run the tests locally in headless mode.

test/src/getExample.js
hbenl: No plans. This is another mode for recording examples that I experimented with. It works, but I found that in CI it's less reliable than the `recordToFile()`/`upload()` combination that we use now.

hbenl:
> Could you add brief descriptions to tests, just briefly explaining what each test is trying to achieve, and how/why you did it in that specific way. Literally a sentence or two, nothing much.

There's a description at the top of each test file (the first argument to the `Test.describe()` call)

abstractalgo (approved):

public/test/examples/bundles/README.md
```diff
@@ -0,0 +1,12 @@
+This folder contains the sources for the bundles used in E2E tests.
```
jcmorrow, line 1: These additional README's are great. My only concern is that docs that aren't linked from the root README might be used less or get stale. What do you think about dropping a single line in the main README like `there are details on things like <source map bundling> and <building the CRA> in the test directory` under the `Testing` heading?

test/manifest.js
```diff
@@ -44,15 +44,13 @@ module.exports = [
     // Not supported on chromium, needs source maps.
     // https://github.com/RecordReplay/chromium/issues/5
     targets: ["gecko"],
-    pending: true,
```
jcmorrow, line 47: Sorry, this came from my ruby lexicon 😄

jcmorrow (approved):
Looks great! What do you think about adding another config key to tests for something like `recordingId`, for tests where we don't want to live record the examples? Maybe that's silly because the whole point is to also exercise the driver/linker?

public/test/examples/bundles/README.md
hbenl: I thought about it, but these are only necessary to rebuild the examples, which is extremely rare (this is the first time we had to do it and I don't expect it to happen again any time soon), so I think adding them to the root README would be a distraction.

hbenl:
> What do you think about adding another config key to tests for something like recordingId, for tests where we don't want to live record the examples?

If you don't want to make new example recordings, just add the ones to use to `test/example-recordings.json`. In fact, if you use the tests locally, every example will be recorded exactly once and then automatically added to that file (that mechanism was broken but is fixed in this PR). If you want to ignore this file and always create a new example recording, you can do so by setting the `SHOULD_RECORD_EXAMPLES` environment variable or adding the `--record-examples` argument.
In CI, the test runs always start with an empty `test/example-recordings.json`, ensuring that we'll always use fresh example recordings.

PR #5294, opened 2022-02-11
Load the global object's properties/variables for autocomplete

### LINKED ISSUE #5261: Autocomplete should include global window properties
Here's an example where typing `cookie` does not suggest `cookieArr` which is available in the global namespace.

[STR Replay](https://app.replay.io/recording/ginger-cafe-sunnyvale-sunnyvale-ca-restaurant-or-menu-delivery-or-seamless--ceaec835-3d7f-4135-adad-674e2ae0d400?point=1622592768422527903165279824249350&time=1118.6139543799118&hasFrames=true)

This is a somewhat hard problem since we would need the set of global window properties to search on, but should be doable.

[Replay of bug](https://app.replay.io/recording/ginger-cafe-sunnyvale-sunnyvale-ca-restaurant-or-menu-delivery-or-seamless--47f0e5b4-11af-499d-8fb6-49708fdbd40d)

jaril:
Fix #5261.

This slows it down a tiny bit. Don't ship until we have some near-unnoticeable debounce mechanism (~500ms) in place to keep things smooth.

### DIFF
--- a/src/devtools/client/webconsole/utils/autocomplete.ts
+++ b/src/devtools/client/webconsole/utils/autocomplete.ts
@@ -149,8 +149,31 @@ function getBinding(name: string, scope: Scope) {
 function getBindingNames(scope: Scope): string[] {
   return scope.bindings.map(b => b.name);
 }
+function getGlobalActor(scopes: Scope) {
+  let globalActor = scopes;
+
+  while (globalActor.parent) {
+    globalActor = globalActor.parent;
+  }
+
+  return globalActor;
+}
 function getGlobalVariables(scopes: Scope) {
-  return getPropertiesForObject(scopes.parent.object?.getObject());
+  const globalActor = getGlobalActor(scopes);
+  const globalFront = globalActor.object;
+
+  if (!globalFront) {
+    return [];
+  }
+
+  // This gets the global object's children in the background. This happen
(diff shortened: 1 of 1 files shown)


src/devtools/client/webconsole/utils/autocomplete.ts
```diff
@@ -149,8 +149,31 @@ function getBinding(name: string, scope: Scope) {
...
+function getGlobalActor(scopes: Scope) {
+  let globalActor = scopes;
+
+  while (globalActor.parent) {
+    globalActor = globalActor.parent;
+  }
+
+  return globalActor;
+}
 function getGlobalVariables(scopes: Scope) {
```
abstractalgo, line 161: Should we maybe trigger this `onMount` or so, as it's done only once, it's very likely by the time it finishes fetching children the user hasn't even started using the autocomplete.

jasonLaster: this is interesting because we should probably do it on mount, but we might also need to do it when we pause because each pause could have new globals

jaril: Exactly. We could be more responsive and start fetching the global's children once paused, but I'd rather take it more conservatively for the first pass.

jasonLaster (approved):

PR #5295, opened 2022-02-11
Use a for loop when loading sources

jcmorrow:
Ummmm... do not ask me how I decided to try this change, I honestly have no idea. I think I've just spent a lot of time on JS twitter recently and that made me suspicious of what the `for ... of` was doing under the hood. So I tried to change it to a regular for loop.

Before:
I tried to time it, but I couldn't get it to work, because my tabs kept crashing

After:
The `await ensureAllSources` and the loop took `9741ms`.
And the memory of the process did not spike at all.

I think there are other wins here, but this one seems huge?

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -357,8 +357,6 @@ class _ThreadFront {
     });
 
     await this.ensureAllSources();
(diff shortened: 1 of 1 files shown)


jcmorrow:
nevermind, I was right not to trust these results, `keys.length` does not behave how I expect here, and returns 0 😆

PR #5299, opened 2022-02-12
Memoize alternate source IDs

jcmorrow:
I will be honest: I don't *really* understand what this code does, so I
am prepared for this be a big mistake. However, I recently profiled a
Replay that was loading slowly. It had 51,000 sources, which was clearly
the thing that was causing issues somehow. I also noticed that when
loading this Replay, my browser tab would often crash with an
out-of-memory error. Trying to *record* the thing loading was, for me,
totally impossible. At some point I had to force quit Replay while
trying to record this thing, because it was currently using 11GB of
memory!

By digging in with the profiler and with Replay (I recorded Replays with
smaller numbers of sources so that I could analyze them for order of
magnitude differences) I saw that one loop in particular was hit a *lot*
of times. It was pretty unclear to me what it was doing, but by counting
the hits during load of the broken Replay I learned that during this
loading process that loop was hit over 69 *million* times. I decided
that since there were only 51,000 sources, it was pretty likely that we
were duplicating work in there *somewhere*, so I memoized the function.
After that it was called 47,000 times, which is... roughly 1:1 with
sources?

Like I said, I did my best to understand what this code was doing, but I
never really got there, so this whole thing might be a mistake, but I
figured it was worth a shot (and I suspect that if this *is* a mistake
then @hbenl will probably set me straight pretty quickly, because I
think he's much more familiar with this code than I am).

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -136,6 +136,8 @@ class _ThreadFront {
   // Map sourceId to info about the source.
   sources = new Map<string, Source>();
 
+  alternateSourceIds: Map<SourceId, Set<SourceId>> = new Map();
+
   private searchWaiters: Map<string, (params: SearchSourceContentsMatch[]) => void> = new Map();
   private fnSearchWaiters: Map<string, (params: FunctionMatch[]) => void> = new Map();
 
@@ -338,8 +340,8 @@ class _ThreadFront {
     this.onSource = onSource;
 
     client.Debugger.findSources({}, sessionId).then(() => {
-      this.groupSourceIds();
       this.hasAllSources = true;
+      this.groupSourceIds();
       this.allSourcesWaiter.resolve();
     });
     client.Debugger.addNewSourceListener(source => {
@@ -1095,7 +1097,7 @@ class _ThreadFront {
     const groups = [];
     while (sourceIds.length) {
       const id = sourceIds[0];
-      const group = this._getAlternateSourceIds(id).filter(id => sourceIds.includes(id));
+      const group = [...this._getAlternateSourceIds(id)].filter(id => sourceIds.includes(id));
       groups.push(group);
       sourceIds = sourceIds.filter(id => !group.includes(id));
     }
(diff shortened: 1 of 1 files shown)


jcmorrow:
I'm fairly sure that this is one good improvement, but there are more that we need, and I think I know *roughly* where the hot spots are.

src/protocol/thread/thread.ts
```diff
@@ -1104,8 +1106,15 @@ class _ThreadFront {
 
   // Get all original/generated IDs which can represent a location in sourceId.
   private _getAlternateSourceIds(sourceId: SourceId) {
+    const original = sourceId;
```
hbenl, line 1109: unused variable

src/protocol/thread/thread.ts
```diff
@@ -136,6 +136,8 @@ class _ThreadFront {
   // Map sourceId to info about the source.
   sources = new Map<string, Source>();
 
+  alternativeSourceIds: Map<SourceId, Set<SourceId>> = new Map();
```
hbenl, line 139: could we call this `alternateSourceIds` (just to be consistent)?

hbenl (changes requested):
This is a great find. The only pitfall is that `_getAlternateSourceIds()` may be called before all sources were loaded (although I'm not sure that it is and if it is I'm not sure whether we should change that) and then its result may change. But usually it's called first from https://github.com/RecordReplay/devtools/blob/ee817ae0ca0f9ae4154337c4a730bdaafa2d600f/src/protocol/thread/thread.ts#L341 right after loading the sources.
So, to ensure that we don't mess anything up we should only cache the alternate sourceIds if `this.hasAllSources` has already been set to `true`. And we should set `thishasAllSources` to `true` before calling `this.groupSourceIds()`. Then we're on the safe side and should still get the full benefits of this cache.

jcmorrow:
> The only pitfall is that _getAlternateSourceIds() may be called before all sources were loaded (although I'm not sure that it is and if it is I'm not sure whether we should change that) and then its result may change.

Yes! Ok I was wondering this as well, and I figured something along the lines of just waiting to memoize would work, thanks @hbenl !

jcmorrow:
Yeah, following up on this: moving the memoization to only happen if `hasAllSources` is true bumped the call count from 49,000 to 51,000. I think waiting would *probably* be fine, but I'm a little more nervous to make a change that would actually block execution, so I think I'll stick with the `if(hasAllSources)` check.

jcmorrow:
> Yeah, following up on this: moving the memoization to only happen if `hasAllSources` is true bumped the call count from 49,000 to 51,000. I think waiting would _probably_ be fine, but I'm a little more nervous to make a change that would actually block execution, so I think I'll stick with the `if(hasAllSources)` check.

Oh, also, I just realized that this recording has 51,000 sources, so probably this bump indicates that it's working *better*!

jasonLaster:
Looks great, but will leave it to @hbenl to do the proper PR. Might be nice to have a replay for posterity here

src/protocol/thread/thread.ts
```diff
@@ -1104,22 +1106,32 @@ class _ThreadFront {
...
+      rv.add(currentSourceId);
+      const sources = this.sources.get(currentSourceId);
       assert(sources);
       const { generatedSourceIds } = sources;
-      (generatedSourceIds || []).forEach(id => worklist.push(id));
-      const originalSourceIds = this.originalSources.map.get(sourceId);
-      (originalSourceIds || []).forEach(id => worklist.push(id));
+      worklist.concat(generatedSourceIds || []);
+      const originalSourceIds = this.originalSources.map.get(currentSourceId);
+      worklist.concat(originalSourceIds || []);
```
hbenl, line 1128: I missed this the first time: `worklist.concat()` will not modify `worklist` but create a new array instead.

hbenl (changes requested):
You should also swap these 2 lines so that we start caching as soon as possible: https://github.com/RecordReplay/devtools/blob/2d1fb5ed51f599ce02a7ead72d2f5de306f3e0f2/src/protocol/thread/thread.ts#L341-L342

hbenl (approved):

PR #5304, opened 2022-02-13
Library cleanup

jonbell-lot23:
Quick Loom explaining changes:
https://www.loom.com/share/589f4358630d46148a7548ba5b133c84

* Edit button has hover effect
* View settings has better touch target
* "Your Library" made consistent (previously we had My Library next to Your Library which looks strange)


jonbell-lot23:
> While we're at it, maybe aligning the top search with the listing might be good too.
> 
> image

The top portion is wider by design:
image

PR #5305, opened 2022-02-14
Fixing selection state when editing title

jonbell-lot23:
Loom: https://www.loom.com/share/ebd4182356bf47208f7f0e886c514a42

I've removed the selection state because it looks awkward.

Prod:
image

Proposed:
image

### DIFF
--- a/src/ui/components/Header/Header.tsx
+++ b/src/ui/components/Header/Header.tsx
@@ -77,7 +77,7 @@ function HeaderTitle({
   const canEditTitle = recording.userRole !== "none";
 
   const className =
-    "ml-2 text-lg p-0.5 bg-transparent border-black whitespace-pre overflow-hidden overflow-ellipsis";
+    "ml-2 text-lg py-0.5 whitespace-pre overflow-hidden overflow-ellipsis title-container";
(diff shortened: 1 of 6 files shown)


src/ui/components/Header/Header.tsx
```diff
@@ -120,7 +119,8 @@ function HeaderTitle({
 
   return (
     <span
-      className={cx(className, "input focus:border-blue-500 focus:ring-primaryAccent", {
+      style={{ outline: "none", background: "inherit" }}
+      className={cx(className, "input m-5 focus:bg-blue-500", {
```
jasonLaster, line 123: I think you might be able to continue using tailwind
https://tailwindcss.com/docs/outline-style#removing-outlines

not sure about the background though

jasonLaster (approved):
Looks good. One tailwind comment though

src/ui/components/Header/Header.tsx
jonbell-lot23: It's crazy, I worked on this for ages and just could not get tailwind working. Had to go with style to make it finally take.

PR #5306, opened 2022-02-14
Revert Prettier Tailwind plugin

abstractalgo:
Revert of the PR #5254.


jonbell-lot23 (commented):
I don't think this revert is necessary anymore. I think the files I thought had been overwritten were just not pushed live yet.

PR #5307, opened 2022-02-14
Revert "Mobile login UI tweak"

jaril:
Reverts RecordReplay/devtools#5303


jaril:
image

@jonbell-lot23 Heads up — I had to revert this because it broke the sharing modal layout. Will need to revisit, let me know if I can help!

PR #5314, opened 2022-02-15
Smaller video

jonbell-lot23:
In order to give the console more space to breathe, I changed our resize values from 30% to 10%. If people want to shrink the window that much, more power to 'em!

### DIFF
--- a/src/ui/components/SecondaryToolbox/index.tsx
+++ b/src/ui/components/SecondaryToolbox/index.tsx
@@ -48,7 +48,7 @@ function PanelButtons({
   };
 
   return (
-    <div className="theme-tab-font-size flex flex-row items-center overflow-hidden narrow:bg-black">
+    <div className="theme-tab-font-size flex flex-row items-center overflow-hidden">
       {!isNode && <NodePicker />}
       <button
         className={classnames("console-panel-button", {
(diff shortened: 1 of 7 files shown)


jasonLaster:
I'd like to see how this feels, but my initial impression is that i'm a bit nervous.

A big part of the aha moment for Replay is seeing what the video looks like at any point of time. If the video is initially 10%, then folks could miss that moment.

jonbell-lot23:
> I'd like to see how this feels, but my initial impression is that i'm a bit nervous.
> 
> A big part of the aha moment for Replay is seeing what the video looks like at any point of time. If the video is initially 10%, then folks could miss that moment.

Totally agree. I meant to make the minimum height smaller, but not the initial load. Will keep working on that.

jonbell-lot23:
> oh cool. yeah - that makes sense.
> 
> It's easy to look at the initial size by evaluating `localStorage.clear()` in the console and refreshing.

Thanks! So I confirmed that this code will work.

1. First load -- normal size (not tiny)
2. Resize video 1 to be tiny
3. Open video 2 -- it's also tiny because of localstorage settings
4. localStorage.clear on video 2 and reload -- normal size (not tiny)

Loom repro steps: https://www.loom.com/share/04c99be32f2e4a328c700fdc4646d766

PR #5316, opened 2022-02-15
Re-add babelrc. fixes #5169

### LINKED ISSUE #5169: re-enable preset-env
I'm finding it difficult to figure out what is going wrong in this [replay](https://app.replay.io/recording/cannot-get-endpoint--12e47a50-9816-480a-a3bd-a7f35f80f7ea?point=10384593718396974585983497085649680&time=4380.796754910333&hasFrames=true) where we do not advance past the loading screen.

One of the things that is making it difficult is that the async/await functions are compiled to generators. This goes away when `preset-env` is set to chrome: "95". There is probably a way to do this with SWC because it is less work :)

https://github.com/RecordReplay/devtools/pull/5049
https://babeljs.io/docs/en/babel-plugin-transform-regenerator

jasonLaster:
This is a temporary fix which for #5169.

I think the next version of Next.js will let us set a preset-env and go back to SWC.

In the meantime, we should re-add babel which will let us bring back native async/await which is a perf win and will make debugging significantly easier

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,15 @@
+{
+  "presets": [
+    [
+      "next/babel",
+      {
+        "preset-env": {
+          "targets": {
+            "chrome": "95"
+          }
+        }
+      }
+    ]
+  ],
+  "plugins": ["@babel/plugin-transform-react-display-name"]
+}


abstractalgo (approved):
Sounds good 👍

PR #5322, opened 2022-02-15
Cleanup of Command Button

abstractalgo:
Just saw a few things along the way, no functional changes.

### DIFF
--- a/src/ui/components/CommandPalette/CommandButton.tsx
+++ b/src/ui/components/CommandPalette/CommandButton.tsx
@@ -1,19 +1,23 @@
 import { formatKeyShortcut } from "devtools/client/debugger/src/utils/text";
-import React, { useEffect, useRef } from "react";
-import { connect, ConnectedProps } from "react-redux";
+import React, { FC, useEffect, useRef } from "react";
+import { useDispatch } from "react-redux";
+import cx from "classnames";
 import { actions } from "ui/actions";
 import { Command } from "./CommandPalette";
 
-type CommandButtonProps = PropsFromRedux & { command: Command; active: boolean };
+type CommandButtonProps = {
+  command: Command;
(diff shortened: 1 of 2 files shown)


src/ui/components/CommandPalette/CommandPalette.tsx
```diff
@@ -58,14 +58,14 @@ const COMMANDS: Command[] = [
   { key: "show_privacy", label: "Show Privacy" },
   { key: "show_replay_info", label: "Show Replay Info" },
   { key: "show_sharing", label: "Show Sharing Options" },
-];
+] as const;
```
jaril, line 61: nice

src/ui/components/CommandPalette/CommandButton.tsx
```diff
@@ -25,21 +29,18 @@ function CommandButton({ command, executeCommand, active }: CommandButtonProps)
     <button
       key={label}
       onClick={onClick}
-      className={`${bgColors} flex justify-between px-6 py-2 transition`}
+      className={cx(
```
jaril, line 32: Small nit: any reason for the cx instead of the default `classnames`? Only bringing it up since we use `cx` for context, but that's something that doesn't really make a whole lot of sense in replay land so will probably be done away with soon.

jaril (approved):
Just left a question but otherwise looks good to me!

src/ui/components/CommandPalette/CommandButton.tsx
abstractalgo: Multiple projects that used `classnames` library imported it as `cx` and nowadays I don't even think twice about it. No special reason, just shorter than "classnames", I guess.

It seems I'm not the only one following this "convention" based on the search through the codebase for "cx" that I did just now.

PR #5327, opened 2022-02-15
Changing div to button

jonbell-lot23:
Comment from Jaril

### DIFF
--- a/src/ui/components/Library/SidebarFooter.tsx
+++ b/src/ui/components/Library/SidebarFooter.tsx
@@ -12,10 +12,7 @@ function SidebarFooter({ setModal }: PropsFromRedux) {
   };
 
   return (
-    <button
-      className="flex flex-row space-x-2 bg-gray-700 p-4 cursor-pointer"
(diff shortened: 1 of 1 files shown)


src/ui/components/Library/SidebarFooter.tsx
```diff
@@ -12,17 +12,18 @@ function SidebarFooter({ setModal }: PropsFromRedux) {
   };
 
   return (
-    <div className="flex flex-row space-x-2 bg-gray-700 p-4 cursor-pointer" onClick={handleSettingsClick}>
+    <button
+      className="flex flex-row space-x-2 bg-gray-700 p-4 cursor-pointer"
```
jaril, line 16: don't need `cursor-pointer` we get that for free with the button! :)

jaril (approved):
Just one quick change but looks good!

src/ui/components/Library/SidebarFooter.tsx
jonbell-lot23: True, thanks!

PR #5333, opened 2022-02-16
Fix material icon loading detection

ryanjduffy:
## Issue

The added specificity of the `width` from `.material-icon-loaded .material-icon` was stomping on the tailwind specificity preventing easy assignment of the width to an icon

image

## Resolution

* Refactoring the "loading" class to be applied by default and then removed when loaded ensures that the higher specificity is only applied when loading and is removed when loaded so the desired size can take effect then.
* Also changed the hook design a bit to keep everyone inside their own boundary:
  * class is now applied to the app root node instead of the body to stay within the app's boundary
  * root App component uses the hook and passes a ref to this node so the hook isn't reaching out beyond it's boundary
  * Added some retry logic so the fonts can be ready within the first 10 seconds of rendering the app

image

### DIFF
--- a/public/fonts/material_icons/material_icons.css
+++ b/public/fonts/material_icons/material_icons.css
@@ -17,12 +17,6 @@
     url(/fonts/material_icons/MaterialIconsOutlined-Regular.otf) format("opentype");
 }
 
-.material-icon-loaded .material-icons,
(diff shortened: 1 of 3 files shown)


jaril (approved):
!!

PR #5336, opened 2022-02-16
Fixing SSO button width

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/5335


jaril:
That fixed it but that's not quite the root problem. We need a `flex-shrink-0` on the button so it doesn't try wrapping and mess up the whole layout.

image

jaril (dismissed):
👆

jaril:
> Thanks for the tip! I didn't know about that trick. But it still isn't fixing it:

I think it's because you added more text. It should work just fine for the current `Sign In` text.

> Ideally larger buttons would push the enclosing div wider.

This one's a bit more involved, since the reason it's overflowing in the first place in that picture is because there's a max width set on the modal in the first place that keeps it from making the container any larger.

jonbell-lot23:
Added the flex-shrink-0 but it also doesn't fix the root problem because it messes up our margins. So I kept the w-48 tweak to help with the margins and later we can make sure the overall component expands correctly when buttons get bigger.

Issue #4748, opened 2021-12-04
Update Accordion component


jasonLaster:
Our current accordion has a couple of issues:

1. the panels can be awkward sizes
2. the panels cannot be resized

this library should be a good starting point
https://github.com/DanFessler/react-panelgroup

jasonLaster:
After doing some more research I found:
* https://github.com/leefsmp/Re-Flex
* https://github.com/tomkp/react-split-pane

re-flex is pretty close, but has some weird behavior when you add in panes that are collapsible

jasonLaster:
closing as this is probably a nerd snipe and we can revisit down the road

jaril:
Bumping.

jcmorrow:
Some notes from our exploration of VS Code's accordiong: https://www.notion.so/replayio/Accordion-06d91f9d65d24cd88020fd9f5b7dc3c5

abstractalgo:
related #5317

PR #5337, opened 2022-02-17
Controlled Accordion API

jaril:
This makes some changes to the new Accordion component so we can apply it to the Source Explorer

https://user-images.githubusercontent.com/15959269/154586865-8865dab1-5af4-4330-9a13-eb2bdedb96a1.mov

.


jaril:
Closes #4748.

PR #5339, opened 2022-02-17
Bump `@apollo/client` and `graphql`

abstractalgo:
Bump Apollo Client and GraphQL because some of the Sentry errors seem like being caused by the internal implementation instead of surfacing failed requests to the requester, so a bump might help.

Changelogs:

- https://github.com/apollographql/apollo-client/releases (no major)
- https://github.com/graphql/graphql-js/releases (major 15→16, with breaking changes along the way but we seem unaffected)


package.json
```diff
@@ -27,10 +27,11 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
-    "@apollo/client": "^3.4.16",
+    "@apollo/client": "^3.5.9",
     "@auth0/auth0-react": "^1.2.0",
     "@headlessui/react": "^1.4.1",
     "@heroicons/react": "^1.0.1",
+    "@recordreplay/playwright": "^1.10.6",
```
hbenl, line 34: This was turned into an optional dependency in https://github.com/RecordReplay/devtools/commit/e74aabdd20bf411deb4b686e06149efd34425063, not sure if that is still necessary.

hbenl (approved):

package.json
abstractalgo: I keep having this problem where `npm i` always adds to `package.json` and I keep removing it manually. I briefly discussed this with Josh, and tried changing versions of node/npm but to no avail.

abstractalgo:
Spoke too soon. xD

`@apollo/client` 3.5.x changed how they export code and `ts-node` now gets confused when running mock tests and doesn't know how to deal with it. As a confirmation, I bumped Apollo only to 3.4.17 (latest before 3.5.x) and it all works.

I tried playing with different tsconfig options, changing `"moduleResolution": "nodenext",` seems to work differently but then it complains about some even weirder things (ES module must be imported dynamically; and types are missing (?)).

Related: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#esm-nodejs

PR #5352, opened 2022-02-18
Remove sourceMaps refence

jasonLaster:
There are a couple of places where our thunks pass in different options (sourceMaps, prefsService, client).

The sourceMaps reference is no longer needed because we now sourceMap on the client


jcmorrow (approved):
Delete it all!

PR #5386, opened 2022-02-18
Make the background a real gradient

jcmorrow:
This is a small change, but on my ultra wide monitor I've noticed that
the edge of the nebula that we use as a background image ends up peeking
out and looking kind of janky. This changes that container to have a
true CSS gradient background, which should both give us more control and
also make it feel more versatile across a lot of screen sizes.

I've also tried to make the way that we style the initial height of the
command panel explicit.

https://www.loom.com/share/92dcb8f76fcd493f92ddf06c2dcc85de

I did change the gradient stops a little from the loom, here's the current version:

CleanShot 2022-02-18 at 13 55 23

### DIFF
--- a/src/devtools/client/debugger/src/components/App.css
+++ b/src/devtools/client/debugger/src/components/App.css
@@ -32,6 +32,13 @@ button {
 
 .editor-container {
   width: 100%;
+  background: linear-gradient(
+    25deg,
+    rgba(255, 255, 255, 1) 0%,
+    rgba(255, 255, 255, 1) 30%,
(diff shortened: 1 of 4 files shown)


jaril (approved):
Looks good to me. As you said I'd try to get to color match it a bit better but apart from that, should be good!

Only other thing worth bringing up is I think @jonbell-lot23 may have meant to intentionally make the palette look like it's always displaying a full set of four items with no "peeking", especially for the default set of four items. There's a chance I totally got that wrong in which case I totally dig this let's just ship it!

PR #5387, opened 2022-02-18
Make sure we're able to measure the accordion container before rendering its panes

jaril:
https://user-images.githubusercontent.com/15959269/154768508-020c647b-b167-4a3f-9bef-010c64955fa5.mov


src/ui/components/NewAccordion/Accordion.tsx
```diff
@@ -195,7 +195,7 @@ export const Accordion: FC<{
 
   return (
     <div className="relative flex h-full flex-col overflow-auto" ref={containerRef}>
-      {newChildren}
+      {state.domHeight ? newChildren : null}
```
jcmorrow, line 198: DOMheight ftw 📐

PR #5389, opened 2022-02-18
Add a mixpanel tracker for whenever font loading times out

jaril:
I'm occasionally seeing fonts just never come in properly. This should give us some idea of how often that's happening.


ryanjduffy:
Should this be in honeycomb instead of mixpanel? Are we tracking other errors in mixpanel?

ryanjduffy (approved):

PR #5393, opened 2022-02-18
Make loading and login screens the same

jcmorrow:
This fixes at least the login/initial tab screen, but I'm guessing there might be other places to fix as well cc @jaril


abstractalgo:
@jonbell-lot23 might be cool to wrap this up, as it is a step towards a more unified look

PR #5412, opened 2022-02-21
Bump `@auth0/auth0-react`

abstractalgo:
At [1.5.0](https://github.com/auth0/auth0-react/releases/tag/v1.5.0) they added TS generic for `useAuth0` hook and changed the `User` type, which now requires more changes to our code as well.

TODO:

- [ ] fix types


abstractalgo:
After brief discussion with Ryan, this is not a priority and we might end up killing auth0 in favor or workos or something else, so really not a priority.

PR #5415, opened 2022-02-21
Dependencies cleanup

abstractalgo:
Some more cleaning of unused packages or moving them to `devDependencies`.


package.json
```diff
@@ -41,7 +41,6 @@
     "@tiptap/extension-placeholder": "^2.0.0-beta.48",
     "@tiptap/react": "^2.0.0-beta.108",
     "@tiptap/starter-kit": "^2.0.0-beta.183",
-    "babel-plugin-add-react-displayname": "0.0.5",
```
hbenl, line 44: Isn't this still needed?
https://github.com/RecordReplay/devtools/blob/b6e32728d1b8144dd8ee9a9e55b726e4da26dbf4/.babelrc#L14

hbenl (approved):

package.json
abstractalgo: Didn't show up in the file search, but VS Code just had a hiccup (doesn't refresh the search panel results always right away). You're right. 👍

PR #5417, opened 2022-02-21
Set up Husky

abstractalgo:
Following the reasoning from https://github.com/RecordReplay/devtools/pull/5278#issuecomment-1046984585, this change adds a Husky pre-commit hook that lints code.

Also, enabled `@typescript-eslint/explicit-function-return-type` rule.

Things to resolve:

- [ ] we run lint GHA as part of CI; does it now become obsolete?
- [ ] TSC vs ESlint checks mismatch in CI


abstractalgo:
It seems that any solution introduces a bit more complexity and has a cost in additional time for processing (Husky, Trunk or anything else). As far additional tooling for linting and formatting, maybe it's best to wait, because...

...a problem of its own is the fact that our linter is wrongly set up. `tsc` is running type checks across the entire repo, while `eslint` runs only partially (due to Next; see https://nextjs.org/docs/basic-features/eslint#linting-custom-directories-and-files). If we correctly set up ESlint for the entire repo, then we already have: `612 problems (559 errors, 53 warnings)`. We'll start addressing those in a separate PR.

PR #5434, opened 2022-02-23
Use custom login page from other login buttons

### LINKED ISSUE #5432: Login button on recording page goes to Google instead of our new auth page
This login button:
Screen Shot 2022-02-22 at 3 51 11 PM

ryanjduffy:
## Issue

Fixed #5432 - A couple login buttons do not redirect through our login page and instead go directly to auth0

## Resolution

Add `loginAndReturn` to our custom `useAuth0` hook which consolidates the `returnTo` query param calculation and uses the nextjs router to redirect to our custom login page.

### DIFF
--- a/src/ui/components/shared/LoginModal.tsx
+++ b/src/ui/components/shared/LoginModal.tsx
@@ -3,13 +3,7 @@ import useAuth0 from "ui/utils/useAuth0";
 import Modal from "ui/components/shared/Modal";
 
 function LoginModal() {
-  const { loginWithRedirect } = useAuth0();
-
-  const onClick: React.MouseEventHandler = e => {
-    loginWithRedirect({
-      appState: { returnTo: window.location.pathname + window.location.search },
-    });
-  };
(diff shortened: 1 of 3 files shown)


src/ui/components/shared/LoginModal.tsx
```diff
@@ -25,7 +19,7 @@ function LoginModal() {
           <div className="flex flex-col items-center">
             <button
               type="button"
-              onClick={onClick}
+              onClick={() => loginAndReturn()}
```
jaril, line 22: Can just do `onClick={loginAndReturn}` here, no?

jaril (approved):
Got a nit but this looks good!

src/ui/components/shared/LoginModal.tsx
ryanjduffy: We could. I tend to avoid passing non-event handlers to event callbacks so things don't accidentally break if we change the signature of `loginAndReturn`. TS will probably catch that for us but it's an easy regression to introduce.

PR #5440, opened 2022-02-23
Correct ESlint setup

abstractalgo:
Following https://github.com/RecordReplay/devtools/pull/5417#issuecomment-1048840332, this is a more correct way to setup ESlint. Until now, it was linting only `/pages` folder for us.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ node_modules
 src/devtools/client/shared/vendor/
 src/devtools/client/shared/sourceeditor/codemirror
 sourcemapVisualizer.js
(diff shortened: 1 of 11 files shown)


markerikson:
While it's not entirely lint-related, any chance we could remove the `"endOfLine": "lf",` line from `.prettierrc`?  I'm on Windows, and so Git checks out all files with `CRLF` line endings by default.  This causes Prettier to flag every single line as an issue :)

abstractalgo:
@markerikson I don't see why not.

hbenl (approved):

PR #5446, opened 2022-02-23
Making the “long Replay” alert more obvious

jonbell-lot23:
Used to look like this:
image

Now looks like this:
image


src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -87,19 +87,9 @@ function Actions({ onDiscard, status }: { onDiscard: () => void; status: Status
...
-            style={{ width: "200px" }}
-          >
-            {`Replays work best when under 2 minutes`}
-          </div>
-        }
-      >
-        <MaterialIcon className="select-none">warning</MaterialIcon>
-      </PortalTooltip>
+    <div className="absolute place-content-center bottom-2 right-2 flex rounded-md bg-gray-500 p-2 text-white shadow-lg text-xs">
+      <MaterialIcon className="select-none">warning</MaterialIcon>{" "}
```
jaril, line 91: Small nit, but I'd take out the `${" }` on line 91 and just add a `space-x-2` or something in the parent div on line 90.

jaril (approved):
One nit but looks good otherwise!

src/ui/components/UploadScreen/UploadScreen.tsx
jonbell-lot23: Totally agree, thanks!

PR #5441, opened 2022-02-23
Clean up the trimming UI

jaril:
Fix #3396.

https://user-images.githubusercontent.com/15959269/155363839-b8b5b628-a50c-4829-85ab-d8dad595979d.mov

PR #5448, opened 2022-02-24
Hook the trimmer to the backend

jaril:
Follow up to #5441.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -1,6 +1,6 @@
 import { ExecutionPoint, PauseId } from "@recordreplay/protocol";
 import { Pause, ThreadFront } from "protocol/thread";
-import { client, log } from "protocol/socket";
+import { client, log, sendMessage } from "protocol/socket";
 import {
   getGraphicsAtTime,
   paintGraphics,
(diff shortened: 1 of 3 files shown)


src/ui/actions/timeline.ts
```diff
@@ -526,3 +526,28 @@ export function updateTrimRegion(operation: TrimOperation): UIThunkAction {
...
+      { region: { begin: 0, end: zoomRegion.endTime } },
+      window.sessionId
+    );
+    await sendMessage(
+      "Session.loadRegion",
+      {
+        region: { begin: trimRegion.startTime, end: trimRegion.endTime },
+      },
+      window.sessionId
+    );
```
jasonLaster, line 551: We should move this to the ThreadFront

```js
export function syncTrimmedRegion(): UIThunkAction {
  return async ({ getState }) => {
    const state = getState();
    const zoomRegion = getZoomRegion(state);
    const trimRegion = getTrimRegion(state);

    if (!trimRegion) {
      return;
    }

    await ThreadFront.trimRegion()
```


```js
trimRegion(startTime, endTime) {
  await Promise.all(client.Session.unloadRegion(), client.Session.loadRegion())
}
```

jaril:
This doesn't quite work yet. 

The idea was that when we close the trimmer, we can "save" the changes by letting the backend know what parts of the recording the user would like loaded. Following the existing [API](https://static.replay.io/protocol/tot/Session/#method-loadRegion), it looks like the way to do that is:
1. unload the entire recording
2. load the trimmed section

From my understanding of the docs that should get us to a point where the backend only loads the trimmed section. But in practice it behaves differently. Will try to get a replay.

jaril:
Documenting the behavior here while using this [replay](http://localhost:8080/recording/hello-world--7aede9d1-37f5-431a-867a-ae8c8bce8106):

1. Unload the full recording (~33s) with `Session.unloadRegion`
2. Loaded Regions keep updating until it stops with the loaded region indicated as `32759-33080`
3. Load regions with `Session.loadRegion` from 0-5s, so `0-5000`
4. Loaded regions go from [`0-527`, `32759-33080`], then [`0-2120`, `32759-33080`] `0-2120`. 
5. Then it overshoots it to [`0-10282`, `32759-33080`]. Then it overshoots to [`0-26913`, `32759-33080`].
6. Then for some reason the start moves from 0, as it moves to [`10808-27430`, `32759-33080`]. 
7. Then that keeps going until there's just [`10808-33080`]

jaril:
For now this is blocked by: [4773](https://github.com/RecordReplay/backend/issues/4773)

PR #5454, opened 2022-02-24
Add trimming and column breakpoint experimental settings

jasonLaster:
Adds a couple of experimental settings


jasonLaster:
This code is a bit ugly. It also doesn't update the trimming button, so i'm thinking about ways it can be improved. Not a high priority though

jasonLaster:
We should also add dark mode!

jaril:
https://user-images.githubusercontent.com/15959269/155587429-9662bf53-64bc-49a6-9dff-2bcbfd5e6e0b.mov

PR #5456, opened 2022-02-24
Don't scroll CodeMirror on space press

jcmorrow:
I *mostly* fixed this months ago in https://github.com/RecordReplay/devtools/commit/cdaea8c11500f563749e6bb552be8b5533e2f1fe but over the past few days I noticed that I still sometimes scroll the editor when I'm trying to toggle play/pause in the video editor. It turns out that when you click on a breakpoint inside of a breakpoint panel the element that gets focus is `CodeMirror-scroll`. I'm not sure why that thing gets focus, but if I had to guess I would say that CodeMirror detects a click on itself and tries to get ready to scroll (which would be a pretty reasonable shortcut in most cases). In our case, though, it's really annoying to scroll the editor when you instead mean to play or pause the video.

### DIFF
--- a/src/ui/components/App.tsx
+++ b/src/ui/components/App.tsx
@@ -107,12 +107,24 @@ function App({ children, modal }: AppProps) {
   useEffect(() => {
     // Stop space bar from being used as a universal "scroll down" operator
     // We have a big play/pause interface, so space makes way more sense for that.
-    window.addEventListener("keydown", function (e) {
-      if (e.code === "Space" && e.target === document.body) {
+
+    const stopCodeMirrorScroll = (e: KeyboardEvent) => {
+      if (e.code !== "Space") {
+        return;
+      }
+
+      if (
+        e.target === document.body ||
+        (e.target?.hasOwnProperty("classList") &&
+          (e.target as Element).classList.contains(".CodeMirror-scroll"))
+      ) {
         e.preventDefault();
       }
-    });
-  });
+    };
+
+    window.addEventListener("keydown", stopCodeMirrorScroll);
+    return () => window.removeEventListener("keydown", stopCodeMirrorScroll);
+  }, []);
 
   useEffect(() => {
     document.body.parentElement!.className = enableDarkMode ? "theme-dark" : "theme-light";


src/ui/components/App.tsx
```diff
@@ -107,8 +107,17 @@ function App({ children, modal, theme }: AppProps) {
...
-      if (e.code === "Space" && e.target === document.body) {
-        e.preventDefault();
+      if (e.code === "Space") {
+        const codeMirrorScroll = document.querySelectorAll(".CodeMirror-scroll");
+        let isCodeMirror = false;
+        for (let el of codeMirrorScroll) {
+          if (el === e.target) {
+            isCodeMirror = true;
+          }
+        }
```
jasonLaster, line 117: could this be `const isCodeMirror = e.target.classList.has(.CodeMirror-scroll)`

jcmorrow: Yep, tsc was yelling at me so I was taking evasive maneuvers but got there eventually 😆

src/ui/components/App.tsx
```diff
@@ -107,8 +107,14 @@ function App({ children, modal, theme }: AppProps) {
     // Stop space bar from being used as a universal "scroll down" operator
     // We have a big play/pause interface, so space makes way more sense for that.
     window.addEventListener("keydown", function (e) {
-      if (e.code === "Space" && e.target === document.body) {
-        e.preventDefault();
+      if (e.code === "Space") {
```
abstractalgo, line 110: Can use early exit to make the code just slightly more readable.

src/ui/components/App.tsx
```diff
@@ -107,8 +107,14 @@ function App({ children, modal, theme }: AppProps) {
...
+        if (
+          e.target === document.body ||
+          (e.target?.hasOwnProperty("classList") &&
+            (e.target as Element).classList.contains(".CodeMirror-scroll"))
+        ) {
+          e.preventDefault();
+        }
       }
     });
   });
```
abstractalgo, line 120: Needs a dependencies array and a return value as a function that cleans up the event listener from `window` (not that it probably makes literally any difference, but it's a good practice).

abstractalgo:
I actually don't know how to repro this, replay/loom would be much appreciated 😄 
I was wondering if there's a way to prevent this in the codemirror itself, because leaking this as a responsibility to a "global handler" is definitely not ideal 😕

jcmorrow:
https://www.loom.com/share/16bb5d82008243d2b6c5245f26422e93 @abstractalgo 

And as mentioned here, we could also probably put the handler on the breakpoint component. I'm not sure if I feel much better about that though 🤷

abstractalgo:
> https://www.loom.com/share/16bb5d82008243d2b6c5245f26422e93 @abstractalgo
> 
> And as mentioned here, we could also probably put the handler on the breakpoint component. I'm not sure if I feel much better about that though 🤷

As you mentioned, it seems reasonable to try to prevent bubbling from within the timeline component for logpoints. But if it's anything more complex, the current solution will suffice. 👍  @jcmorrow

abstractalgo (approved):

abstractalgo (approved):

abstractalgo:
I'm sorry, my Internet connection was acting up 😄

jcmorrow:
> > https://www.loom.com/share/16bb5d82008243d2b6c5245f26422e93 @abstractalgo
> > And as mentioned here, we could also probably put the handler on the breakpoint component. I'm not sure if I feel much better about that though 🤷
> 
> As you mentioned, it seems reasonable to try to prevent bubbling from within the timeline component for logpoints. But if it's anything more complex, the current solution will suffice. 👍 @jcmorrow

Yeah I tried this, it felt jankier and less effective (because there are several click targets that can end up focusing that element so you would have to add the handler to all of them 😦 )

PR #5457, opened 2022-02-24
Only wait for the first 5 seconds of video

jcmorrow:
OK! Believe it or not it took me a week to write these 50 lines of code 😅 

Originally, I thought that this PR was going to be using the new backend paints work to load paints separately from the rest of the basic processing resources and get the user into the viewer immediately. When I thought that was going to be the approach, there was a problem:

- We wait on the dispatcher promise that gets returned from `findPaints` to know when it's ok to do things like trying to get paint contents, play video, etc. and on the dispatcher side, that promise gets blocked by the rest of basic processing data, even though from a linker perspective they are gathered separately and don't have all that much to do with each other anymore. In that world there were more dispatcher changes required to do anything like what we're doing here.
- It turns out a lot of that is not necessary, and in fact, speeding up the end of the `findPaints` promise isn't even the best case if it was an easy thing to do. Instead, the insight here is this: the user waits a long time right now to view long replays, with long basic processing times (usually a bad run-to-point ratio). But *while they are waiting the front-end is receiving all of the data it needs to do video things for most of the video*. It doesn't matter when the paints promise resolves, because we're getting lots of useful paints well before it resolves.
- The second insight is something like this: previously we blocked on getting paint *points*, before we even attempted to get paint *data*. Now, as soon as we start getting points, we also start fetching contents (at least for the first five seconds of the video, so that at least we can show the user *something* to confirm that the recording worked, etc). Once they are in the viewer, they can click around the timeline, but if they go past the end of the paint points that we currently know about, we'll pause to buffer up to that point before playing the video.
- There's probably more to do here, for instance, being in the DevTools is not particularly enjoyable while they are loading (although, to our credit, it's not really *that* broken and some stuff like Network Monitor data already loads asynchronously).
- Also, we can probably do more to signal to the user the various states of loading that we are in at any time. As I discussed with @jasonLaster , it's theoretically possible for paints to come in out of order, such that we are missing a chunk of paint points in the middle of the recording, and we accidentally try to play the video over them, which would result in the video being stuck in the last retrieved frame for the duration of the missing section. This should be pretty rare, and only happen if something bizarre happens like several crashes in one paints branch.

This has also made me think more about our video loading in general and after getting some feedback from @jasonLaster  I suspect there are some major improvements to be made by doing something like sending paints region-by-region (or branch section by branch section) as compressed data rather than one at a time, uncompressed over the socket.

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -1,7 +1,7 @@
 // Routines for managing and rendering graphics data fetched over the WRP.
 
 import { ThreadFront } from "./thread";
-import { assert, binarySearch } from "./utils";
+import { assert, binarySearch, defer, Deferred } from "./utils";
 import { DownloadCancelledError, ScreenshotCache } from "./screenshot-cache";
 import ResizeObserverPolyfill from "resize-observer-polyfill";
 import {
@@ -18,6 +18,9 @@ import { Canvas } from "ui/state/app";
 import { setCanvas, setEventsForType, setVideoUrl } from "ui/actions/app";
 import { setPlaybackPrecachedTime, setPlaybackStalled } from "ui/actions/timeline";
 import { getPlaybackPrecachedTime, getRecordingDuration } from "ui/reducers/timeline";
+import { max } from "lodash";
+
+const FIVE_SECONDS = 5000;
 
 const { features } = require("ui/utils/prefs");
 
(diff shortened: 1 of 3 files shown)


src/protocol/graphics.ts
```diff
@@ -110,10 +111,32 @@ const gMouseClickEvents: MouseEvent[] = [];
...
     const paintHash = screenShots.find(desc => desc.mimeType == "image/jpeg")!.hash;
     insertEntrySorted(gPaintPoints, { point, time, paintHash });
+
+    if (gAllNecessaryPointsReceived) {
+      // We are all set on the loading front, no need to practively grab these
+      // paints
+      return;
+    }
+
+    if (time < 5_000) {
```
abstractalgo, line 131: We can move (the magic) `5_000` to a reusable const somewhere.

hbenl (approved):

src/protocol/graphics.ts
```diff
@@ -193,7 +218,11 @@ class VideoPlayer {
 export const Video = new VideoPlayer();
 
 let onRefreshGraphics: (canvas: Canvas) => void;
-let paintPointsWaiter: Promise<findPaintsResult>;
+let paintPointsWaiter: Deferred<void> = defer();
+let hasAllPaintPoints = false;
+paintPointsWaiter.promise.then(() => {
```
loganfsmyth, line 223: It doesn't look like this ever resolves?

jcmorrow: Good point. This is a combination of a couple of different implementations that I tried 😓

Issue #5396, opened 2022-02-20
"Visualise sourcemap" link shouldn't appear if there are no source maps


jonbell-lot23:
<img width="423" alt="image" src="https://user-images.githubusercontent.com/9154902/154828773-c2687338-34c2-4059-9991-83f264d497a5.png">

hbenl:
This is already implemented. The reason for the confusion is the "No sourcemaps found" claim which is wrong, see #5345

jonbell-lot23:
We may still have a bug. This Replay has no sourcemaps, yet shows the "visualise sourcemaps" link:
https://app.replay.io/recording/flicker-bug--7703cce9-ece6-4a58-9737-2d3049ab39c2?point=4218741197564930438686037636022395&time=1583&hasFrames=false#

PR #5458, opened 2022-02-24
Fix detection of sourcemaps for pretty-printed sources (#5396)

hbenl:


jasonLaster:
Can you share a replay of the fix? It would be nice see these IF statements hit

hbenl:
This example shows a non-pretty-printed source with sourcemap: https://app.replay.io/recording/docprodbundle-replay--4cb8e348-bf71-4969-b070-4e3a19183d1a
This example shows two pretty-printed sources, first without and then with sourcemap: https://app.replay.io/recording/docprodbundle-replay-replay--bc17d3d4-a445-40b2-9945-1bf9b89897bb

jasonLaster (approved):

PR #5471, opened 2022-02-25
Add eslint recommend

jasonLaster:


src/ui/setup/dynamic/devtools.ts
```diff
@@ -1,13 +1,6 @@
 import { bindActionCreators, Store } from "redux";
 
-import {
-  sessionError,
-  uploadedData,
-  CommandMethods,
-  CommandParams,
-  CommandResult,
-  CommandHasPauseId,
```
jasonLaster, line 9: eslint found these unused variables

PR #5475, opened 2022-02-25
Only wait for the first 5 seconds of video (2)

jcmorrow:
This is a fixed up version of https://github.com/RecordReplay/devtools/pull/5457. The problem with 5457 was that recordings could be:

- Node based, in which case waiting for video makes no sense
- Shorter than 5 seconds, in which case waiting for 5 seconds of paints to come in also makes no sense.

Now, we properly handle both of those cases.

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -1,7 +1,7 @@
 // Routines for managing and rendering graphics data fetched over the WRP.
 
 import { ThreadFront } from "./thread";
-import { assert, binarySearch } from "./utils";
+import { assert, binarySearch, defer, Deferred } from "./utils";
 import { DownloadCancelledError, ScreenshotCache } from "./screenshot-cache";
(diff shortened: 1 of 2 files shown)


src/protocol/graphics.ts
```diff
@@ -19,6 +19,8 @@ import { setCanvas, setEventsForType, setVideoUrl } from "ui/actions/app";
 import { setPlaybackPrecachedTime, setPlaybackStalled } from "ui/actions/timeline";
 import { getPlaybackPrecachedTime, getRecordingDuration } from "ui/reducers/timeline";
 
+const FIVE_SECONDS = 5000;
```
jasonLaster, line 22: lets use a semantic variable like `MIN_VIDEO_DURATION`

at some point we'll change the value and this will be a lie

jasonLaster (approved):

PR #5476, opened 2022-02-25
Expand the console filter drawer by default

### LINKED ISSUE #5427: Console - show the drawer by default
We should show the console drawer by default which will be more discoverable

jaril:
Fix #5427.

image


jaril:
@jonbell-lot23 heads up!

PR #5477, opened 2022-02-25
Add messaging when adding a team member fails

### LINKED ISSUE #5346: Inviting users to an expired team fails silently
GraphQL throws an error but we don't message the user

ryanjduffy:
## Issue

Adding a member to a team can fail and we don't message it at all. This was worse when the backend threw when adding members to expired teams but can still occur.

Fixes #5346

## Resolution

* Catch errors and display a message
* Call `trackEvent` to track failures


jaril (approved):
Looks good!

PR #5485, opened 2022-02-25
Track anonymous user usage in mixpanel

### LINKED ISSUE #4684: Start instrumenting our guest flow UX

jaril:
Fix #4684.

Right now we only start tracking user activity once we confirm that the user is authenticated. This makes it so that we do the same for anonymous users.

https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelidentify

Some notes:
- Mixpanel saves a distinct ID in cookies that it uses to identify a user.
- Without any user info (which is the case for anonymous users), Mixpanel will identify a user using that distinct ID.
- Mixpanel is [smart enough](https://help.mixpanel.com/hc/en-us/articles/360039133851) to notice when a previously-signed in person is logged out by matching the distinct IDs. This means we will continue to get events properly for joe@schmoe.com in case he's using replay but forgot to log in.
- Cookies are cookies so these IDs are very finicky. For example, two separate incognito sessions opening the same replay  resulted in two anonymous users being identified. It's possible that this ends up hammering our Mixpanel headcount so gotta keep an eye on it.

### DIFF
--- a/src/ui/setup/index.ts
+++ b/src/ui/setup/index.ts
@@ -12,7 +12,7 @@ import { setAccessTokenInBrowserPrefs, setUserInBrowserPrefs } from "ui/utils/br
 import { getUserInfo } from "ui/hooks/users";
 import { getUserSettings } from "ui/hooks/settings";
 import { initLaunchDarkly } from "ui/utils/launchdarkly";
-import { maybeSetMixpanelContext } from "ui/utils/mixpanel";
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):

jasonLaster:
One formatting issue

```
src/ui/setup/index.ts:1:1
 1:1  high  Incorrect formatting, autoformat by running 'trunk fmt'  prettier
```

PR #5496, opened 2022-02-26
Show jest tests which ran in the recording as console messages

bhackett1024:
WIP for https://github.com/RecordReplay/devtools/issues/5439

This identifies the jest tests in a recording and adds console messages for each of them, which can be clicked on to jump to the start of the test.  This still needs to deal with test failures.

### DIFF
--- a/src/protocol/find-tests.ts
+++ b/src/protocol/find-tests.ts
@@ -1,0 +1,251 @@
+// Perform some analysis to find and describe automated tests that were recorded.
+
+import { ThreadFront } from "protocol/thread/thread";
+import analysisManager, { AnalysisHandler, AnalysisParams } from "./analysisManager";
+import { Helpers } from "./logpoint";
+import { assert } from "protocol/utils";
+import { client } from "./socket";
+import {
+  AnalysisEntry,
+  ExecutionPoint,
+  Message,
+  Location,
+  PointDescription,
+} from "@recordreplay/protocol";
+
+// Information about a jest test which ran in the recording.
+interface JestTestInfo {
+  // Hierarchical names for this test, from child to parent.
+  names: string[];
+
+  // Start point of the test.
+  startPoint: PointDescription;
+
+  // If the test failed, the place where it failed at.
+  errorPoint?: ExecutionPoint;
+
+  // If the test failed, description of the failure.
+  errorText?: string;
+}
+
+// Mapper returning analysis results for jest test callback invocations
+// which are associated with a test instead of a hook, returning the name
+// of the test and the point where the callback is invoked.
+const JestTestMapper = `
+${Helpers}
+const { point, time, pauseId } = input;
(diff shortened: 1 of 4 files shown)


hbenl (approved):

src/protocol/find-tests.ts
```diff
@@ -0,0 +1,445 @@
...
+    }
+  }
+  return null;
+}
+
+// Look for places in the recording used to run jest tests.
+async function setupJestTests(): Promise<JestTestState | null> {
+  // Look for a source containing the callAsyncCircusFn function which is used to
+  // run tests using recent versions of Jest.
+  const circusUtilsSourceId = await findMatchingSourceId("jest-circus/build/utils.js");
```
SimenB, line 312: note that we might start bundling (https://github.com/facebook/jest/pull/12348) at which point at least the filename will break and _possibly_ the function name (depending on sourcemaps).

Not sure how to handle that. It would be better to have a real hook for this (e.g. [`handleTestEvent` in the env](https://jestjs.io/docs/configuration#testenvironment-string)?), but that might not be feasible - I have no idea how this project works.

bhackett1024: Definitely, the pattern matching on file names and code patterns in this PR is not robust at all, and longer term it would be great to have a way of identifying places where tests started and failed that doesn't rely on detecting these patterns.  We have a mechanism to add annotations to a recording which could be used to find these points.  It isn't currently exposed in node, but if it was we could have something like:

```
if (process.addRecordReplayAnnotations) {
  process.addRecordReplayAnnotation("JestTestStart", { ...info about test... });
}
```

I don't know if it would be possible to use `handleTestEvent` for this --- we would need these annotations to be as close as possible to where the test starts running, so we can find the entry point to the test function which was passed in.  Right now we pattern match to detect the calls to the test function, and then step in to each of those calls to find the point to jump to when someone clicks on that test.

SimenB: We have [`--testLocationInResults`](https://jestjs.io/docs/cli#--testlocationinresults). Would that be enough? I would assume that info is also included in a `handleTestEvent` call, but I haven't verified. In that case you can at the very least point to where the test function is

bhackett1024: I think that would be enough.  If the test event callback is notified about every test, and has the name of that test, the file/line/column of the test callback, and whether the test passed, that should be enough information to reconstruct everything we're getting from the pattern matching we're doing.  The test callback would call process.addRecordReplayAnnotation with this information, and then the devtools can read that from the recording and use it to figure out where the tests ran.

SimenB: Yeah, its called when test is declared, execution starts, execution enda etc. I'm not 100% sure the location info is present for all cases, but if not we can probably fix that

PR #5501, opened 2022-02-27
Added Url and Path columns to NetMonitor

jasonLaster:
This adds a couple of additional columns 

fixes https://github.com/RecordReplay/devtools/issues/5452


abstractalgo (approved):
lgtm 👌 

Not a fan of single-word function names, but is ok 😄

PR #5507, opened 2022-02-27
Standardising tab styles

jonbell-lot23:
Here's a loom explaining: https://www.loom.com/share/6d5f0e6911024b2ea3f16ab483f1b496

We were building three different tabs in three different ways. This standardising all three on a single set of design tokens so they all work in the same way. The components are still different, but at least the UX is the same.


jasonLaster:
Looks great!

PR #5531, opened 2022-02-28
small change to correct shortcut for function search in quick open modal

### LINKED ISSUE #5530: "Search for a function" shortcut in quick open modal incorrect
**Description:**
Currently says `Cmd+Shift+P` but should be `Cmd+Shift+O`, PR to follow.

**Steps to Reproduce:**

**Expected Result:**

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:

ceceliacreates:
Just a small text change, the shortcut said `Cmd+Shift+P` but should be `Cmd+Shift+O`. Resolves #5530

### DIFF
--- a/src/ui/components/CommandPalette/CommandPalette.tsx
+++ b/src/ui/components/CommandPalette/CommandPalette.tsx
@@ -47,7 +47,7 @@ const COMMANDS: readonly Command[] = [
   { key: "open_viewer", label: "Open Viewer" },
   { key: "open_file_search", label: "Search for file", shortcut: "CmdOrCtrl+P" },
   { key: "open_full_text_search", label: "Search full text", shortcut: "CmdOrCtrl+Shift+F" },
-  { key: "open_function_search", label: "Search for a function", shortcut: "CmdOrCtrl+Shift+P" },
+  { key: "open_function_search", label: "Search for a function", shortcut: "CmdOrCtrl+Shift+O" },
   { key: "open_outline", label: "Open Outline" },
   { key: "open_print_statements", label: "Open Print Statements" },
   { key: "open_react_devtools", label: "Open React DevTools", settingKey: "showReact" },


jasonLaster (approved):

jasonLaster (commented):
Actually - global function search should be `CmdOrCtrl+O`. `CmdCtrl+Shift+O` is for searching functions within a file. 

Maybe we want both here?

ceceliacreates:
> Actually - global function search should be `CmdOrCtrl+O`. `CmdCtrl+Shift+O` is for searching functions within a file.
> 
> Maybe we want both here?

I assumed that searching within the current file was the preferred scope and this was just a typo. The shortcut menu (screenshot below) shows `Cmd+Shift+O` for "Find function" and `Cmd+O` for "Find all functions". As far as what shortcuts to include, that feels like a design/UX decision, but I did want to fix this because currently it's incorrect.

Screen Shot 2022-02-28 at 11 35 31 AM

FYI, I did also update our docs to include keyboard shortcuts here: https://docs.replay.io/docs/shortcuts-e42e370b166d45738df25f537e16b230

jaril (approved):
🎉

jaril:
@ceceliacreates totally brushed over the discussion on this yesterday. Yes we have global function search and it's intentionally on Cmd+Shift+O. 

We don't have current file function search yet. My preference would just be to leave it as it is for now (Global function search -> Cmd+Shift+O) and then revisit when we decide to add current file function search. That will require spec-ing out what we want file function search to look like (especially since we have a search box in the outline view!), and since it's not super straightforward I'd rather punt on the decision til then.

tldr I'd prefer we leave it at Cmd+Shift+O, so correct the command palette so `Search for a function` -> `Cmd+Shift+O`, and update the docs.

Objections? @jonbell-lot23 @jasonLaster

jonbell-lot23:
Agree with punting and taking this on as a broader design exploration later. (I do have concerns about how our current search opens in outline view, plus our inconsistent use of input boxes overall)

jasonLaster:
Actually cmd+shift+o is per file and cmd+o is global

Here is a replay that shows the behavior
https://app.replay.io/recording/function-search--15ccb4b6-bb0a-4500-a843-ed782812bbfd 

they both open the quick open modal

jaril:
> Actually cmd+shift+o is per file and cmd+o is global
> 
> Here is a replay that shows the behavior https://app.replay.io/recording/function-search--15ccb4b6-bb0a-4500-a843-ed782812bbfd
> 
> they both open the quick open modal
> 
> cc @ceceliacreates @jonbell-lot23 @jaril

Yup got that wrong, follow Jason's lead! 👆

I think we're good to merge this and the docs are correct as well :)

PR #5537, opened 2022-02-28
Timeline fix

jonbell-lot23:
We had to revert because I wasn't handling buffering/loading states correctly. Now it's fixed.
Loom: https://www.loom.com/share/94f3924a5e0949119e9c2cb5e2070402

### DIFF
--- a/src/devtools/client/themes/variables.css
+++ b/src/devtools/client/themes/variables.css
@@ -180,7 +180,8 @@
   --theme-text-field: var(--theme-base-95);
   --theme-console-input: var(--theme-body-background);
 
-  --progressbar-preview: var(--theme-base-85);
+  --progressbar-preview-min: var(--theme-base-80);
+  --progressbar-preview-max: var(--theme-base-85);
   --progressbar-unplayed: var(--theme-base-100);
 
   --unloaded-region-img: url("/images/dotted-mask-light.svg");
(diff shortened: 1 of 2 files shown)


src/devtools/client/themes/variables.css
```diff
@@ -366,7 +367,8 @@
   --theme-text-field: var(--theme-base-95);
   --theme-console-input: var(--theme-text-field);
 
-  --progressbar-preview: var(--theme-base-70);
+  --progressbar-preview-min: var(--theme-base-80);
+  --progressbar-preview-max: var(--theme-base-70);
```
jcmorrow, line 371: What do `min` and `max` refer to here?

jonbell-lot23: When you're moving the mouse cursor over the timeline, we have a range of colors:

* The background of the whole timeline
* The blue bar showing where you've gotten to
* One shade of grey for showing where the buffer has gotten to
* Another shade of grey showing a hover state

To map to these, the timeline has always used these CSS classes to show the buffer state's min and max value, based on where you've hovering:

```
.timeline .progress-line.preview-min
```

```
.timeline .progress-line.preview-max
```

So these variables map to them.

jcmorrow: Ah, ok. I guess I meant more of a `min` of what and `max` of what sense. It would be nice for them to have some meaning, but it sounds like that might not be a term that you've introduced, just something that is elsewhere. I still think it might be nice to have a better name for it, but it doesn't need to block fixing a bug 👍

jonbell-lot23: Makes sense, thanks!

PR #5547, opened 2022-03-01
Coloring the icons in console

jonbell-lot23:
Loom: https://www.loom.com/share/4cb71c2f29b6474dac5845e6754e0ebd

Our icons used to look too dark, like this:
image

Now they'll look like this:
image

### DIFF
--- a/src/image/image.css
+++ b/src/image/image.css
@@ -94,22 +94,28 @@
   background-image: url(~devtools/client/themes/images/dropmarker.svg) !important;
 }
 .message.command > .icon {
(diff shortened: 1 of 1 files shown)


jcmorrow (approved):
If most icons are going to be that color it might be nice to declare this once for all `.icon`, but maybe that makes less sense than it seems just from this PR 🤷

jonbell-lot23:
> If most icons are going to be that color it might be nice to declare this once for all `.icon`, but maybe that makes less sense than it seems just from this PR 🤷

Yeah, we use "icons" in a bunch of places and contexts so it's a bit hairy right now. We'll be splitting it out and making it more defined later on.

PR #5548, opened 2022-03-01
Fix console loading state

jaril:
The logic was reversed in #4902. This fixes it.


jcmorrow (approved):
I'm confused about how this wasn't completely broken before but this works 👍

jaril:
> I'm confused about how this wasn't completely broken before but this works 👍

somehow, it was failing the right way 🤣

PR #5550, opened 2022-03-01
Convert eager-eval utils to actions

jasonLaster:
wip to be able to fetch the selected frame from the store

PR #5552, opened 2022-03-01
Fetch the selected frame from the store for eager eval purposes

jaril:
Re-do of #5550.

@jasonLaster the previous patch doesn't really work well as an action. This wraps the util in a hook so that we can (in effect) do the same thing, where we fetch the selected frame from the store and use that value to perform the evaluation.

Demo: https://share.descript.com/view/abrz2fWpGUQ

### DIFF
--- a/src/devtools/client/webconsole/utils/autocomplete-eager.ts
+++ b/src/devtools/client/webconsole/utils/autocomplete-eager.ts
@@ -1,5 +1,6 @@
 import { getSelectedFrame } from "devtools/client/debugger/src/selectors";
 import { ThreadFront, ValueFront } from "protocol/thread";
+import { useCallback } from "react";
 import { useSelector } from "react-redux";
 import { getPropertiesForObject } from "./autocomplete";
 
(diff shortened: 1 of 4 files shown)


jasonLaster (commented):
Code looks solid. Will defer to @abstractalgo on testing, ts, etc...

src/devtools/client/webconsole/utils/autocomplete-eager.ts
```diff
@@ -49,3 +55,24 @@ export async function eagerEvaluateExpression(expression: string): Promise<Value
...
+}
+export function useEagerEvaluateExpression() {
+  const frame = useSelector(getSelectedFrame);
+
+  if (!frame) {
+    return () => null;
+  }
+
+  return async (expression: string) =>
+    await eagerEvaluateExpression(expression, frame.asyncIndex, frame.protocolId);
```
jcmorrow, line 77: Is this the same as leaving out the `async/await` entirely?

src/devtools/client/webconsole/utils/autocomplete-eager.ts
```diff
@@ -49,3 +55,24 @@ export async function eagerEvaluateExpression(expression: string): Promise<Value
...
+
+export function useGetEvaluatedProperties() {
+  const frame = useSelector(getSelectedFrame);
+
+  if (!frame) {
+    return () => null;
+  }
+
+  return async (expression: string) =>
+    await getEvaluatedProperties(expression, frame.asyncIndex, frame.protocolId);
```
jcmorrow, line 67: Same question as below.

jcmorrow (approved):
One small question but otherwise LGTM! 👍

src/devtools/client/webconsole/utils/autocomplete-eager.ts
jaril: Good call, think so. Let me get that in.

PR #5553, opened 2022-03-01
Standardising our icons further

jonbell-lot23:
A few days ago we noticed [this bug](https://github.com/RecordReplay/devtools/issues/5500) that showed the network monitor menu with ugly black boxes. This was caused because we needed to move away from Material icons in order to make sure hover states and dark mode can be supported, and I missed a few dropdowns when I moved us over.

The short term fix was to go back to Material icons, but that caused bugs elsewhere. The Full Credit Answer is to move the Network Monitor icons to the new and improved approach, and that's what this PR addresses.

Loom walking through the changes: https://www.loom.com/share/00243e8dc060474796bba4f6caaf8d34

NOTE: I'd like Jaril to take a careful look at this to make sure we're not using the dropdown component somewhere else that I didn't discover.


abstractalgo:
Are we moving away from or towards Material icons?
What's the ideal outcome for icons implementation(s)?

@jonbell-lot23

jonbell-lot23:
> Are we moving away from or towards Material icons?
> 
> What's the ideal outcome for icons implementation(s)?
> 
> 
> 
> @jonbell-lot23 

We're moving away from Material icons because they're not giving us hover states. They're also aesthetically inconsistent and lower quality.

jaril:
> > Are we moving away from or towards Material icons?
> > What's the ideal outcome for icons implementation(s)?
> > @jonbell-lot23
> 
> We're moving away from Material icons because they're not giving us hover states. They're also aesthetically inconsistent and lower quality.

Just in case it's relevant info, we _could_ make it so that material icons work well with hover states.

My impression was that material icons were too rigid for the look that we wanted to achieve the look for the UI, and so we're creating our own icon set to have more control over it.

jonbell-lot23:
Yup. 

Probably the best long term solution is for both Material and not-Material to work with hover states. Then it's just a matter of aesthetic taste.

jaril (approved):

PR #5558, opened 2022-03-01
Start tracking guest flows in Mixpanel

jaril:
Picking back up from #5485.

The last PR was reverted because there were too many anonymous users (~1k in ~6 hours). 

My current hunch is that those users are from QA Wolf tests where we're testing some anonymous user UX behavior. Since the anonymous user tracking is cookie-based, it's reasonable to assume that each test instance would count as one, and that contributed to the high numbers.

This delays tracking the user until after the login screen, which should alleviate it somewhat. But ideally we should be able to start identifying the user in `setup/index`. 

Checking this with Chris now to see what we can do about it.

### DIFF
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -10,11 +10,12 @@ import NonDevView from "./Views/NonDevView";
 import WaitForReduxSlice from "./WaitForReduxSlice";
 import ReplayLogo from "./shared/ReplayLogo";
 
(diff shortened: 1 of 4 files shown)


jasonLaster (commented):
Curious what @ryanjduffy thinks.

Also it would be nice to explicitly check to see if the guest user is viewing a recording created from an external user.

jaril:
Continued in #5558.

PR #5563, opened 2022-03-01
Run e2e tests on pushes to master

bhackett1024:
We should run our test suite on all pushes to master.  This would make it easier to understand what is going on with https://github.com/RecordReplay/devtools/issues/5562

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,10 @@
 name: Run test suites
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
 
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):

jcmorrow (approved):

ryanjduffy:
I think we'll have to rework this. The `wait-for-vercel-preview` action doesn't work with `master` merges because it looks for a `pull_request` from GH to find the deployment. I think it could be reworked to support either use case but it doesn't atm.

I'm going to revert this since it doesn't work and can't without other substantive changes to the workflow.

jcmorrow:
Right. Ideally this would run against prod once the vercel deploy finishes.

PR #5566, opened 2022-03-02
Dark theme breakpoints

jonbell-lot23:
Here's a loom explaining what I did here:
https://www.loom.com/share/337d75afe81446e2b4e4040b34e92934

Basically I moved a ton of stuff to CSS variables so our dark theme looks good when you're setting breakpoints. It was pretty broken before.


jasonLaster (approved):
Looks great!

PR #5568, opened 2022-03-02
Ensure that only one session is created

hbenl:
We call `createSession()` from a react component that is currently rendered twice when the devtools start up.
As a result, we were creating two sessions for the recording and in tests the test script was run twice in parallel (obviously leading to some test failures).


hbenl:
> were you able to dig into why it was being rendered twice?

No, I didn't look into that.

PR #5569, opened 2022-03-02
Adding docs

abstractalgo:
Docs for core concepts, objects and types - an overview of how `devtools` is conceptualized and why, and perhaps a partial overview of the architecture.

- docs should mostly help answer "why" and "how", and not just "what" or layout the features (build intuition on why things are needed and for which part and why they are shaped in these particular ways)
- these dev docs are different from regular user-facing docs in way that's more concerned with implementation details that aren't necessarily visible via UI and a dev should be knowledgeable about while working with core features

Let's populate it with basic info and then improve it over time.

- [ ] concepts
- [ ] the flow of information when the recording is being loaded, played and inspected
- [ ] how are different panels' information derived from the basic info provided by the protocol
- [ ] an example (a demo recording of some simple app and a walkthrough of how objects from the execution get represented)


docs/concepts.md
```diff
@@ -0,0 +1,55 @@
+concepts:
+
+```
+grip
+reps
+front types
```
jasonLaster, line 6: It would be nice for the `src/protocol` docs to be separate because that directory will be a standalone package that other projects will import. It is also helpful to differentiate between devtools, src/protocol, and https://www.npmjs.com/package/@recordreplay/protocol

jasonLaster: https://replay.io/protocol

PR #5583, opened 2022-03-03
Only show the focus modal once

jasonLaster:


jonbell-lot23:
@jasonLaster I disagree with this because without our single line of text it's just a big dimmed UI and it looks awkward.

Previously we had too much text, but now that the text is so brief, I think it's better to show simple helper text each time.

(True story: Microsoft learned that if you're not using a feature nearly every day, you forget how to use it every time. This feature is novel enough that we lose more than we gain by trying to remove helper text)

Issue #5270, opened 2022-02-09
Add WHYs to our Asserts


jasonLaster:
It would be great if Sentry asserts had some additional context. this should be as simple as adding an additional label for our asserts

jaril:
@hbenl mind taking a look at this?

PR #5588, opened 2022-03-03
Add failure reasons to asserts (#5270)

hbenl:


abstractalgo (approved):
You weren't kidding by saying that a lot of them are quite generic. 😄

PR #5592, opened 2022-03-03
Prevent selecting recordings by non-owners

ryanjduffy:
## Issue

Non-owners of recordings could select them in bulk edit mode

## Resolution

* Suppress the selection handler for non-owners which was still activated by clicking the row even though the checkbox was not rendered
* Add some styling to make it more clear what can be interactive when bulk editing

### DIFF
--- a/src/ui/components/Library/RecordingRow.tsx
+++ b/src/ui/components/Library/RecordingRow.tsx
@@ -11,6 +11,7 @@ import { connect, ConnectedProps } from "react-redux";
 import { actions } from "ui/actions";
 import { getDisplayedUrl } from "ui/utils/environment";
 import { getRecordingURL } from "ui/utils/recording";
+import classNames from "classnames";
 
 export function getDurationString(durationMs: number) {
   const seconds = Math.round(durationMs / 1000);
@@ -77,10 +78,12 @@ function RecordingRow({
   const isOwner = userId == recording.user?.id;
 
   const toggleChecked = () => {
-    if (selected) {
-      removeSelectedId(recording.id);
-    } else {
-      addSelectedId(recording.id);
+    if (isOwner) {
+      if (selected) {
+        removeSelectedId(recording.id);
+      } else {
+        addSelectedId(recording.id);
+      }
     }
   };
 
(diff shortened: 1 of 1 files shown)


ryanjduffy:
@jonbell-lot23 - Proposing some visual changes here for recording rows that can't be selected. In addition to the reduced opacity (60%), I've suppressed the `cursor-pointer` which is misleading since the row can be selected.

image

ryanjduffy:
@jasonLaster - I agree in general though moving a replay could be problematic if the owner of the replay doesn't have access to the destination. Here's the position I shared in slack. If there's debate on how to proceed, I'd like to land this fix so things aren't broken and then "refix" shortly after.

image

jonbell-lot23:
The primary action is loading a replay, and I am concerned that opacity 60% communicates that the replay itself is invalid or not playable.

Just like how a file on a desktop or a Unix filesystem looks the same regardless of author, I am not convinced we need such a blunt tool to communicate ownership.

I'd tuck this into the overflow menu instead, and dim things that aren't accessible. We could also modify the overflow menu to say why things are dimmed but a) that's pretty aggressive and non standard b) I am not convinced of the use case that warrants that much attention on this issue

ryanjduffy:
@jonbell-lot23 - I should have clarified that the opacity only applied when bulk editing was active in which case those options are not interactive.

That said, @jasonLaster and I discussed this last night and decided to adjust the business rules on this to allow any team member to move or delete recordings from a workspace. The only open item is how to handle the option to move to "my library" for recordings that you did not create. My Library is currently implemented as "recordings I own that aren't in a workspace" so the notion of moving someone else's recording to My Library doesn't work as it would require changing the owner.

My inclination right now is to handle this as you suggested above: The "My Library" option under "Move" would be disabled and visually dimmed if the user selected recordings that they do not own.

There are some other edge cases (e.g. User A moving a recording User B owns to a workspace to which User B does not have access) that may cause us to change our minds later on this.

ryanjduffy:
I think we should roll this fix into #5609 since they are effectively solving the same problem from different perspectives.

jonbell-lot23:
Thanks for documenting the thinking here!

PR #5253, opened 2022-02-08
Upgrade to Tailwind 3

abstractalgo:
https://tailwindcss.com/docs/upgrade-guide

PR #5593, opened 2022-03-03
Fix edit button styles

ryanjduffy:
## Issue

* The edit button is white on white
* The down arrow on the selection dropdown is too tight to the text and misaligned

image

## Resolution

* Remove the `bg-white` class from the button. Interestingly, this class has been there a long time but was overridden by `bg-primaryAccent` on the button until recently when the CSS was rendered in a different order changing the precedence.
* Undoes a bit of #5336 to bring back the `inline-flex` styling on buttons so we can align the contents more easily.


ryanjduffy:
Button color was a regression from #5253

jaril (approved):

PR #5594, opened 2022-03-03
Put a loader in place of the console terminal when paused on an unloaded region

### LINKED ISSUE #5211: Evaluations in unloaded regions disappear
The console permits evaluations while in an unloaded region, but after pressing enter the evaluation and its result are not added to the console.

https://app.replay.io/recording/disappearing-console-eval--156b3ec9-3430-483b-a481-117a514e3df8?point=62632080870259492550044119640049259&time=23310&hasFrames=false

jaril:
Fix #5211.

Console evaluations don't work while paused in an unloaded region, which means that the whole evaluation gets dropped and the experience just feels broken.

This keeps us from being in a state where the user is able to evaluate something in the console while paused in a problematic pause point in the first place.

image

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -435,7 +435,8 @@ html #webconsole-notificationbox {
 }
 
 /* CodeMirror-powered JsTerm */
-.jsterm-input-container .CodeMirror {
+.jsterm-input-container .CodeMirror,
+.jsterm-input-container .load-container {
   /* aim for a 32px left space (a descendent has 4px padding) */
   padding-inline-start: calc(var(--console-inline-start-gutter) - 4px);
   /* Create a new stacking context */
--- a/src/devtools/client/webconsole/components/App.js
+++ b/src/devtools/client/webconsole/components/App.js
@@ -57,7 +57,7 @@ class App extends React.Component {
       return;
     }
 
-    window.jsterm.editor.focus();
+    window.jsterm?.editor.focus();
   };
 
   render() {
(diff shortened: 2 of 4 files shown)


jaril:
This seems to have angered the test gods. Working on it.

jasonLaster:
@jaril the spinner feels like a bit of a misuse here.

@jonbell-lot23 curious what you think would be a good treatment. Probably similar to what we do when the user is not a developer role and does not have permission to evaluate

jaril:
I just put it in for now as a placeholder so Jon and I can talk through it later during sync. 

There's two cases to think about:
1) Temporary unloaded regions: over time this region will become loaded.
2) Permanently unloaded regions: the backend has yanked it away from you without your input and will stay unloaded.

We've agreed that we should just kneecap console evaluations in unloaded regions, almost too harshly, so that we know what it feels like in practice and we can pull it back to the appropriate amount. The boring version of it is literally just having nothing, so I slapped a spinner just so I know which one's which.

> Probably similar to what we do when the user is not a developer role and does not have permission to evaluate

Forgot that we had that. Might be consistent to copy the treatment there. Let me know if you know where that visual is/how to force it, in case you get to it faster than I do.

jaril:
So we do still have it! 

image

jonbell-lot23:
Hey gang! Talking to Jaril right now and the best pattern for this is:

* Some text
* Greyed out
* Italicised (probably)
* That says "Loading region..."

jaril:
image

jasonLaster:
This looks good. My question is whether Loading... is honest if you first have to focus on the region to load it.

jonbell-lot23:
> This looks good. My question is whether Loading... is honest if you first have to focus on the region to load it.

Great point, I may have misunderstood this detail.

I thought this was temporary and would often be swapped out successful as the region loads.

If not, I totally support changing the text to clarify that it's an unloaded region.

jasonLaster:
Talked to jaril and it is temporary so we're good

PR #5607, opened 2022-03-04
The dangers of using Objects as Maps

hbenl:
... especially when writing a javascript debugger:
While working on support for getters in the ObjectInspector I saw this in a simple example recording:
Screenshot_2022-03-04_15-06-40
These getters (except `foobar`) look like the properties of a `ValueFront`, but this was not a recording of Replay and the example didn't contain these getters anywhere, so what happened?
I was collecting the getters in an Object (code slightly simplified):
```
const rv = {};
for (const { name, get } of this._object.preview.properties) {
  if (get) {
    rv[name] = get;
  }
}
return rv;
```
But the object whose getters I was collecting contained a getter named `__proto__`, so I was effectively doing `rv.__proto__ = get` where `get` is of type `ValueFront`.
Then the ObjectInspector iterated over `rv` using `for ... in` and ... :boom:

This PR changes `ValueFront.previewValueMap()` and `ValueFront.previewGetters()` to return a Map instead of an Object.

### DIFF
--- a/src/devtools/packages/devtools-reps/object-inspector/utils/value.tsx
+++ b/src/devtools/packages/devtools-reps/object-inspector/utils/value.tsx
@@ -124,13 +124,13 @@ export class ValueItem implements IItem {
 
     const value = this.contents;
     const previewValues = value.previewValueMap();
-    const rv: Item[] = Object.entries(previewValues).map(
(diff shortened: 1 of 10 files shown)


abstractalgo (approved):
lgtm, good gatch :)

PR #5609, opened 2022-03-04
Allow any team member to see the dropdown for a node recording

### LINKED ISSUE #5332: Recording Item menu in Library is sometimes missing
Reports that some node recordings in the team do not have a menu to delete the recording.

jaril:
Fix #5332.

This makes it so that:
- every team member can see rename/delete/toggle privacy options for replays in the team.
- in the recording dropdown, users will only see the **Move to: Your Library** option if they're the owner of the recording.

### DIFF
--- a/src/ui/components/Library/RecordingOptionsDropdown.tsx
+++ b/src/ui/components/Library/RecordingOptionsDropdown.tsx
@@ -11,46 +11,9 @@ import classNames from "classnames";
 import MoveRecordingMenu from "./MoveRecordingMenu";
 import { useConfirm } from "../shared/Confirm";
 import { useIsPublicEnabled } from "ui/utils/org";
-import { useGetNonPendingWorkspaces } from "ui/hooks/workspaces";
 import { getWorkspaceId } from "ui/reducers/app";
 import { setModal } from "ui/actions/app";
-
-export function useGetPermissions(recording: Recording) {
-  const { userId, loading: userIdLoading } = hooks.useGetUserId();
-  const { workspaces, loading: workspacesLoading } = useGetNonPendingWorkspaces();
-  const isOwner = userId == recording.user?.id;
-
-  if (userIdLoading || workspacesLoading) {
-    return {
-      loading: true,
-      permissions: {
-        move: false,
-        moveToLibrary: false,
-        privacy: false,
-        rename: false,
-        delete: false,
-      },
-    };
-  }
-
-  const sameTeam =
-    recording.workspace?.id && workspaces.find(w => w.id === recording.workspace?.id);
-  // Here we are inferring that recordings without a user are node recordings, since they've
-  // been update using the workspace's API key.
-  const isNodeRecording = !recording.user;
-  const isPrivilegedUser = isOwner || (isNodeRecording && sameTeam);
-
-  return {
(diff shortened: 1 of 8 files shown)


src/ui/components/Library/RecordingOptionsDropdown.tsx
```diff
@@ -8,28 +8,47 @@ import MaterialIcon from "../shared/MaterialIcon";
...
+
+  const isOwner = userId == recording.user?.id;
+  // Node recordings don't have an owner since they're uploaded using the workspace's
+  // API key. We add this check here so the team is able to access the dropdown.
+  const sameTeam =
+    recording.workspace?.id && workspaces.find(w => w.id === recording.workspace?.id);
+  const showDropdown = isOwner || (!recording.user && sameTeam);
+
+  return { showDropdown, loading };
+}
```
abstractalgo, line 39: This is actually checking belonging/ownership. The name of the hook can be something like `useOwnershipInfo` or so.

The fact that this info is being used to show/hide a dropdown is less relevant to the actual job of the hook and it would be good to decouple those two.

src/ui/components/Library/RecordingOptionsDropdown.tsx
```diff
@@ -8,28 +8,47 @@ import MaterialIcon from "../shared/MaterialIcon";
...
   currentWorkspaceId,
   recording,
   setModal,
 }: RecordingOptionsDropdownProps) {
   const [isPrivate, setIsPrivate] = useState(recording.private);
   const [expanded, setExpanded] = useState(false);
+  const { showDropdown, loading: loading1 } = useGetShowDropdown(recording);
   const deleteRecording = hooks.useDeleteRecordingFromLibrary();
-  const { workspaces, loading } = hooks.useGetNonPendingWorkspaces();
+  const { workspaces, loading: loading2 } = hooks.useGetNonPendingWorkspaces();
```
abstractalgo, line 50: `loading1` can then become `loadingOwnershipInfo` or so, and `loading2` `loadingWorkspaces`.

I'd strongly recommend against using numbers, as they carry very little information and usually just make things more confusing.

ryanjduffy:
Can we roll the fix for https://github.com/RecordReplay/devtools/pull/5592 into this as well?

jaril:
> Can we roll the fix for #5592 into this as well?

Should be an easy fix. Gimme a sec.

ryanjduffy (approved):

jaril:
We need to change this test because of this PR:
- [ ] There should be dropdowns for every single replay
- [ ] Since the user is in the same team as the replay, they should see the delete/rename/move options in the dropdown menu
- [ ] If they are the owner, they should see a Move To: My Library option in the dropdown
- [ ] If they are not the owner, there should be no Move To: My Library option in the dropdown
- [ ] Recordings in a team that the user belongs to should be selectable during bulk edit
- [ ] In bulk edit, If they are the owner of all replays selected, they should see a Move To: My Library option in the dropdown
 - [ ] In bulk edit, If they are not the owner of any of the selected replays, they should NOT see a Move To: My Library option in the dropdown

https://www.qawolf.com/run/cl114mbe314498651rshlu7jx7z

PR #5614, opened 2022-03-04
More flexible layout options

jasonLaster:
Today, when we hide the editor, it is no longer available. This breaks the links that we have in Comments, Elements, Net Monitor, Console, etc ...

What we should do instead is show the Editor alongside the other panels. 
* We'll need to update the language in the layout dropdown to reflect this. 
* We'll also need to update the UI of the Source Tabs to fit better in this context
* We might also want to offer a mode where the panels are below the video 

Screen Shot 2022-03-04 at 8 01 12 AM


src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -68,6 +73,15 @@ const PanelButtons: FC<PanelButtonsProps> = ({
           <div className="label">Elements</div>
         </button>
       )}
+      <button
+        className={classnames("console-panel-button", {
+          expanded: selectedPanel === "debugger",
+        })}
+        onClick={() => onClick("debugger")}
+      >
+        <div className="label">Sources</div>
```
abstractalgo, line 82: Considering that the actual "Sources" panel can still be visible, and that the item in the dropdown to hide it says "Hide Editor", we can name this "Editor". Seems more consistent.

src/devtools/client/debugger/src/components/Editor/EditorPane.js
```diff
@@ -0,0 +1,48 @@
...
+const Pane = ({ children }) => {
+  const [paneNode, setPanelNode] = useState(null);
+  const nodeWidth = useWidthObserver(paneNode);
+
+  return (
+    <div
+      className={classNames("editor-pane overflow-hidden rounded-lg", {
+        narrow: nodeWidth && nodeWidth < 240,
+      })}
+      ref={node => setPanelNode(node)}
```
abstractalgo, line 23: A place for a ref, not a state (`useRef`).

jasonLaster: good call. Updated

src/ui/components/SecondaryToolbox/index.tsx
jasonLaster: Yep the naming could certainly be improved. Sources is useful though because that's what chrome does

jasonLaster:
cc @jonbell-lot23 would love advice on treatment for the toolbox options dropdown

https://www.loom.com/share/c5b5b2ebac00405ea5505edb99c81158

cc @abstractalgo it would be nice to have your help stress testing this and fixing the rough edges with CodeMirror and perhaps our types.

jonbell-lot23 (changes requested):
This is foundational UI, and requires deeper conversation and consideration than we've allowed in this PR. Strongly recommend holding off and letting this go through a real product design process.

* We all agree the UI needs flexibility
* We all agree this is something to work on soon

But as I mentioned on Friday, we have alignment on insights but not implementation. We should find that first, because this is too important to treat as a small tweak to the UI.

jasonLaster:
Happy to discuss. 

I wouldn't worry too much about this implementation. The work here was primarily on the engineering side to be prepared.

jasonLaster:
Here are some screens.

Screen Shot 2022-03-07 at 1 29 47 PM
Screen Shot 2022-03-07 at 1 29 03 PM


We also discussed adding contextual actions in the viewer
We also discussed adding more actions to the command palette

jonbell-lot23:
I have an idea that I'm pretty excited about.

1. It's a no-brainer to remove the "hide video" menu item and instead just hide the video when you drag it too small. That's ten times better than making someone search for the right menu item, because it's contextual to the user's action.
2. Then what if we used the same pattern for the editor? Drag the editor window too small and boom, we go to single column view.

I think these two changes are foundational things that need to exist no matter what. But it doesn't mean they exist in a vacuum. We can still add actual UI to make it more discoverable. So:

1. We add "Bring back the video" and "Bring back two column view" UI. Not sure exactly how it would look yet, but there should be a button somewhere. And probably not tucked away under an icon, but more prominent.

With these changes in place, we have UI clearly visible, but also some "of course" design where panels snap into place as expected.

jasonLaster:
Love these suggestions.  These are things we can layer on top once we have the core logic in place. The first step is to just support these modes.

jonbell-lot23:
> Love these suggestions. These are things we can layer on top once we have the core logic in place. The first step is to just support these modes.

This kind of prototype has a clever habit of becoming the actual UI, and I'm trying to avoid that. I think this dropdown would be a mistake to add into the UI, especially if it stays that way for a while.

So I'd recommend either pushing the backend with no user-facing UI yet, or waiting until we have user-facing UI and pushing it out together. I really don't want this dropdown to exist in production, even temporarily and behind a flag.

jasonLaster:
I don't think there's a conflict here. 

We can get the logic in and keep the UI changes to a minimal. Hide Editor just adds the Editor panel to the secondary toolbox (fixes the bug)

Then in terms of UX. We need an easy way to solve for 

1. toggling the viewer (close)
2. toggling between split/unified (close)
3. flipping orientation devtools/viewer (horizontal/vertical)

Of course all of this can be done in the palette, but we should also expose this in the UI. The dropdown is the easy option and perhaps more generalizable (UIs do grow in complexity), but there are other ways as well with in-context options.

jonbell-lot23:
Yup, these screens are what I am saying are not ready to go live:

image

image

But they will be, and it won't take long. But we're not there yet.

jonbell-lot23:
Ok here's where I think we are. We're pretty close to being able to go live.

**Tweaking the dropdown text**
image

Let's go with this, each with an icon on the side:
* Dock to bottom right
* Dock to bottom
* Dock to left
* Hide/show video (which is what is on prod)

**Tweaking the dropdown image state itself**
We made a big breakthrough with the UX by deciding to swap the image out to represent the current state. So instead of a generic icon, we'll be showing one of three icons: dock to bottom right, dock to bottom, or dock to left. Here's an image:

image

**Our new two-tier tab system**
This is how it currently looks: 
image

We can improve this iteratively. We don't need to hold off on this to get it right. I'll get in there and fine-tune it.

jaril:
https://user-images.githubusercontent.com/15959269/157989511-4f10c3b7-ad32-4f35-81f4-297bbb164fba.mov

PR #5618, opened 2022-03-04
Add dynamic search throttling based on number of sources

### LINKED ISSUE #5436: Quick Open modal takes a while to show search results
https://app.replay.io/recording/source-search-takes-a-second-to-start--72ae1e14-393c-4fc7-abf8-bc87abc4aa2b?point=29855706938647575316662754272608323&time=11023.053042121684&hasFrames=true

jaril:
Fix #5436.


src/devtools/client/debugger/src/components/QuickOpenModal.js
```diff
@@ -172,7 +177,7 @@ export class QuickOpenModal extends Component {
     }
 
     return this.searchSources(query);
-  }, updateResultsThrottle);
+  }, this.getDebounceMs());
```
jasonLaster, line 180: I'd be surprised if this works if the source count could change.

There should be a way to set updateResults whenever the `sourceCount` updates in a functional components way that would be a useEffect...perhaps here it is in `componentDidUpdate` and `this.updateResults = () => {}` this should not be hard but it is silly

jaril: Yep. Fixed it so that we update the callback with a new debounced function with the right `ms` whenever the `sourceCount` changes.

jasonLaster (approved):

PR #5623, opened 2022-03-06
Dark theme launchprep 1

jonbell-lot23:
Getting down to some nitty gritty now. Here's the Notion page where I'm tracking all the small stuff:
https://www.notion.so/replayio/Dark-Theme-GTM-Plan-dff728b7f58e405a9750fd7fc748488e

Here's a copy/paste from the Notion doc:

**Highest priority**

- [x]  White box in bottom-right corner of sources
- [x]  Sources top right magnifying glass highlights white, should be dark
- [x]  When sources is minimised there’s a white line
- [x]  Elements highlight colours is much too white
- [x]  Headers have big ugly white lines through them
- [x]  Network headers hover state is too dark
- [x]  Network headers has border line on the left
- [x]  Network URL text is too light
- [x]  When breakpoint loads, it has white background ... not looking good

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
@@ -11,7 +11,7 @@
 }
 
 .sources-list {
-  height: 100%;
+  /* height: 100%; */
 }
 
 .sources-list .img.blackBox {
--- a/src/ui/components/Timeline/Timeline.css
+++ b/src/ui/components/Timeline/Timeline.css
@@ -129,7 +129,6 @@
 .breakpoint-navigation-timeline .progress-line {
   width: 0%;
   height: 5px;
-  border-radius: 4px;
   /* border-top: 3px dashed var(--primary-accent); */
   border-top: 5px solid var(--primary-accent);
   position: absolute;
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ next-env.d.ts
 out
 tsconfig.tsbuildinfo
 yarn.lock
+.nova/
--- a/packages/accordion/index.tsx
+++ b/packages/accordion/index.tsx
@@ -113,7 +113,6 @@ export function AccordionPane({
     >
       {isResizable && <ResizeHandle onResizeStart={onResizeStart} isResizing={isBeingResized} />}
       <div className="flex h-full w-full flex-col overflow-hidden">
-        <div className={classNames("border-b", index! > 0 ? "" : "border-transparent")} />
         <div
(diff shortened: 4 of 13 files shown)


src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
```diff
@@ -11,7 +11,7 @@
 }
 
 .sources-list {
-  height: 100%;
+  /* height: 100%; */
```
jasonLaster, line 14: this does not seem dark theme related. Probably should be it's own PR

src/ui/components/Timeline/Timeline.css
```diff
@@ -129,7 +129,6 @@
 .breakpoint-navigation-timeline .progress-line {
   width: 0%;
   height: 5px;
-  border-radius: 4px;
```
jasonLaster, line 132: this also doesn't seem dark theme related. curious what inspired this

jasonLaster (approved):
Overall looks good. Two questions

src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
jonbell-lot23: It is because this line is forcing a visual flaw in dark theme and I couldn't see any regression in light when I removed it

src/ui/components/Timeline/Timeline.css
jonbell-lot23: Rather than rounding nicely on both the top and bottom, it's only rounding on the top. Probably because the timeline isn't tall enough. I didn't notice it in light theme but it was very noticeable in dark.

jasonLaster: Got it - i guess we can fix this up in a follow up. It feels boxy now - but probably better than the alternative

src/devtools/client/debugger/src/components/PrimaryPanes/Sources.css
jasonLaster: Cool - i think @jaril wants to revisit this as well so that the source pane by default is a bit larger

jonbell-lot23: Ah good to know, thanks. I suspected as much, which is why I commented it out. CC'ing @jaril to make sure I'm not being dumb.

src/ui/components/Timeline/Timeline.css
jonbell-lot23: Agree it feels boxy, yeah. Looking forward to fixing this soon.

jasonLaster:
closing this since #2 landed which i assume supercedes

PR #5648, opened 2022-03-07
Completely remove isDemo

abstractalgo:


jcmorrow (approved):
What is that diff noise in `package.json`? Different `npm` version or something?

abstractalgo:
moved to https://github.com/RecordReplay/devtools/pull/5649

PR #5660, opened 2022-03-07
Fix Download Replay onboarding copy

jaril:
Don't show an inviter if there's no inviter in the first place. That happens when a user joins replay using an invite link to a team.


abstractalgo (approved):
tempalteception 😵‍💫 😄

PR #5668, opened 2022-03-08
Manual Vercel deployment

abstractalgo:
This PR changes how the deployment onto Vercel works.

The problem was that Vercel deployments got triggered automatically via Github app for Vercel, but there are some of the GHA jobs that are dependent on the preview build being available. To make sure the order is correct (deployment first and then tests that depend on it), we had a job that would poll for a preview build's availability. However, this behavior and the action we used were quite unreliable and would often time out, and that would cause the tests to fail, rendering most of CI pretty useless.

Instead of relying on automatic deployment, we now actually disable them with `vercel.json`'s `github.enabled` flag, and we use a different Github action ([vercel-action](https://github.com/amondnet/vercel-action#readme)) to manually trigger deployments. Internally, that action uses just a regular [Vercel CLI](https://vercel.com/docs/cli). This makes it possible to not poll for availability, but instead, have a completely natural order of deployment steps and jobs that simply follow one after the other. This should make the entire process much more reliable, and arguably more controllable.

Work:
- [ ] move secrets to `secrets` and please the linters (do a force push to clean up the git history)
- [x] make it work for preview builds of PRs
- [x] make it work for `master`
- [x] add Storybook step (this is just extra)

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -1,0 +1,13 @@
+name: Storybook
+
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  build-storybook:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Deps
+        run: npm install
+      - name: Build Storybook
+        run: npm run build-storybook
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,10 @@
 name: Run test suites
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
 
 jobs:
   download-browser:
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -39,19 +44,25 @@ jobs:
...
+      - uses: amondnet/vercel-action@v20
+        id: vercel-deploy
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          max_timeout: 500
-          check_interval: 30
-      - run: echo ${{steps.waitFor200.outputs.url}}
+          # for posting a comment upon completion
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          vercel-token: "aLNsahI9JClTxsQuxV7Kf5SD"
```
jcmorrow, line 56: Let's make sure we revoke this token if it's useful for anything.

.github/workflows/test.yml
```diff
@@ -39,19 +44,25 @@ jobs:
...
+          vercel-org-id: "team_o3Dz7BlGrXUzHeSUxDXdLfxP"
+          vercel-project-id: "prj_t1WZoKI1cEWvD9FAfXw7JXuspF7y"
```
jcmorrow, line 58: Honestly it would be fine by me to just shove these into secrets too. There's not really any reason to broadcast an ID that could be used to target our services if we don't have to.

.gitignore
```diff
@@ -15,3 +15,4 @@ out
 tsconfig.tsbuildinfo
 yarn.lock
 .nova/
+.vercel
```
jcmorrow, line 18: I'm very surprised we've gotten this far without doing this 🤔

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,13 @@
+name: Storybook
```
jcmorrow, line 1: What will this do? Is the idea to make storybook publicly accessible with each deploy? I think that's a neat idea, but what do you think about making this into a separate change?

jcmorrow (commented):
I've read this over a couple of times. I think I'm good with all of these changes. We should see if once it's rebased the tests will work again (I suspect they are timing out because of the async-storage error, but I could be wrong, I did not look super closely).

.github/workflows/test.yml
abstractalgo: It's my personal token, if we end up using this, we'd probably want to generate some more stable token and shove all of those keys into GH secrets.

.github/workflows/storybook.yml
abstractalgo: Not about public access but a CI step so that we catch when/if we break Storybook asap.

abstractalgo:
This will require some nicer investigation into our GHA steps, checking why our builds and tests take so long (doesn't seem like they should be taking _that_ long, perhaps reporting back raw timing numbers and maybe some other data, so that we can track what's happening, such as reporting bundle size changes, or reporting web metrics (lighthouse scores) for pages (e.g. library and 1-2 test recordings).

I'll close the PR for now as this PR is not solving _the_ problem that it was after and we'll revisit.

PR #5670, opened 2022-03-08
Bundle loading tips icons

abstractalgo:
Moved the loading tips icons from regular SVGs loaded on-demand through `` tag to SVG embedded in the JS bundle. 

This makes icons available as soon as they are needed (i.e. when the tip changes), instead of having a very brief period when the next icon is not loaded yet and there's a state where the current tip's icon is mismatched with the last tip's icon - which basically manifests itself as double the number of transitions (one for text, one for icon) and due to the small gap between those two events it can be registered as a flicker. This change makes the entire exp smoother and looks less buggy and unstable.

---

**To consider:** Due to sudden visual change, it might be wise to make the loading tip box const size and perhaps due a fade-in/fade-out transition between the tips. Then, there would be no abrupt changes, which would give the best experience.


src/ui/components/shared/LoadingTip.tsx
```diff
@@ -1,61 +1,63 @@
...
-];
+] as const;
 
-export default function LoadingTip() {
-  const loadingPageTipIndex = useSelector(selectors.getLoadingPageTipIndex);
-  const dispatch = useDispatch();
-  const { title, description, icon } = TIPS[loadingPageTipIndex || 0];
-  const key = useRef<any>();
+export const LoadingTip: FC = () => {
+  const [tipIdx, setTipIdx] = useState(0);
```
jaril, line 33: Is there some better naming we can use for `tipIdx`?

src/ui/components/shared/LoadingTip.tsx
```diff
@@ -1,61 +1,63 @@
...
     title: "Private by default",
     description: "Replay will never view or analyze your replays without your explicit permission.",
-    icon: "zypsy-security-icon3.svg",
+    icon: ShieldIcon,
   },
-];
+] as const;
 
-export default function LoadingTip() {
-  const loadingPageTipIndex = useSelector(selectors.getLoadingPageTipIndex);
```
jaril, line 33: The reason we stored this in redux is so that the index is maintained even though the loading page is re-rendered from different parent components multiple times. This doesn't preserve that, no?

abstractalgo: If it's rerendered, then the timer gets reset as well, so no global storage for the index would fix it alone without also storing the remaining timer timeout or so.

I'll revisit this to confirm what's going on, but thanks for the heads up.

src/ui/components/shared/LoadingTip.tsx
abstractalgo: `tipGripIdx` 😅 
jk, I'll come up with something.

jonbell-lot23:
Looks good to me. Once you and Jaril are synced, I'm thumbs up

jaril (approved):

Issue #5665, opened 2022-03-08
Dark theme release ticket


jonbell-lot23:
Here's where we're going to track the last dark theme stuff we're looking to get to:

**Assigned to Jaril**

- [x]  Console cursor is black and should be much lighter (caret)
- [x]  Elements has big ugly white line across the bottom
- [x]  “Search HTML” magnifying glass needs to be lighter
- [x]  Network has white box in bottom right corner
- [x]  “Filter by event type” and “Filter Output” have different placeholder colors
- [x] Codemirror highlighting is broken
- [x] Console auto-complete
- [x] Loading screen needs dark theme SVG -- currently too white
- [x] We need to move this checkbox from Experiments to Preferences

**Post-launch**

- [x]  Adding a comment shows a flash of yellow
- [x]  Loading screen flashes white (possibly fixed by #5677)
- [x]  Dropdowns don't have a dark mode
- [x] Network missing its splitter
- [x] Counters missing their squirqle
- [x] Library doesn’t have a dark mode version (https://github.com/RecordReplay/devtools/pull/5720)
- [x] Loading screen flashes white
- [ ] Hats are pink, should be darker red to fit dark theme

abstractalgo:
Can be done completely separately... but at some point we need to dark-mode Library as well (just a small reminder; it is only the first time that I realized that 😅).

jaril:
<img width="238" alt="image" src="https://user-images.githubusercontent.com/15959269/157277891-2b579dcd-08f1-41ab-b84d-758507a4c609.png">

Let's get these nav icons lighter and match the other icon colors, they're subtle.

jaril:
Parts fixed in: https://github.com/RecordReplay/devtools/pull/5676

jaril:
> Elements has big ugly white line across the bottom

Can't reproduce this. @jonbell-lot23 mind seeing if you can still reproduce it?

jonbell-lot23:
Can't reproduce either, @jaril ! I vaguely remember fixing it in a flurry of other stuff, so I think we're good :)

jonbell-lot23:
> <img alt="image" width="238" src="https://user-images.githubusercontent.com/15959269/157277891-2b579dcd-08f1-41ab-b84d-758507a4c609.png">
> 
> Let's get these nav icons lighter and match the other icon colors, they're subtle.

On it

jonbell-lot23:
[...]

PR #5677, opened 2022-03-08
Fix "Loading screen flashes white"

jaril:
Related to #5665.


jaril:
There's a full mark version for this that goes along the lines of https://www.joshwcomeau.com/react/dark-mode/ but for now this ought to do the trick.

PR #5680, opened 2022-03-08
Detect mouse move events outside the timeline which scrubbing the timeline

### LINKED ISSUE #5597: It should be possible to drag the clown nose outside of the footer
we should have the same treatment for dragging the timeline as we have for focus mode.

jaril:
Fix #5597.

This moves us from using a `<div>` mask to detect mousemove events outside of the target area (e.g. timeline) while the user is doing some dragging action. The originally approach worked fine, but ran into problems in cases where you might have to keep track of two masks, e.g. dragging the focus region handlebars. 

Another approach I tried was to move us away from needing two masks in the first place. But it got messy, so in the interest of time I'm taking the L and using this short-term solution for now. We probably should revisit that timeline implementation soon:tm:.

https://user-images.githubusercontent.com/15959269/157914899-ce6a4088-dc0c-4251-a890-423fe42db7d5.mov


jaril:
This totally torpedo-ed focusing. Going back to the drawing board!

jasonLaster (approved):

Issue #5414, opened 2022-02-21
Splitters being too fast


abstractalgo:
You can see in the video that dragging certain splitter handles makes them go too fast, i.e. they don't stay consistent with the actual mouse position, and instead have a multiplier of the diff since the beginning of the drag action.

https://user-images.githubusercontent.com/1355455/154956887-c0d81aa8-bca5-47f7-9f72-eac89c06eb96.mp4

jasonLaster:
Good find. Could you add a replay so we can see what's going on?

abstractalgo:
Hah, can't add a replay - happens in Chrome, but not in Replay browser. 😅

jcmorrow:
Are you zoomed? 😈

abstractalgo:
Nope 😇

PR #5685, opened 2022-03-08
Unify, fix and rewrite splitters

abstractalgo:
We had two almost identical implementations of `SplitBox` component. One of them was buggy (see in #5414). I didn't go into discovering why the other was buggy, I simply chose to swap all references with the correct one, but their interfaces were different and I find it very difficult to work with plain JS so I rewrote to TS and discovered a couple of inconsistencies along the way (sizes sometimes being string, sometimes number, styles that are invalid, wrong CSS etc).


abstractalgo:
I've highlighted all of the places where SplitBox is being used (a total 5 of them):

image
image

- Network panel (x1)
- Element panel (x1)
- main layout with each being a SplitBox as well (x1 + (x1 + x1))

jonbell-lot23 (commented):
I clicked around in dark and light theme and everything seemed good to me. Thanks, this is great investment in standardising our code and ux!

PR #5698, opened 2022-03-09
Variables cleanup

jonbell-lot23:
**What's this?**
Things got pretty messy during my dark mode deep dive, so this is me cleaning things up again as an investment in our future with these variables.

Here's my notepad as I worked through items and tracked the status of each:
image

Take a look at variables.css to see the best summary of what I did.

**Gosh, how much has this been tested?**

I did a fair amount of testing, but there's no rush. We can push this in a few days, and that will give me more time to really make sure I don't have any regressions.


jaril:
This is definitely a nit, but my preference would be to go from `bgcolor` to just `bg` in places that use it, since the additional specificity is redundant. Should be an easy:tm: find and replace job if we agree! @jonbell-lot23

jonbell-lot23:
> This is definitely a nit, but my preference would be to go from `bgcolor` to just `bg` in places that use it, since the additional specificity is redundant. Should be an easy™️ find and replace job if we agree! @jonbell-lot23

Collapsing bgcolor and background into a single property is a common shorthand but it confuses the semantics when I want to use a gradient background. I'd like to stick with the 1:1 mapping with how css works rather than combining them.

color -- foreground color
bgcolor -- background colour
background -- gradients, images, etc

(https://techstacker.com/difference-between-background-and-background-color-in-css/)

jasonLaster:
to continue the nit train, we should try and be consistent with what tailwind does, so `bg` where they use `bg` and yada yada. the good thing is their docs are great https://tailwindcss.com/docs

jonbell-lot23:
> to continue the nit train, we should try and be consistent with what tailwind does, so `bg` where they use `bg` and yada yada. the good thing is their docs are great https://tailwindcss.com/docs

Precisely. 

image

Background doesn't mean "background color", it's a broader category that refers to 9 different properties, of which "color" is just one. Thus we should refer to it as a "background color" as tailwind and css do.

jasonLaster:
Sorry - should have included this in the write up, but i think @jaril 's point is that tailwind uses `bg-<color>-<level>` not `bgColor` or `bg-color` e.g. `bg-blue-500`

https://tailwindcss.com/docs/background-color#setting-the-background-color

jcmorrow (approved):
Visually I just compared dark mode between the preview branch and production and did not find any differences 👌

Issue #4766, opened 2021-12-07
Decouple original and generated sources from a pause point


jaril:
Right now, we can only toggle between a selected source's alternate source when we're paused. It should be possible to do so without being paused.

https://github.com/RecordReplay/devtools/blob/master/src/devtools/client/debugger/src/reducers/pause.js#L368-L370

jaril:
Related to #5345.

PR #5699, opened 2022-03-09
Decouple SourcemapToggle from the current pause (#4766)

### LINKED ISSUE #5345: "No sourcemaps found" shown even if there are sourcemaps
If you look at the source of a bundle while _not_ being paused in that source, Replay shows:
Screenshot_2022-02-17_18-48-54
But this is not necessarily the case: the issue is that there are many original sources associated with a bundle and we don't know which one you're interested in, so we can't select an alternate source.

hbenl:
Currently it is only possible to switch to an alternate source if we're paused in the current source.
With this PR, if we're not paused in the current source we try to find a unique alternate source and allow the user to switch to that.
If there are no alternate sources, the "No sourcemaps found" message is shown.
If there are multiple alternate sources, the SourcemapToggle is hidden.
This also fixes #5345 (and some assertion messages and types in `ThreadFront`).

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -1066,7 +1066,7 @@ class _ThreadFront {
           return { sourceId: id };
         }
         kind = minifiedInfo.kind;
-        assert(kind != "prettyPrinted", "source kind must be prettyPrinted");
+        assert(kind != "prettyPrinted", "source kind must not be prettyPrinted");
       }
       if (kind == "sourceMapped") {
         originalId = id;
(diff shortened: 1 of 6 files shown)


src/protocol/thread/thread.ts
```diff
@@ -1173,7 +1173,7 @@ class _ThreadFront {
     return kind == "sourceMapped";
   }
 
-  preferSource(sourceId: SourceId, value: SourceId) {
```
jaril, line 1176: lol

jaril (approved):

Issue #5696, opened 2022-03-09
ObjectInspector shows "Invalid object"


hbenl:
Reported by @Andarist in Discord: https://discord.com/channels/779097926135054346/840240266569187380/950892152827822110
STR: open this link: https://app.replay.io/recording/composed-conditional-type-expands-into-union-of-its-its-branches--3362459f-d68d-40af-997b-7f407643f642?point=20686870556443395751999&time=2.2467751632049273&hasFrames=true and hover over `array` in the editor.

PR #5700, opened 2022-03-09
Don't assume an array-like object must have a length property (#5696)

hbenl:


abstractalgo (approved):

Andarist:
Out of curiosity - what’s an array-like in this context? I’ve always called array-likes objects that **have** length property and numeric properties. It seems that the first one doesnt have to always be true for those 🤔

hbenl:
It's purely based on classname:
https://github.com/RecordReplay/devtools/blob/3258b2d1099dae85cb5b48c135f8ddda900123de/src/devtools/packages/devtools-reps/reps/array.js#L175-L192
And the title of this PR is slightly inaccurate: it should say "... must have a length *own* property".
In the example there is one array that apparently is an instance of a class derived from `Array`. This instance's prototype (i.e. the derived class) does not have a `length` own property (but the `Array` class interestingly has one, always set to 0).

PR #5702, opened 2022-03-09
Add Storybook to CI on PRs

abstractalgo:
Based on https://github.com/RecordReplay/devtools/pull/5668#discussion_r822208891, this adds Storybook to CI as a job that enables us to catch when we break Storybook asap.

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -1,0 +1,13 @@
+name: Storybook
+
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  build-storybook:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  build-storybook:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Deps
+        run: npm install
+      - name: Build Storybook
+        run: npm run build-storybook
```
jasonLaster, line 13: would love to go one step further at some point and add a jest test for viewing the different stories

PR #5714, opened 2022-03-10
DevTools loading

### LINKED ISSUE #5063: DevTools can take awhile to load
When we talked to Victor, the devtools panel took awhile to load. 

Here's what we saw:

1. The viewer opened quickly
2. The ViewToggle showed up quickly
3. Clicking the DevTools button transitioned quickly
4. The NonDev view seemed stuck and did not transition to the DevView quickly

abstractalgo:
Closes #5063.

- [ ] profile
  - [x] Chrome profiler
  - [ ] Firefox profiler
  - [x] React DevTools profiler
- [ ] compare: base behavior vs lazy vs lazy+preload vs lazy+preload+commit hiddden
- [ ] figure out why layout thrasing happens for some of the panels, fix it (just React deciding to flush changes so far?)
- [ ] lazy load each of the panels separately
  - [x] Console
  - [ ] Editor
    - [ ] can we swap CodeMirror for something more lightweight?
  - [x] ~~Elements?~~ (Elements is pretty lightweight and always loads progressively, so not an issue)
- [ ] see what better strats there are for CodeMirror (https://github.com/RecordReplay/devtools/pull/5789)
  - [ ] e.g. plain text render (uncolored) and lazy render CodeMirror
- [ ] optimize sidepanel rerender (don't re-mount it)

Removing `framer-motion` was actually significant (https://github.com/RecordReplay/devtools/pull/5413).

Nifty resources:
- https://reactjs.org/docs/react-api.html#reactlazy
- https://medium.com/hackernoon/lazy-loading-and-preloading-components-in-react-16-6-804de091c82d
- https://www.youtube.com/watch?v=XgDus4lPmR8
  - https://philipwalton.com/articles/idle-until-urgent/

### Sample capture no. 1

Visible DevTools panels: Console, everything else just default.

image

image

Timings: 
- other panels and components (DebuggerLoader, SecondaryToolbox, ToolboxOptions, ObjectInspectors,...): ~100ms
- Console (JSTerm + recalc style + layout + CodeMirror): ~170ms

**Conclusion:** layout thrashing and CodeMirror (Console's) biggest culprits.

### Sample capture no. 2

Visible DevTools panels: Editor, Console, everything else pretty much just default.

image

Timings:
- other panels: ~140ms
- Console (layout thrashing + CodeMirror): ~300ms
- Editor (CodeMirror + reupdates): ~130ms
- pause: ~60m
- CodeMirror update: ~60ms
- SourcesTree: ~60ms

---

**Finding:** the comments pane gets re-mounted and that stalls things as well. This includes re-fetching comments data and also not-so-trivial TipTap editor rendering. (not just comments, but entire `<SidePanel>`)


abstractalgo:
Thinking about some unified approach where each component can load itself async, render itself async, so that everything in the UI that depends on such components could get unblocked asap.

This can also be very tricky because we may trash the layout a couple of times instead of ~once. But we should be able to minimize the scope of thrashing, because every component never reaches outside its bounds.

Basically, we want the main layout (viewer or devtools), with its general section sizes, to appear close to immediately and then have spinners/loaders for each of the sections until they load themselves. This should also fit well with changes to the current point (i.e. click on a timeline) - some components need to re-fetch themselves (or their data loaded from somewhere above), and some are (more or less) fine between changes to the current point.

Does this kind of behavior point to the larger architectural need to satisfy it?

abstractalgo:
As discussed, we should try out a "two-phase loading" - delay rendering until some further point in time instead of rendering all at once.

markerikson (approved):

PR #5720, opened 2022-03-11
Add dark mode to library

jonbell-lot23:
Here's a Loom:
https://www.loom.com/share/5139ff3b89ef4cd98aeb85f4dfd0e10c

Here's how it looks:
image


jasonLaster:
Don't have a better suggestion but I'm not a big fan of the background colors for the left sidebar and table. 

Can you walk through how you landed on these colors and what some other options might be?

jonbell-lot23:
I tried adding some contrast but here's a screenshot of how it would look if we stuck with our standard Replay dark theme colours:

image

jasonLaster:
That feels a bit better to me

jasonLaster (approved):

PR #5727, opened 2022-03-11
Continue to here

jaril:
This adds a shortcut for getting a "Continue to here" button:
- We currently show a `+` when hovering over a line in the editor
- Holding down **Cmd/Ctrl** will swap the `+` for a "Fast-forward to this line" button
- Holding down **Cmd+Shift/Ctrl+Shift** will swap the `+` for a "Rewind to line" button
- Buttons are grayed out if:
  1. The hits for that line are still being loaded
  2. The action is not-applicable, i.e. user cannot fast forward to a line if there are no future hits from the current pause point.

Follow up items:
- Clicking on the button does not block CodeMirror click events. This leads to weird behavior like multiple cursors (Cmd+Click), or highlighting ranges (Cmd+Shift+Click). Not sure what to do here but didn't want it to block this shipping. Either there's some way to have a click on the button to not trigger CodeMirror click events, or we refactor the button so that it's rendered as an overlay that's not a child of the CodeMirror element.

https://user-images.githubusercontent.com/15959269/157912886-b22248cc-8605-43a4-9b39-7d270f31b534.mov

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/index.js
+++ b/src/devtools/client/debugger/src/components/Editor/index.js
@@ -32,7 +32,7 @@ import HighlightLine from "./HighlightLine";
 import HighlightLines from "./HighlightLines";
 import EditorLoadingBar from "./EditorLoadingBar";
 import { EditorNag } from "ui/components/shared/Nags/Nags";
-import WelcomeBox from "../WelcomeBox";
+import { KeyModifiersContext } from "ui/components/KeyModifiers";
 
 import {
   showSourceText,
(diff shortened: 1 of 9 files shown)


src/devtools/client/debugger/src/components/Editor/index.js
```diff
@@ -437,7 +437,11 @@ class Editor extends PureComponent {
         <EmptyLines editor={editor} />
         <Breakpoints editor={editor} cx={cx} />
         <Preview editor={editor} editorRef={this.$editorWrapper} />
-        <LineNumberTooltip editor={editor} />
+        {
+          <KeyModifiersContext.Consumer>
+            {keyModifiers => <LineNumberTooltip editor={editor} keyModifiers={keyModifiers} />}
+          </KeyModifiersContext.Consumer>
```
jasonLaster, line 443: Nice!

PR #5730, opened 2022-03-11
Use launch darkly for maintenance mode

jazzdan:


ryanjduffy (approved):

src/ui/utils/launchdarkly.ts
```diff
@@ -1,28 +1,41 @@
...
-  if (!ready) {
-    return defaultValue;
+function useLaunchDarkly() {
+  const [ready, setReady] = useState(false);
+
+  useEffect(() => {
+    readyPromise.then(r => setReady(r));
+  }, [setReady]);
+
+  function getFeatureFlag(name: string, defaultValue: any) {
```
abstractalgo, line 30: `name: string` can be narrowed to a more precise union of string literals.

It could also be combined with a predefined set of default values.

```ts
const DEFAULT_VALUES = {
  "maintenance-mode": false
} as const;

function getFeatureFlag<F extends keyof DEFAULT_VALUES>(flagName: F) {
   if (!ready) {
     return DEFAULT_VALUES[flagName]
   }
   return client.variation(name, DEFAULT_VALUES[flagName])
}
```

abstractalgo (approved):

PR #5772, opened 2022-03-14
Migrate Console app to TS

abstractalgo:
Gotta love those `@ts-ignore`s 😓


src/devtools/client/webconsole/components/App.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { FilterDrawer } from "./FilterDrawer";
+import Warning from "ui/components/shared/Warning";
+import { FC, MouseEventHandler, useEffect } from "react";
+import { useSelector } from "react-redux";
+import JSTerm from "devtools/client/webconsole/components/Input/JSTerm";
+
+/**
+ * Console root Application component.
+ */
+const App: FC = () => {
```
markerikson, line 16: :hand: Nit: the React+TS community recommends _against_ using `React.FC`:

- https://github.com/facebook/create-react-app/pull/8177
- https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/

Instead the recommendation is to explicitly type `props` if it exists.

It especially doesn't add any help here, because this is a function component with no children.

You should be able to just drop the `: FC` here

src/devtools/client/webconsole/components/App.tsx
```diff
@@ -0,0 +1,72 @@
...
+  // @ts-ignore
```
markerikson, line 17: :question: What TS errors are you getting here?  Is it just that TS doesn't know what type `state` is?

markerikson (approved):
It'll run, so feel free to merge if you want.

That said, it _should_ be fairly straightforward to pull in the right type for `state` here.  I see that we've got a `UIState` type in use in other files, but from context this seems like it might be a different Redux store instance/state structure entirely, in which case I don't know if there's a type for its state yet.

src/devtools/client/webconsole/components/App.tsx
abstractalgo: Yeah, this state is still in JS files. Didn't want to bloat this PR with extra changes since this is already coming from another PR as a partial update that can be quickly merged.

PR #5776, opened 2022-03-14
Move selected network request ID to redux

jaril:


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -111,23 +112,23 @@ export const NetworkMonitor = ({
...
-                selectedRequest={selectedRequest}
+                selectedRequest={data.find(request => request.id === selectedRequestId)}
               />
             }
             endPanel={
-              selectedRequest ? (
+              selectedRequestId ? (
                 loadedRegions &&
-                getPointIsInLoadedRegion(loadedRegions, selectedRequest.point.point) ? (
+                getPointIsInLoadedRegion(loadedRegions, data.find(request => request.id === selectedRequestId)!.point.point) ? (
```
markerikson, line 124: :question: Looks like we've got `data.find()` multiple times.  Can we do that once earlier in the rendering logic and find the current request ahead of time?

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -1,51 +1,52 @@
...
   loading,
-  loadedRegions,
   requestBodies,
   requests,
   responseBodies,
   seek,
   selectFrame,
 }: PropsFromRedux) => {
-  const [selectedRequest, setSelectedRequest] = useState<RequestSummary>();
+  const selectedRequestId = useSelector(getSelectedRequestId);
```
jcmorrow, line 43: When this was state that was local to this component, it made sense to pass it through to its child `RequestDetails`. Now that this is coming from the global store, I think it would be better to fetch it separately from each component, so that they can be more self-contained (e.g. in the future, you could render a `RequestDetails` somewhere else and you as the renderer would not need to go asking the global store for the request ID to do that). It means duplicating this line across the two components, but I think it's closer to the `many small connected components` idea which is the modern norm (calling on @markerikson to correct me if I'm wrong on that one 😉 )

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -289,13 +296,18 @@ const RequestDetails = ({
...
   const activeTabs = tabs.filter(t => t.visible);
+  const closePanel = () => dispatch(hideRequestDetails());
 
   useEffect(() => {
     if (!activeTabs.find(t => t.id === activeTab)) {
       setActiveTab(DEFAULT_TAB);
     }
   }, [activeTab, activeTabs]);
 
+  if (!(loadedRegions && request && getPointIsInLoadedRegion(loadedRegions, request.point.point))) {
```
jcmorrow, line 307: I don't have a concrete suggestion here, but I think we should keep an eye on how many places stuff like this (check if we have changed our trim region) creeps in. I suspect there might be a more centralized way of doing these things to avoid future shotgun surgery.

jcmorrow (approved):
Couple of nits, but looks good 👍  I'm always a little sad to have to lift local state into the global store, but I think it's the right thing to do here.

src/ui/components/NetworkMonitor/index.tsx
markerikson: Hmm. We definitely encourage more components reading their own data from the store as a general principle.  In this case, though, I'm not sure we _want_ to do that.

Reading data from the store forces a component to re-render when that data changes. I'm not sure of the component hierarchy in this area of the UI, but it sounds list-y.

Ideally, if I have list item 4 selected, then click on list item 7, _only_ a max of three components render: item 4, item 7, and likely their parent list.

If the list item component reads a "current selected ID" value, now _every_ list item will re-render whenever that changes.  

But, if we read that "current selected ID" value in the list _parent_, the same thing happens because React re-renders recursively by default.

So, _if_ we want to optimize re-rendering, the best approach here would be:

- Parent reads "current selected item ID" and a list of all item IDs from the store
- Parent renders all list item children and passes in `selected` and `itemId` as props
- Child list item components are wrapped in `React.memo()` to avoid re-renders if props haven't changed
- Child list items read their own data from the store by ID

That way:

- List parent re-renders because it got the new `selectedItemId`
- List item 4 re-renders because it sees `selected` change, as does list item 7
- All other components get the same props and don't re-render

That's the hypothetical ideal. I don't know this area of the code yet, so I don't know whether it's _necessary_ here, or if we're even trying to optimize rendering perf around this part of the UI atm.

jcmorrow: https://www.loom.com/share/a1e09049f49742f39879e22322d18f05

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -111,28 +112,18 @@ export const NetworkMonitor = ({
...
-                    frames={frames[selectedRequest?.point.point]}
-                    selectFrame={selectFrame}
-                  />
-                ) : (
-                  <RequestDetailsUnavailable closePanel={closePanel} />
-                )
+              selectedRequestId ? (
+                <RequestDetails
+                  cx={cx}
+                  request={data.find(request => request.id === selectedRequestId)!}
```
jcmorrow, line 125: What do you think about making these things selectors as well rather than passing them in? If it's difficult don't worry about it, but it seems like since RequestDetails gets 90% of its data straight from the store it's weird to have it rely on a couple of things that come from the store through its parent 🤷

jcmorrow (approved):
One suggestion, but only do it if it makes sense to you and doesn't feel like a big lift. 🏋️

PR #5782, opened 2022-03-14
Set up Redux DevTools Extension and migrate quick-open code to TS

markerikson:
This PR:

- Adds Redux Toolkit as a dependency (yay!)
- Configures the Redux store to enable use of the Redux DevTools Extension in dev mode only:
  - Adds the DevTools extension enhancer
  - Configures the enhancer to "sanitize" large portions of the state (such as source file contents) to improve perf in both the extension and the UI, by making copies of the state and replacing those large values with an `"<OMITTED>"` string

- Converts the `quick-open` reducer and actions files to TS
- Replaces the existing quick open reducer + action creators with RTK's `createSlice`
- Adds TS types for `state.quickOpen` and the associated actions, and adds them to `UIState` and `UIActions`

Will clean up code in a follow-up commit tomorrow


src/ui/setup/store.ts
```diff
@@ -29,12 +37,67 @@ export function bootstrapStore(initialState: { app: AppState; layout: LayoutStat
...
+      };
+
+      // Use Immer to simplify nested immutable updates when making a copy of the state
+      return customImmer.produce(state, (draft: any) => {
+        if (draft.app) {
+          // TODO This is a DOM node in the Redux state and shouldn't even be here anyway
+          draft.app.videoNode = OMITTED;
+        }
+
+        sanitizeContents(draft.sources?.focusedItem.contents);
```
markerikson, line 86: :hand: Missed a `?.` here

src/ui/setup/store.ts
```diff
@@ -1,4 +1,6 @@
-import { combineReducers, applyMiddleware, Store } from "redux";
+import { combineReducers, applyMiddleware, Store, compose } from "redux";
+import { devToolsEnhancer, EnhancerOptions } from "@redux-devtools/extension";
+import { Immer } from "immer";
```
abstractalgo, line 3: Do we want to make `immer` an explicit dependency? Or is there a reason not to do that?

abstractalgo (approved):
lgtm 👍

src/ui/setup/store.ts
markerikson: We could. It comes with RTK, so we know the dep exists. The main question is whether it ought to be listed as a separate dep given that we import from it specifically.

I can go ahead and add it for completeness sake.

markerikson: Sure, and the other reason to do it is that skipping the explicit dep listing relies on `npm` to hoist transitive deps. In practice that _should_ work, but doesn't hurt to be safe here.

PR #5788, opened 2022-03-15
Fetch all hit counts at the same time

jcmorrow:
This pairs with https://github.com/RecordReplay/backend/pull/4899 to fetch all hit counts for a source in one go, and without fetching actual Point data, just a straight hit-count. This makes it possible to do things like https://github.com/RecordReplay/backend/issues/4750. In addition to adding the ability to call the new protocol endpoint, we have a few supporting changes:

- adds an experimental flag for turning this feature on. This feature only works on new recordings that support fetching hit counts for locations in batches
- this caches based on the sourceID, and on the current trim begin and end, so that we will refetch whenever the user changes focus
- in order to make that work I've made the action that fetches the hitcounts for the whole file run on every line hover. We use a memoizable action to short circuit most of the time, so we only fetch the hitpoint counts once for each combination of source and focus window, but anytime any of those change a line hover will fetch the latest hit count

https://user-images.githubusercontent.com/5903784/159845411-e270be48-1416-4801-8f60-e96593fd2b1b.mp4

### DIFF
--- a/src/devtools/client/debugger/src/actions/sources/breakableLines.js
+++ b/src/devtools/client/debugger/src/actions/sources/breakableLines.js
@@ -4,28 +4,28 @@
 
 //
 
-import { getSourceActorsForSource, getBreakableLines } from "../../selectors";
-import { setBreakpointPositions } from "../breakpoints/breakpointPositions";
-import { loadSourceActorBreakableLines } from "../source-actors";
-
-function calculateBreakableLines(positions) {
-  const lines = [];
-  for (const line in positions) {
-    if (positions[line].length > 0) {
-      lines.push(Number(line));
-    }
-  }
-
-  return lines;
-}
+import { getSourceActorsForSource } from "../../selectors";
+import {
+  loadSourceActorBreakableLines,
+  loadSourceActorBreakpointHitCounts,
+} from "../source-actors";
 
 export function setBreakableLines(cx, sourceId) {
-  return async (dispatch, getState, { client }) => {
-    let breakableLines;
+  return async (dispatch, getState) => {
     const actors = getSourceActorsForSource(getState(), sourceId);
 
     await Promise.all(
       actors.map(actor => dispatch(loadSourceActorBreakableLines({ id: actor.id, cx })))
+    );
+  };
+}
+
+export function setBreakpointHitCounts(sourceId) {
+  return async (dispatch, getState) => {
+    const actors = getSourceActorsForSource(getState(), sourceId);
+
+    await Promise.all(
+      actors.map(actor => dispatch(loadSourceActorBreakpointHitCounts({ id: actor.id })))
     );
   };
 }
--- a/src/devtools/client/debugger/src/reducers/source-actors.js
+++ b/src/devtools/client/debugger/src/reducers/source-actors.js
@@ -51,6 +51,14 @@ export default function update(state = initial, action) {
       state = updateBreakpointColumns(state, action);
       break;
 
+    case "SET_SOURCE_ACTOR_BREAKPOINT_HIT_COUNTS":
+      state = updateBreakpointHitCounts(state, action);
+      break;
+
+    case "set_trim_region":
+      state.values = clearBreakpointHitCounts(state);
+      break;
+
     case "SET_SOURCE_ACTOR_BREAKABLE_LINES":
       state = updateBreakableLines(state, action);
       break;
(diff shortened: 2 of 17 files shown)


src/devtools/client/debugger/src/reducers/source-actors.js
```diff
@@ -101,6 +105,31 @@ function updateBreakableLines(state, action) {
...
 }
 
+function updateBreakpointHitCounts(state, action) {
+  const value = asyncActionAsValue(action);
+  if (value.state === "pending") {
+    return state;
+  }
+
+  const hitCountsMap = value.value.hits.reduce((acc, { location, hits }) => {
+    acc[location.line] = { ...(location.line || {}), [location.column]: hits };
```
markerikson, line 115: :hand: Strictly speaking I don't think we have to make a copy here. We're starting with a fresh `acc` object and providing new empty objects inside the loop. That means it _is_ safe to literally mutate those line objects inside of this reducer, because they're brand new.  Not sure on the _actual_ perf cost here, but I think this would technically be a bit faster:

```ts
if (!acc[location.line]) {
  acc[location.line] = {}
}

acc[location.line][location.column] = hits;
return acc;
```

(yes I hate the "fill in a default value" logic here - really wish JS had an equivalent of Python's `defaultdict` type)

src/devtools/client/debugger/src/actions/sources/breakableLines.js
```diff
@@ -4,28 +4,28 @@
...
-import { getSourceActorsForSource, getBreakableLines } from "../../selectors";
-import { setBreakpointPositions } from "../breakpoints/breakpointPositions";
-import { loadSourceActorBreakableLines } from "../source-actors";
+import { getSourceActorsForSource } from "../../selectors";
+import {
+  loadSourceActorBreakableLines,
+  loadSourceActorBreakpointHitCounts,
+} from "../source-actors";
 
-function calculateBreakableLines(positions) {
```
jcmorrow, line 11: It looks a little weird to be replacing this function with a function that does a totally different thing, but that's just a trick of the diff - this was unused code.

jcmorrow:
The things I want to fix before putting this up for a final review are:

- Mark's feedback on the data munging (which can definitely be improved 😂 )
- Add focus section to the things that will query/requery hitCounts (so if we change the focus we should requery for this stuff, because the hitcounts will be different)
- Feature-flag (and maybe also by linker capability) whether we get the hitCounts this way or via the old analysis route.
- Make sure actually *setting* breakpoints still works 😄

jcmorrow:
Actually, let's just feature-flag this, I'll fix everything else after that 😄

src/devtools/client/debugger/src/reducers/source-actors.js
```diff
@@ -101,6 +105,31 @@ function updateBreakableLines(state, action) {
   return updateResources(state, [{ id: sourceId, breakableLines: value }]);
 }
 
+function updateBreakpointHitCounts(state, action) {
+  const value = asyncActionAsValue(action);
+  if (value.state === "pending") {
+    return state;
+  }
+
+  const hitCountsMap = value.value.hits.reduce((acc, { location, hits }) => {
```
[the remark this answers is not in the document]
jcmorrow: That's a good question. This returns one entry for each possible breakpoint in a file. For a small file, that's probably a couple dozen, for a large file, potentially thousands or maybe even tens of thousands.

src/devtools/client/debugger/src/reducers/source-actors.js
jcmorrow: yeah, your aversion to the `if` is exactly why I did it this way, but you're right, we should try not to make thousands of copies for no reason!

src/devtools/client/debugger/src/reducers/source-actors.js
```diff
@@ -101,6 +105,31 @@ function updateBreakableLines(state, action) {
...
+    acc[location.line] = { ...(location.line || {}), [location.column]: hits };
+    return acc;
+  }, {});
+
+  const lineHitCountMap = Object.keys(hitCountsMap).reduce((acc, line) => {
+    const columns = hitCountsMap[line];
+    acc[line] = columns[Math.min(...Object.keys(columns).map(Number))];
```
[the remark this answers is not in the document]
jcmorrow: yeah, it's actually probably a little cleaner to do this in the first reduce, but this is an interesting way to optimize in JS that I had never heard of before, so I'm glad you mentioned it!

src/ui/hooks/settings.ts
```diff
@@ -97,6 +98,7 @@ function convertUserSettings(data: any): ExperimentalUserSettings {
     enableEventLink: settings.enableEventLink,
     enableTeams: settings.enableTeams,
     showReact: settings.showReact,
+    liveCodeAnalysis: settings.liveCodeAnalysis,
```
jcmorrow, line 101: not sure if this is right, I'm just following the pattern here.

src/devtools/client/debugger/src/reducers/source-actors.js
```diff
@@ -101,16 +109,53 @@ function updateBreakableLines(state, action) {
...
+
+  if (!hasResource(state, sourceId)) {
+    return state;
+  }
+
+  return updateResources(state, [{ id: sourceId, breakpointHitCounts: hitCountsMap.hits }]);
+}
+
+function clearBreakpointHitCounts(state) {
+  const withoutBreakpointHitCounts = makeReduceAllQuery(
```
markerikson, line 142: :hand: FWIW I personally find this particular function a bit confusing / hard to read.  (Then again, I also just searched the codebase for `makeReduceAllQuery` and discovered the set of "make cached resource" functions for the first time, so I had no clue those existed until 2 minutes ago :)  )

BTW, as a side note: now that we've got Redux Toolkit in the codebase, we've got Immer available to us, which we can use to simplify any immutable update logic.  I started to see what this snippet might look like with Immer and decided it isn't as helpful here because we're converting the array into a lookup table anyway, but FYI that it's available for other spots where we're doing immutable updates.

markerikson (approved):
Don't have enough codebase context yet to comment on the sources changes specifically, but seems generally good at first glance

src/devtools/client/debugger/src/reducers/source-actors.js
jcmorrow: haha, two things: *I* had never seen this function before last night, and I could not agree more. I know how map reduce works, but I still just like sat scratching my head and looking at it for ten minutes. But I wanted to follow the established patterns as much as I could, so I'm sticking with it until we can do an overhaul on this reducer. And *yes* to immer, I have not used it a ton, but where I have it's been great, excited to have that on this project.

PR #5789, opened 2022-03-15
Purge unused CodeMirror addons

abstractalgo:
Part of #6033.

- [x] purge some of unused addons and modes


jasonLaster:
I like where this is headed. And yeah - lets land these as its own small PR. This looks like a win by itself

jasonLaster (approved):

src/devtools/client/webconsole/components/Input/CodeMirror.tsx
```diff
@@ -3,16 +3,6 @@ import React, { FC, useRef } from "react";
 import { Controlled as CodeMirror } from "react-codemirror2";
 
 import "codemirror/mode/javascript/javascript";
-import "codemirror/mode/htmlmixed/htmlmixed";
```
jasonLaster, line 6: i'm pretty sure we're not using htmlmixe or jsx

abstractalgo: Me too, the problem is still that we have another set of similar/same imports for CoreMirror for Editor.
I'm still trying to chase how we exactly init CodeMirror because we have a global instance and a number of things that expect to be able to get that instance. I'm hoping that I can dynamically load both CodeMirror and its plugins only when needed, but that seems involved as well, and would also require merging different instances together. CodeMirror registers itself globally and processes all its editors "together", so maybe creating some sort of context provider for CodeMirror is also a good idea. However, a lot of these are just assumptions of what _might_ work, as we currently just kinda wing it, but to be able to precisely control the perf, we want to control how and when things init.

Issue #5792, opened 2022-03-15
Pretty-printed original sources are shown as generated


hbenl:
See https://app.replay.io/recording/source-map-button-only-shown-for-generated-file--9d9e9c67-ae48-4169-82ec-0a95255dd9e4?point=39266744999444417451074577407938523&time=22492&hasFrames=true

PR #5793, opened 2022-03-15
Show pretty-printed original sources as original (#5792)

hbenl:
`ThreadFront.isSourceMappedSource()` also takes into account that original sources may also be pretty-printed (though that's rare).

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -1168,7 +1168,7 @@ class _ThreadFront {
         return false;
       }
       kind = minifiedInfo.kind;
-      assert(kind != "prettyPrinted", "source kind must be prettyPrinted");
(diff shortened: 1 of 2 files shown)


src/protocol/thread/thread.ts
```diff
@@ -1168,7 +1168,7 @@ class _ThreadFront {
         return false;
       }
       kind = minifiedInfo.kind;
-      assert(kind != "prettyPrinted", "source kind must be prettyPrinted");
+      assert(kind != "prettyPrinted", "source kind must not be prettyPrinted");
```
abstractalgo, line 1171: Non-blocking nit: move to `!==`, but this pattern of eqeq checks seems to be present in more than one place.

abstractalgo (approved):

PR #5812, opened 2022-03-16
Remove unused local prefs

jaril:
- Removes prefs/features that were dead code (i.e. wasn't being used anywhere else in the database)
- Removes features that have been on for a while and have been stable (e.g. comments, launch browser, event count, smooth playback).


abstractalgo (approved):
Let's test things that these flags should (not) affect, I guess that's the best approach with this.

\+ Typecheck fix, but otherwise seems reasonable

PR #5815, opened 2022-03-16
Issue 5202 toggle animation delay

### LINKED ISSUE #5202: Delay viewMode change to allow for toggle animation when changing mode from command palette
> That's fair. I would have expected the toggle to not animate if using the command palette.

It looks like this happens because we adjust localViewMode when it is out of sync with viewMode. This triggers a refresh of the ViewToggle, which animates to align with the viewMode when it is updated elsewhere. (Added a comment in the replay where this occurs).

Because the ViewToggle animation is reactive to the viewMode change, it will always happen after the layout has already updated. 

I made the changes to fix the sidebar, but will create a separate issue for the toggle animation timing issue.

I believe a potential fix would be to change `localViewMode` to `toggleViewMode` that is tracked in Redux instead of locally. Then, when the setViewMode function is called, we would update `toggleViewMode` to trigger the animation, have a setTimeout to allow for the animation, then move forward with dispatching the `selectedPrimaryPanel` and `viewMode` updates. We'd want to ensure `toggleViewMode` is only ever updated in the same function as `viewMode` to keep them in sync. Not sure if this is the best approach though.

_Originally posted by @ceceliacreates in https://github.com/RecordReplay/devtools/issues/4939#issuecomment-1027261238_

ceceliacreates:
moves toggle delay logic out of the component and into the setViewMode action. 

I had to change the useRef for the toggleTimeoutKey to just a variable because you can't have useRef outside a component. It seems to be working but let me know if there's a better approach there.

I also made a small change to the Types so all the layout actions have the same name format (some said Action at the end, some didn't). 

Resolves #5202 

Replay: https://app.replay.io/recording/toggle-animation-update--83ecbaed-288d-4fc3-b6bc-6287c38720a3


src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.css
```diff
@@ -25,3 +25,8 @@
   background-color: var(--grey-20);
   cursor: pointer;
 }
+
+/* Override common.css line 100 button pointer-events behavior so group-hover will work*/
+button > * {
```
jaril, line 30: Wrong PR?

src/ui/actions/layout.ts
```diff
@@ -63,6 +69,22 @@ export function setViewMode(viewMode: ViewMode): UIThunkAction {
...
+    let toggleTimeoutKey: NodeJS.Timeout | null = null;
+
+    //Update toggleMode to start toggle animation
+    dispatch(setToggleMode(viewMode));
+
+    // Delay updating the viewMode in redux so that the toggle can fully animate
+    // before re-rendering all of devtools in the new viewMode.
+    if (toggleTimeoutKey) {
+      clearTimeout(toggleTimeoutKey);
+    }
```
jaril, line 82: This won't be toggled because `toggleTimeoutKey` here will always be undefined, since it doesn't get set until well after.

src/ui/state/layout.ts
```diff
@@ -5,10 +5,13 @@ export type LayoutState = {
   selectedPrimaryPanel: PrimaryPanelName;
   selectedPanel: SecondaryPanelName;
   viewMode: ViewMode;
+  toggleMode: ToggleMode;
   toolboxLayout: ToolboxLayout;
 };
 
 export type ViewMode = "dev" | "non-dev";
+export type ToggleMode = "dev" | "non-dev";
```
jaril, line 13: We can just use `ViewMode` as the type where `ToggleMode` is currently being used.

src/ui/components/Header/ViewToggle.tsx
```diff
@@ -78,7 +54,7 @@ function ViewToggle({ viewMode, setViewMode }: PropsFromRedux) {
 
 const connector = connect(
   (state: UIState) => ({
-    viewMode: getViewMode(state),
+    toggleMode: getToggleMode(state),
```
jaril, line 57: `ToggleMode`'s kind of confusing without any context. Maybe `ViewToggleMode`? Even then I'm not a fan of the double verb, but at least it's somewhat connected. I'll leave it to you.

jaril (changes requested):

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.css
```diff
@@ -25,3 +25,8 @@
...
+  pointer-events: auto;
```
jaril: Let's ✂️ this before shipping.

jaril (approved):

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.css
jaril: Nevermind this comment not sure why it showed up 😬

PR #5822, opened 2022-03-16
Add a subtle flash whenever there's activity in the editor and it's not the selected panel

### LINKED ISSUE #5804: Light up the sources tab when we open the editor in the background
**The issue**

When you're in one of our new layouts, clicking stuff in the console or network monitor will open a new file in the editor, but you won't see it because it's opening in a separate sources tab. This is fine for now, but we need to at least message to the user that something happened.

**The proposal**

This is a UX regression. The best way to fix this quickly and subtly is to simply light up the sources tab with a white color that then fades back to the default gray over 2 seconds. This gives an indication of state change.

**Other ideas**

Jaril talked about completely suppressing things opening in the editor and Jason talked about navigating you to the editor. I disagree with both approaches because they're both too extreme imo, but let's leave that for the next iteration of this feature.

Short term, a simple color state change. Longer term we can debate whether we suppress (Jaril), yank (Jason), or message (Jon). But let's not let the product regression live on prod any longer than we need to. Part of shipping things we know are rough is being able to quickly address regressions, so let's quickly address this regression and iterate from there.

jaril:
Fix #5804.

@jonbell-lot23 got a version working decently well. For now it just overlays a 50% yellow layer to simulate the flash, but clearly that's the rough version. Let me know what you think v1 should look like!

https://user-images.githubusercontent.com/15959269/158672578-586eae95-e24f-4b7e-8095-742e8ed7d12f.mov

### DIFF
--- a/src/devtools/client/debugger/src/reducers/sources.d.ts
+++ b/src/devtools/client/debugger/src/reducers/sources.d.ts
@@ -21,9 +21,15 @@ export type Source = {
   extensionName?: string | null;
   introductionType?: "scriptElement";
 };
+type SelectedLocation = {
+  sourceId: string;
+  line: number;
+  column?: number;
+  sourceUrl: string;
+}
 
 export function getSelectedSourceWithContent(state: UIState): Source;
 export function getIsSourceMappedSource(state: UIState): boolean;
 export function getSources(state: UIState): Source[];
 export function getSelectedSource(state: UIState): Source;
-export function getSelectedLocation(state: UIState): Source;
+export function getSelectedLocation(state: UIState): SelectedLocation;
--- a/src/ui/components/SecondaryToolbox/TabSpotlight.tsx
+++ b/src/ui/components/SecondaryToolbox/TabSpotlight.tsx
@@ -1,0 +1,39 @@
+import classNames from "classnames";
+import { getSelectedLocation } from "devtools/client/debugger/src/reducers/sources";
+import { useEffect, useRef, useState } from "react";
+import { useSelector } from "react-redux";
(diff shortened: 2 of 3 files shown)


jaril:
```
Great prototype, thanks! A few comments, from specific to meta:

Specific: I'd make it white, and fade out earlier (nearly instant) and slower. It can be a super gentle tap on the shoulder, and I think this is more loud than we need.

Meta: We're kicking off a broader conversation about this. There are strong opinions about yank versus message versus suppress, so I'd stay tuned.
```

jonbell-lot23:
@jaril Following up on our chat in Discord:

This code assumes we put a new div on top of the other div and use opacity to fade it out. I think a better approach is to actually change the div itself. I was thinking the logic would go like this:

-> On state change, add a className
-> That class uses CSS transition to slowly fade away

And that means we could design anything without worrying about the limitations of an overlapping transparent div. So if we design something where the foreground and background colours should change, this code approach would allow that. Whereas the current transparency means we end up with an ugly transparent yellow color. 

Can we rework the code to apply to the div itself?

jasonLaster (approved):
Awesome to see the animation working.

I think we could probably make the animation more subtle, perhaps with an alternative to the background like a bottom border or badge...

Also we should evaluate the actions that navigate users to the source vs highlight the sources pane. A good example of this is as a user, if i select a source in the source explore my primary intention is to view the source and we should navigate there. If i click on a source link in the console that's also my primary intention. Fastforward/rewind that is a secondary intention and badging is fine.

jonbell-lot23:
Thanks for the approval but it's possible there's a whole file we don't even need to apply. Seey comments above -- I'd like to get badging working on the div itself, not with a div overlay that brings some animation issues.

PR #5828, opened 2022-03-16
The Great Thunk Migration of 2022

markerikson:
- Removes a dead `test-head.js` file
- Replaces uses of `toolbox` as an object arg with `window.gToolbox`
- Replaces uses of `parser` as an object arg with direct imports, and cleans up several other thunk arg usages
- Replaces the existing store setup with Redux Toolkit's `configureStore`
- **REWRITES ALL THE THUNK ARGUMENTS YAYYYYY!!!!!**
  - Replaced every single "object thunk arg" usage, like `{dispatch, getState, client}`, with the standard Redux thunk signature of `(dispatch, getState, extra)` instead
  - Updated thunk-related types to match the thunk middleware being used
- Disabled the immutability and serializability check middleware in `configureStore`, as we know we have both mutation and non-serializable values like `Map`s in state right now and we'll fix those another time

This is a _lot_ of changes, and if there was any other way to break them up I would have.

But this had to all be done in one shot.

### Status

~~I do still need to do additional cleanup - there's a bunch of dead code, including the old thunk middleware and its types.~~

~~I'm also seeing a bunch of "Not allowed" messages in the console in local dev, which appear to be coming from this function:~~

```ts
function NotAllowed() {
  console.error("Not allowed");
}
```

~~which in turn is used by `DisallowEverythingProxyHandler`.  Need to investigate that further.~~

The "Not allowed" messages were due to `ValueFront` and `app.videoNode` values in the state, as the Redux DevTools extension tried to serialize those.  Fixed that in #5854.

Cleaned up the dead code.

One of the QA Wolf tests was failing.  Jaril identified that the comments pane wasn't being selected as the app loaded, and that it was likely due to the async initial state not being used.  I fixed `store.ts` to actually use the passed-in `initialState`, which I completely missed earlier.

Assuming the checks pass, I _think_ this is good to go.


markerikson:
Updated the PR with:

- Fixes from #5854 to avoid the "Not allowed" message spam in dev
- Rebase against `master`

src/devtools/client/debugger/src/actions/sources/loadSourceText.js
```diff
@@ -5,6 +5,7 @@
 //
 
 import { PROMISE } from "ui/setup/redux/middleware/promise";
+import { parser } from "devtools/client/debugger/src/utils/bootstrap";
```
jaril, line 8: is this still needed?

src/devtools/client/debugger/src/actions/toolbox.js
```diff
@@ -9,19 +9,19 @@
  * @static
  */
 export function openLink(url) {
-  return ({ toolbox }) => {
-    toolbox.getPanel("debugger")?.openLink(url);
+  return () => {
+    window.gToolbox.getPanel("debugger")?.openLink(url);
```
jaril, line 13: 😬 someday it won't be on the window…

src/devtools/client/debugger/src/client/commands.d.ts
```diff
@@ -1,2 +1,48 @@
 export const setupCommands: object;
-export const clientCommands: object;
+
+export type $FixTypeLaterFunction = Function;
+
+// TODO All of these functions need to be given actual types, preferably by converting this file to TS
```
jaril, line 5: 🤣

src/devtools/client/debugger/src/utils/bootstrap.js
```diff
@@ -15,7 +15,6 @@ export function bootstrapWorkers(panelWorkers) {
 
   parser.start(() => new Worker(new URL("../workers/parser/worker", import.meta.url)), "parser");
   search.start(() => new Worker(new URL("../workers/search/worker", import.meta.url)), "search");
-  return { ...panelWorkers, parser, search };
```
jaril, line 18: Nice.

src/devtools/client/debugger/src/utils/memoizableAction.js
```diff
@@ -31,7 +31,18 @@ import { validateContext } from "./context";
  */
 export function memoizeableAction(name, { getValue, createKey, action }) {
   const requests = new Map();
-  return args => async thunkArgs => {
+  return args => async (dispatch, getState, extraArgs) => {
+    // TODO Consider rewriting usage sites to not need thunkArgs as an object
+
+    // This mimics the existing pattern from the legacy "object thunk args" middleware.
```
jaril, line 37: 👍

src/ui/utils/thunk.ts
```diff
@@ -45,46 +49,47 @@ export interface ThunkDispatch<TState, TExtraThunkArg, TBasicAction extends Acti
...
-  ): {
-    [TActionCreatorName in keyof TActionCreators]: ReturnType<
-      TActionCreators[TActionCreatorName]
-    > extends ThunkAction<any, any, any, any>
-      ? (
-          ...args: Parameters<TActionCreators[TActionCreatorName]>
-        ) => ReturnType<ReturnType<TActionCreators[TActionCreatorName]>>
-      : TActionCreators[TActionCreatorName];
-  };
+// declare module "redux" {
```
jaril, line 68: Is this intentionally left as a comment or just missed deletion?

jaril (approved):
Small nits but lgtm!

src/devtools/client/debugger/src/actions/sources/loadSourceText.js
markerikson: Yes, `parser` is used on line 46 - this is one of the couple files that referenced `parser` as a thunk extra arg, and I've eliminated that by switching to a direct import.

jaril:
You know what I think we're good! We have two intermittent tests — `stepping-05` and `inspector-05`. They're unrelated to this PR 👍

PR #5843, opened 2022-03-17
Fixing breakpoints dead-end with better text

jonbell-lot23:
Fixes: https://github.com/RecordReplay/devtools/issues/5830

Was:
image

image

Now:
image

image

Features include:

* An indication of how to fix the issue
* A link to Focus Mode in our docs
* Alignment with our other error colours
* Fixed a rounded edge bug -- was looking squared off and wrong in dark theme


jonbell-lot23:
@ceceliacreates note that I'm sending this link to Focus Mode in the docs rather than the original link that was talking more generally about the error. We should update the general error docs to have updated screenshots when this PR goes live.

jasonLaster (approved):

PR #5864, opened 2022-03-18
Preserve comments during editing

abstractalgo:
Part of https://github.com/RecordReplay/devtools/issues/5573.

Fixes https://github.com/RecordReplay/devtools/issues/5437.
Fixes https://github.com/RecordReplay/devtools/issues/5090.
Fixes https://github.com/RecordReplay/devtools/issues/5816.

To be able to preserve comment's content between panel switches we can utilize either app store or some storage mechanism. To be able to preserve the comment's content between page reloads (manually or because of being signed out and then signing in), we have to use some storage mechanism.

Using handlers on unmounts is not sufficient because they don't manage to handle the cases where the user suddenly closes the page. Thus leaving us an option of persisting comments along the way.

To do that, we debounce storage of a serialized version of the comment's content. It's unlikely that the user will input much new content before the timeout gets triggered.

To make more room for persisting partial comments between different recordings, we store partial comments per recording by assigning them a recording-specific storage key. We also add an extra suffix if the comment is not a pending comment but a reply or edit of an existing comment, because the UI allows for zero or one pending comment, zero or more reply comments and zero or more existing comments to exist at the same time.

The performance of the serialization and the storage seem good even for longer comments.

It's unlikely that the user will store hundreds of huge unsent comments and replies, so the storage demands are very low, and they get cleaned up after being done.

I've picked local storage as the storage mechanism for no special reason other than its simplicity and straightforwardness.

---

Updates:

- `useCommentsLocalStorage.ts`: a new hook that does ops with storage
- `CommentEditor.tsx`: a component shared by `<ExistingCommentEditor>` and `<NewCommentEditor>` that passes handlers to `<TipTapEditor>`
  - `TipTapEditor.tsx`: adds new handlers for some actions, i.e. what triggers hook's functions
- `ExistingCommentEditor.tsx`: has an actual new logic utilizing the new hook
- `NewCommentEditor.tsx`: has an actual new logic utilizing the new hook
- `CommentCard.tsx`: some types update


src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
```diff
@@ -59,7 +66,21 @@ function ExistingCommentEditor({
...
+        handleSubmit={handleSubmit}
+        onCreate={({ editor }) => {
+          const storedComment = commentsLocalStorage.get();
+          if (storedComment) {
+            editor.commands.setContent(JSON.parse(storedComment));
+          }
+        }}
+        onUpdate={debounce(({ editor }) => {
+          commentsLocalStorage.set(JSON.stringify(editor.getJSON()));
+        }, PERSIST_COMM_DEBOUNCE_DELAY)}
```
jaril, line 81: It looks like the onCreate and onUpdate logic are repeated in both `ExistingCommentEditor` and `NewCommentEditor` — is there a way to just pass the `mode` down and call the `useCommentsLocalStorage` at the `CommentEditor` component??

src/ui/components/Comments/TranscriptComments/CommentEditor/useCommentsLocalStorage.ts
```diff
@@ -0,0 +1,53 @@
+import { useGetRecordingId } from "ui/hooks/recordings";
+
+// types of editable comments:
+// - existing comments (zero or more active)
```
jaril, line 4: what's zero or more active referring to here?

jaril (approved):
Small nits but looks good to me!

src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
abstractalgo: Since we already have split components (`Exisiting` and `New`) to handle rendering and ~some logic, then it should be the responsibility of those components to provide their corresponding handlers. They're similar, but not quite the same.

src/ui/components/Comments/TranscriptComments/CommentEditor/useCommentsLocalStorage.ts
abstractalgo: Active/visible in the UI. Will add info.

PR #5866, opened 2022-03-18
Add descriptions in our assert calls

jaril:
This adds some descriptions so we know what assert errors map out to in Sentry.


abstractalgo (approved):

abstractalgo:
Extra file modified from another PR. @jaril

Issue #2908, opened 2021-06-21
Video source data gets reset(?) when switching from viewer to devtools


kannanvijayan:
When webm video playback is enabled, switching from viewer to devtools, or vice-versa, causes the video element's src contents seem to reset due to some interaction with react's component system switching views.

ryanjduffy:
Also fixed by #2921

Issue #2907, opened 2021-06-21
Fix video playback in frontend


kannanvijayan:
Video playback in frontend does not work as expected, when experimental feature is enabled.

ryanjduffy:
I think this will be fixed by #2921

kannanvijayan:
Almost good to go, but a couple of remaining issues:

1. I managed to hit the "play" button and still replicate the transparent playback issue (checkerboard background).
2. The replay playback button should be grayed out until the video blob is loaded.  I kept clicking it and there was no visual indication of when the video playback was ready.

kannanvijayan:
Closing to move over to new issue.

PR #5868, opened 2022-03-18
Remove the video DOM node ref from the Redux store

markerikson:
DOM nodes are not "state", and not serializable, so they shouldn't
go in the Redux store.

Here, the Redux store was being used to pass around the DOM node for
the playing video, so that `<Video>` could save the node ref, and
`graphics.ts` could later call methods on it.  Understandable as a
hack, but not a good idea.

We can accomplish the same thing by just having a tiny module with
a getter/setter pair, and store the DOM node ref there.

For previous background on why this DOM node was in the Redux store in the first place, see:

- #2921
- #2907
- #2908

Did some manual testing before PR to confirm that I can start playing a video in one mode, toggle to the other, and it keeps on playing.


jaril (approved):
LGTM!

abstractalgo (approved):

PR #5872, opened 2022-03-18
GraphQL types codegen

### LINKED ISSUE #5870: Generate GraphQL types
Consider using one of many codegen tools to auto-generate Typescript types for `devtools` based on RecordReplay/devtools GraphQL schema.

The general approach could be:
- run some `npm` script to fetch the latest schema, but using the "secret" so the schema stays private for the general public (as we use Hasura, this is a suggestion on how to do it: https://hasura.io/docs/latest/graphql/core/guides/export-graphql-schema.html#using-apollo-cli)
- run the codegen tool with the latest schema
- use types
- profit $$$

We'll be able to introduce types progressively into the codebase, instead of doing it in one large PR.

---

- codegen
- progressively improve hooks with types
- add types for cache and context

abstractalgo:
Closes #5870.

- [x] the script (`npm run gql`)
  - [x] script to fetch the schema (`npm run gql:schema`)
  - [x] script to autogen the types (`npm run gql:gen`)
- [x] add instructions to README
- [x] add a CI step
- [x] add generated files (entire `src/graphql` dir)


abstractalgo:
image

🤔 🤔 🤔 🤔

src/ui/hooks/settings.ts
```diff
@@ -102,10 +102,8 @@ function convertUserSettings(data: any): ExperimentalUserSettings {
 
 function getUpdateUserSettingQuery(key: SettingItemKey, type: "uuid" | "Boolean") {
   return gql`
-    mutation UpdateUserSettings($newValue: ${type}) {
-      updateUserSettings(
-        input: { ${key}: $newValue },
-      ) {
+    mutation UpdateUserSettings($newValue: Boolean) {
+      updateUserSettings(input: { showReact: $newValue }) {
```
abstractalgo, line 106: This is fun because dynamic string confuses the parser. We'll see what's the workaround for this. 🤔

abstractalgo:
Not even having types, already discovering inconsistencies:

image

🎉

abstractalgo:
Editing output `gql_schema.graphql` file (much easier to read and edit than `.json` file, and codegen can still work with both types):

image

Aaaaaand, we have types!! 🎉 🎉 🎉 🎉 🎉 🎉 

image

abstractalgo:
A couple of details about the script:
```
apollo codegen:generate --localSchemaFile=gql_schema.graphql --target=typescript --tagName=gql
```
- `localSchemaFile`: file with schema (can be either graphql or json)
- `target`: typescript
- `tagName`: keyword to recognize how we create `DocumentNode`s from strings (almost always is `gql`, that's a de facto standard)

This is the output with that regular script:

image

Basically, creates `__generated__` folders with types next to files that would use them.

But there is also this CLI argument:

- `--outputFlat`: By default, TypeScript/Flow will put each generated file in a directory next to its source file using the value of the "output" as the directory name. Set "outputFlat" to put all generated files in the directory relative to the current working directory defined by "output".

So we can have a flat output, instead of nested next to files, e.g.:
image

And the actual script's shape should be:
```
apollo client:codegen [OUTPUT] # we not "output" then '__generated__' is used
```

...so we can set where to put files (image above has it set as `apollo client:codegen src/graphql`).

Having a flat output in a single folder seems less messy and less repetitive and could make PR reviews easier by not polluting scope of changes across many palces at once, and instead keeping them contained in a single folder. Also, importing with flat structure would be easier - it's always the same folder to import from. So I chose to go with flat structure in this PR.

abstractalgo:
`next build` failing 🤔

package.json
```diff
@@ -146,6 +149,7 @@
     "eslint-plugin-react": "^7.29.2",
     "eslint-plugin-react-hooks": "^4.3.0",
     "fs-extra": "^10.0.1",
+    "graphql": "15.8.0",
```
jasonLaster, line 152: why are we using a lower version?

jasonLaster (approved):

package.json
abstractalgo: People have reported the codegen script not working on `16.x`, so a downgrade works.
Our codebase was referencing `14.7`, `15.8` and `16.3`, and only `15.8` works so far.

markerikson (approved):
First time I've seen any GraphQL-generated types, so not a lot of context here, but seems plausible at first glance

package.json
```diff
@@ -24,11 +24,14 @@
     "upload-e2e-mapped-styles-bundle-sourcemap": "cd public/test/examples/ && replay-sourcemap-upload --group e2e-test --key rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ mapped_styles.js*",
     "upload-e2e-mapped-minified-bundle-sourcemap": "cd public/test/examples/ && replay-sourcemap-upload --group e2e-test --key rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ mapped_minified.js*",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "gql:schema": "gq https://graphql.replay.io/v1/graphql -H \"X-Hasura-Admin-Secret: $HASURA_KEY\" --introspect > schema.graphql",
```
hbenl, line 28: We shouldn't rely on globally installed npm packages, so `graphqurl` (which provides the `gq` command) should be added to `devDependencies`.

hbenl (changes requested):
We should get the schema from `api.replay.io` and not from Hasura (i.e. `graphql.replay.io`)

package.json
abstractalgo: I had problems with getting it to work with a dev dep, but will revisit.

.github/workflows/typecheck.yml
```diff
@@ -11,3 +11,18 @@ jobs:
         run: npm install
       - name: Run tsc
         run: npm run typecheck
+  gql:
+    name: GraphQL schema typesdsa
```
hbenl, line 15: "typesdsa" ?

.github/workflows/typecheck.yml
```diff
@@ -11,3 +11,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Deps
+        run: npm install && npm install -g graphqurl
```
hbenl, line 20: is it still necessary to install graphqurl globally here?

package.json
```diff
@@ -24,11 +24,14 @@
     "upload-e2e-mapped-styles-bundle-sourcemap": "cd public/test/examples/ && replay-sourcemap-upload --group e2e-test --key rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ mapped_styles.js*",
     "upload-e2e-mapped-minified-bundle-sourcemap": "cd public/test/examples/ && replay-sourcemap-upload --group e2e-test --key rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ mapped_minified.js*",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "gql:schema": "./node_modules/.bin/gq https://graphql.replay.io/v1/graphql -H \"X-Hasura-Admin-Secret: $HASURA_KEY\" --introspect > schema.graphql",
```
hbenl, line 28: `./node_modules/.bin/` is unnecessary.

hbenl (approved):

PR #5881, opened 2022-03-21
Fixing error message

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/5856

We wrote this copy correctly [in February](https://www.notion.so/replayio/Back-end-error-messages-47afdf09b20f419cb8610138bd053d26) but it looks like we decided to collapse the Fatal and OldBuild errors. This puts it back to the original intent, which is giving a descriptive error when there's a fatal. Also used Brian's recommendation to talk about recording a second time, rather than a general "oops" dead-end.

### DIFF
--- a/src/ui/setup/dynamic/devtools.ts
+++ b/src/ui/setup/dynamic/devtools.ts
@@ -80,12 +80,11 @@ enum SessionError {
 const SessionErrorMessages: Record<number, string> = {
   [SessionError.BackendDeploy]: "Please wait a few minutes and try again.",
   [SessionError.NodeTerminated]: "Our servers hiccuped but things should be back to normal soon.",
-  [SessionError.KnownFatalError]:
-    "This error has been fixed in an updated version of Replay. Please try upgrading Replay and trying a new recording.",
+  [SessionError.KnownFatalError]: "Darn! There was an error. Please try recording again.",
   [SessionError.OldBuild]:
     "This error has been fixed in an updated version of Replay. Please try upgrading Replay and trying a new recording.",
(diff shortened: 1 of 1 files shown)


src/ui/setup/dynamic/devtools.ts
```diff
@@ -80,12 +80,11 @@ enum SessionError {
 const SessionErrorMessages: Record<number, string> = {
   [SessionError.BackendDeploy]: "Please wait a few minutes and try again.",
   [SessionError.NodeTerminated]: "Our servers hiccuped but things should be back to normal soon.",
-  [SessionError.KnownFatalError]:
-    "This error has been fixed in an updated version of Replay. Please try upgrading Replay and trying a new recording.",
+  [SessionError.KnownFatalError]: "Darn! There was an error. Please try recording again.",
```
bhackett1024, line 83: This doesn't look right to me.  It sounds good to have the same error message for KnownFatalError and OldBuild like we did before, but this PR is updating the message for KnownFatalError and not UnknownFatalError.  I think this should leave the message for KnownFatalError alone, and use this new message for UnknownFatalError.

jonbell-lot23: Good catch, will fix now

jonbell-lot23: Moving to new PR: https://github.com/RecordReplay/devtools/pull/5907

jonbell-lot23:
Moving to new PR: https://github.com/RecordReplay/devtools/pull/5907

PR #5896, opened 2022-03-21
Migrate webconsole/*/filters.js to RTK+TS

### LINKED ISSUE #5894: Migrate webconsole filters logic to RTK+TS
Ref: https://github.com/RecordReplay/devtools/issues/5893

markerikson:
- Migrated filters reducer to createSlice
- Removed unused "clear messages" case
- Moved `getAllFilters` selector into the reducer file
- Added filters state to UIState type

Fixes #5894


src/devtools/client/webconsole/actions/filters.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type MessageFilters = "error" | "warn" | "info" | "debug" | "log";
+
+const FILTER_PREF_MAP: Record<MessageFilters, PrefsKeys> = {
+  error: "filterError",
+  warn: "filterWarn",
+  info: "filterInfo",
+  debug: "filterDebug",
+  log: "filterLog",
+};
```
abstractalgo, line 26: `as const`

abstractalgo (approved):
I would prefer return types on functions, again just to help distinguish what's a pure function, and what an action (`void`) etc to make reading just slightly easier, but I guess it is a personal preference to a good degree, I most certainly won't block on it.

Overall, seems good 👍

PR #5901, opened 2022-03-21
Fix persistence of webconsole prefs

### LINKED ISSUE #5845: Console timestamps are shown by default
**Description:**
I bet @markerikson recent console pref change removed a pref which was storing console timestamps

Bug: console logs now show timestamps by default
ER: the pref is off by default and persisted

markerikson:
This PR:

- Adds `consoleUI` to the `UIState` definition
- Restores use of `prefs.timestampsVisible` in defining `consoleUI` initial state
- Restructures webconsole prefs persistence to centralize the persistence logic and actually set persistence values reactively

Previously, the "log level" prefs were being persisted in a thunk,
in `actions/filters`.  Jason wants pref persistence to be reactive
and in a single location, in `ui/setup/prefs`.

Also, the `timestampsVisible` flag wasn't being persisted properly,
and the `node_modules` pref looks like it was never persisted.  Both
of those are now persisted correctly.

Fixes #5845

### DIFF
--- a/src/devtools/client/webconsole/reducers/ui.ts
+++ b/src/devtools/client/webconsole/reducers/ui.ts
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import { createSlice, createNextState } from "@reduxjs/toolkit";
 
-interface WebconsoleUIState {
+export interface WebconsoleUIState {
   timestampsVisible: boolean;
 }
 
--- a/src/ui/setup/prefs.ts
+++ b/src/ui/setup/prefs.ts
@@ -1,6 +1,7 @@
 import { UIStore } from "ui/actions";
 import { UIState } from "ui/state";
 import { prefs, asyncStore } from "ui/utils/prefs";
+import { prefs as webconsolePrefs } from "devtools/client/webconsole/utils/prefs";
 import { getRecordingId } from "ui/utils/recording";
 import {
   getConsoleFilterDrawerExpanded,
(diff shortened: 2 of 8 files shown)


src/devtools/client/webconsole/reducers/ui.ts
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 import { createSlice, createNextState } from "@reduxjs/toolkit";
 
-interface WebconsoleUIState {
+export interface WebconsoleUIState {
```
jaril, line 6: This is the smallest nit but what's your take on camel case for acronyms — WebconsoleUIState, or WebconsoleUiState?

src/ui/setup/prefs.ts
```diff
@@ -37,20 +38,48 @@ export function registerStoreObserver(
...
   }
 
   updatePref("viewMode", getViewMode);
+
   updateAsyncPref("eventListenerBreakpoints", (state: UIState) => state.eventListenerBreakpoints);
   updateAsyncPref("commandHistory", (state: UIState) => state.messages?.commandHistory);
 
+  // This is lazy-loaded, so it may not exist on startup
+  if (state.consoleUI && oldState.consoleUI) {
+    console.log(state);
```
jaril, line 68: :salt:

src/ui/setup/prefs.ts
```diff
@@ -37,20 +38,48 @@ export function registerStoreObserver(
...
+      return state.consoleUI.timestampsVisible;
+    });
+  }
+
+  if (state.filters && oldState.filters) {
+    updateWebconsolePref("filterError", (state: UIState) => state.filters.error);
+    updateWebconsolePref("filterWarn", (state: UIState) => state.filters.warn);
+    updateWebconsolePref("filterInfo", (state: UIState) => state.filters.info);
+    updateWebconsolePref("filterDebug", (state: UIState) => state.filters.debug);
+    updateWebconsolePref("filterLog", (state: UIState) => state.filters?.log);
```
jaril, line 79: Is there something special about log that we need the optional chaining in `state.filters?.log`?

jaril (approved):
Left non-blocking comments. Just need to take out the `console.log` and we're good to go!

src/devtools/client/webconsole/reducers/ui.ts
markerikson: Uh... I've never had to formally think about this before? :)

https://stackoverflow.com/questions/15526107/acronyms-in-camelcase looks like it has some good discussion

I've fallen towards `userId` as of late, although "ID" isn't an acronym.

Microsoft's advice suggests capitalizing both letters in a two-letter acronym.

But no hugely strong opinion here.

PR #5907, opened 2022-03-22
Tweaking error messages

jonbell-lot23:
I misunderstood the previous ask. Here's where we are now:

* SessionError.OldBuild and SessionError.KnownFatalError have the same message: upgrade and try again
* SessionError.UnknownFatalError talks about trying again (no upgrade message)


bhackett1024 (approved):
Thanks!

PR #5922, opened 2022-03-22
Extract EditorWithAutocomplete for reuse

jaril:
Rebase on top of #5915 when ready.


abstractalgo (approved):
lgtm

abstractalgo:
Oh yeah, there's a weird lock file change, that probably shouldn't be there.

PR #5930, opened 2022-03-22
Add scaffolding for React Testing Library tests inside of Jest

markerikson:
This PR:

- Adds React Testing Library and related packages
- Adds deps needed to run additional Jest tests
- Configures Jest to seemingly load our client codebase without exploding
  - Configured TS to recognize Jest globals
  - Added Jest support for `src` folder aliases
  - Added `babel-plugin-import-meta` to handle worker files
  - Added a Jest transformer for text files (used by `.properties`)
  - Added setup file to stub out globals like `Worker`
- Rewrites some weird import statements to work okay under Jest
- Adds a `testUtils.tsx` file that exports `createTestStore` and `render` utils for use in component tests
- Adds a tiny test file that _should_ (🤞) create a test store without exploding and verify that a lazy-loaded slice of state exists

Note that it's a miracle that the new "create a store without exploding" test can work at all, given how complex the store setup logic is and how many sub-dependencies are involved.  (See [`docs/devtools-setup-sequence.md`](https://github.com/RecordReplay/devtools/blob/c8dc10a34760276d1e5537d6ace2ea1b609ee586/docs/devtools-setup-sequence.md) for details on what that setup sequence actually includes.)


src/devtools/client/webconsole/components/FilterBar/__tests__/FilterBar.test.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { render, createTestStore } from "test/testUtils";
+
+describe("First example test", () => {
+  it("Creates a store without exploding", async () => {
+    const store = await createTestStore();
+    expect(store.getState().consoleUI).toBeTruthy();
+  });
```
abstractalgo, line 8: Isn't it the idea to auto-wrap renders with context that is a store (+ provide partial initial state) so that we can just render without having to explicitly sync on store creation? Or because of other things that may need initializing, it's fine to sync manually on it?

src/test/testUtils.tsx
```diff
@@ -0,0 +1,46 @@
...
+// This type interface extends the default options for render from RTL, as well
+// as allows the user to specify other things such as initialState, store. For
+// future dependencies, such as wanting to test with react-router, you can extend
+// this interface to accept a path and route and use those in a <MemoryRouter />
+interface ExtendedRenderOptions extends Omit<RenderOptions, "queries"> {
+  preloadedState?: Partial<UIState>;
+  store?: UIStore;
+}
+
+export async function createTestStore(preloadedState: Partial<UIState> = {}) {
```
abstractalgo, line 22: Would it make sense to consider `DeepPartial`? There probably is some value in that proposal, but considering how we generally tended to organize store so far (pretty flat), maybe not so.

test/jest/jest-text-transformer.js
```diff
@@ -0,0 +1,39 @@
+"use strict";
```
abstractalgo, line 1: I hope you didn't write this from memory. 😅

abstractalgo (commented):
Truly is a miracle, I would most definitely expect something to blow up. :D But we also don't yet initialize some important global objects, and those carry the majority of "hard business logic", so this might even seem reasonable.

Would it make sense to add tests (even these simple ones) to CI to pick up on our mistakes asap? Or they are already included as part of unit tests, right?

test/jest/jest-text-transformer.js
markerikson: Heh, no, that came from https://github.com/janouma/jest-text-transformer.

Actually given that I copy-pasted it, I should attribute the source in here.

src/test/testUtils.tsx
markerikson: No, you really _can't_ do `DeepPartial` - each slice reducer either gets all of its slice state or `undefined`, so it can't be "deep"

src/devtools/client/webconsole/components/FilterBar/__tests__/FilterBar.test.tsx
markerikson: In an actual test, this will be:

```js
const { getByText } = render(<MyComponent />)
```

But for this first PR I just wanted to verify that creating the store runs without exploding, and loads the lazy reducers.

markerikson:
This is part of the Jest "unit test" suite.  If you look at https://github.com/RecordReplay/devtools/runs/5651642052?check_suite_focus=true you can see that the tiny "create a store" test did indeed run as part of this PR!

I will work on adding a first "real" test in a follow-up PR.  For now I just wanted to get the scaffolding in place.

src/devtools/client/webconsole/components/FilterBar/__tests__/FilterBar.test.tsx
abstractalgo: Got it. Cool cool cool 👍

abstractalgo (approved):

PR #5943, opened 2022-03-23
Implement autocomplete using createPortal

jaril:
Rebase once #5942 is merged.

The current implementation of autocomplete clashes when it's a child of CodeMirror, since CodeMirror is snoopy and listens in on the autocomplete's events. For example, scrolling autocomplete while in the breakpoint panel (which is child of the CodeMirror elements) ends up scrolling the entire editor.

Using a portal circumvents that problem entirely.


jaril:
Hard yolo-ing @jasonLaster

PR #5947, opened 2022-03-23
Configure Jest to run React Testing Library component tests

markerikson:
This PR:

- Sets up additional Jest config (`jest-dom` types, more folder aliases)
- Sets up Apollo's `<MockedProvider>` for use with test-rendering components
- Mocks out websockets
- Adds an initial test that renders the entire `<WebConsoleApp>` component
- Reworks several imports to avoid a "React is not initialized" error that occurred _only_ when running `jest --coverage`
- Moves the test file to `WebConsoleApp.test.tsx`
- Adds utils to filter out unwanted spammy console messages in tests


bvaughn:
Looking at the history of the `WebConsoleApp.test` file that was added with this PR, I see that it passed (locally for me) on this initial commit (5a40b7a1ed30a754a59b1f00a037c18fb1d715aa) but fails on every commit after (starting with 3fa9900783b73fbfda0952e6a6ec591b8ad4589a).

PR #5950, opened 2022-03-24
0323 5874 and 5944 darktheme tweaks

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/5944

Before:
image

After:
image

@jaril remember to check in dark and light theme, and we should never be setting a bgcolor without a color. (Maybe we can automate this?)

### DIFF
--- a/src/ui/components/shared/CodeEditor/AutocompleteMatches.tsx
+++ b/src/ui/components/shared/CodeEditor/AutocompleteMatches.tsx
@@ -26,7 +26,7 @@ function Match({
   return (
     <button
       className={classNames(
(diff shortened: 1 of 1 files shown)


jasonLaster (approved):

jaril:
@jonbell-lot23 With CSS I'm cautious of additive tasks, since ideally the cascade should just keep you from getting into undesirable situations like this.

Filing a separate fix just to keep us from getting into that situation in the first place!

jonbell-lot23:
Left comments -> https://github.com/RecordReplay/devtools/pull/5956#pullrequestreview-920765871

PR #5955, opened 2022-03-24
Update SourceMap toggle

### LINKED ISSUE #5954: Update Source Map button
It'd be simpler if the toggle read "Use source map". We might be able to simplify the footer as well.
 
image (1)

jasonLaster:
fixes #5954

### New version
Screen Shot 2022-03-24 at 9 52 14 AM

### Old version
Screen Shot 2022-03-24 at 9 52 37 AM


jonbell-lot23:
100%, it's such a strong implicit connection that we don't lose a lot. But we gain a simpler visual

PR #5958, opened 2022-03-24
Keep presses on editor buttons from propagating to codemirror

### LINKED ISSUE #5839: Using arrow keys to move cursor in editor also moves the recording timeline
**Description:**
https://app.replay.io/recording/using-arrow-keys-in-editor-also-moves-in-timeline--d80a6626-46d8-496d-9a45-f81b914b7883?point=48028745944638700543075875408578059&time=18380&hasFrames=true

**Steps to Reproduce:**

**Expected Result:**

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay:

jaril:
Fix #5839.

We used to have this "trick" for keeping CodeMirror from reacting to clicks to its children. It's only necessary in this case since we have clickable items (e.g. the + button) that, when clicked, ends up moving CodeMirror's cursor.

This puts it back.


jasonLaster:
Nice

PR #5962, opened 2022-03-24
Add integration test for the webconsole messages list

markerikson:
This PR:

- Adds docs notes on the structure of the webconsole UI components and Redux usage
- Does some more config work to enable processing of recorded websocket messages in a Jest environment
- Silences a bunch of annoying error/warn/log messages (some just in tests, plus a couple actual log statements that were running in real code)
- Fills out the TS types for the `state.messages` slice, which for some reason only listed one field. (Some of the types are still placeholders, but at least all fields exist now.)
- Adds a util script to parse a HAR network archive, extract all Replay websocket messages, filter out large / unwanted messages, and write the rest to a JSON file for use in a fixture
- Reworks the webconsole UI integration test to actually load those recorded socket messages into the store, verify the right list of messages is rendered, and check that text filtering works as expected


test/jest/exportWebsocketMessages.js
```diff
@@ -0,0 +1,53 @@
...
+
+function main() {
+  const [node, script, harPath] = process.argv;
+
+  if (typeof harPath !== "string" || !harPath.endsWith(".har" || !fs.existsSync(harPath))) {
+    console.log("Must provide a valid path to a .har file");
+    process.exit(1);
+  }
+
+  processHar(harPath);
```
jasonLaster, line 50: I am pretty sure that we do not support network events in the console. which means we can probably remove this test and delete more code

markerikson: The `"Network.requests"` messages were definitely in the data sent via websocket from the backend.  They actually contained a bunch of request info.  Either that's being used by the Network Monitor, or the backend is sending them and they're not being used.

I stripped those out of the export to help shrink the size of the fixture file I committed, but also because I figured it was better to blanket remove those for now rather than try to do spend time combing through the exported websocket messages to make sure there wasn't any personal/sensitive info in the fixture.

jasonLaster: oh interesting - maybe they're shown in console warnings :/ good thing we're testing them :)

src/protocol/thread/thread.ts
```diff
@@ -294,7 +294,7 @@ class _ThreadFront {
 
     client.Session.addLoadedRegionsListener((parameters: loadedRegions) => {
       // Log loaded regions to help with diagnostics.
-      console.log("LoadedRegions", parameters);
+      // console.log("LoadedRegions", parameters);
```
jaril, line 297: Let's keep this in (for now): https://github.com/RecordReplay/devtools/commit/7b961a6d6f258c5a0ec0e8fc2a67ceb8a699f142

jaril (approved):

PR #5968, opened 2022-03-25
Default to prefs rather than big dumb blank page

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/5840

When you select settings, we should default to something useful. Personal isn't super useful.

Was:
image

Now:
image


jasonLaster:
Agreed that preferences is more useful,  but still think Personal should be top of the list.

jonbell-lot23:
> Agreed that preferences is more useful, but still think Personal should be top of the list.

Top of the list but default to Preferences? Jon answer: ok

Or scrap the PR? Jon answer: disagree. Even if this page was full of personal details, it's the definition of a "set and forget" screen that shouldn't take top billing. But as it is, it's just a log out button. So it's earned the spot even less.

jaril:
> Or scrap the PR? Jon answer: disagree. Even if this page was full of personal details, it's the definition of a "set and forget" screen that shouldn't take top billing. But as it is, it's just a log out button. So it's earned the spot even less.

Was gonna say that, as a general rule, I would I find it odd for it to take any more than 2 clicks to log out from the main screen. But practically, I don't see people logging out to a big thing that users do, so meh.

I say keep Personal as top of the list, but default to Preferences is a-okay.

PR #5979, opened 2022-03-25
Allow an initialHeight to be set in accordion items

jcmorrow:
This was really bothering me, because our `sources` item sits at the top of our default accordion, and often its content is large, but its initial size is tiny, while the bottom panel is huge and has very little content. I have added an optional `initialHeight` prop that can be provided to start the sections off at certain heights.  I thought about trying to do something fancy like defaulting the sizes of panels to `100/n` percent, where `n` is the number of accordion items, but that seemed like sometimes it would not be what the user wanted (and if they *do* want it, they can make that happen with just an `initialHeight` prop anyways).

Before:

CleanShot 2022-03-25 at 14 01 30

After:

CleanShot 2022-03-25 at 14 02 27

Also, we were trying to set an initial `expanded` state, but I don't think it was working, because we were not looking at `props`, which should be where that was set.

### DIFF
--- a/packages/accordion/index.tsx
+++ b/packages/accordion/index.tsx
@@ -30,17 +30,18 @@ import {
 
 // When I wrote this code God and I knew what I was doing. Now only God knows.
 
+export type SectionHeight = string | number;
+export type CollapsedState = boolean[];
+export type CreasesState = number[];
 export interface AccordionItem {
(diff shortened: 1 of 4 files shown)


jaril (approved):

jaril:
🚀

PR #5982, opened 2022-03-26
Add experimental hit counts to the outline view

jasonLaster:
Screen Shot 2022-03-25 at 5 47 21 PM

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
@@ -8,11 +8,10 @@ import MaterialIcon from "ui/components/shared/MaterialIcon";
 import hooks from "ui/hooks";
 import { Nag } from "ui/hooks/users";
 import { selectors } from "ui/reducers";
-import { prefs } from "ui/utils/prefs";
-import { prefs as prefsService } from "devtools/shared/services";
+import { prefs, features } from "ui/utils/prefs";
 import { trackEvent } from "ui/utils/telemetry";
 import { shouldShowNag } from "ui/utils/user";
-import { getHitCounts, getSelectedSource } from "../../reducers/sources";
+import { getHitCountsForSelectedSource, getSelectedSource } from "../../reducers/sources";
 import StaticTooltip from "./StaticTooltip";
 
 const { runAnalysisOnLine } = require("devtools/client/debugger/src/actions/breakpoints/index");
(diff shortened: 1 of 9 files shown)


src/devtools/client/debugger/src/components/SourceOutline/getOutlineSymbols.tsx
```diff
@@ -2,19 +2,34 @@ import { fuzzySearch } from "../../utils/function";
 import { groupBy, keyBy } from "lodash";
 import { SourceSymbols, ClassSymbol, FunctionSymbol } from "../../types";
 
-export function getOutlineSymbols(symbols: null | SourceSymbols, filter: string) {
+function addHitCountsToFunctions(functions, hitCounts) {
+  return functions.map(functionSymbol => {
```
markerikson, line 6: :hand: Got a `map()` that doesn't _look_ like it returns in all cases.  Does the loop always hit something, or are there times we wouldn't find anything and would return `undefined`?

jasonLaster: Might have written this code in 10 minutes before running to dinner 😉

src/devtools/client/debugger/src/components/SourceOutline/getOutlineSymbols.tsx
```diff
@@ -2,19 +2,34 @@ import { fuzzySearch } from "../../utils/function";
...
     return null;
   }
   const { classes, functions } = symbols;
+  const functionsWithHits = addHitCountsToFunctions(functions, hitCounts);
+  console.log("yo", functionsWithHits);
   const classNames = new Set(classes.map(s => s.name));
-  const functionsByName = keyBy(functions, "name");
-  const filteredFunctions = functions.filter(
+  const functionsByName = keyBy(functionsWithHits, "name");
+  const filteredFunctions = functionsWithHits.filter(
```
jasonLaster, line 26: In the future - I think we'll transition to asking the server for the source's symbols. At that point the server can just return the hitcounts for the symbols as well.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -87,17 +86,19 @@ export default function LineNumberTooltip({
...
   const analysisPoints = useSelector(selectors.getPointsForHoveredLineNumber);
 
   let analysisPointsCount: number | undefined;
 
-  if (prefsService.getBoolPref("devtools.features.codeHeatMaps")) {
-    analysisPointsCount =
-      hitCounts && lastHoveredLineNumber.current
-        ? hitCounts[lastHoveredLineNumber.current]
-        : undefined;
+  if (features.codeHeatMaps) {
```
jasonLaster, line 95: we should use `prefs` and `features` where we can

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -116,20 +117,15 @@ export default function LineNumberTooltip({
     if (lineNumber !== lastHoveredLineNumber.current) {
       lastHoveredLineNumber.current = lineNumber;
     }
-    if (prefsService.getBoolPref("devtools.features.codeHeatMaps")) {
-      setTimeout(() => {
-        if (lineNumber === lastHoveredLineNumber.current) {
+    setTimeout(() => {
+      if (lineNumber === lastHoveredLineNumber.current) {
```
jasonLaster, line 121: lifted these conditions up so that it was a bit more obvious what the feature flag was doing

jasonLaster (commented):
Some small tweaks. The biggest remaining issue is that the data seems to lie to us:

1. the line number tooltip values are different from what you get if you click to add a print statement
2. the function hit counts are different than what you'd get if you hover on the first line.  We can probably special case the logic for one line functions vs. multiple line functions

jasonLaster:
Just added some logic for differentiating between multi-line and single line functions. Not sure why the hit counts are still different though because at this point the locations are matching up.

jcmorrow:
Ah, OK, I am 90% sure I know why we sometimes return an incorrect number for lines - I think sometimes we are double-counting. This might end up being slightly tricky on the backend. I'll write it up in Discord.

src/devtools/client/debugger/src/components/SourceOutline/getOutlineSymbols.tsx
jcmorrow: This *should* always return something. If it doesn't, then we will return `0` correctly down on line 19 👍

PR #6011, opened 2022-03-28
Bump @sentry packages

abstractalgo:
Still don't understand why, but seems that these two had problems specifically that no other bumps had:

- https://github.com/RecordReplay/devtools/pull/5999
- https://github.com/RecordReplay/devtools/pull/5990

Closing those two PRs in favor of this one.


abstractalgo:
...aaaaand now it's fine. 🤔 🤔 🤔 

I'm not going to pretend to know what happened here. It even failed when running locally. But now it's fine. Maybe it just likes 6xxx numbers on PRs better than 5xxx, dunno.

PR #6023, opened 2022-03-28
Migrate existing messages reducer/selectors and `ConsoleOutput` component to TS

markerikson:
This PR:

- Adds a bunch of notes on the logic inside `reducers/messages`
- Converts `reducers/messages`, `selectors/messages`, `selectors/index` and `components/ConsoleOutput` from JS to TS
- Removes apparently dead code for dealing with console groups from `reducers/messages`


docs/codebase-notes/messages-reducer.md
```diff
@@ -0,0 +1,262 @@
+# Messages Reducer Notes
```
jasonLaster, line 1: wonder if this should be closer to the Web Console codebase. Is this going to go stale...

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -160,44 +179,18 @@ class ConsoleOutput extends React.Component {
...
-        className: "webconsole-output",
-        role: "main",
-        ref: node => {
-          this.outputNode = node;
-        },
-      },
-      <ConsoleLoadingBar />,
-      messageNodes
+    return (
+      <div className="webconsole-output" ref={this.outputNode} role="main">
```
jasonLaster, line 183: maybe i'm being pedantic, but `ref={this.outputNode}` could be it's own PR as well and much easier to review and see if something broke...

jasonLaster (commented):
Some general comments. Will review more later... lots here

docs/codebase-notes/messages-reducer.md
markerikson: We don't have any kind of a nested `docs` folder atm, and I did specifically add a `docs/codebase-notes/README.md` that says "this material could go stale as we're refactoring, but I'm leaving it here so that it's saved somewhere and other folks can see it".

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
markerikson: I'd classify it as a relatively low-risk change, especially given that the file is now TS.  All uses of `this.outputNode` now specifically use `this.outputNode.current!` and the types line up.  Plus, I would have had to define `outputNode = HTMLDivElement | null` anyway if I was going to stick with the callback form, so doing it as a ref object now simplifies that.

PR #6030, opened 2022-03-29
Stabilize comment editing

jasonLaster:
This is a proof of concept PR that does a reasonably good job at fixing the comment flickering with a couple of strategies. 

1. Stabilize the comment keys. The keys now use createdAt directly.
2. Use Josh's trick for stabilizing the list of comments in the transcript.
3. Don't render the transcript while we're still loading the user, recording, or comments. 
4. Don't remove the pending comment until we've finished creating or updating the comment

Note, the localStorage cache for pending comments is still pretty broken, but it does not contribute to flickering so i left it in. If you'd like to see how comments work without it you can remove the commit.

---

NOTE: We can discuss tomorrow how we'd like to move forward. This version is not perfect. For instance focus state is sometimes dropped. I have seen some flickering although i think it might also be hot module reloading related. I also think there are some architectural improvements we could make if we would like to. The goal for this PR is to try and find as many sources as flickering as possible and also rule out other sources that seem like they could be causing flickering, but don't.

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentCard.tsx
@@ -53,7 +53,7 @@ function CommentItemHeader({
 
   useEffect(() => {
     setRelativeDate(formatRelativeTime(new Date(comment.createdAt)));
-  }, []);
+  }, [comment.createdAt]);
 
   if (!comment.user) {
     return null;
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
@@ -43,11 +43,11 @@ function ExistingCommentEditor({
     commentId: comment.id,
   });
 
-  const handleSubmit = (inputValue: string) => {
+  const handleSubmit = async (inputValue: string) => {
     if (type === "comment") {
-      updateComment(comment.id, inputValue, (comment as Comment).position);
+      await updateComment(comment.id, inputValue, (comment as Comment).position);
     } else if (type === "reply") {
-      updateCommentReply(comment.id, inputValue);
+      await updateCommentReply(comment.id, inputValue);
     }
     commentsLocalStorage.clear();
     clearPendingComment();
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
@@ -33,20 +33,19 @@ function NewCommentEditor({ clearPendingComment, data, setModal }: NewCommentEdi
       : "video"
   );
 
-  const handleSubmit = (inputValue: string) => {
+  const handleSubmit = async (inputValue: string) => {
     if (!isAuthenticated) {
       setModal("login");
       return;
     }
 
     if (data.type == "new_reply") {
-      handleReplySave(data.comment, inputValue);
+      await handleReplySave(data.comment, inputValue);
     } else {
-      handleNewSave(data.comment, inputValue);
+      await handleNewSave(data.comment, inputValue);
     }
 
     commentsLocalStorage.clear();
-
     clearPendingComment();
   };
 
--- a/src/ui/components/Transcript/Transcript.tsx
+++ b/src/ui/components/Transcript/Transcript.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useMemo } from "react";
 import { useSelector } from "react-redux";
 import { selectors } from "ui/reducers";
 import sortBy from "lodash/sortBy";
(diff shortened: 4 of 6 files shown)


src/ui/components/Comments/TranscriptComments/CommentCard.tsx
```diff
@@ -53,7 +53,7 @@ function CommentItemHeader({
 
   useEffect(() => {
     setRelativeDate(formatRelativeTime(new Date(comment.createdAt)));
-  }, []);
+  }, [comment.createdAt]);
```
jasonLaster, line 56: linting

src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
```diff
@@ -33,20 +33,19 @@ function NewCommentEditor({ clearPendingComment, data, setModal }: NewCommentEdi
...
       return;
     }
 
     if (data.type == "new_reply") {
-      handleReplySave(data.comment, inputValue);
+      await handleReplySave(data.comment, inputValue);
     } else {
-      handleNewSave(data.comment, inputValue);
+      await handleNewSave(data.comment, inputValue);
     }
```
jasonLaster, line 46: we need to keep the pending comment around until we've created the initial comment

src/ui/components/Transcript/Transcript.tsx
```diff
@@ -7,27 +7,40 @@ import { Comment } from "ui/state/comments";
...
   const pendingComment = useSelector(selectors.getPendingComment);
   const { isAuthenticated } = useAuth0();
 
-  if (loading) {
-    return null;
-  }
+  const displayedComments = useMemo(() => {
+    const displayedComments: Comment[] = [...comments].filter(
+      comment => !pendingComment || commentKey(comment) != commentKey(pendingComment.comment)
+    );
```
jasonLaster, line 25: it's important to default to the pending comment as long as it is here

src/ui/components/Transcript/Transcript.tsx
```diff
@@ -7,27 +7,40 @@ import { Comment } from "ui/state/comments";
...
-  if (pendingComment?.type == "new_comment") {
-    displayedComments.push(pendingComment.comment);
-  }
+    const sortedComments = sortBy(displayedComments, ["time", "createdAt"]);
 
-  const sortedComments = sortBy(displayedComments, ["time", "createdAt"]);
-  const keys = commentKeys(sortedComments);
+    if (uniqBy(sortedComments, commentKey).length != sortedComments.length) {
+      console.warn("Duplicate comments detected");
+      debugger;
```
jasonLaster, line 35: this check helped me see where there were dupes so i kept it in to share the approach. It's safe to remove now

src/ui/components/Transcript/Transcript.tsx
```diff
@@ -7,27 +7,40 @@ import { Comment } from "ui/state/comments";
...
+    }
+
+    return sortedComments;
+  }, [comments, pendingComment]);
+
+  if (loading || auth.isLoading || recording.loading) {
+    return null;
+  }
```
jasonLaster, line 43: these loading states were causing several unecessary renders

src/ui/utils/comments.ts
```diff
@@ -37,12 +37,11 @@ export function formatRelativeTime(date: Date) {
...
-  const createdAt = orderedByApproximatelyCreatedAt(comments);
-  return comments.map((c, i) => commentKey(c, createdAt[i]));
+  return comments.map((c, i) => commentKey(c));
 }
 
-function commentKey(comment: Comment | Reply, createdAtOrder: number): string {
-  return compact([createdAtOrder + 1, commentIdentifiers(comment)]).join("-");
+export function commentKey(comment: Comment | Reply): string {
+  return compact([comment.createdAt, commentIdentifiers(comment)]).join("-");
 }
```
jasonLaster, line 45: I am pretty sure that these keys are more stable when you use the createdAt field directly which I confirmed was not causing flickering

src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
```diff
@@ -43,11 +43,11 @@ function ExistingCommentEditor({
...
   });
 
-  const handleSubmit = (inputValue: string) => {
+  const handleSubmit = async (inputValue: string) => {
     if (type === "comment") {
-      updateComment(comment.id, inputValue, (comment as Comment).position);
+      await updateComment(comment.id, inputValue, (comment as Comment).position);
     } else if (type === "reply") {
-      updateCommentReply(comment.id, inputValue);
+      await updateCommentReply(comment.id, inputValue);
```
abstractalgo, line 50: Do I correctly understand that we just synchronously wait? Thus bypassing React rerenders in the meantime, thus avoiding flicker?

src/ui/utils/comments.ts
jcmorrow: Hmmmmmm... `createdAt` should change when persisting to the server, which is why I didn't use it originally.

src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
jasonLaster: I'm actually a bit confused why we have the call to `clearPendingComment` below.  I added this change to be consistent with `NewCommentEditor`, but if we can remove the call to `clearPendingComment` it is probably not necessary.

src/ui/utils/comments.ts
jasonLaster: That's what I thought as well, but we actually don't update the Apollo representation when it comes back.

jasonLaster:
Next steps:
1. Update backend to support ID
2. Update CommentCard to disable Reply button
3.

jcmorrow (approved):
Looks like it mostly solves the "new comment" flickers, I still see some weird stuff with Replies, but it's a step in the right direction.

PR #6041, opened 2022-03-29
Refactor analysis points to include more precise error data

### LINKED ISSUE #5687: Investigate whether it's possible to tell if an analysis error is because it's 10k+, or because of another reason
```
jlast — Today at 3:08 PM
@Jaril (he/him) i think we're looking into bugs with the 10k hits
jlast — Today at 3:08 PM
Oh interesting InvalidObject is what we show when we fail to understand the object. CC @hbenl (he/him)
@Andarist is it possible to share this replay with us with comments at these spots?
Andarist — Today at 3:11 PM
invalid object stuff here
https://app.replay.io/recording/composed-conditional-type-expands-into-union-of-its-its-branches--3362459f-d68d-40af-997b-7f407643f642
10k+ hits are reported almost everywhere, can be seen on the same line
Jaril (he/him) — Today at 3:13 PM
I think we infer that something has 10k+ hits when the analysis returns an error. Which it might have, but it's also possible that any other number of other errors could have happened, and it's just being masked by the 10k+ label
Andarist — Today at 3:14 PM
might be it, i've refreshed and i don't experience this right now
aaaaand i have it again 😄
jlast — Today at 3:16 PM
@Jaril (he/him) shouldnt we ahve an error code for the 10K to differentiate
Andarist — Today at 3:16 PM
actually, it behaves slightly differently now - i dont see it everywhere now, but i have some lines on which this appears and im pretty sure that this info is invalid on those lines
oh, sorry - i was looking at it wrong, this time this actually might be reported correctly here as its a super generic helper, i was just htinking about its parent call which isnt called as much and the info is OK there 
```

jaril:
Fix #5687.

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
@@ -30,11 +30,11 @@ function BreakpointNavigation({
       seek(point.point, point.time, true);
     }
   };
-  const isEmpty = analysisPoints && (analysisPoints === "error" || analysisPoints?.length == 0);
+  const isEmpty = analysisPoints && (analysisPoints.error || analysisPoints.data.length == 0);
 
   let prev, next;
 
-  if (executionPoint && analysisPoints !== "error" && analysisPoints?.length > 0) {
+  if (executionPoint && analysisPoints.error && analysisPoints.data.length > 0) {
     prev = findLast(analysisPoints, p => compareNumericStrings(p.point, executionPoint) < 0);
     next = find(analysisPoints, p => compareNumericStrings(p.point, executionPoint) > 0);
   }
@@ -47,8 +47,8 @@ function BreakpointNavigation({
 
   useEffect(() => {
     if (analysisPoints) {
-      trackEvent(analysisPoints.length > 0 ? "breakpoint.has_hits" : "breakpoint.no_hits", {
-        hits: analysisPoints.length,
+      trackEvent(analysisPoints.data.length > 0 ? "breakpoint.has_hits" : "breakpoint.no_hits", {
+        hits: analysisPoints.data.length,
       });
     }
   }, [analysisPoints]);
@@ -58,7 +58,7 @@ function BreakpointNavigation({
       {!isEmpty ? (
         <BreakpointNavigationCommands prev={prev} next={next} navigateToPoint={navigateToPoint} />
       ) : null}
-      {analysisPoints !== "error" ? (
+      {analysisPoints.error ? (
         <BreakpointTimeline breakpoint={breakpoint} setZoomedBreakpoint={setZoomedBreakpoint} />
       ) : (
         <div className="flex-grow" />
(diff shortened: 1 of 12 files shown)


src/ui/state/app.ts
```diff
@@ -100,8 +100,14 @@ export interface AppState {
...
 
-export interface AnalysisPoints {
-  [key: string]: PointDescription[] | "error";
+export type AnalysisPoints = Record<string, AnalysisPayload>;
+export type AnalysisPayload = {
+  data: PointDescription[];
+  error: AnalysisError | null;
+};
+export enum AnalysisError {
+  HellaHits = "hella-hits",
```
jcmorrow, line 109: 🤨

src/ui/state/app.ts
```diff
@@ -126,3 +132,7 @@ export interface Canvas {
   top: number;
   width: number;
 }
+
+export enum ProtocolError {
+  TooManyPoints = 55,
```
jcmorrow, line 137: We don't export this type from protocol huh?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
```diff
@@ -58,7 +58,7 @@ function BreakpointNavigation({
       {!isEmpty ? (
         <BreakpointNavigationCommands prev={prev} next={next} navigateToPoint={navigateToPoint} />
       ) : null}
-      {analysisPoints !== "error" ? (
+      {analysisPoints.error ? (
```
hbenl, line 61: Should be `!analysisPoints.error`

src/protocol/logpoint.ts
```diff
@@ -253,8 +250,8 @@ async function setMultiSourceLogpoint(
   if (showInConsole && primitiveFronts) {
     const points = getAnalysisPointsForLocation(store.getState(), locations[0], condition);
     if (points) {
-      if (points !== "error") {
-        showPrimitiveLogpoints(logGroupId, points, primitiveFronts);
+      if (points.error) {
```
hbenl, line 253: should be `!points.error`

src/ui/reducers/app.ts
```diff
@@ -237,28 +264,16 @@ export const getAnalysisPointsForLocation = (
...
   const location = getHoveredLineNumberLocation(state);
-  const points = getAnalysisPointsForLocation(state, location);
-  const focusRegion = getFocusRegion(state);
-
-  if (focusRegion && points && points !== "error") {
-    return points.filter(p => isInTrimSpan(p.time, focusRegion));
-  }
-
-  return points;
+  return getAnalysisPointsForLocation(state, location);
```
hbenl, line 276: no `getPointsInTrimSpan()`?

hbenl (approved):
Found 2 missing negations and a possible oversight. Otherwise LGTM.

src/ui/reducers/app.ts
jaril: It's applied inside `getAnalysisPointsForLocation` so no need here. Let me know if I missed something!

PR #6042, opened 2022-03-29
Correct env for e2e on master (take two)

abstractalgo:


hbenl (approved):

abstractalgo:
Spend some time debugging the GHA action but couldn't find obvious wrong things with it in that time. https://github.com/RecordReplay/devtools/pull/5668 is worth resurrecting.

Issue #4578, opened 2021-11-19
Elements Panel steals keyboard arrow events from the comments pane


jasonLaster:
https://app.replay.io/recording/44662e48-c367-49ff-8780-1a1f8c362d9d

jasonLaster:
Left a bunch of comments - 

1. feels related to this PR     `Call preventDefault on cursor key events in the markup panel` (#4423)
2. i dont trust our breakpoint locations there are a bunch of spots in KeyShortcuts that feel like they should be hit

hbenl:
It's not the Elements panel that's doing the stealing but the ReactDevTools panel, see https://app.replay.io/recording/5895bcc2-c7d2-47c5-8456-597198e3993d
We probably didn't notice this before because the global keydown event listener is only registered when ReactDevTools is rendered and in the past we only rendered it when it was visible. But since #4537 we're always rendering it (even if it's disabled in the user's settings).
We could fix this by asking the ReactDevTools developers to not register their event listener on the document node or by putting ReactDevTools in an iframe.

jasonLaster:
Thanks. It sounds like scoping the event listener is a pretty good short-term fix, but we might have other collisions in the future as well.

jasonLaster:
It looks like [views/DevTools.js](https://github.com/facebook/react/blob/main/packages/react-devtools-shared/src/devtools/views/DevTools.js#L239) is adding a global keydown handler. We can either fix this by running React DevTools in an iframe or scoping this handler to React DevTools root node.

cc @bvaughn

jcmorrow:
Here's a comment in React Devtools mentioning the issue, might be a good intervention point: https://github.com/facebook/react/blob/8b8817845a8efd8cf53be395fafbb1d0f62c4368/packages/react-devtools-shared/src/devtools/views/Components/Tree.js#L129-L131

jasonLaster:
This feels more like an important issue than a project to me. Glad we're now triaging these in the weekly frontend meeting

bvaughn:
Happy to lend a hand here in a week or two (if not sooner) :)

jasonLaster:
@bvaughn looks good!

bvaughn:
Quick update that [4.24.3](https://github.com/facebook/react/blob/main/packages/react-devtools/CHANGELOG.md#4243) was published to NPM today with my fix for this.

bvaughn:
[...]

PR #6075, opened 2022-03-30
Update react-devtools-inline, react, and react-dom dependencies

### LINKED ISSUE #5344: Bump React DevTools
Bump React DevTools for frontend and backend.

Because of bridge protocol version mismatches between the latest version and what we have now (v2 vs v1), this would likely invalidate previous recordings, so this is not without a risk.

This will require a ticket for the backend/gecko as well.

bvaughn:
Resolves #5344 and #4578; relates to https://github.com/RecordReplay/gecko-dev/pull/796

This `react-devtools-inline` update introduces a newer [bridge protocol version](https://github.com/facebook/react/pull/21331) than what is currently being embedded in the backend. That means older Replay sessions containing React data won't work with it.

I think the best short-term solution for this is for the devtools/client to check the protocol version embedded in the Replay, and load the `react-devtools-inline` version that supports it. This is a little hacky but it takes the pressure off of our ability to upgrade React DevTools without breaking backwards compat.

# Checklist
- [x] Update `react-devtools-inline` to latest (4.24.3).
- [x] Update backend `react-devtools-inline` as well: https://github.com/RecordReplay/gecko-dev/pull/796
- [x] Update frontend to check protocol and dynamically load the React DevTools package that supports that version.
- [x] Fix TypeScript types
- [x] Add inline comments explaining what/why
- [ ] Maybe add tests (once we have shared Replay sessions with the new protocol version?)


jasonLaster:
@bvaughn makes sense.

Here are the [docs](https://github.com/RecordReplay/gecko-dev/tree/webreplay-release/devtools/server/actors/replay/react-devtools) on updating the `react_devtools_backend`. Updating the backend is sufficient for being able to land this PR and view the latest replays. 

re backward compatibility: agreed. If the frontend knows the version we should be able to load it. There are some funny hoops to jump through because the `react-devtools-inline` is currently in our package.json, so i'm not too sure how we'd get multiple versions, but that feels more of an opsy challenge. Related, we don't currently have backward compat on the q2 roadmap because we assumed Component Stacks and some other features took precedence, but that's all flexible.

hbenl:
> Perhaps a solution to this could be to change our backend profile to include metadata about which backend version was used to record the ["operations" message](https://github.com/facebook/react/blob/main/packages/react-devtools/OVERVIEW.md#serializing-the-tree).

FYI: we're already fetching the ReactDevTools protocol version by sending the `getBridgeProtocol` message to the backend in the recording:
https://github.com/RecordReplay/devtools/blob/b36e7f0e2188a1a8788d92a4b8e1b01d93131b22/src/ui/components/SecondaryToolbox/ReactDevTools.tsx#L82-L83

bvaughn:
@hbenl and I tested and verified that the changes in this PR and https://github.com/RecordReplay/gecko-dev/pull/796 are compatible. The remaining trouble is that– if we were to land these PRs, old Replays would "break" due to the React DevTools protocol version.

I'm going to see if I can't figure out a workaround to make the _newer_ DevTools frontend support the _older_ pre-recorded profiles.

bvaughn:
I think the short term path forward here is to embed _both_ versions of the `react-devtools-inline` package (4.17 which supports the older bridge protocol version and 4.24 which supports the newer one). Then we can look at the Replay and load the right frontend version based on the backend protocol that was captured.

This is kind of hacky but it unblocks us to update React DevTools.

src/ui/components/SecondaryToolbox/react-devtools-inline.d.ts
```diff
@@ -187,3 +187,7 @@ https: declare module "react-devtools-inline/frontend" {
     bridgeAndStore: { bridge?: FrontentBridge; store?: Store }
   ): typeof ReactDevTools;
 }
+
+https: declare module "react-devtools-inline_4_17_0/frontend" {
+  export * from "@types/react-devtools-inline/frontend";
```
bvaughn, line 192: I am fairly new to TypeScript. Is this a valid way to tell TypeScript that these two modules have the same API?

hbenl (changes requested):
This works for me with recordings made with both the old and the new backend. The only issue is that the ReactDevTools UI is not updated when jumping to a different point in time.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -229,6 +239,54 @@ function createReactDevTools(
...
+    asyncIndex: 0,
+    text: ` __RECORD_REPLAY_REACT_DEVTOOLS_SEND_MESSAGE__("getBridgeProtocol", undefined)`,
+  });
+  const json: any = await response?.returned?.getJSON();
+  const version = json?.data?.version || 0;
+
+  let ReactDevTools = null;
+  if (version >= 2) {
+    const reactDevToolsInlineModule = await import("react-devtools-inline/frontend");
+    ReactDevTools = await createReactDevTools(
```
hbenl, line 267: The `createReactDevTools()` call could be deduplicated by moving it out of the if/then/else block

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -238,6 +296,34 @@ function ReactDevtoolsPanel({
...
+  const initializationRef = useRef<boolean>(false);
+
+  // Lazy-load the right version of react-devtools-inline based on embedded metadata.
+  // Eventually this lazy-loading code could be done with Suspense, but for now use an effect.
+  //
+  // TODO This lint disabling shouldn't be necessary; omitting the dependencies array (in favor of a ref) is valid.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
+    if (initializationRef.current) {
+      // Only initialize the DevTools UI once, once conditions have been met.
```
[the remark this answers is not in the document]
bvaughn: > We need to initialize the DevTools UI every time currentPoint changes, otherwise the React panel will not be updated when the user jumps to a different point in the timeline.

Ah, if we are re-initializing it every time the point changes then I think I have some substantial gap in my knowledge about how React DevTools works within Replay vs in a regular browser context. (That sounds really expensive. Do we have to re-initialize?)

> For performance reasons we should then also separate requesting the protocol version from the initialization of the DevTools UI so that we don't re-request the version every time the user jumps to a different point.

Yes. This makes sense, given the above.

Thanks for sharing this info. Let me take another pass at this given this new knowledge.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -17,6 +17,13 @@ import { setHasReactComponents, setProtocolCheckFailed } from "ui/actions/reactD
 import Highlighter from "highlighter/highlighter";
 import NodePicker, { NodePickerOpts } from "ui/utils/nodePicker";
 import { sendTelemetryEvent, trackEvent } from "ui/utils/telemetry";
+import { consoleOverflowed } from "devtools/client/webconsole/reducers/messages";
```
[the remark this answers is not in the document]
bvaughn: Ah, thanks. I think VSCode must have auto-imported this and I didn't even notice it.

As part of starting this new job, I'm trying to make the update from Sublime to VSCode and it's a bit confusing to me.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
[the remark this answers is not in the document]
bvaughn: Re-reading the code, and I guess now I understand why Replay was doing a "bad thing" before and initializing a component during render:
https://github.com/RecordReplay/devtools/blob/ce03f35d4f536f24a3b9cb5200de9189bf4abfa0/src/ui/components/SecondaryToolbox/ReactDevTools.tsx#L286-L307

I thought this was a _mistake_ rather than intentional.

I would love to chat some about the constraints here to see if there's a way we could avoid recreating DevTools entirely (b'c of the cost of this) but for now I'll refactor my approach here. :)

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -207,12 +216,17 @@ class ReplayWall implements Wall {
...
   annotations: Annotation[],
   currentPoint: ExecutionPoint,
   enablePicker: (opts: NodePickerOpts) => void,
   disablePicker: () => void,
   onShutdown: () => void
-) {
+): ReactDevToolsType {
+  const createBridge: typeof createBridgeType = reactDevToolsInlineModule.createBridge;
+  const createStore: typeof createStoreType = reactDevToolsInlineModule.createStore;
+  const initialize: typeof initializeType = reactDevToolsInlineModule.initialize;
```
bvaughn, line 228: TypeScript seems weird to me. When I used `import type` I didn't have to also use `typeof` but it seems that makes it happy?

bvaughn:
Okay. I think things should work more now like they used to, except for the async module-loading hop. Sorry for not catching the jumping-between-points issue during my smoke testing. I assumed that if DevTools loaded _once_ it would work as usual from there on out– because of my mental model of how it runs in the browser under normal conditions.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
hbenl: Interestingly, even though you're using `import type`, `createBridgeType` etc. refer to the functions, not to their types. I'm surprised that Typescript doesn't complain about using `import type` for something that isn't a type but I guess it makes sense when you want to import a function's (or object's) type without creating a runtime dependency and the exporting module doesn't export the types separately.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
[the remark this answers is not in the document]
hbenl: The reason is simple: we need to feed all recorded backend messages from the beginning of the recording to the current point to the frontend. Then, when you jump to an *earlier* point in time, we'd have to *remove* some of these messages, which isn't supported by the frontend. If the frontend had a way to reset it (so that it forgets about all the messages we already fed to it) then we wouldn't need to recreate it.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
bvaughn: Interesting. Is there a better way to type these functions?

I guess I could just `declare createBridgeType` in the `.d.ts` module

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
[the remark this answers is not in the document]
bvaughn: Yeah, that makes sense (at least when stepping to an earlier point).

> If the frontend had a way to reset it (so that it forgets about all the messages we already fed to it) then we wouldn't need to recreate it.

We _could_ do that 🙂 It would be a little hacky, but so is what we're already doing.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
hbenl: You could add `type createBridgeType = typeof createBridge;` to the `.d.ts` file.
But an even easier way would be to write `reactDevToolsInlineModule: typeof import("react-devtools-inline/frontend")` and then remove the explicit types from the functions that you extract from `reactDevToolsInlineModule`.

bvaughn: Ahhhh, _this_ (`reactDevToolsInlineModule: typeof import("react-devtools-inline/frontend")`) is what I was looking for but didn't know how to do.

PR #6078, opened 2022-03-30
Fallback to analyses for old recordings

jcmorrow:


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -119,8 +120,9 @@ export default function LineNumberTooltip({
     }
     setTimeout(() => {
       if (lineNumber === lastHoveredLineNumber.current) {
-        if (features.codeHeatMaps) {
-          dispatch(setBreakpointHitCounts(source.id));
+        if (codeHeatMaps) {
+          dispatch(setBreakpointHitCounts(source.id, () => setCodeHeatMaps(false)));
+          dispatch(runAnalysisOnLine(lineNumber));
```
jasonLaster, line 125: I think this is here for the perf win, if HeatMaps fails, but it re-adds the run analysis message which will slow the backend down all the time. It's probably alright to remove this because if HeatMaps fails we'll kick this off anyways

src/devtools/client/debugger/src/actions/source-actors.js
```diff
@@ -88,11 +88,14 @@ export const loadSourceActorBreakpointHitCounts = memoizeableAction(
       ]).join("-");
     },
     getValue: ({ id }, { getState }) => getSourceActorBreakpointHitCounts(getState(), id),
-    action: async ({ id }, { dispatch, getState, client }) => {
+    action: async ({ id, onFailure }, { dispatch, getState, client }) => {
       await dispatch({
```
jasonLaster, line 92: You probably dont need to pass `onFailure` in. Typically using `.catch` is nicer.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -119,8 +120,9 @@ export default function LineNumberTooltip({
     }
     setTimeout(() => {
       if (lineNumber === lastHoveredLineNumber.current) {
-        if (features.codeHeatMaps) {
-          dispatch(setBreakpointHitCounts(source.id));
+        if (codeHeatMaps) {
+          dispatch(setBreakpointHitCounts(source.id, () => setCodeHeatMaps(false)));
```
jasonLaster, line 124: we could do `dispatch(setBreakpointHitCounts(source.id).catch` here

jcmorrow: in my experience something closer to the protocol was already catching, so the catch here was not getting hit

PR #6079, opened 2022-03-30
Change default flag for codeHeatMaps

jcmorrow:


src/ui/utils/prefs.ts
```diff
@@ -27,7 +27,6 @@ pref("devtools.features.networkRequestComments", true);
 pref("devtools.features.useMultipleControllers", false);
 pref("devtools.features.multipleControllerUseSnapshots", false);
 pref("devtools.features.breakpointPanelAutocomplete", true);
-pref("devtools.features.codeHeatMaps", false);
```
jasonLaster, line 30: In this change, I'd probably keep the flag, but just default it to `true`. This way - we can still disable it down the line if we find that it is bad.

src/ui/components/shared/UserSettingsModal/ExperimentalSettings.tsx
```diff
@@ -50,11 +50,6 @@ const EXPERIMENTAL_SETTINGS: ExperimentalSetting[] = [
     key: "multipleControllerUseSnapshots",
     secret: true,
   },
-  {
-    label: "Code Heatmaps ",
-    description: "Calculate hit counts for editor files all at once",
-    key: "codeHeatMaps",
-  },
```
jasonLaster, line 57: lets keep this here for a bit. There might be some inconsistencies where we can recommend to users that they disable it on a one off basis

Issue #5850, opened 2022-03-17
Add an e2e test for video playback


jasonLaster:
We should write a simple test that asserts that when you click the play button the video plays.

A simple assertion would be to check `window.currentScreenshotHash` every 50ms and assert that there are atleast N unique values when the video finishes.

gideonred:
This ticket has been moved to [FE-133](https://linear.app/replay/issue/FE-133/add-an-e2e-test-for-video-playback)

PR #6095, opened 2022-03-31
Add e2e test for video playback (#5850)

hbenl:
This test is currently unreliable due to performance issues: it tries to look at the hash of the currently displayed screenshot every 50ms, but the 50ms delay can actually take several seconds as can be seen in this recording: http://localhost:8080/recording/replay-of-localhost8080--b2d8651f-962d-44ed-917c-0845043b0b88


jaril:
@hbenl should we close this for now until we get playwrighted?

jaril:
Closing this for now — @hbenl lmk if that's the wrong call!

Issue #6098, opened 2022-03-31
URL parameters not updated when navigating


hbenl:
Using the stepping buttons in the debugger panel doesn't update the `point`/`time`/`hasFrames` URL parameters to their new values.

PR #6099, opened 2022-03-31
Always update the URL when seeking to a new point (#6098)

hbenl:


jasonLaster (approved):
I like this change, but curious where it came from

PR #6110, opened 2022-03-31
Allow user to use global text search but include/exclude node_modules

### LINKED ISSUE #5418: Support searching in excluded files
The full text search feature is confusing when trying to search in sources that are excluded from the search.

STR:

1. Visit https://app.replay.io/recording/6d4337b7-95d5-4a8a-a795-c85b0b497718
2. Search in all files for "onTestFileStart" or anything else which appears in the recording's javascript.

Actual results: No results are returned.

The problem here is that the devtools are excluding almost all of the sources in this recording from searching, I think because they are in a node_modules folder. The devtools should support searching in all the recording's sources (maybe using a checkbox), or at least indicate that it isn't actually going to search through all the sources.

jaril:
Fix #5418.

[Demo](https://user-images.githubusercontent.com/15959269/161144979-e188630c-96c5-4367-8dc9-2df959e3f602.mov) 

Light | Dark
--- | ---
image | image

### DIFF
--- a/src/devtools/client/debugger/src/components/FullTextSearch/index.js
+++ b/src/devtools/client/debugger/src/components/FullTextSearch/index.js
@@ -17,6 +17,7 @@ import {
 import { trackEvent } from "ui/utils/telemetry";
 import { FullTextFilter } from "./FullTextFilter";
 import { FullTextResults } from "./FullTextResults";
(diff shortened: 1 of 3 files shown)


jaril:
@ceceliacreates this'll be a new docs thing 🙋‍♂️

PR #6115, opened 2022-03-31
Don't try to parse content if it is an object

jcmorrow:
We are trying to parse `content` even when it came from Redux, but when it comes from Redux it is already an object, not a string, in which case we can hand it directly to TipTap. With @jasonLaster's change to persist the `pendingComment` slice of redux this means we get comment persistence across switching modes and across tab refreshes, sessions (on the same device), etc.

### DIFF
--- a/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
+++ b/src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
@@ -1,11 +1,12 @@
 import React from "react";
-import { connect, ConnectedProps } from "react-redux";
+import { connect, ConnectedProps, useDispatch } from "react-redux";
 
 import { Comment, Reply } from "ui/state/comments";
-import CommentEditor, { PERSIST_COMM_DEBOUNCE_DELAY } from "./CommentEditor";
+import CommentEditor, { PERSIST_COMMENT_DEBOUNCE_DELAY } from "./CommentEditor";
 import { useCommentsLocalStorage } from "./useCommentsLocalStorage";
 import debounce from "lodash/debounce";
 import { CommentData } from "../types";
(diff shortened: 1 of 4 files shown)


src/ui/components/Comments/TranscriptComments/CommentEditor/TipTapEditor.tsx
```diff
@@ -90,7 +94,7 @@ const TipTapEditor = ({
         }),
       ],
       editorProps: { attributes: { class: "focus:outline-none" } },
-      content: tryToParse(content),
+      content: typeof content === "string" ? tryToParse(content) : content,
```
jasonLaster, line 97: this feels like something that might live in `tryToParse`

`content: tryToParse(content)` could just return if it's an object

src/ui/components/Comments/TranscriptComments/CommentEditor/NewCommentEditor.tsx
```diff
@@ -21,22 +22,16 @@ function NewCommentEditor({ onSubmit, data }: NewCommentEditorProps) {
...
-      onCreate={({ editor }) => {
-        const storedComment = commentsLocalStorage.get();
-        editor.commands.setContent(storedComment ? JSON.parse(storedComment) : null);
-      }}
+      handleSubmit={inputValue => onSubmit(data, inputValue)}
       onUpdate={debounce(({ editor }) => {
-        commentsLocalStorage.set(JSON.stringify(editor.getJSON()));
-      }, PERSIST_COMM_DEBOUNCE_DELAY)}
+        dispatch(updatePendingCommentContent(editor.getJSON()));
+      }, PERSIST_COMMENT_DEBOUNCE_DELAY)}
```
jasonLaster, line 34: Nice catch

jasonLaster (approved):
Looks good!

PR #6116, opened 2022-03-31
Save pending comments to replay sessions

jasonLaster:
Having tested this, but wanted to throw it up to get some eyeballs.

**Context**
At the moment we have 3 stores (localStorage, redux, apollo). Redux is the source of truth for pending comments. This is what we will use when you want to show the same pending comment when you switch from viewer to devtools. We also have an established pattern for saving redux state to localStorage and indexdb. This uses that so when the user comes back the same pending comments will be shown.

**How is it different?**
The old approach had some special logic around storing multiple comments with different values. This doesnt do that. When we want to do that we can add these rules to redux and get that for free.


pages/_app.tsx
```diff
@@ -155,8 +155,9 @@ function Routing({ Component, pageProps }: AppProps) {
   const [store, setStore] = useState<Store | null>(null);
   const { getFeatureFlag } = useLaunchDarkly();
 
+  debugger;
   useEffect(() => {
-    bootstrapApp().then((store: Store) => setStore(store));
+    bootstrapApp("sdf").then((store: Store) => setStore(store));
```
jcmorrow, line 160: sdf

src/ui/components/Comments/TranscriptComments/CommentEditor/ExistingCommentEditor.tsx
```diff
@@ -58,21 +52,10 @@ function ExistingCommentEditor({
...
         }}
-        onCreate={({ editor }) => {
-          const storedComment = commentsLocalStorage.get();
-          if (storedComment) {
-            editor.commands.setContent(JSON.parse(storedComment));
-          }
-        }}
-        onUpdate={debounce(({ editor }) => {
-          commentsLocalStorage.set(JSON.stringify(editor.getJSON()));
-        }, PERSIST_COMM_DEBOUNCE_DELAY)}
```
jcmorrow, line 73: Just FYI this is definitely going to merge conflict with me because I'm using this place to update Redux, just a note for whichever of us merges second.

src/ui/setup/prefs.ts
```diff
@@ -159,7 +160,13 @@ async function maybeUpdateReplaySessions(state: UIState) {
     selectedPanel: getSelectedPanel(state),
     localNags: [...(previousReplaySession?.localNags || [])],
     tabs: persistTabs(getTabs(state)) || [],
+    pendingComment: selectors.getPendingComment(state),
   };
 
   asyncStore.replaySessions = { ...previousReplaySessions, [recordingId]: currentReplaySession };
 }
+
+export async function getReplaySession(recordingId: string) {
```
jcmorrow, line 169: we could use this in a few other places as well.

jcmorrow (approved):

src/ui/setup/prefs.ts
jasonLaster: I bet

jasonLaster:
The missing piece is getting the recordingId into the bootstrapStore call. Should be fine since it's coming from the router, but something that i couldnt do with minimal internet

PR #6120, opened 2022-04-01
Refactor our view modes

jaril:
We currently have a big rendering job that happens when you toggle between VIewer and DevTools because there's a lot of shared components that end up being re-rendered. This was mostly due to janky component structuring.

This minimizes the mounting / unmounting required when switching between those two modes.


src/devtools/client/debugger/src/components/App.js
```diff
@@ -165,39 +165,8 @@ class Debugger extends Component {
...
-    const { toolboxLayout } = this.props;
-
-    if (toolboxLayout == "ide") {
-      return <EditorPane toolboxLayout={toolboxLayout} />;
-    }
-
-    return null;
-  }
-
   renderLayout = () => {
```
jcmorrow, line 168: What do you think about inlining this now that it's only one line?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Breakpoint.js
```diff
@@ -11,11 +11,25 @@ import { connect } from "devtools/client/debugger/src/utils/connect";
...
+    <div
+      className={classnames("editor new-breakpoint", {
+        "folding-enabled": features.codeFolding,
+      })}
+      onContextMenu={onContextMenu}
+      onClick={onClick}
+      onMouseEnter={onMouseEnter}
+      onMouseLeave={onMouseLeave}
+    >
+      <svg viewBox="0 0 60 15" width="60" height="15">
```
jcmorrow, line 27: What do you think about pulling this into a little named component?

src/devtools/client/debugger/src/components/SecondaryPanes/CommandBar.js
```diff
@@ -70,28 +69,6 @@ function formatKey(action) {
 }
 
 class CommandBar extends Component {
-  componentWillUnmount() {
```
jcmorrow, line 73: Do we just not need this anymore or did it move somewhere else?

src/ui/actions/layout.ts
```diff
@@ -71,13 +70,6 @@ export function setViewMode(viewMode: ViewMode): UIThunkAction {
     //Update viewToggleMode to start toggle animation
     dispatch(setViewToggleMode(viewMode));
 
-    // Delay updating the viewMode in redux so that the toggle can fully animate
-    // before re-rendering all of devtools in the new viewMode.
-    const delayPromise = new Promise<void>(resolve => {
-      setTimeout(() => resolve(), TOGGLE_DELAY);
-    });
```
jcmorrow, line 78: I am so happy to see this ✂️

src/ui/components/Transcript/Transcript.css
```diff
@@ -0,0 +1,27 @@
+.sidebar {
```
jcmorrow, line 1: How would you feel about using a CSS module here? I can't tell if we've decided that we're actually trying to go that direction for new style files or not?

src/ui/components/Transcript/Transcript.css
```diff
@@ -0,0 +1,27 @@
...
+
+.sidebar-toolbar-item {
+  font-size: 15px;
+  line-height: 19px;
+  font-weight: 400;
+  color: var(--body-color);
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  padding: 8px;
```
jcmorrow, line 26: It's totally up to you, but I find that sorting in styling (when it won't change behavior) can be really nice. It tends to group together like things (in this case the two font properties) and it makes it easier to see when you might accidentally have two conflicting properties.

jcmorrow (approved):
Champion!

src/devtools/client/debugger/src/components/Editor/Breakpoints/Breakpoint.js
jasonLaster: I think this was this way for a reason, but I don't remember :)

It's probably safe to do this, but it could also cause some performance issues. For historical reasons, why are you changing this here and is it possible to do this in it's own commit so it's easier to revert if needed or compare quickly

jasonLaster (commented):
This is awesome!

jasonLaster:
Adding a breakpoint failed, so there's probably some more testing that's needed there

src/devtools/client/debugger/src/components/Editor/Breakpoints/Breakpoint.js
jaril: I was running into a problem earlier where this file was being rendered serverside where the window was unavailable. We don't have that problem anymore so I could just undo it.

src/ui/components/Transcript/Transcript.css
jaril: No strong opinion — would have inlined it but this was faster.

src/devtools/client/debugger/src/components/SecondaryPanes/CommandBar.js
jaril: This was running into the same serverside rendering problem bug earlier. I didn't feel particularly strongly about keeping step debugging shortcuts so I was quick to hit delete. I think I can bring it back and we can decide what to do with it when we finally undertake the big keyboard shortcuts discussion.

jasonLaster (commented):
I think this is somehow missing here and causing the border to show up in the source explorer
```
#toolbox-content-debugger .tree-indent {
  width: 16px;
  margin-inline-start: 0;
  border-inline-start: 0;
}
```

PR #6129, opened 2022-04-02
Handle unauthenticated errors from the graphql API

ryanjduffy:
V2 of #6125 

This adds an `!isAuthenticated` check to prevent try to refresh the token when there was no token to begin with.

The error condition (there are probably others) was opening a public replay anonymously and using the rewind button in the console to navigate backwards. This triggered `dismissNag` which fails because the user isn't authenticated.


jcmorrow (approved):
It is weird to me that we will *try* to hit GraphQL when we know we are not authenticated. I would expect us to intercept that codepath earlier, before getting back an unauthenticated response. Anyways, that's not really relevant here, because it still makes sense to have this `unauthenticated` logic in place, in case we try and do that elsewhere as well 👍

PR #6161, opened 2022-04-04
Add Next CLI installation step to README

bvaughn:
Looks like this step was missing from the setup steps.


bvaughn:
This step _may not actually be necessary_ as it appears I had forgotten to copy `.env.sample .env` on this laptop!

PR #6162, opened 2022-04-04
Add a hitCount line limit

jcmorrow:
Right now we have a very naive algorithm: fetch hitCounts for all lines when a source file gets hovered for the first time. This can swamp a controller if a file is tens of thousands of lines long, but it's *very* useful in the case that a file is of a reasonable length. In my experience, 1000 line fetches at once still feels pretty instantaneous, and I think many files that people are interested in having analyzed completely will be under 1000 lines. The hardest part of this change was just understanding how `memoizableAction` works for the third time 🙃 

In case the above did not make sense, here's an example:

- The user loads a file
- SourceOutline sees that a new file is loaded and asks for the hitCount on line 1
- memoizableAction will see that the line requests was 1, and actually ask for the hits for lines 1-1000.
- If the user scrolls to line 1005 (for example) and hovers, memoizableAction will fetch the hits for lines 1000-2000, etc.

This does not totally resolve https://github.com/RecordReplay/backend/issues/5132 and https://github.com/RecordReplay/backend/issues/5108, but it helps a *lot*.

### DIFF
--- a/src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
+++ b/src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
@@ -47,7 +47,8 @@ export function SourceOutline({
 
   useEffect(() => {
     if (selectedSource) {
-      dispatch(setBreakpointHitCounts(selectedSource.id));
+      // We start by loading the first N lines of hits, where N is the line limit.
+      dispatch(setBreakpointHitCounts(selectedSource.id, 1));
     }
   }, [selectedSource?.id]);
 
--- a/src/devtools/client/debugger/src/reducers/source-actors.js
+++ b/src/devtools/client/debugger/src/reducers/source-actors.js
@@ -116,7 +116,25 @@ function updateBreakpointHitCounts(state, action) {
     return state;
   }
 
-  return updateResources(state, [{ id: sourceId, breakpointHitCounts: action.value.hits }]);
+  const {
+    min: currentMin,
+    max: currentMax,
+    breakpointHitCounts: currentBreakpointHitCounts,
+  } = {
+    min: Infinity,
+    max: 0,
+    breakpointHitCounts: [],
+    ...state.values[sourceId],
+  };
+
+  return updateResources(state, [
+    {
+      id: sourceId,
+      breakpointHitCounts: [...currentBreakpointHitCounts, ...action.value.hits],
+      min: Math.min(currentMin, action.value.min),
+      max: Math.max(currentMax, action.value.max),
+    },
+  ]);
 }
 
 function clearBreakpointHitCounts(state) {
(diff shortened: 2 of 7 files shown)


src/devtools/client/debugger/src/reducers/source-actors.js
```diff
@@ -198,8 +216,11 @@ export function getAllThreadsBySource(state) {
   return queryThreadsBySourceObject(state.sourceActors);
 }
 
-export function getSourceActorBreakpointHitCounts(state, id) {
-  const { breakpointHitCounts } = getResource(state.sourceActors, id);
+export function getSourceActorBreakpointHitCounts(state, id, lineNumber) {
+  const { breakpointHitCounts, min, max } = getResource(state.sourceActors, id);
+  if (lineNumber && (!breakpointHitCounts || max < lineNumber || min > lineNumber)) {
```
jaril, line 221: Is it possible for `lineNumber` to be falsy?

src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
```diff
@@ -47,7 +47,8 @@ export function SourceOutline({
 
   useEffect(() => {
     if (selectedSource) {
-      dispatch(setBreakpointHitCounts(selectedSource.id));
+      // We start by loading the first N lines of hits, where N is the line limit.
+      dispatch(setBreakpointHitCounts(selectedSource.id, 1));
```
jaril, line 51: Does this mean that the hits in the Outline view will only reflect hits in lines 1-500?

jaril: And it reads like it will update to be a more accurate number, but only if the user hovers on other lines in the editor (which forces the load of hits in other lines). Did I understand that correctly?

jcmorrow: So, we just talked about this at backend standup, and it brings up a great point. Here's a further improvement:

- As soon as a file loads, grab the locations that are outside of function definitions (and so ran as soon as the file loaded)
- Also grab the first location (what we use for outline view) inside of all functions and ask for the hits at each of those locations
- If any of those hits come back as 0, we don't need to fetch any further hitCounts for that function, because no later location can be non-0 if the first location is 0
- Theoretically, you could do this for each level of scope/function nesting, but that's probably a bridge farther than we need to go to for this version.

I think I have a good follow-up here, which would be to differentiate between 0 and null for the outline view, which might be less misleading. I'll do that first, then follow-up with the above points.

src/devtools/client/debugger/src/reducers/source-actors.js
jcmorrow: I think it *was* when I started writing this and now it's not. Let me double check that.

jcmorrow: Removed!

src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
jcmorrow: OK, so I've made it so that if we aren't loading the line that that function definition falls on it will *not* have a hitCount marker. As soon as we load it we will show the hitCounts (but you have to go to that line and hover there). I am happy with this compromise because:

A) I think most of the time people will not be using these features on 1000+ line files
B) If they *are* doing that, there is still a way for them to get the info, it's just not perfect
C) All of this is behind a feature flag 😄

PR #6164, opened 2022-04-04
Convert "resources" abstraction logic to TS

markerikson:
This PR:

- Converts all of the logic in `utils/resources` to TS
- Adds my initial notes on the sources reducer file

Since the resources code is heavily used by the sources logic, and frankly it's incredibly abstract and hard to follow, converting this code to TS first seemed like a foundational step.

All code changes are _just_ adding TS types.


docs/codebase-notes/sources-reducer.md
```diff
@@ -0,0 +1,251 @@
...
+
+Just sets `state.focusedItem`.
+
+Dispatched by: `components/PrimaryPanes/SourcesTree.js`, line 134 when a sources tree item is focused
+
+### `"SOURCES_LOADED"`
+
+Just sets `state.sourcesLoading = false`.
+
+Dispatched by: `debugger/src/client/index.js` line 44, after waiting for `ThreadFront.findSources()` to complete
```
jasonLaster, line 128: These docs are great, but I worry that they might be hard to maintain. Would it make sense to start adding more inline documentation and use a docgen tool internally?

Fine to add these now and then come back later

jasonLaster (approved):
Looks good. This code is the most abstract code we have in the codebase

docs/codebase-notes/sources-reducer.md
markerikson: These aren't meant to be permanent. These are my scratch notes as I'm just trying to sort through what this codebase is doing in the first place. In a sense, them having value to anyone else is an incidental bonus.  

But, the info _is_ potentially useful in the near term, so it's worth committing these instead of me just keeping them locally on my own machine.

PR #6165, opened 2022-04-04
Switch from pushState to replaceState

### LINKED ISSUE #6050: Browser forward/back buttons don't do anything
The forward/back buttons update the URL to change the current point/time, but the UI doesn't update to match.

jaril:
Fix #6050.

This keeps the browser back/forward buttons working [as I expect it should](https://github.com/RecordReplay/devtools/issues/6050#issuecomment-1087932342), i.e. pressing back brings you to the previous page you were on, instead of moving back to a previous pause position in the same replay.


jasonLaster (approved):
Smart. Best of both worlds

PR #6166, opened 2022-04-04
Fix global shift-based commands

### LINKED ISSUE #6136: When leaving a comment, keyboard shortcuts should be disabled
As reported by Jason, it's possible to press Shift-T in comments and see the entire theme shift. That's a bug.

But there are other issues to consider as well. We have multiple places where keyboard input is received, including:

* Comments
* Console
* Editing the Replay title
* Print statements
* etc

And every one of them might fall into this same keyboard shortcut issue.  Here's what I propose:

1. Remove our Shift-T shortcut. Command-K, D-A (for "Dark Theme"), Enter-> is how I toggle quickly and it's fine.
2. Disable keyboard shortcuts when in comments or editing the Replay title
3. Let's keep an eye on this issue, because it's a broader system we need to think through

jasonLaster:
Fixes #6136


jasonLaster:
Landing this because it is a pretty significant issue. @jaril feel free to leave feedback after the fackt

PR #6169, opened 2022-04-05
Showing env variables message

### LINKED ISSUE #6065: Show that node recordings include environment variables in security sensitive information
All node recordings include the contents of the environment in which they ran.  Once we support redaction this will change, but for now we should indicate this when sharing a recording publicly.  Since node recordings are only created via API keys, and API key recordings are not public, a recording needs to be explicitly made public to expose this information, but the "share" screen needs to indicate this.  See also https://github.com/RecordReplay/devtools/issues/6064

jonbell-lot23:
Fixes #6065

Was:
image

Now:
image

What's changed:

1. Removed the external link -- that's not relevant for node recordings
2. Removed "potentially sensitive" note if there are 0 domains
3. Added "This node recording contains all environment variables"

Notes:

I'm guessing that the code that was changed to

```
recording.url != ""
```

is probably not best practice. Please let me know what way is better!

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -71,7 +71,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
           ) : null}
         </div>
         <div className="group">
-          {recording.url != "" ? (
+          {recording.url ? (
             <Row>
               <Icon
                 filename="external"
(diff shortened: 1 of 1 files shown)


src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -69,33 +71,49 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
...
-              filename="external"
-              className="cursor-pointer bg-iconColor group-hover:bg-primaryAccent"
-            />
-            <div className="overflow-hidden overflow-ellipsis whitespace-pre" title={recording.url}>
-              <a href={recording.url} target="_blank" rel="noopener noreferrer">
-                {getDisplayedUrl(recording.url)}
-              </a>
-            </div>
-          </Row>
+          {recording.url != "" ? (
```
jaril, line 74: I'd switch this to `recording.url ? ( … ) : null`

jonbell-lot23: Thanks, done!

Issue #5970, opened 2022-03-25
Preview popup larger than screen, contents not readable


bgirard:
While inspecting a ShadowRoot, I couldn't fit the `preview-popup` on my screen:
<img width="1405" alt="Screen Shot 2022-03-24 at 8 26 51 PM" src="https://user-images.githubusercontent.com/793565/160048623-3616185d-d00c-4c8f-b953-9efa1e809e93.png">

jaril:
It looks like we have the appropriate [max-width for the popover component](https://github.com/RecordReplay/devtools/blob/285fba8621fad161a36d35ab42962981395eb213/src/devtools/client/debugger/src/components/Editor/Preview/Popup.css#L7) so this is not a CSS problem.

The problem is that the preview for the object is being miscategorized as a tooltip. The tooltip doesn't have a max width, which explains this behavior.

The reason why it's being miscategorized appears to be related to the ObjectInspector changes that we landed a few weeks ago:
- https://github.com/RecordReplay/devtools/commit/149fe8bbaa48f3e2a125bda23ab517cc7c7b8738
- https://github.com/RecordReplay/devtools/commit/aafb88531ec55fc931b29a18b7db8edf3adb6492

I left comments in [Jason's replay of the bug](https://app.replay.io/recording/popup-fills-the-width--ecdfde48-ac49-4fa4-8f80-acf484335dda?point=73341193192422716350803440895525710&time=35517.42245964713&hasFrames=true). 

The main culprit looks like the preview object itself having an undefined `properties` object. 

<img width="736" alt="image" src="https://user-images.githubusercontent.com/15959269/161605031-fce8194b-c324-4fbb-9d86-ab83940d9bec.png">

@hbenl Mind taking a look??

PR #6191, opened 2022-04-05
Fix preview popup width (#5970)

hbenl:
Removing some outdated logic that doesn't make sense after the ObjectInspector refactor (and probably didn't make sense even before that).


jaril:
@hbenl lint error but looks good!

PR #6204, opened 2022-04-05
Fixed comment.tests

bvaughn:
Fixes comment tests that were broken in #6030

There are several ways to fix tests like this. Happy to chat about alternatives.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "dev": "next dev -p 8080 ",
     "export": "next build && next export -o dist",
     "test": "jest",
+    "test:watch": "jest --watch",
     "start": "next start -p 8080",
     "build": "next build",
     "lint": "next lint",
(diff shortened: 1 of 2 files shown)


markerikson (approved):
Not fully up to speed on the comments behavior, but these test changes seem reasonable and the test passes.

(There's still a failure related to the "web console" test I wrote, which is bizarre because it used to pass and runs fine on _my_ machine :)  but that's separate from this PR)

package.json
```diff
@@ -7,6 +7,7 @@
     "dev": "next dev -p 8080 ",
     "export": "next build && next export -o dist",
     "test": "jest",
+    "test:watch": "jest --watch",
```
bvaughn, line 10: This is not strictly related to the `comments.test` fixes but I wanted a way to iterate on tests with a watcher.

bvaughn:
> (There's still a failure related to the "web console" test I wrote, which is bizarre because it used to pass and runs fine on my machine :) but that's separate from this PR)

I have another (local) branch now where I'm looking at this test. :)

PR #6206, opened 2022-04-05
Fix WebConsoleApp tests

bvaughn:
These tests were adding a wrapper `act()` call around `@testing-library` actions, but `@testing-library` automatically calls `act()`. (React was warning about the overlapping calls via `console.error`.)

I believe the most straight forward fix for this is to remove our  wrappers.


markerikson:
Nice! I'm just curious why this has continued to pass when I run it locally, _and_ seemed to be passing in CI for a while before starting to fail consistently.

markerikson (approved):

bvaughn:
> I'm just curious why this has continued to pass when I run it locally, and seemed to be passing in CI for a while before starting to fail consistently.

Not sure why ti would pass for you locally. Something to do with your environment?

The initial test didn't have any calls to `act` (and passes for me locally). It started failing when it got a little more complex.

PR #6209, opened 2022-04-05
Convert `source-actors` reducer to TS

markerikson:
This PR:

- Moves `asyncActionAsValue` from the promise middleware file over into `utils/async-value`, because that's where the related utils are.  (Also, `asyncActionAsValue` is _only_ used by the `source-actors` reducer anyway)
- Converts `utils/async-value` to TS
- Updates some of the "resource" TS types to pass through an `Arg` generic type
- Updates the resources logic to handle a weird case where an `id` field was actually receiving a single-item array like `["42"]`, and then that array was being used as an object key lookup
- Converts the actual `source-actors` reducer file to TS


docs/codebase-notes/sources-reducer.md
```diff
@@ -14,13 +14,40 @@ interface Location {
...
+  introductionUrl?: string;
+  isBlackBoxed: boolean;
+  isExtension: boolean;
+  isOriginal: boolean;
+  isPrettyPrinted: boolean;
+  relativeUrl?: string;
+  url?: string;
+}
+
 // Several types TBD here
```
jcmorrow, line 44: This seems like it probably isn't adding much, how would you feel about removing it now that you've got us to a full-TS place?

jcmorrow (commented):
I have a comment about an unrelated comment. Otherwise... seems good to me! (Disclaimer: I have never really understood how this thing worked).

docs/codebase-notes/sources-reducer.md
markerikson: Yeah, that's left over from the start of my notes on what the types looked like. I'll remove that in tomorrow's work on the actions files.

PR #6213, opened 2022-04-06
Comments polish 

jonbell-lot23:
Fixes https://github.com/RecordReplay/devtools/issues/6185

1. Dealing with transparency error:

Was:
image

Now:
image

2. Also removing the gray dim effect. @jcmorrow just confirming you have no concerns with this? I don't see the downside of removing this, but want to make sure I'm not missing anything.


jcmorrow (approved):
LGTM!

PR #6222, opened 2022-04-06
Changed comments to sort by 'point' rather than 'time'

bvaughn:
Resolves https://github.com/RecordReplay/customer-support/issues/48

I don't have a great corpus of testing data for this, but I verified that this change fixed the incorrectly-ordered comments in the parent issue.

### DIFF
--- a/src/ui/components/Transcript/Transcript.tsx
+++ b/src/ui/components/Transcript/Transcript.tsx
@@ -27,7 +27,7 @@ export default function Transcript() {
       displayedComments.push(pendingComment.comment);
     }
 
-    const sortedComments = sortBy(displayedComments, ["time", "createdAt"]);
+    const sortedComments = sortBy(displayedComments, ["point", "createdAt"]);
     return sortedComments;
   }, [comments, pendingComment]);
 


jaril:
I'd check the sorting behavior here — `point` is a long string and I think sometimes we may have had to pull out `BigInt` to get it right?

bvaughn:
> I'd check the sorting behavior here — `point` is a long string and I think sometimes we may have had to pull out `BigInt` to get it right?

@jaril Oh thank you!

Why would that make a difference? Sorting a collection on an attribute that's a string containing only numbers should function the same as sorting a collection on an attribute containing numbers I think – so long as the string doesn't also contain non-numeric characters? Do you think `point` might contain non-numeric characters?

I could replace the "point" param with a function that casts to `BigInt` before comparing. It just seems unnecessary (although maybe I'm missing something).

bvaughn:
@jaril Is that unexpected? `5192296858926797881666186703274057` _is_ smaller than `519229685892679788166618670327405799999` isn't it?

Just to sanity check...
```js
function compareValues(a, b) {
  const stringResult = a > b;
  const intResult = parseInt(a) > parseInt(b);
  const bigIntResult = BigInt(a) > BigInt(b);
  console.log(`Compare ${a} > ${b}\n  string: ${stringResult}\n  int: ${intResult}\n  bigInt: ${bigIntResult}`);
}

compareValues("5192296858926797881666186703274057", "5192296858926797881666186703274057");
compareValues("5192296858926797881666186703274057", "519229685892679788166618670327405");
compareValues("5192296858926797881666186703274057", "519229685892679788166618670327405799999");
```
Prints:
```
Compare 5192296858926797881666186703274057 > 5192296858926797881666186703274057
  string: false
  int: false
  bigInt: false
Compare 5192296858926797881666186703274057 > 519229685892679788166618670327405
  string: true
  int: true
  bigInt: true
Compare 5192296858926797881666186703274057 > 519229685892679788166618670327405799999
  string: false
  int: false
  bigInt: false
```

jaril (approved):

bvaughn:
No worries at all! Happy to confirm 😄

PR #6228, opened 2022-04-06
Refactor the way we looking at indexing progress

### LINKED ISSUE #5851: Loading status not showing
I wrote a backend PR to cleanup the loaded / loading regions which are sent to the client (https://github.com/RecordReplay/backend/pull/4967), but this information is not shown appropriately while waiting for indexing to complete.

1. The loading progress indicator on the play button never updates while loading is happening.
2. Unloaded regions are not shown until after indexing completes.  I think it would be clearer if the unloaded regions reflect the latest information from the backend.

There are also a couple weird UI problems I noticed:

1. Most of the timeline is very indistinct and hard to see.
2. There is a spinner on the right which never goes away and has no tooltip, I don't know what it's for.

This replay shows all of these problems:

https://app.replay.io/recording/weird-loading-behavior--e09f6427-7141-455b-ba15-b54e6a96ca1f?point=217427430961111295803169762487312225&time=78448&hasFrames=false

jaril:
Fix #5851.

- [x] Use the new indexed value from LoadedRegions
- [x] Don't differentiate between loaded/unloaded regions in the timeline

To recap:
- Now that we've landed the [backend patch](https://github.com/RecordReplay/backend/pull/5186), this fixes the way we keep track of the `indexed` value as regions become loaded/unloaded.
- This changes the way we display things in the timeline — within the focused region, we used to expose the parts that were loaded vs unloaded by differentiating them visually. This gets rid of that and just says, in the timeline, the user only cares about what's focused and what's not focused. Unloaded regions don't prohibit clicking on parts of the timeline, so this was deemed okay.
- That said, each region has an indexing state. If your current focused regions has regions are not fully-indexed (and non-indexed region also means that they're not loaded), the information in DevTools (e.g. hit counts, console messages) could be misleading. So while we're indexing, it's nice to be able to signal to the user that the UI is not fully dressed yet.
- The way we'll do this is with the indexing loader that wraps the play button. If that indexing loader/spinner is anything but 100%, then replay's still getting dressed. If it's 100%, then the user can go nuts.
- It used to be that we would hide the indexing loader once it reaches 100%. But because "We've 100% indexed everything!" is an important signal, I thought we should always show that. That's a clearer signal since we'll always display some progress between 0-100%. The current approach displays 0-99.9%, then hides the spinner when it gets to 100%. I wasn't a fan of the idea that the green light would be hidden since having to infer the green light opens up the possibility of user misinterpretation, so this was my preference. It's less subtle, but that gives us space to dial it back if we deem it too noisy and unhelpful. (This could be controversial 🙋‍♂️ @jonbell-lot23)

Follow ups:
- Remove the unloaded regions code from the `BreakpointTimeline`.
- Backend is returning indexed regions that are not strictly a subset of the loading regions. This should be fixed on the backend.
- Haven't deployed the new protocol yet so this relies on its own `LoadedRegions` type instead of importing it from the protocol.

### DIFF
--- a/src/ui/reducers/app.ts
+++ b/src/ui/reducers/app.ts
@@ -259,22 +259,30 @@ export const getIndexingProgress = (state: UIState) => {
     return null;
   }
 
-  const indexedProgress = regions.indexed.reduce(
-    (sum, region) => sum + (region.end.time - region.begin.time),
-    0
-  );
-  const loadingProgress = regions.loading.reduce(
+  const { loading, indexed } = regions;
+
+  const indexedProgress = indexed
+    .filter(indexedRegion => {
+      // It's possible for the indexedProgress to not be a subset of
+      // loadingRegions. This acts as a guard if that should happen.
+      // Todo: Investigate this on the backend.
+      return loading.some(
+        loadingRegion =>
+          indexedRegion.begin.time >= loadingRegion.begin.time &&
+          indexedRegion.end.time <= loadingRegion.end.time
+      );
+    })
+    .reduce((sum, region) => sum + (region.end.time - region.begin.time), 0);
+  const loadingProgress = loading.reduce(
     (sum, region) => sum + (region.end.time - region.begin.time),
     0
   );
 
-  console.log({indexedProgress, loadingProgress});
-  
-  return indexedProgress / loadingProgress * 100;
+  return (indexedProgress / loadingProgress) * 100 || 0;
 };
 export const getIsIndexed = (state: UIState) => {
   return getIndexingProgress(state) === 100;
-}
+};
 export const getLoading = (state: UIState) => state.app.loading;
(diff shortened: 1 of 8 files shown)


src/ui/reducers/app.ts
```diff
@@ -252,8 +252,37 @@ export const isInspectorSelected = (state: UIState) =>
...
+export const getIndexingProgress = (state: UIState) => {
+  const regions = getLoadedRegions(state);
+
+  if (!regions) {
+    return null;
+  }
+
+  const { loading, indexed } = regions;
+
+  const indexedProgress = indexed
```
markerikson, line 264: :hand: This looks like a good candidate for a memoized Reselect `createSelector` version rather than a plain function.  Not sure how big `indexed` or `loading` are, but a `.some()` inside a `.filter()` looks like it's worth avoiding if we don't need to recalculate it.

PR #6235, opened 2022-04-07
Fix eslint

jasonLaster:
the accordion package might have broken our eslint config


.eslintrc.json
```diff
@@ -20,6 +20,7 @@
   "rules": {
     "react/prop-types": 0,
     "react/require-default-props": 0,
+    "react-hooks/exhaustive-deps": 0,
```
bvaughn, line 23: TBH I'm not sure I'd do this.

I think the rule is kind of silly in the case of an effect that I want to manage via a ref (valid but not super common) _but_ it's not the worst thing to have to add eslint-ignore comment. I'm worried that removing it might lead us to accidentally omit dependencies from the array without realizing, which could cause bugs.

bvaughn: So today, this is a possible false positive:
```js
const ref = useRef(false);

useEffect(() => {
  if (ref.current) {
    return;
  }

  ref.current = true;

  // Do whatever I want with whatever references I want...
});
```

I'm worried that disabling this rule might lead to a false negative:
```js
useEffect(() => {
  // Do something with depA and depB...
}, [depA]);
```

jasonLaster: That makes sense.

PR #6236, opened 2022-04-07
Add security info to share dialog

### LINKED ISSUE #6064: Show security sensitive information when sharing a recording publicly
The upload screen for a recording shows security sensitive information when sharing a recording publicly, but if a recording has been created and then the "share" screen is used to make it public, the security sensitive information is not shown.  Sensitive information in a recording should be shown when the "share" screen when opening a recording for public access.

jonbell-lot23:
Fixes #6064 (Not quite, but it gets a fair amount done)

Addressed:
* Node recordings get a node message
* Normal recordings get a cookies message

Not addressed:
* I dropped the "learn more" link because it was going to cause a modal on top of a modal from a tiny box with three tight links. The overall result was ridiculous; what we really need to do is address the overall design. Perhaps with a pattern similar to our upload screen, where clicking more info shows a little pane to the side.

Node recording before change:
image

Node recording after change:
image

Normal recording before change:
image

Normal recording after change:
image

### DIFF
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -158,17 +158,19 @@ export default function PrivacyDropdown({ recording }: { recording: Recording })
   }
 
   return (
-    <PortalDropdown
-      buttonContent={<DropdownButton>{summary}</DropdownButton>}
-      buttonStyle={"overflow-hidden"}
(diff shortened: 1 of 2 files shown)


jaril (approved):
LGTM. It's a bit in your face but that's kind of the point. @jasonLaster quick look?

jasonLaster:
Looks good to me

PR #6239, opened 2022-04-07
Fix theme bug

jaril:
Selecting `system` as the theme is currently not displaying correctly in the user's settings. This fixes that, and removes some `console.logs` on the way.


src/ui/components/Library/TeamTrialEnd.tsx
```diff
@@ -13,8 +13,6 @@ export default function TeamTrialEnd({ currentWorkspaceId }: { currentWorkspaceI
   const router = useRouter();
   const { userId: localUserId } = hooks.useGetUserId();
 
-  console.log("hello");
```
bvaughn, line 16: Goodbye!

PR #6256, opened 2022-04-07
Fix showing expired badge for teams with a payment method

ryanjduffy:
The logic for determining that a sub was expiring was failing because the "date in calendar days" was 0 but since the customer has a payment method, that's okay!

https://app.replay.io/recording/replay-with-expired-fixed--b15d847e-c096-43a2-827a-5609ee38e6ae

### DIFF
--- a/src/ui/utils/workspace.ts
+++ b/src/ui/utils/workspace.ts
@@ -16,22 +16,20 @@ export function subscriptionEndsIn(workspace: Workspace, date?: Date | number) {
   const subscription = workspace.subscription;
 
   if (!subscription) {
-    return 0;
+    return Infinity;
   }
 
   if (
     subscription.status === WorkspaceSubscriptionStatus.Canceled ||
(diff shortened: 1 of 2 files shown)


src/ui/utils/workspace.ts
```diff
@@ -40,7 +38,7 @@ export function subscriptionEndsIn(workspace: Workspace, date?: Date | number) {
 export function subscriptionExpired(workspace: Workspace, date?: Date) {
   const expiresIn = subscriptionEndsIn(workspace, date);
 
-  return expiresIn <= 0;
+  return expiresIn < 0;
```
jaril, line 41: Haven't wrapped my head around this fully. We're looking for a negative `expiresIn` value to indicate that it's expired?

jaril: Oh we count calendar days got it got it.

jaril (approved):

PR #6272, opened 2022-04-08
Fix credit card in dark theme

### LINKED ISSUE #5989: Credit card field looks odd in dark theme
Hey @ryanjduffy I remember we talked about how we pull this information externally, so I'm not sure how to fix this:

image

(It's also dinnertime so it's possible the fix is easy but I can't address right now)

See also https://github.com/RecordReplay/devtools/issues/5987 that I just fixed on my way to discovering this

jonbell-lot23:
Fixes #5989

Was:
image

(Notice how the text isn't showing at all -- that's because it's black)

Now:
image

image


ryanjduffy (approved):

bvaughn:
Scanning recent PRs, I was curious why this PR actually passed all CI checks– but looking at it, it looks like only 3 checks ran (vs 12 on other PRs). Why is this?

PR #6280, opened 2022-04-08
fix: getSystemColorSchemePreference condition to check system theme

### LINKED ISSUE #6260: System theme doesn't quite work
This helper function isn't quite right:
https://github.com/RecordReplay/devtools/blob/ce03f35d4f536f24a3b9cb5200de9189bf4abfa0/src/ui/utils/environment.ts#L143-L145

See discord thread [here](https://discord.com/channels/779097926135054346/795737681589174312/961713022089437194), stack overflow [here](https://stackoverflow.com/questions/56393880/how-do-i-detect-dark-mode-using-javascript/57795495#57795495).

ashwinkhode:
Fixes #6260 

This is my first time contributing. Please let me know if there can be any improvements :)

### DIFF
--- a/src/ui/utils/environment.ts
+++ b/src/ui/utils/environment.ts
@@ -141,5 +141,7 @@ export function getDisplayedUrl(url: string) {
 }
 
 export function getSystemColorSchemePreference() {
-  return window.matchMedia("(prefers-color-scheme:dark)") ? "dark" : "light";
+  const prefersDarkTheme = window.matchMedia("(prefers-color-scheme:dark)").matches
+
+  return prefersDarkTheme ? "dark" : "light";
 }
--- a/src/ui/utils/environment.ts
+++ b/src/ui/utils/environment.ts
@@ -141,7 +141,7 @@ export function getDisplayedUrl(url: string) {
 }
 
 export function getSystemColorSchemePreference() {
-  const prefersDarkTheme = window.matchMedia("(prefers-color-scheme:dark)").matches
+  const prefersDarkTheme = window.matchMedia("(prefers-color-scheme:dark)").matches;
 
   return prefersDarkTheme ? "dark" : "light";
 }


ashwinkhode:
@jaril please review my PR 🙈

src/ui/utils/environment.ts
```diff
@@ -141,5 +141,7 @@ export function getDisplayedUrl(url: string) {
 }
 
 export function getSystemColorSchemePreference() {
-  return window.matchMedia("(prefers-color-scheme:dark)") ? "dark" : "light";
+  const prefersDarkTheme = window.matchMedia("(prefers-color-scheme:dark)").matches
```
jaril, line 144: Small lint fix — this needs a `;` at the end. Running `trunk check` should also take care of it!

john-msftech:
Hi @ashwinkhode ! I was just about to start looking at this issue as my first contribution as well. 

The only thing I notice with your change is that while it will now respect the system setting it will also need to watch for if the system theme changes while having the system setting configured in the app.

Hope that makes sense and I'd be happy to elaborate!

jaril (changes requested):
Thanks for contributing! It just needs a small fix otherwise let's get it in :)

src/ui/utils/environment.ts
ashwinkhode: fixed it @jaril . Also the above mentioned issue by @john-msftech seems like a different issue because as of now, whenever theme is set to system it gets changed dark or light based on system default theme. If we want to listen to system theme change then we need to keep theme set to system only and not change it to dark/light 🤔

jaril (approved):

markerikson:
Thanks!

jonbell-lot23:
🎉 Seconded, thanks very much! Looking forward to getting your next contributions!

PR #6283, opened 2022-04-09
Save console filters on a per-replay session basis

### LINKED ISSUE #6134: Save console filters in ReplaySessions
**Description:**

The console filters should be saved within replay sessions so they do not persist for all replays

jaril:
Fix #6134.

### DIFF
--- a/src/ui/setup/prefs.ts
+++ b/src/ui/setup/prefs.ts
@@ -1,6 +1,6 @@
 import { UIStore } from "ui/actions";
 import { UIState } from "ui/state";
-import { prefs, asyncStore, features } from "ui/utils/prefs";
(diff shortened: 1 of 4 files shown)


src/ui/setup/prefs.ts
```diff
@@ -74,19 +76,6 @@ export function updatePrefs(state: UIState, oldState: UIState) {
       return state.consoleUI.timestampsVisible;
     });
   }
-
-  if (state.messages?.filters && oldState.messages?.filters) {
-    updateWebconsolePref("filterError", (state: UIState) => state.messages.filters.error);
```
markerikson, line 79: :question: Where do we do the persistence step now?

markerikson (approved):

PR #6288, opened 2022-04-10
Show risky experimental settings (fixes #6245)

### LINKED ISSUE #6245: Update super secret experimental settings
We have two experimental settings which require pressing shift nine times. We should update this so that they're always shown, but have a disclaimer for the section that explains that there's an elevated chance that this will cause replaying issues.

jasonLaster:
Screen Shot 2022-04-10 at 4 57 53 PM

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -155,8 +155,8 @@ export function createSession(recordingId: string): UIThunkAction {
       const experimentalSettings: ExperimentalSettings = {
         listenForMetrics: !!window.app.prefs.listenForMetrics,
         disableCache: !!window.app.prefs.disableCache,
-        useMultipleControllers: !!window.app.features.useMultipleControllers,
-        multipleControllerUseSnapshots: !!window.app.features.multipleControllerUseSnapshots,
+        useMultipleControllers: !!window.app.features.tenMinuteReplays,
+        multipleControllerUseSnapshots: !!window.app.features.tenMinuteReplays,
       };
 
       dispatch(showLoadingProgress());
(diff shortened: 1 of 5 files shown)


bhackett1024:
These settings should be combined into one, "long recording support".  Controllers and snapshots are backend internal concepts that don't belong in the devtools.  I don't know what the point of marking these as especially risky is, experimental settings are experimental for a reason.

bhackett1024:
Here is a patch I was working on to combine the settings, I think these changes should be folded into this PR.

https://gist.github.com/bhackett1024/67145e65b9605f3b350a45e38c7bb4b3

src/ui/actions/session.ts
```diff
@@ -155,8 +155,8 @@ export function createSession(recordingId: string): UIThunkAction {
       const experimentalSettings: ExperimentalSettings = {
         listenForMetrics: !!window.app.prefs.listenForMetrics,
         disableCache: !!window.app.prefs.disableCache,
-        useMultipleControllers: !!window.app.features.useMultipleControllers,
-        multipleControllerUseSnapshots: !!window.app.features.multipleControllerUseSnapshots,
+        useMultipleControllers: !!window.app.features.tenMinuteReplays,
```
jaril, line 158: can't this just be `tenMinuteReplays`?

jaril: Oh I see — nevermindnevermind

jaril (approved):

jaril:
Re-ran the tests but this looks like this should be ready to go out @jasonLaster

PR #6305, opened 2022-04-11
Add loading tooltip while React DevTools picker initializes

### LINKED ISSUE #5535: React DevTools Node Picker should have a loading... tooltip
It takes ~2 seconds for the Node picker to start. We should show a Loading... tooltip while you hover in the canvas (similar to the elements panel).

bvaughn:
Resolves #5535

This commit adds a new state (_initializing_) for the React DevTools node-picker and a new reusable `<Tooltip>` component that can be used to show text aligned to the mouse cursor.

Loom demo of new functionality:
https://www.loom.com/share/d6a483cd2ff848958162cb96a9e672ee


src/ui/components/shared/Tooltip.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+export default function Tooltip({ label, targetID }: TooltipProps) {
+  const [coordinates, setCoordinates] = useState<Coordinates | null>(null);
+
+  const elementsRef = useRef<Elements>({
+    container: null,
+    tooltip: null,
+  });
+
+  const tooltipRefSetter = useCallback((tooltip: HTMLDivElement | null) => {
```
bvaughn, line 29: Using a ref-setter function lets us do a one-time lookup of the nearest positioned parent and cache the value for later.

src/ui/components/shared/Tooltip.tsx
```diff
@@ -0,0 +1,142 @@
...
+    if (target) {
+      const onMouseLeave = () => {
+        setCoordinates(null);
+      };
+      const onMouseMove = (event: MouseEvent) => {
+        setCoordinates({ pageX: event.pageX, pageY: event.pageY });
+      };
+
+      target.addEventListener("mouseleave", onMouseLeave);
+      target.addEventListener("mousemove", onMouseMove as (e: Event) => void);
```
bvaughn, line 51: This TypeScript type refinement seems silly, but I was getting an error without it:
> Argument of type '(event: MouseEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.

I'm not sure why this doesn't happen other places, like
https://github.com/RecordReplay/devtools/blob/f94f3d8afe599efdfc9f614718a3c3f68510fc27/src/ui/components/shared/CommentTool.tsx#L87-L94

jaril (approved):

src/ui/actions/app.ts
```diff
@@ -85,6 +85,9 @@ export type SetHoveredLineNumberLocation = Action<"set_hovered_line_number_locat
   location: Location | null;
 };
 export type SetIsNodePickerActive = Action<"set_is_node_picker_active"> & { active: boolean };
+export type SetIsNodePickerInitializing = Action<"set_is_node_picker_initializing"> & {
```
markerikson, line 88: :speech_balloon: As an FYI for future reference: this whole pattern of hand-defining TS types for action objects and unioning them together is actually something that the Redux team specifically advises _against_:

https://redux.js.org/usage/usage-with-typescript#avoid-action-type-unions

As we slowly migrate the Redux logic over to Redux Toolkit, that can go away - RTK's `createSlice` generates correctly-typed action creators for you automatically.

But, it's the pattern that's in this codebase already, so it doesn't _hurt_ to add this.

markerikson (approved):
Seems reasonable to me.  I suppose there might be some other way to handle the "is initializing" state, but given that these look like two widely separated areas of the codebase, throwing that into Redux is a valid way to solve the issue for now.

src/ui/actions/app.ts
bvaughn: Thanks for the added info.

I'm mostly trying to avoid thinking too hard about our Redux setup and how it feels to work with it until after you've made the changes you're working on.

src/ui/components/shared/Tooltip.tsx
jaril: Played around with this a bit. Looks like the difference is that `videoNode` could be `window`. This quiets TS down:

```diff
diff --git a/src/ui/components/shared/Tooltip.tsx b/src/ui/components/shared/Tooltip.tsx
index 0b0bb50c..c14e3fe1 100644
--- a/src/ui/components/shared/Tooltip.tsx
+++ b/src/ui/components/shared/Tooltip.tsx
@@ -38,7 +38,7 @@ export default function Tooltip({ label, targetID }: TooltipProps) {
 
   // Window event listeners only need to run on mount/unmount.
   useLayoutEffect(() => {
-    const target = targetID ? document.getElementById(targetID) : window;
+    const target = document.getElementById(targetID!);
     if (target) {
       const onMouseLeave = () => {
         setCoordinates(null);
@@ -48,10 +48,10 @@ export default function Tooltip({ label, targetID }: TooltipProps) {
       };
 
       target.addEventListener("mouseleave", onMouseLeave);
-      target.addEventListener("mousemove", onMouseMove as (e: Event) => void);
+      target.addEventListener("mousemove", onMouseMove);
       return () => {
         target.removeEventListener("mouseleave", onMouseLeave);
-        target.removeEventListener("mousemove", onMouseMove as (e: Event) => void);
+        target.removeEventListener("mousemove", onMouseMove);
       };
     }
   }, [targetID]);
```

jaril: Didn't dig in more past that :P

bvaughn: >  Looks like the difference is that `videoNode` could be `window`.

But this is okay:
```js
window.addEventListener("mousemove", onMouseMove);
```

And this is okay:
```js
const target = document.getElementById(targetID!);
if (target) {
  target.addEventListener("mousemove", onMouseMove);
}
```

Just not this:
```js
const target = targetID ? document.getElementById(targetID) : window;
if (target) {
  target.addEventListener("mousemove", onMouseMove);
}
```

I don't understand why.

PR #6312, opened 2022-04-11
Make "Quick Open" modal accessible from "Viewer" mode panel 

### LINKED ISSUE #5539: Cmd+P (file finder) should work from the viewer
I almost print a paper copy of Replay about 4 times a day because I open up a Replay, see what I'm looking for the in the viewer, and promptly hit `cmd+P` to go to a file. It feels like it would be reasonable to make `cmd+P` bring up the fuzzy finder in the viewer (so you can still hit escape if you want to avoid switching modes), but once the user decides on a file, transfer to devtools and open that file in the editor.

markerikson:
This PR depends on #6311 .  (I'm trying out a sort of pseudo-"stacked" PRs approach here. I think I've seen that merging the other PR first may end up causing this one to auto-close, so we'll see how that works out.)

Fixes #5539 .

The "Quick Open" modal is accessed via several keyboard shortcuts,
including `CTRL-P`. Previously, those shortcuts were only set up
when the "DevTools" mode panel was on screen.

We want the "Quick Open" modal to be accessible even when the user
has "Viewer" mode active, and if they do select a file, it should
switch over the "DevTools" mode so they can see the contents.

This required several changes:

- `<QuickOpenModal>` is now rendered by `ui/components/App.tsx`,
  in order for it to always be displayed regardless of active mode
- The shortcuts handling has moved up to `KeyboardShortcuts.tsx`,
  which both centralizes the behavior and makes it always globally
  turned on.  This also includes the "Escape" key handling to close
  the "Quick Open" modal (and some "Active Search" UI).
- The previous rendering logic was also tied into showing a
  "shortcuts" modal left over from the FF days. We're not sure how
  we want to keep that going forward. For now, I've removed the
  shortcut/state/rendering for that modal.


jaril:
Related issues:
https://github.com/RecordReplay/devtools/issues/6279
https://github.com/RecordReplay/devtools/issues/6310

jaril (approved):
Needs a rebase but generally looks good!

[Walkthrough here](https://www.loom.com/share/afe80fb71c19484bae17d3e1cdb0aa5b). The only thing I'd address before merging is that the loading text shown in the quickopen before the sources populate has a noticeably larger font size.

jaril:
One thing I noticed is that while quick open and command palette are mutually exclusive (only one can appear at one time), it's possible for there to be a modal and quick open/command palette. This kinda looks janky. I'm not sure if this means we should only allow one to be open at a time, or we should disallow opening one if another one is open. 

This is already existing behavior so it's not a blocking comment. @jonbell-lot23 this might be a separate ticket. 🙋‍♂️

jaril:
@ceceliacreates docs maybe but I think we're okay!

markerikson:
Looks like the whole DevTools area has a `text-xs` applied which caused the "Loading" area to end up as 12pt, and now the modal is outside that.  (The text input is fixed at 14pt.)

I'll try adding a `text-xs` to the modal itself and see if that helps.

markerikson:
Backend is having issues atm, but the UI here looks good.

PR #6323, opened 2022-04-12
Dependabot should only update dependencies once a month

### LINKED ISSUE #6319: Make dependabot more helpful
We usually wake up to ~25 PRs every monday for individual dependencies for us to bump. Then we have to figure out which ones are safe to merge vs one not to. Once we've decided which are to be merged, we have to do them one-by-one since they usually result in merge conflicts with each other in `package-lock.json`.

This takes a non-insignificant amount of time to just herd the process along. We talked at standup about whether we should just remove dependabot, or relax its config.

Close this ticket once we've figured out the course of action.

jaril:
Fix #6319.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
   - package-ecosystem: npm
     directory: /
     schedule:
-      interval: weekly
+      interval: monthly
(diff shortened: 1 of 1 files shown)


bvaughn (approved):
LGTM 👍🏼

Issue #6233, opened 2022-04-06
Comment sticks around


jonbell-lot23:
Check this out:
https://app.replay.io/recording/adding-comments--d4afaf63-1766-4281-a0ef-b06e03e41a19?point=100925270199248397248981589133626899&time=34817&hasFrames=false

<img width="527" alt="image" src="https://user-images.githubusercontent.com/9154902/162846966-66b06b40-cea4-4b0e-82d7-0f0e89ce13d2.png">

In this example I wrote "Hello" in the comment field and hit enter. Then I hit enter again. But the comment never successfully sticks. It just sits there in edit state.

jonbell-lot23:
Updated, thanks!

bvaughn:
This looks to me like the same bug as #6502 (which I just dug into).

There is a problem in the way the frontend tries to hand off the pending comment from Redux to Apollo when saving:
https://github.com/RecordReplay/devtools/blob/d187caadf0932c485eebb0316a8ba7355b3130a9/src/ui/hooks/comments/useAddComment.tsx#L32-L93

What ends up happening is that there is a copy of the pending comment in the Apollo cache (that doesn't get removed) and then we also add the newly-saved pending comment to that cache as well. This creates visual duplicates and leaves the input field hanging around still visible (because the UI chooses whether to render the new comment input or not based on the magic "PENDING" id).

PR #6335 refactors this logic and should fix this issue along the way, so I'm going to roll this into that.

bvaughn:
Closed via #6335

PR #6335, opened 2022-04-12
Comment client architecture overhaul

### LINKED ISSUE #6138: Comment flicker (part 3)
We've identified most of the difficult race conditions with comment flickering. I think the remaining ones are fairly straight forward

1. Comments flicker when transitioning from one draft to another ([replay](https://app.replay.io/recording/comments-flicker-when-replacing-a-draft--17018258-f1fd-49e6-8d9d-78d169a533bf?point=105144011415874916729333207047078683&time=38579.125322516105&hasFrames=true))
2. Comments flicker when toggling sidebar panels ([replay](https://app.replay.io/recording/comments-flicker-when-replacing-a-draft--64c9f59d-0939-43ea-88be-7dfc0442657d?point=41213856331603456710710582096957179&time=29137&hasFrames=false))

In both cases, the problem is that the comment card initially renders without the editor. And then the editor comes in. We can probably handle that one of two ways. the first is to animate in the editor, the second is to not show the comment card until we have the editor. This might cause a similar flicker if we simply flipped it and the editor appeared before the rest of the comment, but I think it could work.

bvaughn:
Resolves #6138, #6233, and #6502.

Loom walk through of the change (with UI demo):
https://www.loom.com/share/8eb1cc84a1fd4eaea2499e3e6701a1e2

# Background information
I left a long trail of comments on #6138 as I dug into the underlying code here (both in Replay and the tiptap editor).

https://github.com/RecordReplay/devtools/issues/6138#issuecomment-1097031057 summaries my findings:
1. Clicking the "comment" button creates a new pending comment in Redux state.
2. This new comment has a `createdAt` date that corresponds to the time the button was clicked.
3. Replay uses `createdAt` and `updatedAt` dates as React keys to forcefully reset state.
4. As a result, clicking "comment" deeply unmounts and remounts the pending comment subtree, including the tiptap editor.

https://github.com/RecordReplay/devtools/issues/6138#issuecomment-1096891604 summaries my findings on the tiptap editor:
1. When mounting initially, this component renders an empty div element.
2. The tiptap `useEditor` hook subscribes to "transaction" events dispatched from the tiptap Editor and schedules React updates.
3. After one of these updates (in `componentDidUpdate`) the React wrapper copies HTML elements (created by the tiptap Editor's shared "core" code) into this empty div.

The two above findings combined explain why we see the "flicker" that was reported.

https://github.com/RecordReplay/devtools/issues/6502#issuecomment-1109811888 summarizes my findings on other related bugs (#6233, #6502):
1. When a pending comment is saved, `useAddComment` optimistically moves it to the Apollo cache.
2. Once the server returns a successful response, the pending comment should be removed and the newly-saved version left in its place.
3. What actually happens in some cases is that the pending comment is left in the Apollo cache alongside of the newly saved version, causing the new comment input to stick around.
4. Typing "Enter" again may cause more duplicate comments to be shown in the UI.

# Proposed changes
Based on my own findings, and after chatting with @jcmorrow, this PR will make the following changes:
- [x] Remove `PENDING_COMMENT_ID` and the `commentKey` and `commentKeys` helper functions. (When we need to have `key`s for React elements, just use `comment.id`).
- [x] (Related to the above) move away from using `createdAt` or `updatedAt` in keys. These values aren't stable and so they cause unnecessary unmounts and remounts within the React tree. (Even for comments that are persisted, Apollo will return a new `createdAt` value.)
- [x] ~~Add new dispatcher/reducer logic to allow views to choose between _create a new pending comment_ and _update/move the current pending comment_. Currently we only support the former, and it causes problems with tiptap.~~
  * I don't think we need this as an _explicitly_ change because of the new id format.
- [x] Remove imperative-ref-hack in `TipTapEditor`. This is only necessary if we recycle _pending_ comment IDs.
- [x] Remove optimistic updates and manually modifying Apollo's in-memory cache because this leads to certain types of race conditions (see https://github.com/RecordReplay/devtools/pull/6335#issuecomment-1109861829 and [this replay](https://app.replay.io/recording/replay-apollo-mutation-hook-caching-bug--14e387b5-5191-43c6-813b-c7ca4cff8d71?point=62632080871353839877609415755436860&time=29784&hasFrames=false#)).

# Future changes
There are several follow up changes we can make after this PR.
## To the way we model comments...
- [ ] Move _pending_ comments out of Redux and into Apollo.
- [ ] Update the server to also be aware of the (currently local-only) `isPending` attribute, but maybe rename it to `isPublished`.
- [ ] Comments that have not been "published" should only be visible to their authors, (so the server should filter them). Submitting a comment (pressing _enter_) should "publish" it.

## To our UX...
- [ ] If the user clicks on a different part of the UI when a pending comment has text entered but not _saved_, we could prompt the user if they want to save or discard (rather than just erasing the content). This change is probably not relevant if we make the above data model changes, since we could just keep multiple pending comments around until the user explicitly cancels them ("Escape" key) or publishes/shares them ("Enter" key).

## To our React components...
- [ ] #6366: Replace Tip Tap with something like [Lexical](https://www.npmjs.com/package/lexical) that's more React-compatible.


jasonLaster (commented):
At a high level the code/reasoning looks great. Also probably useful regardless of whether we go off of TipTap as in, these changes will make the component more stable regardless.

markerikson (approved):
Don't have much context on the comment logic beyond what you wrote in the issue, but changes seem reasonable!

jcmorrow:
I think @bvaughn already knows this because we talked about it at the end of the day (and this is still a draft PR), but I just want to flag given that it is getting some ✅ that in its current state the preview branch has some pretty broken stuff that definitely should not go out yet:

https://user-images.githubusercontent.com/5903784/163096401-d110f9b7-80fa-4fa2-a1f8-051678a6f8c1.mp4

We've got a good plan to tackle this though and I have high hopes!

bvaughn:
Thanks for the feedback, folks! 😄 

There are a few things left to do on this branch before it's ready to land, but I'll keep you posted.

bvaughn:
I think this PR is pretty close but there are a few scenarios that I want to outline the expected behavior for (to make sure we're all on the same page).

Note that if we refactor our internal data model to support multiple simultaneous pending comments, and if we save them to Apollo so the only difference is whether they've been "published" or not (aka shared with other people), then some of the below functionality changes.

## The Viewer/DevTools panel

### Steps
1. Click on one point of the viewer.
2. Maybe enter some text (maybe not).
3. My click wasn't perfect, so I click again (or "drag").

### Expected outcome
If I entered text between steps 1 and 3, the pending comment (text) should be preserved. The comment should just be "moved" to a different visual point on the screen.

I think the stakes of doing "the wrong thing" here are small. If I am surprised to see my pending comment moved (because I wanted to make a new one) I can click back where I had before and hit "enter" (to save the comment). If we took the opposite approach of discarding and recreating the comment (what we currently do) then I lose what I typed. If that's not the behavior I expected, I might be frustrated.

We could try to be smart here and decide whether to recreate or reuse based on the "distance" between points that were clicked on (some percentage of the size of the viewer) but that seems like overkill at this point.

## The console

### Steps
1. Click "comment" above a specific print statement.
2. Maybe enter some text (maybe not).
3. Click "comment" above a different print statement.

### Expected outcome
A new pending comment is created. Different print statements may or may not be related to each other.

I think it's easier to click a specific print statement/line than it is to click a precise x/y coordinate. (We also don't show the tooltip text "move the marker" here so...)

If this difference in behavior feels uncomfortable, (it does to me), then maybe that suggests that we need to add a feature to prompt users before erasing a pending comment with text in it. ("Would you like to save this comment?") This is only relevant so long as our data model only supports one pending comment at a time, so hopefully this discrepency goes away soon.

## The source code panel

### Steps
1. Click "comment" above a specific line of code.
2. Maybe enter some text (maybe not).
3. Click "comment" above a different line of code.

### Expected outcome
A new pending comment is created. Different lines of code are not related to each other.

(Same caveats as above.)

bvaughn:
I think this PR is pretty close but there are a few scenarios that I want to outline the expected behavior for (to make sure we're all on the same page).

Note that if we refactor our internal data model to support multiple simultaneous pending comments, and if we save them to Apollo so the only difference is whether they've been "published" or not (aka shared with other people), then some of the below functionality changes.

## The Viewer/DevTools panel

### Steps
1. Click on one point of the viewer.
2. Maybe enter some text (maybe not).
3. My click wasn't perfect, so I click again (or "drag").

### Expected outcome
If I entered text between steps 1 and 3, the pending comment (text) should be preserved. The comment should just be "moved" to a different visual point on the screen.

I think the stakes of doing "the wrong thing" here are small. If I am surprised to see my pending comment moved (because I wanted to make a new one) I can click back where I had before and hit "enter" (to save the comment). If we took the opposite approach of discarding and recreating the comment (what we currently do) then I lose what I typed. If that's not the behavior I expected, I might be frustrated.

We could try to be smart here and decide whether to recreate or reuse based on the "distance" between points that were clicked on (some percentage of the size of the viewer) but that seems like overkill at this point.

## The console

### Steps
1. Click "comment" above a specific print statement.
2. Maybe enter some text (maybe not).
3. Click "comment" above a different print statement.

### Expected outcome
A new pending comment is created. Different print statements may or may not be related to each other.

I think it's easier to click a specific print statement/line than it is to click a precise x/y coordinate. (We also don't show the tooltip text "move the marker" here so...)

If this difference in behavior feels uncomfortable, (it does to me), then maybe that suggests that we need to add a feature to prompt users before erasing a pending comment with text in it. ("Would you like to save this comment?") This is only relevant so long as our data model only supports one pending comment at a time, so hopefully this discrepency goes away soon.

## The source code panel

### Steps
1. Click "comment" above a specific line of code.
2. Maybe enter some text (maybe not).
3. Click "comment" above a different line of code.

### Expected outcome
A new pending comment is created. Different lines of code are not related to each other.

(Same caveats as above.)

bvaughn:
There's a small bug I need to sort out before this PR is ready for review:
When a new (top level) comment is added, the UI shows a duplicate. I expect this is because we aren't cleaning up the "pending" comment on success, so `Transcript` pushes it onto the merged comments array.

src/ui/hooks/comments/useAddComment.tsx
```diff
@@ -32,7 +32,7 @@ export default function useAddComment() {
     return addComment({
       variables: {
         input: {
-          ...omit(comment, ["id", "createdAt", "updatedAt", "replies", "user"]),
+          ...omit(comment, ["id", "createdAt", "updatedAt", "isUnpublished", "replies", "user"]),
```
bvaughn, line 35: For now, this is a local-only field. Even once it's a field Apollo supports, we'll probably still want to filter it out (just like we filter out the id and dates) since Apollo will determine the correct value for comments that are successfully published.

src/ui/components/Transcript/Transcript.tsx
```diff
@@ -19,15 +18,13 @@ export default function Transcript() {
   const { isAuthenticated } = useAuth0();
 
   const displayedComments = useMemo(() => {
-    const displayedComments: Comment[] = [...comments].filter(
-      comment => !pendingComment || commentKey(comment) != commentKey(pendingComment.comment)
-    );
```
bvaughn, line 24: We don't need to do this dedupe pass anymore since we removed the optimistic cache update for new comments.

src/ui/hooks/comments/useAddComment.tsx
```diff
@@ -19,7 +19,8 @@ export default function useAddComment() {
           }
         }
       }
-    `
+    `,
+    { refetchQueries: ["GetComments"] }
```
bvaughn, line 23: When we save a new comment, we should refetch the list of comments too (so the comment doesn't disappear).

src/ui/hooks/comments/useAddComment.tsx
```diff
@@ -29,67 +30,22 @@ export default function useAddComment() {
...
-          data: newData,
-        });
-      },
+        onCompleted: (data: Comment | {}) => {
+          resolve(data);
+        },
+        onError: (error: ApolloError) => {
+          reject(error);
+        },
+        awaitRefetchQueries: true,
```
bvaughn, line 47: We should not consider the operation complete (aka we shouldn't invoke our `onCompleted` handler) until the comment has been saved _and_ the list has been refetched. Otherwise the new comment will disappear and reappear a second or two later.

**TODO* This doesn't seem to work as expected. Most of the time it does, but some times there will still be a few seconds delay where the comment disappears and then pops back in. I'm not sure why yet.

jcmorrow: Maybe this is complicated by our long-polling for comments? Whenever I have the phenomenon of "it goes away and comes back a few seconds later" it's because my initial update is failing and then long-polling eventually fixes it.

bvaughn: > it's because my initial update is failing and then long-polling eventually fixes it.

Can you describe this scenario a little more?

Why would an initial update fail silently with Apollo? Is that common/expected?

bvaughn:
I'm going to sit this PR down for a bit and work on something else. I need a break.

The `useMutation` hook _should_ refetch queries and hold on "completing" until the queries have been refetched. In practice, this seems to fail sometimes for reasons I'm not sure about. The resulting user experience is not great.

I also tried using the `optimisticResponse` option to temporarily move the pending comment into the Apollo cache but it exhibited the same behavior. Maybe we're doing something non-standard with the cache that's causing this? I'm not sure.

bvaughn:
Given the current broken state of comments (e.g. #6233 and #6502) I am half tempted to say this PR– even with its quirks– is already an improvement 😄  The case it _doesn't_ handle well is when you delete a few comments and then add a new one – before the Apollo client has finished syncing the deletion requests.

I'm going to try to dig back into that specific case today and see if I can figure out how to fix it.

jasonLaster (approved):

bvaughn:
This is ready for review 😄

src/ui/components/Comments/TranscriptComments/CommentCard.tsx
```diff
@@ -266,7 +279,8 @@ function CommentCard({
     <div
       className={classNames(
         `comment-card relative mx-auto w-full cursor-pointer border-b border-splitter transition`,
-        isOutsideFocusedRegion ? "opacity-30" : "bg-bodyBgcolor"
+        isOutsideFocusedRegion ? "opacity-30" : "bg-bodyBgcolor",
+        isUpdating ? "bg-themeBase-90" : ""
```
bvaughn, line 283: This is the style that makes the in-flight request _look_ disabled.

Separately, we were already hiding the actions "..." button.

src/ui/actions/comments.ts
```diff
@@ -58,13 +57,21 @@ export function createComment(
     const primaryLabel = labels?.primary;
     const secondaryLabel = labels?.secondary;
 
+    // If a comment is saved to Apollo, it will be assigned a persisted ID.
+    // Local, pending comments just need stable+unique IDs.
+    // Avoid using createdAt or updatedAt in these IDs though as these values are not stable.
+    const id = sourceLocation
```
jcmorrow, line 63: So, this is fine, but I just want to clear up some confusion around how `commentKey` works vs. what this change is doing. And a feelings disclaimer, I wrote that function and I feel like it's getting short-shrift here, and I know it's silly but I put a lot of work into it at the time and I feel a little defensive about it.

In the PR description it says:

> move away from using createdAt or updatedAt in keys. These values aren't stable and so they cause unnecessary unmounts and remounts within the React tree. (Even for comments that are persisted, Apollo will return a new createdAt value.)

It is true that after [this change](https://github.com/RecordReplay/devtools/commit/ab24d3f649689209e9ff3f2783ceb5b7ca878dbe#diff-9168052ad32419af17b20e0eaab3d17efed5744ba76105290e3d9c4b21d5295aR44) `commentKey`  no longer worked as intended, because it included an unstable value (`createdAt`) as you mention. However, previous to that change, this line, which generates an identifier for a comment by looking at its unique, stable identifiers, is *exactly* what `commentKey` was doing. Now, it's fine, we can delete all of that code and the accompanying tests, but I just want to point out that we already have all of this logic in a tested, (I think) well-named helpers file, and we are not moving away from that idea, but removing a regression introduced by the PR I linked above.

OK, sorry, rant over! I feel silly leaving this comment, but as I'm sure you can appreciate, it took up so much space in my brain when thinking about these problems, and I think we arrived at extremely similar solutions, which is great and gives me confidence in the direction that we are going (not because *I* was one of them, but because multiple people working independently made the same decisions).

markerikson (approved):
Based on what you showed in the last Loom, the approach seems reasonable overall.  Not much feedback on the code changes - you've clearly spent a lot of time thinking through this, and I could generally see the intent of the changes based on the video description.

src/ui/utils/comments.ts
```diff
@@ -36,14 +36,6 @@ export function formatRelativeTime(date: Date) {
   return "Now";
 }
 
-export function commentKeys(comments: (Comment | Reply)[]): string[] {
```
jcmorrow, line 39: if we are going to entirely get rid of these functions, I think you can delete `orderByApproximatelyCreatedAt`(was unused even before this) and `commentIdentifiers` as well.

jcmorrow (approved):
I think this all looks pretty good, I fully support throwing all optimistic Apollo updates into the 🔥 . I am a little sad to see us inlining functions that previously had names and tests, but I'm probably just being sentimental about my own code, I'll get over it 🙂

jcmorrow:
I actually have one more thing I want to ask about, making a replay now

jcmorrow:
My question was: doesn't the entire component unmount and remount upon persisting, because the ID changes? And the answer is yes (as captured and commented [here](https://app.replay.io/recording/comment-creation-and-ids--086ea96b-b3ff-40e2-99b9-a25af8089d09?point=83401268305296702943639935809227867&time=30317&hasFrames=false)), but because it happens at the same time that we transition to showing the full card with the user's name and avatar, it's a lot less obvious that the editor is completely unmounting and remounting.

bvaughn:
> I am a little sad to see us inlining functions that previously had names and tests, but I'm probably just being sentimental about my own code, I'll get over it 🙂

Ooh, which functions are you referring to? I don't think I "inlined" anything. I just removed the previous comment-key-mapping functions (and their tests) since we're now just using `comment.id` directly. Am I misunderstanding?

bvaughn:
> My question was: doesn't the entire component unmount and remount upon persisting, because the ID changes? And the answer is yes (as captured and commented [here](https://app.replay.io/recording/comment-creation-and-ids--086ea96b-b3ff-40e2-99b9-a25af8089d09?point=83401268305296702943639935809227867&time=30317&hasFrames=false)), but because it happens at the same time that we transition to showing the full card with the user's name and avatar, it's a lot less obvious that the editor is completely unmounting and remounting.

Yeah. Since we're keying off of ID, we do remount some things. That's also bound to happen anyway I think because the tree structure is different for a pending vs persisted comment. One uses `CommentCard` -> `CommentItem` and one uses `CommentCard` -> `PendingCommentCard`, so React would do an unmount and remount already

jcmorrow:
> Yeah. Since we're keying off of ID, we do remount some things. That's also bound to happen anyway I think because the tree structure is different for a pending vs persisted comment. One uses CommentCard -> CommentItem and one uses CommentCard -> PendingCommentCard, so React would do an unmount and remount already (even though the DOM happens to end up being similar).

Yeah, I'm not sure when that switch was made, but it was definitely very recently. That's why we used to use commentKey for all comments, even persisted ones, because we were using the same component for new and persisted comments, and the goal was for it to have absolutely no unmounting/remounting (which was really difficult to achieve!). But this sidesteps that whole problem and I think that's for the best.

src/ui/actions/comments.ts
bvaughn: Ooops, I was replying to your comments in order and replied to the one above before seeing this one (so I deleted my outer reply and I'm going to respond here directly).

I appreciate the disclaimer ❤️  I'm sincerely sorry if my comments or code changes seemed inconsiderate or dismissive. That was not my intent.

When I stepped into this code, I was _really_ confused and it took me a lot of digging and notes-taking to wrap my brain around how it worked. A lot of the comments I left on this PR (and the related issues) were my best attempt and sorting through the mental space and organizing my thoughts around it.

You're right in pointing out that the new ID is the very similar to the previous one, minus the date. And yeah, if I look at an older commit, without the date, it's the same. So maybe I should find another way to (re)add tests for the spirit of what was there before.

Even though the code is doing a very similar thing, I think this is a small but nice change still because:
* Unique ID rather than a "magic" one.
* Setting `key={id}` rather than `key={someMappingFunction(object)}` is a more standard usage with React components.

Again though, I understand how this might have made you feel not great and I'm really sorry for that.

src/ui/components/Comments/TranscriptComments/CommentCard.tsx
bvaughn: Oooh, I wonder if that was b'c the previous change was opacity or because @jonbell-lot23 didn't like the style treatment similarity?

I'll follow up with Jon on this. Thank you!

bvaughn:
I'm going to write a follow up issue or two for things in the PR description, and think about how I can revive some of the comment ordering tests (using our new test helper utils)– but I'm going to land this first since it fixes a few paper cuts that would be really nice to have fixed for users.

I'll also follow up with Jon RE the styling.

src/ui/actions/comments.ts
bvaughn: Will re-add some comment ordering tests in #6567

PR #6337, opened 2022-04-12
Fixed the tooltip to center with the button #6331

### LINKED ISSUE #6331: Tooltips for sidebar buttons are not vertically centered with the icon
image

The tooltip looks like it's anchored to the bottom edge of the icon container instead of being centered vertically.

Hoshhh:
Setting the margin top to half of the original value seems to have centered it for all of the toolbar buttons. If you would like me to look for an alternative solution, id be glad to try!

Fix #6331.


jaril (approved):

Hoshhh:
@jaril Not a problem, we all get busy sometimes! I just appreciate the opportunity. Seems like an awesome project!

PR #6350, opened 2022-04-13
Use the heatmap hit counts in the breakpoint panel's status's denominator

### LINKED ISSUE #6356: Don't show the 10k error in the panel summary if the error is specifically not a 10k analysis error
https://app.replay.io/recording/replay-breakpoints-broken--419bbc81-6768-476e-86d5-21a44af45856?point=63281118031871153968300277358397405&time=29714&hasFrames=true#

Analyses can fail. If it has more than 10k+ hits, we consider it failed. But there are also other reasons for it to fail. In those cases, we're currently applying the 10k error treatment to the UI even though that might not apply. 

Jason debugs it [here](https://www.loom.com/share/383d9fb76ccd495d9d1baad14a58a8c0).

Mark from Glide ran into that problem so let's fix that for him.

jaril:
This PR does a few things:
- Abstracts away the hits implementation (heatmaps or analyses) by extracting the logic for identifying the number of hits, error/load states into a new `useGetHitsForPanel` hook. That's then used by all the panel's children components so that we can rely on the data being presented to be consistent, instead of having different ways to evaluate hits/loading/error.
- Allows us to show the hits as the breakpoint status's denominator separately, even if we may not have the exact position (numerator) available immediately.
- TS-ifies the panel component
- Moves us from react-redux's `connect` to `useSelector`/`useDispatch` where possible.

It will mainly fix #6356, but generally, this should keep us from having cases where we're displaying data inconsistently in the breakpoint panel.


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const useGetHitsForPanel = (breakpoint: Breakpoint) => {
+  const analysisPoints = useSelector((state: UIState) => getAnalysisPointsForBreakpoint(state, breakpoint));
+  const hitCounts = useSelector(getHitCountsForSelectedSource)
+  const hitCountFromSources = hitCounts?.filter(
+    hitCount => hitCount.location.line === breakpoint.location.line
+  )[0];
+
+  const hits = analysisPoints?.data.length || hitCountFromSources.hits || 0;
```
markerikson, line 40: :hand: Outta curiosity, is it better to use `??` here to only fall back if those are `null | undefined`, rather than "falsy/`0`"?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -7,49 +7,58 @@ import classnames from "classnames";
...
+  breakpoint: Breakpoint;
+  editor: CodeMirror;
+  insertAt: number;
 }) {
+  const dispatch = useDispatch();
   const [editing, setEditing] = useState(false);
   const [showCondition, setShowCondition] = useState(Boolean(breakpoint.options.condition)); // nosemgrep
-  const [width, setWidth] = useState(getPanelWidth(editor)); // nosemgrep
-  const [inputToFocus, setInputToFocus] = useState("logValue");
+  const [width, setWidth] = useState(getPanelWidth(editor.editor)); // nosemgrep
```
markerikson, line 58: :speech_balloon:  heh, I'm gonna have to remember `// nosemgrep` :)

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -7,49 +7,58 @@ import classnames from "classnames";
...
-
   return editor.getScrollInfo().clientWidth - panelIndent;
 }
 
-function Panel({
-  analysisPoints,
+const useGetHitsForPanel = (breakpoint: Breakpoint) => {
+  const analysisPoints = useSelector((state: UIState) => getAnalysisPointsForBreakpoint(state, breakpoint));
+  const hitCounts = useSelector(getHitCountsForSelectedSource)
+  const hitCountFromSources = hitCounts?.filter(
```
markerikson, line 36: :hand: This should probably be a `.find()` instead

jaril:
Nobody look at this yet!!

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -1,35 +1,58 @@
...
 import { getExecutionPoint } from "../../../selectors";
+import { HitsContext } from "../../Editor/Breakpoints/Panel/Panel";
+
+const createStatusString = (
+  hits: number | null,
+  executionPoint: string | null,
+  analysisPoints: AnalysisPayload
+) => {
+  const denominator = hits;
+  let numerator = "…";
```
bvaughn, line 14: What does a status string like "…/2" mean? Should we maybe just return "0/2" in that case? or even just "…"?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const useGetHitsForPanel = (breakpoint: Breakpoint) => {
+  const analysisPoints = useSelector((state: UIState) =>
+    getAnalysisPointsForBreakpoint(state, breakpoint)
+  );
+  const hitCounts = useSelector(getHitCountsForSelectedSource);
+  const hitCountFromSources = hitCounts?.find(
+    hitCount => hitCount.location.line === breakpoint.location.line
+  );
```
bvaughn, line 59: How often do we think this hook will run and with how many `hitCounts` matches? (Wondering if this should be memoized.)

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const loading = analysisPoints === null || !hitCountFromSources;
+  const hits = analysisPoints?.data.length ?? hitCountFromSources?.hits ?? null;
```
bvaughn, line 62: Ooo, this is interesting. Two sources of truth for the number of hits. I don't think I understand the surrounding code well enough to have an informed opinion on this but it jumps out to me.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
jcmorrow: I would vote `...` or `?/2`

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
jcmorrow: it will only run until the first match, but there could be hundreds or thousands of `hitCounts` entries to iterate over).

bvaughn: Definitely seems worth. memoizing then rather than doing every render.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
jcmorrow: The summary here is that there are two ways of knowing how many times a particular location was hit. The first, older, and more thorough way is by running an analysis on that location. This will only return results for the given location, and it will also include tons of other information about the hits, not just the count. The second, newer, but sometimes more finicky route is using `heatMaps` (under the hood the `getHitCounts` [protocol message](https://static.replay.io/protocol/tot/Debugger/#method-getHitCounts). We switch between the two based on:

- Features the user has enabled (you can turn heatMaps off)
- How old the recording is (older recordings don't support heatMaps)
- How many locations the file will query for (files with thousands of locations can trigger an error on the backend when querying for heatMaps)

In all cases we fall back to running analyses.

jcmorrow:
The more I think about this, the more I think it's probably not a change worth making. Here's my thinking:

- In a perfect world, these numbers always agree, so it shouldn't matter which one you use, and trying to favor one over the other is complexity that doesn't add anything functionally.
- In a world where these numbers disagree:
  - It is almost certainly `heatMaps` which is incorrect
  - Favoring heatMaps means the hits will agree, but you might end up with other broken scenarios like `100/50` hits, which makes no sense

I think we should continue to use analyses numbers for the breakpoint panel, and when we see disagreements we should file issues and use those as an opportunity to improve the heatMaps feature.

bvaughn:
That reasoning makes sense.

Issue #5384, opened 2022-02-18
Source does not exist


jasonLaster:
Here is a replay of the problem
https://app.replay.io/recording/what-is-this-sourcemap-error--1363310a-7d37-4f4c-9be9-1db1bc1d1bd2?point=18497557561419090466109470070014025&time=7821.187132625835&hasFrames=true

This commit caused the regression https://github.com/RecordReplay/devtools/commit/4856970341ac4e2e078cd28a20ad17e228878d32

PR #6357, opened 2022-04-13
Remove source-queue (#5384)

hbenl:
Recently we've seen several "source xyz does not exist" errors, these were caused by a source queue in which sources were collected before being added to the redux store, so requesting a source from the store could fail even after receiving all sources from the backend because it was still waiting in the queue.
We've inherited that queue from Firefox devtools, but it doesn't make any sense here because we're waiting for all sources to be received from the backend before adding them to the store anyway. So the queue only added an unnecessary delay (in addition to causing the "source does not exist" bug).


markerikson (approved):

jaril:
https://sentry.io/organizations/replay/issues/3188330864/?project=5399075&query=is%3Aunresolved

PR #6358, opened 2022-04-13
Only instantiate the apolloclient once

jaril:
Replay of the changes: https://app.replay.io/recording/upload-screen-check--c0a243d3-6d6a-451b-9fc5-412e1b25b0aa?point=3458764513820540931&time=94&hasFrames=false 
Walkthrough of changes: https://www.loom.com/share/9be50eab1d5d409497c26333e31e609d

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -133,17 +133,6 @@ interface AuthProps {
   apiKey?: string;
 }
 
-// We need to ensure that we always pass the same handleAuthError function
-// to ApolloWrapper, otherwise it will create a new apolloClient every time
-// and parts of the UI will reset (https://github.com/RecordReplay/devtools/issues/6168).
-// But handleAuthError needs access to the current values from useAuth0(),
-// so we use a constant wrapper around the _handleAuthError() function that
-// will be recreated with the current values.
-let _handleAuthError: () => Promise<void>;
-function handleAuthError() {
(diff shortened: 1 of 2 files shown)


hbenl:
I'm not sure this will work correctly if there is an auth error: you create the apollo client with the first instance of `handleAuthError()`, which captures `isAuthenticated` from the first call to `AppUtilities`. If the value of `isAuthenticated` changes in later calls to `AppUtilities`, that change never makes it into the apollo client, i.e. if there is an auth error later, `handleAuthError()` will get called but it will still use the initial value of `isAuthenticated`. (I don't know in which situations `handleAuthError()` is called, so I'm not sure this could actually lead to wrong behavior)

jaril:
> I'm not sure this will work correctly if there is an auth error: you create the apollo client with the first instance of `handleAuthError()`, which captures `isAuthenticated` from the first call to `AppUtilities`. If the value of `isAuthenticated` changes in later calls to `AppUtilities`, that change never makes it into the apollo client, i.e. if there is an auth error later, `handleAuthError()` will get called but it will still use the initial value of `isAuthenticated`. (I don't know in which situations `handleAuthError()` is called, so I'm not sure this could actually lead to wrong behavior)

I agree it does seem somewhat flimsy. I'll just leave it as is for now then.

PR #6371, opened 2022-04-13
Migrate all remaining existing reducers to TS

markerikson:
This PR:

- MIGRATES THE REST OF OUR REDUCERS TO TS YAYYYY!
- Removes a dependency on `lodash-move`, which was only used in one file, and inlines a copy of `array-move` instead


src/devtools/client/debugger/src/reducers/file-search.ts
```diff
@@ -92,9 +92,6 @@ function update(state = createFileSearchState(), action: AnyAction) {
   }
 }
 
-// NOTE: we'd like to have the app state fully typed
-// https://github.com/firefox-devtools/debugger/blob/master/src/reducers/sources.js#L179-L185
```
jcmorrow, line 96: 🎉

jcmorrow (approved):
Super exciting!

PR #6386, opened 2022-04-14
Switch to getEventHandlerCount fixes #6369

### LINKED ISSUE #6369: Feature flag event listener counts
We currently send a find analysis points call for each event listener. We should put this logic behind a feature flag and disable it for now by default. 

As a follow up, we can add a new protocol method that can return the counts upon request which will be much faster and less strain on the backend

jasonLaster:
Fixes #6369

Pretty excited about the performance win here

Screen Shot 2022-04-13 at 8 39 47 PM

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-//
-
 import React, { Component } from "react";
 import classnames from "classnames";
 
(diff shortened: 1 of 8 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/EventListeners.js
```diff
@@ -293,15 +289,15 @@ class EventListeners extends Component {
...
 
   renderListenerEvent(event, category) {
-    const { activeEventListeners, eventTypePoints } = this.props;
+    const { activeEventListeners, eventTypeCounts } = this.props;
     const { searchText } = this.state;
 
-    const points = eventTypePoints[event.id];
-    if (!points || points.length == 0) {
+    const count = eventTypeCounts[event.id];
+    if (!count || count == 0) {
```
jcmorrow, line 296: The `|| count == 0` won't do anything here since `0` is falsey.

jcmorrow (approved):
Nice 💯

PR #6387, opened 2022-04-14
Switch to getEventHandlerCount fixes #6369

### LINKED ISSUE #6369: Feature flag event listener counts

jasonLaster:
Fixes #6369

Pretty excited about the performance win here

Screen Shot 2022-04-13 at 8 39 47 PM


jcmorrow (approved):
Sweeeet 👍

PR #6394, opened 2022-04-14
Migrate remaining selectors to TS

markerikson:
This PR:

- Migrates all of our remaining selector files to TS

The only exceptions I can find are a couple of files for "changes" and "markup" buried down in `client/inspector/`, and those don't seem very relevant.

This should help improve type safety in all our components going forward.

I do see that we've still got some stubbed-out function declarations in `selectors/index.ts` that I had moved over from the prior `index.d.ts` file.  I tried removing those and got compile errors, so I'm leaving a `// TODO` and bailing on that for now.


src/devtools/client/debugger/src/reducers/types.ts
```diff
@@ -83,10 +83,10 @@ export type SourceActorLocation = {
...
  */
 export type SourceLocation = {
-  readonly sourceId: SourceId;
-  readonly line: number;
-  readonly column?: number;
-  readonly sourceUrl?: string;
+  sourceId: SourceId;
+  line: number;
+  column?: number;
+  sourceUrl?: string;
```
bvaughn, line 89: Seems like it would be nice to keep these attributes `readonly` and just update the way this code creates the location object to not rely on mutation:
https://github.com/RecordReplay/devtools/blob/f314b7b9804aa7a99c9718682ff2d8dde26b8aa8/src/devtools/client/debugger/src/utils/breakpoint/index.js#L59-L67

aka
```js
export function makeBreakpointLocation(state, location) {
  const source = getSource(state, location.sourceId);
  if (!source) {
    throw new Error("no source");
  }

  let sourceUrl;
  let sourceId;

  if (source.url) {
    sourceUrl = source.url;
  } else {
    sourceId = getSourceActorsForSource(state, source.id)[0].id;
  }

  return {
    line: location.line,
    column: location.column,
    sourceUrl,
    sourceId,
  };
}

```

From a quick look, it seems like that's the only place we'd need to change.

We don't _really_ want locations being mutated. That could break React component functionality.

bvaughn (approved):
I just quickly scanned this but LGTM.

src/devtools/client/debugger/src/reducers/types.ts
markerikson: FWIW, the only reason there's a `readonly` on those fields at all is because I just generated all the TS types in that file from the original Flow types in the Mozilla codebase.

I don't think any of our hand-written types actually use `readonly` at all.

Another consideration is that with Redux Toolkit, we _do_ want to allow "mutating" syntax for values, because Immer lets us do that safely.  Not saying it matters in this particular case, but as a whole we'd want to define most types as standard fields rather than `readonly` fields because of that.

That said, I like the proposed change here, and I'll go ahead and make that.

bvaughn: > Another consideration is that with Redux Toolkit, we do want to allow "mutating" syntax for values, because Immer lets us do that safely.

Let's chat about this sometime in realtime?

My concern is that these types aren't restricted to scopes inside of an Immer callback, and so it would be nice if they warned about unsafe mutations (since mutations are _generally_ unsafe within the context of React).

PR #6399, opened 2022-04-14
Fix comment badge behavior

jaril:
The comment badge was showing regardless of whether or not there was a comment. This fixes that.

[Replay of the changes](https://app.replay.io/recording/todomvc-comment-badge--9aff7784-5af2-4ee5-aff1-a60b4b57e269)

### DIFF
--- a/src/ui/components/Toolbar.tsx
+++ b/src/ui/components/Toolbar.tsx
@@ -1,18 +1,18 @@
 import React, { useEffect, useState } from "react";
 import classnames from "classnames";
-import { connect, ConnectedProps, useDispatch, useSelector } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
 import { actions } from "../actions";
 import { selectors } from "../reducers";
-
 import IconWithTooltip from "ui/components/shared/IconWithTooltip";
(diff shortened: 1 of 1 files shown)


src/ui/components/Toolbar.tsx
```diff
@@ -88,13 +83,21 @@ export default function Toolbar() {
   const isPaused = useSelector(selectors.hasFrames);
   const viewMode = useSelector(selectors.getViewMode);
   const selectedPrimaryPanel = useSelector(getSelectedPrimaryPanel);
-  const [hideCommentsBadge, setHideCommentsBadge] = useState(false);
+  const [showCommentsBadge, setShowCommentsBadge] = useState(false);
+  const recordingId = useGetRecordingId();
+  const { comments, loading } = hooks.useGetComments(recordingId);
+
+  useEffect(() => {
```
markerikson, line 90: :hand: It feels like more of this should be "derived state" rather than necessarily setting state in a `useEffect`, but I'd have to think about the intended behavior here more.

markerikson (approved):
The `useEffect` logic feels a bit questionable conceptually, but might as well fix the actual bug now

PR #6408, opened 2022-04-15
Add CMD+F (search) to Console

### LINKED ISSUE #3834: Console search (cmd+f)
### Discussed in https://github.com/RecordReplay/devtools/discussions/3803

<div type='discussions-op-text'>

<sup>Originally posted by **jasonLaster** October  1, 2021</sup>
The console should have a simple cmd+f search ui</div>

bvaughn:
Resolves #3834

* **UI demo**: https://www.loom.com/share/dfcb19ef243f423d8e6aa95c78613939 
* **Architecture walk through**: https://www.loom.com/share/0e9ea476fc3148f68bd396ccfd70919d

## MVP changes
- [x] Implement basic search functionality.
- [x] Highlight the console row that contains the currently-focused search result.
- [x] Add unit tests for `useSearch` and `useConsoleSearch` hooks.

## Follow up changes
- [ ] Refocus the input inside of `JSTerm` when the CMD+F search input is hidden. (Currently this is refocused if you click the "x" to hide the search, but not if you hide via the "Escape" key.) I don't think this is straight forward to change because it requires an _imperative_ method call on a DOM element rendered deep inside of the "react-codemirror2" library we are using.
- [ ] Wrap subtext matches with `<mark>` tags to highlight (and highlight the currently "selected" match in a special way). 
   * Note that the pre-existing searches, like find-in-file, only highlight the _current_ result– but I think it would be better UX to highlight all matches– with a highlight that updates as you type, but a current cursor that only jumps when you type "Enter").

## Design considerations
* Search is done in the main UI thread to avoid the heavy serialization cost of a data source (console messages) that potentially changes often. (Search-in-file functionality is implemented in a worker, but file content is _static_ for the session.) If this is problematic, we could look into an amortization strategy that only sends _message diffs_ to a worker, but I think the overhead would still potentially not be worth the effort. There's also the precedent of filtering console rows in the main thread for things like the selected timeline range, loaded sections, and top-filter text.)
* The console search state is not global. It's scoped to a small part of the Replay component tree. Because of this, I choose to store it in React state rather than Redux.
* Replay already depends on an experimental version of the "react" NPM package because of [react-devtools-inline](https://github.com/facebook/react/blob/main/packages/react-devtools-inline/README.md#react-devtools-inline). Because of this, I decided to go ahead and build on top of the `useDeferredValue` hook to prepare for _when_ we update Replay to use the [`createRoot` API](https://reactjs.org/docs/react-dom-client.html#createroot).
  * Note this design decouples the query text and current results such that it's expected a component may temporarily render a query that doesn't match its results. This is being done _intentionally_ so that the search input feels responsive and updates quickly, leaving the slower search work to be done in a lower priority update. (Note that this won't matter so long as we use the legacy `render` method.)
* The bulk of the search login has been implemented as a resuable `useSearch` hook (with some basic unit tests). The console-specific Message search logic is implemented in a small `useConsoleSearch` decorator hook. (Neither hook is very large but this pattern should allow us to build additional searches more easily if we wanted to.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -161,6 +161,7 @@
     "fs-extra": "^10.0.1",
     "graphql": "15.8.0",
     "graphqurl": "^1.0.1",
+    "identity-obj-proxy": "^3.0.0",
     "mini-css-extract-plugin": "^2.6.0",
     "node-polyfill-webpack-plugin": "^1.1.4",
     "postcss": "^8.4.12",
--- a/src/devtools/client/debugger/src/components/shared/SearchInput.js
+++ b/src/devtools/client/debugger/src/components/shared/SearchInput.js
@@ -32,6 +32,7 @@ class SearchInput extends Component {
   $input;
 
   static defaultProps = {
+    className: "",
     expanded: false,
     hasPrefix: false,
     selectedItemId: "",
@@ -183,6 +184,7 @@ class SearchInput extends Component {
 
   render() {
     const {
+      className,
       expanded,
       handleClose,
       onChange,
@@ -214,7 +216,7 @@ class SearchInput extends Component {
     };
 
     return (
-      <div className="search-outline">
+      <div className={`search-outline ${className || ""}`}>
         <div
           className={classnames("search-field rounded-lg", size)}
           role="combobox"
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -16,6 +16,8 @@ import { MessageContainer } from "devtools/client/webconsole/components/Output/M
 import ConsoleLoadingBar from "./ConsoleLoadingBar";
 
 import constants from "devtools/client/webconsole/constants";
+import { StateContext } from "../Search";
+import { Message } from "../../reducers/messages";
 
 function mapStateToProps(state: UIState) {
   return {
@@ -44,10 +46,16 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
 class ConsoleOutput extends React.Component<PropsFromRedux> {
   outputNode = React.createRef<HTMLDivElement>();
 
+  _scrollTimeoutID: number | null = null;
+  _prevSearchResultMessage: Message | null = null;
+
+  static contextType = StateContext;
+
   componentDidMount() {
     if (this.props.visibleMessages.length > 0) {
       scrollToBottom(this.outputNode.current!);
     }
+    this.scrollCurrentSearchResultIntoView();
   }
 
   componentDidUpdate(prevProps: PropsFromRedux) {
@@ -59,17 +67,12 @@ class ConsoleOutput extends React.Component<PropsFromRedux> {
     }
 
     this.maybeScrollToResult(prevProps);
+    this.scrollCurrentSearchResultIntoView();
   }
 
-  scrollToClosestMessage() {
-    const { closestMessage } = this.props;
-
-    if (!closestMessage) {
-      return;
-    }
-
+  ensureMessageIsInView(message: Message) {
     const element: HTMLElement = this.outputNode.current!.querySelector(
-      `.message[data-message-id="${closestMessage.id}"]`
+      `.message[data-message-id="${message.id}"]`
     )!;
 
     if (!element) {
@@ -81,9 +84,43 @@ class ConsoleOutput extends React.Component<PropsFromRedux> {
       return;
     }
 
+    if (this._scrollTimeoutID) {
+      clearTimeout(this._scrollTimeoutID);
+    }
+
     // Chrome sometimes ignores element.scrollIntoView() here,
     // calling it with a little delay fixes it
-    setTimeout(() => element.scrollIntoView({ block: "center", behavior: "smooth" }));
+    this._scrollTimeoutID = setTimeout(() =>
+      element.scrollIntoView({ block: "center", behavior: "smooth" })
+    );
+  }
+
+  scrollCurrentSearchResultIntoView() {
+    const { currentIndex, results, visible } = this.context;
+
+    if (!visible) {
+      return;
+    }
+
+    const message =
+      currentIndex >= 0 && currentIndex < results.length ? results[currentIndex] : null;
+
+    // Only programmatically scroll after changes to the selected search result.
+    if (this._prevSearchResultMessage === message || message === null) {
+      return;
+    }
+
+    this._prevSearchResultMessage = message;
+
+    this.ensureMessageIsInView(message);
+  }
+
+  scrollToClosestMessage() {
+    const { closestMessage } = this.props;
+
+    if (closestMessage) {
+      this.ensureMessageIsInView(closestMessage);
+    }
   }
 
   maybeScrollToResult(prevProps: PropsFromRedux) {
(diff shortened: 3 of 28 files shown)


package.json
```diff
@@ -161,6 +161,7 @@
     "fs-extra": "^10.0.1",
     "graphql": "15.8.0",
     "graphqurl": "^1.0.1",
+    "identity-obj-proxy": "^3.0.0",
```
bvaughn, line 164: Our Jest config was setup to use this plugin but we didn't actually have it as a dependency. It just so happened that we have so few tests that none of them had hit this path before I added a CSS module in the new `ConsoleSearch` component.

src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+    TestComponent = () => {
+      const [state, actions] = useConsoleSearch();
+      lastCommitedSearchActions = actions;
+      lastCommitedSearchState = state;
+
+      return null;
+    };
+  });
+
+  // TODO [bvaughn] This is pretty gross; can we collocate this (closer to the Redux code if nothing else)?
```
bvaughn, line 38: cc @markerikson Any thoughts about this kind of test?

src/devtools/client/webconsole/selectors/messages.ts
```diff
@@ -35,15 +35,30 @@ export const getVisibleMessages = createSelector(
...
+      // Filter out messages that aren't within the focused region.
+      if (focusRegion) {
+        if (!isInTrimSpan(executionPointTime, focusRegion)) {
+          return false;
+        }
+      }
+
+      // Filter out messages that haven't yet been loaded.
+      if (
+        !loadedRegions!.loaded.some(
```
jcmorrow, line 53: Kind of a nit but I think this is the same as saying 

```
return loadedRegions!.loaded.some(region => executionPointTime >= region.begin.time && executionPointTime <= region.end.time);
```

Which it might be nice to pull out into a `isInLoadedRegions` function, since that's a pretty reasonable check that we might use elsewhere. Actually, maybe a better signature would be like `isInRegions(point, regions)`?

jcmorrow (approved):
👏  Really happy to see this get added *with tests*! I think all of my comments are style-related, feel free to take or leave!

src/devtools/client/webconsole/components/Search/ConsoleSearchContext.tsx
```diff
@@ -0,0 +1,8 @@
+import { createContext } from "react";
+
+import type { Actions, State } from "./useConsoleSearch";
+
+// Storing state and dispatch in separate contexts allows parts of the tree to access the (state) dispatch function
+// without re-rendering when the state value changes.
+export const ActionsContext = createContext<Actions>(null as any);
```
jaril, line 7: _takes notes_ 👍

src/devtools/client/webconsole/components/Search/useConsoleSearch.ts
```diff
@@ -0,0 +1,67 @@
...
+import { getMessages } from "../../selectors";
+
+import type { Actions as SearchActions, State as SearchState } from "./useSearch";
+
+function search(query: string, messages: Message[]): Message[] {
+  const results: Message[] = [];
+
+  const needle = query.toLocaleLowerCase();
+  messages.forEach(message => {
+    console.log(message.id, 'messageText:', message.messageText);
```
jaril, line 14: ✂️

src/devtools/client/webconsole/components/Search/useConsoleSearch.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (typeof message.messageText === "string" && message.messageText.toLocaleLowerCase().includes(needle)) {
+      results.push(message);
+    } else {
+      message.parameters?.some(parameter => {
+        console.log(message.id, 'parameter:', parameter);
```
jaril, line 19: ✂️

jaril (approved):
LGTM and played around with it a bit — though trunk format/removing console logs pending.

The refocus thing was the only thing I noticed that was off but that's already addressed in follow up items (and id rather see this shipped sooner than later!)

src/devtools/client/debugger/src/components/shared/SearchInput.js
```diff
@@ -214,7 +216,7 @@ class SearchInput extends Component {
     };
 
     return (
-      <div className="search-outline">
+      <div className={`search-outline ${className || ""}`}>
```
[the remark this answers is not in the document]
bvaughn: I don't think I understand this bit of feedback. Don't we normally use it in the same way as it's used here?

Do you have a suggested alternative name?

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -151,36 +187,29 @@ class ConsoleOutput extends React.Component<PropsFromRedux> {
...
-          isPaused: closestMessage?.id == messageId,
-          isFirstMessageForPoint: this.getIsFirstMessageForPoint(i, visibleMessages),
-          isPrimaryHighlighted,
-          shouldScrollIntoView,
-        });
+    const messageNodes = visibleMessages.map((messageId, i) => {
+      const message = messages.entities[messageId]!;
+      // @ts-ignore ExecutionPoint/string mismatch
+      const isPrimaryHighlighted = hoveredItem?.point === message.executionPoint;
+      const shouldScrollIntoView = isPrimaryHighlighted && hoveredItem?.target !== "console";
```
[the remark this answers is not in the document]
bvaughn: Looks like we store "hovered item" in the Redux state (`state.timeline.hoveredItem`). FWIW I didn't write this, just moved the surrounding code a bit.
https://github.com/RecordReplay/devtools/blob/1b65b594016980128185df9b7baaf093e5d04e36/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx#L154-L183

My initial guess is that this check is to prevent the logs from scrolling/jumping while the user is hovering over the console (perhaps inspecting or selecting some logged text) but as I log that value, it always appears to be `undefined` when I'm I test things in the console so 🤷🏼 

cc @hbenl (45ddaa2ad6ab521ad4fdd347b746c9fd65dc8adf) for a better explanation maybe?

src/devtools/client/webconsole/components/Output/MessageContainer.js
```diff
@@ -33,6 +36,8 @@ export class MessageContainer extends React.Component {
     };
   }
 
+  static contextType = StateContext;
```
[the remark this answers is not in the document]
bvaughn: It's the newest [React Context API](https://reactjs.org/docs/context.html#classcontexttype) that still supports class components (like `MessageContainer`). Function components would just use the [`useContext` hook](https://reactjs.org/docs/hooks-reference.html#usecontext).

src/devtools/client/webconsole/components/Output/message-types/ConsoleCommand.js
```diff
@@ -31,7 +33,7 @@ export default function ConsoleCommand(props) {
     source,
     type,
     level,
-    topLevelClasses: [],
+    topLevelClasses: topLevelClassName ? [topLevelClassName] : [],
```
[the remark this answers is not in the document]
bvaughn: What is `[].compact()` ? Do you mean something like `[...].filter(i => i)`?

I have a slight preference to avoid unnecessarily using array manipulation methods.

src/devtools/client/webconsole/components/Output/message-types/EvaluationResult.js
```diff
@@ -17,14 +17,23 @@ EvaluationResult.propTypes = {
...
 };
 
 EvaluationResult.defaultProps = {
   open: false,
 };
 
 export default function EvaluationResult(props) {
-  const { dispatch, message, timestampsVisible, maybeScrollToBottom, open, isPaused } = props;
+  const {
+    dispatch,
```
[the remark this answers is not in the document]
bvaughn: Generally I have this preference to!

I think I'd argue _against_ it in this case for _pre-existing code_ because it causes churn.

I'd love for us to _standardize_ this (with an auto-fix Lint rule) in which case I'd be thrilled to do it.

src/devtools/client/webconsole/components/Search/useConsoleSearch.ts
bvaughn: 🤡

src/devtools/client/webconsole/components/Search/useConsoleSearch.ts
bvaughn: 🤡 🤡

src/devtools/client/webconsole/selectors/messages.ts
bvaughn: > Kind of a nit but I think this is the same as saying

Yeah, definitely. I know this sort of thing is totally subjective, but I kind of find the format I used to be a little more readable in terms of the sequence of conditionals that we check.

I don't feel strongly over this though.

🧐

https://github.com/RecordReplay/devtools/blob/1b65b594016980128185df9b7baaf093e5d04e36/src/ui/reducers/app.ts#L359-L363

bvaughn:
I think Typecheck and Vercel tasks are failing after rebase because of #6413

src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
markerikson: Heh, my brain's fried enough from being social at the conf today I'm not sure I can come up with specific feedback atm :)

Totally agreed that we're going to want a better testing environment going forward. There's just so much complexity around the store setup and data loading in the codebase atm.

src/devtools/client/webconsole/components/Search/ConsoleSearch.module.css
```diff
@@ -0,0 +1,11 @@
+.SearchBar {
+  width: 100%;
+  height: var(--editor-searchbar-height);
+  display: flex;
+  align-items: center;
+  border-top: 1px solid var(--theme-splitter-color);
```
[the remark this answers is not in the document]
bvaughn: Follow up PR #6450

PR #6412, opened 2022-04-16
Remove unnecessary console logging

bvaughn:
These log statements are adding a lot of noise to the console unnecessarily. It's getting in the way of my local logs.

### DIFF
--- a/src/protocol/socket.ts
+++ b/src/protocol/socket.ts
@@ -120,7 +120,6 @@ const doSend = makeInfallible(msg => {
 });
 
 function onSocketOpen() {
-  console.log("Socket Open");
   gPendingMessages.forEach(msg => doSend(msg));
   gPendingMessages.length = 0;
   gSocketOpen = true;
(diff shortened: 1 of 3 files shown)


markerikson:
I felt the same way, but @bhackett1024 specifically requested that we keep those for debugging purposes.

bhackett1024:
I would like to keep these in, but I also feel like I'm the only one.  Can we gate them behind a setting maybe?  The LoadedRegions one especially is handy for understanding precisely what the backend is doing.

bvaughn:
Seems odd to me to leave console logging on for everyone for a niche use case like this. Can we maybe move it behind a flag and at least log to `console.debug` so I can hide it by turning off "verbose" logging? :)

bvaughn:
Also which file(s) are we talking about? all of them?

bhackett1024:
> Seems odd to me to leave console logging on for everyone for a niche use case like this. Can we maybe move it behind a flag or at least log to `console.debug` so I can hide it by turning off "verbose" logging? :)

Sure either of these is fine.  Deleting the logs besides the LoadedRegions one is also fine.

bvaughn:
Great! I'm happy with that :) I'll remove the other two and change the loaded regions one to `console.debug` so it's there, but hidden by default :)

PR #6413, opened 2022-04-16
Specify load point when creating sessions

bhackett1024:
For https://github.com/RecordReplay/backend/issues/5169.  The backend now supports specifying a load point when creating sessions.  If the recording is still in progress, it will only be loaded up to the specified point.  If the recording is not in progress this has no effect, and since the only way we can get points for a recording that is still in progress is in node with the `process.recordreplay.currentPointURL()`, this should only affect that use case.

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -161,8 +161,11 @@ export function createSession(recordingId: string): UIThunkAction {
 
       dispatch(showLoadingProgress());
 
+      const loadPoint = new URL(window.location.href).searchParams.get("point");
+
       const { sessionId } = await sendMessage("Recording.createSession", {
         recordingId,
+        loadPoint,
         experimentalSettings,
       });
 
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -165,7 +165,7 @@ export function createSession(recordingId: string): UIThunkAction {
 
       const { sessionId } = await sendMessage("Recording.createSession", {
         recordingId,
-        loadPoint || undefined,
+        loadPoint: loadPoint || undefined,
         experimentalSettings,
       });
 
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -165,7 +165,7 @@ export function createSession(recordingId: string): UIThunkAction {
 
       const { sessionId } = await sendMessage("Recording.createSession", {
         recordingId,
-        loadPoint,
+        loadPoint || undefined,
         experimentalSettings,
       });
 


jasonLaster (approved):
Looks good.

There's a small TS issue with `ExperimentalSettings` defined above

```

src/ui/actions/session.ts(168,9): error TS2345: Argument of type '{ recordingId: string; loadPoint: string | undefined; experimentalSettings: ExperimentalSettings; }' is not assignable to parameter of type 'createSessionParameters'.
```

jaril (approved):

bvaughn:
Think this broke Flow on main branch:
```
src/ui/actions/session.ts:168:9 - error TS2345: Argument of type '{ recordingId: string; loadPoint: string | undefined; experimentalSettings: ExperimentalSettings; }' is not assignable to parameter of type 'createSessionParameters'.
  Object literal may only specify known properties, and 'loadPoint' does not exist in type 'createSessionParameters'.

168         loadPoint: loadPoint || undefined,
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

bhackett1024:
Huh, the package protocol is already updated but I don't know how to update the devtools so that it will be used and get this type check to pass.

bvaughn:
Seems like maybe the type change from this PR hasn't made it into devtools yet?

https://github.com/RecordReplay/backend/pull/5209

bvaughn:
I thought maybe we needed to run...
```sh
HASURA_KEY=<key> npm run gql
```

...from this page:
https://github.com/RecordReplay/devtools/blob/master/src/graphql/README.md

But that doesn't seem to have updated anything locally :)

bvaughn:
I think we need to update the types in: `node_modules/@recordreplay/protocol/js/client/build/client.d.ts`

I don't know how that gets generated and published.

bhackett1024:
> I think we need to update the types in: `node_modules/@recordreplay/protocol/js/client/build/client.d.ts`
> 
> I don't know how that gets generated and published.
> 
> Maybe here? https://github.com/RecordReplay/protocol

The protocol is part of the backend repository now and published from there, but the latest version should have the loadPoint change.  Is the devtools set up to load the latest version automatically, or does that need to be bumped manually?

bvaughn:
I don't know how these projects fit together, but i see a protocol that's published to NPM and seems to live here:
https://github.com/RecordReplay/protocol/blob/970510204bec19eed0243171f287aad815b07609/json/protocol.json#L237-L246

It looks like you've updated it before? e.g. https://github.com/RecordReplay/protocol/pull/21/

Seems like someone needs to update that JSON, then run `npm run build` and _publish_ the updated protocol to NPM

Then bump the client dependency:
https://github.com/RecordReplay/devtools/blob/ff7cf09a02def1c126d9a0af1697332e7049064b/package.json#L38

But maybe I'm wrong here? cc @jaril @jasonLaster

bhackett1024:
I just bumped the protocol version manually, since the new 0.23.0 release has already gone out to NPM and has the loadPoint change.  I accidentally pushed that directly to the main branch (I didn't know we could do that still?  cc @gideonred)

https://github.com/RecordReplay/devtools/commit/985eaeb1c0759ed2f1ef6fba871bb06463c49d62

bvaughn:
Oh! That's interesting :)

bhackett1024:
BTW https://github.com/recordreplay/protocol is now archived / read only, here is the new location for where the package is published from: https://github.com/RecordReplay/backend/tree/master/packages/protocol

bvaughn:
@bhackett1024 https://github.com/RecordReplay/devtools/pull/6448

PR #6421, opened 2022-04-17
Remove console ComponentMap

jasonLaster:
This does not remove all the grossness, but it is a start


markerikson:
Fwiw I usually lean towards a plain object lookup table, like :

```js
const Component To Render = components[someTypeEnum] ?? SomeDefault 
```

jasonLaster:
closing as a dupe #6422

jasonLaster:
@markerikson same - in this case I held off because of the branching, but in general, that's my preference too.

PR #6422, opened 2022-04-17
Remove console ComponentMap

jasonLaster:
This does not remove all the grossness, but it is a start

### DIFF
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -4,10 +4,8 @@
 "use strict";
 
 import React from "react";
-import ReactDOM from "react-dom";
 import { Dispatch } from "@reduxjs/toolkit";
 import { connect, ConnectedProps } from "react-redux";
-import PropTypes from "prop-types";
 
 import { actions } from "ui/actions";
 import { isVisible } from "ui/utils/dom";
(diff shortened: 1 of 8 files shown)


markerikson (approved):

markerikson:
Heh, shoulda looked at the check results first, but I _am_ at 30K feet right now :) so it's a miracle I can see any of this at all

src/devtools/client/debugger/src/components/Editor/ToggleWidgetButton.tsx
```diff
@@ -182,7 +177,7 @@ function ToggleWidgetButton({ editor, cx, breakpoints }: ToggleWidgetButtonProps
       editor.codeMirror.off("lineMouseEnter", onLineEnter);
       editor.codeMirror.off("lineMouseLeave", onLineLeave);
     };
-  }, []);
+  }, [editor]);
```
bvaughn, line 180: I don't think this is the right change here. We're referencing functions `onLineEnter` and `onLineLeave` that are declared in the render body, but we're only re-running the effect when `editor` changes. This likely means the functions are going to be stale.

It shouldn't matter in this specific case since they close over state update functions (which are themselves stable) but it seems like a future foot-gun.

bvaughn:
The test that's failing on CI passes for me locally:
```sh
node ./test/run.js --pattern inspector-05
```

bvaughn (approved):

jasonLaster:
oops, didnt mean to push this editor stuff.

PR #6434, opened 2022-04-18
Add a unicorn to onboarding 

jonbell-lot23:
Addresses RecordReplay/design#22 (but is not working yet)

Copied from that thread:

Hey @jaril and @jasonLaster, I came up with a quick hack for our onboarding problems. I added a quick unicorn emoji for our first print statements, so at least it will show up in the console. And that means we're not waiting on our Rainbow Console designs to land.

I have it half working, but need help with the other half:
https://www.loom.com/share/656f9a0a7c144039afc2716b1635e14f

(In the interest of time, I'd rather just have someone fix this up and send it out. Learning opportunities are great, but it takes a lot more time to set up a call, do the screen share, talk about it, etc than just shipping it async)

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/SummaryExpression.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/SummaryExpression.tsx
@@ -6,6 +6,8 @@ import Popup from "./Popup";
 import hooks from "ui/hooks";
 const { getCodeMirror } = require("devtools/client/debugger/src/utils/editor/create-editor");
 const { prefs } = require("ui/utils/prefs");
+import { Nag } from "ui/hooks/users";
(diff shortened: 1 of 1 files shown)


jaril:
Gonna be gone next week for PTO so let's check in later to see how we want to proceed here.

Last time we talked about how to insert that unicorn emoji into the breakpoints specifically for onboarding. The summary was:
- It's possible to do it without Rainbow Console changes landing.
- It's pretty hacky.
- Even if we were to commit to it, Jon would need help writing it.

_Implementation details: Have a `useEffect` that runs whenever a breakpoint panel is mounted that checks if the user is should get unicorns (define what the condition for that is). If they should, then quickly+stealthily edit the breakpoint so that we sneak in a `"🦄",` to the beginning of the current print statement's default string._

jonbell-lot23:
Thanks, @jaril ! We're going to close this out for now and address in other ways. (They shipped earlier this week, in fact!)

PR #6441, opened 2022-04-18
Update heatmap hits count when the backend unloads a region forcibly

jaril:
[Replay with the changes](https://app.replay.io/recording/focusing-works--8e41c9f6-b83f-4172-82cf-2be67834bf99)

image

This:
- fixes an error where we had a `null` value is possibly being iterated on.
- refetches the heatmap hit count whenever the focus mode regions are altered.


jcmorrow (approved):
💯

PR #6450, opened 2022-04-19
Add ESLint rules for sorting keys, vars, and imports

bvaughn:
There are too many violations of these rules in place for us to fix them. (Even `lint --fix` fails on hundreds of files.) This PR then makes use of [Trunk's hold-the-line feature](https://docs.trunk.io/check/overview#hold-the-line) to incrementally improve our code base as we edit old files and add new ones.

This commit also updates Trunk and ESLint versions, and changes our NPM lint script (`npm run lint`) to use Trunk so we are using the exact same linting setup locally as on CI.

---

PR #6451 tests this new setup by modifying some existing files and verifying that [CI fails](https://github.com/RecordReplay/devtools/runs/6077213467?check_suite_focus=true):
Screen Shot of CI failing due to lint rule violation


jasonLaster:
Works well!

Was surprised that I actually had to change an import to get an error. I assumed it would be file based, but i guess it's line based.

```
diff --git a/src/devtools/client/debugger/src/components/Editor/HighlightLines.js b/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
index 492165a29..77a125176 100644
--- a/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
+++ b/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
@@ -2,10 +2,11 @@

-import { Component } from "react";

 import range from "lodash/range";
 import isEmpty from "lodash/isEmpty";
+import { Component } from "react";
+
 import { connect } from "../../utils/connect";
 import { getHighlightedLineRange } from "../../selectors";
```

bvaughn:
I'm going to leave this in Draft mode for now because it seems like no matter what combo of ESLint plug-ins I try, "autofix" doesn't work for a lot of cases in our code base 😦 and I'm not sure everyone is willing to sign up to manually fixing ordering for a while until things settle.

package.json
```diff
@@ -10,8 +10,8 @@
     "test:watch": "jest --watch",
     "start": "next start -p 8080",
     "build": "next build",
-    "lint": "next lint",
-    "lint:fix": "npm run lint -- --fix",
+    "lint": "trunk check",
+    "lint:fix": "trunk fmt",
```
dapirian, line 14:
```suggestion
    "lint:fix": "trunk check --fix",
    "fmt": "trunk fmt",
```
- `trunk fmt` only autoformats, which does not include eslint autofixes
- `trunk check --fix` (of `-y` for shorthand) _also_ autofixes lint errors

dapirian:
> Works well!
> 
> Was surprised that I actually had to change an import to get an error. I assumed it would be file based, but i guess it's line based.
> 
> ```diff
> diff --git a/src/devtools/client/debugger/src/components/Editor/HighlightLines.js b/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
> index 492165a29..77a125176 100644
> --- a/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
> +++ b/src/devtools/client/debugger/src/components/Editor/HighlightLines.js
> @@ -2,10 +2,11 @@
>   * License, v. 2.0. If a copy of the MPL was not distributed with this
>   * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
> 
> -import { Component } from "react";
> 
>  import range from "lodash/range";
>  import isEmpty from "lodash/isEmpty";
> +import { Component } from "react";
> +
>  import { connect } from "../../utils/connect";
>  import { getHighlightedLineRange } from "../../selectors";
> ```

Actually, it's better than line based 😉 - it's relatively easy to change a line and cause new lint issues downstream of that line, and `trunk` also detects those as new issues.

jcmorrow (approved):
As discussed in our call, I think it would still be net-positive to include the autofixes that *do* work, and depend on hold-the-line for the remaining violations.

bvaughn:
> As discussed in our call, I think it would still be net-positive to include the autofixes that do work, and depend on hold-the-line for the remaining violations.

Okay, I can do that. Will post a (largeish) PR here in a bit that we'll hopefully just stamp and land quickly.

bvaughn:
Going to open a new PR for this.

Issue #6237, opened 2022-04-07
Audit linting story


jasonLaster:
Changes:
- [x] Update our ESLint config
  - [x] #6453
  - [x] #6456
  - [x] #6454
  - [x] #6463
  - [x] #6457
- [x] Update local NPM scripts
  - [x] #6452
  - [x] #6462
- [x] Fix pre-existing lint issues
  - [x] #6458
  - [x] ~~Fix pre-existing lint errors~~
- [ ] Setup Trunk/CI to hold the line going forward
  - [x] #6461

---

We currently have three lint commands

* `npm run lint`
* `next lint`
* `trunk check` 

we run `trunk check` in ci and `next lint` locally. We also effectively disabled `next lint` a month ago when the accordion package was added.

We should
a) standardize on one tool
b) audit the react-hooks warnings

**What is trunk**
Trunk is a new service (https://docs.trunk.io/) that runs all the linters in a performant way. It can also work across languages, so C++, markdown, and perhaps some other compliance checks. The devtools project is in the next ecosystem, so `next lint` would probably be sufficient too

```
info  - Loaded env from /Users/jasonlaster/src/replay/devtools/.env
warn  - You have enabled experimental feature(s).
warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use them at your own risk.

warn  - The Next.js plugin was not detected in your ESLint configuration. See https://nextjs.org/docs/basic-features/eslint#migrating-existing-config

./src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/FirstEditNag.tsx
30:6  Warning: React Hook useEffect has a missing dependency: 'step'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps

./src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
61:6  Warning: React Hook useEffect has missing dependencies: 'dismissNag', 'editor.editor', and 'updateWidth'. Either include them or remove the dependency array.  react-hooks/exhaustive-deps

./src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Widget.tsx
34:6  Warning: React Hook useEffect has missing dependencies: 'editor.codeMirror', 'insertAt', 'location.line', and 'node'. Either include them or remove the dependency array.  react-hooks/exhaustive-deps

./src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
[...]

PR #6454, opened 2022-04-19
Elevate ESLint react-hooks rules to error level

bvaughn:
Relates to #6237

We have a lot of lint violations for the react-hooks rules. Particularly "react-hooks/exhaustive-deps". In many cases, it's probably not going to cause a bug– but there is potential for subtle bugs (like stale closures) and so we should be treating these violations more seriously.

Now that #6452 has landed, `npm run lint` will run with [Trunk's hold-the-line](https://docs.trunk.io/check/overview#hold-the-line approach)– meaning that it won't complain about pre-existing lint errors.

Given that, this PR raises the ESLint level for the react-hooks from "warn" to "error" to prevent new regressions from being added to the repo.

---

Looks like there's an interesting challenge here. Next runs its own linter, which balks at these rule changes. I think this approach only works if we rely on Trunk for linting. Because of this, I've _disabled_ Next's linting during build. I'm not sure how controversial that may be.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,6 +29,8 @@
     "react/button-has-type": 0,
     "no-nested-ternary": 0,
     "no-restricted-syntax": 0,
+    "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
+    "react-hooks/exhaustive-deps": "error", // Checks effect dependencies
     "react/no-string-refs": 0,
     "react/no-find-dom-node": 0,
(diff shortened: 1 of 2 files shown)


jasonLaster:
re disabling next linting: I support it. It is better for the preview branch to build and get e2e test runs going. I believe we can block deploys on the trunk check which is similar

jcmorrow (approved):
🏅

PR #6457, opened 2022-04-19
Enable stylelint

dapirian:
This uses `stylelint-config-standard` and `stylelint-config-prettier`, however there are a _lot_ of existing issues in the codebase, and someone from the replay team should checkout this branch and do a quick audit to make sure stylelint is tuned how you want it. Thanks!


bvaughn:
Thanks for taking the time to share this approach!

I think we're going to pass on it for now because our styling story is kind of still up in the air, and that seems like it should maybe be tackled first. I'll revisit once we have it a bit more settled.

PR #6458, opened 2022-04-19
Address pre-existing 'react-hooks/exhaustive-deps' rule violations.

bvaughn:
In most cases, I tried to _fix_ the rule violation.

When the fix seemed unclear (e.g. I wasn't sure if the change would "break" the desired behavior) then I left a `TODO [authorname]` comment for the original author. (We'll rely on Trunk hold-the-line to prevent new ones from being added.)

Things to keep in mind about the following changes:
* `useEditor` from `@tiptap/react` returns a _stable_ "editor" provided the dependencies array values don't change.
* `useDispatch` from `react-redux` returns a _stable_ `dispatch` function so long as you don't change the Redux `store` (which we don't).
* There were a few places we could have been memoizing that we weren't, so I added them (e.g. `useGetRecording()` -> `convertRecording(recording.data)`).

### DIFF
--- a/packages/accordion/index.tsx
+++ b/packages/accordion/index.tsx
@@ -102,6 +102,9 @@ export function AccordionPane({
   // do this by intercepting the expanded state in the Accordion so we only have one
   // expanded prop (the one from the Accordion) being considered by the AccordionPane.
   // https://gist.github.com/jaril/dfad5343f141c175d767d21cd6fdaab2
+  //
+  // TODO [jcmorrow] Fix react-hooks/exhaustive-deps
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   useEffect(() => dispatch(expanded ? expandSection(index!) : collapseSection(index!)), [expanded]);
 
   return (
--- a/src/ui/components/shared/CommentTool.tsx
+++ b/src/ui/components/shared/CommentTool.tsx
@@ -83,81 +83,87 @@ function CommentTool({
   const { userId } = useGetUserId();
   const captionNode = useRef<HTMLDivElement | null>(null);
 
-  const addListeners = () => {
+  // Re-register the listener on every update to prevent the props used by the handler functions from being stale.
+  useEffect(() => {
     const videoNode = document.getElementById("graphics");
-
     if (videoNode) {
+      const onMouseDown = (evt: MouseEvent) => {
+        if (!pendingComment || !document.activeElement) {
+          return;
+        }
+
+        const pendingCommentEditorId = getCommentEditorDOMId(pendingComment.comment);
+        // this uses `[id="..."]` because comment ids can have "="s in them!
+        const isEditorFocused = !!document.activeElement.closest(
+          `[id="${pendingCommentEditorId}"]`
+        );
+
+        // If the pending comment's editor is focused, comment tool clicks should not take focus from it.
+        if (isEditorFocused) {
+          evt.preventDefault();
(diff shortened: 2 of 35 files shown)


src/ui/components/shared/PortalDropdown.tsx
```diff
@@ -66,7 +58,7 @@ export default function PortalDropdown(props: PortalDropdownProps) {
               <div
                 className={`content ${position}`}
                 style={{ ...style, ...contentPosition }}
-                ref={dropdownRef}
+                ref={setDropdownNode}
```
bvaughn, line 61: The ref-setter was just passing through to the state-updater, so I just used it directly.

src/ui/components/shared/CommentTool.tsx
```diff
@@ -83,81 +83,87 @@ function CommentTool({
   const { userId } = useGetUserId();
   const captionNode = useRef<HTMLDivElement | null>(null);
 
-  const addListeners = () => {
+  // Re-register the listener on every update to prevent the props used by the handler functions from being stale.
+  useEffect(() => {
     const videoNode = document.getElementById("graphics");
-
     if (videoNode) {
+      const onMouseDown = (evt: MouseEvent) => {
```
bvaughn, line 90: This looks like a lot of changes but the bulk of it is moving event handlers inside of the effect body. We should be doing this as a general rule, to avoid potential stale closure issues.

src/ui/hooks/recordings.ts
```diff
@@ -170,7 +171,8 @@ export function useGetRecording(recordingId: RecordingId | null | undefined): {
     console.error("Apollo error while getting the recording", error);
   }
 
-  const recording = convertRecording(data?.recording);
+  const recording = useMemo(() => convertRecording(data?.recording), [data]);
```
bvaughn, line 174: Custom hooks should consider memoizing like this to make them easier to use.

src/ui/utils/useAuth0.ts
```diff
@@ -23,9 +23,8 @@ const TEST_AUTH = {
 
 export type AuthContext = Auth0ContextInterface | typeof TEST_AUTH;
 
-/**
- * A wrapper around useAuth0() that returns dummy data in tests
- */
+// TODO [hbenl, ryanjduffy] This function should `useMemo` to memoize the "user" object it returns.
+// As it is, this hooks prevents components like CommentTool from limiting how often their effects run.
```
bvaughn, line 27: This seems like a good thing to do but I didn't want to mix it into this PR.

markerikson (approved):
Looks as good as I can tell given that I'm reviewing from my phone :)

packages/accordion/index.tsx
```diff
@@ -102,6 +102,9 @@ export function AccordionPane({
   // do this by intercepting the expanded state in the Accordion so we only have one
   // expanded prop (the one from the Accordion) being considered by the AccordionPane.
   // https://gist.github.com/jaril/dfad5343f141c175d767d21cd6fdaab2
+  //
+  // TODO [jcmorrow] Fix react-hooks/exhaustive-deps
```
jcmorrow, line 106: oh god the accordion.

bvaughn:
Good old `inspector-05` test is failing but that's nothing new. The test passes for me locally on this branch so I'm going to merge this and keep moving forward. Next step will be to run auto-fix on the repo before holding the line.

PR #6460, opened 2022-04-19
Wait for basic processing before fetching messages

jaril:
Testing this out to see if more time is all the console needs to be able to load its messages.

Ideally we write the tests with some intentional performance targets in mind, and we will. But for now, this just needs to be less flaky so we're not wasting mental energy trying to figure out if a failing test is okay or something more serious.

### DIFF
--- a/src/devtools/client/webconsole/actions/messages.ts
+++ b/src/devtools/client/webconsole/actions/messages.ts
@@ -2,9 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import * as Sentry from "@sentry/browser";
 import type { UIStore, UIThunkAction } from "ui/actions";
-
-import { getAllFilters } from "../selectors/messages";
 
 import {
   prepareMessage,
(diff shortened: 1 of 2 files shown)


bvaughn (approved):

src/devtools/client/webconsole/actions/messages.ts
```diff
@@ -42,7 +41,7 @@ export function setupMessages(store: UIStore) {
   ThreadFront.findConsoleMessages(
     (_, msg) => store.dispatch(onConsoleMessage(msg)),
     () => store.dispatch(onConsoleOverflow())
-  ).then(() => store.dispatch(messagesLoaded()));
+  ).then(() => store.dispatch(messagesLoaded()), Sentry.captureException);
```
jasonLaster, line 44: This si fine, but it would be better if we didn't have to sprinkle it in.

Perhaps we can add an exception higher up in the redux or protocol level

bhackett1024:
I'm having trouble understanding this change, the first comment is talking about flaky tests but adding these waits can have an effect on console loading for users, because the devtools can load before basic processing finishes and it would be good to proactively fetch the console messages like were doing before so they can be displayed as they arrive.  If the tests are being flaky it seems better to address that within the tests themselves, by changing how they wait for things on the page to load.

src/protocol/thread/thread.ts
```diff
@@ -928,6 +928,8 @@ class _ThreadFront {
     onConsoleOverflow: () => void
   ) {
     const sessionId = await this.waitForSession();
+    // Wait for basic processing to start before fetching console messages.
+    await this.ensureProcessed("basic");
```
loganfsmyth, line 932: This needs to be passed the `sessionId` :P

jaril: Does it?? https://github.com/RecordReplay/devtools/blob/9e8035861d1f30e4ef64d33dd8f3feada26b0e68/src/protocol/thread/thread.ts#L274-L276

PR #6461, opened 2022-04-19
Setup Trunk Webapp to sync nightly

bvaughn:
Following the setup instructions here: https://docs.trunk.io/web-app/upload-check

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,19 @@
+name: Nightly
+on:
+  schedule:
+    - cron: 0 8 * * 1-5
+  workflow_dispatch:
+
+jobs:
+  trunk_check:
+    name: Trunk Check Upload
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Trunk Check
+        uses: trunk-io/trunk-action@v1.0.0
+        with:
+          trunk-token: ${{ secrets.TRUNK_TOKEN }}
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -14,3 +14,8 @@ lint:
         - public
         - src/graphql
         - docs/codebase-notes
+repo:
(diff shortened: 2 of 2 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Trunk Check
+        uses: trunk-io/trunk-action@v1.0.0
+        with:
+          trunk-token: ${{ secrets.GITHUB_TOKEN }}
```
dapirian, line 19: This should be a token you get from the trunk web app and store as a GH action secret, not the GitHub token

dapirian (commented):
Nice!

.github/workflows/nightly.yaml
bvaughn: @dapirian Ooo! That wasn't clear to me from the instructions. I don't see any part of the Trunk web app where I can get this from. Maybe related to the GitHub connection issue you're looking into though?

dapirian: oh we'll make that more clear! yes, you can't see it because of that issue - stay tuned :)

jasonLaster (approved):

jaril (approved):

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Trunk Check
+        uses: trunk-io/trunk-action@v1.0.0
+        with:
+          trunk-token: ${{ secrets.TRUNK_TOKEN }}
```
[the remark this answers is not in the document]
bvaughn: `trunk check --fix` does not fix this 🤷🏼 

This is the exact specific type of thing that _needs_ to be automated or it will drive people crazy 😄

PR #6464, opened 2022-04-19
Eslint sort autofix

bvaughn:
Relates to #6237 and builds on #6463.

We might as well try to auto-fix what we can as part of updating our rules:
```sh
trunk check --fix --all
```


jaril (approved):

bvaughn:
Maybe I should close this PR in favor of one fix _per_ lint rule.

Issue #6459, opened 2022-04-19
Progress bar in the console can be scrolled away from view


jaril:
We have a blue progress bar at the top of the console messages to indicate that we haven't finish loading all the messages yet.

It's a pretty important signal. But apparently it's possible to scroll and have it be hidden from view. 

https://user-images.githubusercontent.com/15959269/164058330-e4cdf345-4b47-4527-9b1d-968f4f4424bd.mov

johnathan-co:
@jaril [PR](https://github.com/RecordReplay/devtools/pull/6469) is ready. 

Sorry for the delay. This bar is utilized by a few different components, and I thought I was running into issues with the bar on the Markup (Elements) tab after the change, but all seems well.

jaril:
No worries — appreciate the help!!!!

johnathan-co:
I've gotta run and won't be able to pick [this](https://github.com/RecordReplay/devtools/pull/6469#issuecomment-1103192241) back up until tomorrow, so @ashwinkhode feel free to jump in if you'd like! Otherwise, I'll be back tomorrow.

jaril:
Fixed by https://github.com/RecordReplay/devtools/pull/6469.

PR #6469, opened 2022-04-19
Stick the progress bar to prevent losing when scrolling

johnathan-co:
@jaril 
#6459


jaril:
@johnathan-co looks like there might be some funkiness around the progress bar being used in the editor with the sticky change.

Tried to capture it in this video — not sure if you'll see but for a split second it displays under line 1.

https://user-images.githubusercontent.com/15959269/164104503-84c7668e-4747-4d5f-939b-6cebbc6a553f.mov

jaril:
Maybe we can create a separate style specifically just for the console?

johnathan-co:
> Maybe we can create a separate style specifically just for the console?

We could and that was the route I was going to go when I _thought_ I saw some funkiness with the Elements tab.

jaril (changes requested):
Looks like this sits well for all the other places that this component is being used in (console, network monitor, elements), but it throws things off in the code editor. 

Let's turn this `position: sticky` into a console specific style and i think we're good to go!

Let's make sticky a specifically

PR #6477, opened 2022-04-20
Add a way to salvage your replay session if the backend chooses to unload some regions for the user

### LINKED ISSUE #6289: Unloaded regions not showing on timeline
In the replay below, unloaded regions are not shown correctly on the timeline.

https://app.replay.io/recording/loaded-regions-not-showing--ba3e6c6e-a062-46a4-be11-ea7bb7d30908?point=33749929584389312161735533083492352&time=27080&hasFrames=false#

jaril:
Fix #6289.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -511,6 +511,19 @@ export function setFocusRegion(focusRegion: {
   return { type: "set_trim_region", focusRegion };
 }
 
+export function setFocusAroundTime(time: number, maxTime: number): UIThunkAction {
+  return (dispatch, getState) => {
+    const state = getState();
+    const zoomRegion = getZoomRegion(state);
+
+    const duration = zoomRegion.endTime;
+    const startTime = Math.max(time - maxTime / 2, 0);
+    const endTime = Math.min(time + maxTime / 2, duration);
+
+    dispatch(setFocusRegion({ endTime, startTime }));
+  };
+}
+
 export function updateFocusRegion(operation: FocusOperation): UIThunkAction {
   return (dispatch, getState) => {
     const state = getState();
@@ -568,9 +581,9 @@ export function syncFocusedRegion(): UIThunkAction {
   };
 }
 
-export function enterFocusMode(): UIThunkAction {
+export function enterFocusMode(instructions?: string): UIThunkAction {
   return (dispatch, getState) => {
-    dispatch(setModal("focusing"));
+    dispatch(setModal("focusing", { instructions }));
     const state = getState();
     const focusRegion = getFocusRegion(state);
     const zoomRegion = getZoomRegion(state);
@@ -582,7 +595,7 @@ export function enterFocusMode(): UIThunkAction {
   };
 }
 
-export function toggleFocusMode(): UIThunkAction {
+export function toggleFocusMode(instructions?: string): UIThunkAction {
   return (dispatch, getState) => {
     const state = getState();
     const isFocusing = getIsFocusing(state);
@@ -592,7 +605,7 @@ export function toggleFocusMode(): UIThunkAction {
       dispatch(hideModal());
     } else {
       trackEvent("timeline.start_focus_edit");
-      dispatch(enterFocusMode());
+      dispatch(enterFocusMode(instructions));
     }
   };
 }
--- a/src/ui/components/Timeline/NonLoadingRegions.tsx
+++ b/src/ui/components/Timeline/NonLoadingRegions.tsx
@@ -1,0 +1,31 @@
+import { useSelector } from "react-redux";
+import { getNonLoadingTimeRanges } from "ui/reducers/app";
+import { getZoomRegion } from "ui/reducers/timeline";
+import { TimeRange } from "ui/utils/app";
+
+const NonLoadingRegion = ({ range }: { range: TimeRange }) => {
+  const { endTime } = useSelector(getZoomRegion)!;
+  const { start, end } = range;
+  const style = {
+    left: `${(start / endTime) * 100}%`,
+    width: `${((end - start) / endTime) * 100}%`,
+  };
+
+  return (
+    <div className="unfocused-regions-container start" style={style}>
+      <div className="unfocused-regions" />
+    </div>
+  );
+};
+
+export default function NonLoadingRegions() {
+  const timeRanges = useSelector(getNonLoadingTimeRanges);
+
+  return (
+    <>
+      {timeRanges.map((r, i) => (
+        <NonLoadingRegion key={i} range={r} />
+      ))}
+    </>
+  );
+}
--- a/src/ui/components/Timeline/Tooltip.tsx
+++ b/src/ui/components/Timeline/Tooltip.tsx
@@ -1,37 +1,38 @@
 import React from "react";
-import { connect, ConnectedProps } from "react-redux";
-import { selectors } from "ui/reducers";
-import { UIState } from "ui/state";
+import { useSelector } from "react-redux";
+import { getIsFocusing, getNonLoadingTimeRanges } from "ui/reducers/app";
+import { getHoverTime, getZoomRegion } from "ui/reducers/timeline";
 import { getVisiblePosition } from "ui/utils/timeline";
 
-export function Tooltip({ hoverTime, zoomRegion, timelineWidth, isFocusing }: Props) {
+const getTimestamp = (time?: number) => {
+  const date = new Date(time || 0);
+  const seconds = date.getSeconds().toString().padStart(2, "0");
+  const minutes = date.getMinutes();
+
+  return `${minutes}:${seconds}`;
+};
+
+export default function Tooltip({ timelineWidth }: { timelineWidth: number }) {
+  const hoverTime = useSelector(getHoverTime);
+  const zoomRegion = useSelector(getZoomRegion);
+  const isFocusing = useSelector(getIsFocusing);
+  const nonLoadingRegion = useSelector(getNonLoadingTimeRanges);
   const shouldHideTooltip = !hoverTime || isFocusing;
+
   if (shouldHideTooltip) {
     return null;
   }
 
   let offset = getVisiblePosition({ time: hoverTime, zoom: zoomRegion }) * timelineWidth;
 
-  const time = new Date(hoverTime || 0);
-  const seconds = time.getSeconds().toString().padStart(2, "0");
-  const minutes = time.getMinutes();
+  const isHoveredOnNonLoadingRegion = nonLoadingRegion.some(
+    r => r.start <= hoverTime && r.end >= hoverTime
+  );
+  const msg = isHoveredOnNonLoadingRegion ? "Unloaded" : getTimestamp(hoverTime);
 
   return (
     <div className="timeline-tooltip" style={{ left: offset }}>
-      {`${minutes}:${seconds}`}
+      {msg}
     </div>
   );
 }
-
-const connector = connect((state: UIState) => ({
-  hoverTime: selectors.getHoverTime(state),
-  zoomRegion: selectors.getZoomRegion(state),
-  isFocusing: selectors.getIsFocusing(state),
-}));
-
-type PropsFromRedux = ConnectedProps<typeof connector>;
-type Props = PropsFromRedux & {
-  timelineWidth: number;
-};
-
-export default connector(Tooltip);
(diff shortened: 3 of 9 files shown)


src/ui/reducers/app.ts
```diff
@@ -298,6 +300,17 @@ export const getLoading = (state: UIState) => state.app.loading;
...
 export const getLoadingFinished = (state: UIState) => state.app.loadingFinished;
 export const getLoadedRegions = (state: UIState) => state.app.loadedRegions;
+export const getNonLoadingTimeRanges = (state: UIState) => {
+  const loadingRegions = getLoadedRegions(state)?.loading || [];
+  const endTime = getZoomRegion(state).endTime;
+
+  return [
+    { start: 0, end: 5000 },
+    { start: 10000, end: 24555 },
+  ];
```
jaril, line 310: Hack I added for now just to let me test how the UI feels. Technically I could have unloaded regions from the console but it's late and that takes more thinking time :P

jaril: Will remove this prior to asking for a review

src/ui/components/Timeline/index.tsx
```diff
@@ -150,6 +156,10 @@ class Timeline extends Component<PropsFromRedux, { isDragging: boolean }> {
       return;
     }
 
+    if (!isFocusing && clickedOnNonLoadingRegion) {
+      enterFocusMode("We're having difficulties keeping up with your current replay. Please select a <30s time range to focus on.");
```
jaril, line 160: @jonbell-lot23 👋 placeholder copy gimme your best shot :)

src/ui/utils/app.ts
```diff
@@ -0,0 +1,30 @@
...
+export type TimeRange = {
+  start: number;
+  end: number;
+};
+
+export function getNonLoadingRegionTimeRanges(
+  loadingRegions: TimeStampedPointRange[],
+  endTime: number
+): TimeRange[] {
+  if (!loadingRegions.length) {
```
jaril, line 11: this was making me nervous so even though technically this is reducer-y logic I extracted it to its own util so that we can write unit tests for it

jasonLaster:
@jaril can you share a loom + replay of this in action?

jonbell-lot23:
Here's his loom https://www.loom.com/share/7ce8bf01bee7420183896894a58d6170

jonbell-lot23:
@jaril see comments on discord - I see three things I think are important even for a v1

jaril:
Latest loom with the following changes:
- `Unloaded` tooltip when hovering over an unloaded region.
- Clicking on an unloaded region in the timeline defaults the new focus to a 30s range with the mouse position as the midpoint.
- Changed the copy to @jonbell's "We're having difficulties with your replay. Please select a shorter range. 30 seconds or less works best!"

src/ui/components/Timeline/NonLoadingRegions.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { getZoomRegion } from "ui/reducers/timeline";
+import { TimeRange } from "ui/utils/app";
+
+const NonLoadingRegion = ({ range }: { range: TimeRange }) => {
+  const { endTime } = useSelector(getZoomRegion)!;
+  const { start, end } = range;
+  const style = {
+    left: `${(start / endTime) * 100}%`,
+    width: `${((end - start) / endTime) * 100}%`,
+  };
```
bvaughn, line 12: Some small bit of confusion about our domain model here, but it relates to the duration question above.

Looks like this `left` positioning logic kind of assumes either that the zoom region's `startTime` is 0 or that ranges are always relative to a specific range. I guess maybe that's true since our `TimelineState` only has a single `zoomRegion` it's just confusing to me initially because I'm used to seeing view logic related to start/time times needing to explicitly account for both.

src/ui/components/Timeline/Tooltip.tsx
```diff
@@ -1,37 +1,38 @@
...
+  const minutes = date.getMinutes();
+
+  return `${minutes}:${seconds}`;
+};
+
+export default function Tooltip({ timelineWidth }: { timelineWidth: number }) {
+  const hoverTime = useSelector(getHoverTime);
+  const zoomRegion = useSelector(getZoomRegion);
+  const isFocusing = useSelector(getIsFocusing);
+  const nonLoadingRegion = useSelector(getNonLoadingTimeRanges);
```
bvaughn, line 19: nit
```suggestion
  const nonLoadingRegions = useSelector(getNonLoadingTimeRanges);
```

src/ui/components/Timeline/Tooltip.tsx
```diff
@@ -1,37 +1,38 @@
...
   }
 
   let offset = getVisiblePosition({ time: hoverTime, zoom: zoomRegion }) * timelineWidth;
 
-  const time = new Date(hoverTime || 0);
-  const seconds = time.getSeconds().toString().padStart(2, "0");
-  const minutes = time.getMinutes();
+  const isHoveredOnNonLoadingRegion = nonLoadingRegion.some(
+    r => r.start <= hoverTime && r.end >= hoverTime
+  );
```
bvaughn, line 30: nit: We have a new helper for this if you wanted to use it:
https://github.com/RecordReplay/devtools/blob/d7683b635fd8ea15717fa0c1e130e91c0b9e2d4c/src/ui/utils/timeline.ts#L149-L151

src/ui/components/Timeline/Tooltip.tsx
```diff
@@ -1,37 +1,38 @@
...
+  const msg = isHoveredOnNonLoadingRegion ? "Unloaded" : getTimestamp(hoverTime);
```
bvaughn, line 31: I'm curious why we just show "unloaded" here vs showing the time _and_ "unloaded" message?

src/ui/components/Timeline/index.tsx
```diff
@@ -138,6 +143,9 @@ class Timeline extends Component<PropsFromRedux, { isDragging: boolean }> {
       focusRegion &&
       (mouseTime < focusRegion.startTime || mouseTime > focusRegion.endTime) &&
       !isFocusing;
+    const clickedOnNonLoadingRegion = nonLoadingTimeRanges.some(
+      r => r.start <= mouseTime && r.end >= mouseTime
```
bvaughn, line 147: nit https://github.com/RecordReplay/devtools/blob/d7683b635fd8ea15717fa0c1e130e91c0b9e2d4c/src/ui/utils/timeline.ts#L149-L151

src/ui/state/app.ts
```diff
@@ -21,6 +21,7 @@ export type ModalOptionsType = {
   view?: string;
   loom?: string;
   comment?: Reply;
+  instructions?: string;
```
bvaughn, line 24: Mildly surprised ESLint didn't flag this as not being sorted but maybe that check doesn't run on types :)

src/ui/utils/app.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        },
+      ],
+      300
+    );
+
+    expect(timeRanges.length).toEqual(1);
+    expect(timeRanges[0].end).toEqual(200);
+    expect(timeRanges[0].start).toEqual(100);
+  });
+});
```
bvaughn, line 90: Nice 👍🏼

src/ui/utils/app.ts
bvaughn: Oooh, do we not write tests for our Redux reducers/state? 😮

src/ui/utils/app.ts
```diff
@@ -0,0 +1,30 @@
...
+    return [{ end: endTime, start: 0 }];
```
bvaughn, line 12: Why do we use `startTime`/`endTime` for some things and `start`/`end` for other things? At least we have TypeScript to help but it's a little confusing :)

It'a also interesting that sometimes `end` refers to the number but other times its an object with a `.time` property 😁

bvaughn (approved):
This looks good 👍🏼  I left a few suggestions, mostly nits.

src/ui/utils/app.ts
jaril: Nope 😬

src/ui/actions/timeline.ts
```diff
@@ -511,6 +511,19 @@ export function setFocusRegion(focusRegion: {
   return { type: "set_trim_region", focusRegion };
 }
 
+export function setFocusAroundTime(time: number, maxTime: number): UIThunkAction {
+  return (dispatch, getState) => {
+    const state = getState();
+    const zoomRegion = getZoomRegion(state);
+
+    const duration = zoomRegion.endTime;
```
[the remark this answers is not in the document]
jaril: Technically, yes. `zoomRegion`'s an outdated concept that really should just be refactored out at some point. It used to be that scrolling vertically while on the timeline would zoom you in and out. That's since been taken out. So `zoomRegion.endTime` is always the end of the recording, and `zoomRegion.startTime` is always 0.

src/ui/actions/timeline.ts
```diff
@@ -511,6 +511,19 @@ export function setFocusRegion(focusRegion: {
...
+    const startTime = Math.max(time - maxTime / 2, 0);
+    const endTime = Math.min(time + maxTime / 2, duration);
```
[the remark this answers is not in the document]
jaril: Yeah, that was the intended behavior so that we try to keep wherever the user clicked as a midpoint, and try to capture up to focus up to 15s on either end. But you're right, say you clicked at 0s, then you'd default the focus to 0s-15s. 

Also in general I have more confidence in the backend being able to focus properly with a smaller duration so I preferred it default to a smaller rather than a larger range. The user can easily adjust it as a follow-up if needed.

src/ui/components/Timeline/NonLoadingRegions.tsx
jaril: Yeah `zoomRegion.startTime` is always 0 and `zoomRegion.endTime` is the duration of the recording. Definitely overdue for a refactor.

src/ui/components/Timeline/Tooltip.tsx
jaril: Ah good catch — fixed!

src/ui/components/Timeline/Tooltip.tsx
jaril: I figured @jonbell-lot23 would have input here that we can address as a follow-up. 

For now, I tried to err on making the signal noisier than quieter, and a big `Unloaded` as well as the timestamp missing felt like a nice unmistakable signal to the user that something's up.

On that note, I just realized that if while hovering on the timeline and you saw unloaded, it might scare off the user from clicking it in the first place. Maybe we need to be more explicit with an even louder message, e.g. "Click here to fix it"?

src/ui/components/Timeline/index.tsx
jaril: Nice catch. Using it!

jaril: Ah I would have, but you'll notice that the `nonLoadingRegions` is a different type from `TimeStampedPointRange[]`, since it doesn't have a point+time for the start and end. It only has the time. 

The reason is there's no easy way to get the last point in a recording. So it's hard to recreate the inverse of `loadedRegions.loading` with the points intact. But we know what the endTime is (thanks to zoomRegion), so that's what I end up using here.

src/ui/utils/app.ts
jaril: Same comment here: https://github.com/RecordReplay/devtools/pull/6477#discussion_r854737015

There's a nice chunk of work that can simplify this once we get around to cleaning up this function, especially [these calls](https://github.com/RecordReplay/devtools/blob/8a1cbf7fa800224d922bfea230a957a0f1669a97/src/ui/actions/timeline.ts#L96-L99). But since this is affecting users I stopped short of doing that bit for now.

src/ui/actions/timeline.ts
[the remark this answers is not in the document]
bvaughn: My gut tells me that we should maybe still avoid hard-coding that dependency into places like this. So long as there's an "end time" attribute (and not a "duration" attribute) then it feels like it should be fair game to change the region's "start time" without expecting it to break other places like this.

src/ui/actions/timeline.ts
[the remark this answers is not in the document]
bvaughn: That's fair. I don't feel strongly about this. Just seems like we're being unnecessarily harsh/strict in this case :D

Issue #6479, opened 2022-04-20
ObjectInspector doesn't ensure all objects are loaded


hbenl:
Sometimes the ObjectInspector tries to show a preview for an object without loading it. For example, a function value may be shown as `Function { ... }` instead of `function <name>()` because the latter preview can only be shown when that value (containing the function's name) has been loaded.
The object-preview-01 e2e test is sporadically failing because of this, e.g. https://app.replay.io/recording/1a12125d-eaf6-4d68-a956-9a90c38df053

PR #6480, opened 2022-04-20
ObjectInspector: ensure all ValueFronts are loaded (#6479)

hbenl:
This fixes several issues that led to the ObjectInspector showing previews for objects without loading their `ValueItem`s first (i.e. their `ValueFronts` contained only an `objectId` and a `className` but not a preview).
- when the ObjectInspector is rendered with a root `ValueItem` that hasn't been loaded yet, it will load it now and then rerender (ideally this should never happen: the ObjectInspector should only be rendered with a root `ValueItem` that has already been loaded, but that isn't always the case right now and I decided to add this logic to make the ObjectInspector more resilient)
- I added a `loaded` property to `ValueItem` to ensure that it is rerendered when `loaded` changes from `false` to `true`
- the root `ValueItem` created in connected-object-inspector is now recreated on every update, ensuring that its `loaded` property will change after it has been loaded
- the children of `KeyValueItem` and `ContainerItem` are now loaded when the items get expanded
- the `ValueItem` children of a `ContainerItem` are now recreated on every update ensuring that their `loaded` property is up to date

### DIFF
--- a/src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
+++ b/src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
@@ -72,6 +72,24 @@ class OI extends PureComponent<ObjectInspectorProps> {
     this.activeItem = this.props.activeItem;
   }
 
+  componentDidMount() {
+    this.ensureRootsLoaded();
+  }
+
+  // TODO [hbenl] this should not be necessary: the ObjectInspector should only
+  // be rendered with a root ValueItem that has already been loaded.
+  // However currently (at least) the console doesn't ensure that this is the case,
+  // so we need this workaround for now.
+  async ensureRootsLoaded() {
+    const rootsToLoad = this.getRoots().filter(
+      (root): root is ValueItem => root.type === "value" && !root.loaded
+    );
+    if (rootsToLoad.length > 0) {
+      await Promise.all(rootsToLoad.map(root => root.contents.load()));
+      this.forceUpdate();
+    }
+  }
+
   getRoots = (): Item[] =>
     this.props.roots instanceof Function ? this.props.roots() : this.props.roots;
 
@@ -98,23 +116,24 @@ class OI extends PureComponent<ObjectInspectorProps> {
     } else {
       this.expandedPaths.delete(item.path);
     }
-    this.forceUpdate();
 
     if (!expand) {
+      this.forceUpdate();
       return;
     }
 
     if (item.type === "getter" && item.valueItem) {
       item = item.valueItem;
     }
-    if (item.type === "value" && item.needsToLoadChildren()) {
+    if ("childrenLoaded" in item && !item.childrenLoaded) {
       try {
         await item.loadChildren();
       } catch {
         this.expandedPaths.delete(item.path);
       }
-      this.forceUpdate();
     }
+
+    this.forceUpdate();
   };
 
   expand = (item: Item): Promise<void> => this.setExpanded(item, true);
--- a/src/devtools/packages/devtools-reps/object-inspector/utils/container.tsx
+++ b/src/devtools/packages/devtools-reps/object-inspector/utils/container.tsx
@@ -3,14 +3,16 @@ import { IItem, Item, LabelAndValue } from ".";
 
 export class ContainerItem implements IItem {
   readonly type = "container";
-  name: string;
-  path: string;
-  contents: Item[];
+  readonly name: string;
+  readonly path: string;
+  readonly contents: Item[];
+  readonly childrenLoaded: boolean;
 
   constructor(opts: { name: string; contents: Item[] } & ({ parent: Item } | { path: string })) {
     this.name = opts.name;
     this.path = "parent" in opts ? `${opts.parent.path}/${opts.name}` : opts.path;
     this.contents = opts.contents;
+    this.childrenLoaded = this.contents.every(item => item.type !== "value" || item.loaded);
   }
 
   isPrimitive(): boolean {
@@ -22,7 +24,13 @@ export class ContainerItem implements IItem {
   }
 
   getChildren(): Item[] {
-    return this.contents;
+    return this.contents.map(item => (item.type === "value" ? item.recreate() : item));
+  }
+
+  async loadChildren() {
+    await Promise.all(
+      this.contents.map(item => item.type === "value" && item.contents.loadIfNecessary())
+    );
   }
 
   shouldUpdate(prevItem: Item) {
--- a/src/devtools/packages/devtools-reps/object-inspector/utils/keyValue.tsx
+++ b/src/devtools/packages/devtools-reps/object-inspector/utils/keyValue.tsx
@@ -2,15 +2,15 @@ import { ValueFront } from "protocol/thread";
 import { assert } from "protocol/utils";
 import { IItem, isValueLoaded, Item, LabelAndValue, ValueItem } from ".";
 import { ObjectInspectorItemProps } from "../components/ObjectInspectorItem";
-const PropRep = require("../../reps/prop-rep");
+import PropRep from "../../reps/prop-rep";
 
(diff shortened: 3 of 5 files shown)


src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
```diff
@@ -72,6 +72,20 @@ class OI extends PureComponent<ObjectInspectorProps> {
...
+
+  async ensureRootsLoaded() {
+    const rootsToLoad = this.getRoots().filter(
+      (root): root is ValueItem => root.type === "value" && !root.loaded
+    );
+    if (rootsToLoad.length > 0) {
+      await Promise.all(rootsToLoad.map(root => root.contents.load()));
+      this.forceUpdate();
+    }
+  }
```
bvaughn, line 87: > - when the ObjectInspector is rendered with a root `ValueItem` that hasn't been loaded yet, it will load it now and then rerender (ideally this should never happen: the ObjectInspector should only be rendered with a root `ValueItem` that has already been loaded, but that isn't always the case right now and I decided to add this logic to make the ObjectInspector more resilient)

I guess this relates to what you were saying in the PR description :grin:

Should we add some kind of warning here? I think if this happens, maybe it suggests that the higher level code that's orchestrating the loading of value items isn't working correct? and this "fixes" things but also in a way that is less efficient because of the loading waterfall (load a little data and render, then load more data and re-render).

src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
```diff
@@ -98,23 +112,24 @@ class OI extends PureComponent<ObjectInspectorProps> {
...
     }
-    if (item.type === "value" && item.needsToLoadChildren()) {
+    if ("childrenLoaded" in item && !item.childrenLoaded) {
       try {
         await item.loadChildren();
       } catch {
         this.expandedPaths.delete(item.path);
       }
-      this.forceUpdate();
     }
```
bvaughn, line 130: This comment higher level than the PR, but as I'm looking at this code now I'm noticing that the way it uses instance properties may cause problems when we eventually migrate to `createRoot` and start using React's concurrent features. Maybe we could chat some time about why we're putting things on the instance rather than in component state. 😁

src/devtools/packages/devtools-reps/object-inspector/utils/container.tsx
```diff
@@ -22,7 +24,13 @@ export class ContainerItem implements IItem {
   }
 
   getChildren(): Item[] {
-    return this.contents;
+    return this.contents.map(item => (item.type === "value" ? item.recreate() : item));
```
bvaughn, line 27: This looks kind of heavy handed, but I don't have a lot of context.

src/devtools/packages/devtools-reps/object-inspector/utils/keyValue.tsx
```diff
@@ -2,15 +2,15 @@ import { ValueFront } from "protocol/thread";
...
-  name: string;
-  path: string;
-  key: ValueFront;
-  value: ValueFront;
-  childrenLoaded: boolean;
+  readonly name: string;
+  readonly path: string;
+  readonly key: ValueFront;
+  readonly value: ValueFront;
+  readonly childrenLoaded: boolean;
```
bvaughn, line 13: Ah, so we _were_ mutating before and that was causing React not to re-render?

src/devtools/packages/devtools-reps/object-inspector/utils/value.tsx
```diff
@@ -216,10 +210,19 @@ export class ValueItem implements IItem {
...
   }
 
+  recreate() {
+    return new ValueItem({
+      contents: this.contents,
+      isInCurrentPause: this.isInCurrentPause,
+      name: this.name,
+      path: this.path,
+    });
+  }
```
bvaughn, line 220: This throws me a little. If these properties are now explicitly marked as read-only, and the new `ValueItem` literally has the same values as the exciting one– why do we need to recreate it? Is this a bandaid over something deeper and more serious?

bvaughn (commented):
> * the root `ValueItem` created in connected-object-inspector is now recreated on every update...
> * the `ValueItem` children of a `ContainerItem` are now recreated on every update...

This seems pretty heavy-handed. Is the concern that we aren't subscribing to changes correctly? Or are we mutating the underlying data such that re-renders aren't otherwise triggered?

Typically React components shouldn't need to be "reset" in this way except for uncommon cases like derived state.

I don't want to block this PR but I'm also not sure I understand why we need to force recreate so much. Maybe we can chat about this?

hbenl:
Some more context concerning the update strategy of the tree component used by the ObjectInspector:
Whenever the ObjectInspector calls `forceUpdate()`, a new item tree is created (using the `getRoots()` and `getItemChildren()` methods), the items in the new tree are associated with items in the old tree whose `path` property has the same value and `shouldItemUpdate(prevItem, nextItem)` is called to determine whether the component for that item should be rerendered.
So it's important that all items that _may_ need to be rerendered are recreated. But recreating an item does not imply it will be rerendered.

src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
hbenl: I don't know the reason, I kept it from legacy code. Given how hard it can be in my experience to reason about ObjectInspector's update strategy, I would very much welcome a thorough refactoring of this code.

src/devtools/packages/devtools-reps/object-inspector/utils/container.tsx
hbenl: The comment I just added on ObjectInspector's update strategy should clarify this.

src/devtools/packages/devtools-reps/object-inspector/utils/keyValue.tsx
hbenl: No, I just added this to clarify that these objects should never be mutated.

src/devtools/packages/devtools-reps/object-inspector/utils/value.tsx
hbenl: These properties are the same, but the `loaded` and `childrenLoaded` properties (which are computed in the item's constructor) may be different because the state of the underlying `ValueFront`s may have changed.

src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
hbenl: I haven't added a warning so far because I already know that the console code doesn't handle this correctly. I'll take another look if I can fix this (and if so, add a warning here).

bvaughn:
Thanks for the extra info.

This seems like an unusual pattern for React component props and memoization.

The combination of these two things concerns me:
> the `loaded` and `childrenLoaded` properties (which are computed in the item's constructor) may be different because the state of the underlying `ValueFront`s may have changed.

> So it's important that all items that _may_ need to be rerendered are recreated. But recreating an item does not imply it will be rerendered.

But I feel like I don't have a thorough enough understanding of the surrounding code to properly state why.

I trust your intuition about whether this fix is an improvement over what's there. 😄 

> Given how hard it can be in my experience to reason about ObjectInspector's update strategy, I would very much welcome a thorough refactoring of this code.

I agree with this sentiment!

bvaughn (approved):

src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
jcmorrow: Just my two cents: it seems like adding a warning and TODO's at the places we already know that violate this would be enough to get this in without having to fix any additional issues. Even though this change is small it's already pretty dense!

PR #6488, opened 2022-04-21
Don't merge ranges with no overlap

### LINKED ISSUE #6439: Hits stuck on "loading" after using file search
See replay and comment below.  This seems related to using file search to move around the file and then look at hits on lines nearby.

https://app.replay.io/recording/hits-loading-forever--3a8e97f8-0ee2-4b10-82f5-98af3a912fc0?point=93461343463184194617469009251011317&time=38057&hasFrames=false#

jcmorrow:
When getting hit counts for large files we chunk up the lines into 1,000 line blocks. So if you request the hit counts for line 20 we actually grab the hit counts for lines 1 - 1,000 (which, for most files, is all of them). If you then scroll to a line of that file beyond the 1,000th line, we will request the 1,000 lines surrounding *that* chunk. If you do this in sequential order (1-1,000 and then 1,001-2,000, etc.) then all is well. But if you jump around, going from line 20 to line 5,020, we load the regions 1-1,000 and 5,001 - 6,000, but *mark* the lines 1 - 6,000 as loaded. So if you then go to look at line 4,020, we don't actually request the hit counts and you get a `loading...` tooltip forever. This change keeps track of lines as ranges, and correctly keeps non-overlapping ranges separate.

Fixes #6439

### DIFF
--- a/src/devtools/client/debugger/src/reducers/source-actors.ts
+++ b/src/devtools/client/debugger/src/reducers/source-actors.ts
@@ -3,8 +3,9 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 import type { AnyAction, Action } from "@reduxjs/toolkit";
-
+import sortBy from "lodash/sortBy";
 import type { UIState } from "ui/state";
+
 import { asSettled, asyncActionAsValue, AsyncValue } from "../utils/async-value";
 import {
   createInitial,
@@ -19,7 +20,10 @@ import {
   makeReduceAllQuery,
 } from "../utils/resource";
 import type { ResourceState } from "../utils/resource/core";
+
 import type { HitCount } from "./sources";
+
+export type LineRange = { min: number; max: number };
 
 export interface SourceActor {
   actor: string;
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/reducers/source-actors.ts
```diff
@@ -164,9 +164,25 @@ function updateBreakableLines(
     return state;
   }
 
-  return updateResources(state, [{ id: sourceId, breakableLines: value }]);
+  return updateResources(state, [{ breakableLines: value, id: sourceId }]);
 }
 
+const mergeRanges = (ranges: { min: number; max: number }[]): { min: number; max: number }[] => {
```
markerikson, line 170: :hand: Might be worth pulling out the `{min, max}` type so it doesn't have to get repeated inline in each spot

markerikson (approved):

src/devtools/client/debugger/src/reducers/source-actors.ts
jcmorrow: Yeah I was also thinking about this, let's do it 👍

PR #6489, opened 2022-04-21
Change "internal error" to a nicer message

jonbell-lot23:
Now it says "Whoops! Something went wrong, can you please try again?"

(Note: I think this is the right place because it's nowhere else in the code. Need confirmation)


jcmorrow (approved):
🌈

PR #6493, opened 2022-04-21
Draft for separate login screen when it's the user's first time opening the browser

jaril:
Tried whipping up what the login screen component would look like if we just treated the first browser open login screen as its own thing. This just isolates some of the same JSX into its own if statement and theoretically we could have gone more generic, but didn't want to go too far out of my way to isolate it completely.

@jonbell-lot23 is this easier to hack on? We can ship this and it should be pretty easy for you to tweak that login page without being too tied down with dry components.


jonbell-lot23 (approved):
Seems good to me! Won't be able to speak to it until I get into the code, but seems promising. Thanks!

PR #6496, opened 2022-04-21
Still show the team dropdown even if the user doesn't have any teams

### LINKED ISSUE #6344: When onboarding, clicking the "edit" link looks wrong
image

The dropdown disappears, leaving this

jaril:
Fix #6344.

image


bvaughn (approved):
Interesting. Should we at least _disable_ the dropdown in the event where there's only one item? Seems a little confusing to show a dropdown with only one choice (akin to showing a radio button with only one option).

Seems fine as a quick fix though for the spacing/alignment issue.

PR #6499, opened 2022-04-21
Use jest to run e2e tests

hbenl:


test/e2e/runTest.ts
```diff
@@ -0,0 +1,37 @@
+import config from "./config";
```
jasonLaster, line 1: this file should probably be `runClassicTest`

test/e2e/tests/breakpoints-01.test.ts
```diff
@@ -0,0 +1,12 @@
+import config from "../config";
+import { runClassicTest } from "../runTest";
+
+it("Test basic breakpoint functionality.", async () => {
```
jasonLaster, line 4: I think these `classic` tests could be expressed through `jest-in-case` as well which would cut down on the duplication quite a lot. I think you'd still be able to get all of the jest watch utilities as well

https://www.npmjs.com/package/jest-in-case

hbenl: I had considered something similar, but failed to get it to work with the VS Code integration for jest. But most importantly this is the structure I think it will have when we move away from the classic tests.

.github/actions/test/action.yml
```diff
@@ -15,7 +14,7 @@ runs:
       shell: bash
     - run: chmod +x ./replay-node/macOS-replay-node
       shell: bash
-    - run: node ${{ github.action_path }}/../../../test/run.js
+    - run: npm run test-e2e -- --shard ${{ env.INPUT_STRIPE }}
```
jcmorrow, line 17: this is a nit, but if we already have the term `stripe` can we keep calling them stripes? I'd prefer not to use two words where one will suffice (or maybe there's actually a difference between a shard and a stripe that I am missing?)

test/e2e/config.ts
```diff
@@ -0,0 +1,22 @@
...
+  headless: !!process.env.RECORD_REPLAY_PLAYWRIGHT_HEADLESS,
+  nodePath: process.env.RECORD_REPLAY_NODE,
+  // This API key allows access to the whole team that holds all these test
+  // recordings. If someone wanted to, they could go in and delete the workspace
+  // or recordings in it or anything. While thats not great, it's also not
+  // the end of the world. If someone does that, we can always change
+  // this code to only run in CI in the main repo and have this as a secret.
+  // It's a lot easier to hardcode it for now though.
+  // trunk-ignore(gitleaks/generic-api-key)
+  replayApiKey: "rwk_7XPbO5fhz0bkhANYXtN2dkm74wNQCchXf2OxVgAerTQ",
```
jcmorrow, line 20: I was going to suggest using this as an opportunity to put this into a GH secret, but then we wouldn't get test recordings on forks, which would be annoying for external contributors. Still, I'd love to ✂️  it.

test/e2e/harness.ts
```diff
@@ -0,0 +1,34 @@
+import { Page } from "@recordreplay/playwright";
+
+declare global {
+  const app: any;
```
jcmorrow, line 4: If we have types for app state and app actions/selectors (which I think we have at least *some* of), it might be nice to give this just a slightly more deliberate type. Eh, on second thought, I don't know if it's worth it. I leave it up to you and other reviewers to decide if there is any value there.

test/e2e/harness.ts
```diff
@@ -0,0 +1,34 @@
...
+        app.selectors.getViewMode() == "dev" &&
+        document.querySelector(".webconsole-app")
+      );
+    });
+  }
+
+  waitForMessage(text: string) {
+    return this.page.waitForFunction(text => {
+      const messages = document.querySelectorAll(".webconsole-output .message");
+      return [...messages].some(message => (message as HTMLElement).innerText.includes(text));
```
jcmorrow, line 31: I'm kind of surprised that there isn't a `expect(this.page).toHaveSelector` or something we could use here?

test/e2e/runTest.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  if (!success) {
+    console.log("Test failed, uploading recording");
+    const recordingId = await uploadLastRecording(testUrl);
+    console.error(
+      `::error ::Failure ${args.script}: https://app.replay.io/recording/${recordingId}`
+    );
+  }
+  expect(success).toBe(true);
```
jcmorrow, line 46: 🤩

test/e2e/tests/autocomplete.test.ts
```diff
@@ -0,0 +1,8 @@
+import { runClassicTest } from "../runTest";
+
+it("Test autocomplete in the console.", async () => {
```
jcmorrow, line 3: Super nit, but can we either turn these `it(...)` statements into sentences, something like `it("autocompletes the console input")`, or use `test("autocomplete in the console")` (I think the latter probably makes more sense here personally).

test/e2e/tests/breakpoints-01.test.ts
jcmorrow: Did it work from the command line? What was breaking in VSCode? I've never run the tests from VSCode, but I'd be happy to help try and get it to work, getting rid of that duplication would be 🤌

test/e2e/utils.ts
```diff
@@ -0,0 +1,22 @@
+import os from "os";
```
jcmorrow, line 1: It's so weird that GH does not syntax highlight these files as typescript? This one doesn't even have the `.test.ts`, so there's really no reason for it not to 🤔

jcmorrow (commented):
Sweeeeeet, this is looking really good. I left some suggestions but nothing too major!

test/e2e/config.ts
jasonLaster: probably a follow up :)

test/package.json
```diff
@@ -7,18 +7,14 @@
...
-    "eslint-plugin-react": "^7.20.6",
-    "eslint-plugin-react-hooks": "^4.3.0"
+    "@jest/console": "^28.0.0-alpha.9",
+    "@replayio/replay": "^0.1.2",
+    "babel-jest": "^28.0.0-alpha.11",
+    "jest": "^28.0.0-alpha.11",
+    "next": "^12.1.5"
   },
   "optionalDependencies": {
-    "@recordreplay/playwright": "^1.10.6"
```
jasonLaster, line 21: These package changes seem like a nice win

jcmorrow:
Oh and also - unit tests are not running for me, it looks like some of the transform config from e2e might have also slipped into the unit test config 😞

test/e2e/tests/console_stacks.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      await harness.waitForMessage("ConsoleWarn");
+      await harness.waitForMessage("ConsoleError");
+      await harness.waitForMessage("Assertion failed: ConsoleAssert");
+      await harness.waitForMessage("Error: UncaughtError");
+      await harness.waitForMessage("Object { number: 42 }");
+      await harness.waitForMessage("Object { number: 12 }");
+      await harness.waitForMessage("uncaught exception: [object Object]");
+    },
+  });
+});
```
jasonLaster, line 29: Love this new test.  What do you think about putting this test into our existing test runner? This way we can 

1. write playwright and classic tests today 
2. look into using jest as a follow up for the new style and migrate the tests over

jasonLaster (commented):
This PR is doing a lot and I can see why it has been difficult to get over the finish line. Frankly our existing test runner has a ton of config options and use cases.

At this point, I think it would be helpful to look for places to break this down and land incrementally. This will also let us balance the test work with the many go/sprint items as well.

package.json
```diff
@@ -164,6 +166,8 @@
     "graphql": "15.8.0",
     "graphqurl": "^1.0.1",
     "identity-obj-proxy": "^3.0.0",
+    "jest": "^28.0.0-alpha.11",
```
bvaughn, line 169: Should we be using [`28.0.1`](https://www.npmjs.com/package/jest/v/28.0.1) rather than [`28.0.0-alpha.11`](https://www.npmjs.com/package/jest/v/28.0.0-alpha.11)
?

test/e2e/runTest.ts
hbenl: It now also contains `runPlaywrightTest`.

.github/actions/test/action.yml
hbenl: This is a jest option, so we can't rename it.

test/e2e/harness.ts
hbenl: We have an `AppHelpers` interface for that, I tried that but it's incomplete and didn't contain the methods I needed so I postponed this.

test/e2e/harness.ts
hbenl: `expect` comes from jest, `page` comes from playwright so I'd be surprised if that exists.

package.json
hbenl: Yes, I've updated the PR.

hbenl:
> unit tests are not running for me, it looks like some of the transform config from e2e might have also slipped into the unit test config

No, this is a regression in jest 28 (which we need for the `--shard` option). See facebook/jest#12750

hbenl:
> At this point, I think it would be helpful to look for places to break this down and land incrementally.

We *could* just remove the change to the GitHub action and the jest version and land this. Then we'd still have our old test runner (and CI would still be using that) in parallel with the new version of the tests.

test/e2e/tests/breakpoints-01.test.ts
hbenl: In the end I decided I could live without VS Code integration, but there's another reason why this approach wouldn't work: jest's `--shard` option, which we need in CI, puts all tests from the same file in the same "shard".
I think we could hack our own stripe/shard implementation and then we could also follow this approach. Not sure if it's worth it though.

jasonLaster (approved):

hbenl:
Here's what changes:
- to run the e2e tests you can now use `npm run test-e2e` instead of `node test/run`
- but the old test runner is still there (for now), so `node test/run` will still work
- devtools CI uses jest for running the tests, the backend still uses the old runner (for now)
- the environment variables for configuring the tests are the same as before (see `test/e2e/config.ts`)
- configuring the tests with cli arguments doesn't work with jest, you'll have to use the corresponding environment variables
- jest doesn't use `test/manifest.js` to register the tests, instead there is one file per test under `test/e2e/tests` that registers the test using jest's `it(description, testFn)`
- to get VS Code integration, install one of the jest test runner extensions and configure it to use `jest-e2e.config.js` instead of `jest.config.js`
- to run a single test on the command line you can use e.g. `npm run test-e2e -- --testMatch "**/breakpoints-01.js"` or `npm run test-e2e -- -t "Test basic breakpoint functionality"`
- we want to rewrite the e2e tests using the Playwright API, I've already done this for `test/e2e/tests/console_stacks.test.ts`, more will follow

.github/actions/test/action.yml
jcmorrow: Ah, ok, makes sense! 👍

bvaughn:
Oh cool. Do the fixture test generating scripts Jason and I added still work with this new approach? Can't run and test at the moment because I'm on a flight.

test/e2e/harness.ts
jcmorrow: It looks like we are at least implicitly defining this as

```
app: {
  actions: any;
  selectors: any;
} | undefined;
```

So that could be worth being explicit about, but not a huge deal.

test/e2e/harness.ts
```diff
@@ -0,0 +1,34 @@
...
+  async start() {
+    await this.page.waitForFunction(() => window.app?.actions);
+    this.page.evaluate(() => window.app.actions.setViewMode("dev"));
+    await this.page.waitForFunction(() => {
+      const loadedRegions = app.selectors.getLoadedRegions();
+      const fullyLoaded =
+        loadedRegions &&
+        loadedRegions.loading.length > 0 &&
+        loadedRegions.loaded.length > 0 &&
+        loadedRegions.loading[0].end.point === loadedRegions.loaded[0].end.point;
```
jcmorrow, line 19: We don't need to wait for indexing as well here?

hbenl:
> Do the fixture test generating scripts Jason and I added still work with this new approach?

Not yet. I'll port them over tomorrow.

test/e2e/harness.ts
jcmorrow: https://playwright.dev/docs/test-assertions#locator-assertions-to-have-text

test/e2e/runTest.ts
```diff
@@ -0,0 +1,74 @@
...
+  const url = `${config.devtoolsUrl}/recording/${exampleRecordingId}`;
+  try {
+    await recordPlaywright(config.browserName, async page => {
+      await page.goto(url);
+      await args.script(page);
+    });
+  } catch (e) {
+    const recordingId = await uploadLastRecording(url);
+    console.error(
+      `::error ::Failure ${args.script}: https://app.replay.io/recording/${recordingId}`
```
jcmorrow, line 70: This is a small one, but what do you think about pulling out a `reportError` function or something since we have very specific formatting here and on line 56?

bvaughn:
Okay. Let me know if you need a hand, when you get there.

test/e2e/recordNode.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+export async function recordNodeExample(scriptPath: string) {
+  if (!config.nodePath) {
+    console.log(`Skipping test: RECORD_REPLAY_NODE not set`);
+    return;
+  }
+
+  const recordingIdFile = tmpFile();
```
jcmorrow, line 30: This can totally be a follow-up, but for simplicity I would love to remove all of this random tempfile stuff, and instead use a predictable scheme based on the test-name and current time. I like to have values that have at least some clear meaning, makes debugging stuff a lot easier!

jcmorrow (approved):
Bit of a bummer to see all of the duplication still around, but it's a WIP, and this is good progress 👍  Our test running code is (I think) much easier to understand now than it was six months ago, and I bet it will be even better in another six months!

test/e2e/harness.ts
hbenl: Apparently not, I copied this verbatim from the old harness: https://github.com/RecordReplay/devtools/blob/09040ae4b06002de3ff629964ac120b2289caeee/src/test/harness.js#L86-L89

test/e2e/harness.ts
hbenl: Oh, nice. However, it doesn't work for some reason: `page.locator()` doesn't exist according to typescript. So I tried with `(page as any).locator()`, which works, but then `expect(locator).toHaveText(...)` throws "TypeError: expect(...).toHaveText is not a function". I'll investigate this in a follow-up.

PR #6516, opened 2022-04-21
Gracefully handle the case where there is no selected request

### LINKED ISSUE #6510: Replay crashes when the zoom region gets too small?
Loom: https://www.loom.com/share/fc4539236ed14722bbf1cf66a48f0a7b

Replay: https://app.replay.io/recording/replay-crashes-when-zoom-region-too-small--d46e4c5b-218c-424f-bb69-08a612510ceb

bvaughn:
Resolves #6510

Previously we were only checking for a selected request ID but not ensuring that we could match it to a request within the loaded region. I'm really new to this code so I'm not sure if this indicates a broader unexpected issue that should be dug into further– but conceptually, it seems like a weak guard we had in place before so I hardened it a little bit.

Best viewed [without whitespace](https://github.com/RecordReplay/devtools/pull/6516/files?w=1).

### DIFF
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -89,49 +89,51 @@ export const NetworkMonitor = ({
 
   return (
     <Table events={events} requests={requests} types={types}>
-      {({ table, data }: { table: any; data: RequestSummary[] }) => (
-        <div className="flex h-full min-h-0 flex-col" ref={container}>
-          <FilterBar types={types} toggleType={toggleType} table={table} />
-          <SplitBox
-            className="min-h-0 border-t border-splitter"
-            initialSize="350px"
-            minSize={selectedRequestId ? "30%" : "100%"}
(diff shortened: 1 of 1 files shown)


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -89,49 +89,51 @@ export const NetworkMonitor = ({
...
-                table={table}
-                data={data}
-                currentTime={currentTime}
-                onRowSelect={row => {
-                  trackEvent("net_monitor.select_request_row");
-                  dispatch(fetchFrames(row.point));
+      {({ table, data }: { table: any; data: RequestSummary[] }) => {
+        let selectedRequest;
+        if (selectedRequestId) {
+          selectedRequest = data.find(request => request.id === selectedRequestId);
```
bvaughn, line 95: If it's unexpected that we'd ever _not_ find a match here, I could add a `console.warn` log or something.

jcmorrow (approved):

PR #6524, opened 2022-04-22
Keep showing the total points in a breakpoint panel even while in playback

### LINKED ISSUE #6420: Print statement panel shows loading when playing
**Description**
The panel summary shows "loading" when the video is playing. This is probably a recent regression
https://app.replay.io/recording/your-first-replay--11acf13c-845b-489a-a1f7-5c2326052ede

jaril:
Fix #6420.

[Replay of the changes](https://app.replay.io/recording/playback-count--f5f028c6-5d34-4fec-b789-37cfa35a8301)
[Loom walkthrough](https://www.loom.com/share/48e6284516094c07ae014312209dcd4b)

## Preview
https://user-images.githubusercontent.com/15959269/164608323-52a01e3c-0502-4cd3-bb7b-443fbc9cde2d.mov

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
@@ -3,6 +3,17 @@ import { getIsIndexed } from "ui/reducers/app";
 import { AnalysisError, AnalysisPayload } from "ui/state/app";
 import { getExecutionPoint } from "../../../selectors";
 
+const getPreviousPointsCount = (analysisPoints: AnalysisPayload, executionPoint: string | null) => {
+  if (!executionPoint) {
+    return "…";
+  }
+
+  const previousPoints = analysisPoints.data.filter(
+    point => BigInt(point.point) <= BigInt(executionPoint)
+  );
+  return previousPoints.length;
+};
+
 export function PanelStatus({ analysisPoints }: { analysisPoints: AnalysisPayload }) {
   const executionPoint = useSelector(getExecutionPoint);
   const isIndexed = useSelector(getIsIndexed);
@@ -11,18 +22,16 @@ export function PanelStatus({ analysisPoints }: { analysisPoints: AnalysisPayloa
 
   if (!isIndexed) {
     status = "Indexing";
-  } else if (!analysisPoints || !executionPoint) {
+  } else if (!analysisPoints) {
     status = "Loading";
   } else if (analysisPoints.error) {
     status = analysisPoints.error === AnalysisError.TooManyPoints ? "10k+ hits" : "Error";
   } else if (analysisPoints.data.length == 0) {
     status = "No hits";
   } else {
-    const points = analysisPoints
-      ? analysisPoints.data.filter(point => BigInt(point.point) <= BigInt(executionPoint))
-      : [];
+    const pointsCount = getPreviousPointsCount(analysisPoints, executionPoint);
 
-    status = `${points.length}/${analysisPoints.data.length}`;
+    status = `${pointsCount}/${analysisPoints.data.length}`;
     maxStatusLength = `${analysisPoints.data.length}/${analysisPoints.data.length}`.length;
   }
 


src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -3,6 +3,17 @@ import { getIsIndexed } from "ui/reducers/app";
 import { AnalysisError, AnalysisPayload } from "ui/state/app";
 import { getExecutionPoint } from "../../../selectors";
 
+const getPreviousPointsCount = (analysisPoints: AnalysisPayload, executionPoint: string | null) => {
+  if (!executionPoint) {
+    return "…";
```
bvaughn, line 8: This branch feels off to me.

I guess fundamentally, I think that a function named "get count" should always return a number.

In this case, we're trying to differentiate between "it's this specific number in the set of numbers" and "it's not in the set of numbers" then maybe `-1` (like `indexOf`) would be a good value to signal that?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -11,18 +22,16 @@ export function PanelStatus({ analysisPoints }: { analysisPoints: AnalysisPayloa
...
   } else if (analysisPoints.data.length == 0) {
     status = "No hits";
   } else {
-    const points = analysisPoints
-      ? analysisPoints.data.filter(point => BigInt(point.point) <= BigInt(executionPoint))
-      : [];
+    const pointsCount = getPreviousPointsCount(analysisPoints, executionPoint);
 
-    status = `${points.length}/${analysisPoints.data.length}`;
+    status = `${pointsCount}/${analysisPoints.data.length}`;
```
bvaughn, line 34: The format of "…/5" doesn't seem meaningful or intuitive to me.

Maybe it's because I think of "…" as a _loading_ indicator? This is meant to indicate "not a match".

My first thoughts are that:
* It's useful to keep showing the total count of things (`analysisPoints.data.length`)
* Maybe we could show the closest previous match but dim it out in some way? This could convey "you're roughly here within the total set of these, but you're not exactly on any one of them".

So like, if I selected the 23rd item the UI would look like this:
Screen Shot 2022-04-22 at 11 08 07 AM

And then if I played ahead a little– but not far enough to hit the 24th item, it might look like this?
Screen Shot 2022-04-22 at 11 08 34 AM

bvaughn (commented):
Left some thoughts. What do you think?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
jaril: Oh good call, yeah let me clean that up. And then interpret the `-1` as "…"

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
jaril: Now that I look at it, maybe the ✱ does look a bit better? Not that it's the best, but at least that way it doesn't look like we're loading.

bvaughn: > The main problem here is that mainly, during playback, we can't figure out what that numerator is. The executionPoint that lets us order your current pause on/in-between hits isn't available.

That makes sense, that there's no `point` if you're scrubbing, but I (maybe naively) assumed that we could use the current _time_ value to find the nearest point. The only ambiguous cases that come to my mind are:
* If there _multiple_ points at the current time _and_ you aren't actively viewing one, in which case I think it's okay for to just show the last one?
* If you scrubbed _before_ the first point, in which case I think it's okay to show a greyed out "0"?

> Now that I look at it, maybe the ✱ does look a bit better? Not that it's the best, but at least that way it doesn't look like we're loading.

TBH I think the character that most comes to my mind as a "doesn't match" or "unavailable" is `-` but (that's also not a great match here).

jasonLaster: Agreed that we can approximate it based on time. while scrolling at best we're dealing with pixel level fidelity...

jonbell-lot23:
Hey @jaril is this good to go? Going through UI polish tickets and this came up.

bvaughn:
> [Edit: I see there's a merge conflict, but other than that -- ready to go?]

I think there's still some minor PR feedback to be addressed.

jcmorrow:
Closing in favor of https://github.com/RecordReplay/devtools/pull/6524

jaril:
> Closing in favor of #6524

You meant #6663 right? :P

jcmorrow:
> > Closing in favor of #6524
> 
> You meant #6663 right? :P

Haha, yes, sorry, I did not even realize GitHub auto-complete would let you tag a PR from itself 🤦

PR #6525, opened 2022-04-22
upgrade trunk and remove deprecated repo section from config

mmatheson:


bvaughn (approved):
Looks fine to me. What's changing in 10.1?

PR #6533, opened 2022-04-22
Add qualifier filters to the library

jaril:
image

This PR adds some of the scaffolding necessary to have [advanced search syntax](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests). For now we're going to support two qualifiers (`created` and `target` ). This allows us to have a dropdown where the user can specify if they want to only see replays recorded in the last 7 days, or replays that targeted node recordings.

Some notes:
- The filtering will still be done client side. It won't be for much longer — the immediate follow-up to this will be to figure out how to ask the backend for a filtered list of recordings based on the current filter string.
- The qualifiers aren't fully implemented. Right now, only the necessary parts to support the dropdown have been implemented. So `created` doesn't support comparison prefixes (eg >, <=), and `target` can only filter for `node` targets.
- One change this introduces is we will no longer filter the displayed replays whenever the filter string changes. Instead, we now require the user to explicit commit to the filtering by pressing enter.
- The URL is updated to take a url param `q` which will contain the currently-applied filter string, so that the filter persists on refresh and users can share the link. Those aren't particularly big use cases for now, but it wasn't much more work to put in and it's something which we will need for the test suite use case anyway, so I decided to include it here.

Fix #6533.

### DIFF
--- a/src/ui/components/Library/FilterBar.tsx
+++ b/src/ui/components/Library/FilterBar.tsx
@@ -1,8 +1,9 @@
-import React, { ChangeEvent, KeyboardEvent } from "react";
+import React, { ChangeEvent, KeyboardEvent, useContext } from "react";
 
 import { TextInput } from "../shared/Forms";
 
 import { FilterDropdown } from "./FilterDropdown";
+import { LibraryFilters, LibraryFiltersContext } from "./useFilters";
 
 export function FilterBar({
   displayedString,
@@ -23,14 +24,32 @@ export function FilterBar({
   };
 
   return (
-    <>
+    <div className="flex flex-grow relative items-center space-x-3">
       <FilterDropdown setAppliedText={setAppliedText} />
-      <TextInput
-        value={displayedString}
-        onChange={onChange}
-        placeholder="Search"
-        onKeyDown={onKeyPress}
-      />
-    </>
+      <div className="flex flex-grow">
+        <TextInput
+          value={displayedString}
+          onChange={onChange}
+          placeholder="Search"
+          onKeyDown={onKeyPress}
+        />
+        <InvalidFilterWarning />
+      </div>
+    </div>
+  );
+}
+
+function InvalidFilterWarning() {
+  const filters = useContext(LibraryFiltersContext);
+  const errantQualifier = Object.entries(filters.qualifiers).find(q => q[1].error);
+
+  if (!errantQualifier) {
+    return null;
+  }
+
+  return (
+    <div className="bg-errorBgcolor rounded-md border border-errorColor text-errorColor p-2 text-xs absolute transform translate-y-full mt-1">
+      Invalid filter string: {errantQualifier[1].error}
+    </div>
   );
 }
--- a/src/ui/components/Library/FilterBar.tsx
+++ b/src/ui/components/Library/FilterBar.tsx
@@ -6,27 +6,25 @@ import { FilterDropdown } from "./FilterDropdown";
 
 export function FilterBar({
   displayedString,
-  setAppliedString,
-  setDisplayedString,
-  applyDisplayedString,
+  setDisplayedText,
+  setAppliedText,
 }: {
   displayedString: string;
-  setAppliedString: (str: string) => void;
-  setDisplayedString: (str: string) => void;
-  applyDisplayedString: () => void;
+  setDisplayedText: (str: string) => void;
+  setAppliedText: (str: string) => void;
 }) {
   const onChange = (e: ChangeEvent<HTMLInputElement>) => {
-    setDisplayedString(e.target.value);
+    setDisplayedText(e.target.value);
   };
   const onKeyPress = (e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === "Enter") {
-      applyDisplayedString();
+      setAppliedText(e.currentTarget.value);
     }
   };
 
   return (
     <>
-      <FilterDropdown setSearchString={setAppliedString} />
+      <FilterDropdown setAppliedText={setAppliedText} />
       <TextInput
         value={displayedString}
         onChange={onChange}
--- a/src/ui/components/Library/FilterBar.tsx
+++ b/src/ui/components/Library/FilterBar.tsx
@@ -1,0 +1,38 @@
+import React, { ChangeEvent, KeyboardEvent } from "react";
+
+import { TextInput } from "../shared/Forms";
+
+import { FilterDropdown } from "./FilterDropdown";
+
+export function FilterBar({
+  displayedString,
+  setAppliedString,
+  setDisplayedString,
+  applyDisplayedString,
+}: {
+  displayedString: string;
+  setAppliedString: (str: string) => void;
+  setDisplayedString: (str: string) => void;
+  applyDisplayedString: () => void;
+}) {
+  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setDisplayedString(e.target.value);
+  };
+  const onKeyPress = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Enter") {
+      applyDisplayedString();
+    }
+  };
+
+  return (
+    <>
+      <FilterDropdown setSearchString={setAppliedString} />
+      <TextInput
+        value={displayedString}
+        onChange={onChange}
+        placeholder="Search"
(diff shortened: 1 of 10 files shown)


src/ui/components/Library/Viewer.tsx
```diff
@@ -47,26 +48,45 @@ function DownloadLinks() {
...
+      : filteredRecordings;
+  filteredRecordings =
+    (qualifiers.created && new Date(qualifiers.created))
+      ? filteredRecordings.filter(
+          r => new Date(r.date).getTime() > new Date(qualifiers.created!).getTime()
+        )
+      : filteredRecordings;
+
+  return filteredRecordings;
+};
```
jasonLaster, line 72: this is actually really interesting. I hadn't thought about how we could parse on the client and then just pass variables to our graphQL recordings query...

jaril:
Old notes: 

This is so not done don't look!

### Spec
Allow the users to use GitHub-style filter syntax in the library's search box.
- [x] It should be able to handle a `created` filter, e.g. `created:YYYY-MM-DD`
- [x] It should be able to handle a `target` filter, e.g. `target:node`
- [x] Don't apply filters unless the user hits enter
- [x] Filters should be persisted in the URL so they stay intact on refresh

### Follow-ups
- [x] ~~Better parsing for dates~~
  - GitHub parses comparison qualifiers (e.g. <2022-04-22, >=2022-04-22). Since we're moving this to the backend anyway, we can just handle that qualifier on the backend when we start working on that bit. 
- [x] ~~More search param magic~~
  - The following items primarily handles sharing URLs between multiple replay users. This version is limited since it's mostly for individual use, so we can skip it for now.
  - [ ] If the user opens a team link with filters in the URL, but they're not a part of that team, they should be redirected to their main library
  - [ ] The library should be able to use the filters in the URL when fetching the recordings
- [x] ~~Changes should apply after the user hits enter~~
  - Since this is a half-measure for now, skipping this. But we _need_ to implement change committing with enter immediately after as we do this properly on the backend.
  - Felt strongly about this even if we were doing the filtering on the server. The current implementation does the filtering on the client (for now), on each keypress, and perf is _bad_. So we should fix this 100%.

src/ui/components/Library/useFilters.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  const applyDisplayedString = () => {
+    setAppliedString(displayedString);
+    updateUrlWithParams({ q: displayedString });
+  };
+  const forceSetAppliedString = (newStr: string) => {
+    setAppliedString(newStr);
+    setDisplayedString(newStr);
+    updateUrlWithParams({ q: newStr });
+  };
```
bvaughn, line 74: The naming in this hook is confusing to me.

There's a "display string" and an "applied string" but then there's a method called "apply displayed string" 🙃 

Also `forceSetAppliedString` updates both applied and displayed strings, but `applyDisplayedString` updates only _applied_ string which is the... opposite of what I'd expect 😆 

Can we change these names to be clearer?

I think `displayedString` and `appliedString` make sense 👍🏼  but the callbacks could be made clearer? Seems like there are 3 use-cases:
1. Update the displayed string only (as a user types).
2. Update the displayed and applied strings (choosing a pre-set filter from the dropdown).
3. Applying the current display string (as a user presses "Enter" after updating the filter).

I think we could simplify to the following two functions:
* `setDisplayText` (alias to state updater) – handles case 1
* `setAppliedText` (updates both displayed and applied texts) – handles cases 2 and 3

bvaughn (commented):
This is a great start! Here's some initial feedback

jaril:
> * The filtering will still be done client side. It won't be for much longer — the immediate follow-up to this will be to figure out how to ask the backend for a filtered list of recordings based on the current filter string.

Adding this but stopping short of syntax highlighting (for now). Will give you a heads up when it's ready!

bvaughn:
Having trouble with this branch. When I run `npm run dev` I get an error about `devtools/client/debugger/src/actions/ui` missing.

src/ui/components/Library/ViewerRouter.tsx
```diff
@@ -21,15 +21,15 @@ function ViewerLoader() {
...
+function MyLibrary() {
   const { recordings, loading } = hooks.useGetPersonalRecordings();
   const { loading: nonPendingLoading } = hooks.useGetNonPendingWorkspaces();
 
   if (loading || nonPendingLoading || recordings == null) {
     return <ViewerLoader />;
   }
 
-  return <Viewer {...{ recordings, workspaceName: MY_LIBRARY, searchString }} />;
+  return <Viewer {...{ recordings, workspaceName: MY_LIBRARY }} />;
```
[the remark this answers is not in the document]
bvaughn: There a reason we're leaving this as-is? Masks potential TypeScript problems 😄  Seems easy to fix!

src/ui/components/Library/ViewerRouter.tsx
```diff
@@ -122,7 +115,7 @@ function ViewerRouter(props: ViewerRouterProps) {
   }
 
   if (currentWorkspaceId === null && features.library) {
-    return <MyLibrary {...props} />;
+    return <MyLibrary />;
```
bvaughn, line 118: 👍🏼

src/ui/components/Library/useFilters.ts
```diff
@@ -0,0 +1,125 @@
...
+  const [displayedString, setDisplayedString] = useState(str);
+
+  const setDisplayedText = (newStr: string) => {
+    setDisplayedString(newStr);
+  };
+  const setAppliedText = (newStr: string) => {
+    setAppliedString(newStr);
+    setDisplayedString(newStr);
+    updateUrlWithParams({ q: newStr });
+  };
```
bvaughn, line 103: Thanks! Looks much clearer now!

src/ui/components/Library/FilterBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: {
+  displayedString: string;
+  setDisplayedText: (str: string) => void;
+  setAppliedText: (str: string) => void;
+}) {
+  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setDisplayedText(e.target.value);
+  };
+  const onKeyPress = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Enter") {
```
bvaughn, line 21: What should we do on _blur_ (or "Escape" key)?

Currently the pending filter text _won't_ be applied, which is good, but there's no visual indicator that the filter text is out of sync with the filters you're viewing.

Seems like we should discard the pending text on blur/escape?

bvaughn (approved):
👍🏼 

Would like to see the blur/escape case explicitly handled, but otherwise this looks good!

src/ui/components/Library/FilterBar.tsx
jaril: I was looking for some examples of how this is handled and it looks like the answer is, it isn't? Specifically in the GitHub filter example, pressing escape while focused is a noop, and if you do blur by clicking elsewhere, the pending change isn't discarded. Not a big fan for the same reasons you indicated. But at the same time, I don't think I'd expect for the text to be discarded on blur.

I think the main problem is that we don't indicate to the user what filter text is currently applied. If that were somewhere, then at least you can somewhat infer whether your change has been applied or not.

Maybe we should show something like, "Showing 89 replays that match <insert-filter-text-here>"? That gets wordy though.

bvaughn: Ah, good point. GitHub is an interesting example.

I agree with you that blowing it away on blur might not be great, but I think it's not a given that people will know to press Enter to confirm. Could we just apply on blur? (And maybe discard on Escape?)

Alternately maybe we could use color to _hint_ at an unsaved/unapplied change?

jaril: It's probably not optimal but I think I'm satisfied with being at parity here (for now) and then we can readjust if it's a big problem. Let me know what you think!

bvaughn: Not worth holding up the merge for this, but I'd prefer we follow up with a fix to more gracefully handle that case 😁  I don't mind doing it.

jaril: Down, thanks! I'll try to race you to it too

Issue #6511, opened 2022-04-21
Recording stop loading certain Network requests


bvaughn:
These requests were loading prior to 20 minutes ago or so, then some of them stopped. Not sure why.

Loom: https://www.loom.com/share/9eadf59ecd71497bafa27df1089c888a

Replay: https://app.replay.io/recording/replay-network-request-details-not-loading--fa66a20b-1eb6-48ee-b9c7-3cb1c0b6c46e

bvaughn:
It seems like when this is happening, `loadedRegions` is undefined. Is that expected?

jcmorrow:
@bvaughn is that coming from Replay or did you see that in the browser JS console?

bvaughn:
@jcmorrow I see it in my locally running Replay. I can see the no-details line being hit in my session too but reasoning about what's happening there is a bit harder due to source maps. Still looking.

jcmorrow:
well, just beware, I made this: https://github.com/RecordReplay/devtools/issues/6512 because I don't believe that `loadedRegions` should be null when this code path gets executed. That just makes no sense to me.

bvaughn:
Do you suspect the Apollo hooks for this? I got side tracked digging into the comments issue b'c of these paper cuts but I think I was seeing some confusing things in terms of what the cache should have (response returned from server after mutation and get) and what was actually being passed to React during render.

jcmorrow:
Sorry I didn't see this comment til now. So, I'm sure you know that Apollo sits between the client and the GraphQL API. So the stuff that goes over that is stuff like comments, basic recording data (id's titles, which workspace it's attached to, etc.), some information about users and orgs, billing info, stuff like that. Everything that goes over the protocol, information about sources, console logs, really most of what makes devtools devtools and not just another web app with users and CRUD resources goes over our Protocol websocket API, and because none of it is GraphQL, none of it touches Apollo (at least... as far as I know 🙃 ). Anyways, I too see 👻 s in Apollo with some regularity, but unless you're dealing with the Library page or comments, it's pretty unlikely that the data you are seeing ever touched Apollo.

jcmorrow:
[...]

Issue #5606, opened 2022-03-04
[UX] Fetching network request details in an unloaded region


jaril:
This is a follow up item to #5605 which addressed #5282.

We need to consider what should happen when the user tries to click on a request that's in an unloaded region.

Things to consider:
- A link to docs about loaded/unloaded regions and why it affects things you can see in the UI.
- If the current unloaded region will eventually be loaded, then some auto-refresh capability so the request details show up immediately once it's ready.
- If the current unloaded region has been permanently yanked away by the backend, some explanation about what had happened and how they should probably use focus mode so that we can load what they're interested in.
- Maybe we shouldn't allow users to click into a request in an unloaded region in the first place.

jonbell-lot23:
Hello! Here's the design proposal.

First, we should allow people to click in rather than disallowing the click at all. This lets us message better.

**1. If it will eventually load**

* Use our standard progress bar across the top
* Text: something like "Loading..."

**If it will never load because it's outside of Focus Mode's window**

* Text: something like "This request is outside of Focus Mode. Click here to resize Focus Mode."

**If it will never load for a reason that the user can't fix**

* Text: "Apologies, this request is not available. Learn more about what can cause this error." <-- link to docs

jonbell-lot23:
Assigning to @jaril for discussion and implementation

jaril:
> Assigning to @jaril for discussion and implementation

Let me take a stab at it the way you proposed and let's feel it out!

jaril:
Carving off #1 — pretend that the unloaded network details will be loaded at some point, and in the meantime we should show the progress bar and loading text. Then swap it out once it's available.

#3 is more difficult so we'll address it separately afterwards.

jaril:
Just noticed that #1 is actually being handled with the current behavior. Will close for now.

bvaughn:
Hey @jaril and @jonbell-lot23!

I think this UX could still use some improvement. (See #6511 where I thought the current behavior was actually a bug/something broken.)

> > 1. **If it will eventually load**
>> Use our standard progress bar across the top
>> Text: something like "Loading..."
>
> Just noticed that https://github.com/RecordReplay/devtools/pull/1 is actually being handled with the current behavior. Will close for now.

[...]

PR #6534, opened 2022-04-22
Unloaded Network requests should be dimmed and unseletable

bvaughn:
Relates to #5606 and #6511

Loom walk through of the change: https://www.loom.com/share/f2366f55c10c4dd3a54b261b27726918

This PR changes the way Network requests are treated for "unloaded" regions:
* Network requests that are unloaded are still _shown_ in the Network panel, but they are dimmed and cannot be opened by clicking on them (or on a comment that references them).
* The logic to filter the selected Network request based on whether it was in a loaded region has been moved out of a React component effect and into a Redux thunk (to prevent invalid state and potential tearing/flickering).

Potential follow up changes:
* Should also disable the fast forward and rewind buttons for these requests too? That would probably be a bigger change since that affects more than just the network state. (If we do that I'd like to see us do it as a separate PR.)


src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -25,10 +28,28 @@ interface Detail {
...
+  point: ExecutionPoint;
+}) {
   const dispatch = useDispatch();
   const closePanel = () => dispatch(hideRequestDetails());
 
+  let message;
+  if (loadedRegions && isPointInRegions(loadedRegions.loading, point)) {
+    message = (
+      <>
+        <LoadingProgressBar />
```
bvaughn, line 45: This progress bar doesn't seem great but in the absence of an indeterminate bar, it seemed like the right one to use here. (I don't think it's worth the effort of us trying to estimate an actual progress value for this use case, but maybe others disagree?)

src/devtools/client/themes/variables.css
```diff
@@ -87,6 +87,7 @@
   --grey-10-a20: rgba(249, 249, 250, 0.2);
   --grey-10-a25: rgba(249, 249, 250, 0.25);
   --grey-10-a30: rgba(249, 249, 250, 0.3);
+  --grey-10-a40: rgba(249, 249, 250, 0.4);
```
bvaughn, line 90: cc @jonbell-lot23 in case this change is controversial. 

I wanted two distinct levels of "dimmed" for this UI, while the text also still remained somewhat legible.

src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -93,6 +101,16 @@ export const NetworkMonitor = ({
...
         if (selectedRequestId) {
           selectedRequest = data.find(request => request.id === selectedRequestId);
+          if (selectedRequest && loadedRegions) {
+            if (!isTimeInRegions(selectedRequest.point.time, loadedRegions.loaded)) {
+              // Unloaded requests should not be selectable.
+              //
+              // TODO I'm not sure this logic is appropriate for the view,
+              // especially a lower level view like NetworkMonitor, to decide.
+              // This risks "tearing" (where different views show different behavior).
+              selectedRequest = undefined;
```
bvaughn, line 111: This is not great. We're still _selecting_ the request in our state; we're just overriding it in the view layer. This kind of follows the existing pattern (in the `useEffect` above) which I think is also something we should clean up, but maybe as a separate follow up PR?

src/devtools/client/themes/variables.css
jonbell-lot23: Interesting, thanks for flagging! Don’t have an opinion at this state, should be ok

src/ui/components/NetworkMonitor/index.tsx
bvaughn: I think two places are currently doing the "selecting" here:
* The comment renders the secondary label using the `NetworkRequestPreview` which has a click-handler that selects the network monitor and shows the request details.
* The Network panel's internal `RequestTable` component has a row-click-handler that does the same.

For both of the above cases, I think...
1. Jumping to the request's timestamp is a good idea, but–
2. We shouldn't update the Redux state to select the request if it's not loaded.

bvaughn: I'm going to go ahead and pull in this scope as well. I'm not _100% confident_ about the approach though so I'm going to tag @markerikson on it specifically for the Redux stuff 🙇🏼

src/ui/actions/network.ts
```diff
@@ -120,10 +121,21 @@ export function fetchFrames(tsPoint: TimeStampedPoint): UIThunkAction {
...
+export function showRequestDetails(requestId: RequestId): UIThunkAction {
+  return (dispatch, getState) => {
+    const state = getState();
+    const request = state.network.requests.find(request => request.id === requestId);
+    const loadedRegions = getLoadedRegions(state);
+
+    // Don't select a request that's not within a loaded region.
+    if (!request || !loadedRegions || !isPointInRegions(loadedRegions.loaded, request.point)) {
+      return;
+    }
```
bvaughn, line 133: This filtering logic was previously done in a React component's effect. This seemed bad because (a) it's an extra, unnecessary render and (b) it could cause view inconsistencies and/or flickering if that component happened to load later.

markerikson (approved):

src/ui/actions/network.ts
jcmorrow: I could be wrong, but it seems like if we're going to be pulling in `getState()` anyways, this would make a lot of sense as a check in a reducer. So we would always dispatch the action, but sometimes it would have an effect and other times it would not.

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -309,7 +309,7 @@ const RequestDetails = ({ cx, request }: { cx: any; request: RequestSummary }) =
     }
   }, [activeTab, activeTabs]);
 
-  if (!(loadedRegions && request && getPointIsInLoadedRegion(loadedRegions, request.point.point))) {
+  if (!(loadedRegions && request && isPointInRegions(loadedRegions, request.point.point))) {
```
jcmorrow, line 312: Do we still need this code? It seems like we should not be able to trigger this anymore, or if we did, returning `null` might be ok?

jcmorrow (commented):
Couple of comments, but overall looks good from a code perspective. I'm a little nervous about the multiple meanings to dimming things, since it's not very clear what a slightly-dimmed vs. very-dimmed row means from a user perspective. At the very least it seems like a good place to make sure we've documented what the treatments mean well (pinging @ceceliacreates to help with that).

src/ui/actions/network.ts
bvaughn: > I could be wrong, but it seems like if we're going to be pulling in `getState()` anyways, this would make a lot of sense as a check in a reducer. So we would always dispatch the action, but sometimes it would have an effect and other times it would not.

This is an interesting point to consider, I think– when logic should be run– at the point when the user performs an action or during render?

I think it should be done at the point the user performs an action (aka in the "thunk"). That's the only time when the UI the user has just interacted with is known to match the user's intent. Given that React might delay before rendering, might render more than once, might render a higher priority thing before a lower priority one, etc– filtering in render could mean an unexpected result in some cases.

Waiting to do the filtering in a reducer is akin to waiting until render time (at least for React's `useReducer` and `useState` updater functions). That actually _might not be true for Redux_. I'm not very familiar with it so I don't know.

That's why I moved it to the thunk though. cc @markerikson for thought about this.

jcmorrow: Oh interesting, my mental model was that redux would process the action and update state as soon as the action was dispatched and then render happens in response to the redux state changing, but it sounds like you're suggesting it could be the other way around in some scenarios? Where react decides to render and that causes the state to update? I'm having trouble picturing how that would work but maybe I'm misunderstanding.

src/ui/components/NetworkMonitor/RequestDetails.tsx
bvaughn: I added logic to prevent selecting a resource that's already outside of the loaded regions, but I didn't add anything to deselect a previously selected resource when a region gets unloaded.

We should probably handle that too, but currently we'll retain the state so that if you reload the region– the item is "reselected" automatically. In the interim, (when you've expected the trim but not yet loaded) we'll render the unloaded message.

If you'd like we can chat about the UX here. I'm not trying to fix all related issues with this PR FWIW though, just one of the more egregious ones I noticed personally.

src/ui/actions/network.ts
markerikson: Dispatching an action is 100% synchronous by default.  When you call `dispatch(someAction)`, by the time it returns:

- The reducer was run
- The store has saved the new state value internally
- The store has notified all UI subscribers

The mini Redux store shown at https://redux.js.org/tutorials/fundamentals/part-4-store#inside-a-redux-store is a 100% valid running example, and in there the `dispatch` method is just:

```js
  function dispatch(action) {
    state = reducer(state, action)
    listeners.forEach(listener => listener())
  }
```

Next we get into the conceptual question of "how do I handle logic in Redux?"

It's entirely valid to dispatch an action that results in 0 state updates.  We encourage thinking of actions as ["events that happened in an app"](https://redux.js.org/style-guide/#model-actions-as-events-not-setters), which any reducer _may_ respond to if appropriate.  This means that it's possible for _no_ reducers to care about an action.

If that happens, the store subscribers ( `connect` and `useSelector`) will still be called, but they both have eager bailouts that skip doing any further checks if the root state hasn't changed.

On the flip side, it's also totally fine to have logic in a thunk or similar that decides _if_ an action should be dispatched or not based on certain conditions.

bvaughn: > but it sounds like you're suggesting it could be the other way around in some scenarios?

No. I don't know what guarantees Redux provides in this case (which is why I tagged Mark). My mental model around reducers is built on React's reducer behavior.

src/ui/components/NetworkMonitor/RequestDetails.tsx
jcmorrow: Goootchya. That makes sense, I have one further suggestion:

It sounds like now the *only* situation in which we would see this RequestUnavailable component is when the currently select request is a region that is not loaded. In that case, would we change that message from "this is unavailable" (which feels like it does not give the user a place to go any further into understanding what is going on) to "This request happened in a part of the recording which is not currently loaded"? That way it at least gives them something more specific to look for information about like "Why did part of my recording unload?"

bvaughn: > This request happened in a part of the recording which is not currently loaded

Yeah, that sounds like a better message. I'll make that change 👍🏼

src/ui/actions/network.ts
jcmorrow: thanks both, it's good to know that there is some difference between the redux and react models (because I have always assumed they are super close to each other), and your explanation lines up with what I expect @markerikson, thanks for the thorough explanation! I'm fine with this living in either place. My own preference would be to stick it in the reducer, but my reasons are small and kind of nitpicky, so I'm fine either way.

bvaughn:
> I'm a little nervous about the multiple meanings to dimming things, since it's not very clear what a slightly-dimmed vs. very-dimmed row means from a user perspective.

I agree. I think we can follow up with @jonbell-lot23 on a better long-term fix for this (in terms of UI). I think it's worth fixing the current even-more-broken behavior in the meanwhile though.

PR #6537, opened 2022-04-22
Only request at most one column's location per line

jcmorrow:
We might actually want to check whether the user has columnBreakpoints
turned on, because theoretically someday we could show different numbers
hovering on different parts of the line, but for right now, we always
show the first location, so let's just ask for that hitCount.

A nice consequence of this is that it is a lot less likely that we will
have to fallback to analyses because we are asking for too many
locations, because the backend's limit is 1,000 locations, we query in
1,000 line chunks, and so if no line asks for more than one location we
should always be under the limit. (This is not actually necessarily true
because generated locations also have a limit but whatever, close
enough).

Recording of heatmaps working on a big minified file which it was choking on before this change:
https://app.replay.io/recording/gethitcounts-on-big-minified-ebay-file--6dfc9d63-55ca-4e76-ad58-b9ed72c0dbf4

### DIFF
--- a/src/devtools/client/debugger/src/client/commands.js
+++ b/src/devtools/client/debugger/src/client/commands.js
@@ -339,12 +339,13 @@ async function getSourceActorBreakpointHitCounts({ id }, lineNumber, onFailure) 
   // explanation of the bounds here.
   const lowerBound = Math.floor(lineNumber / MAX_LINE_HITS_TO_FETCH) * MAX_LINE_HITS_TO_FETCH;
   const upperBound = lowerBound + MAX_LINE_HITS_TO_FETCH;
-  const locationsToFetch = locations.filter(
-    location => location.line >= lowerBound && location.line < upperBound
-  );
+  const locationsToFetch = locations
+    .filter(location => location.line >= lowerBound && location.line < upperBound)
+    .map(location => ({ ...location, columns: [location.columns.sort((a, b) => a - b)[0]] }));
+
   return {
-    min: lowerBound,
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/client/commands.js
```diff
@@ -339,12 +339,13 @@ async function getSourceActorBreakpointHitCounts({ id }, lineNumber, onFailure)
   // explanation of the bounds here.
   const lowerBound = Math.floor(lineNumber / MAX_LINE_HITS_TO_FETCH) * MAX_LINE_HITS_TO_FETCH;
   const upperBound = lowerBound + MAX_LINE_HITS_TO_FETCH;
-  const locationsToFetch = locations.filter(
-    location => location.line >= lowerBound && location.line < upperBound
-  );
+  const locationsToFetch = locations
+    .filter(location => location.line >= lowerBound && location.line < upperBound)
+    .map(location => ({ ...location, columns: [location.columns.sort((a, b) => a - b)[0]] }));
```
bvaughn, line 344: We'll always have at least one item in the `columns` array? It's an `any` type so I'm not sure.

bvaughn (changes requested):

bvaughn (approved):

src/devtools/client/debugger/src/client/commands.js
jcmorrow: Yeah, the backend won't return the line as an entry if there is not at least one breakable column.

PR #6539, opened 2022-04-23
Create command errors inside the command promise chain for better stacks.

loganfsmyth:
By creating the error inside the promise chain instead creating the error inside the message callback, the engine can be way better about stack traces.

### DIFF
--- a/src/protocol/socket.ts
+++ b/src/protocol/socket.ts
@@ -31,10 +31,25 @@ let gSocketOpen = false;
 let gPendingMessages: Message[] = [];
 let gNextMessageId = 1;
 
+type CommandResponse =
+  | {
+      id: number;
+      result: unknown;
+      error?: undefined;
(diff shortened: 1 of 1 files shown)


markerikson:
_Just_ out of curiosity, can you show a before/after of the stack trace results?

loganfsmyth:
Totally! Here's the results of running `app.sendMessage("Pause.fakeThing", {})` in the browser console during local dev:

## Currently:
Screen Shot 2022-04-22 at 6 32 50 PM

## With this change:
Screen Shot 2022-04-22 at 6 31 58 PM

loganfsmyth:
As you can see, instead of including all the context that lead up to the `onmessage` handler firing, it now includes all the context that lead to `sendMessage` running, which is _much_ more useful.

markerikson (approved):

PR #6542, opened 2022-04-23
Clean up breakpoint edit footer

jasonLaster:
fixes https://github.com/RecordReplay/design/issues/23


jonbell-lot23 (approved):
Yeah, I think I follow this. Mostly I'm confused that the code was doing the inline conditionals in the previous example, but the new example is quite clear.

* Show this if editing
* Then here's the other stuff

I don't quite follow the conditional that disables it. I see the ifEmpty but I don't understand how it's set.

jasonLaster:
@jonbell-lot23 yep conditionals in JSX are hard for me to follow as well. 

re isEmpty: I recommend recording some examples of print statement panels and then looking for cases where `isEmpty` is true :)

PR #6543, opened 2022-04-23
Generate fixtures from e2e tests

jasonLaster:
This is a fun experiment for generating fixture data from E2E tests

https://www.notion.so/replayio/Automating-fixtures-c4c0da5fd7984057bf941330de0533ff

There are some really nice benefits of this approach

1. it's easy to create fixtures which can be used in unit tests
2. it's easy to re-generate fixtures when they're invalidated
3. it's easy to check for invalidation in a unit test in ci

### Checklist
- [ ] Decide on strategy for where we store fixture JSON data.
- [x] Make test runner check if fixture data changes between runs and warn locally or fail on CI.
- [x] Add `manifest.json` support for specifying which integration tests should generate fixture data.

### Follow up
- [ ] Decide on strategy for invalidating fixtures when protocol changes. (Perhaps encode the version into the fixture for comparison?)


bvaughn:
79c613a adds our first unit test built with one of the new fixtures. I didn't touch the `runTest` script but my test depends on your path changes for the fixtures :)

bvaughn:
I went ahead and updated the test fixture logic to add support for the arg `--updateWebsocketLogs` to instruct the runner to update fixtures. By default, it should do this when running on GitHub but not when running locally.

bvaughn:
> Typecheck / GraphQL schema types (pull_request) 

This CI job is a burden 😬

bvaughn:
Sorry for the poor form of force-pushing to your branch. I don't think you're around or actively working on this and I just didn't want to clutter the commit log with my silliness :)

bvaughn:
I'm really excited about this direction. IMO the biggest downside is that these fixtures look pretty big. I wonder how folks would feel about putting them into a sibling repo or a submodule so they don't bloat this repo?

The `breakpoints-01` test for example generates a fixture that's > 80k lines long.

bvaughn:
While generating fixture data for some of the other tests, I discovered that I can't even run some of these locally (like `console_warp-01.js`) because of the error:
```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at Object.slice (node:buffer:593:37)
    at Buffer.toString (node:buffer:811:14)
    at Object.readFileSync (node:fs:497:41)
    at readRecordingFile (/Users/bvaughn/Documents/git/replay/devtools/test/node_modules/@replayio/replay/src/main.js:34:13)
    at readRecordings (/Users/bvaughn/Documents/git/replay/devtools/test/node_modules/@replayio/replay/src/main.js:65:17)
    at Object.listAllRecordings (/Users/bvaughn/Documents/git/replay/devtools/test/node_modules/@replayio/replay/src/main.js:222:22)
    at upload (/Users/bvaughn/Documents/git/replay/devtools/test/src/getExample.js:75:20)
    at recordExample (/Users/bvaughn/Documents/git/replay/devtools/test/src/getExample.js:108:23)
    at async getExample (/Users/bvaughn/Documents/git/replay/devtools/test/src/getExample.js:130:30)
    at async runMatchingTests (/Users/bvaughn/Documents/git/replay/devtools/test/run.js:52:34) {
  code: 'ERR_STRING_TOO_LONG'
```

At first I thought maybe we'd broken something, but it's that way on master as well 🤪

test/src/runTest.js
```diff
@@ -34,6 +50,10 @@ async function recordBrowser(state, test, testPath, browserName) {
     await browser.close();
   }
 
+  if (updateWebsocketLogs) {
```
jasonLaster, line 53: it'd also be nice to have a field in `manifest.json` for specifying if we should save a fixture

jasonLaster: We should compare the new and old fixture and warn locally and fail in CI.

src/devtools/client/webconsole/components/__tests__/WebConsoleAppSearch.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    // Initialize state using exported websocket messages,
+    // sent through the mock environment straight to socket parsing
+    fixtureData.forEach((message: any) => {
+      window.mockEnvironment?.sendSocketMessage(JSON.stringify(message));
+    });
+
+    // Give everything time to settle
+    await sleep(100);
+  });
```
jasonLaster, line 61: it would be nice to extract this sooner rather than later to a util which takes the name of the fixture.

jasonLaster (commented):
Hoping to be able to tweak this later tonight :)

test/src/runTest.js
bvaughn: > We should compare the new and old fixture and warn locally and fail in CI.

Originally I thought we'd just run tests and only complain if they failed but I like this suggestion better.

src/devtools/client/webconsole/components/__tests__/WebConsoleAppSearch.test.tsx
bvaughn: Agreed.

bvaughn: Done!

src/devtools/client/webconsole/components/Search/ConsoleSearch.tsx
```diff
@@ -48,18 +48,32 @@ export default function ConsoleSearch({}: Props) {
...
+      if (results.length === 1) {
+        summaryMsg = "1 result";
+      } else {
+        summaryMsg = `${index + 1} of ${results.length} results`;
+      }
+    } else {
+      summaryMsg = "No results found";
+    }
+  }
+
```
bvaughn, line 63: Going to extract this change into its own PR (#6544) since it's only tangentially test related.

src/test/testUtils.tsx
```diff
@@ -60,7 +63,46 @@ export const filterLoggingInTests = (
...
+
+  // Give everything time to settle
+  await new Promise(resolve => setTimeout(resolve, 100));
+
+  return store;
+}
+
 export const filterCommonTestWarnings = () => {
+  // Skip LoadedRegions debug messages
+  filterLoggingInTests(message => message === "LoadedRegions", "debug");
```
bvaughn, line 105: These debug logs are kind of annoying so I added this (for now). Maybe we should always filter out _all_ `console.debug` logs from test output.

bvaughn:
I added a checklist to the PR description that gather the comments above into a single place listing what's left before this draft is done. Thoughts?

jasonLaster:
###  Add manifest.json support for specifying which integration tests should generate fixture data.

Added a `saveFixture` field which will keep us from writing unnecessary fixtures to disk.

### Make test runner check if fixture data changes between runs and warn locally or fail on CI.

Added a naive string equality check and quickly realized that the messages were very different. (pause ids, session ids, timings, ...). We might want to encode the protocol version into the fixture, but at that point it seems like this can be something we can do when we have more data points.

bvaughn:
> We might want to encode the protocol version into the fixture, but at that point it seems like this can be something we can do when we have more data points.

Fair!

bvaughn:
Okay. Jason and I chatted out of band and decided to land this as-is with a single test fixture, then we'll follow up with a cleanup (that Jason will write an issue for) that removes the fixture and instead saves a Replay recording id in Git. Then we can extract the "fixture" messages _from the Replay recording itself_ (and save it in a Git-ignored .cache directory for offline support).

jasonLaster:
Here's a quick [gist](https://gist.github.com/jasonLaster/8f1ff65479c642c87300459491dce8b8) of what a replay ID based fixture approach would look like. We can reference this when we follow up

PR #6550, opened 2022-04-25
Add testFixtureUtils with types from @recordreplay/protocol

bvaughn:
Update useConsoleSearch.test to use the new fixture approach.

Related to #6543.

### DIFF
--- a/src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
+++ b/src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
@@ -1,17 +1,26 @@
+import { ThreadFront, ValueFront } from "protocol/thread";
 import React from "react";
 import { act } from "react-dom/test-utils";
+import {
+  DEFAULT_SESSION_ID,
+  createConsoleMessage,
+  createLoadedRegions,
+  createSource,
+  createValue,
+  sendLoadedRegionsToMockEnvironment,
+  sendMessageToMockEnvironment,
+  sendSourceToMockEnvironment,
+  createPointDescription,
+} from "test/testFixtureUtils";
 import {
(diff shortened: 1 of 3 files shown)


src/test/testFixtureUtils.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+export function sendLoadedRegionsToMockEnvironment(value: loadedRegions): void {
+  window.mockEnvironment?.sendSocketMessage(
+    JSON.stringify({
+      method: "Session.loadedRegions",
+      params: value,
+      sessionId: DEFAULT_SESSION_ID,
+    })
+  );
+}
```
bvaughn, line 185: These messages aren't typed. (There's no mapping of `method` string to expected `params` value that I'm aware of). That's unfortunate.

src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
```diff
@@ -35,107 +45,56 @@ describe("useConsoleSearch", () => {
...
+        argumentValues: [createValue({ value: "another string value" })],
+      })
+    );
+
+    sendLoadedRegionsToMockEnvironment(
+      createLoadedRegions({
+        beginTime: 0,
+        endTime: 1000,
+      })
+    );
```
bvaughn, line 90: This is still pretty verbose, although I tried to use as many optional params with intelligent defaults as I could.

jasonLaster (approved):

bvaughn:
```js
sendMessagesToMockEnvironment(
  createSource({ kind: "scriptSource" }),
  createConsoleMessage({ text: "Plain string message" })
)
```
Love this as a follow up suggestion!

Issue #6538, opened 2022-04-22
Unable to change default workspace to "Your Library" on upload screen


loganfsmyth:
<img width="753" alt="Screen Shot 2022-04-22 at 3 46 18 PM" src="https://user-images.githubusercontent.com/132260/164815002-b05f249d-b193-4d5d-8c82-567b337e1690.png">

We've regressed this before, can we please fix this UI so IDs and UI strings never overlap?

loganfsmyth:
I assume the issue here is that https://github.com/RecordReplay/devtools/blob/3ccd585328fbeca847ad0ff7a1d356820daf572c/src/ui/components/UploadScreen/Sharing.tsx#L34 should be using the `MY_LIBRARY` constant, which means this regressed in https://github.com/RecordReplay/devtools/pull/5304 which also means it's been broken for 2 months.

We definitely also need to figure out why this error isn't being reported to Sentry

ryanjduffy:
Replay that confirms @loganfsmyth 's theory and points out why this didn't hit sentry.

https://app.replay.io/recording/upload-screen-error-when-selecting-my-library--ce147d99-217c-47e4-908f-5c06c7d0e5fa

The sentry change was intentional but worth debating so i'll open a pair of PRs so we can land a fix for this bug separately from the sentry discussion.

PR #6553, opened 2022-04-25
Improve apollo error handling

ryanjduffy:
Reverts RecordReplay/devtools#5669

## Issue

We missed #6538 because there was no notification that something was going wrong

## Resolution

* Revert suppressing all 500 errors to force us to handle server-side errors on the client.
* Improve the console log output to provide more context to sentry

image

### DIFF
--- a/src/ui/utils/apolloClient.tsx
+++ b/src/ui/utils/apolloClient.tsx
@@ -126,7 +126,9 @@ export const createApolloClient = memoizeLast(function (
  * Apollo's error reporting is inconsistent and GraphQL errors may appear as
  * network errors containing the GraphQL error in an undocumented property.
  */
-export function extractGraphQLError(error: ApolloError | undefined): string | undefined {
+export function extractGraphQLError(
+  error: Partial<Pick<ApolloError, "graphQLErrors" | "networkError" | "message">> | undefined
+): string | undefined {
   if (!error) {
     return undefined;
   }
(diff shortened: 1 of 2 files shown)


src/ui/utils/apolloClient.tsx
```diff
@@ -181,9 +183,15 @@ function createHttpLink(token: string | undefined) {
...
-  return onError(({ graphQLErrors, networkError }) => {
+  return onError(err => {
+    const { graphQLErrors, networkError } = err;
     if (graphQLErrors) {
-      console.error("[Apollo GraphQL error]", graphQLErrors);
+      console.error(
+        "[Apollo GraphQL error]",
+        err.operation.operationName,
+        extractGraphQLError(err),
+        graphQLErrors.length > 1 ? graphQLErrors : graphQLErrors[0]
```
jcmorrow, line 193: I'm a little confused by this. So we log

```
[error]
```
if there is only one, or
```
error
```
if there is more than one? Since this is just a `console.error` call I'm sure it's fine either way, but just wondering if I'm missing something meaningful here.

src/ui/utils/telemetry.ts
```diff
@@ -14,7 +14,6 @@ export function setupTelemetry() {
     "Current thread has changed",
     "Failed to load Stripe.js",
     "Stripe.js not available",
-    "Received status code 500",
```
jcmorrow, line 17: So are we still going to send all Apollo 500 errors to sentry with the same subject line? I was picturing us stopping these here, but then doing something like:

```
Sentry.captureException(new Error(graphQLErrors.join(": "));
```

in the Apollo error link.

src/ui/utils/apolloClient.tsx
ryanjduffy: There's usually only one error so if that's the case I'm pulling that object out so sentry doesn't try to be too smart and collapse the output to "Array" because it's a deep object. But, if there are multiple errors I'm not trying to be too fancy and just logging them all.

src/ui/utils/telemetry.ts
ryanjduffy: I considered this but didn't want to intro customized error handling. So we still `console.error` these errors but with a more descriptive log and sentry will still capture them as before.

src/ui/utils/apolloClient.tsx
jcmorrow: Gotchya, I still wish we had something more specific at the top-level other than `[Apollo GraphQL error]`, just for Sentry's grouping reasons, but this seems better than silencing everything.

jcmorrow (approved):

Issue #6555, opened 2022-04-26
Ensure ValueFronts are fully loaded before showing them


hbenl:
https://github.com/RecordReplay/devtools/blob/d187caadf0932c485eebb0316a8ba7355b3130a9/src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx#L79-L83

TODO:
- fix the console
- add telemetry so that we see if the same issue exists in other places

PR #6556, opened 2022-04-26
Ensure ValueFronts are fully loaded before showing them (#6555)

hbenl:


src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
```diff
@@ -76,15 +79,14 @@ class OI extends PureComponent<ObjectInspectorProps> {
     this.ensureRootsLoaded();
   }
 
-  // TODO [hbenl] this should not be necessary: the ObjectInspector should only
+  // this should not be necessary: the ObjectInspector should only
   // be rendered with a root ValueItem that has already been loaded.
-  // However currently (at least) the console doesn't ensure that this is the case,
-  // so we need this workaround for now.
```
bvaughn, line 82: Curious why we're removing the TODO part of this comment but leaving the "_this should not be necessary_..." bit?

Is the implied todo here that we'll remove it later so long as we don't see "devtools-oi-root-not-loaded" events being logged? (Maybe we can spell that out more clearly for someone like me, if so?)

src/protocol/thread/thread.ts
```diff
@@ -925,7 +927,7 @@ class _ThreadFront {
   }
 
   async findConsoleMessages(
-    onConsoleMessage: (pause: Pause, message: Message) => void,
+    onConsoleMessage: (pause: Pause, message: WiredMessage) => void,
```
bvaughn, line 930: Why are we still passing `pause: Pause` as the first param?

The only place we use `ThreadFront.findConsoleMessages` – we add a wrapper function that ignores the first param.

bvaughn (approved):

src/devtools/packages/devtools-reps/object-inspector/components/ObjectInspector.tsx
hbenl: The TODO was intended for fixing the known console issue and adding telemetry. So this entire workaround should not be necessary anymore but I thought we could keep it to make the ObjectInspector resilient in case it is used incorrectly again.

src/protocol/thread/thread.ts
hbenl: I don't know, I just didn't want to start refactoring stuff in this PR.

bvaughn:
I see this error pretty often in the local app:
Screen Shot 2022-05-23 at 8 35 36 AM

Seems related to this PR? I'd provide a Replay but auth is not working for me at the moment.

cc @hbenl

Issue #6483, opened 2022-04-20
Our most failed command is DOM.repaintGraphics on Node recordings


gideonred:
This Honeycomb query is showing how many DOM.repaintGraphics are failing for recordings it doesn't apply to:
https://ui.honeycomb.io/replay/datasets/backend/result/j9EjMZc9a4/a/nP91qUWR22f

Could we make the frontend not do these unnecessary protocol commands?

PR #6559, opened 2022-04-26
Don't try to repaint in node recordings (#6483)

hbenl:


src/protocol/graphics.ts
```diff
@@ -286,6 +286,10 @@ export function setupGraphics(store: UIStore) {
 }
 
 export async function repaint(force = false) {
+  const recordingTarget = await ThreadFront.recordingTargetWaiter.promise;
+  if (recordingTarget === "node") {
```
jasonLaster, line 290: Small nit, but would prefer if the waiters are private and expose a cleaner interface...

```js
if (await ThreadFront.getTarget() == "node") {
  return;
}
```

PR #6563, opened 2022-04-26
Add login button for Replay browser-initiated auth flow

ryanjduffy:


ryanjduffy:
Found out that ext auth has never worked for linux so this will have to be pushed out until https://github.com/RecordReplay/gecko-dev/issues/832 is resolved

ryanjduffy:
Added an macOS check so we can get this out and come back to linux support.

PR #6567, opened 2022-04-26
More comment unit tests

bvaughn:
- [x] Refactor test factory functions to be a little less verbose on the test side.

Before:
```js
sendSourceToMockEnvironment(
  createSource({
    kind: "scriptSource",
  })
);
sendMessageToMockEnvironment(
  createConsoleMessage({
    text: "Plain string message",
  })
);
// And so on...
```

After:
```js
sendValuesToMockEnvironment(
  createSource({
    kind: "scriptSource",
  }),
  createConsoleMessage({
    text: "Plain string message",
  }),
  // And so on...
);
```

- [x] Add TypeScript types to GraphQL mocks.
- [x] Fixed some bad mock data (missing `indexed` array in loaded regions mock).

- [ ] ~~Add comment sorting tests to replace ones removed in #6335.~~ Punted for now due to effort.

### DIFF
--- a/src/test/testFixtureUtils.tsx
+++ b/src/test/testFixtureUtils.tsx
@@ -26,9 +26,12 @@ import { UIStore } from "ui/actions";
 import { createTestStore } from "./testUtils";
 
 export const DEFAULT_SOURCE_ID = "fake-source-id";
+export const DEFAULT_SOURCE_URL = "fake-source-url";
 export const DEFAULT_SESSION_ID = "fake-session-id";
 
 let uidCounter = 0;
+
+type MessageTuple = ["Console.newMessage", Message];
 
 export function createConsoleMessage({
   argumentValues = [],
@@ -56,8 +59,8 @@ export function createConsoleMessage({
   stack?: CallStack;
   text?: string;
   url?: string;
-}): Message {
-  return {
+}): MessageTuple {
+  const value: Message = {
     argumentValues,
     column,
     data,
@@ -71,7 +74,11 @@ export function createConsoleMessage({
     text,
     url,
   };
+
+  return ["Console.newMessage", value];
 }
+
+type LoadedRegionsTuple = ["Session.loadedRegions", loadedRegions];
 
 export function createLoadedRegions({
   beginPoint,
@@ -87,17 +94,19 @@ export function createLoadedRegions({
   endTime: number;
   isIndexed?: boolean;
   isLoaded?: boolean;
-}): loadedRegions {
+}): LoadedRegionsTuple {
   const timeRange: TimeStampedPointRange = {
     begin: createTimeStampedPoint({ point: beginPoint, time: beginTime }),
     end: createTimeStampedPoint({ point: endPoint, time: endTime }),
   };
 
-  return {
+  const value: loadedRegions = {
     indexed: isIndexed ? [timeRange] : [],
     loaded: isLoaded ? [timeRange] : [],
     loading: [timeRange],
   };
+
+  return ["Session.loadedRegions", value];
 }
 
 export function createLocation({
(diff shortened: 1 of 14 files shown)


src/test/testFixtureUtils.tsx
```diff
@@ -215,34 +227,43 @@ export async function loadFixtureData(testName: string): Promise<UIStore> {
...
+            params: value,
+            sessionId: DEFAULT_SESSION_ID,
+          })
+        );
+        break;
+      case "Console.newMessage":
+        window.mockEnvironment?.sendSocketMessage(
+          JSON.stringify({
+            method,
+            params: {
```
markerikson, line 248: :question: I notice a difference in `params, sessionId` between these different messages. Is that correct?  (If it is, it bothers my sense of symmetry :)  )

src/test/testFixtureUtils.tsx
```diff
@@ -215,34 +227,43 @@ export async function loadFixtureData(testName: string): Promise<UIStore> {
...
-      params: value,
-      sessionId: DEFAULT_SESSION_ID,
-    })
-  );
+export function sendValuesToMockEnvironment(...values: Array<Tuple>): void {
+  values.forEach(([method, value]) => {
+    switch (method) {
+      case "Session.loadedRegions":
+        window.mockEnvironment?.sendSocketMessage(
+          JSON.stringify({
```
markerikson, line 237: :speech_balloon: This is all test util code and not really a big deal to keep DRY. At the same time, I'm looking at this and feeling like it could be turned into either a lookup table, or something that does formatting first and then a single call to `sendSocketMessage()` at the end?

src/test/testFixtureUtils.tsx
bvaughn: Nope, you're right. `sessionId` should be outside of `params`. Doesn't seem to affect anything though but I'll fix.

src/test/testFixtureUtils.tsx
bvaughn: This is a draft PR and that's a nit pick 😉

bvaughn: A more serious answer: I had hoped we'd be able to eventually type the values we're passing to `sendSocketMessage` message too (and going too DRY like you're suggesting might interfere with that). That being said, I'm not sure if it's possible (if we even have that typing information available) so....

markerikson (approved):

jcmorrow (approved):
I have not looked at these Apollo mocks in a long time, but I think this all looks good to me 💯

bvaughn:
527d685 is interesting.

Mock tests are failing on CI. Maybe something about my changes caused them to fail? Not sure. So I started digging in but I was surprised to see that I could not really debug them locally.

For example, I could run:
```sh
node test/mock/run.js --pattern invalid-recording-01
```

On master this would pass. On my branch it would fail. In both cases all it's really doing is launching an instrumented browser at the following URL: http://localhost:8080/recording/40408917-b1b5-49fc-bbf1-511655f647f1?mock=1

To debug this, I assumed the easiest thing would be to open that URL myself and inspect it. Funny thing though– that URL was just a blank white page.

Fast forward a bit until Josh and Ryan explained to me that I can view the logs using the `@replayio/replay` CLI and I spotted some errors in the output:
> TypeError: indexed is undefined

Digging in a little further, I noticed that the "loaded regions" object only had `loading` and `loaded` keys. Was this a problem with our types:
https://github.com/RecordReplay/devtools/blob/dc19be26c6fd0a0537d18af167649678471e5e4c/src/ui/state/app.ts#L74-L78

Or with our fixtures? I guessed it was a problem with the latter, so I manually added empty `indexed` arrays to two of our fixtures that I noticed were missing them.

After this– the mock test still fails on this branch, but now the URL (above) is no longer an empty white page. This `undefined` error was causing the entire app UI to crash and be blank. Yet somehow our mock tests still passed. (Not sure how or why that is?)

cc @hbenl in case this last part is surprising.

bvaughn:
The mock test was failing due to changes made to `test/mock/src/graphql/recordings.ts`. I've fixed that test in fee443a.

PR #6573, opened 2022-04-27
Update tab headers to mimic VS Code overflow behavior

### LINKED ISSUE #6504: Tabs overflow shouldn't be ellipsis-ized
When tab file names overflow in the editor, we currently handle the overflow with ellipsis. We should just let it get clipped so that we're not wasting valuable real estate. This is in line with what VS Code does.

Us:

VS Code:
image

bvaughn:
Resolves #6504.

~~https://www.loom.com/share/8f93bc1048b047888619a1ccc3e77f9d~~

https://www.loom.com/share/fcecf251c8ec4be5ba5f57d3e48d8f2d

### DIFF
--- a/src/ui/components/SecondaryToolbox/SecondaryToolbox.css
+++ b/src/ui/components/SecondaryToolbox/SecondaryToolbox.css
@@ -100,3 +100,17 @@ overflows its container (https://drafts.csswg.org/css-flexbox-1/#min-size-auto) 
 .secondary-toolbox .splitter {
   background-clip: padding-box;
 }
+
+.secondary-toolbox-right-buttons-container {
+  position: relative;
+  overflow: visible;
+}
+
+.secondary-toolbox-scroll-overflow-gradient {
+  position: absolute;
+  top: 0;
+  left: -2rem;
+  width: 2rem;
+  height: 100%;
+  background-image: linear-gradient(to left, rgba(0, 0, 0, 1), rgba(0, 0, 0, 0));
+}
--- a/src/ui/components/SecondaryToolbox/SecondaryToolbox.css
+++ b/src/ui/components/SecondaryToolbox/SecondaryToolbox.css
@@ -17,7 +17,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  overflow: hidden;
+  overflow: auto;
 }
 
 .secondary-toolbox-header button {
@@ -26,6 +26,7 @@
   cursor: pointer;
   transition: color 200ms;
   display: flex;
+  flex: 0 0 auto;
   flex-direction: row;
   align-items: flex-end;
   overflow: hidden;
@@ -34,16 +35,6 @@
 
 .secondary-toolbox-header .panel-buttons button {
   padding: 8px 12px;
-}
-
-.secondary-toolbox-header .panel-buttons button {
-  padding: 8px 12px;
-}
-
-.secondary-toolbox-header button .label {
-  overflow-x: hidden;
-  white-space: nowrap;
-  text-overflow: ellipsis;
 }
 
 .secondary-toolbox-header button > *:not(:last-child) {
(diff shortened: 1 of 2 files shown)


src/ui/components/SecondaryToolbox/SecondaryToolbox.css
```diff
@@ -36,16 +37,6 @@
   padding: 8px 12px;
 }
 
-.secondary-toolbox-header .panel-buttons button {
-  padding: 8px 12px;
-}
```
bvaughn, line 41: GitHub isn't doing a good job of showing this, but this rule was a duplicate:
https://github.com/RecordReplay/devtools/blob/78628cf4485a20e22ce4d30fc79240a295aa98fa/src/ui/components/SecondaryToolbox/SecondaryToolbox.css#L35-L41

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
```diff
@@ -17,7 +17,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  overflow: hidden;
+  overflow: auto;
```
bvaughn, line 20: Horizontally scroll if we need to.

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
```diff
@@ -26,6 +26,7 @@
   cursor: pointer;
   transition: color 200ms;
   display: flex;
+  flex: 0 0 auto;
```
bvaughn, line 29: Buttons should fit their content (without growing or shrinking).

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
```diff
@@ -36,16 +37,6 @@
...
 
-.secondary-toolbox-header .panel-buttons button {
-  padding: 8px 12px;
-}
-
-.secondary-toolbox-header button .label {
-  overflow-x: hidden;
-  white-space: nowrap;
-  text-overflow: ellipsis;
-}
```
bvaughn, line 47: Never clip the button label text.

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
jcmorrow: Have you checked this on the other SecondaryPanels? Sometimes those things have an annoying amount of interdependency on the same CSS rules.

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
jcmorrow: I never know if basis should be auto or content. Is there a rule of thumb where you use content over auto?

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
jcmorrow: I'll leave this comment here because it's good context but I just clicked through on the preview branch and didn't see anything obvious 👍

jcmorrow (approved):

markerikson (approved):

src/ui/components/SecondaryToolbox/SecondaryToolbox.css
bvaughn: TBH I'm unclear on that as well. I've always used "auto".

bvaughn:
> Sometimes those things have an annoying amount of interdependency on the same CSS rules.

Sounds like (and looks like) our CSS could maybe use some cleanup at some point too huh?

bvaughn:
Based on Jon's most recent comment in chat, I have added the gradient overhang like I proposed.

I'm not hugely happy with the CSS class names and stuff but 🤷🏼

jasonLaster (approved):
❤️

PR #6580, opened 2022-04-28
Prefix badge

jasonLaster:
We should probably re-write most of this code, it's real rough.

https://www.loom.com/share/68ee0d067a234b26976556ca5e33bbf6


jasonLaster:
Landing this despite the code needing some work for three reasons

1. it is safe behind a feature flag
2. the styling is easier to improve once it lands
3. reactathon is tuesday and this is great for demos

PR #6593, opened 2022-04-29
Change console search bgcolor so it doesn't blend in with the console

jonbell-lot23:
Was:
image

image

Now:
image

image

This extra band of colour isn't there until you trigger it, so I think this is a good balance between showing contrast and being too extreme


jasonLaster (approved):
I like it! 

Wonder if the editor footer should get a similar treatment

jonbell-lot23:
Closes https://github.com/RecordReplay/devtools/issues/6515

PR #6596, opened 2022-04-29
Use the content hash when grouping identical sources (#6594)

### LINKED ISSUE #6594: Use the content hash when grouping identical sources
Currently we consider sources to be identical if they have the same URL, but that may not always be the case, see RecordReplay/customer-support#69 and RecordReplay/devtools#3429.
RecordReplay/backend#5462 adds a content hash to `Debugger.newSource` events, this should be used in addition to the URL when grouping identical sources.

hbenl:
Needs RecordReplay/backend#5462 to be deployed first.
Fixes #6594.

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -62,6 +62,7 @@ export interface Source {
   kind: SourceKind;
   url?: string;
   generatedSourceIds?: SourceId[];
+  contentHash?: string;
 }
 
 export interface PauseEventArgs {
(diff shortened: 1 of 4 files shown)


src/protocol/thread/thread.ts
```diff
@@ -1302,6 +1309,30 @@ class _ThreadFront {
...
+        assert(
+          source.generatedSourceIds?.length === 1,
+          "a pretty-printed source should have exactly one generated source"
+        );
+        const minifiedSource = this.sources.get(source.generatedSourceIds[0]);
+        assert(minifiedSource?.contentHash, "no contentHash found for minified source");
+        hash = "minified:" + minifiedSource?.contentHash;
+      } else {
+        assert(hash, "no contentHash found for source that is not pretty-printed");
+      }
```
bhackett1024, line 1330: These asserts feel dangerous.  If they fail, will we get problems like https://github.com/RecordReplay/devtools/issues/6547 ?  I think it would be better if we warned on getting unexpected source info from the backend, and still made a best-effort attempt to do the right thing.

bhackett1024 (approved):

PR #6598, opened 2022-04-29
Integration test runner captures Apollo fixtures for unit tests

bvaughn:
### ⚠️ Commits are atomic and may be easier to review individually.

The overall purpose of this PR is to build on top of the work Jason and I started in #6543 and add support for Apollo fixture data as well, so that unit tests have a more complete picture of a realistic session.

Fixture data can be updated by running:
```sh
node test/run.js --update-fixtures --pattern <test-name>
```

This will generate two fixture files:
```
test/fixtures/console_messages.apollo.json
test/fixtures/console_messages.replay.json
```

Unit tests can then use these fixtures to run more realistic test scenarios like so:
```js
import { loadFixtureData } from "test/testFixtureUtils";

await loadFixtureData("console_messages");
```

### DIFF
--- a/src/test/testFixtureUtils.tsx
+++ b/src/test/testFixtureUtils.tsx
@@ -213,9 +213,7 @@ export async function loadFixtureData(
       const mock = {
         request: {
           operationName: requestData.operationName,
-          query: gql`
-            ${requestData.query}
-          `,
+          query: gql(requestData.query),
           variables: requestData.variables,
         },
         result: responseData,


src/ui/components/__tests__/AppModal.test.tsx
```diff
@@ -1,57 +1,56 @@
...
+    recording = response.recording;
+    store = response.store;
+
+    const url = getRecordingURL(recording);
+    global.location.pathname = url;
+    useRouter.mockImplementation(() => ({
+      asPath: url,
+      pathname: url,
+      query: { id: recording.id },
+    }));
```
bvaughn, line 31: This block is probably something we should move to test utils at some point 😄  A few tests are doing this.

src/ui/components/UploadScreen/Privacy.tsx
```diff
@@ -83,7 +83,7 @@ function PrivacyData({ icon, name, urls }: { icon: string; name: string; urls: s
 
 export function Privacy() {
   const { recording } = useGetRecording(getRecordingId());
-  const { operations } = recording!;
+  const { operations } = recording ?? {};
```
bvaughn, line 86: I think there's no guarantee that `recording` won't be undefined (still `loading`), in which case this component would fall down.

jasonLaster (approved):
This looks nice from the perspective that the unit tests are not any more complicated and perhaps a bit simpler and we unlock the ability to test components that depend on some of the apollo queries

bvaughn:
> and we unlock the ability to test components that depend on some of the apollo queries

Yes! This is huge 😁

PR #6601, opened 2022-04-29
[DRAFT] Initial Redux DevTools integration POC

markerikson:
This PR:

- Adds a dependency on the `@redux-devtools/app` package, which has the prebuilt Redux DevTools UI components and logic
- Adds a new "Redux" panel to the toolbox pane
- Renders the RDT UI
- Actually manages to feed in actual Redux actions received via annotations from the backend
- Actually manages to use timeline updates to reset the list of actions and feed in just the ones that were dispatched up to that point in time
- Uses [`next-global-css`](https://github.com/bem/next-global-css) to work around Next throwing errors due to the RDT `<Editor>` component importing CSS from within `node_modules`

## Status

Totally a draft POC and absolutely _not_ ready to ship in any way :)

Things that this needs for minimum productionization:

- Consideration of how we actually want to handle annotations between React and Redux integration
- The Redux panel should likely only import the RDT UI if we actually _have_ Redux-related annotations.  (Possibly the same for the React panel later?).  There's a bunch of bundle size here, and we shouldn't pay that if it's not relevant.  
- For that matter, showing the "Redux" panel should also be an experimental setting
- The current "tie displayed actions to the timeline progress" technique is hacky. But it does seem to work!
- The RDT UI theme should be controlled by Replay's own theme selection
- The RDT logic has a bunch of "feature" settings. We should probably turn off a bunch of options, at least for now.  (Like, "skip" doesn't make sense atm.)
- Probably need to correctly handle "instances" - there can be many "stores" connected to the DevTools in one page
- Should test this against other frameworks that use the RDT, like Zustand, Jotai, and NgRx
- Should test this with some really large or frequent actions.  (Like, say.... Replay itself! :)  )

This is also entirely dependent on the FF content script changes sitting over in https://github.com/RecordReplay/gecko-dev/pull/831 - you've got to use an FF build with those to make the recording, which also means you can't even try this PR out because our prod backend doesn't recognize this FF build!  We may need to just go merge in those changes and release a new FF build first to unblock the chicken/egg issue here.


src/ui/actions/reactDevTools.ts
```diff
@@ -57,8 +20,6 @@ export function addAnnotations(annotations: Annotation[]): AddAnnotationsAction
   return { type: "add_annotations", annotations };
 }
 
-export function setProtocolCheckFailed(): UIThunkAction {
-  return async dispatch => {
-    dispatch({ type: "set_protocol_fail" });
-  };
+export function setProtocolCheckFailed(): SetProtocolCheckFailedAction {
```
markerikson, line 23: :speech_balloon: This action creator really didn't need to be a thunk anyway, since it's just dispatching a single basic action.

markerikson:
Status update:

- We've published a Firefox build that _should_ capture Redux data. Unfortunately, it doesn't seem to be doing that :(  We're going to have to investigate further.
- I've updated this PR to do a few of the "productionization" items:
  - The "Redux" tab only shows up if there are Redux annotations loaded into the client
  - The `@redux-devtools/app` package is lazy-imported on demand when you switch over to that tab for the first time
- I've run this against a Jotai CodeSandbox, and confirmed that A) this _does_ work with other non-Redux packages, and B) it _does_ work okay with multiple "instances" in one page:

First instance - a single Jotai "count" atom:

image


Second instance - a "combined atoms" view:

image

markerikson:
I've updated this to keep the RDT theme in sync with the Replay app theme.  Also tried passing down some flags to mark all of the RDT "features" as disabled, in the hopes that its UI would turn off buttons like "Jump" and "Skip", but doesn't seem to be doing what I want.

That said, I _think_ this is basically working at a v1 level at this point.

Caveats:

- I have not tried it with a larger-scale recording with many actions, ie, a recording of Replay itself
- We don't yet know why the latest browser build does not appear to be capturing Redux actions as annotations properly, and this PR is kinda useless until we get that figured out

jasonLaster (approved):
🚀

PR #6602, opened 2022-04-29
Define "ready" as "indexed and loaded"

jcmorrow:
We can get into a state where things are indexed, but not loaded, and that results in some bad UI where things no longer look like they are loading, but not all of the data is present either. This is more of a bandaid for https://github.com/RecordReplay/backend/issues/5429 than a real solution, but at least it communicates to the user when we are still waiting on things and they shouldn't expect everything to work or be accurate

### DIFF
--- a/src/devtools/client/debugger/src/utils/sources-tree/utils.test.js
+++ b/src/devtools/client/debugger/src/utils/sources-tree/utils.test.js
@@ -5,15 +5,7 @@
 
 import { makeMockSource } from "devtools/client/debugger/src/utils/test-mockup";
 
-import {
-  createDirectoryNode,
-  getRelativePath,
-  isExactUrlMatch,
-  isDirectory,
-  isNotJavaScript,
-  getPathWithoutThread,
-} from "./utils";
-import { addToTree } from "./addToTree";
+import { getRelativePath, isExactUrlMatch, isNotJavaScript, getPathWithoutThread } from "./utils";
 
 describe("sources tree", () => {
   describe("isExactUrlMatch", () => {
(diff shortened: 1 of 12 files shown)


jasonLaster (approved):
Okay - so I like the thinking. We should be talking to the user in terms of the intersection of indexed and loading, but I'm not sure I love the implementation...

src/ui/components/OgreTimeline.tsx
```diff
@@ -34,7 +34,7 @@ const Spans = ({
   const { endTime } = useSelector(getZoomRegion)!;
```
jasonLaster, line 34: Lets rename this component `ProtocolTimeline` since we'll show all the data we have from the protocol

src/ui/state/app.ts
```diff
@@ -75,6 +75,7 @@ export interface LoadedRegions {
   loading: TimeStampedPointRange[];
   loaded: TimeStampedPointRange[];
   indexed: TimeStampedPointRange[];
+  indexedAndLoaded: TimeStampedPointRange[];
```
jasonLaster, line 78: Lets leave a TODO here that we want to move (`loading`,`indexed`, `loaded`) into a protocol field and rename `indexedAndLoaded` to `loading` and use a more relevant `loaded`

jasonLaster: can `indexedAndLoaded` be a selector?

jcmorrow: I'm not so sure on moving `indexedAndLoaded` to `loading`, I think those two are closer to `loaded`, and the protocol definition of `loading` is probably fine? (Maybe when we add soft-focus `loading` would become `focusedAndProtocolLoading` or something?

jcmorrow: oh yeah, sorry, `indexedAndLoaded` **is** a selector now, and I just forgot to update the usage. Good catch 👍

jaril (approved):
lgtm

PR #6629, opened 2022-05-02
Handle messages without a frame in ConsoleOutput

hbenl:
http://localhost:8080/recording/replay-breaks--b879b88e-9908-4b08-b4c7-04a2e79971b1


hbenl:
@jasonLaster A good example why Typescript warnings shouldn't be ignored ;)

bvaughn (approved):

Issue #6634, opened 2022-05-02
Sort comments


hbenl:
The 2 comments on [this recording](https://app.replay.io/recording/step-over-fail--4fc68349-5f89-49fb-b086-99f47c69abc1) are shown in the wrong order: the comment that is shown first is the second in the recording's timeline and is also the second I entered.
In [this recording](https://app.replay.io/recording/view-long-replay--1afe2dc1-646f-4403-941c-5f68c5e8e11a) the last 2 comments were entered last, but they're at the beginning of the recording's timeline, so I would have expected them to be at the top.

PR #6651, opened 2022-05-04
Fix comment sorting (#6634)

hbenl:


src/ui/components/Transcript/Transcript.tsx
```diff
@@ -24,7 +24,7 @@ export default function Transcript() {
       clonedComments.push(pendingComment.comment);
     }
 
-    const sortedComments = sortBy(clonedComments, ["point", "createdAt"]);
+    const sortedComments = sortBy(clonedComments, [c => BigInt(c.point), "createdAt"]);
```
bvaughn, line 27: I'm surprised this is necessary. I thought JavaScript's normal string sorting would properly handle numeric string values in cases like this.

Taking the points from one of the Replays you mentioned, it looks like this is just a difference in behavior between Lodash `sortBy` and native JavaScript `sort`: https://codesandbox.io/s/boring-sound-lic91z?file=/src/index.js

TIL.

bvaughn (approved):

src/ui/components/Transcript/Transcript.tsx
hbenl: Strings are always sorted lexicographically, in the codesandbox you use `(a, b) => a.point - b.point` as the sort function, which will coerce the strings to numbers, `sortBy` will not do that type coercion.

PR #6661, opened 2022-05-06
Cache annotations in thread front for use by different callers

bhackett1024:
While working on https://github.com/RecordReplay/devtools/pull/6660 I noticed a problem where if multiple places have asked for annotations from the thread front, we get errors because there can be only one listener at a time for the annotations in the recording.  This impacts the redux devtools as well I think (cc @markerikson ).  This PR pulls out the changes I made in that PR so that we only fetch the annotations once, avoid this problem, and avoid refetching data from the network.

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -8,6 +8,7 @@
 // helps with adapting the devtools to the WRP.
 
 import {
+  Annotation,
   BreakpointId,
   ExecutionPoint,
   FrameId,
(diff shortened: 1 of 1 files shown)


markerikson:
Interesting! Yeah, I specifically moved the one existing `ThreadFront.getAnnotations()` call to a new location in a `<ReduxAnnotationsProvider>` component, and it has filtering logic to send the React annotations one way and the Redux annotations another way - see [`cc8b3d8` (#6601)](https://github.com/RecordReplay/devtools/pull/6601/commits/cc8b3d8d124c29c656047361382024618da3c916) for the diff.

Longer term, it might be really helpful to be able to set up multiple annotations subscriptions and have the backend filter them.

jasonLaster (approved):

PR #6663, opened 2022-05-06
Handle hot analyses and analyses with errors

jcmorrow:
OK, this is kind of a few fixes rolled into one, so let me see if I can break them all down.

- Display error messages when analyses error
  - I think the situation we were trying to avoid here was when a user has used an analysis to correctly get the points for a given line, but then some further analysis errored and we overwrote the count which we had before. There are a few reasons we don't need to do this anymore:
    - We use getHitCounts by default now, so it's likely that we won't be overwriting an analysis result which was being used
    - We differentiate between TooManyPoints and other errors now, so we won't accidentally show 10K+ hits like we did in https://github.com/RecordReplay/customer-support/issues/18
  - I think there might be fancier things to do later around analyses and errors, but for now, this feels like a major simplification which will make room for additional complexity later as it makes sense
- Including `Loading` as a possible text length, and use `width` rather than `maxWidth`. Using `width` stops things from bouncing around when the user is playing/pausing, and including `Loading` in the calculation means that we won't get stuck rendering things as 3 characters wide because we expect things like `2/5` when in fact `Loading` is also a possible expression.
- Favor `Loading` over `Indexing` as a term we display to the user.
- Use the current timestamp to show what hit is closest to the current time, even for cases where the user cannot edit the print statement
- Show a breakpoint timeline and allow the user to move around it even when the print statement is not editable
- Reduce the number of points we will show on a compressed timeline. If there are 10,000 hits, we don't want to render 10,000 circles which mostly overlap with each other. Instead, we show at most 1 circle per half-second (so the max number is 120 * number of minutes in a recording, in most cases it will be a lot less of course) and also the current pause, if it lines up with any of the hits. Note: this is just for display, the arrows still go backwards and forwards just fine.

Fixes https://github.com/RecordReplay/devtools/issues/6420

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
@@ -121,7 +121,12 @@ function Panel({
               toggleEditingOn={toggleEditingOn}
             />
           )}
-          <BreakpointNavigation {...{ breakpoint, editing, showCondition, setShowCondition }} />
+          <BreakpointNavigation
+            breakpoint={breakpoint}
+            editing={editing}
+            showCondition={showCondition}
+            setShowCondition={setShowCondition}
+          />
         </div>
       </div>
     </Widget>
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
@@ -1,46 +1,56 @@
-import React, { useState, useRef } from "react";
+import React, { useState, useRef, useMemo } from "react";
 import classnames from "classnames";
 
 import { actions as UIActions } from "ui/actions";
 import { selectors } from "ui/reducers";
 import { timelineMarkerWidth as pointWidth } from "ui/constants";
 const BreakpointTimelinePoint = require("./BreakpointTimelinePoint").default;
-const { prefs } = require("ui/utils/prefs");
 import { getVisiblePosition } from "ui/utils/timeline";
 import PortalTooltip from "ui/components/shared/PortalTooltip";
-import { mostRecentPaintOrMouseEvent } from "protocol/graphics";
+import { mostRecentPaintOrMouseEvent, precacheScreenshots } from "protocol/graphics";
 
 import TimeTooltip from "devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/TimeTooltip";
 import { UIState } from "ui/state";
 import { connect, ConnectedProps, useSelector } from "react-redux";
-import { getExecutionPoint } from "devtools/client/debugger/src/reducers/pause";
 import { HoveredItem } from "ui/state/timeline";
 import { UnloadedRegions } from "ui/components/Timeline/UnloadedRegions";
 import { AnalysisPayload } from "ui/state/app";
+import { PointDescription } from "@recordreplay/protocol";
+import { getExecutionPoint } from "../../../selectors";
 
 function Points({
   analysisPoints,
-  hoveredItem,
   breakpoint,
+  hoveredItem,
 }: {
-  breakpoint: any;
   analysisPoints: AnalysisPayload;
+  breakpoint: any;
   hoveredItem: HoveredItem | null;
 }) {
   const executionPoint = useSelector(getExecutionPoint);
-  let displayedPoints = [...analysisPoints.data];
-
-  // Even if we're over maxHitsDisplayed, we should still display a single paused
-  // point if the user happens to be paused on that this breakpoint.
-  if (analysisPoints.data.length > prefs.maxHitsDisplayed && executionPoint) {
-    displayedPoints = displayedPoints.filter(
-      point => BigInt(point.point) == BigInt(executionPoint)
-    );
-  }
+  const duration = useSelector(selectors.getRecordingDuration);
+  const displayedPoints = useMemo(() => {
+    if (!duration) {
+      // We haven't loaded yet, bailing early
+      return [];
+    }
+    let previousDisplayed: PointDescription;
+    return analysisPoints.data.filter(p => {
+      if (
+        !previousDisplayed ||
+        executionPoint === p.point ||
+        p.time - previousDisplayed.time > duration * 0.01
+      ) {
+        previousDisplayed = p;
+        return true;
+      }
+      return false;
+    });
+  }, [analysisPoints.data, executionPoint, duration]);
 
   return (
     <>
-      {displayedPoints.map((p, i) => (
+      {displayedPoints?.map((p, i) => (
         <BreakpointTimelinePoint
           breakpoint={breakpoint}
           point={p}
@@ -55,10 +65,6 @@ function Points({
 
 type BreakpointTimelineProps = PropsFromRedux & {
   breakpoint: any;
-};
-type Coordinates = {
-  x: number;
-  y: number;
 };
 
 function BreakpointTimeline({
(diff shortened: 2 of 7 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
```diff
@@ -121,7 +121,9 @@ function Panel({
               toggleEditingOn={toggleEditingOn}
             />
           )}
-          <BreakpointNavigation {...{ breakpoint, editing, showCondition, setShowCondition }} />
+          <BreakpointNavigation
+            {...{ breakpoint, editing, isHot, showCondition, setShowCondition }}
+          />
```
bvaughn, line 126: Curious to see us using a spread operator here when we're also enumerating props, rather than just listing them the "normal" way :grin:
```js
<BreakpointNavigation
  breakpoint={breakpoint}
  editing={editing}
  isHot={isHot}
  showCondition={showCondition}
  setShowCondition={setShowCondition}
/>
```

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -6,21 +6,25 @@ import { getExecutionPoint } from "../../../selectors";
...
   analysisPoints: AnalysisPayload;
+  isHot: boolean;
   prefixBadge: string;
 }) {
   const executionPoint = useSelector(getExecutionPoint);
   const isIndexed = useSelector(getIsIndexed);
   const theme = useSelector(getTheme);
   let status = "";
-  let maxStatusLength = 0;
+  let maxStatusLength = "Loading".length;
```
bvaughn, line 20: "Indexing" is longer than "Loading" 😁  Seems like maybe we should just wait and measure the status length after we've set the status string? (Why are we pre-measuring the max? Now that we're using CSS `width` rather than `minWidth` anyway– maybe this can just be changed too?)

bvaughn (approved):
I don't fully understand some of these fixes (the surrounding code) but this looks fine 😁

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
jasonLaster: It would be nice to replace Indexing with Loading in the UI since we're transitioning to a simpler timeline as well

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
markerikson: It's a somewhat common pattern to avoid that annoying prop name repetition :)  Slightly surprised it sounds like you haven't seen it before?

bvaughn: I haven't seen it before. Seems kind of wasteful (unnecessary object allocation and spread operator) but to be fair, not in a way that _actually_ matters so 🤷🏼 😆

bvaughn: Also just to play devil's advocate against myself– there are _tons_ of objects and functions allocated for every React element rendered so I realize my comment above is silly 😄

jcmorrow: I left this here because it was already like this but I've never liked this short-hand, personally. I am sympathetic to cutting the repetition, but I just feel like it always looks awkward and adds more cognitive load than just reading the words twice (although it is less typing and less area for typos!). I know this is impossible but I *really* wish that React could interpret an attribute name without an explicit value to the local scope of that token instead of `true`. So:

```
<BreakpointNavigation
  breakpoint
  editing
  isHot
  showCondition
  setShowCondition />
```

There's probably lots of reasons why that is a bad idea but it *feels* right to me 🤷

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
jcmorrow: > Why are we pre-measuring the max?

My goal with pre-measuring the max was to avoid the length of the pill/timeline from jumping. I agree that it's awkward, but a better solution did not seem immediately obvious to me. Good call on the loading vs. indexed thing though. Re: changing the actual word in the UI... that's an interesting point! I wonder why we have distinguished between those in the past. I can't think of a compelling reason off of the top of my head right now.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.js
markerikson: So this is actually a common proposal for a notional "JSX spec 2.0" - the term is "prop punning".  The Rescript language actually does that by default in its JSX variation.

Dan even mentioned this today as a thing that would theoretically happen someday:

https://twitter.com/dan_abramov/status/1523343682476326912

but followed it with "no immediate changes planned, likely more of a “years from now” kind of thing".

Given all that, yes, `<MyComponent {...{prop1, prop2, prop3}} />` is about as close as we're gonna get for a long time.

jcmorrow: wow I'm surprised it's even being seriously considered, it seems like a pretty huge change! I guess that's what major versions of things are for though.

bvaughn: > I know this is impossible but I *really* wish that React could interpret an attribute name without an explicit value to the local scope of that token instead of `true`.

I suspect one reason it doesn't do this is that it would be a backwards breaking change that would be difficult to grep for. (All existing shorthand props would start evaluating as `undefined` instead of `true`.)

Though I think this could be code modded.

jcmorrow: oh yeah I forgot about typing. That seems like a good enough reason to me.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
```diff
@@ -28,19 +30,24 @@ function Points({
...
-      point => BigInt(point.point) == BigInt(executionPoint)
-    );
-  }
+  const displayedPoints = useMemo(() => {
+    let previousDisplayed: PointDescription;
+    return analysisPoints.data.filter(p => {
+      if (
+        !previousDisplayed ||
+        executionPoint === p.point ||
+        p.time - previousDisplayed.time > HALF_A_SECOND
```
jaril, line 39: Need to double check but isn't time here in ms, so `HALF_A_SECOND` should be 500?

jaril (approved):

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
jcmorrow: Yeah I was also confused about this, but after double-checking I think A) you are correct and B) we should probably have it be larger than 5ms but also smaller than 500. I think I'll play with it a little and decide what feels right.

jcmorrow: What I've realized should have been obvious - it depends on how long the recording is. I've made it 1% of the recording, so the dots are thick enough that they can run together, but bursty activity gets sampled down to max 100 events.

PR #6666, opened 2022-05-08
Build a new Redux DevTools UI on top of their store

jasonLaster:
This is a fun experiment to see what it would look like if we build a new UI on top of Redux DevTools' redux actions/reducers/selectors.

* I'm directly importing all of the modules. We can lazy load the panel at the top, but otherwise this will be come cumbersome.
* I'm trying to use all of the selectors/actions so that we can re-use Redux DevTools' business logic

NOTE: This would be much harder without being able to use replay to see what's going on


src/ui/components/SecondaryToolbox/redux-devtools/ActionsList.tsx
```diff
@@ -0,0 +1,23 @@
...
+    <div>
+      {actions.map(action => (
+        <div>{action.action.type}</div>
+      ))}
+    </div>
+  );
+}
+
+export default connect(state => {
+  const id = getActiveInstance(state.instances);
```
markerikson, line 18: :speech_balloon: you know this could be a `useSelector` instead :)

jasonLaster:
Couple of notes from a pairing session with @markerikson 

1. The current version in tree will be 0.5
2. We'll create a new `/react-devtools` route to host the official Redux DevTools and associated monitors
3. We'll create a new Redux DevTools panel which is based off of the instances reducer


### New Panel

1. We'll use a shared annotations reducer for raw annotations
2. We'll create a new redux-devtools reducer and potentially some action creators
3. The Redux DevTools panel will be lazily loaded to avoid bundle size bloat
4. The panel will have a list view which will show the current paused red line
5. The panel will have a summary view which will show one of four modes (payload, state, diff, stack)

jaril:
@jasonLaster this looks stale — closing it for now, feel free to open if that's a bad call!

PR #6670, opened 2022-05-09
Fix white text on bright badges UX bug

jonbell-lot23:
White text doesn't show up on yellow very well. I went to fix that, which is best handled with CSS rather than embedding a bunch of conditionals inline. I moved things over to css modules to make sure we have the flexibility we need.

Quick walkthrough: https://www.loom.com/share/29851b422acd4f2498c5ced1469e9cb2


src/ui/components/PrefixBadge.module.css
```diff
@@ -0,0 +1,55 @@
...
+}
+
+:root:global(.theme-light) .yellow {
+  background-color: #e4cd00;
+  color: rgb(99, 79, 5);
+}
+
+:root:global(.theme-dark) .yellow {
+  background-color: #fee608;
+  color: rgb(99, 79, 5);
```
jcmorrow, line 54: Would it be ok if we converted these to hex just to make things consistent?

jcmorrow (commented):
There's quite a few unused things, and a little clean up to do I think

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -1,9 +1,10 @@
 import sortedLastIndex from "lodash/sortedLastIndex";
 import { useSelector } from "react-redux";
-import { getBadgeColor, isColorPrefix } from "ui/components/PrefixBadge";
+import { isColorPrefix } from "ui/components/PrefixBadge";
```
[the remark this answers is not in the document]
jonbell-lot23: Yup, keeping

jonbell-lot23:
Ideally I'd have been able to fix the code as @jcmorrow described, but I did what I could and made a separate issue for the stuff I couldn't figure out -> https://github.com/RecordReplay/devtools/issues/6693

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -40,15 +41,10 @@ export function PanelStatus({
...
-  const style = isColorPrefix(prefixBadge)
-    ? { backgroundColor: getBadgeColor(prefixBadge, theme), color: "white" }
-    : {};
-
   return (
     <div className="breakpoint-navigation-status-container">
       <div
-        className="rounded-2xl bg-breakpointStatusBG px-3 py-0.5 text-breakpointStatus"
-        style={style}
+        className={`rounded-2xl bg-breakpointStatusBG px-3 py-0.5 text-breakpointStatus ${styles[prefixBadge]}`}
```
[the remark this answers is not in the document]
jcmorrow: It's a little gross but I think it will work?

jonbell-lot23:
Hey @jcmorrow this is almost there. I can't fix my last bug because Replay is being slow.

Take a look! Bask in my brilliance! Then maybe we can figure out the last issue, where the default state isn't loading correctly. (For example if I dismiss the color, a gray color persists in the console. I think I know why but can't test easily with Replay being slow)

jcmorrow:
> But I can't fix my last bug because Replay is being slow.

If possible, when you see a horrendously slow Replay, run `sessionId` in the JS console to grab the session ID and either throw it in discord or in [this issue](https://github.com/RecordReplay/backend/issues/5517) so we have more information for performance work!

src/ui/components/PrefixBadge.module.css
```diff
@@ -0,0 +1,88 @@
...
+
+:root:global(.theme-light) .pickerunicorn,
+:root:global(.theme-dark) .pickerunicorn {
+  background-image: url(~devtools/client/debugger/images/sources/unicorn-pixel.svg) !important;
+  background-size: auto 100%;
+  border-radius: 0;
+  background-color: transparent;
+}
+
+:root:global(.theme-light) .empty,
```
jcmorrow, line 39: Yeah, so we are saying here that the color "empty" is `theme-base-90`, so that is what we are getting both in the picker and in the console messages. As I mentioned with the unicorn thing: the picker, the console, and the panel status are all pretty different in terms of how these different prefix choices interact with them, so it would probably be *best* to have separate components for each, rather than trying to make them all work with exactly the same styles. However, if you wanted to stick with this approach, I would suggest making `empty` equivalent to a `transparent` background color rather than `theme-base-90`. That means the picker will *actually* be empty (meaning the circle will be white and thus invisible), but that is sort of the trade-off we have to make if we insist that each place these choices show-up has to use the same styles.

jonbell-lot23:
Heyyyy @jcmorrow I think this works now. Thoughts?

https://www.loom.com/share/f698001410dc4324a573d5de6c319c45

src/ui/components/PrefixBadge.jsx
```diff
@@ -135,32 +90,16 @@ function PrefixBadgeButton({ breakpoint, theme, setBreakpointPrefixBadge }) {
     return null;
   }
 
-  console.log({ showPrefixBadge });
-
   const prefixBadge = breakpoint.options.prefixBadge;
   const isColor = isColorPrefix(prefixBadge);
```
jcmorrow, line 94: Pretty sure we can delete lines 93 and 94, and then we can delete `isColorPrefix`... and then we can delete `PREFIX_COLORS`

src/ui/components/PrefixBadge.module.css
jcmorrow: I still have this feedback 😄

jcmorrow:
Rebased and merged as #6741

Issue #6673, opened 2022-05-09
Fix circular imports in client


markerikson:
Part of #6552 .

Our client codebase has a ton of circular imports ( `a` imports `b` imports (possibly indirectly through a dozen other files) `a`).  This is likely responsible for some of the "side effectful imports" problems we're seeing.

Untangling those imports is a good first step for cleaning things up.

I've previously added the Webpack `circular-dependencies-plugin` to `next.config.js`, which errors and prints circle details if detected during build.  That's behind an environment variable, so it has to be turned on explicitly for now.

If I can get _all_ of these sorted out, I may turn it on permanently to avoid regressions.

PR #6675, opened 2022-05-09
Circular import fixes, part 1

markerikson:
This PR does some initial code refactoring to try to fix some of our circular imports:

- Imports of ` { client } from "./socket"` are moved to use `await import("./socket")` inside the functions where that's used
- The "set error" thunks were extracted from `actions/session` and moved into their own file
- The "async layout initial state" calculation logic was moved to `ui/setup/index`, where it's actually used
- `<ApolloWrapper>` was moved out to its own file
- `addTab()` was moved over to the file where it's used

These appear to resolve the first few circular deps errors I'm seeing, but there's many more where those came from :(

See #6673 for some examples of what the circle errors look like when logged by the Webpack plugin.

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -14,7 +14,7 @@ import LogRocket from "ui/utils/logrocket";
 import { registerRecording, sendTelemetryEvent, trackEvent } from "ui/utils/telemetry";
 import { extractGraphQLError } from "ui/utils/apolloClient";
 
-import { ExpectedError, UnexpectedError } from "ui/state/app";
+import type { ExpectedError, UnexpectedError } from "ui/state/app";
 import { getRecording } from "ui/hooks/recordings";
 import { getRecordingId } from "ui/utils/recording";
 import { getUserId, getUserInfo } from "ui/hooks/users";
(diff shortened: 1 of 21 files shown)


src/ui/actions/session.ts
```diff
@@ -14,7 +14,7 @@ import LogRocket from "ui/utils/logrocket";
 import { registerRecording, sendTelemetryEvent, trackEvent } from "ui/utils/telemetry";
 import { extractGraphQLError } from "ui/utils/apolloClient";
 
-import { ExpectedError, UnexpectedError } from "ui/state/app";
+import type { ExpectedError, UnexpectedError } from "ui/state/app";
```
jaril, line 17: Didn't know about this!

src/ui/setup/index.ts
```diff
@@ -27,6 +34,62 @@ declare global {
...
+  }
+
+  if (!recording) {
+    return syncInitialLayoutState.selectedPrimaryPanel;
+  }
+
+  return recording.comments.length ? "comments" : syncInitialLayoutState.selectedPrimaryPanel;
+};
+
+export async function getInitialLayoutState(): Promise<LayoutState> {
```
jaril, line 49: This is definitely gnarly — if you've got ideas on how we could simplify it 👋

jaril (approved):

src/ui/actions/session.ts
markerikson: Yeah, TS will let you mix them, like `import {realVariable, SomeType} from "wherever"`, but you can also clearly specify "this line is _only_ type imports" too. Probably are some arguments why that's mostly better, although it can also lead to multiple lines to import types and values separately from one file.

PR #6677, opened 2022-05-09
Circular import fixes, part 3: breakpoint actions structure

markerikson:
More circular import fixes, per #6673 .

This PR:

- Moves all the functions out of `actions/breakpoints/index.js` into `actions/breakpoints/breakpoints.ts`, as the index file was both doing mass re-exports _and_ defining its own functions.  That was causing a bunch of circles.
- Converts those moved functions to TS, albeit with some looseness around the various uses of `Location`
- Fixes any cases of `import { func } from "."`, which was causing some circles
- Renames `index.js` to TS to help with inference


jaril (approved):

bvaughn:
My console is full of these messages now:
Screen Shot 2022-06-01 at 3 56 59 PM

For example, if you open this recording you should see a ton of them:
https://app.replay.io/recording/tooltip-should-be-hidden--611c1587-916b-4ab3-a264-3b2b81b54b32?point=44134523303659330792475110726959171&time=23101.2016985138&hasFrames=true

Related to this PR?

PR #6682, opened 2022-05-10
Make a better unicorn

jonbell-lot23:
The original unicorn was looking pretty squished. (Notice in the palette you can't even see the horn) By making it a headshot, we can make it look more like a unicorn, more personalised, and have it fit the pixel art look better.

Was:
image

Now:
image

Figma zoomed-in comparison:
image


jaril:
> The original unicorn was looking pretty squished. (Notice in the palette you can't even see the horn) By making it a headshot, we can make it look more like a unicorn, more personalised, and have it fit the pixel art look better.
> 
> Was: image
> 
> Now: image
> 
> Figma zoomed-in comparison: image

Just noticed that these are both such bangers. Stardew Valley should hire us to design sprites.

PR #6690, opened 2022-05-10
Use the payload when setting awaitingSourcemaps

jcmorrow:
I think this was just lost in the massive diff of the cleanup in
https://github.com/RecordReplay/devtools/commit/02d24f59a7f2551c47b8f3384fa70da6f4b448dd

Fixes https://github.com/RecordReplay/devtools/issues/6687


markerikson (approved):
Yep, my bad, sorry!

PR #6697, opened 2022-05-11
Copy events before passing them to redux

jcmorrow:
I guess redux is freezing the payload of this action after the changes in https://github.com/RecordReplay/devtools/commit/02d24f59a7f2551c47b8f3384fa70da6f4b448dd, and that causes a bunch of errors to be raised when we try to extend the array by pushing more stuff onto it.

CleanShot 2022-05-10 at 18 31 30

Replay of the issue: https://app.replay.io/recording/mouse-event-errors-in-console--cedc8b70-6186-41c4-8a05-9d13b00d03ba

I *think* this will fix https://github.com/RecordReplay/devtools/issues/6679

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -164,7 +164,7 @@ function onMouseEvents(events: MouseEvent[], store: UIStore) {
     }
   });
 
-  store.dispatch(setEventsForType({ events: gMouseClickEvents, eventType: "mousedown" }));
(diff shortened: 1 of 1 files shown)


markerikson:
Ahhh, great point! Yes, the reducer is putting the array in the store, which means it's getting frozen.

markerikson (approved):

PR #6699, opened 2022-05-11
Wait for basic processing before we run analyses

### LINKED ISSUE #6549: "No hits" shown incorrectly on print statement
This seems to happen when adding a print statement before indexing finishes.  Replay with comments:

https://app.replay.io/recording/invalid-no-hits--8c51609c-f563-4f53-8857-dba1005ee64a?point=546164725844869027945326080604256823&time=232105&hasFrames=false

jcmorrow:
There's a common problem in our analyses which is that they sometimes run before the recording is fully loaded. #6549 is one example of this problem (replay at https://app.replay.io/recording/invalid-no-hits--8c51609c-f563-4f53-8857-dba1005ee64a?point=546164725844869027945326080604256823&time=232105&hasFrames=false)

There's another set of problems around *re-running* analyses, and I have a plan for that, but this is just an easy win to start with.

Fixes #6549

### DIFF
--- a/src/protocol/logpoint.ts
+++ b/src/protocol/logpoint.ts
@@ -226,7 +226,7 @@ export async function setLogpoint(
   condition: string,
   showInConsole: boolean = true
 ) {
-  await ThreadFront.ensureAllSources();
+  await Promise.all([ThreadFront.ensureAllSources(), ThreadFront.ensureProcessed("basic")]);
   const sourceIds = ThreadFront.getCorrespondingSourceIds(location.sourceId);
   const { line, column } = location;
   const locations = sourceIds.map(sourceId => ({ sourceId, line, column }));


src/protocol/logpoint.ts
```diff
@@ -226,7 +226,7 @@ export async function setLogpoint(
   condition: string,
   showInConsole: boolean = true
 ) {
-  await ThreadFront.ensureAllSources();
+  await Promise.all([ThreadFront.ensureAllSources(), ThreadFront.ensureProcessed("basic")]);
```
jcmorrow, line 229: I'm *pretty* sure that just waiting for basic processing works here, but I could be wrong?

jasonLaster: not opposed, but isn't the backend equipped to handle analysis before basic processing is finished?

hbenl: Is this the right place to wait? There are other places that start analyses (e.g. `setEventLogpoint()`), so I think we should wait in `AnalysisManager.runAnalysis()` - see also #6701

bhackett1024:
I don't think this patch will meaningfully change anything.  We're already waiting for all the sources, which requires basic data, so waiting for the basic processing to finish as well will (a) finish at about the same time, and (b) not actually be required for any of the later computation that is happening.  Nothing ever really needs to wait for all basic processing to finish vs. waiting for the specific data it needs, and we should be moving away from waits for basic processing.

It's hard to tell if waiting for indexed processing to complete would affect anything.  I added some more comments in the replay, but the analysis we ran to find the points finished before indexing finished in the loading range, which seems strange and I just filed https://github.com/RecordReplay/backend/issues/5554 to investigate this.  Either way though this doesn't seem like the real problem here.  As you said above we need to re-run analyses after the loaded regions change --- in this case we ran an analysis at line 5508, and focus mode was used to change the loaded regions, but no new analysis ran at that line later and we showed incorrect information in the UI after indexing in the new loaded regions finished.

bhackett1024 (changes requested):
See comment above

src/protocol/logpoint.ts
jcmorrow: The problem is that it's not a lack of sources which is the issue here. The sources were loaded before the "wrong"  analysis was run in the case of https://app.replay.io/recording/invalid-no-hits--8c51609c-f563-4f53-8857-dba1005ee64a?point=546164725844869027945326080604256823&time=232105&hasFrames=false. But it's true that if we just fix the re-running, then this kind of error will be a lot less problematic, since the analysis will re-run when everything is loaded.

PR #6702, opened 2022-05-11
Increase timeouts for e2e tests

hbenl:
This PR increases 2 timeouts for our e2e tests:
- CI test runs still fail quite often waiting for the Vercel preview branch
- I noticed that [this test run](https://github.com/RecordReplay/devtools/actions/runs/2303277954) had a failed test but didn't upload a recording for it - the reason was that the test ran into the jest timeout and not a timeout within the test

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,8 +52,8 @@ jobs:
         id: waitFor200
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          max_timeout: 480
-          check_interval: 60
+          max_timeout: 960
+          check_interval: 10
           environment: ${{ fromJSON('["", "production"]')[github.ref == 'refs/heads/master'] }}
     outputs:
       url: ${{ steps.waitFor200.outputs.url }}
--- a/jest-e2e.config.js
+++ b/jest-e2e.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   setupFilesAfterEnv: ["./test/e2e/setupEnv.ts"],
   testMatch: ["**/e2e/tests/*.test.ts"],
-  testTimeout: 240_000,
+  testTimeout: 480_000,
   transform: {
     // Use babel-jest to transpile tests with the next/babel preset
     // https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object


.github/workflows/test.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         id: waitFor200
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          max_timeout: 480
+          max_timeout: 960
```
jcmorrow, line 55: I'm pretty sure that most of these failures are not timeouts sadly 😞

jest-e2e.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   setupFilesAfterEnv: ["./test/e2e/setupEnv.ts"],
   testMatch: ["**/e2e/tests/*.test.ts"],
-  testTimeout: 240_000,
+  testTimeout: 480_000,
```
jcmorrow, line 4: The only question I have here is if there are some tests that will wait this whole time period before failing, because then the whole suite is more likely to time out?

.github/workflows/test.yml
hbenl: I'm not sure either if this will help, but I thought we should at least try?

jest-e2e.config.js
hbenl: Obviously this *can* happen, depending on how slow backend and CI runner are when the tests are run. If it does happen frequently we'll have to use more stripes.

.github/workflows/test.yml
hbenl: After some experimentation with the action I realised this: https://github.com/RecordReplay/devtools/issues/6730#issuecomment-1125171169. So I'm doubling down on this being a timeout issue.

jcmorrow: Oh, interesting, so that's saying that the check at `480` (for example) might actually be checking the result of the check at `420`? I'm ok with bumping this, but we should watch our for API rate-limiting errors, because that is what prompted me to bump the timeout so high, and if we have more builds pinging we might end up tripping that again.

jest-e2e.config.js
hbenl: I don't think that's necessary: when a test fails it does so due to the timeouts within the test - I've never seen a test "hang" anywhere that made the timeout for the entire test necessary (but I definitely want to keep that timeout just in case such a hang ever occurs).

.github/workflows/test.yml
jcmorrow: Wait whaaaaaaat? That's not very intuitive! Well, that's great, thanks for finding this, let's 🤞 !

jcmorrow (approved):

jest-e2e.config.js
jcmorrow: I definitely have seen this, but I think that was like 3 test runners ago 🙃

.github/workflows/test.yml
jcmorrow: One more thought on this: given what you've discovered, would it make sense to bump the timeout to some slightly odd number. Like would making it 961 be any better? It would be nice if the effect of this setting were as close as possible to the actual number we set it to 😆

hbenl: No, that wouldn't help: the action is always waiting for `check_interval`, so the timeout value is effectively rounded to a multiple of `check_interval`. But I have decreased `check_interval` now, so that should help.

jcmorrow: OK, just a heads up - I very much expect that we will be triggering the API rate limit with this change, but let's see what happens.

hbenl: So *that* was why it was changed from the default. I'll change it back since this is not really necessary (I just did it to make the tests start a *little* earlier)

jest-e2e.config.js
hbenl: Now I've seen this too, but it wasn't the test taking too long, for some reason the upload of the example recording stalled.

PR #6714, opened 2022-05-11
Add the received paint points to the ProtocolTimeline

jcmorrow:
What we show in the viewer is always a little tricky. See this note from `actions/timeline.ts`:

>  Seek to the exact time provided, even if it does not match up with a
>  paint event. This can cause some slight UI weirdness: resumes done in
>  the debugger will be relative to the point instead of the time,
>  so e.g. running forward could land at a point before the time itself.
>  This could be fixed but doesn't seem worth worrying about for now.

This ends up *really* lagging behind sometimes, if it's a long recording and the paints are still loading. Like, you can end up with execution and a paint from 10 seconds before where your timeline says you are.  This seems bad? I think maybe we should do more to signal to the user that a particular section of the recording is still loading and jumping there is not necessarily going to be successful either in terms of execution *or* in terms of video.

I walk through some potential solutions here: https://www.loom.com/share/fcdfddf52a43464dae6e864634692e94 but in the meantime, here's a little additional piece of timeline info: how far into the recording do we have paint points for on the `ProtocolTimeline` (it would be nice if we stored these in redux rather than this global variable, but I think it's probably ok to just recompute the paints each render because when they are loading is also when other stuff is going to be loading, so it will be rendering fairly often anyways.

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -106,7 +106,9 @@ export interface TimeStampedPointWithPaintHash extends TimeStampedPoint {
 // All paints that have occurred in the recording, in order. Include the
 // beginning point of the recording as well, which is not painted and has
 // a known point and time.
-const gPaintPoints: TimeStampedPointWithPaintHash[] = [{ point: "0", time: 0, paintHash: "" }];
+export const gPaintPoints: TimeStampedPointWithPaintHash[] = [
+  { point: "0", time: 0, paintHash: "" },
+];
 
 // All mouse events that have occurred in the recording, in order.
 const gMouseEvents: MouseEvent[] = [];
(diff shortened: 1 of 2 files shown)


jaril (approved):
The protocol timeline change lgtm!

As for what to do when the paints aren't available yet, 🤷‍♂️. I don't entirely hate the blurred idea, also. 

To be fair, regular video players (youtube?) also have that lag between you navigating to a new time, and the video buffering and getting the appropriate frame for that pause point. 

If I understood it right, the unique challenge here is that unlike a regular video player, paints just take noticeably longer. So when the paint's not ready, it's possible that the wait time is actually >10s? Which is a lot heavier than what I'd expect for video players, which is ~1-2s.

PR #6717, opened 2022-05-11
Fix setFullTextQuery name and re-export UI action creators

### LINKED ISSUE #6716: Can't type in search full text input
Search full text is broken for me. I tried it in a newer replay too (the one below is older) and had the same issue.

Replay: https://app.replay.io/recording/replay-cant-type-in-search-full-text-input--a207b7f1-b383-479a-8534-82d38ec9bcf6
Loom: https://www.loom.com/share/315ff868d92e4e449f8be2e58892fcc8

markerikson:
This PR:

- Fixes a typo in `setFullTextQuery`'s name (the 't' was lower-cased), which caused that action creator to not be referenced correctly
- Ensures that all action creators from `reducers/ui` are re-exported from `actions/ui`

also I hate this re-export pattern WHY DO WE DO THIS

Fixes #6716 , where typing text in the full text search input threw an error because the action creator didn't exist .

### DIFF
--- a/src/devtools/client/debugger/src/actions/ui.ts
+++ b/src/devtools/client/debugger/src/actions/ui.ts
@@ -46,13 +46,21 @@ import { selectSource, selectLocation } from "./sources/select";
 
 export {
   closeActiveSearch,
-  highlightLineRange,
   clearHighlightLineRange,
+  closeProjectSearch,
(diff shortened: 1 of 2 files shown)


markerikson:
Confirmed that full text search works again, merging

PR #6721, opened 2022-05-11
Focus Mode 1.1 "mouseleave" logic fix

bvaughn:
Fixes https://github.com/RecordReplay/customer-support/issues/89

### DIFF
--- a/src/ui/components/Timeline/Focuser.tsx
+++ b/src/ui/components/Timeline/Focuser.tsx
@@ -45,7 +45,7 @@ function Focuser({ editMode, setEditMode }: Props) {
     }
 
     // Stop dragging on "click"
-    const onWindowClick = (event: MouseEvent) => {
+    const onDocumentClick = (event: MouseEvent) => {
       switch (editMode.type) {
         case "drag":
         case "resize-end":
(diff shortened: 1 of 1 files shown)


src/ui/components/Timeline/Focuser.tsx
```diff
@@ -125,7 +125,17 @@ function Focuser({ editMode, setEditMode }: Props) {
     };
 
     // Stop all drag operations when the mouse leaves the window.
-    const onWindowMouseLeave = () => {
+    const onDocumentMouseLeave = (event: MouseEvent) => {
```
bvaughn, line 128: Looking at the Replay, "mouseleave" was being dispatched for things other than the outer window/document. This is not how I expected it to work (or how it worked for me on MacOS + Chrome during testing) but it seems like _at least one fix_ here is to bounds check before assuming the mouse has _actually_ left the document.

Testing on Chrome, Firefox, and Safari seems good for me after this change.

@markerikson Can you verify on Windows?

jcmorrow: Woah, weird but good thinking!

jcmorrow (commented):
I'll wait for Mark to 👍 on the windows fix but LGM

src/ui/components/Timeline/Focuser.tsx
markerikson: Seems smooth to me on both Chrome and FF!

markerikson (approved):

PR #6722, opened 2022-05-11
Circular import fixes, part 6: assorted imports and async reducer initial state

markerikson:

This PR:

- Moves a couple of the "async initial state" functions out of their reducer files and into `ui/setup/index.ts`, which is the only file that uses them.  This fixes some circles due to importing prefs functions into reducers.
- Copy-pastes a couple trivial selectors
- Reworks the `visibleColumnBreakpoints` file and associated code
  - There was a `selectedSource` param being passed through several functions, but turns out it was never actually being used. Removed that param from all related functions
- General rearrangement of several other imports

Thanks to the `circular-webpack-plugin` output changes (which I haven't committed yet), this PR gets us down to "only" 49 cycles, and we started with 174.  So, 72% fixed so far.


src/devtools/client/debugger/src/reducers/tabs.ts
```diff
@@ -31,13 +29,7 @@ export interface TabsState {
...
 }
 
-const EMPTY_TABS: Tab[] = [];
-
-export const getInitialTabsState = async () => {
-  const session = await getReplaySession(getRecordingId()!);
-
-  return { tabs: session?.tabs ?? EMPTY_TABS };
-};
+export const EMPTY_TABS: Tab[] = [];
```
jaril, line 32: Why not move this along with getInitialTabsState?

jaril: Wait I just saw it nevermind!

src/devtools/client/debugger/src/utils/editor/get-token-location.js
```diff
@@ -3,16 +3,3 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 //
```
jaril, line 5: Can we delete this file then?

jaril (approved):

src/devtools/client/debugger/src/utils/editor/get-token-location.js
markerikson: Oh yeah, I should have done that earlier! :)

PR #6725, opened 2022-05-12
Deploy new focus mode icon

jonbell-lot23:
Was:
image

Now:
image

Was:
image

Now:
image

Was:
image

Now:
image

--

Notice that the foreground colour used to be white, and now it's black. We can go back to white, I just had trouble getting it working.

Also note that a new Global Loader design is coming, so we don't need to get too fancy with this just yet. This is in prep for that.


bvaughn (approved):
The new icon is a much better fit for focus!

I'm surprised to see the foreground color changed from white to black on the focus button! I guess it is a much better contrast ratio though.

CI was failing because of LInt errors so I went ahead and fixed them.

jonbell-lot23:
"Notice that the foreground colour used to be white, and now it's black. We can go back to white, I just had trouble getting it working."

Cc @bvaughn

bvaughn:
Yeah, so this is an issue where our Tailwind and theme colors don't really work well. I think we need to create some new Tailwind styles for toggle on/of background and foreground colors. I'll do that this afternoon at some point.

bvaughn:
#6740

PR #6727, opened 2022-05-12
Disable graphql mutations in e2e tests

hbenl:
Recordings of failed e2e tests often show the next.js error popup triggered by a failed graphql mutation, making them hard to work with:

Screenshot from 2022-05-12 14-11-03

https://app.replay.io/recording/replay-of-localhost8080--c4044322-54f3-48c7-b140-b163e860fc50


jaril (approved):
LGTM!

We also perform mutations using `useMutation`, I wonder if we need to do it there as well?

hbenl:
> We also perform mutations using useMutation, I wonder if we need to do it there as well?

Haven't seen that get in the way yet. We'll see.

PR #6738, opened 2022-05-12
Only fetch annotations if we might use them

jcmorrow:
These redux annotations can be huge, if the feature is turned off we
should not be fetching them.


markerikson:
Trunk is yelling at you for the lack of `isReduxEnabled` in the deps array, but seems good otherwise

PR #6739, opened 2022-05-12
Circular import fixes, part 7: sources-tree and object-inspector

markerikson:

This PR:

- Shuffles functions and imports in `sources-tree`
- Completely changes the structure in `object-inspector/utils/`:
  - Renamed the folder to `object-inspector/items/`
  - Moved all of the `SomeItem` classes into `index.tsx`, because they all rely on each other
  - Moved all of the util functions into `utils.ts`

The total diff is accurate, but also bigger than it looks like.  This was primarily a "cut from `someItem.tsx`, paste into `index.tsx`" job.


jaril (approved):

jcmorrow (approved):
I played around with the ObjectInspector and SourcesTree on the preview branch, seems solid! 👍

PR #6740, opened 2022-05-12
Focus Mode 1.1 UI tweaks

bvaughn:
Loom: https://www.loom.com/share/352cdf641fe147fea22d013772208e80

- [x] Focus toggle button has distinct on/off states now
- [x] Discard button is always enabled

### DIFF
--- a/src/devtools/client/themes/variables.css
+++ b/src/devtools/client/themes/variables.css
@@ -197,6 +197,7 @@
   --theme-toggle-inner: var(--cool-gray-200);
   --theme-toggle-selected: white;
   --theme-toggle-bgcolor: var(--theme-base-85);
+  --theme-toggle-color: var(--grey-70);
   --theme-toolbar-hover: #fff;
   --toolbarbutton-focus-bgcolor: var(--theme-base-90);
   --toolbox-bgcolor: var(--theme-base-100);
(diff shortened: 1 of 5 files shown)


jcmorrow (approved):
Love this. On a side-note: during your loom I noticed that the width of the timestamps jumps when you toggle Focus Mode selection on and off and now I can't unsee it 👀

bvaughn:
> I noticed that the width of the timestamps jumps when you toggle Focus Mode selection on and off and now I can't unsee it 👀

Good catch! Let me look.

jonbell-lot23 (approved):
I noticed the timestamp jiggle also, so +1 to that.

Looking good! I wonder if the word "cancel" maybe fits better, but that's for me to muse on later, not something that would prevent this release

bvaughn:
Had "Cancel" at one point locally and decided to go with "Discard" instead because that felt more like what you were doing (_discarding_ the focus region) – but I don't feel strongly about it!

bvaughn:
As for the jiggled, I had noticed that at one point and meant to follow up on it but then I just...stopped seeing it 😅 

There's a 1 pixel width difference between the editable and non-editable state, despite my setting width and max-width, as well as border, margin, and padding to 0. I'll take another pass at it.

bvaughn:
Fixed the jiggle https://www.loom.com/share/8f6445a6de60482d9266f2c546ecb477

PR #6742, opened 2022-05-12
add AddCommentButton component

souporserious:
This adds a new `AddCommentButton` component under `packages/components` as well as a `shift+D` keyboard shortcut to trigger adding a comment.

This will be the first of many components of our design system that will be located here. Eventually, we may move these under the design repo, but this keeps things simple for now. I will have a follow-up RFC to this PR posted in the design repo to discuss architecture decisions for the larger design system.

## Future Considerations

Most of this component will eventually be extracted into specific design system components and I'd also like to revisit the animation once we have better tools in place since the composition isn't as nice as I'd like it be 😄

## Storybook

https://user-images.githubusercontent.com/2762082/168173685-f86124ee-87eb-4394-ab35-1f46f13d5df7.mov

## Editor

https://user-images.githubusercontent.com/2762082/168176445-f95cd195-7c14-4587-8caa-39a883f04806.mov


packages/components/AddCommentButton/AddCommentButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  </svg>
+);
+
+const StyledButton = styled.button<{ $isOpened: boolean; $isPausedOnHit: boolean }>(
```
souporserious, line 64: This was a quick spike using `styled-components` and hard-coding all of the values to start moving portions over to a CSS in JS solution, most of this will be ripped out in the near future which I'll detail more of in the upcoming RFC 😇

packages/components/AddCommentButton/AddCommentButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+const StyledText = styled.span({
+  whiteSpace: "nowrap",
+  transition: `all ${duration}ms 40ms ease-out`,
+  transform: "scale(0.98)",
+  opacity: 0,
+  margin: "0 0.25rem",
+});
+
+const TextFade = ({ children, visible }: { children: React.ReactNode; visible: boolean }) => (
+  <Transition in={visible} timeout={duration}>
```
souporserious, line 106: I'd eventually like to have some simple animation primitives like this in our system as well.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/PanelSummary/index.tsx
```diff
@@ -92,6 +91,28 @@ function PanelSummary({
     }
   };
 
+  /**
+   * Hacky solution that attaches a keyboard shortcut to "shift+c" for adding comments.
+   * TODO: move to a first-class keyboard shortcut hook that we can use to
+   * display a keyboard shortcut modal
+   *
+   */
+  const addCommentRef = useRef<any>(null);
```
souporserious, line 100: The ref here is to avoid having a huge dependency chain in `useEffect` below since `addComment` is not memoized. Happy to memoize and add the dependencies for `addComment` if we think this is an issue though!

jonbell-lot23 (approved):
Woo! What do you think about making the button into a full circle rather than a squirqle?

souporserious:
@jonbell-lot23 ah forgot to mention I had this, but noticed it was messing with other buttons like "Save". So you'll go from round to square in one flow. I can follow up when re-working these into the system, also happy to just go for it now if you want 🤠

jonbell-lot23:
> @jonbell-lot23 ah forgot to mention I had this, but noticed it was messing with other buttons like "Save". So you'll go from round to square in one flow. I can follow up when re-working these into the system, also happy to just go for it now if you want 🤠

Happy to come around to it later!

jaril (approved):

packages/components/AddCommentButton/AddCommentButton.tsx
bvaughn: This component is pretty hard for me to read 😅 I'm a little concerned that our new animation code is making things a lot hard (for me) to read (and by extension, to maintain/modify). Do you know how soon we'll be replacing it with something more long-term?

souporserious: Yeah, this was one of my first contributions here and wasn't sure what styling solution to use tbh 😅. We should start replacing these next month once the marketing videos and polish are in a good spot. Once Mark is done with the exploration around how we're going to manage the monorepo I'll get a docs site going and start fleshing out the design system tokens and primitive components.

bvaughn: Cool! Looking forward to talking about that stuff once you have proposals in mind 😄

PR #6744, opened 2022-05-12
Clamp to make CSS work

jcmorrow:
Turns out `Infinity%` does not play nicely with CSS widths.


src/ui/components/ProtocolTimeline.tsx
```diff
@@ -64,7 +65,7 @@ export default function ProtocolTimeline() {
             ? [
                 {
                   begin: { point: firstPaint.point, time: firstPaint.time },
-                  end: { point: "", time: Infinity },
```
jcmorrow, line 67: This is not accurate in this case, but it's not used either.

jaril (approved):

jaril:
Gotta love clamps.

PR #6745, opened 2022-05-12
Lower protocol timeline z-index not to cover hover timestamp

bvaughn:
https://www.loom.com/share/aa6cba64f828483ebe45d8f534c6a940


jcmorrow (approved):
Yeeeeee nice change. I always see this and then forget to actually change it.

bvaughn:
You and me both!

PR #6746, opened 2022-05-13
add initial Figma icons script

souporserious:
This adds an initial Node script for building icons from our [design system file](https://www.figma.com/file/ASas6u2DMihEEzw8jPT1XC/Replay-Component-Library?node-id=134%3A2898) and exports them to React components. In the future, this will take on more responsibility for illustrations among other sources which are detailed in the [RFC](https://github.com/RecordReplay/design/discussions/49).

## Future Considerations
This isn't extremely useful yet since the component's don't allow passing props. I think we should be able to easily fix this in a follow-up with [svgo](https://github.com/svg/svgo) or [svgr](https://react-svgr.com/).


jonbell-lot23 (approved):
I haven't actually read the code closely but I am excited that this is progressing!

bvaughn:
Before I dig into the code, I wanted to say that I am _thrilled_ to see us investing in automation. Decisions to automate build/release/sync tasks in the past have always been some of the most impactful in my experience.

packages/icons/README.md
```diff
@@ -0,0 +1,19 @@
...
+FIGMA_TOKEN=personal-token-here
+```
+
+Once that is setup, you can run the following script to download and generate the icons:
+
+```bash
+npm run build
+```
+
+This will generate an `index.js` file that can be committed to the project.
```
bvaughn, line 19: Super handy. Love seeing us adding in-context READMEs

packages/icons/scripts/icons.js
```diff
@@ -0,0 +1,35 @@
...
+const dotenv = require("dotenv");
+const { fetchImages } = require("figma-tools");
+const prettier = require("prettier");
+const svgtojsx = require("svg-to-jsx");
+
+dotenv.config({ path: "../../.env" });
+
+const pretterOptions = JSON.parse(fs.readFileSync("../../.prettierrc", "utf-8"));
+
+console.log("Fetching icons...");
```
bvaughn, line 14: How long does the fetch step generally take?

Could add something like [progress-estimator](https://www.npmjs.com/package/progress-estimator) 😆  but I realize this is overkill

packages/icons/scripts/icons.js
```diff
@@ -0,0 +1,35 @@
...
+fetchImages({
+  fileId: "ASas6u2DMihEEzw8jPT1XC",
+  format: "svg",
+  filter: component => component.pageName === "0.3 Icons",
+}).then(async svgs => {
+  console.log("Building icons...");
+
+  const jsx = await Promise.all(svgs.map(svg => svgtojsx(svg.buffer)));
+  const data = svgs
+    .map((svg, index) => `export const ${pascalCase(svg.name)}Icon = () => ${jsx[index]}\n`)
```
bvaughn, line 25: Worth adding a warning if we ever encounter a duplicate key? I imagine that would be unexpected but 🤷🏼

bvaughn (approved):
Love it!

packages/icons/scripts/icons.js
souporserious: Yeah, definitely! It's fairly quick right now, but it can slow as the Figma file grows over time and we add more. I had ora in there before for loading, but was having ESM issues 🙄. I'll circle back on this though and clean things up! My hope is we can start building a little design system for CLI tooling using react-ink.

packages/icons/README.md
souporserious: ❤️ yeah! I love colocating as much as possible. The new docs architecture I'm going to propose will be similar where each component/hook/util will use local `README.mdx` files to build up all of the documentation.

PR #6747, opened 2022-05-13
Don't allow prefix badges to shrink

jcmorrow:
I was doing this with min-width before and:

A) that was just a lack of imagination on my part, forgetting to reach for flexbox fundamentals
B) that is no longer used as of the latest refactor anyways, so it was resulting in squished unicorns!

Before:

Poor unicorn!

Poor unicorn!

After:

Hooray!

Hooray!

In case anyone is wondering about the interaction here between things like width, min-width, flex, flex-shrink, etc.

Every item inside of a flex item gets some traits that decide how much it grows, shrinks, and what it's "basis" (the amount of room it will take up if it neither grew nor shrank) is. By default, everything is `0 1 auto`. You'll often see flex item traits in 3's like this. It's a shorthand, you should read it as `<flex-grow> <flex-shrink> <flex-basis>`. So by default stuff will shrink if it needs to, will not grow, and will use its width as its basis. Now, when stuff grows and shrinks in a flexbox, it grows and shrinks linearly by both its basis and its `flex-grow`, `flex-shrink` values. We can ignore flex-grow here because our items are already overflowing - they need to shrink. Since both items have a flex-shrink of `1` they are going to try and shrink by the same absolute amount - but one of them (the console message) is much larger, so while it is also shrinking, it's shrinking very little in proportion to how large it was to begin with, while the icon has gotten *way* smaller. In fact, the icon has gotten so much smaller that it is at its default min-width of 1px. Since flexbox respects min-height and min-width, it can shrink no further, and the rest of the shrinking will be done by the remaining items, which in this case is just the console message. By telling the icon "don't shrink" (or, technically, setting its flex-shrink to 0) we are saying "don't shrink at all, let the other items do all the shrinking", which is exactly what we want! Fun times with flexbox!

### DIFF
--- a/src/devtools/client/webconsole/components/Output/MessageIcon.js
+++ b/src/devtools/client/webconsole/components/Output/MessageIcon.js
@@ -39,22 +39,6 @@ export function MessageIcon(props) {
   }
 
   if (prefixBadge) {
-    const style =
-      prefixBadge == "unicorn"
-        ? {
-            height: 16,
-            marginLeft: 4,
-            marginRight: 8,
-            marginTop: 2,
-          }
-        : {
-            height: 7,
-            marginLeft: 9,
-            marginRight: 12,
-            marginTop: 8,
-            width: 7,
-            minWidth: 7,
-          };
     return <PrefixBadge prefixBadge={prefixBadge} />;
   }
 
--- a/src/ui/components/PrefixBadge.module.css
+++ b/src/ui/components/PrefixBadge.module.css
@@ -1,5 +1,6 @@
 :root:global(.theme-light) .consoleBadge,
 :root:global(.theme-dark) .consoleBadge {
+  flex-shrink: 0;
   width: 12px;
   height: 12px;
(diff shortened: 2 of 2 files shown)


jonbell-lot23 (approved):
Lovely, thank you!

I am going to copy/paste that flex explanation because I suspect I'm gonna need that reference :)

src/devtools/client/webconsole/components/Output/MessageIcon.js
```diff
@@ -39,22 +39,6 @@ export function MessageIcon(props) {
...
-            marginTop: 2,
-          }
-        : {
-            height: 7,
-            marginLeft: 9,
-            marginRight: 12,
-            marginTop: 8,
-            width: 7,
-            minWidth: 7,
-          };
```
jasonLaster, line 57: nice - this was particularly ugly code

jasonLaster (approved):
Seems legit to me

PR #6750, opened 2022-05-13
Global Loader 1.5

### LINKED ISSUE #6671: Global loader 1.5
We're updating our global loader!
[Link to design ticket](https://github.com/RecordReplay/design/issues/40)

**The problem we're trying to solve**
We handle buffering, loading, and indexing, and it can be hard to give people enough information without being overwhelming. This design attempts to be more clear (by using a literal progress indicator) without showing too much (we are calling everything "loading" and we're retiring the word "indexing" from this view.

image

**Details**

* In the default view, things will work the same as before
* There's a new unified "capsule" design to visually group everything
* We changed the icon to look more like a trim (rather than "full screen")
* When things are slow, we'll show a turtle icon
* When things fail, we'll show a storm icon
* Hovering over the left side could show indexing information in the timeline (to discuss)
* Hovering over the right side should show a highlgiht
* When focus mode is deployed, the right side should stay lit up to show state

**Design notes**

I need to cut some of these icons. Keep me posted before starting work and we can make sure we're synced!

### Code changes
* #6750
* #6769

bvaughn:
Relates to #6671

Loom: https://www.loom.com/share/04814b28824646539db56412cd5aac37

- [x] Add new capsule loader UI
- [x] Add new radial progress loading indicator
- [x] Detect loading/indexing error state (add timeout logic for now?)
- [x] Remove radial progress loading indicator from around play/pause button
- [x] Change the way we're calculating the completed percentage values as described here: https://github.com/RecordReplay/devtools/pull/6750/files#r872666374
- [x] Finalized wording in really-slow instructions popup modal
- [ ] Hover to show loaded/indexed sections
- [ ] Figure out a way to write some unit or e2e tests for the threshold logic.

I think this PR is ready for an initial review pass, but I will write some tests before it lands. (Probably on Monday morning.)

Also pending the outcome of this comment (https://github.com/RecordReplay/devtools/pull/6750#issuecomment-1126514196) there may be some changes.


src/ui/actions/app.ts
```diff
@@ -70,7 +71,46 @@ export function setupApp(store: UIStore) {
...
+  // The backend doesn't give up on loading and indexing; apparently it keeps trying until the entire session errors.
+  // Practically speaking though, there are cases where updates take so long it feels like things are broken.
+  // In that case the UI should show an error state.
+  //
+  // We can rely on the fact that even when loading takes a long time, we should still be getting regular progress updates.
+  // If too much time passes between these updates, we can infer that things are either slow, or we're in a stuck state (aka an "error" for all practical purposes).
+  //
+  // If another update eventually comes in we will reset the slow/timed-out flag.
+  const SLOW_THRESHOLD = 2500;
+  const TIMED_OUT_THRESHOLD = 10000;
```
bvaughn, line 83: These values are kind of arbitrary. Do we have better values to use here?

src/ui/components/Timeline/Capsule.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { getIndexingProgress, getLoadingStatusWarning, LoadingStatusWarning } from "ui/actions/app";
+
+import Icon from "../shared/Icon";
+
+import styles from "./Capsule.module.css";
+import { EditFocusButton } from "./EditFocusButton";
+import FocusInputs from "./FocusInputs";
+
+export default function Capsule() {
+  const indexingProgress = useSelector(getIndexingProgress) || 0;
```
bvaughn, line 13: Is this the right selector to be using?

I think I might want to switch to a selector that calculates 0-100% differently. Rather than the union of index and loaded, I think it might make more sense to say 100% is 100% of both of those (that way we wouldn't show no updates for a while even when e.g. indexing is updating).

jonbell-lot23:
(I am blown away by how quickly our UX issues are improving) 🥰

bvaughn:
@jonbell-lot23 Thinking through the error case– really this isn't an "error" in the typical sense. It's just that things are going so slow, we think maybe they're broken. We don't know _for sure_ though. So showing an error modal feels a little heavy handed. Maybe we should reconsider this state?

Screen Shot 2022-05-13 at 4 47 40 PM

Maybe instead we should just show the red error badge and if a user interacts with it (hovers? clicks?) we could show more– although we'd just be guessing. Reloading _might_ help but maybe not if the backend is down.

src/ui/components/Timeline/Capsule.tsx
bvaughn: I wrote a new one. Feels better.

src/ui/actions/app.ts
bvaughn: Let's go with them and iterate over time.

jonbell-lot23:
Ok I'm awake now @bvaughn :)

* It's important that errors provide some way forward. I had intended the error state to throw an apology modal that says to try again with a refresh or something. (We should also track every single time this happens to someone)

* Same with the turtle -- not sure exactly what a tooltip could say in this state, but worth considering in a follow-up. (We should also track all turtles as a key metric)

* In this design, the error still allows someone to change the window. There's a big benefit to this in cases where changing the window will work. But if changing the window will hit the same issues, I wonder if this is another reason for failure to force a modal and disable all other actions?

* The ease/speed that you're able to see loading status, focus on a region, drag it to the side, and see the video updating in real time is just extraordinary. Feels so smooth.

jonbell-lot23:
Oops our messages passed in transit @bvaughn !

I trust you to find the right balance here :) I just want to make sure totally fatal errors are honest about them. If we can't determine when they're fatal versus slow, I'd err on the side of really clear messaging. The current experience is too vague.

bvaughn:
> Also, going from "error" back into a state that's peachy seems odd. Errors are the end of a road, not a transient UI state.

Unfortunately, in this case, there is no "end of the road" (short of the whole session dying) – so if that's the approach we want, we'll have to drop the errored state entirely.

jonbell-lot23:
Totally agree that's how the code works, but from a user/mental model standpoint I'd propose we invent one. After x seconds (and the number can be very generous) I do think we want to trigger an "end of the road" error even if technically that's not the signal the backend is sending.

Otherwise we get infinite loading or error states replaced with not-error states, and both lead to confusion.

bvaughn:
I agree that going from an error state to a not-error state, without any user action, seems weird and we should probably avoid it.

In this case, my intuition is that maybe we should re-think the "error" state to be something softer / less final.

I propose we have 3 states:
1. Things are loading normally; here's the current progress.
2. Things are loading a little slowly; here's the current progress.
3. Things are loading really slowly; (there may be a problem;) here's your progress.

jonbell-lot23:
That's fair. So maybe it's:

* Show percentage
* Show turtle
* Show cloud

But instead of saying "error" maybe we keep the percentage. And maybe it doesn't need to change color so dramatically either. Maybe just a foreground color and no pink background color.

And then when you click into it we can explain things in a more wordy way, perhaps in a modal. We won't say "error" or force the entire UI to stop, but we can say things are going slower than expected and recommend some things to try.


(Separate philosophical discussion that will be fun to have later -- some of the best UX gets that way because it does indeed break from the implementation in order to align more closely with user expectations. Seems counter-intuitive but it's one of my favourite topics when it comes to making great interactions)

bvaughn:
https://www.loom.com/share/04814b28824646539db56412cd5aac37

jonbell-lot23:
Looking good @bvaughn ! As for word smithing, here's a quick bit of copy we can iterate on:

## Loading

Loading is going slower than expected. Try using <link>Focus Mode</> to narrow down further, and please let us know on our [Discord](https://discord.com/channels/779097926135054346/798693728999047203) if you think something is wrong.

--

And I'd add a drop shadow on the modal, right now it's sort of floating in space.

bvaughn:
Screen Shot 2022-05-13 at 8 07 55 PM
Screen Shot 2022-05-13 at 8 07 57 PM

jonbell-lot23:
> Screen Shot 2022-05-13 at 8 07 55 PM
> 
> Screen Shot 2022-05-13 at 8 07 57 PM

Holy cow that's long.

## Loading is running slowly

Use link-> Focus Mode <- to narrow what's being loaded. [Chat with us on Discord](https://discord.com/channels/779097926135054346/798693728999047203)

bvaughn:
Haha, yes! We agree!

bvaughn:
Screen Shot 2022-05-13 at 9 01 24 PM
Screen Shot 2022-05-13 at 9 01 26 PM

bvaughn:
I _think_ the `react_devtools` test failure is expected. It's happening on master too.

jasonLaster (approved):

bvaughn:
Jason and I did a remote review of this code and we decided to land it today so that @jonbell-lot23 has the chance to play with it and gather feedback (or make tweaks) on Monday.

Then on my Monday, I'll do the remaining two items:
* Show loaded/indexed progress on hover
* Write unit tests

(Also incorporate any changes Jon requests.)

PR #6752, opened 2022-05-13
Circular import fixes, part 8: migrate breakpoints reducers to RTK

markerikson:

This PR:

- Removes the dead portion of the "context" middleware
- Updates the "context" middleware to handle looking for `action.meta.cx`, to work with RTK actions
- Removes an attempt to debounce prefs persistence, which was breaking persistence because the new/old state values were ending up the same somehow
- Migrates the `breakpoints and `pendingBreakpoints` reducers to RTK

This doesn't fix many import cycles by itself.  However, a bunch of the cycles go through `utils/breakpoint/index.ts`, which imports `ThreadFront` just to read `ThreadFront.recordingId`.  _That_ in turn gets used while generating some ID strings.

The next step is to refactor our `ThreadFront` usages to minimize imports, including injecting it into `extraThunkArgs`.  From there I can update these actions to accept `recordingId` as part of the payload, and _that_ will let me knock out about another 10 cycles.

I've done some checks and at least confirmed that I can still add and remove breakpoints and that the "requested breakpoints" are getting persisted into IndexedDB correctly.  But, some additional hands-on QA checks would be helpful.

### DIFF
--- a/src/devtools/client/debugger/src/actions/breakpoints/breakpoints.ts
+++ b/src/devtools/client/debugger/src/actions/breakpoints/breakpoints.ts
@@ -12,6 +12,11 @@ import {
   getBreakpointsForSource,
   getBreakpointsForSourceId,
 } from "../../reducers/breakpoints";
+import {
+  removeRequestedBreakpoint,
+  removeBreakpoints as removeBreakpointsAction,
+} from "../../reducers/breakpoints";
+import { setBreakpoint } from "../../reducers/breakpoints";
 import { Source } from "../../reducers/sources";
 import {
   getBreakpointsList,
@@ -29,7 +34,6 @@ import { setBreakpointPositions } from "./breakpointPositions";
 import {
   _removeBreakpoint,
   removeBreakpointOption,
-  removeRequestedBreakpoint,
   addBreakpoint,
   enableBreakpoint,
   disableBreakpoint,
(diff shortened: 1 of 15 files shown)


src/devtools/client/debugger/src/actions/breakpoints/breakpoints.ts
```diff
@@ -317,14 +321,14 @@ export function addBreakpointAtColumn(cx: Context, location: Location): UIThunkA
...
-  return {
-    breakpoint: {
-      ...breakpoint,
-      options: { ...breakpoint.options, prefixBadge },
-    },
-    type: "SET_BREAKPOINT",
-  };
+export function setBreakpointPrefixBadge(
+  breakpoint: Breakpoint,
+  prefixBadge: Breakpoint["options"]["prefixBadge"]
```
jaril, line 326: Just curious — what's the benefit of what looks to be a computed (?) type here instead of just string? I'm guessing it's just so if we change prefixBadge this gets updated as well?

jaril (approved):

src/ui/setup/prefs.ts
```diff
@@ -99,12 +105,14 @@ const actualUpdatePrefs = (state: UIState, oldState: UIState) => {
...
+
+  if (state.pendingBreakpoints && oldState.pendingBreakpoints) {
+    // @ts-ignore `asyncStoreHelper` is untyped, so TS doesn't know keys here
+    updateDebuggerAsyncPrefs(state, oldState, "pendingBreakpoints", getPendingBreakpoints);
+  }
   maybeUpdateReplaySessions(state);
 };
 
-// Avoid a small bit of overhead of checking prefs after _every_ dispatch
-export const updatePrefs = debounce(actualUpdatePrefs, 50);
```
jaril, line 106: :(

src/devtools/client/debugger/src/actions/breakpoints/breakpoints.ts
markerikson: The actual type is an enum/union of some kind, so it was faster to just point to "whatever the type of that field is" than look it up :)

PR #6753, opened 2022-05-13
Query annotations separately by kind

### LINKED ISSUE #6749: Reenable React DevTools
#6738 disabled fetching annotations if the user hasn't enabled *Redux* DevTools. But if we don't fetch any annotations, this also disables  *React* DevTools.

jcmorrow:
We recently added support for [Redux DevTools](https://github.com/RecordReplay/devtools/commit/ac248bd53e5687bce15b06928879aa90cbe0513e) to our existing support for React DevTools. This is great, but there's been a couple of realizations since then:

A) The Redux DevTools annotations can be quite large (for a replay of Replay that annotation is around 50MB)
B) We already knew this, but this really highlights why it would be nice to have filtering by annotation kind.

[We have added the ability to filter by annotation kind when requesting annotations](https://github.com/RecordReplay/backend/pull/5585/), so now we just have to use that capability. The tricky part about this is that our protocol's `Session` only expects to have one `annotationsListener` no matter how many queries we have out for annotations. The way I've gotten around this is having a couple of `Map`s that we track in thread and call the correct callbacks when we get results of a particular type. There could be a subtle conflict here where if you have queried for both `all` annotation kinds and a particular annotation kind, you will get your callback called twice with the same item. I think we could actually solve that by returning the protocol message ID in the response (which I feel like maybe we should just always do?). I've also added back in the separate initialization of React DevTools, and left the querying of Redux DevTools annotations behind the feature flag, so it will still have to be turned on for us to even ask about Redux DevTools annotations from the backend.

Fixes #6749 

P.S. Don't merge this yet, it doesn't seem like the backend is responding to `kind` yet.

### DIFF
--- a/src/protocol/thread/thread.ts
+++ b/src/protocol/thread/thread.ts
@@ -28,7 +28,6 @@ import {
   TimeStamp,
   unprocessedRegions,
   loadedRegions,
-  annotations,
   SameLineSourceLocations,
   RequestEventInfo,
   RequestInfo,
@@ -36,7 +35,9 @@ import {
   FunctionMatch,
   responseBodyData,
   requestBodyData,
+  findAnnotationsResult,
 } from "@recordreplay/protocol";
+import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
 
 import { MappedLocationCache } from "../mapped-location-cache";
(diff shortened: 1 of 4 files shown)


markerikson (approved):
Well, this _looks_ plausible :)  Let's ship it as soon as the backend cooperates!

PR #6758, opened 2022-05-13
Circular import fixes, part 9: Remove most `ThreadFront` imports

markerikson:

This PR: 

- Adds `ThreadFront` to `extraThunkArgs`
- Migrates a bunch of JS thunk files to TS
- For _all_ thunk files, changes their usage of `ThreadFront` to read from `extraThunkArgs` and removes direct imports of `ThreadFront`
- Updates the "set/remove breakpoint" actions to accept `recordingId` as an argument, and puts it into the action.  This includes updating dispatch sites to add that argument.
- Updates `utils/breakpoints/index.ts` to take `recordingId` as an argument
- Extracts some source visualizer functions that needed `ThreadFront` into their own file, so that `utils/source.js` doesn't have to import that.

This knocks things down to 37 cycles left.

Also, we're down to maybe 15-20 files that directly import `ThreadFront`, and most of them are component files.


jcmorrow (commented):
Happy to read through this today, but since you'll have the most context and its near EOD for you should we just say this is a merge-on-Monday situation?

markerikson:
Yep, not at all critical to merge today, and I'm signing off in the next couple minutes :)

src/devtools/client/webconsole/actions/network.ts
```diff
@@ -11,6 +10,8 @@ import {
 let onResponseBodyPart: (responseBodyParts: responseBodyData) => void;
 
 export const setupNetwork = async (store: UIStore) => {
+  const ThreadFront = store.dispatch((dispatch, getState, { ThreadFront }) => ThreadFront);
```
jasonLaster, line 13: i'm not sure if i like this suggestion, but might be cleaner to shove this somewhere

kinda regret saying this, but maybe `const ThreadFront = store.getThreadFront()` this is gross, but doing the same thing and atleast `ThreadFront` isn't actually a global.

Alternatively, we could pass ThreadFront into `setupNetwork` which is fine in these cases

markerikson: Yeah, passing `ThreadFront` into these setup functions would also be an option, actually.  Runs the same either way, but at least it's more explicit.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -127,8 +126,9 @@ export default async function DevTools(store: AppStore) {
 
   bootstrapWorkers();
 
-  const extraThunkArgs = {
+  const extraThunkArgs: ThunkExtraArgs = {
     client: clientCommands,
+    ThreadFront: ThreadFront,
```
jaril, line 131: 👍

jaril (approved):

markerikson:
Rebased on top of `master` after merging #6752 ("Fixes Part 8")

PR #6760, opened 2022-05-13
add bulk privacy toggle

### LINKED ISSUE #6495: Library bulk edit should show the privacy toggle
from @ceceliacreates 
>I was expecting "Make public" to be an option here when I bulk select. 🤔

image

souporserious:
Adds a bulk action for toggling selected items privacy state.

closes #6495

## Future Considerations

I think we can clean the UI up a little more once we can disable `Dropdown` items to add a signal when the selected items are already in a particular state and disabling the action (e.g. all selected items are private so disable/signal this in the Make Private item).

It would also be nice to figure out some sort of optimistic update mechanism that immediately changes the state or shows indication something is happening.

## Recording

https://user-images.githubusercontent.com/2762082/168399331-cf55d9ce-1f37-4988-a9f4-cf0c8c3a4ebb.mov

### DIFF
--- a/src/ui/components/Library/BatchActionDropdown.tsx
+++ b/src/ui/components/Library/BatchActionDropdown.tsx
@@ -13,7 +13,7 @@ import PortalDropdown from "../shared/PortalDropdown";
 import MoveRecordingMenu from "./MoveRecordingMenu";
 import { useConfirm } from "../shared/Confirm";
 import { Recording } from "ui/types";
-import { isPrivilegedUser } from "ui/hooks/users";
+import { useIsPublicEnabled } from "ui/utils/org";
 
 const getConfirmOptions = (count: number) => {
   if (count === 1) {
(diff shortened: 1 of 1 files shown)


src/ui/components/Library/BatchActionDropdown.tsx
```diff
@@ -13,7 +13,7 @@ import PortalDropdown from "../shared/PortalDropdown";
 import MoveRecordingMenu from "./MoveRecordingMenu";
 import { useConfirm } from "../shared/Confirm";
 import { Recording } from "ui/types";
-import { isPrivilegedUser } from "ui/hooks/users";
```
souporserious, line 16: This was unused in the file 🧹

jaril (approved):

src/ui/components/Library/BatchActionDropdown.tsx
```diff
@@ -105,6 +119,12 @@ function BatchActionDropdown({
       distance={0}
     >
       <Dropdown menuItemsClassName="z-50">
+        {isPublicEnabled ? (
+          <>
+            <DropdownItem onClick={() => setSelectedIdsIsPrivate(true)}>Make private</DropdownItem>
+            <DropdownItem onClick={() => setSelectedIdsIsPrivate(false)}>Make public</DropdownItem>
```
bvaughn, line 125: If we were feeling fancy 😁 we could disable either of these options when all of the selected IDs were already private or public

bvaughn (approved):

src/ui/components/Library/BatchActionDropdown.tsx
souporserious: Ah yeah, I'll do this soon since our current `DropdownItem` implementation doesn't accept `disabled`. Our `DropdownButton` component does, but doesn't accept `onClick` 😅

PR #6762, opened 2022-05-14
Remove semgrep manual integration

jcmorrow:
We now run this as part of Trunk, so we don't need a separate task for it. I think this also means there is also a repo secret that we can clear out or something?


jcmorrow:
NVM, I think I was fooled. Last time we talked to the Trunk people we had talked about using Trunk to run semgrep so it would be easier to run locally, and then I saw this:

CleanShot 2022-05-13 at 21 36 33

And figured that meant that we must have moved over and just not deleted the old action, but I think that's actually a "semgrep trunk" check, not a "Trunk semgrep" check. Anyways, we should move over, it'd be nice to have an easier to change config and to see results locally.

PR #6764, opened 2022-05-14
Adds a session callbacks API so that it is easier to observe session level events

jasonLaster:
1. adds SessionCallbacks
2. Moves createSession into `socket.ts`
3. Extracts onSocketError and onSocketClose out of `socket.ts`

### DIFF
--- a/src/protocol/socket.ts
+++ b/src/protocol/socket.ts
@@ -8,17 +8,13 @@ import {
   CommandParams,
   CommandResult,
 } from "@recordreplay/protocol";
-import { UIStore, UIThunkAction } from "ui/actions";
-import { setUnexpectedError } from "ui/actions/errors";
-import { UnexpectedError } from "ui/state/app";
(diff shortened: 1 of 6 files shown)


src/protocol/socket.ts
```diff
@@ -78,9 +110,20 @@ export function initSocket(store: UIStore, address: string) {
...
-  const onclose = makeInfallible(() => store.dispatch(onSocketClose()));
-  const onerror = makeInfallible((evt: Event) => store.dispatch(onSocketError(evt, false)));
-  const oninitialerror = makeInfallible((evt: Event) => store.dispatch(onSocketError(evt, true)));
+
+  const onclose = makeInfallible(() => {
+    gSocketOpen = false;
+    gSessionCallbacks?.onSocketClose(willClose);
+  });
+
+  const onerror = makeInfallible((evt: Event) =>
```
bvaughn, line 119: 🤣  "makeInfallible"

jasonLaster: if only... 🤣

PR #6765, opened 2022-05-14
Add a sidepanel for seeing protocol messages

jcmorrow:
This is just a basic addition on top of #6764 , putting stuff into Redux, displaying it in the SidePanel, etc. There are definitely some things to be improved, but I think this is a helpful starting point!

https://user-images.githubusercontent.com/5903784/168417383-60c1b139-67f0-4dc7-8e40-dd9aabe33660.mp4

### DIFF
--- a/src/ui/components/ProtocolViewer.tsx
+++ b/src/ui/components/ProtocolViewer.tsx
@@ -1,0 +1,91 @@
+import classNames from "classnames";
+import { padStart } from "lodash";
+import { useSelector } from "react-redux";
+import { getProtocolRequests, RequestSummary } from "ui/reducers/protocolMessages";
+
+const msAsMinutes = (ms: number) => {
+  const seconds = Math.round(ms / 1000.0);
+  return `${Math.floor(seconds / 60)}:${padStart(String(seconds % 60), 2, "0")}`;
+};
+
+const fullMethod = (request: RequestSummary): string => {
+  return `${request.class}.${request.method}`;
+};
+
+type RequestSummaryChunk = {
+  count: number;
+  method: string;
+  pending: boolean;
+  errored: boolean;
+  startedAt: number;
+};
+
+type ChunkReducer = {
+  chunks: RequestSummaryChunk[];
+  current: RequestSummaryChunk;
+};
+
+const chunkedRequests = (requests: RequestSummary[]): RequestSummaryChunk[] => {
+  return requests.reduce(
+    (acc: ChunkReducer, request: RequestSummary) => {
+      const { current } = acc;
+      if (
+        current.method === fullMethod(request) &&
+        current.pending === request.pending &&
+        current.errored === request.errored
+      ) {
+        current.count++;
+      } else {
+        acc.chunks.push(current);
+        acc.current = {
+          count: 1,
+          errored: request.errored,
+          method: fullMethod(request),
+          pending: request.pending,
+          startedAt: request.recordedAt,
+        };
+      }
+      return acc;
+    },
+    {
+      chunks: [],
+      current: {
+        count: 0,
+        method: "",
+        pending: false,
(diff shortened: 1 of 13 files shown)


markerikson:
For reference: how do you turn on this panel or that enhanced timeline view?

src/ui/components/ProtocolViewer.tsx
```diff
@@ -0,0 +1,91 @@
...
+        errored: false,
+        startedAt: 0,
+      },
+    }
+  ).chunks;
+};
+
+const ProtocolViewer = () => {
+  const requests = useSelector(getProtocolRequests);
+  const chunks = chunkedRequests(requests);
```
markerikson, line 65: :speech_balloon: was debating if it's worth memoizing this, but I guess this will only re-render when that list changes anyway

jcmorrow:
Good question @markerikson ! You can enable it by running `app.features.protocolTimeline = true` in your console 😄

src/ui/reducers/protocolMessages.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export interface ProtocolMessagesState {
+  events: (ProtocolEvent & Recorded)[];
+  requests: RequestSummary[];
+  responses: (CommandResponse & Recorded)[];
+  errors: (CommandResponse & Recorded)[];
+}
+
+const protocolMessagesSlice = createSlice({
```
markerikson, line 28: :speech_balloon: YAYYYYY MORE RTK USAGE! :)

src/ui/reducers/protocolMessages.ts
```diff
@@ -0,0 +1,74 @@
...
+      const request = state.requests.find(r => r.id === action.payload.id);
+      if (request) {
+        request.pending = false;
+        request.errored = true;
+      }
+    },
+    requestSent(state, action: PayloadAction<CommandRequest & Recorded>) {
+      state.requests.push({
+        ...action.payload,
+        class: action.payload.method.slice(0, action.payload.method.indexOf(".")),
```
markerikson, line 58: :hand: tbh I have a hard time reading the `.slice()` bits here :)

I _think_, from reading, that this is trying to take something like `Graphics.refreshSomething`, and split it?

Could you just do `const [ requestClass, requestMethod ] = action.payload.method.split(".")` and use those?

src/ui/reducers/protocolMessages.ts
```diff
@@ -0,0 +1,74 @@
...
+        errored: state.errors.some(r => r.id === action.payload.id),
+      });
+    },
+  },
+});
+
+export const { errorReceived, eventReceived, requestSent, responseReceived } =
+  protocolMessagesSlice.actions;
+
+export default protocolMessagesSlice;
```
markerikson, line 70: :hand: Normally we just export the reducer itself: `export default protocolMessageSlice.reducer`.  Not critical either way, but this matches the "ducks pattern" - default export the reducer, named export the action creators.

markerikson (approved):
Couple nitpicks that would be nice to tweak, but nothing blocking.

src/ui/reducers/protocolMessages.ts
jcmorrow: haha, yes, thank you, that would have a been a much more intelligent way of phrasing it.

src/ui/reducers/protocolMessages.ts
jcmorrow: Ah, perfect, thanks yeah I was trying to figure out the "right" way to do this.

src/ui/components/ProtocolViewer.tsx
jcmorrow: yep, I had the same thought basically

PR #6769, opened 2022-05-15
Global loader 1.5

bvaughn:
Follow up for #6750. Relates to #6671.

Loom: https://www.loom.com/share/c4217f31400d4542a77fda8d19815b93

## Notes from Brian:
- [x] Remove "really-slow" enum and popup.
- [x] Hover to show loaded/indexed sections
  - [x] Fix TypeScript type on `array.map`
  - [x] Settle on hovered color for dark theme (@jonbell-lot23 do you have a preference?) (see this comment: https://github.com/RecordReplay/devtools/pull/6769#issuecomment-1127069755)
- [ ] ~~Write some unit or e2e tests for the threshold logic.~~

## Notes from Jon:
Hello @bvaughn ! Here are some minor UX tweaks plus a Loom walking through them:
https://www.loom.com/share/a3d1c97418cb4eea8db006693f6cca1f

**Small tweaks**

- [x]  The / is jumping when I go into edit mode
- [x]  Hovering over the capsule after loading shouldn’t change the cursor or light up the region in light blue. It’s just for the loading state
- [x]  Icon in the description should have a white background instead of dark
- [x]  Description should be “Focus Mode lets you specify a region for your debugging.”

**Cancel vs discard**

- [x]  Cancel versus discard: we’re canceling, not discarding
- [x]  Would be good to bring discard back — can do it later though

**Colors**

- [x]  Textfield in dark theme needs a background color
- [x]  Color should match the circular loader in both modes (see comment below with details)


src/ui/components/Timeline/LoadingProgressBars.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return regions.map(({ begin, end }, index) => (
+    <div
+      key={index}
+      className={styles.Region}
+      style={{
+        left: `${(begin.time / endTime) * 100}%`,
+        width: `${clamp(((end.time - begin.time) / endTime) * 100, 0, 100)}%`,
+      }}
+    />
+  )) as any;
```
bvaughn, line 26: Why is TypeScript complaining about the return type of this map?

```
src/ui/components/Timeline/index.tsx:125:38 - error TS2786: 'LoadingProgressBars' cannot be used as a JSX component.
  Its return type 'Element[]' is not a valid JSX element.
    Type 'Element[]' is missing the following properties from type 'ReactElement<any, any>': type, props, key

125             {showLoadingProgress && <LoadingProgressBars />}
                                         ~~~~~~~~~~~~~~~~~~~
```

This error means nothing to me.

jonbell-lot23:
Hello @bvaughn ! Here are some minor UX tweaks plus a Loom walking through them:
https://www.loom.com/share/a3d1c97418cb4eea8db006693f6cca1f

**Small tweaks**

- [ ]  The / is jumping when I go into edit mode
- [ ]  Hovering over the capsule after loading shouldn’t change the cursor or light up the region in light blue. It’s just for the loading state
- [ ]  Description should be “Focus Mode lets you specify a region for your debugging.”
- [ ]  Icon in the description should have a white background instead of dark

**Cancel vs discard**

- [ ]  Cancel versus discard: we’re canceling, not discarding
- [ ]  Would be good to bring discard back — can do it later though

**Colors**

- [ ]  Textfield in dark theme needs a background color
- [ ]  Color should match the circular loader in both modes [Stay tuned for final colors, I'm working on them]

src/ui/components/Timeline/LoadingProgressBars.tsx
markerikson: This is a known weakness in the React TS types.

As _you_ well know, it's been legal to return an array from rendering ever since React 16.

But, the TS types somehow struggle with this.

The solution is to wrap the array in a seemingly-unnecessary fragment to make TS happy.

jonbell-lot23:
Ok, here's what I've come up with.

* Use primary accent to represent things that have loaded
* Use a color close to the chrome background to show things not yet loaded

image

Details below ->

**Dark theme**

Dotted line: #01ACFD (should be standard primaryAccent, let me know if not)
Indexed solid line: #01ACFD (should be the same)
Completed radial loader state: #01ACFD (should be the same)
-
Not yet indexed line color: #00567F (our accent color made darker)
Not yet completed radial loader state: #00567F (should be the same)
-
Capsule background color: #3D4552
Capsule hover color: #232938



**Light theme**

Dotted line: #01ACFD (should be standard primaryAccent and same as dark theme)
Indexed solid line: #01ACFD (should be the same)
Completed radial loader state: #01ACFD (should be the same)
-
Not yet indexed line color: #ffffff
Not yet completed radial loader state: #ffffff (should be the same)
-
Capsule background color: #E6E6E6
Capsule hover color: #D8D8D8

bvaughn:
>  Hovering over the capsule after loading shouldn’t change the cursor or light up the region in light blue. It’s just for the loading st

Ooo I also fixed this in #6769 but no problem, whoever merges first is fine :)

bvaughn:
Okay I think I made all of the changes that you suggested, @jonbell-lot23: https://www.loom.com/share/c4217f31400d4542a77fda8d19815b93

src/ui/components/Timeline/LoadingProgressBars.tsx
bvaughn: Goofy.

bvaughn:
I'm going to work on adding some automated tests for the loader now, but the code should be ready to review at any point.

src/ui/actions/app.ts
```diff
@@ -83,14 +83,13 @@ export function setupApp(store: UIStore) {
 
   // The backend doesn't give up on loading and indexing; apparently it keeps trying until the entire session errors.
   // Practically speaking though, there are cases where updates take so long it feels like things are broken.
-  // In that case the UI should show an error state.
+  // In that case the UI should show a visual indicator that hte loading is slow.
```
jcmorrow, line 86: little typo here on `the`

src/ui/components/Timeline/LoadingProgressBars.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { getZoomRegion } from "ui/reducers/timeline";
+import { overlap } from "ui/utils/timeline";
+
+import styles from "./LoadingProgressBars.module.css";
+
+export default function LoadingProgressBars() {
+  const loadedRegions = useSelector(getLoadedRegions);
+  const zoomRegion = useSelector(getZoomRegion);
+
+  const endTime = zoomRegion ? zoomRegion.endTime : 0;
```
jcmorrow, line 13: I'm guessing that the false branch here is something we don't expect to hit? `0` seems like it would make for an odd value of `endTime` in the calculations below.

src/ui/components/Timeline/index.tsx
```diff
@@ -120,14 +122,15 @@ export default function Timeline() {
...
+            {showLoadingProgress && <LoadingProgressBars />}
             <CurrentTimeIndicator editMode={editMode} />
             <Focuser editMode={editMode} setEditMode={setEditMode} />
           </div>
 
           <Tooltip timelineWidth={timelineDimensions.width} />
         </div>
 
-        <Capsule />
+        <Capsule setShowLoadingProgress={setShowLoadingProgress} />
```
jcmorrow, line 133: I'm curious because I haven't used `React.Context` a ton - when do you decide "I should make a context" vs. "I should pass a setter as a prop"? Is there a rule of thumb there? Or is it less about how much you're passing and more about how many consumers there are and how far apart they are in the hierarchy?

jcmorrow (approved):
Couple of comments, but looks good to me!

src/ui/components/Timeline/LoadingProgressBars.tsx
bvaughn: `zoomRegion` can be null, so we need to account for it somehow. Agreed it wouldn't make much sense though. Probably best to just early-return `null` in this case.

src/ui/components/Timeline/index.tsx
bvaughn: In general, the Context API exists so that you don't have to do "props drilling" – but there's a small overhead when reading from a context. In this case, since I'm directly rendering a child component, I just chose to pass the value directly as a prop and avoid the second context lookup.

This is pretty subjective and never something I'd block a PR on.

bvaughn:
Going to punt the unit tests to a follow up so we can land the UI improvements in this PR.

Issue #6774, opened 2022-05-16
Sporadic failures of console_async_eval e2e test


hbenl:
console_async_eval fails sporadically because the `executeInConsole()` function from the test harness doesn't trigger an execution.
The reason is a typical `useState()` bug:
- we create a state variable `value`:
https://github.com/RecordReplay/devtools/blob/7707c393248900004fc5e9c3f073e1659f3f5064/src/devtools/client/webconsole/components/Input/JSTerm.tsx#L51
- then we set `executeRef.current` to a function capturing that variable:
https://github.com/RecordReplay/devtools/blob/7707c393248900004fc5e9c3f073e1659f3f5064/src/devtools/client/webconsole/components/Input/JSTerm.tsx#L69-L72
After `setValue()` is called, this function will still reference the *old* captured `value`. Eventually `JSTerm` will be rerendered due to the call to `setValue()` and that will set `executeRef.current` to a new function referencing the new `value`. But if `executeRef.current()` is called before that happens, it will work with an outdated `value`.

See the comments in this recording: https://app.replay.io/recording/replay-of-devtools-64leppweg-recordreplayvercelapp--d60a4f29-7aa3-48f3-bcb9-76ab5a3563b8

Issue #6366, opened 2022-04-13
Investigate replacing Tip Tap with Lexical


bvaughn:
I dug into our comments code as part of investigating #6138. After digging in, I'm not super thrilled with how Tip Tap works within a React app (e.g. https://github.com/RecordReplay/devtools/issues/6138#issuecomment-1096891604). I think it would be worth looking into replacing it with [Lexical](https://www.lexical.dev/) ([github.com/facebook/lexical](https://github.com/facebook/lexical)). Lexical was designed to work well with React specifically, and a lot of focus was put into making it super performant and accessible.

jcmorrow:
I think my primary investigation would center around this piece: https://github.com/facebook/lexical/blob/main/examples/decorators.md since we have lots of plans for the future of the editor (mentions, tighter integrations with external platforms, etc). Also - I know that there are some open PR's on tiptap to improve React integration - https://github.com/ueberdosis/tiptap/pull/2626 and https://github.com/ueberdosis/tiptap/pull/2562 - which might improve some parts of the TipTap internal without us having to put in the effort to switch over to another editor.

bvaughn:
> I think my primary investigation would center around this piece: https://github.com/facebook/lexical/blob/main/examples/decorators.md since we have lots of plans for the future of the editor

Yup, great call out. Lexical _is_ extensible, although I can't attest to how easy it is or how well it works.

> I know that there are some open PR's on tiptap to improve React integration - https://github.com/ueberdosis/tiptap/pull/2626 and https://github.com/ueberdosis/tiptap/pull/2562 - which might improve some parts of the TipTap internal without us having to put in the effort to switch over to another editor.

Those PRs don't look like they change the Tip Tap / React integration in a fundamental way, unless I'm misreading them.

trueadm:
Please feel free to reach out to the Lexical team, either on our github or discord and we can help you folks wherever possible! :)

jcmorrow:
> Those PRs don't look like they change the Tip Tap / React integration in a fundamental way, unless I'm misreading them.

Yeah maybe I was too optimistic about the problems these solve. I saw [this `requestAnimationFrame` deleted](https://github.com/ueberdosis/tiptap/pull/2626/files#diff-13acb5e551812e195c1140c10ba5ac298a0005996d07756cfd77c2d4194cb350L20-L26) and was hopeful that it could solve some of our issues, but I didn't realize that it's just shuffling stuff around to a different hook 🙁

[...]

PR #6775, opened 2022-05-16
Add workaround for #6774

hbenl:
@bvaughn I don't like this workaround, but the interaction between all those `useState()` and `useRef()` hooks in `JSTerm` and `ControlledCodeMirror` is making my head spin. Please have a look at #6774, perhaps you've got an idea for a proper fix.

### DIFF
--- a/src/test/harness.js
+++ b/src/test/harness.js
@@ -548,9 +548,12 @@ async function toggleScopeNode(text) {
   return toggleObjectInspectorNode(node);
 }
 
-function writeInConsole(value) {
+async function writeInConsole(value) {
   window.jsterm.setValue(value);
-  return waitUntil(() => window.jsterm.editor.getValue() === value);
+  await waitUntil(() => window.jsterm.editor.getValue() === value);
+  // workaround for #6774
+  // TODO [hbenl] remove this workaround once JSTerm is fixed (#6778)
+  await waitForTime(100);
 }
 
 async function executeInConsole(value) {


src/test/harness.js
```diff
@@ -548,9 +548,11 @@ async function toggleScopeNode(text) {
   return toggleObjectInspectorNode(node);
 }
 
-function writeInConsole(value) {
+async function writeInConsole(value) {
   window.jsterm.setValue(value);
-  return waitUntil(() => window.jsterm.editor.getValue() === value);
+  await waitUntil(() => window.jsterm.editor.getValue() === value);
+  // workaround for #6774
```
bvaughn, line 554: Can we add a TODO comment here and pin it to a task? This (and the code it's working around) should be tracked.

bvaughn (approved):
Other than adding a TODO comment to that code, I'm not familiar with the components you mentioned. Looking at `JSTerm` now, and modifying a ref during render is an [unsafe side effect](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects) that we should probably remove:
https://github.com/RecordReplay/devtools/blob/7707c393248900004fc5e9c3f073e1659f3f5064/src/devtools/client/webconsole/components/Input/JSTerm.tsx#L69-L83

Particularly given the way we're using it to also run side effects from a previous render:
https://github.com/RecordReplay/devtools/blob/7707c393248900004fc5e9c3f073e1659f3f5064/src/devtools/client/webconsole/components/Input/JSTerm.tsx#L54

This component looks like it should be refactored and cleaned up. Maybe we could do it as part of #6366.

PR #6785, opened 2022-05-16
add keyboard shortcut for adding a comment

### LINKED ISSUE #5741: It should be possible to add a comment from anywhere
As noted in https://github.com/RecordReplay/devtools/issues/5527 , we need a keyboard shortcut that will add a shortcut in all scenarios.

* If the user is paused at a point in time, we use time
* If the user is paused at an execution point, we use that

Once we have the back-end logic wired up to a keyboard shortcut, the remaining work is just designing a button and slapping it into the Comments pane.

(For now, the shortcut can be shift-C, but this can be modified as we go)

souporserious:
Adds a "Shift+C" global keyboard shortcut for adding a comment from anywhere in the application.

closes #5741

### DIFF
--- a/src/ui/actions/comments.ts
+++ b/src/ui/actions/comments.ts
@@ -47,7 +47,7 @@ export function updatePendingCommentContent(content: string): UpdatePendingComme
 export function createComment(
   time: number,
   point: string,
-  user: User,
+  user: User, // TODO: user is no longer required and should be removed
   recordingId: RecordingId,
   options: CommentOptions
 ): UIThunkAction {
--- a/src/ui/components/KeyboardShortcuts.tsx
+++ b/src/ui/components/KeyboardShortcuts.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo } from "react";
+import { useEffect, useMemo } from "react";
 
 import { connect, ConnectedProps } from "react-redux";
 import { UIState } from "ui/state";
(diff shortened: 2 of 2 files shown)


jasonLaster (approved):

src/ui/actions/comments.ts
```diff
@@ -47,7 +47,7 @@ export function updatePendingCommentContent(content: string): UpdatePendingComme
 export function createComment(
   time: number,
   point: string,
-  user: User,
+  user: User, // TODO: user is no longer required and should be removed
```
souporserious, line 50: Chatted with Josh, `user` is no longer required so we should safely be able to refactor this and get rid of it in the future.

src/ui/components/KeyboardShortcuts.tsx
```diff
@@ -113,6 +120,29 @@ function KeyboardShortcuts({
       toggleQuickOpen(query, project);
     };
 
+    const addComment = (e: KeyboardEvent) => {
+      if (!e.target || !isEditableElement(e.target)) {
+        e.preventDefault();
+        const commentContent: any = pendingComment?.comment.content;
+        const hasComment =
```
souporserious, line 127: This feels gross, but not sure how else to determine when Tiptap has content 😕

src/ui/components/KeyboardShortcuts.tsx
```diff
@@ -113,6 +120,29 @@ function KeyboardShortcuts({
...
+          commentContent === ""
+            ? false
+            : Boolean(commentContent?.content && commentContent.content[0].content);
```
souporserious, line 130: This feels gross, but not sure how else to determine when Tiptap has content 😕

souporserious:
@jonbell-lot23 when you get a chance can you give this a quick test and let me know if it's working how you expected? Also, not sure who to ask, but should we track this event the same as the other `addComment` button using `trackEvent("breakpoint.add_comment")`?

jonbell-lot23 (approved):
Looks good!

Let's track a follow-on ticket for these items:

* Let's add a comment button into the header of the comments pane
* Let's add "Add comment" to the command palette. Maybe we don't know the shortcut just yet, we're still deciding if it should be shift-C

Nice work, woo!

Issue #6781, opened 2022-05-16
Line text comparisons do not memoize correctly, causing perf issues


markerikson:
@Andarist reported lag in the UI when debugging a giant TS compiler source file.

We initially thought this might be due to me tweaking some of the code in `getLineText` to remove a seemingly-unused argument, per [`06bfd4e` (#6722)](https://github.com/RecordReplay/devtools/pull/6722/commits/06bfd4e82ba195766a364ac28b56a70d1d179955) .  

However, after investigation, I've concluded that we're really dealing with some very broken attempts to memoize code that does a `sourceContent.split("\n")`, and that function is getting called _many_ times.

I've made a pair of replays that demonstrate the number of hits:

- Before my PR's cleanup changes: https://app.replay.io/recording/replay-scrolling-visible-breakpoints-pre-part-6-edits--461bab25-829e-4800-819d-8ef835ab15c5
- and after those changes: https://app.replay.io/recording/replay-scrolling-visible-breakpoints--fbabccd1-2b45-46c1-8769-1724505ad31e

The reproduction steps are:

- Open a source file with more lines of code than fit on screen (say, 100 or so)
- Add 2+ print statements to that file
- Now, scroll up and down to change the viewport

There's three functions that are at fault here.  First, the outer "visible breakpoints" selector:

```ts
export const visibleColumnBreakpoints = createSelector(
  getVisibleBreakpoints,
  getVisibleRequestedBreakpoints,
  getViewport,
  getSelectedSourceWithContent,
  getVisibleBreakpointPositions,
  (breakpoints, requestedBreakpoints, viewport, selectedSource, breakpointPositions) => {
    if (!selectedSource) {
      return [];
    }

    // We only want to show a column breakpoint if several conditions are matched
    // - it is the first breakpoint to appear at an the original location
    // - the position is in the current viewport
    // - there is atleast one other breakpoint on that line
    // - there is a breakpoint on that line
    // @ts-ignore nested field mismatch
    const allBreakpoints: Breakpoint[] = [...(breakpoints ?? []), ...(requestedBreakpoints ?? [])];

    // NOTE: column breakpoints are disabled by default because
    // we should first verify if the results make sense
    let positions;
    if (features.columnBreakpoints) {
      positions = breakpointPositions;
    } else {
      // collect all breakpoint positions but make sure that we don't have 2 positions
      // for the same line in the same source
      positions = uniqBy(
        allBreakpoints.map(bp => bp.location),
[...]

PR #6787, opened 2022-05-16
Optimize source-splitting behavior when checking line content

markerikson:
Will hopefully be a solution to #6781 .

Currently, this PR:

- Converts `utils/source.js` to TS
- Adds `WeakMap`-based caching of line-split source content strings, to avoid re-splitting every time the user scrolls in an editor when there's multiple breakpoints

I suspect there's some more memoization/caching wins to be had around here, but want to at least get this up as a preview

/cc @Andarist  - this may help with your giant TS compiler file?


jasonLaster (approved):

markerikson:
Well, @jasonLaster stamped this, so it clearly _must_ be ready to ship to prod! :)

PR #6789, opened 2022-05-16
Do not let the login page infinite loop

jcmorrow:
@ryanjduffy are there any other URL's I should ban from this param for similar reasons?

Fixes https://github.com/RecordReplay/devtools/issues/6788

### DIFF
--- a/src/ui/components/shared/Login/Login.tsx
+++ b/src/ui/components/shared/Login/Login.tsx
@@ -1,7 +1,7 @@
 import { gql } from "@apollo/client";
+import Services from "devtools/shared/services";
 import Link from "next/link";
 import React, { ReactNode, useEffect, useState } from "react";
-import Services from "devtools/shared/services";
 import { query } from "ui/utils/apolloClient";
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/Login/Login.tsx
```diff
@@ -177,6 +177,10 @@ export default function Login({ returnToPath = "" }: { returnToPath?: string })
   const { loginWithRedirect } = useAuth0();
   const [sso, setSSO] = useState(false);
 
+  if (returnToPath.startsWith("/login")) {
```
ryanjduffy, line 180: Would this be better handled in `pages/login.tsx`?

ryanjduffy (approved):

src/ui/components/shared/Login/Login.tsx
jcmorrow: I wanted to make put it as late in the chain of components as possible so I was confident that we wouldn't be missing a path, since `<Login />` is rendered from quite a few places.

ryanjduffy: I think the only case of the loop would be if it were rendered from the login page.

PR #6797, opened 2022-05-17
Move filtering to the backend

jaril:
https://github.com/RecordReplay/backend/pull/5604 needs to be merged first to add support for filtering in the backend, while not breaking the client side filtering feature.

This can be merged afterwards to move filtering to the backend.

### DIFF
--- a/src/ui/components/Library/ViewerRouter.tsx
+++ b/src/ui/components/Library/ViewerRouter.tsx
@@ -1,4 +1,5 @@
-import React, { useEffect } from "react";
+import React, { useContext, useEffect } from "react";
+
 import Viewer from "./Viewer";
 
 import { connect, ConnectedProps } from "react-redux";
@@ -13,6 +14,8 @@ import { BlankViewportWrapper } from "../shared/Viewport";
 import Base64Image from "../shared/Base64Image";
 import { sendTelemetryEvent } from "ui/utils/telemetry";
 
+import { LibraryFiltersContext } from "./useFilters";
+
 function ViewerLoader() {
   return (
     <div className="grid h-full w-full items-center justify-items-center bg-chrome">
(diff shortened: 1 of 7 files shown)


jaril:
@ceceliacreates before I get too far with this, do we already have any documentation on how filtering works in the library?

I don't think we do yet. I'm happy to write it to take it off your plate, but just wanted to make sure I check first on where it should go :)

src/ui/hooks/recordings.ts
```diff
@@ -421,11 +421,12 @@ export function useIsOwner() {
   return userId === recording.owner.id;
 }
 
-export function useGetPersonalRecordings():
-  | { recordings: null; loading: true }
-  | { recordings: Recording[]; loading: false } {
+export function useGetPersonalRecordings(
+  filter: string
```
ryanjduffy, line 425: optional?

src/ui/hooks/recordings.ts
```diff
@@ -444,13 +445,14 @@ export function useGetPersonalRecordings():
 }
 
 export function useGetWorkspaceRecordings(
-  currentWorkspaceId: WorkspaceId
+  currentWorkspaceId: WorkspaceId,
+  filter: string
```
ryanjduffy, line 449: optional?

ryanjduffy (commented):
Everything seems fine. Mind adding a replay of the changes running against cloud dev?

ryanjduffy (approved):

src/ui/hooks/recordings.ts
jaril: Does it need to be optional? I figured since it defaults to "", it should be okay to just pass it in as either an empty string or a real filter string, and then just handle the no filter case in the backend.

ryanjduffy: It's optional on the backend so it makes sense to mark it as such here too, I think. It doesn't really matter particularly because we won't encounter errors by being more restrictive on the front end.

ryanjduffy (approved):

src/ui/components/Library/ViewerRouter.tsx
```diff
@@ -22,9 +25,12 @@ function ViewerLoader() {
 }
 
 function MyLibrary() {
-  const { recordings, loading } = hooks.useGetPersonalRecordings();
+  const { filter } = useContext(LibraryFiltersContext);
+  const { recordings, loading } = hooks.useGetPersonalRecordings(filter);
   const { loading: nonPendingLoading } = hooks.useGetNonPendingWorkspaces();
 
+  console.log({ loading, recordings });
```
bvaughn, line 32: Can we remove this? 😁

PR #6805, opened 2022-05-17
Circular import fixes, part 10: fix _almost_ all remaining cycles

markerikson:

This PR:

- Migrates the `reducers/timeline` file to RTK to pull out a few more action creators
  - removes the "set zoom" action, because it isn't being used, but leaves `state.timeline.zoom` alone
- Refactors `Pause` to take `ThreadFront` as its constructor arg
- Extracts some specific utils and variables that were being circularly-imported
- Updates `next.config.js` with the "show cycle warnings" settings I've been using to check these

As of this PR, we've dropped from 174 cycles originally, down to 19.  Those last 19 are both not easily fixable, and _probably_ not causing problems, so I'm going to wrap up the task here.


src/devtools/client/debugger/src/utils/source.ts
```diff
@@ -398,7 +398,10 @@ export const getLineText = (asyncContent: SourceWithContent["content"], line: nu
   return lineText || "";
 };
 
-export function getTextAtPosition(asyncContent: SourceWithContent["content"], location: Location) {
+export function getTextAtPosition(
+  asyncContent: SourceWithContent["content"],
+  location: { column?: number; line?: number }
```
bvaughn, line 403: Curious why this change? Doesn't seem like we changed any imports to e.g. remove the `Location` import so it seems unnecessary?

src/protocol/thread/pause.ts
```diff
@@ -18,7 +18,7 @@ import {
 import { client } from "../socket";
 import { defer, assert, Deferred } from "../utils";
 import { ValueFront } from "./value";
-import { ThreadFront } from "./thread";
+import type { ThreadFront as ThreadFrontType } from "./thread";
```
bvaughn, line 21: I notice the call sights passing in `ThreadFront` now. Is this (`import type` vs `import`) not sufficient to fix the circular dependency/import issue?

src/ui/components/Comments/TranscriptComments/types.ts
```diff
@@ -1,7 +1,15 @@
...
   | { type: "new_comment"; comment: Comment }
   | { type: "reply"; comment: Reply }
   | { type: "new_reply"; comment: Reply };
+
+export const FocusContext = React.createContext({
+  autofocus: false,
+  isFocused: false,
+  blur: () => {},
+  close: () => {},
+});
```
bvaughn, line 15: I don't think the `FocusContext` belongs in a file named `types`. Make it it's own file?

src/ui/components/Events/eventKinds.tsx
```diff
@@ -0,0 +1,36 @@
...
+type EventKindIcon = "keyboard" | "ads_click" | "navigation";
+type EventKindLabel = "Key Down" | "Key Up" | "Key Press" | "Click" | "Navigation";
+export type EventKindKey =
+  | "event.keyboard.keydown"
+  | "event.keyboard.keyup"
+  | "event.keyboard.keypress"
+  | "event.mouse.click";
+
+export const getReplayEvent = (kind: Kind) => EVENT_KINDS[kind];
+export const getEventLabel = (event: ReplayEvent) => {
```
bvaughn, line 22: This function is only ever used inside of `src/ui/components/Events/Event.tsx` – why remove it from that file?

bvaughn (commented):
I don't think I understand some of the changes made 😅

src/devtools/client/debugger/src/utils/source.ts
markerikson: I changed a couple of the types for functions that _called_ this one, and TS didn't like the mismatches. (Frankly, we have a plethora of variations on `Location`-type objects in this codebase, and there's a _lot_ of inconsistency in what fields and types get passed around).  So, narrowing this down to just the fields that it actually uses fixed a compilation error

src/protocol/thread/pause.ts
markerikson: There were two things that happened here:

- Actually changing the `Pause` constructor to take a `ThreadFront` value as its argument
- Modifying the import statement to tell TS "just import the type, not the actual runtime class value".  We needed the type of the `ThreadFront` class in the constructor signature, and at the same time dropping the runtime import is actually part of the circular import fix

src/ui/components/Comments/TranscriptComments/types.ts
markerikson: Yeah, I can do that - I just saw that it was a non-component file that wasn't part of any circles yet and figured it was a decent place to move the context declaration.

src/ui/components/Events/eventKinds.tsx
markerikson: Yeah, this one could probably go back. I was cutting that whole chunk of code from the component file.

markerikson:
Well, looks like I was overly-aggressive in moving some things, and besides the comments I also totally borked the build.  Fixing...

jcmorrow (approved):
I did not carefully check every line of this, but it all looks pretty reasonable, and the preview branch seems ok 🤷

src/ui/reducers/timeline.ts
```diff
@@ -17,47 +17,43 @@ function initialTimelineState(): TimelineState {
     stalled: false,
     timelineDimensions: { left: 1, top: 1, width: 1 },
     unprocessedRegions: [],
+    /** @deprecated This appears to be obsolete for now? */
```
jaril, line 20: It is. Right now, startTime is always 0, endTime is always the recording duration, and scale is 1. We mostly use this as a replacement for a `duration` value now. Could use a refactor soon but just hesitant to do it/ask anybody to do it unless it's adjacent to something somebody's working on.

jaril (approved):

PR #6806, opened 2022-05-17
add initial NextJS docs site

souporserious:
The first step towards https://github.com/RecordReplay/design/issues/47. Also updates us to the latest NextJS version and sets up all of the `packages` directory to use NPM workspaces.


jasonLaster (approved):
Awesome! btw i think you can add npm workspaces with a pretty simple `npm -w` command.  It works well on the backend

PR #6807, opened 2022-05-17
Focus Mode 1.2: Context menu

### LINKED ISSUE #6735: Focus Mode 1.2
Build on top of #6571

- [x] Implement design in https://github.com/RecordReplay/design/issues/46, [Figma](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=12220%3A92306)
- [x] ~~Jon reaching out to @souporserious to learn how to do the cool auto-sync icon thing~~

bvaughn:
Resolves #6735

Loom: https://www.loom.com/share/432f63ecb2b4417895f1063f1942002f

- [x] Add context menu to Network requests table

Context menu for focus selection in Network panel

- [x] Add context menu to Console rows

Context menu for focus selection in Console panel

- [x] ~~Get finalized focus icons from @jonbell-lot23~~ New icons can be landed in a follow up.
- [x] Write tests for new actions

### DIFF
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -121,7 +121,6 @@
   font-size: var(--theme-code-font-size);
   line-height: var(--console-output-line-height);
   color: var(--console-message-color);
-  cursor: context-menu;
 }
 
 @media (min-width: 1000px) {
--- a/src/devtools/client/themes/webconsole.css
+++ b/src/devtools/client/themes/webconsole.css
@@ -121,6 +121,7 @@
   font-size: var(--theme-code-font-size);
   line-height: var(--console-output-line-height);
   color: var(--console-message-color);
+  cursor: context-menu;
 }
 
 @media (min-width: 1000px) {
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -235,13 +235,13 @@ class ConsoleOutput extends React.Component<PropsFromRedux, State> {
             <Dropdown>
               <DropdownItem onClick={this.setFocusStart}>
                 <>
-                  <Icon filename="set-focus-start" className="mr-4 bg-iconColor" />
+                  <Icon filename="set-focus-start" className="mr-4 bg-iconColor" size="large" />
                   Set focus start
                 </>
               </DropdownItem>
               <DropdownItem onClick={this.setFocusEnd}>
                 <>
-                  <Icon filename="set-focus-end" className="mr-4 bg-iconColor" />
+                  <Icon filename="set-focus-end" className="mr-4 bg-iconColor" size="large" />
                   Set focus end
                 </>
               </DropdownItem>
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -2,7 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { PointDescription } from "@recordreplay/protocol";
 import { Dispatch } from "@reduxjs/toolkit";
 import type { AnyAction } from "@reduxjs/toolkit";
 import { SourceLocation } from "devtools/client/debugger/src/reducers/types";
@@ -10,11 +9,9 @@ import { MessageContainer } from "devtools/client/webconsole/components/Output/M
 import { StateContext } from "devtools/client/webconsole/components/Search";
 import constants from "devtools/client/webconsole/constants";
 import { Frame, Message } from "devtools/client/webconsole/reducers/messages";
-import analysisManager from "protocol/analysisManager";
-import { ThreadFront } from "protocol/thread";
 import React from "react";
 import { connect, ConnectedProps } from "react-redux";
-import { setFocusRegion, syncFocusedRegion } from "ui/actions/timeline";
+import { setFocusRegionEndTime, setFocusRegionStartTime } from "ui/actions/timeline";
 import { ContextMenu } from "ui/components/ContextMenu";
 import { Dropdown, DropdownItem } from "ui/components/Library/LibraryDropdown";
 import Icon from "ui/components/shared/Icon";
@@ -22,6 +19,7 @@ import { selectors } from "ui/reducers";
 import { getFocusRegion } from "ui/reducers/timeline";
 import type { UIState } from "ui/state";
 import { isVisible } from "ui/utils/dom";
+import { convertPointToTime } from "ui/utils/time";
 
 import ConsoleLoadingBar from "./ConsoleLoadingBar";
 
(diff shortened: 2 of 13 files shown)


src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -202,12 +222,125 @@ class ConsoleOutput extends React.Component<PropsFromRedux> {
...
+    const startTime = focusRegion ? focusRegion.startTime : 0;
+    const endTime = time!;
+
+    dispatch(
+      setFocusRegion({
+        endTime,
+        startTime,
+      }) as unknown as AnyAction
+    );
+    dispatch(syncFocusedRegion() as unknown as AnyAction);
```
bvaughn, line 316: cc @markerikson This is a silly type refinement. Any suggestions?

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -232,11 +366,44 @@ function mapStateToProps(state: UIState) {
...
   dispatch,
-  openLink: actions.openLink,
 });
 
 const connector = connect(mapStateToProps, mapDispatchToProps);
 
 export default connector(ConsoleOutput);
+
+// TODO Is there a more central location for this utility method?
+async function getTimeForMessage(message: Message): Promise<number> {
```
jcmorrow, line 379: seems like it could reasonably be described as a timeline util? 🤷

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -232,11 +366,44 @@ function mapStateToProps(state: UIState) {
...
+  const { executionPoint, executionPointTime, timeStamp } = message;
+  if (timeStamp != null) {
+    return timeStamp;
+  } else if (executionPointTime != null) {
+    return executionPointTime;
+  } else if (executionPoint != null) {
+    return new Promise(async resolve => {
+      await analysisManager.runAnalysis(
+        {
+          effectful: true,
```
jcmorrow, line 389: you should be able to get away with passing `false` here since we are guaranteed not to be mutating anything I think?

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -30,14 +42,65 @@ const RequestTable = ({
...
+    setContextMenuData(null);
+  };
+
+  const setFocusEnd = () => {
+    setContextMenuData(null);
+
+    if (contextMenuData) {
+      const time = contextMenuData.row.original?.end;
+      if (time != null) {
+        // If this is the first time the user is focusing, start at the beginning of the recording (or zoom region).
```
jcmorrow, line 66: What do you think about DRYing this up with the same section in `ConsoleOutput`?

jcmorrow (approved):
Couple of small comments, but overall looks great!

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
bvaughn: Yeah that was my _first_ instinct too, but the generic concept of points and time isn't really specific to the timeline. I dunno. Maybe that's still a reasonable enough place to put it though.

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
bvaughn: Ah! I thought you said it should be `true` because of potential diverging. But I can change it. I don't know what it should be given the docs :D

src/ui/components/NetworkMonitor/RequestTable.tsx
bvaughn: Sure. I'll put a method in `utils/time` and also move the point-to-time lookup function there too.

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
bvaughn: I have an idea here.

src/devtools/client/themes/webconsole.css
```diff
@@ -121,6 +121,7 @@
   font-size: var(--theme-code-font-size);
   line-height: var(--console-output-line-height);
   color: var(--console-message-color);
+  cursor: context-menu;
```
markerikson, line 124: :speech_balloon: I had no idea this was even a _thing_ until today!

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
jcmorrow: Sorry! I was probably confusing true/false for `effectful` and `pure`, since they are opposites 😵

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
markerikson: Yeah, the issue here is that the base `Dispatch` type from Redux knows nothing about thunks, and `syncFocusedRegion` is a thunk.

On top of that, `connect` by itself only knows about the base `Dispatch` type.  Same with `useDispatch`, but it's a lot easier to create a custom hook that uses the _real_ type inferred from `store.dispatch`: https://redux.js.org/tutorials/typescript-quick-start#define-typed-hooks

The right answer here is to do `import type { AppDispatch } from "ui/setup/store";`, which has the correct types for the thunk middleware baked in because that was part of the store setup.  Then we need to re-type `dispatch` inside of this block as an `AppDispatch` instead.  (Not sure if there's a good way to do it at the `connect` level, unless we were to manually provide a `mapDispatch` function and cast in there).

That will make it safe to do `dispatch(syncFocusedRegion())` without actually casting on that line.

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
bvaughn: Okay. `effectful: false` makes more sense to me 👍🏼

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -30,14 +42,65 @@ const RequestTable = ({
 }) => {
   const { columns, getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } = table;
 
+  const dispatch = useDispatch();
```
markerikson, line 45: :hand: FYI this actually suffers from the same `dispatch` type problem as above, and we really _should_ be using the "typed hooks" solution I linked in the other comment.  But, there's a quirk in the React-Redux v7.x types that lets you sorta get away with dispatching thunks here, because it's really accepting `any` somehow.  But ideally this would be `const dispatch = useAppDispatch()` with the right type baked in.

markerikson (changes requested):
Left some notes on fixing the `dispatch` type issues with thunks.  

If you've got a couple minutes, it would be nice to copy-paste the "pre-typed hooks" recipe from the Redux docs into our codebase - surprised we don't have that set up anywhere.

But at a minimum, using the `AppDispatch` type inside that connected component will fix the cast issues.

jonbell-lot23:
@bvaughn as for ping, I thought maybe we could just do a single ping from the start/end point when it's set. Just a little way to draw attention to the new window. I'll noodle on this and report back.

PR #6817, opened 2022-05-18
add Icon component

souporserious:
This updates our Figma icon script to use [svgo](https://github.com/svg/svgo) to optimize and turn the raw svgs from Figma into an exported TypeScript components. This also adds a new `Icon` component for easily accessing and using icons.

@jonbell-lot23 I'll record a loom for this later, but for now the `packages/icons` README has all of the info for running the script. I already ran the script for the latest updates so the latest icons are in this PR if they are currently in the `Icons` page.

closes #6825

## Icon Component

The `Icon` component will be responsible for rendering icons. It currently takes the following `name` prop:

#### name: generated union from icon build

The name prop accepts a string that autocompletes with the available icons from the set.

```tsx
<Icon name="trash" />
```

## Future Considerations

We can eventually add a `size` prop that pulls from our design system tokens, but for now we'll keep them static at `20px` as they are in Figma and adjust as needed. I also want to do a pass on linting and error handling as it's lacking right now.

## Example

https://user-images.githubusercontent.com/2762082/168940735-61b6c9ea-a6d9-4155-a801-2de5006b1709.mov


packages/icons/icons.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function BasicUsage() {
+  const [filterValue, setFilterValue] = useState("");
+  return (
+    <div style={{ display: "flex", flexDirection: "column", gap: "2rem" }}>
+      {/* <input
+        placeholder="Filter icons..."
+        value={filterValue}
+        onChange={event => setFilterValue(event.target.value)}
+        style={{ padding: "0.25rem 0.5rem", border: "1px solid grey" }}
+      /> */}
```
bvaughn, line 20: Did you meant to leave this block commented out? Or did you mean to delete it and the `filterValue` state above?

packages/icons/scripts/icons.js
```diff
@@ -20,16 +17,41 @@ fetchImages({
...
-  const jsx = await Promise.all(svgs.map(svg => svgtojsx(svg.buffer)));
-  const data = svgs
-    .map((svg, index) => `export const ${pascalCase(svg.name)}Icon = () => ${jsx[index]}\n`)
+  const allIconExports = (
+    await Promise.all(
+      svgs.map(svg =>
+        transform(
+          svg.buffer.toString(),
+          {
+            icon: 20,
```
bvaughn, line 26: Do we have any automated checks that ensure the icon view boxes are all the same and what we expect?

packages/icons/scripts/icons.js
```diff
@@ -20,16 +17,41 @@ fetchImages({
...
+            replaceAttrValues: {
+              black: "currentColor",
+              "#383838": "currentColor",
+              "#38383D": "currentColor",
```
bvaughn, line 30: Ooo, this is interesting. Too bad svgr doesn't support pattern matching here for hex codes. Maybe we can add a verification step into our script that looks over the generated TypeScript and warns if there are any hard-coded colors that we missed?

bvaughn (approved):
Love it.

packages/icons/icons.stories.tsx
souporserious: Oops, yeah I'll get this hooked up! Was trying to push it out too quick last night 😅

packages/icons/scripts/icons.js
souporserious: No, but this is a great idea! I'll get this in a follow-up PR and add some better logs and error handling all around.

packages/icons/scripts/icons.js
souporserious: I'm moving from svgr to svgo and some simple string parsing so it will just use what's declared in Figma now, this way we can do a regex for these 😎. I also cleaned them up in the Figma file since they were a little messy.

packages/icons/scripts/icons.js
```diff
@@ -1,35 +1,58 @@
...
+    )
+  );
+  const jsxSvgs = await Promise.all(optimizedSvgs.map(svg => svgToJsx(svg.data)));
+  const allIconExports = ['import { SVGProps } from "react"']
+    .concat(
+      svgs.map((svg, index) => {
+        const name = `${pascalCase(svg.name)}Icon`;
+        const jsx = jsxSvgs[index]
+          /** Spread props for controlling the svg element. */
+          .replace('">\n', '" {...props}>\n');
```
souporserious, line 43: In the future, we can add babel or some simple AST parsing to do linting and more reliable transformations like this.

packages/icons/scripts/icons.js
souporserious: Didn't end up needing regex here since svgo takes care of removing fill so we can control it at the top-level.

souporserious:
@jonbell-lot23 and @bvaughn pinging again for another review since I reworked a bunch 😄. I ran into an issue today where the Figma API was down so I added an option for us to export svgs manually and use the same script for creating components if that occurs. I'm also checking svgs into source now which gives us the added benefit of screenshot diffing in GitHub 😎.

jonbell-lot23 (approved):
Don't wait on me -- I saw what it did yesterday and understand the changes you're describing here so that's good enough from me for now

bvaughn (approved):
Cool 😎

PR #6818, opened 2022-05-18
First attempt at refetching console messages

jcmorrow:
There's a long comment in here about the work left to be done.

Will close https://github.com/RecordReplay/customer-support/issues/67 once it is unblocked by https://github.com/RecordReplay/backend/issues/5622

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -55,13 +55,8 @@ function now(): number {
   return Date.now();
 }
 
-let thread: typeof ThreadFrontType | undefined;
 export const refetchDataForTimeRange = (focusRegion: FocusRegion): UIThunkAction => {
-  return async (dispatch, getState) => {
-    if (!thread) {
-      return;
-    }
-
+  return async (dispatch, getState, { ThreadFront }) => {
     // Technically, we only need to do this in *some* circumstances.
     // I think the cases where we need to are:
     // - We are enlarging or moving the focus zone beyond what we have most recently loaded.
@@ -85,14 +80,14 @@ export const refetchDataForTimeRange = (focusRegion: FocusRegion): UIThunkAction
     // without refetching*.
     // BTW - we don't have to figure out *all* of this right now :)
     dispatch(clearMessages());
-    const endpoint = (await sendMessage("Session.getEndpoint", {}, thread.sessionId!)).endpoint;
+    const endpoint = (await sendMessage("Session.getEndpoint", {}, ThreadFront.sessionId!)).endpoint;
     const beginning = (
       await sendMessage(
         "Session.getPointNearTime",
         {
           time: focusRegion.startTime,
         },
-        thread.sessionId!
+        ThreadFront.sessionId!
       )
     ).point as TimeStampedPoint;
     // @ts-ignore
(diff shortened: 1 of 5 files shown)


bvaughn:
Just to leave a comment trail about our video chat earlier 👍🏼  I'm going to take over this branch.

src/ui/actions/app.ts
```diff
@@ -53,6 +55,64 @@ function now(): number {
...
+    // any of this code. If we have overflow, and focus on, let's say the first
+    // 30 seconds of the recording, and we *don't* get overflow on the first
+    // thirty seconds, then we know that if the user were to change the window
+    // to the first 15 seconds, we don't need to refetch, and we also should not
+    // update the "most recently loaded" boundaries in the store, because that
+    // way the user could then focus on, for example, seconds 10 to 20 *also
+    // without refetching*.
+    // BTW - we don't have to figure out *all* of this right now :)
+    dispatch(clearMessages());
+    const endpoint = (await sendMessage("Session.getEndpoint", {}, ThreadFront.sessionId!)).endpoint;
```
markerikson, line 83: :speech_balloon: two quick thoughts here:

- It feels like `endpoint` is something that could be fetched once, on app startup, and we wouldn't have to fetch it again here
- Can the `getPointNearTime` requests be made in parallel instead of serial?  Probably not a big deal in terms of execution time, but doesn't _seem_ like we'd need to wait for the first to complete before requesting the second

src/ui/actions/app.ts
```diff
@@ -53,6 +55,61 @@ function now(): number {
...
+    // thirty seconds, then we know that if the user were to change the window
+    // to the first 15 seconds, we don't need to refetch, and we also should not
+    // update the "most recently loaded" boundaries in the store, because that
+    // way the user could then focus on, for example, seconds 10 to 20 *also
+    // without refetching*.
+    // BTW - we don't have to figure out *all* of this right now :)
+    dispatch(clearMessages());
+
+    // TODO [bvaughn] This value is probably pretty lightweight _but_ we could cache it for the whole session.
+    const endpoint = (await sendMessage("Session.getEndpoint", {}, ThreadFront.sessionId!))
```
bvaughn, line 87: Oof, my most recent commit clobber the endpoint check. I'll re-add that. Just ignore it for now. :)

bvaughn:
https://github.com/RecordReplay/backend/pull/5635 has been deployed to production but I'm still seeing failures for `Console.findMessagesInRange` :(

bvaughn:
I am no longer getting the error reported in https://github.com/RecordReplay/backend/issues/5653:
> Internal error

Now I am getting a different error:
> The linker version used to make this recording does not support this action

_But_ I made a new recording (with the latest Replay) and the request seems to be working okay for it!

https://app.replay.io/recording/react-window-lots-of-scrolling--1232b3d1-c368-439a-a8fb-e63deb8b0d64

src/ui/actions/app.ts
```diff
@@ -53,6 +55,79 @@ function now(): number {
...
+    // 30 seconds of the recording, and we *don't* get overflow on the first
+    // thirty seconds, then we know that if the user were to change the window
+    // to the first 15 seconds, we don't need to refetch, and we also should not
+    // update the "most recently loaded" boundaries in the store, because that
+    // way the user could then focus on, for example, seconds 10 to 20 *also
+    // without refetching*.
+    // BTW - we don't have to figure out *all* of this right now :)
+    dispatch(clearMessages());
+
+    // TODO [bvaughn] This value is probably pretty lightweight _but_ we could cache it for the whole session.
```
jcmorrow, line 86: I'm pretty sure we have this in the store actually, but I suspect that there could be a situation where we don't have it *yet* when this code runs 🤷

Issue #6590, opened 2022-04-28
Soft Focus


jcmorrow:
### Background

We currently support Focus Mode. Now, we *think* of focus mode as one thing, but in fact, it is two things: Focus mode tells the backend to load and unload particular regions of the recording using [Session.loadRegion](https://static.replay.io/protocol/tot/Session/#method-loadRegion) and [Session.unloadRegion](https://static.replay.io/protocol/tot/Session/#method-unloadRegion). This is how we think of what focus mode *is*, but there's actually some nuance here. I don't want to go into the full details because I've already [written them up elsewhere](https://www.notion.so/replayio/Loading-Region-Problems-b8dc0d03c49840f8b504e967c7ea5778), but the short version is: unloading some regions is not enough for millisecond control of your focus region, since regions are usually about half a second in length. Instead, we have to *pretend* to unload stuff that falls outside of our focus window (but still falls inside of the loaded region). I am calling this approach Soft Focus (for the 📷  nerds). Why is this important? Well, it's important for two reasons:

1) The combination of Soft Focus and region unloading can result in inconsistent results from the debugger (see the linked explainer above)
2) If point 1 were not true, if instead we never needed to actually unload sections of the recording, and instead we could just *pretend* to unload them, well then we should do that!

### Todos

- [x] add a feature flag that turns all focus into soft-focus? Some stuff will be broken but that's ok, I want to see how much is broken and what else we need to fix besides the stuff on this list.
- [x] allow some protocol methods which currently do not allow for filtering by time to filter by time
	- [x] `getHitCounts`
	- [x] `Console.findMessages`
- [x] https://github.com/RecordReplay/devtools/issues/7036
- [ ] Produce a much better `getPointNearTime` which will permit a wider release for soft focus

We need to do #7036 for Friday, but the `getPointNearTime` can be a follow-on.

jasonLaster:
Love the branding!

re hit counts: what are your concerns with continuing to use `getHitCounts` as the default?

jcmorrow:
> what are your concerns with continuing to use getHitCounts as the default?

Well, this just requires doing the work of allowing the endpoint to accept begin and end times. And that might end up being more involved than it sounds, because even after the recent position/location cache improvement, I don't think it's possible to say with certainty the hit counts in an arbitrary span without doing a full findAnalysisPoints of the bookend regions.

jcmorrow:
[...]

Issue #6576, opened 2022-04-27
Failing analyses should re-run when loading is complete for a different subset of regions


jcmorrow:
When an analysis errors because it results in too many points, two important things need to happen:

- If we focus, we need to re-run that analysis, because now it might not result in too many points
- We need to tell the user that this analysis is no longer loading and is in an error state, and they should probably use focus to try and get a narrower working window

It's a little hard to tell from this replay that the log message for "Too many points" is indeed linked to the request to run the analysis, so I created https://github.com/RecordReplay/devtools/pull/6575 to make sure that we are also logging the protocol request ID. Using that locally I confirmed that this is what is happening here.

https://app.replay.io/recording/analysis-never-moves-beyond-loading--ff87d6b7-bebd-4c6b-8806-6a4b6e478bb5

jcmorrow:
I think that #6591 is just a better, more detailed version of this card 😄

PR #6819, opened 2022-05-18
Rerun analyses and clear currentPause when needed

jcmorrow:
I started down [a long road](https://github.com/RecordReplay/devtools/compare/jm-analyses-reducer?expand=1), understanding all of the ways that we should change breakpoints/analyses to make them more stable and compatible with #6590. But a little ways down that road I saw some cheap and quick wins, so while I have uhhh... [kind of a lot written about where to go next](https://kinopio.club/a-user-walks-into-a-print-statement-Ubk9AJUa-o3vxPcxySACY), this at least gets us at a dumb place where:

- We will not error when we were at a Pause that now falls in an unloaded region
- We will refetch analyses when we select a new focus region (see the docs linked above about future improvements on this, it is super naive)

Is a slightly dirty (but viable) fix for #6576

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -20,7 +20,7 @@ import { openQuickOpen } from "devtools/client/debugger/src/actions/quick-open";
 import { getRecordingId } from "ui/utils/recording";
 import { prefs } from "devtools/client/debugger/src/utils/prefs";
 import { shallowEqual } from "devtools/client/debugger/src/utils/resource/compare";
-import type { ThreadFront as ThreadFrontType } from "protocol/thread";
+import { ThreadFront as ThreadFrontType } from "protocol/thread";
 import { getShowVideoPanel } from "ui/reducers/layout";
 import { toggleFocusMode } from "./timeline";
 import { getTheme } from "ui/reducers/app";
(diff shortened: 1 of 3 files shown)


src/ui/actions/app.ts
```diff
@@ -20,7 +20,7 @@ import { openQuickOpen } from "devtools/client/debugger/src/actions/quick-open";
 import { getRecordingId } from "ui/utils/recording";
 import { prefs } from "devtools/client/debugger/src/utils/prefs";
 import { shallowEqual } from "devtools/client/debugger/src/utils/resource/compare";
-import type { ThreadFront as ThreadFrontType } from "protocol/thread";
+import { ThreadFront as ThreadFrontType } from "protocol/thread";
```
markerikson, line 23: :hand: Any reason this got changed to remove the `import type` ? I specifically wanted to avoid directly importing `ThreadFront` here if possible.

jcmorrow: Yep, I accidentally imported it, then realized I had ruined your work, and tried to reverse my mistake, but I must have not reverted the actual import! Sorry about that, this will not ship with this change for sure!

jcmorrow:
Whoops, this should be closed. It was done in https://github.com/replayio/devtools/pull/6980 I think

PR #6820, opened 2022-05-18
Don't show outdated graphics

### LINKED ISSUE #4289: Timeline inconsistency

### LINKED ISSUE #6546: Viewing later parts of loading recording shows wrong graphics
When the graphics from the backend haven't fully loaded, clicking on the timeline in later parts of the recording shows images from earlier parts, presumably the last loaded paint.  We should be showing a "loading" indicator or something instead.  I haven't been able to get a recording of this yet.

jcmorrow:
And also when seeking try to get the nearest pause point possible, combining the backend's knowledge of the timeline with the paints array. Also - disable playback if we have no paints here. This is better (I think) than having a broken video player that just never plays. But we should also do some other things that we've talked about: make the hover cursor a disabled sign when we don't have paints that far out yet, maybe gray out the play button if it's going to be impossible (so it doesn't just feel like a focus issue)?

Fixes #6546 
Fixes #4289

### DIFF
--- a/src/protocol/graphics.ts
+++ b/src/protocol/graphics.ts
@@ -236,6 +236,12 @@ export const Video = new VideoPlayer();
 let onRefreshGraphics: (canvas: Canvas) => void;
 export let hasAllPaintPoints = false;
 
+export const setHasAllPaintPoints = (newValue: boolean) => {
+  hasAllPaintPoints = newValue;
+};
+export const timeIsBeyondKnownPaints = (time: number) =>
+  !hasAllPaintPoints && gPaintPoints[gPaintPoints.length - 1].time < time;
+
 export function setupGraphics(store: UIStore) {
   onRefreshGraphics = (canvas: Canvas) => {
     store.dispatch(setCanvas(canvas));
@@ -254,6 +260,12 @@ export function setupGraphics(store: UIStore) {
 
     client.Session.findMouseEvents({}, sessionId);
     client.Session.addMouseEventsListener(({ events }) => onMouseEvents(events, store));
+
+    const recordingTarget = await ThreadFront.recordingTargetWaiter.promise;
+    if (recordingTarget === "node") {
+      // Make sure we never wait for any paints when trying to do things like playback
+      setHasAllPaintPoints(true);
+    }
 
     if (features.videoPlayback) {
       client.Graphics.getPlaybackVideo({}, sessionId);
@@ -293,6 +305,9 @@ export async function repaint(force = false) {
   if (recordingTarget === "node") {
     return;
   }
+  let graphicsFetched = false;
+  // Show a stalled message if the graphics have not fetched after half a second
+  setTimeout(() => !graphicsFetched && store.dispatch(setPlaybackStalled(true)), 500);
   const { mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
   const point = ThreadFront.currentPoint;
   await ThreadFront.ensureAllSources();
@@ -303,9 +318,6 @@ export async function repaint(force = false) {
   const pause = ThreadFront.currentPause;
   assert(pause, "no pause");
 
-  let graphicsFetched = false;
-  // Show a stalled message if the graphics have not fetched after half a second
-  setTimeout(() => !graphicsFetched && store.dispatch(setPlaybackStalled(true)), 500);
   const rv = await pause.repaintGraphics(force);
   graphicsFetched = true;
   store.dispatch(setPlaybackStalled(false));
@@ -394,7 +406,7 @@ export async function getGraphicsAtTime(
   forPlayback = false
 ): Promise<{ screen?: ScreenShot; mouse?: MouseAndClickPosition }> {
   const paintIndex = mostRecentIndex(gPaintPoints, time);
-  if (paintIndex === undefined) {
+  if (paintIndex === undefined || timeIsBeyondKnownPaints(time)) {
     // There are no graphics to paint here.
     return {};
   }
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -10,6 +10,7 @@ import {
   precacheScreenshots,
   snapTimeForPlayback,
   Video,
+  timeIsBeyondKnownPaints,
 } from "protocol/graphics";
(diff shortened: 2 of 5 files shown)


src/ui/actions/timeline.ts
```diff
@@ -256,34 +237,44 @@ export function seek(
     } else {
       ThreadFront.timeWarp(point, time, hasFrames);
     }
+    dispatch(setTimelineToTime(time));
```
hbenl, line 240: I don't think this is a good idea: `setTimelineToTime()` paints the graphics but `ThreadFront.timeWarp()` will emit a `paused` event and that will also trigger painting the graphics: 
https://github.com/RecordReplay/devtools/blob/6f3198df7f6720231e6c1a5a71265e0a7b1854f0/src/protocol/graphics.ts#L264-L288
Not only are we painting twice, in the worst case this may actually paint a recorded screenshot over a repainted one, which would be a bug.

src/ui/actions/timeline.ts
```diff
@@ -256,34 +237,44 @@ export function seek(
...
-      // so e.g. running forward could land at a point before the time itself.
-      // This could be fixed but doesn't seem worth worrying about for now.
-      dispatch(seek(event.point, targetTime, false));
+    if (Math.abs(pointNearTime.time - targetTime) > Math.abs(nearestEvent.time - targetTime)) {
+      dispatch(seek(nearestEvent.point, targetTime, false));
+    } else {
+      // I would prefer that we also use pointNearTime.time here, for accuracy,
+      // but it would be super annoying when it is off. Maybe when we have a
+      // more exact method.
+      dispatch(seek(pointNearTime.point, targetTime, false));
```
hbenl, line 263: I think we should try using `pointNearTime.time` here, otherwise we're not fixing #4289, we're just reducing the size of the lie in the Timeline. That lie has caused so much confusion that I consider it worse than showing the truth (i.e. the time that you're *really* paused at) to the user, even if the truth can also be confusing.
Likewise we should try using `nearestEvent.time` above.

src/ui/components/ProtocolViewer.tsx
```diff
@@ -40,7 +40,7 @@ type ChunkReducer = {
 };
 
 const chunkedRequests = (requests: RequestSummary[]): RequestSummaryChunk[] => {
```
bvaughn, line 42: This is an example of a thing I wish we had inline documentation for. What is a "chunked request"? What does it mean to be "chunked"?

Is it merging protocol requests that have certain overlapping properties into a single object with a count that signifies how many separate requests were merged into the single one?

src/ui/components/ProtocolViewer.tsx
```diff
@@ -74,7 +74,8 @@ const chunkedRequests = (requests: RequestSummary[]): RequestSummaryChunk[] => {
         startedAt: 0,
       },
     }
-  ).chunks;
+  );
+  return [...chunks.chunks, chunks.current];
```
bvaughn, line 78: So this is handling the fact that we don't initialize the `chunks` array with our first `current` and so each time we make a new `current` we push the _previous_ one to the array?

I wonder if we could write this a little more clearly by restructuring the code a bit?
```js
const mergedRequests = (requests: RequestSummary[]): RequestSummaryChunk[] => {
  let current = {
    count: 0,
    ids: [],
    method: "",
    pending: false,
    errored: false,
    startedAt: 0,
  };

  const merged = [current];

  requests.forEach((request: RequestSummary) => {
    if (
      current.method === fullMethod(request) &&
      current.pending === request.pending &&
      current.errored === request.errored
    ) {
      current.count++;
      current.ids.push(request.id);
    } else {
      current = {
        count: 1,
        ids: [request.id],
        errored: request.errored,
        method: fullMethod(request),
        pending: request.pending,
        startedAt: request.recordedAt,
      };

      merged.push(current);
    }
  });

  return merged;
};
```

This is a nit so feel free to ignore.

src/ui/actions/timeline.ts
jcmorrow: Oh! Interesting! This is also what I expected, and then I was watching the protocol monitor and it did *not* issue a request for a repaint (and the graphics did not update) so I added this 😢

src/ui/actions/session.ts
```diff
@@ -232,6 +232,8 @@ export function createSocket(
 
       // We don't want to show the non-dev version of the app for node replays.
       if (recordingTarget === "node") {
+        // Make sure we never wait for any paints when trying to do things like playback
+        setHasAllPaintPoints(true);
```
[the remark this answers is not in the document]
jcmorrow: I do really dislike that we do this check in multiple places, agreed, I'll add this to the graphics setup.

src/protocol/graphics.ts
```diff
@@ -394,7 +398,10 @@ export async function getGraphicsAtTime(
   forPlayback = false
 ): Promise<{ screen?: ScreenShot; mouse?: MouseAndClickPosition }> {
   const paintIndex = mostRecentIndex(gPaintPoints, time);
-  if (paintIndex === undefined) {
+  if (
+    paintIndex === undefined ||
+    (!hasAllPaintPoints && gPaintPoints[gPaintPoints.length - 1].time < time)
```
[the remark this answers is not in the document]
jcmorrow: Oh, and we initialize the array with 1 member, so there's no empty array possibility. I think in the situation that you described it is true that we might not have any more paints to load, but we can't possibly know that until the promise has resolved telling us that all of the paints are in, so I think it's impossible to fix that potential false positive?

src/ui/actions/timeline.ts
jcmorrow: We talked about this, since it's still possible for the point to be noticeably different than the time, we're going to continue to lie, for now, but I will make a backend issue for making `getPointNearTime` more accurate, and hopefully after that we can stop lying 🙃

Issue #6316, opened 2022-04-12
Show original classname in object previews


hbenl:
With RecordReplay/gecko-dev#778 we use the name of the prototype's constructor as an object's className, but that will use the *generated* name of the class. We should show the original classname instead or possibly even switch between original and generated classname depending on the user's choice of showing the original or generated source.

PR #6823, opened 2022-05-18
Show original classname in ObjectInspector (#6316)

hbenl:
I've put this behind a feature flag and disabled it by default because it may degrade ObjectInspector performance.
Needs RecordReplay/backend#5614 to be deployed first.

### DIFF
--- a/packages/protocol/scope-map-cache.ts
+++ b/packages/protocol/scope-map-cache.ts
@@ -1,0 +1,22 @@
+import { Location } from "@replayio/protocol";
+
+import { ThreadFront } from "./thread/thread";
+import { defer } from "./utils";
+
+export default class ScopeMapCache {
+  private cache = new Map<string, Promise<Record<string, string>>>();
+
+  getScopeMap(location: Location): Promise<Record<string, string>> {
+    const cacheKey = `${location.sourceId}:${location.line}:${location.column}`;
+    let promise = this.cache.get(cacheKey);
+    if (!promise) {
+      promise = (async () => {
+        const { client } = await import("./socket");
+        const { map } = await client.Debugger.getScopeMap({ location }, ThreadFront.sessionId!);
+        return map ? Object.fromEntries(map) : {};
+      })();
+      this.cache.set(cacheKey, promise);
+    }
+    return promise;
+  }
+}
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -42,6 +42,7 @@ import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
 
 import { MappedLocationCache } from "../mapped-location-cache";
+import ScopeMapCache from "../scope-map-cache";
 import { client, log } from "../socket";
 import { defer, assert, EventEmitter, ArrayMap } from "../utils";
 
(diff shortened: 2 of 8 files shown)


packages/protocol/scope-map-cache.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    if (this.cache.has(cacheKey)) {
+      return this.cache.get(cacheKey)!;
+    }
+
+    const { promise, resolve } = defer<Record<string, string>>();
+    this.cache.set(cacheKey, promise);
+    const { client } = await import("./socket");
+    const { map } = await client.Debugger.getScopeMap({ location }, ThreadFront.sessionId!);
+    const scopeMap: Record<string, string> = map ? Object.fromEntries(map) : {};
```
jcmorrow, line 20: Is `Record<string, string>` the correct type here? I would have expected it to be something like `{[key: string]: string}` or `Record<string, string | undefined>`?

jcmorrow: I also thought this, but it's not! `Record<A, B>` says "For every A I will have you back a B. I will never hand back a null or undefined, only a B". `{[key: string]: string}` says "I'm an object that only contains string keys and string values, but I don't necessarily have an entry for every possible key string, sometimes I will hand back `undefined`". At least, I'm 95% sure it works this way, since it bit me on a previous side project 😄

packages/protocol/scope-map-cache.ts
```diff
@@ -0,0 +1,24 @@
...
+    const cacheKey = `${location.sourceId}:${location.line}:${location.column}`;
+
+    if (this.cache.has(cacheKey)) {
+      return this.cache.get(cacheKey)!;
+    }
+
+    const { promise, resolve } = defer<Record<string, string>>();
+    this.cache.set(cacheKey, promise);
+    const { client } = await import("./socket");
+    const { map } = await client.Debugger.getScopeMap({ location }, ThreadFront.sessionId!);
```
loganfsmyth, line 19: We've really gotta harden the frontend around exceptions more. I'd also say this is a good example of why I'm generally against using the `defer` pattern, because if this did throw, the promise will never resolve or reject, it'll just hang forever. 

This function returns a promise, so we should just cache the promise directly, e.g.
```javascript
  async getScopeMap(location: Location): Promise<Record<string, string>> {
    const cacheKey = `${location.sourceId}:${location.line}:${location.column}`;
    let result = this.cache.has(cacheKey);
    if (!result) {
      result = (async () => {
        const { client } = await import("./socket");
        const { map } = await client.Debugger.getScopeMap({ location }, ThreadFront.sessionId!);
        const scopeMap: Record<string, string> = map ? Object.fromEntries(map) : {};
        return scopeMap;
      })();
      this.cache.set(cacheKey, result);
    }
    return result;
  }
```

packages/protocol/thread/thread.ts
```diff
@@ -1021,6 +1028,17 @@ class _ThreadFront {
...
   }
 
+  getGeneratedLocation(locations: MappedLocation) {
+    const sourceIds = new Set<SourceId>(locations.map(location => location.sourceId));
+    return locations.find(location => {
+      const generated = this.getGeneratedSourceIds(location.sourceId);
+      if (!generated) {
+        return true;
+      }
+      return !generated.some(generatedId => sourceIds.has(generatedId));
```
loganfsmyth, line 1038: This feels like it could use a comment because I'm finding this check a bit hard to follow.

loganfsmyth (approved):

PR #6832, opened 2022-05-18
Attempt to fix Redux DevTools UI crash bug

### LINKED ISSUE #6831: Redux DevTools crashes in some scenarios
**Description:**

In some cases, clicking the "Redux" tab causes the entire UI to explode.  Worse, it repeats itself the next time you refresh the page, because we persist the active timestamp and the last selected webconsole tab.

I don't know why this is taking down the entire UI, especially given that we have an `<ErrorBoundary>` around the app.

A partial workaround is to delete the `localStorage` pref that says "the Redux tab was last selected", but that's ugly.

**Steps to Reproduce:**

It's tricky and I don't have a full guaranteed set of steps.  However, Ryota Murakami did record this happening in https://app.replay.io/recording/break-app-after-replay-loaded-immediately--3d313379-eac1-4192-afc1-0bca3bdd614d .

As best as I can tell, it's because the logic is trying to limit how many actions show up in the RDT UI based on the current time, and it's possible to end up in a situation where an `"INIT_INSTANCE"` action has been dispatched, but no follow-on "here's some actions for that instance" actions fall in the same time range.  So, the RDT UI components think that "instance 1/1" is active, but there's no corresponding state values, and the `mapState` readout explodes.

I seem to be able to reproduce it by browsing to https://app.replay.io/recording/nsx-random-input--bfc7fe51-de07-43ab-82be-7189aad44d9e?point=6165852525064618446129693546514554&time=3540&hasFrames=false .  That timestamp is in between the "init" and "some actions" updates.

markerikson:
This PR :

- Attempts to fix the Redux DevTools crash bug that I and Ryota Murakami have seen
  - Removed attempt to only show "actions up to this point in time",  which was failing due to edge cases (instance was initialized, but the first "here's some state" action for the instance was never loaded in)
- Cranks up the "maxAge" value for how many actions will get held  in memory by the RDT store before compacting old ones

(Hopefully) fixes #6831 .

The bug can currently be reproduced in prod by opening up https://app.replay.io/recording/nsx-random-input--bfc7fe51-de07-43ab-82be-7189aad44d9e?point=6165852525064618446129693546514554&time=3540&hasFrames=false and selecting the "Redux" tab, at which point the UI should crash.  That particular timestamp is in between when the RDT data had an "instance" entry initialized, and when it actually sees the first action with data for that instance.

Opening up that same recording + params on the preview branch (make sure you're signed in and have the Redux DevTools setting enabled) _should_ work if this fix is good.

### DIFF
--- a/src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
@@ -75,35 +75,68 @@ export const ReduxDevToolsPanel = () => {
     const rootComponent = rootRef.current;
     const store = rootComponent.store!;
 
-    const matchingAnnotationsByTimeRange = reduxAnnotations.filter(annotation => {
-      return currentTimestamp != null && annotation.time < currentTimestamp;
+    /*
+
+     The Redux DevTools UI has its own Redux store with its own reducers.
+     The primary part that we care about is the slice reducer that manages 
+     which "store instances" are connected, and calculates the "lifted state" 
+     for each store (ie, "after 3 increment actions, the state is `{value: 3}`".
+     See [`reducers/instances.ts` in the Redux DevTools repo for the logic](https://github.com/reduxjs/redux-devtools/blob/b82de745928211cd9b7daa7a61b197ad9e11ec36/packages/redux-devtools-app/src/reducers/instances.ts).
+
+     The RDT internals first pass along an `INIT_INSTANCE` action to let the UI 
+     know that a given store instance exists, and then later pass along the serialized  
+     actions  for processing to update that instance's calculated state.
+
+     Internally, the `instances` reducer appears to have handling for a 
+     "default" state, and falls back to that value if no instances are connected.  
+     In my experimenting I found I needed to dispatch an action related to that 
+     as part of resetting the internal state.
+
+    */
+
+    const allInstanceIds: string[] = ["default"];
+
+    reduxAnnotations.forEach(annotation => {
+      // @ts-ignore ignore "this comparison never matches" error
+      if (annotation.action?.request?.type === "INIT_INSTANCE") {
+        // @ts-ignore-error TS is _really_ getting confused here
+        allInstanceIds.push(annotation.action.request.instanceId);
+      }
     });
 
-    if (!matchingAnnotationsByTimeRange.length) {
-      return;
-    }
-
     batch(() => {
+      // We may receive the Redux annotations in several groups over time.
+      // Rather than try to figure out which ones we have and haven't loaded into
+      // the RDT Redux store, we'll just try to reset its internal instance state entirely,
+      // and always reload all of the received annotation actions.
+
       // TODO THIS IS A TERRIBLE IDEA COME UP WITH SOMETHING BETTER AND YET THIS WORKS
-      // Seemingly reset the DevTools internal knowledge of actions,
-      // so we can re-send in all actions up to this point in the recording
+      // Seemingly reset the DevTools internal knowledge of actions.
+      // @ts-ignore this is legal to dispatch
       store.dispatch({
-        type: "devTools/UPDATE_STATE",
-        request: {
-          type: "LIFTED",
-          id: "default",
-        },
+        type: "socket/DISCONNECTED",
       });
 
-      // TODO Is this how we want to actually behave here?
-      // Update the store with the actions up to this point in time
-      matchingAnnotationsByTimeRange.forEach(annotation => {
+      // For each connection instance that we know will exist from annotations,
+      // re-create an empty state value in the instances data.
+      allInstanceIds.forEach(instanceId => {
+        store.dispatch({
+          type: "devTools/UPDATE_STATE",
+          request: {
+            type: "LIFTED",
+            id: instanceId,
+          },
+        });
+      });
+
+      // Originally I had this showing only actions up to the current pause point,
+      // but that led to a situation where there could be partly-initialized state and
+      // the RDT UI would end up crashing. Now, always load everything we've received.
+      reduxAnnotations.forEach(annotation => {
         store.dispatch(annotation.action);
       });
     });
-
-    // TODO This only gets updated when we actually _pause_, not during playback.
-  }, [ReduxDevToolsAppRoot, currentTimestamp, reduxAnnotations]);
+  }, [ReduxDevToolsAppRoot, reduxAnnotations]);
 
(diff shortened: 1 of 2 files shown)


markerikson:
Loading up the same replay in the preview branch appears to work (including after I turned on the Redux setting, selected the "Redux" tab, and reloaded the page):

https://devtools-git-bugfix-6831-rdt-crashes-recordreplay.vercel.app/recording/nsx-random-input--bfc7fe51-de07-43ab-82be-7189aad44d9e?point=6165852525064618446129693546514554&time=3540&hasFrames=false

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -75,35 +75,50 @@ export const ReduxDevToolsPanel = () => {
     const rootComponent = rootRef.current;
     const store = rootComponent.store!;
 
-    const matchingAnnotationsByTimeRange = reduxAnnotations.filter(annotation => {
-      return currentTimestamp != null && annotation.time < currentTimestamp;
+    const allInstanceIds: string[] = ["default"];
+
+    reduxAnnotations.forEach(annotation => {
+      annotation.action.request.instanceId;
```
bvaughn, line 81: What's this line doing? Seems like an accident?

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -75,35 +75,50 @@ export const ReduxDevToolsPanel = () => {
     const rootComponent = rootRef.current;
     const store = rootComponent.store!;
 
-    const matchingAnnotationsByTimeRange = reduxAnnotations.filter(annotation => {
-      return currentTimestamp != null && annotation.time < currentTimestamp;
+    const allInstanceIds: string[] = ["default"];
```
bvaughn, line 78: Could you leave a comment about what the significance of the "default" instance ID is?

It's not obvious to me what "default" or "INIT_INSTANCE" means really.

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -75,35 +75,50 @@ export const ReduxDevToolsPanel = () => {
...
+        });
       });
 
-      // TODO Is this how we want to actually behave here?
-      // Update the store with the actions up to this point in time
-      matchingAnnotationsByTimeRange.forEach(annotation => {
+      // Originally I had this showing only actions up to the current pause point,
+      // but that led to a situation where there could be partly-initialized state and
+      // the RDT UI would end up crashing. Now, always load everything we've received.
+      reduxAnnotations.forEach(annotation => {
```
bvaughn, line 117: Won't this double-dispatch the "INIT_INSTANCE" actions? Since we also dispatch them above? Or is that okay?

src/ui/components/SecondaryToolbox/redux-devtools/redux-annotations.ts
```diff
@@ -58,6 +58,12 @@ export const createDevtoolsAction = <S, A extends Action<unknown>>(
     action.request.instanceId = instanceId;
   }
 
+  // Crank up the max actions the RDT logic will process before compacting old ones
+  // This is usually 50 by default, but hey, let's be generous.  (We'll see if this
+  // ends up causing any memory issues or anything)
+  // @ts-ignore
+  request.maxAge = 500;
```
bvaughn, line 65: What happens when things are "compacted" ? Will it cause runtime errors or just prevent users from interacting with the "compacted" ones ?

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
markerikson: These are all internals of the Redux DevTools "instances" reducer.

The Redux DevTools UI has its own Redux store with its own reducers.  The primary part that we care about is the slice reducer that manages which "store instances" are connected, and calculates the "lifted state" for each store (ie, "after 3 increment actions, the state is `{value: 3}`".  See [`reducers/instances.ts` in the Redux DevTools repo for the logic](https://github.com/reduxjs/redux-devtools/blob/b82de745928211cd9b7daa7a61b197ad9e11ec36/packages/redux-devtools-app/src/reducers/instances.ts).

The RDT internals first pass along an `INIT_INSTANCE` action to let the UI know that a given store instance exists, and then later pass along the serialized actions for processing to update that instance's calculated state.

Internally, the `instances` reducer appears to have handling for a "default" state, and falls back to that value if no instances are connected.  In my experimenting I found I needed to dispatch an action related to that as part of resetting the internal state.

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
markerikson: Yep, shoulda removed that.

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
markerikson: I'm not actually _dispatching_ the `INIT_INSTANCE` actions above - just trying to identify them so I can ensure that we have an empty but valid initial set of state data for each one.

But, beyond that, the intent here is to fully reset the internals of the `instances` reducer every time this logic runs, so that we reload the actions data from scratch.

src/ui/components/SecondaryToolbox/redux-devtools/redux-annotations.ts
markerikson: Prevents the user from seeing the old ones.

When the `maxAge + 1`th action is dispatched, the `instances` reducer calculates the next state as usual, but consolidates the oldest action to effectively remove it from the selectable list.

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
bvaughn: > I'm not actually dispatching the INIT_INSTANCE actions above

Aren't you though? I'm referring to this:
```js
        store.dispatch({
          type: "devTools/UPDATE_STATE",
          request: {
            type: "LIFTED",
            id: instanceId,
          },
        });
```

bvaughn: Nevermind my previous comment. I see what you're saying now about the dispatch being different values, though I still don't understand _why_ it's needed 😁

bvaughn (approved):
I don't really understand a lot of what's going on in this PR TBH 😁

PR #6854, opened 2022-05-19
Tweaked protocol viewer UI

bvaughn:
https://www.loom.com/share/e31c583a16fe48d0a90f672058111aa8

### DIFF
--- a/src/ui/components/ProtocolViewer.tsx
+++ b/src/ui/components/ProtocolViewer.tsx
@@ -12,6 +12,8 @@ import {
   RequestSummary,
 } from "ui/reducers/protocolMessages";
 
+import styles from "./ProtocolViewer.module.css";
+
 const ReactJson = dynamic(() => import("react-json-view"), {
   ssr: false,
 });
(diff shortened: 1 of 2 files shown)


src/ui/components/ProtocolViewer.tsx
```diff
@@ -123,38 +131,54 @@ const ProtocolViewer = () => {
...
-          >
-            <span>
-              {chunk.method}
-              {chunk.count > 1 ? `(${chunk.count})` : null}
-            </span>
-            <span>{msAsMinutes(chunk.startedAt)}</span>
-          </div>
-        );
-      })}
+    <div className={styles.Container}>
```
jcmorrow, line 134: My only comment is: this could be a good place for an accordion 🤷‍♂️

jcmorrow (approved):

src/ui/components/ProtocolViewer.tsx
bvaughn: Oh! That is a fantastic idea.

bvaughn: We seem to have at least **four** accordion components in the devtools repo 😁  

The first two I've looked at don't support dynamic children (e.g. rendering with only the requests pane at first and adding details panes later). They will throw if you do that. So they wouldn't work very well for this case.

I think I'll leave that out for now 😁

bvaughn:
Btw @jcmorrow– as I look at the `chunkedRequests` method again– I'm wondering what's the purpose of the default chunk we push into the accumulated array? Given that its' method is "" – do we expect any protocol requests to match it? I would assume we _don't_ and we'll just end up with an empty chunk at the top of every array.

I'm going to try removing it (and updating the reducer to account for that).

Issue #6591, opened 2022-04-28
Restructure how we interact with and store results and errors from the Analysis API


jcmorrow:
Problems that are hard to solve with our current state shape:

### Context for loading and reloading Analyses

- Right now, analyses don't always get re-run upon loading changes, when the loading/focus state transition from one state to another, sometimes that should trigger re-runs.
- This is probably not going to be *that* common of a requirement, because, for example, if an analysis gets run *once* at the findPoints level, we will know the correct points for the entire recording and every possible focus region of the recording, and if the analysis has run successfully, we will also know the evaluation at those points, so we will not need to requery until the user changes the expression)
  - A note about points vs location: right now we always run analyses by adding the location of the print statement, but we *could* add a subset of points when we know there will be too many to run a full analysis. This is not a top priority, but it will be a requirement for whatever comes out of https://github.com/RecordReplay/design/issues/36 and should be the kind of operation that our changes will support in the future.
- (In a lot of ways, this is going to line up with other work that we're doing under the umbrella of #6590 , because the loading and focus transitions that would cause analyses to re-run, will also trigger other data refetching)

# Tasks which are blocked by this

- [ ] #6851
- [x] #7033 
- [x] #7036

---

### Context for the next cards

- Related to loading states, there is a pretty specific series of lifecycle events for analyses, (some are documented in https://static.replay.io/protocol/tot/Analysis/), and I've tried to capture some of this in an xState machine:

![CleanShot 2022-05-19 at 10 02 31](https://user-images.githubusercontent.com/5903784/169357104-f03fde97-388f-418e-8da0-b1b4cb94070c.png)

[Link](https://stately.ai/viz/9f4a5f8c-4d3c-4baf-aa8c-a483f108771f)
- Wherever we end up in terms of storing analysis states, it should probably roughly align with that state machine
- One way to think about the lifecycle of an analysis, is that it can either be a "partial" analysis, or a "full" analysis, depending on which steps it gets to.
- All analyses will:
  - Be created
  - have something `add`ed to it (you can see which things you can add by looking at the methods inside of the Analysis domain which start with the word `add`.
  - have `findPoints` called, which can error or result in a list of points being emitted, this is a "partial" analysis
  - (maybe) have `runAnalysis` called, which will result in a fork at each of the points and an evaluation of whatever mapper was given during creation. If an analysis gets to this step, it should be considered "full". (fair warning: I find these terms useful, but they are not official).
  - All analyses will also get `release`d after this.
[...]

PR #6857, opened 2022-05-19
Migrate `client/commands` to TS

markerikson:
This is a precursor to work for #6591 .  I was browsing the code to see where some of the analysis logic is used, and kept tracing through `client/commands.js`, couldn't see what the args were or what code was used, and decided to migrate it.

This PR:

- Migrates `client/commands.js` to TS
  - Actually added types to client/commands
  - Deleted many functions that appear to be entirely dead (anecdotally confirmed by Jason)
  - Removed watchpoints logic in Scopes.js that appears to be dead
  - Removed irrelevant `setupCommands` function
- Migrates `client/index.js` to TS

### DIFF
--- a/src/ui/components/Comments/CommentMarker.tsx
+++ b/src/ui/components/Comments/CommentMarker.tsx
@@ -36,7 +36,10 @@ class CommentMarker extends React.Component<CommentMarkerProps> {
     return comments[index];
   }
 
-  onClick = () => {
+  onClick = (e: React.MouseEvent) => {
+    // This should not count as a click on the timeline, which would seek to a
+    // particular time.
+    e.stopPropagation();
     const { comment, seekToComment } = this.props;
     trackEvent("timeline.comment_select");
     seekToComment(comment);


markerikson:
All the E2E tests have failed, but this seems to be something that's occurring in other PRs today like #6854 .

hbenl (changes requested):
The reason for the e2e test failures is that you removed the resume/stepIn/stepOver/stepOut/rewind/reverseStepOver commands which are kinda central to a debugger.

> All the E2E tests have failed, but this seems to be something that's occurring in other PRs today like #6854 .

In that PR I only see one test failing, unlike this PR. Please don't ignore e2e test failures anymore, the tests have become more stable and since they have recordings (which are easily accessible from the [Summary page](https://github.com/RecordReplay/devtools/actions/runs/2354674729) for the "Run test suites" action) it's easy to investigate them.

markerikson:
Hmm. I did remove those functions, but based on my initial searches only the _action creators_ with the same names were used.

But yeah, let me go back and review the changes.

hbenl:
I *think* this is where they are used:
https://github.com/RecordReplay/devtools/blob/82a00fe58eb4421f4fef071cb8365b82adf2bd0e/src/devtools/client/debugger/src/actions/pause/commands.js#L43

markerikson:
AHHH, I think I see the problem.

Yeah, these functions aren't used _in the app code_, but they _are_ exposed as globals and used in some of the test harness:

```js
  getVarsForTests() {
    assert(this.isReady, "DebuggerPanel not ready");
    return {
      store: this._store,
      selectors: this._selectors,
      actions: this._actions,
      client: this._client,
    };
  }

// snip
const dbgSelectors = {};
for (const [name, method] of Object.entries(dbg.selectors)) {
  dbgSelectors[name] = (...args) => method(dbg.store.getState(), ...args);
}
```

so yeah, I was right about the UI not using these, but apparently these are critical for tests.

Oooo-kay, reverting those changes...

hbenl:
> so yeah, I was right about the UI not using these, but apparently these are critical for tests.

No, they're also used by the UI: when you view a recording with the preview deployment of this PR, the stepping buttons in the debugger are broken.

markerikson:
@hbenl oh drat, you may be right there as well.  Lovely.

(also, there were other test failures yesterday due to Node 16 getting rolled out until Brian H reverted that, which is part of what was confusing me here)

markerikson:
okay, gonna discard all the changes to `commands.ts` and start over, with just the TS conversion part

markerikson:
Well, it's _less_ broken. 3 E2E tests pass, 2 fail, and I think Holger said that test `breakpoints-07.js` has been failing due to #6870 .

PR #6859, opened 2022-05-19
Stop propagation of the click on CommentMarker

### LINKED ISSUE #6858: Clicking on a comment marker does not mark that comment as selected
I suspect this has something to do with how we detect clicks on Comment Markers vs. the raw timeline. We should probably have a Comment Marker onClick handler that does not bubble.

https://app.replay.io/recording/clicking-to-seek-to-comment-does-not-work--45798e58-0038-4b27-8835-34b3c2949f81

jcmorrow:
Clicking on a comment marker should seek *to that comment*, not to the
time that that comment is sitting on. This fix was pretty easy and is
not too crazy I think, but it might be better to resolve this difference
by changing the DOM hierarchy?

Fixes #6858

### DIFF
--- a/src/ui/components/Comments/CommentMarker.tsx
+++ b/src/ui/components/Comments/CommentMarker.tsx
@@ -36,7 +36,10 @@ class CommentMarker extends React.Component<CommentMarkerProps> {
     return comments[index];
   }
 
-  onClick = () => {
+  onClick = (e: React.MouseEvent) => {
+    // This should not count as a click on the timeline, which would seek to a
+    // particular time.
+    e.stopPropagation();
     const { comment, seekToComment } = this.props;
     trackEvent("timeline.comment_select");
     seekToComment(comment);


markerikson (approved):

bvaughn (approved):

bvaughn:
> I think, but it might be better to resolve this difference by changing the DOM hierarchy?

I am curious to hear what you have in mind!

jaril (approved):

jcmorrow:
> > I think, but it might be better to resolve this difference by changing the DOM hierarchy?
> 
> I am curious to hear what you have in mind!

My main thought here is just that the timeline itself (the blue bar, not the whole `Timeline/index.tsx` component does not *contain* the comment marker. So it would make sense to me if the blue bar and the markers on the blue bar were siblings, rather than having the markers be children, and then the markers sit above the timeline because of `z-index`, so you automatically get the right behavior, `onClick` for the marker goes to it, `onClick` for the timeline goes to it. But I'm a little afraid that change might unravel into something larger, so I've been sticking with this for now.

PR #6861, opened 2022-05-19
Protocol Viewer 2.0 🎉

bvaughn:
Loom demo: https://www.loom.com/share/05be793f207b4dc984892fb638fa8d88

- [x] Added filter text and clear support to Protocol Viewer
- [x] Add duration tracking (and show as colored chips)
- [x] Improve performance (replace arrays with inline `.find` with maps, memoize/cache, limit loading of request/error results to 10 by default)
- [x] ~~Allow "merged" chunks to be expanded~~ (The details pane and accordion is sufficient for this.)

### DIFF
--- a/src/ui/components/ProtocolViewer.tsx
+++ b/src/ui/components/ProtocolViewer.tsx
@@ -1,13 +1,21 @@
 import { filter, padStart } from "lodash";
 import dynamic from "next/dynamic";
 import { CommandResponse } from "protocol/socket";
-import React, { MutableRefObject, useLayoutEffect, useMemo, useRef, useState } from "react";
+import React, {
+  MutableRefObject,
+  useDeferredValue,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import { useSelector } from "react-redux";
 import Icon from "ui/components/shared/Icon";
 import { getTheme } from "ui/reducers/app";
 import {
   getFullRequestDetails,
-  getProtocolRequests,
+  getProtocolRequestsMap,
+  getProtocolResponseMap,
   Recorded,
   RequestSummary,
 } from "ui/reducers/protocolMessages";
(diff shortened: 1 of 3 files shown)


src/ui/reducers/protocolMessages.ts
```diff
@@ -20,47 +20,53 @@ export type RequestSummary = Omit<CommandRequest, "method"> & {
 
 export interface ProtocolMessagesState {
   events: (ProtocolEvent & Recorded)[];
-  requests: RequestSummary[];
-  responses: (CommandResponse & Recorded)[];
-  errors: (CommandResponse & Recorded)[];
+  idToResponseMap: { [id: number]: CommandResponse & Recorded };
+  idToRequestMap: { [id: number]: RequestSummary };
+  idToErrorMap: { [id: number]: CommandResponse & Recorded };
```
jcmorrow, line 25: As someone who did not do this when first writing this class, I understand the temptation, but especially now that we are grouping things by ID anyways, how would you feel about using entity adapters here?

src/ui/components/ProtocolViewer.tsx
```diff
@@ -115,49 +135,151 @@ const ProtocolRequestDetail = ({
...
+  const doesFilteredChunksContainSelectedChunk = useMemo(
+    () => selectedChunk !== null && filteredChunks.includes(selectedChunk),
+    [filteredChunks, selectedChunk]
+  );
+
+  const onFilterTextInputChange = (event: React.ChangeEvent) => {
+    setFilterText((event.currentTarget as HTMLInputElement).value);
+  };
+
+  const onClearButtonClick = () => {
```
jcmorrow, line 172: This pattern is striking to me. It seems clearer to have the logic (especially when it is 1 line) embedded on the element itself, rather than having to read the variable assignment, then remember that that variable exists and should be used later, and then try and remember what it does when it gets used eventually, many lines later. I totally get it if the action is complicated, but in that case, it seems like it should not be defined in the component at all. What are your considerations here? When does something go inline vs. defined outside?

jcmorrow (approved):
These are good improvements, I left some questions, take or leave them!

src/ui/reducers/protocolMessages.ts
bvaughn: I'm not familiar with that term. Looks like it's a Redux thing? Ah, looking at the [Redux docs](https://redux-toolkit.js.org/api/createEntityAdapter) this looks like what we use with Messages.

I don't think I understand the motivation to add that helper (in this case, or in general). Could you explain?

In this case, I think it would actually add more over-head since it _looks like_ it would result in 3 separate (mostly synced?) id arrays (one per adapter)?

P.S. Can't get past the fact that that API defines a method named `upsertOne` 🤨

src/ui/components/ProtocolViewer.tsx
bvaughn: I generally don't use inline functions because I think they're often less readable. Pulling them out makes it easier to add inline comments (something that's kind of awkward in JSX). It's pretty subjective though.

> and then try and remember what it does when it gets used eventually, many lines later.

IMO clear naming overcomes this, but I guess it's pretty subjective and YMMV.

> What are your considerations here? When does something go inline vs. defined outside?

It's a subjective preference 😄

bvaughn:
One final tweak this morning to improve render performance:
https://www.loom.com/share/dec6d606e173449f81426ca08b3217bc

Going to land once CI is done.

src/ui/reducers/protocolMessages.ts
bvaughn: Decided to skip this one for now. Happy to follow up on it though! Also really interested to hear your thoughts about the benefits.

jcmorrow: Yeah fair. Honestly I think @markerikson would do a better job of selling it than I probably would. I think in this particular case, an entity adapter probably wouldn't get us a ton *yet*, since the operations that we're doing are really simple, but it seems like the kind of thing that might be nice to have as the code base grows and the way that actions update the state gets more complicated. (I'm sure you've probably seen the redux updates that are like:

```
state: {
  ...state,
    [action.payload.topId]: {
      ...action.payload,
      thisOneTrait: action.payload.thisOtherTrait,
   }
}
```

And while I think Immer takes care of a lot of that entity adapters can also be helpful, particularly for batch updates.

src/ui/components/ProtocolViewer.tsx
jcmorrow: The inline comment thing makes sense. I think the naming would feel more natural if it was like `onClick={clearBefore(index)}`. The pattern of `<button onClick={onButtonClick}>` (or, even worse, `onClick={handleOnClick}` which I find border line maddening), feels like it's not adding much to me.

src/ui/reducers/protocolMessages.ts
markerikson: I'd need a better understanding of what the actual data being stored is to be able to say whether an entity adapter is the right fit here.  I do see the lookup tables being used, but from Brian's comment I'm not sure what their relationship is.

src/ui/components/ProtocolViewer.tsx
bvaughn: Yeah that's fair. In many cases, `onClick={() => clearBefore(index)}` is fine. When a function does more than 1-2 things, I prefer to pull it out, but it's subjective.

PR #6867, opened 2022-05-20
add ExperimentFeature: ScaleFontSize

ryota-murakami:
PR for #6667

I added font scale feature like following video.

https://user-images.githubusercontent.com/5501268/169492030-ca9c5912-136d-49fb-9ac5-e12b1c1e85ec.mov

This feature scale `source file and folder`, `Editor`, `console output` font size.  

As https://github.com/RecordReplay/devtools/discussions/6667#discussioncomment-2710552 @jasonLaster says I thought add 'small', 'middle', 'large' option initially, but support multiple option is difficult to current feature toggle based implementation because that treat boolean basically.
So I added simple font size scale as Experimental Feature.  

If this feature is well received, it could be improved by moving it from the "Experimental Feature" to the "Preferences" tab.  For example, to allow more detailed font size adjustment using a slider like the one below.

Screen Shot 2022-05-20 at 18 10 28


jonbell-lot23 (commented):
From a designer's standpoint, this looks good to me!

I'd probably change "Scale Font Size" to "Large Text." Scale made me wonder where the slider was, because I thought I could customise it by scaling. Whereas "Large Text" is more explicitly saying what you're opting into by clicking the checkbox.

As you mention, the next step once we see how this feels is to make this fully customisable. That's an important feature.

Thank you for your contribution!

ryota-murakami:
@jonbell-lot23 I replaced `Scale Font Size` to `Large Text`. 👨‍💻

https://user-images.githubusercontent.com/5501268/169620899-253e9c87-f98b-4aac-bf53-ebd4eb872488.mov

src/devtools/client/debugger/src/components/PrimaryPanes/index.tsx
```diff
@@ -4,7 +4,10 @@
 
 //
 
+import clsx from "clsx";
```
markerikson, line 7: :hand: Can you switch from `clsx` to `classnames`?  Looks like we normally use `classnames` for this kind of step, and the only reason we have `clsx` in the codebase is because it's a dependency for `react-draggable`.

markerikson (changes requested):
I think this just needs the one tweak for `classnames`, and otherwise seems reasonable.

src/devtools/client/debugger/src/components/PrimaryPanes/index.tsx
ryota-murakami: @markerikson Thank you for your review!
I switched from `clsx` to `classnames`. 😀

markerikson (approved):

markerikson:
Awright, let's get this in!

ryota-murakami:
@markerikson Thank you! 😄

PR #6876, opened 2022-05-20
Migrate `client/commands` to TS, attempt 2

markerikson:
I originally was working to migrate `client/commands.js` to TS in #6857 , but was running into problems with E2E tests.  I tried force-pushing it back to be equal with `master` to compare, but Github took that as a signal to close the PR entirely :(

~~So, restarting this with a tiny dummy commit off `master` to see what E2E does with no actual changes.~~

I've reworked the `commands.ts` changes so I don't remove _any_ functions. not even the ones that are absolutely 100% clearly dead because they use variables that are never initialized.


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -124,6 +124,8 @@ export default function LineNumberTooltip({
         if (lineNumber === lastHoveredLineNumber.current) {
           if (codeHeatMaps) {
             dispatch(
+              // TODO This is bad API design. Instead of passing in an `onFailure` callback,
+              // this action  should return a promise
```
jcmorrow, line 128: yeah, I wrote this code, there was a reason that using a promise here was really difficult, but I don't remember what it was off the top of my head.

jcmorrow (approved):

markerikson:
Okay, going to finally merge this!

PR #6882, opened 2022-05-20
Removed border radius from unicorn badge

bvaughn:
I can't repro the border radius issue you emailed me about but does this fix it?

### DIFF
--- a/src/ui/components/PrefixBadge.module.css
+++ b/src/ui/components/PrefixBadge.module.css
@@ -70,10 +70,6 @@
   background-size: auto 100%;
 }
 
-.Empty {
-  background-color: transparent;
-}
-
 /* Background fill colors */
 
 :root:global(.theme-light) .unicorn {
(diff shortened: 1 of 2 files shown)


jasonLaster:
Looks like we had a regression with the print statement badge color which used to be a darker grey
Screen Shot 2022-05-20 at 3 26 30 PM

Also the picker still has a border when no badge is selected. I think that is an SVG issue
Screen Shot 2022-05-20 at 3 26 21 PM

jonbell-lot23 (commented):
1. Horn looks fixed!
2. As Jason mentioned, the little grey capsule got lost so we need to fix that.
3. The border we have on the picker is by design until we get a better icon design. It looks wrong with no border, so we can keep this way for now.

bvaughn:
> Looks like we had a regression with the print statement badge color which used to be a darker grey

Oops! Thanks for catching this.

I think I misunderstood what the desired behavior was for that state. I'll push a fix before landing this change 😄

bvaughn:
Going to land this to verify that the Vercel deployments are picking up from `main` okay.

ryota-murakami:
@bvaughn Nice Work! 😄

PR #6902, opened 2022-05-21
Batched dependabot package updates:

bvaughn:
This PR bumps the following packages:
- [x] #6901
- [x] #6900
- [x] #6899
- [x] #6898
- [x] #6897
- [x] #6896
- [x] #6895
- [x] #6894
- [x] #6893
- [x] #6892
- [x] #6891

Hopefully GitHub will eventually fix this issue (https://github.com/github/roadmap/issues/148)

Note that an update to `@replayio/replay` was landed (#6890) separately and reverted (#6904) because it caused e2e test failures in CI. That's unexpected but I'm not going to dig in for now.


bvaughn:
Just a dependencies bump so I'm going to land it. I'll keep an eye out.

ryota-murakami:
@bvaughn Thanks!
If Dependbot group update issue doesn't look like going to fix, I think [Depfu](https://depfu.com/) is good alternate.  
I'm using it since 3 years ago, I had never any trouble.  
Zero configuration start and flexible setting that scheduled update, group update, reasonable stay version setting etc.. available.  
It can easy to try any public repo doesn't matter personal or organization, I think you'll like it. It's just an option, though.

bvaughn:
Thanks for the tip!

Issue #6472, opened 2022-04-20
Having hover text on network monitor table fields that may be cropped would be really nice


loganfsmyth:
I'm looking at the network monitor on a complicated site with lots of deep paths and long domains can be a bit painful.

For example:
<img width="707" alt="Screen Shot 2022-04-19 at 6 05 45 PM" src="https://user-images.githubusercontent.com/132260/164126539-b26d4fd1-2b25-4f1c-83b3-fbc886f6d226.png">

The domain was "3ce165edb7aa0fe3c3d2f297cc48839.safeframe.googlesyndication.com". I could select the entry and look at the general request data to get that, but that's not a trivial amount of work to process manually and visually.

Similarly for "Name" I could do that, or enable the "Path" column instead, but that makes skimming harder and can make the table overflow.

For both of these, my first reaction was to hover my mouse over the table cell, hoping for a longer descriptive version. We don't have it now but sure would be nice!

jonbell-lot23:
Going through UI polish tasks and adding comments:

Sounds like the best thing to do here is add a simple tooltip.

jonbell-lot23:
I tried to do this and couldn't figure it out. Waiting for eng to pick it up.

jaril:
I misjudged this. This sounds like it's fairly low effort, and it does sound like a bad debugging dead end to try to figure out what that domain value is but not be able to view the entire string.

Adding it to this sprint!

gideonred:
@markerikson to follow up on PR that was merged that may have fixed this.

markerikson:
Yep, should be resolved by #6905

PR #6905, opened 2022-05-21
Display long network rows values 

yeion7:
Issue #6472

I'm adding the full cell value as a title, this will display a build-in tooltip with the name (same approach that chrome use)

https://user-images.githubusercontent.com/9127504/169671642-5591b484-c8f2-4439-ad81-e922492c2fff.mp4

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestRow.tsx
+++ b/src/ui/components/NetworkMonitor/RequestRow.tsx
@@ -1,9 +1,9 @@
 import classNames from "classnames";
 import React, { useLayoutEffect, useRef } from "react";
+import { Row } from "react-table";
 
 import styles from "./RequestTable.module.css";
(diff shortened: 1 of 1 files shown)


src/ui/components/NetworkMonitor/RequestRow.tsx
```diff
@@ -1,9 +1,9 @@
 import classNames from "classnames";
 import React, { useLayoutEffect, useRef } from "react";
+import { Row } from "react-table";
```
yeion7, line 3: I'm not sure why vscode moved this import but I think it make sense since is grouping the library imports at the top

markerikson (approved):
I think it's safe to say that the E2E test failures are _not_ related to this :)

markerikson:
Thank you!

PR #6910, opened 2022-05-22
Dropdown/ContextMenu refactor

### LINKED ISSUE #6909: Dropdown interaction issues
**Description:**

Dropdown have a few issues with a11y,

* Focus doesn't return to the button on select item
* Can't close on scape
* Can't navigate on options with arrows keys

**Steps to Reproduce:**

Try to use a dropdown with keyboard 

**Expected Result:**

Users can use a dropdown only with the keyboard

**Additional Context:**

https://user-images.githubusercontent.com/9127504/169709488-dc6dcdf6-3ed6-4cec-b1c4-f2eee50cf73e.mp4

yeion7:
fix #6909 

I figured out that actually dropdown is using `@headlessui/react`, that have most of the required keyboard interactions build out-of-the-box, the issue is the way that component is being used

In this PR I'm proposing a new Dropdown API

```
<Dropdown trigger={<Icon ... />}>
  <DropdownItem>Option 1</DropdownItem>
  <DropdownItem>Option 1</DropdownItem>
</Dropdown>
```

* Dropdown supports keyboard interactions ([more](https://headlessui.dev/react/menu#keyboard-interaction))
* Dropdown list is render in a portal always
* All the Dropdown positioning is handle by `@popperjs/core`

This is removing `PortalDropdown`, and `shared/Dropdown`, in favor of `Library/LibraryDropdown`

https://user-images.githubusercontent.com/9127504/169709870-3697ab46-c13c-4089-8976-52402ec0d551.mp4


yeion7:
Four usages are failing in types checking (thanks TS)

Those are quite interesting, IMO in those cases we're trying to use a `ContextMenu` component, but hacking the `Dropdown` component to use it 

I think the best is to create (I'm not sure if exist) a `ContextMenu` component for these cases


* Network columns header
image

* Source gutter
image

* Console output
image

* Request row
image

yeion7:
> I think the best is to create (I'm not sure if exist) a ContextMenu component for these cases

Following this idea, I found that in all these cases, the `Dropdown` component is wrapped by `ContextMenu`, so, In deep they're context menus

I just added `ContextMenu` component to `Library`, for now I didn't change to much how it works, I just moved with the same implementation and I unified the logic with the one in `ui/components/ContextMenu`

### ContextMenu

```
<ContextMenu close={function} x={number} y={number}>
      <ContextMenuItem onClick={function}>Add comment</ContextMenuItem>
</ContextMenu>
```

yeion7:
I don't know how to fix the `trunk` linter issues 🤔

src/ui/components/NetworkMonitor/ColumnsDropdown.tsx
```diff
@@ -1,28 +0,0 @@
...
-  };
-
-  return (
-    <Dropdown>
-      {columns.map(column => (
-        <ColumnItem key={column.id} column={column} />
-      ))}
-    </Dropdown>
-  );
-}
```
jasonLaster, line 28: Nice!

jasonLaster (commented):
After skimming the code for 5 minutes, my primary takeaway is how much more consistent everything is. This is a huge help @yeion7

jasonLaster:
https://www.loom.com/share/982c007dc08642d79085d0a7a7e4b539

yeion7:
> https://www.loom.com/share/982c007dc08642d79085d0a7a7e4b539

Nice catch the shadow, I just removed

jaril:
@bvaughn added you as a reviewer 🙋‍♂️

bvaughn:
Skimmed the code _super_ quickly. Looked at the preview branch. Seems nice! :) I could take a closer look in the morning though

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -232,20 +231,18 @@ class ConsoleOutput extends React.Component<PropsFromRedux, State> {
...
+                <Icon filename="set-focus-start" className="mr-4 bg-iconColor" size="large" />
+                Set focus start
+              </>
+            </ContextMenuItem>
+            <ContextMenuItem onClick={this.setFocusEnd}>
+              <>
+                <Icon filename="set-focus-end" className="mr-4 bg-iconColor" size="large" />
+                Set focus end
+              </>
+            </ContextMenuItem>
```
bvaughn, line 245: Keyboard navigation doesn't work in this case, (or the one in Network monitor). Do you know why?

bvaughn (approved):
Skimmed again this morning and I think this looks good 👍🏼 

Noticed a couple of things while testing but I don't think they're blockers:
* The context menu options in both the Network and Console panels don't have keyboard arrow support.
* The context menu shown in the sources view is still the old one:
Screen Shot 2022-05-24 at 8 39 57 AM

Might be nice to update these but wouldn't need to block this PR for it.

bvaughn:
Going to hold off and hear @yeion7's thoughts on the above two things before landing but otherwise I think we're okay to move forward.

yeion7:
> The context menu options in both the Network and Console panels don't have keyboard arrow support.

this is expected, we're using only the `MenuItems` from `@headlessui/react`, this is kind of a hacky way to use that component, that component was build for Dropdowns and unfortunally the library doesn't expose the hooks or something to enabled the keyboard navigation for a single list, I'm thinking to work with that ContextMenu in a follow up PR. 

In this PR I focus to moved the implementation to a different component and not use the same for Context Menus and Dropdowns

> The context menu shown in the sources view is still the old one:

Yes, I'd like to have a better version of `ContextMenu` with a good component DX, and then migrate the old ones 


btw, thanks for the review

bvaughn:
Okay. This seems like a nice incremental improvement then :)

bvaughn:
This was reverted (#6941) because it caused an error:
image

Not sure why we didn't see this on the preview branch.

PR #6915, opened 2022-05-23
Console focus find messages in range

bvaughn:
Resolves https://github.com/RecordReplay/customer-support/issues/67

Test recordings:
* [successful-logpoint-02--db5e1246-520b-4e63-8211-3fa150780fba](https://devtools-ln29rwanc-recordreplay.vercel.app/recording/successful-logpoint-02--db5e1246-520b-4e63-8211-3fa150780fba)
* [recording-with-a-lot-of-console-messages--36fd857e-ca04-4155-a309-287501c99c14](https://devtools-ln29rwanc-recordreplay.vercel.app/recording/recording-with-a-lot-of-console-messages--36fd857e-ca04-4155-a309-287501c99c14)
* ~~too-many-console-messages--1c17b653-15ab-4460-bb6d-5ec27a09141c~~ (doesn't support new `findMessagesInRange` API)

- [x] When focus mode has been set, re-run analysis to find console messages in the new focus range.
- [x] Add "soft focus" support to avoid unnecessarily querying the backend when focus is narrowed (assuming no overflow)
- [x] Dry up console "wired message" factory code
- [x] Write tests

### Bugs
- [x] Fix regression in log point when focus window moves around.


src/ui/actions/app.ts
```diff
@@ -53,6 +58,107 @@ function now(): number {
...
+          value => new ValueFront(pause, value)
+        );
+      }
+
+      if (message.sourceId) {
+        message.sourceId = ThreadFront.getCorrespondingSourceIds(message.sourceId)[0];
+      }
+
+      dispatch(onConsoleMessage(wiredMessage));
+    });
```
bvaughn, line 156: This feels pretty clunky. Would be nice to DRY this sort of stuff up _at least_ as it would be easy to _forget_ to do one of these things (as I did at one point) which would lead to errors.

src/devtools/client/webconsole/actions/messages.ts
```diff
@@ -231,3 +239,128 @@ export const messagesClearEvaluations = messageEvaluationsCleared;
 export const messagesClearLogpoint = logpointMessagesCleared;
 export const messageOpen = messageOpened;
 export const messageClose = messageClosed;
+
+function isNewFocusRegionSubset(
```
markerikson, line 243: :speech_balloon: I like this, very readable!

src/devtools/client/webconsole/actions/messages.ts
```diff
@@ -231,3 +239,128 @@ export const messagesClearEvaluations = messageEvaluationsCleared;
...
+    // 1. The most recent time it requested data "overflowed" (too many messages to send them all), or
+    // 2. The new focus region is outside of the most recent region we fetched messages for.
+    //
+    // There are two things to note about the second bullet point above:
+    // 1. When devtools is first opened, there is no focused region.
+    //    This is equivalent to focusing on the entire timeline, so we often won't need to refetch messages when focusing for the first time.
+    // 2. We shouldn't compare the new focus region to the most recent focus region,
+    //    but rather to the most recent focus region that we fetched messages for (the entire timeline in many cases).
+    //    If we don't need to refetch after zooming in, then we won't need to refetch after zooming back out either,
+    //    (unless our fetches have overflowed at some point).
```
markerikson, line 279: :speech_balloon: ditto, love the explanation here - I think this will really help me a lot on my side

src/devtools/client/webconsole/actions/messages.ts
```diff
@@ -231,3 +239,128 @@ export const messagesClearEvaluations = messageEvaluationsCleared;
...
+      },
+      ThreadFront.sessionId!
+    );
+
+    // Store the result of the new analysis window for "soft focus" comparison next time.
+    dispatch(setLastFetchedForFocusRegion(focusRegion));
+    dispatch(setConsoleOverflowed(overflow === true));
+
+    // Copied from ThreadFront.findConsoleMessages():
+    // TODO [bvaughn] Would be nice if this shared code with ThreadFront.
```
markerikson, line 338: :question: is the issue that messages _normally_ come over the websocket via the standard "get some messages" request, and here we have an alternate "fetch messages" request that isn't used anywhere else, so we have to duplicate the processing?

src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -259,12 +265,6 @@ const messagesSlice = createSlice({
 
       return removeMessagesFromState(state as MessageState, removedIds);
     },
-    //  This is only here to force recalculation after filters are updated
```
markerikson, line 262: 🎉

src/devtools/client/webconsole/actions/messages.ts
bvaughn: Yup! And if we don't do the mapping of message values to `ValueFront`s, it can cause things to break later (see https://github.com/RecordReplay/devtools/pull/6556#issuecomment-1134621051)

src/devtools/client/webconsole/reducers/messages.ts
bvaughn: Yeah, happened to notice that this wasn't being referenced anywhere so ✂️  😁

src/devtools/client/webconsole/actions/messages.ts
jcmorrow: This is all great. As mentioned in standup today and in https://github.com/RecordReplay/devtools/issues/6591, this logic is the same for all resources that could overflow, so we will likely want to reuse this particular check elsewhere at some point.

bvaughn: Yeah, good call. Wasn't sure where to put this for now since it could apply to other resource types.

Maybe the answer is to move the subset helper method into `utils/timeline` so it can be referenced by other code that cares about such things, but keep the state (both overflow and `overflow` and `lastFetchedForFocusRegion`) in messages since they're very message-specific. (We might want to re-fetch some random resource even if we don't happen to need to refetch messages for example.)

jcmorrow: yep, that's exactly right. The logic for "given a new focus window, a most recently fetched window, and an overflow,  flag, should I fetch?" is the same across everything, but the "last fetched" and "overflow" portions are state that are unique to each resource that cares about this check.

src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -285,6 +282,20 @@ const messagesSlice = createSlice({
         state as MessageState
       );
     },
+    clearMessages(state) {
+      removeMessagesFromState(state as MessageState, state.messages.ids as string[]);
+      state.overflow = false;
+      state.messagesLoaded = false;
+    },
```
bvaughn, line 289: This looks unnecessarily slow, since it's doing iteration and individual deletion:
https://github.com/RecordReplay/devtools/blob/66136c37002ab85c152b477e41252351f012d959/src/devtools/client/webconsole/reducers/messages.ts#L443-L459

I think we can just clear and recreate both the id and entities object directly.

bvaughn:
Okay I think this is ready for review.

src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -285,6 +282,26 @@ const messagesSlice = createSlice({
...
+      state.visibleMessages = [];
+    },
+    setConsoleOverflowed(state, action: PayloadAction<boolean>) {
+      state.overflow = action.payload;
+    },
+    setLastFetchedForFocusRegion(state, action: PayloadAction<FocusRegion | null>) {
+      state.lastFetchedForFocusRegion = action.payload;
+    },
+    setMessagesLoaded(state, action: PayloadAction<boolean>) {
+      state.messagesLoaded = action.payload;
```
markerikson, line 303: :question: I see this got changed to take a boolean payload, but it doesn't look like we're dispatching `setMessagesLoaded(false)` anywhere - it gets set as `state.messagesLoaded = false` above.  Do we anticipate dispatching that later?

src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -259,12 +262,6 @@ const messagesSlice = createSlice({
 
       return removeMessagesFromState(state as MessageState, removedIds);
     },
-    //  This is only here to force recalculation after filters are updated
-    // TODO Find a way to rework `visibleMessages` as derived data and remove this
-    filterStateUpdated(state, action: PayloadAction<FiltersState>) {},
```
markerikson, line 264: :speech_balloon: Yeah, this was a holdover from when I was doing some early RTK reducer migration.  I had to temporarily dispatch an extra action to keep this reducer in sync with the `filtersReducer`, and eventually resolved it by moving the filters state inside of this slice.  Looks like I just forgot to remove `filterStateUpdated`, so good catch!

src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -285,6 +282,26 @@ const messagesSlice = createSlice({
         state as MessageState
       );
     },
+    clearMessages(state) {
+      // Note that we don't use removeMessagesFromState() in this case, because we're removing ALL messages.
+      // That method does extra work to handle the case where we are only removing SOME messages.
+      messagesAdapter.removeAll((state as MessageState).messages);
```
markerikson, line 288: :hand: Do we need to reset `state.logpointMessages` here too?

Tbh I'm still not entirely sure what that field is used for, and looking at my local copy it _seems_ like it's only being used within this file.  But it seems like a thing we'd want to clear out in here also.

jcmorrow (approved):
You're a champion. I was so daunted when first investigating how the heck we were going to deal with this. To see the theoretical solution turn into a working preview branch feels so good! 🙌

src/devtools/client/webconsole/reducers/messages.ts
bvaughn: Hmm, I'm actually not sure. I _think_ log points are local (in memory) as opposed to messages that are fetched from the server. Would we need to clear log points out before refetching messages? Maybe @jaril or @jcmorrow knows?

src/devtools/client/webconsole/reducers/messages.ts
bvaughn: Oh, that's fair. I was confused by the fact that we had an action with the same name as the state (`messagesLoaded` and `messagesLoaded`). I wanted to use the state in my "do we need to refetch?" check again so I didn't want them to clash, so I renamed it to `setMessagesLoaded` but then it seemed weird not to accept a value. I don't feel strongly about this though TBH>

src/devtools/client/webconsole/reducers/messages.ts
bvaughn: I just tried with and without clearing the `logpointMessages` and I think it's kind of a weird user experience either way. The log point either disappears or becomes kind of half broken when you change the focus.

I'm not sure I understand the purpose of this bit of message state. Looks like it's been around since Brian Hackett imported the code ~two years ago in fc08cf873c1844587d97895bf89174eca4bf511c.

bvaughn: I think it _is_ referenced in quite a few places though I'm not yet sure what it's used for. This is what removing it would look like, I think:
```diff
diff --git a/src/devtools/client/debugger/src/actions/breakpoints/logpoints.ts b/src/devtools/client/debugger/src/actions/breakpoints/logpoints.ts
index dbe648c4a..4c980328a 100644
--- a/src/devtools/client/debugger/src/actions/breakpoints/logpoints.ts
+++ b/src/devtools/client/debugger/src/actions/breakpoints/logpoints.ts
@@ -15,9 +15,6 @@ import { _removeBreakpoint, removeBreakpointOption } from "./modify";
 export function removeLogpointsInSource(cx: Context, source: Source): UIThunkAction {
   return async (dispatch, getState) => {
     const breakpoints = getLogpointsForSource(getState(), source.id);
-    for (const breakpoint of breakpoints) {
-      dispatch(removeLogpoint(cx, breakpoint));
-    }
     const requestedBreakpointLocations = getRequestedBreakpointLocations(getState());
     for (const location of Object.values(requestedBreakpointLocations)) {
       if (location.sourceId === source.id) {
@@ -29,11 +26,6 @@ export function removeLogpointsInSource(cx: Context, source: Source): UIThunkAct
 
 export function toggleLogpoint(cx: Context, line: number, bp?: Breakpoint): UIThunkAction {
   return dispatch => {
-    if (bp?.options.logValue) {
-      trackEvent("breakpoint.minus_click");
-      return dispatch(removeLogpoint(cx, bp));
-    }
-
     trackEvent("breakpoint.plus_click");
     return dispatch(addLogpoint(cx, line));
   };
diff --git a/src/devtools/client/debugger/src/client/commands.ts b/src/devtools/client/debugger/src/client/commands.ts
index ac391b5b1..9aff7bc89 100644
--- a/src/devtools/client/debugger/src/client/commands.ts
+++ b/src/devtools/client/debugger/src/client/commands.ts
@@ -15,7 +15,6 @@ import {
   newLogGroupId,
   setEventLogpoints,
   setExceptionLogpoint,
-  removeLogpoint,
 } from "protocol/logpoint";
 import { ThreadFront, createPrimitiveValueFront, ValueFront } from "protocol/thread";
 import { WiredNamedValue } from "protocol/thread/pause";
@@ -115,19 +114,11 @@ function maybeGenerateLogGroupId(options: InitialBreakpointOptions): FinalBreakp
   return options;
 }
 
-async function maybeClearLogpoint(location: SourceLocation) {
-  const bp = breakpoints[locationKey(location)];
-  if (bp && bp.options.logGroupId) {
-    removeLogpoint(bp.options.logGroupId);
-  }
-}
-
 function hasBreakpoint(location: SourceLocation) {
   return !!breakpoints[locationKey(location)];
 }
 
 function setBreakpoint(location: SourceLocation, options: InitialBreakpointOptions) {
-  maybeClearLogpoint(location);
   const finalOptions = maybeGenerateLogGroupId(options);
   breakpoints[locationKey(location)] = { location, options: finalOptions };
 
@@ -157,7 +148,6 @@ function setBreakpoint(location: SourceLocation, options: InitialBreakpointOptio
 }
 
 function removeBreakpoint(location: SourceLocation) {
-  maybeClearLogpoint(location);
   delete breakpoints[locationKey(location)];
 
   const { line, column, sourceUrl, sourceId } = location;
@@ -320,9 +310,6 @@ function setEventListenerBreakpoints(ids: string[]) {
 let gExceptionLogpointGroupId: string | null;
 
 function logExceptions(shouldLog: boolean) {
-  if (gExceptionLogpointGroupId) {
-    removeLogpoint(gExceptionLogpointGroupId);
-  }
   if (shouldLog) {
     gExceptionLogpointGroupId = newLogGroupId();
     setExceptionLogpoint(gExceptionLogpointGroupId);
diff --git a/src/devtools/client/debugger/src/components/SecondaryPanes/LogpointsPane.tsx b/src/devtools/client/debugger/src/components/SecondaryPanes/LogpointsPane.tsx
index e139342f4..693e3e860 100644
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/LogpointsPane.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/LogpointsPane.tsx
@@ -3,7 +3,7 @@ import { useDispatch, useSelector } from "react-redux";
 import { getLogpointSources } from "../../selectors/breakpointSources";
 import Breakpoints from "./Breakpoints";
 import MaterialIcon from "ui/components/shared/MaterialIcon";
-import { removeLogpoint, removeLogpointsInSource } from "../../actions/breakpoints/logpoints";
+import { removeLogpointsInSource } from "../../actions/breakpoints/logpoints";
 import { Context } from "../../selectors";
 import { Breakpoint, Source } from "../../reducers/types";
 
@@ -25,12 +25,8 @@ export default function LogpointsPane() {
       type="print-statement"
       emptyContent={emptyContent}
       breakpointSources={logpointSources}
-      onRemoveBreakpoint={(cx: Context, breakpoint: Breakpoint) =>
-        dispatch(removeLogpoint(cx, breakpoint))
-      }
-      onRemoveBreakpoints={(cx: Context, source: Source) =>
-        dispatch(removeLogpointsInSource(cx, source))
-      }
+      onRemoveBreakpoint={(cx: Context, breakpoint: Breakpoint) => {}}
+      onRemoveBreakpoints={(cx: Context, source: Source) => {}}
     />
   );
 }
diff --git a/src/devtools/client/webconsole/actions/messages.ts b/src/devtools/client/webconsole/actions/messages.ts
index ae555aa78..c69bb8237 100644
--- a/src/devtools/client/webconsole/actions/messages.ts
+++ b/src/devtools/client/webconsole/actions/messages.ts
@@ -21,7 +21,6 @@ import { isFocusRegionSubset } from "ui/utils/timeline";
 
 import {
   clearMessages,
-  logpointMessagesCleared,
   Message as InternalMessage,
   messagesAdded,
   messageClosed,
@@ -42,7 +41,6 @@ export function setupMessages(store: UIStore, ThreadFront: typeof ThreadFrontTyp
     store.dispatch(onLogpointLoading(logGroupId, point, time, location));
   LogpointHandlers.onResult = (logGroupId, point, time, location, pause, values) =>
     store.dispatch(onLogpointResult(logGroupId, point, time, location, pause, values));
-  LogpointHandlers.clearLogpoint = logGroupId => store.dispatch(messagesClearLogpoint(logGroupId));
   TestMessageHandlers.onTestMessage = msg => store.dispatch(onConsoleMessage(msg));
 
   ThreadFront.findConsoleMessages(
@@ -235,7 +233,6 @@ export function messagesAdd(
 }
 
 export const messagesClearEvaluations = messageEvaluationsCleared;
-export const messagesClearLogpoint = logpointMessagesCleared;
 export const messageOpen = messageOpened;
 export const messageClose = messageClosed;
 
diff --git a/src/devtools/client/webconsole/reducers/messages.ts b/src/devtools/client/webconsole/reducers/messages.ts
index 7689e9b33..c82809c31 100644
--- a/src/devtools/client/webconsole/reducers/messages.ts
+++ b/src/devtools/client/webconsole/reducers/messages.ts
@@ -95,11 +95,6 @@ export interface Message {
   };
 }
 
-interface LogpointMessageEntry {
-  messageId: MessageId;
-  key: string;
-}
-
 interface FilteredMessagesCount {
   debug: number;
   info: number;
@@ -141,8 +136,6 @@ export interface MessageState {
   filteredMessagesCount: FilteredMessagesCount;
   /** List of message IDs that are opened */
   messagesUiById: MessageId[];
-  /** Lookup table of which messages are "logpoint" messages */
-  logpointMessages: EntityState<LogpointMessageEntry>;
   /** Set of logpoint IDs that have been removed */
   removedLogpointIds: MessageId[];
   /** Any execution point we are currently paused at, when replaying. */
@@ -171,9 +164,6 @@ export const appendToHistory = (command: Command, history: Command[]): Command[]
 };
 
 export const messagesAdapter = createEntityAdapter<Message>();
-const logpointMessagesAdapter = createEntityAdapter<LogpointMessageEntry>({
-  selectId: entry => entry.key,
-});
 
 export const syncInitialMessageState = (overrides: Partial<MessageState> = {}): MessageState => {
   // Realistically, we only expect filters and commandHistory
@@ -190,7 +180,6 @@ export const syncInitialMessageState = (overrides: Partial<MessageState> = {}):
         messagesUiById: [],
         commandHistory: [],
         exceptionLogpointError: null,
-        logpointMessages: logpointMessagesAdapter.getInitialState(),
         removedLogpointIds: [],
         pausedExecutionPoint: null,
         pausedExecutionPointTime: 0,
@@ -248,20 +237,6 @@ const messagesSlice = createSlice({
 
       return removeMessagesFromState(state as MessageState, removedIds);
     },
-    logpointMessagesCleared(state, action: PayloadAction<string>) {
-      const logpointId = action.payload;
-      const removedIds = [];
-
-      for (const [id, message] of Object.entries(state.messages.entities)) {
-        if (message!.logpointId == logpointId) {
-          removedIds.push(id);
-        }
-      }
-
-      state.removedLogpointIds = Array.from(new Set([...state.removedLogpointIds, logpointId]));
-
-      return removeMessagesFromState(state as MessageState, removedIds);
-    },
     exceptionLogpointErrorCleared(state) {
       state.exceptionLogpointError = null;
     },
@@ -323,7 +298,6 @@ const messagesSlice = createSlice({
 
 export const {
   clearMessages,
-  logpointMessagesCleared,
   messageClosed,
   messageEvaluationsCleared,
   messageOpened,
@@ -366,22 +340,6 @@ function addMessage(newMessage: Message, state: MessageState): MessageState {
     state.hasExecutionPoints = true;
   }
 
-  // When replaying, we might get two messages with the same execution point and
-  // logpoint ID. In this case the first message is provisional and should be
-  // removed.
-  const removedIds = [];
-  if (newMessage.logpointId || (newMessage.evalId && newMessage.type === MESSAGE_TYPE.RESULT)) {
-    const key = `${newMessage.logpointId || newMessage.evalId}:${newMessage.executionPoint}`;
-    const existingMessage = state.logpointMessages.entities[key];
-    if (existingMessage) {
-      log(`LogpointFinish ${newMessage.executionPoint}`);
-      removedIds.push(existingMessage.messageId);
-    } else {
-      log(`LogpointStart ${newMessage.executionPoint}`);
-    }
-    logpointMessagesAdapter.upsertOne(state.logpointMessages, { key, messageId: newMessage.id });
-  }
-
   // Immer will freeze anyway, but this might help skipping the `ValueFront` fields
   const addedMessage = Object.freeze(newMessage);
   messagesAdapter.upsertOne(state.messages, addedMessage);
@@ -399,7 +357,7 @@ function addMessage(newMessage: Message, state: MessageState): MessageState {
     state.filteredMessagesCount[addedMessage.level as FilterCountKeys]++;
   }
 
-  return removeMessagesFromState(state, removedIds);
+  return state;
 }
 
 interface MessagesFilters {
diff --git a/src/protocol/logpoint.ts b/src/protocol/logpoint.ts
index e0609c1cc..75bca6970 100644
--- a/src/protocol/logpoint.ts
+++ b/src/protocol/logpoint.ts
@@ -38,7 +38,6 @@ export const LogpointHandlers: {
     time: number,
     location: Location
   ) => void;
-  clearLogpoint?: (logGroupId: string) => void;
 } = {};
 
 let store: UIStore;
@@ -433,7 +432,6 @@ export function newLogGroupId() {
 export async function setEventLogpoints(eventTypes: string[]) {
   for (const eventType in eventTypeLogGroupId) {
     if (eventTypes.indexOf(eventType) < 0) {
-      removeLogpoint(eventTypeLogGroupId[eventType]);
       delete eventTypeLogGroupId[eventType];
     }
   }
@@ -533,7 +531,6 @@ export async function setExceptionLogpoint(logGroupId: string) {
   try {
     await analysisManager.runAnalysis(params, handler);
   } catch (e: any) {
-    removeLogpoint(logGroupId);
     let msg;
     if (e.code === ProtocolError.TooManyPoints) {
       msg = "There are too many exceptions. Please focus to a smaller time range and try again.";
@@ -544,9 +541,3 @@ export async function setExceptionLogpoint(logGroupId: string) {
     store.dispatch(exceptionLogpointErrorReceived(msg));
   }
 }
-
-export function removeLogpoint(logGroupId: string) {
-  if (LogpointHandlers.clearLogpoint) {
-    LogpointHandlers.clearLogpoint(logGroupId);
-  }
-}
```

bvaughn:
I noticed a small regression for log points when you move the focus region around. We might be clearing message state too aggressively (and wiping out local log points too?) Going to dig in a bit. Will hold off merging for now.

src/devtools/client/webconsole/reducers/messages.ts
jcmorrow: @bvaughn can you describe what the "half-broken" state that you're seeing is?

bvaughn: What I was seeing earlier (and will dig into in the morning) is that the logpoint UI would stick around, but the messages would get cleared from the console and never show up again _unless_ you removed and re-saved the logpoint. I have some ideas about why this is (though I'm still not sure I understand the purpose of the piece of state we're discussing above).

bvaughn: Update: I was doing something stupid (clearing _all_ messages in memory, including log points). Since log point state is stored in a few places, the log point overlay stuck around but the log points never came back in the console.

I've fixed it with the most recent push. I'll probably go ahead and land this in a few minutes, assuming CI is happy, and we can iterate.

bvaughn:
The two failing e2e tests (`stepping-05.js` and `breakpoints-07.js`) are also failing on every recent commit to `main` so I'm going to move forward and land this.

jaril:
I think this is related to https://github.com/RecordReplay/devtools/issues/6851 so just linking it here!

PR #6918, opened 2022-05-23
NetworkMonitor ResizeObserver

yeion7:
The network monitor resize observe is being call on each render of the component, unfortunally this component get's render a lot when some panel get resize or the timeline change 

https://user-images.githubusercontent.com/9127504/169877220-61c3f33c-e91a-494e-acbf-5433dc23b3b7.mp4

Reading the [spec](https://drafts.csswg.org/resize-observer/#dom-resizeobserver-observe)

```
observe(target, options)

Adds target to the list of observed elements.

1. If target is in [[observationTargets]] slot, call unobserve(target).

2. Let resizeObservation be new ResizeObservation(target, options).

3. Add the resizeObservation to the [[observationTargets]] slot.
```

Basically on each render we're removing and setting back the observer again

With this PR, the observer will be set once and then clean up on NetworkMonitor unmount

### DIFF
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -56,10 +56,19 @@ export const NetworkMonitor = ({
   );
 
   useEffect(() => {
-    if (container.current) {
-      resizeObserver.current.observe(container.current);
+    const observer = resizeObserver.current;
+    const splitBoxContainer = container.current;
+
+    if (splitBoxContainer) {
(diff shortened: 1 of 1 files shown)


markerikson (changes requested):
Needs to be reformatted, but the actual change looks good!

yeion7:
> Needs to be reformatted, but the actual change looks good!

Done

markerikson (approved):

markerikson:
Thanks!

PR #6927, opened 2022-05-24
An initial (and naive) take on soft-focus

jcmorrow:
### Motivation and Changes

Soft focus has been a theoretical goal of the front-end team for close to a month at this point (although it was really just me tinkering and thinking for the first couple of weeks). And that has produced some results, but there's still quite a bit in motion, as well as some unknowns. Technically, all we need to do to enable V1 of soft focus is *not* onload regions upon the focus window changing (we still issue the load command, because if we are already loaded it is a noop, and if the backend has unloaded a portion of the recording that the user has asked for, we need to ensure that we reload it). One thing that jumps out as soon as you enable soft-focus is that we need more specific points to pass for ranges to backend functions like `getHitCounts`. I tried just using `getPointNearTime` with the focus window times, but that is no good, because it always returns the beginning of the region, so sub-region focus windows end up looking like windows of 0 length in point-terms. Instead, I do a more sophisticated version of what we were doing previously with timeline seeks:

 Any time we get a timestamped point from the backend, we store it in a sorted array in redux. Then, when the user sets the focus window, we use that point array to approximate the point that lies near the user's selected times. This is sometimes not very fine-grained, but here's the elegant part: if there is stuff that *matters* around the area that is being selected (for instance, analysis hits), then we will have stored fine-grained point correlations already, and we won't be vastly off. Conversely, if the points we have are really rough approximations, it is unlikely that they will affect anything anyways, because the user has not looked at anything of interest around them.

I was prepared for this to be terrible, but in my relatively limited experience... it's pretty good? I think at the very least it is good enough to enable it internally and poke around.

Followups:

- We should actually soft-focus analyses, but I will wait on Mark's rewrite of the analysisManager before making that change.

### DIFF
--- a/src/devtools/client/debugger/src/reducers/source-actors.ts
+++ b/src/devtools/client/debugger/src/reducers/source-actors.ts
@@ -4,6 +4,8 @@
 
 import type { AnyAction, Action } from "@reduxjs/toolkit";
 import sortBy from "lodash/sortBy";
+import { setLoadedRegions } from "ui/reducers/app";
+import { setFocusRegion } from "ui/reducers/timeline";
 import type { UIState } from "ui/state";
 
 import { asSettled, asyncActionAsValue, AsyncValue } from "../utils/async-value";
@@ -108,8 +110,12 @@ export default function update(state = initial, action: AnyAction) {
       state = updateBreakpointHitCounts(state, action as SetSourceActorBreakpointHitCountsAction);
       break;
 
-    case "app/setLoadedRegions":
-      state.values = clearBreakpointHitCounts(state);
+    case setLoadedRegions.type:
+      state = { ...state, values: clearBreakpointHitCounts(state) };
+      break;
+
+    case setFocusRegion.type:
+      state = { ...state, values: clearBreakpointHitCounts(state) };
       break;
 
     case "SET_SOURCE_ACTOR_BREAKABLE_LINES":
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -1,5 +1,6 @@
 import { ExecutionPoint, PauseId } from "@recordreplay/protocol";
 import { refetchMessages } from "devtools/client/webconsole/actions/messages";
+import sortedIndexBy from "lodash/sortedIndexBy";
 import {
   getGraphicsAtTime,
   paintGraphics,
@@ -13,7 +14,7 @@ import {
   Video,
   timeIsBeyondKnownPaints,
 } from "protocol/graphics";
-import { client, log, sendMessage } from "protocol/socket";
+import { client, log } from "protocol/socket";
 import { ThreadFront } from "protocol/thread";
 import { Pause } from "protocol/thread/pause";
 import { PauseEventArgs } from "protocol/thread/thread";
@@ -28,23 +29,16 @@ import {
   getZoomRegion,
   getShowFocusModeControls,
 } from "ui/reducers/timeline";
-import { HoveredItem, FocusRegion } from "ui/state/timeline";
+import { HoveredItem } from "ui/state/timeline";
 import { getPausePointParams, getTest, updateUrlWithParams } from "ui/utils/environment";
 import KeyShortcuts, { isEditableElement } from "ui/utils/key-shortcuts";
 import { features } from "ui/utils/prefs";
 import { trackEvent } from "ui/utils/telemetry";
 
 import {
+  setFocusRegion as newFocusRegion,
   setHoveredItem as setHoveredItemAction,
   setPlaybackStalled,
-  setTimelineState,
-  setTrimRegion,
-} from "../reducers/timeline";
-
-export {
-  setPlaybackPrecachedTime,
-  setPlaybackStalled,
-  setTrimRegion,
   setTimelineState,
 } from "../reducers/timeline";
 
@@ -226,10 +220,9 @@ export function seekToTime(targetTime: number): UIThunkAction {
     }
 
     const nearestEvent = mostRecentPaintOrMouseEvent(targetTime) || { point: "", time: Infinity };
-    const pointNearTime =
-      // @ts-ignore
-      (await sendMessage("Session.getPointNearTime", { time: targetTime }, ThreadFront.sessionId!)) // @ts-ignore
-        .point;
+    const pointNearTime = (
+      await client.Session.getPointNearTime({ time: targetTime }, ThreadFront.sessionId!)
+    ).point;
 
     if (Math.abs(pointNearTime.time - targetTime) > Math.abs(nearestEvent.time - targetTime)) {
       dispatch(seek(nearestEvent.point, targetTime, false));
@@ -456,7 +449,9 @@ export function clearHoveredItem(): UIThunkAction {
   };
 }
 
-export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
+export function setFocusRegion(
+  focusRegion: { startTime: number; endTime: number } | null
+): UIThunkAction {
   return (dispatch, getState) => {
     const state = getState();
     const currentTime = getCurrentTime(state);
@@ -517,14 +512,41 @@ export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
         }
       }
 
+      // We now try and create a focus region with a time and a point, from just a
+      // time. In the future, it would be better if this were something like
+      // "setFocusRegionByTime" and we had another method which accepted
+      // TimeStampedPoints, because setting via context menu will often happen on
+      // resources where we already know the time *and* point. Also, we probably
+      // should *not* do this on every mouse movement in focus mode. This is
+      // actually foregoing most of the `getPointNearTime`, and instead relying on
+      // points that we mostly have already. I think often this will be OK, but
+      // probably we should also run "getPointNearTime", store that, and take the
+      // closest thing we can find of the points we have (assuming we don't have
+      // an exact match.)
+      const startIndex = sortedIndexBy(
+        state.timeline.points,
+        { time: startTime, point: "" },
+        p => p.time
+      );
+      const start =
+        startIndex > 0 ? state.timeline.points[startIndex - 1] : { point: "", time: startTime };
+      const endIndex = sortedIndexBy(
+        state.timeline.points,
+        { time: endTime, point: "" },
+        p => p.time
+      );
+      const end = endIndex > 0 ? state.timeline.points[endIndex - 1] : { point: "", time: endTime };
+
       dispatch(
-        setTrimRegion({
-          endTime,
+        newFocusRegion({
+          start,
           startTime,
+          end,
+          endTime,
         })
       );
     } else {
-      dispatch(setTrimRegion(null));
+      dispatch(newFocusRegion(null));
     }
   };
 }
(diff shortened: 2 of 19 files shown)


jonbell-lot23:
Anything in particular I should be thinking about here @jcmorrow ? I've been following along from a distance but I'm not sure if there's messaging I need to be considering

hbenl (approved):
LGTM

src/devtools/client/debugger/src/reducers/source-actors.ts
```diff
@@ -112,6 +112,10 @@ export default function update(state = initial, action: AnyAction) {
       state.values = clearBreakpointHitCounts(state);
       break;
 
+    case "timeline/setTrimRegion":
+      state.values = clearBreakpointHitCounts(state);
```
markerikson, line 116: ⚠️ FYI this is actually mutating `state`, because we don't have `createSlice` here and we aren't using Immer at all.  Gotta do a full `{...state, values: clearBreakpointHitCounts(state)}` here.

src/ui/actions/timeline.ts
```diff
@@ -467,65 +469,93 @@ export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
       dispatch(stopPlayback());
     }
 
-    if (focusRegion !== null) {
-      const zoomRegion = getZoomRegion(state);
-      const { endTime: prevEndTime, startTime: prevStartTime } = getFocusRegion(state) || {};
-
-      let { endTime, startTime } = focusRegion;
+    if (focusRegion === null) {
```
markerikson, line 472: :speech_balloon: not suggesting this has to be done now, but most of this calculation _really_ feels like something that belongs in a reducer instead if at all possible.

src/ui/reducers/timeline.ts
```diff
@@ -42,6 +45,12 @@ const timelineSlice = createSlice({
     setTrimRegion(state, action: PayloadAction<FocusRegion | null>) {
       state.focusRegion = action.payload;
     },
+    pointsReceived(state, action: PayloadAction<TimeStampedPoint[]>) {
+      state.points = sortBy(
```
markerikson, line 49: :hand: This _almost_ feels like something we could use an entity adapter for?

If `p.time` is effectively acting like a unique ID, we could do an `adapter.upsertMany(action.payload)`, and also use its `comparator` option to manage the sorting.

Probably not a big difference either way - I'm just pattern-matching on tools I know here.

src/ui/utils/prefs.ts
```diff
@@ -66,6 +67,7 @@ export const features = new PrefsHelper("devtools.features", {
   unicornConsole: ["Bool", "unicornConsole"],
   showRedux: ["Bool", "showRedux"],
   enableLargeText: ["Bool", "enableLargeText"],
+  softFocus: ["Bool", "softFocus"],
```
bvaughn, line 70: Ooo, should we go back and add this feature flag to #6915? I could open a PR that does that.

src/devtools/client/debugger/src/reducers/source-actors.ts
markerikson: Actually it looks like we also already do this with `"app/setLoadedRegions"` above as well - can you fix that too?

src/ui/reducers/timeline.ts
```diff
@@ -42,6 +45,12 @@ const timelineSlice = createSlice({
...
+        uniqBy(state.points.concat(action.payload), p => p.time),
+        p => p.time
+      );
+    },
```
bvaughn, line 53: This code is succinct but it also looks kind of expensive. How often do we think it will run?

For example, I think we could write a more performant version like this (but admittedly it's a lot less readable):
```js
const points = state.points.concat();

let actionIndex = 0;
let stateIndex = 0;

while (actionIndex < action.payload.length) {
  const actionPoint = action.payload[actionIndex];

  let stateIndexPoint = points[stateIndex];

  if (stateIndexPoint == null) {
    points.push(actionPoint);
    actionIndex++;
  } else if (actionPoint.time === stateIndexPoint.time) {
    // Don't add duplicates.
    actionIndex++;
  } else if (actionPoint.time < stateIndexPoint.time) {
    points.splice(stateIndex, 0, actionPoint);
    actionIndex++;
    stateIndex++;
  } else {
    stateIndex++;
  }
}
```

I think this would be significantly faster (see [jsben.ch/xtKWW](https://jsben.ch/xtKWW)) but maybe it's not worth unless we think it's likely to be run a lot?

In general I'm concerned about how many inline maps and sorts we do– on particularly large sets of data– without memoization.

src/ui/actions/timeline.ts
markerikson: Yes, absolutely.  That's part of the point of Redux.  We _want_ as much logic as possible in a reducer, because pure functions are easier to test and predict.

We also specifically recommend dispatching fewer actions, and modeling actions as "events" rather than "setters":

- https://redux.js.org/style-guide/#put-as-much-logic-as-possible-in-reducers
- https://redux.js.org/style-guide/#model-actions-as-events-not-setters
- https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially
- https://redux.js.org/style-guide/#allow-many-reducers-to-respond-to-the-same-action

src/devtools/client/debugger/src/reducers/source-actors.ts
```diff
@@ -112,6 +112,10 @@ export default function update(state = initial, action: AnyAction) {
       state.values = clearBreakpointHitCounts(state);
       break;
 
+    case "timeline/setTrimRegion":
```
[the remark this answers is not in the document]
markerikson: Ideally, we'd actually be importing the real `setTrimRegion` action creator from the other slice file, and doing:

```js
extraReducers: (builder) => {
  builder.addCase(setTrimRegion, (state, action) => {
    // reducer updates
  })
}
```

and at that point we wouldn't be caring about action type strings at all any more.  But, this reducer hasn't been updated to RTK yet.

We _could_ still import the action creator and do `case setTrimRegion.type` to get the same behavior, but we haven't done that yet.  

(there's also the question of whether doing that adds a circular import.)

src/devtools/client/debugger/src/reducers/source-actors.ts
jcmorrow: > we don't have createSlice here

Oh shoot I totally did not realize this. Thanks for the catch! You've done such a thorough job I just assume that we have `createSlice` everywhere 😅

src/ui/actions/timeline.ts
```diff
@@ -467,65 +469,93 @@ export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
...
+      dispatch(setTrimRegion(null));
```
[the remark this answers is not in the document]
jcmorrow: Yes. I have read `set_trim` for months and everytime I see it I think "Is that like focus? Or is it like zoom? Or is it something else?"

src/ui/reducers/timeline.ts
markerikson: Yeah, part of my reason for suggesting an entity adapter here _was_ to potentially simplify perf behavior, per your comment below :)

but also I was just trying to read this line of code and understand what it's doing, and it looks like it's doing a combo of "unique ID by timestamp" and "keep this sorted", and those are also things that an entity adapter can help with.

src/ui/actions/timeline.ts
```diff
@@ -467,65 +469,93 @@ export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
...
+        dispatch(setTimelineToTime(currentTime));
+      }
     }
+
+    // We now try and create a focus region with a time and a point, from just a
+    // time. In the future, it would be better if this were something like
+    // "setFocusRegionByTime" and we had another method which accepted
+    // TimeStampedPoints, because setting via context menu will often happen on
+    // resources where we already know the time *and* point. Also, we probably
+    // should *not* do this on every mouse movement in focus mode. This is
```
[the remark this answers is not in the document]
jcmorrow: I thought that this happened on every mouse move when dragging the left or right handle of the focus window. But it sounds like I might be misunderstanding when this code runs?

src/ui/actions/timeline.ts
```diff
@@ -467,65 +469,93 @@ export function setFocusRegion(focusRegion: FocusRegion | null): UIThunkAction {
...
+    // actually foregoing most of the `getPointNearTime`, and instead relying on
+    // points that we mostly have already. I think often this will be OK, but
+    // probably we should also run "getPointNearTime", store that, and take the
+    // closest thing we can find of the points we have (assuming we don't have
+    // an exact match.)
```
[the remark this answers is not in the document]
jcmorrow: So I think the *ideal* case would be:

We are trying to translate a time to a point.

- We see if we have any exact matches (and if so, just use that, done).
- If we don't have any exact matches, go ask the backend for the nearest point, and then store the result (so next time we won't have to hit the backend)

However, that assumes that the points that the backend are returning are actually close enough to use (in some cases they are, in some they are not, the *average* amount that a backend response to `getPointNearTIme` right now is "off" is about a quarter of a second, but it has a long tail of higher errors like 10 seconds or more). We should improve the accuracy of that backend endpoint, and then we should centralize logic for the above, so we never have to wonder "which technique should I use here?"

However, we are not in that world yet, and I did not want to put a protocol call into something that I suspected might be getting hit on every mouse-move while setting the focus window, so I forewent the `getPointNearTime` here, for the reasons explained in the PR body. Let me know what you think here! I'm still trying to figure out the best way to do this!

src/ui/reducers/timeline.ts
jcmorrow: yeah, I was definitely a little lazy when writing this, I'm happy to add a more tuned version.

src/ui/reducers/timeline.ts
bvaughn: Fair!

src/ui/actions/timeline.ts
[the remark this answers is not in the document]
bvaughn: You're totally right. For some reason I thought we were in the sync-focus-region bit that only applied on save. Yeaahhh, this is a lot to do on every drag event. I see slowness already on some recordings with a lot of filtering done in response to tight events like that.

src/ui/actions/timeline.ts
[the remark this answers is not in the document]
bvaughn: Yeah, that's fair. Thanks for elaborating.

I think in general, we should probably debounce some of this stuff (or use an approach like `useDeferredValue` or something). We definitely don't want to be firing off protocol messages on mouse-move.

I trust your intuition about whether this is the right approach. I was just uneasy because I wasn't sure how you'd gotten to that conclusion (or how I'd know to do the same thing if I were writing the code).

src/ui/state/timeline.ts
```diff
@@ -7,7 +7,9 @@ export interface ZoomRegion {
 }
 
 export interface FocusRegion {
+  end: TimeStampedPoint;
   endTime: number;
+  start: TimeStampedPoint;
```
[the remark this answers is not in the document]
bvaughn: I think that's okay as an intermediate state. Let's just clearly document which one should take precedent (maybe where we define the type)?

bvaughn (approved):
I don't want to block this PR from landing, so 👍🏼 

I'm not sure I fully understand some of it (like the point-to-time lookup bit) but the same could be said about the code that's already on main so 😂

PR #6928, opened 2022-05-24
Organization only can edit team name

### LINKED ISSUE #6799: Disable editing a team name if the team is not an organization
We have an edit team name feature that a user can access by going into team settings > profile, but it's only enabled for [organizations](https://github.com/RecordReplay/backend/blob/f5c4edac6fe15add730cc5828f956d2df05e6639/src/graphql-api/models.ts#L2205).

Right now, we have the same input box for the title regardless of whether or not the team is an organization. We should disable that input box if the team is not an organization to keep the user from trying to edit it in the first place.

image

ryota-murakami:
fix #6799

Disabled text input if team is not Organization.

### Screen Shot (When edit disabled)
Screen Shot 2022-05-24 at 19 06 11


ryota-murakami:
@jaril I submitted this PR, I appreciate your review!

jaril (approved):

jaril:
@ceceliacreates user-facing change 🙋‍♂️

PR #6930, opened 2022-05-24
Intentionally break things

bvaughn:
Sanity checking our e2e test


bvaughn:
Ah, looking at the Vercel deployment associated with this PR I see that it shows _both_ domains:
Screen Shot 2022-05-24 at 9 16 52 AM

That seems okay then. So why does the UI look different for me?

Ah! It's because we only show the "missing" controls to "internal" users, and the e2e tests aren't running with that permission.

https://github.com/RecordReplay/devtools/blob/b4bde0cfe15e89b37aac0477b5b2c3892d493c7c/src/ui/components/ProtocolViewer.tsx#L139-L148

PR #6945, opened 2022-05-25
Add a test result indicator for replays with test metadata

jaril:
image

### DIFF
--- a/src/ui/components/Library/RecordingRow.tsx
+++ b/src/ui/components/Library/RecordingRow.tsx
@@ -145,6 +145,13 @@ function RecordingRow({
                 <div className="overflow-hidden overflow-ellipsis whitespace-pre font-light text-gray-400">
                   {getDisplayedUrl(recording.url)}
                 </div>
+                {
+                  recording.metadata?.test ? (
+                    <div className="overflow-hidden overflow-ellipsis whitespace-pre font-light text-gray-400">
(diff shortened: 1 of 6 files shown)


src/ui/types/index.ts
```diff
@@ -144,6 +144,7 @@ export interface Recording {
   operations?: OperationsData;
   resolution: { resolvedAt: string; resolvedBy: string };
   collaboratorRequests: CollaboratorRequest[];
+  metadata: any;
```
ryanjduffy, line 147: Would be good to tighten this up. It's something like `Record<string, null | Record<string, unknown>>` which is a bit weird.

src/ui/components/Library/RecordingRow.tsx
```diff
@@ -139,6 +145,13 @@ function RecordingRow({
                 <div className="overflow-hidden overflow-ellipsis whitespace-pre font-light text-gray-400">
                   {getDisplayedUrl(recording.url)}
                 </div>
+                {
+                  recording.metadata?.test ? (
```
ryanjduffy, line 149: This probably won't affect layout but it might be worth checking if `file` is set here so we don't add a node that is empty.

e.g. `recording.metadata?.test?.file`

jaril:
Cleaned up types a bit. Let me know if this looks a tad better!

PR #6955, opened 2022-05-25
Removing external deps from new protocol package

bvaughn:
Follow up for #6953.

- [x] Adds new "shared" package with constants referenced by both protocol and Replay devtools app
- [x] Removes references to devtools test utils from protocol packages
- [ ] ~~Remove remaining references to devtools app code from protocol package~~ Moving these to a follow up.


packages/protocol/utils.ts
```diff
@@ -1,5 +1,6 @@
 import * as Sentry from "@sentry/browser";
-import { isDevelopment } from "ui/utils/environment";
+// TODO [bvaughn] protocol
+import { isDevelopment } from "shared/utils/environment";
```
jasonLaster, line 3: feels funny for the protocol to need to know about `environment`. If anything we could pass an environment in when creating a session.

packages/protocol/socket.ts
```diff
@@ -8,7 +8,7 @@ import {
   CommandParams,
   CommandResult,
 } from "@recordreplay/protocol";
-import { isMock, mockEnvironment, waitForMockEnvironment } from "ui/utils/environment";
+import { isMock, mockEnvironment, waitForMockEnvironment } from "shared/utils/environment";
```
jasonLaster, line 11: I bet you can avoid importing `shared/utils/environment` from `packages/protocol`. Might need other utils, but probably not this one

jasonLaster (commented):
Looking great

packages/protocol/utils.ts
bvaughn: Yeah! Maybe so. Need to think about that more.

I don't think this (packages/protocol) is very close to what we'll eventually publish to NPM btw! The main goal of these PRs is to get things in a place where I can start prototyping something on top of the shared packages/protocol code. The "something" will, hopefully be the thing we eventually publish for others to build on top of.

packages/protocol/socket.ts
bvaughn: The "packages/shared" package is based on the idea from React's package of a similar name– things that are used by multiple projects in the monorepo. Any code in the package would get inlined during build (aka there's no "shared" NPM package). I think conceptually, this may make sense for Replay too.

Now maybe environment isn't a good use case for that. Totally.

I don't want to get too hung up on designing a perfect end state for this package though, because I don't think it's compelling to publish and consume anywhere other than Replay _for now_. Just taking baby steps. :)

bvaughn: Removed in 41378d5a822ce33bb5e23cec70df202a1ab2ef58

packages/protocol/utils.ts
```diff
@@ -1,5 +1,14 @@
...
+
+let errorHandler: ErrorHandler = (error: Error) => {
+  throw error;
+};
+
+// By default, this package throw Errors.
+// You can override that behavior with this method (e.g. to log errors to Sentry).
+export function setErrorHandler(customErrorHandler: ErrorHandler) {
+  errorHandler = customErrorHandler;
+}
```
jasonLaster, line 11: yeah - this is clever.

pages/_document.tsx
```diff
@@ -1,6 +1,16 @@
-import React from "react";
+import * as Sentry from "@sentry/browser";
 import crypto from "crypto";
 import Document, { Html, Head, Main, NextScript } from "next/document";
+import { setErrorHandler } from "protocol/utils";
+import React from "react";
+import { isDevelopment } from "ui/utils/environment";
+
+if (!isDevelopment()) {
+  // Configure protocl to log errors to Sentry in production environments.
```
jasonLaster, line 9:
```
Configure protocol
```

jasonLaster (commented):
Looks good to me

packages/protocol/socket.ts
```diff
@@ -22,9 +21,29 @@ export interface Request<T extends CommandMethods> {
...
+export function setCustomSendMessageHandler(
+  sendMessage: SendMessageMethod
+): setCustomSendMessageHandlerResponse {
+  customSendMessageHandler = sendMessage;
+
+  return {
+    flushQueuedMessages,
+    responseDataHandler: socketDataHandler,
+  };
+}
```
jasonLaster, line 45: I like this pattern of setting overrides.

src/ui/utils/environment.ts
```diff
@@ -58,29 +51,32 @@ export function isTest() {
...
 export function isMock() {
   return !!url.searchParams.get("mock");
 }
 
-export async function waitForMockEnvironment() {
+// Helper method to retrieve the mock environment that was injected by a test runner.
+export async function getMockEnvironmentForTesting(): Promise<MockEnvironment> {
   if (!isMock()) {
-    return null;
+    throw Error("Not a mock/test environment");
```
jasonLaster, line 64: Yesss!

jasonLaster (approved):
This is great!

test/mock/src/mockEnvironment.ts
```diff
@@ -112,14 +114,29 @@ export function doInstall(options: MockOptionsJSON) {
...
+  // Configure the protocol to use our mock WebSocket message method.
+  const { flushQueuedMessages, responseDataHandler } = setCustomSendMessageHandler(
+    mockEnvironment.sendSocketMessage
+  );
+
+  // Configure the protocol's default socket message handler to receive messages from the mocked WebSocket.
+  mockEnvironment.setSocketDataHandler(responseDataHandler);
+
+  // Now that the mock environment has been configured, flush any pending messages.
+  flushQueuedMessages();
```
bvaughn, line 135: Ah, I don't think any of these lines of code can run from within the context of `page.evaluate` 😢  Need to re-think.

packages/protocol/thread/thread.ts
```diff
@@ -518,12 +517,13 @@ class _ThreadFront {
   async getHitCounts(
     sourceId: SourceId,
     locations: SameLineSourceLocations[],
-    focusRegion: FocusRegion | null
+    beginExecutionPoint?: ExecutionPoint,
+    endExecutionPoint?: ExecutionPoint
```
jcmorrow, line 521: It might be better to make a `range` param here, because that is what we will ultimately call the protocol with, and passing in just one of these would be veering towards "hard to define" behavior.

packages/shared/constants.ts
```diff
@@ -0,0 +1,3 @@
+// The PageStyle actor flattens the DOM CSS objects a little bit, merging Rules and their Styles into one actor.
+// For elements (which have a style but no associated rule) we fake a rule with the following style id.
+export const ELEMENT_STYLE = 100;
```
jcmorrow, line 3: wow this is weird. I had not seen this yet. Given that this will now be the `git blame` for this line, it might be worth throwing in a better `sha` to check for history, or just a pointer to somewhere else to learn more?

bvaughn: There was no original context. See commit 171ea70586c4b26e2509c0698cde1e25636af1cb with message:
> Add inspector client

packages/protocol/thread/thread.ts
bvaughn: Sure!

PR #6956, opened 2022-05-25
Only use start and end points when soft focus is enabled

jcmorrow:
There's just a little too much fussiness with focus regions I think for this to be generally released. A lot of the best benefits will already be had just by passing ranges to console logs and getHitCounts, and those will be available for all users. Meanwhile, us internal folks can mess around with this, figure out when it feels good, when it feels bad, and roll it out a little slower.


src/ui/utils/timeline.ts
```diff
@@ -22,6 +23,14 @@ export function getPixelDistance({
   return Math.abs((toPos - fromPos) * overlayWidth);
 }
 
+export function startTimeForFocusRegion(focusRegion: FocusRegion) {
```
jcmorrow, line 26: These are just wrappers that know about checking prefs, which keeps the logic contained.

src/ui/actions/timeline.ts
```diff
@@ -523,19 +532,19 @@ export function setFocusRegion(
       // probably we should also run "getPointNearTime", store that, and take the
       // closest thing we can find of the points we have (assuming we don't have
       // an exact match.)
-      const startIndex = sortedIndexBy(
+      const startIndex = sortedLastIndexBy(
```
jcmorrow, line 535: `sortedLastIndex - 1` is equivalent to saying "the very last index we could use while still containing the entire asked-for region.

src/ui/actions/timeline.ts
```diff
@@ -535,7 +545,7 @@ export function setFocusRegion(
         { time: endTime, point: "" },
         p => p.time
       );
-      const end = endIndex > 0 ? state.timeline.points[endIndex - 1] : { point: "", time: endTime };
+      const end = endIndex > 0 ? state.timeline.points[endIndex] : { point: "", time: endTime };
```
jcmorrow, line 548: `endIndex - 1` is wrong, we want the point we would have inserted this *before*, not the point we would have inserted it *after*.

jasonLaster:
> There's just a little too much fussiness with focus regions I think for
this to be generally released. A lot of the best benefits will already
be had just by passing ranges to console logs and getHitCounts, and
those will be available for all users. Meanwhile, us internal folks can
mess around with this, figure out when it feels good, when it feels bad,
and roll it out a little slower.

Agreed

markerikson (approved):

src/ui/utils/timeline.ts
bvaughn: I like that we're encapsulating this logic in a single place.

I'm concerned though that there is no static analysis checks to prevent people from reading `focusRegion.start` or `focusRegion.startTime` directly.

I wonder if it would be worth making this type opaque so that people _couldn't_ directly interact with those attributes (without a type cast or ts-ignore)? That might help avoid references from leaking in.

jcmorrow: Good idea. I've made the `FocusRegion` type empty, and made an `UnsafeFocusRegion` which you can cast to if you need to directly access fields on it.

src/ui/state/timeline.ts
```diff
@@ -8,8 +8,13 @@ export interface ZoomRegion {
...
   // We are moving towards using TimeStampedPoints for all ranges on the client,
-  // to that end, we should avoid using `endTime` and `startTime` from this
-  // object whenever possible
+  // to that end, we should avoid using fields directly from this object, and
+  // instead use the getters `startTimeForFocusRegion` and
+  // `endTimeForFocusRegion`. If you must access the fields directly, you can
+  // cast instead to an `UnsafeFocusRegion`
+}
+
+export interface UnsafeFocusRegion {
```
bvaughn, line 17: Nice 👍🏼  This gives us better guards for the interim.

src/ui/utils/timeline.ts
bvaughn: Thanks! 🙇🏼

bvaughn (approved):

PR #6962, opened 2022-05-26
Show duration for individual requests

jcmorrow:
In the protocol viewer, sometimes you see that a group of requests was slow, but only *some* were slow, and seeing their durations in the detail view helps to narrow down.

CleanShot 2022-05-25 at 23 57 01


markerikson (approved):

jasonLaster:
Small pet-peeve, could we round the durations?

Issue #6548, opened 2022-04-25
Clicking on breakpoint timeline does nothing


bhackett1024:
In this replay, there were a few times when I tried to click on the breakpoint timeline to jump to one of the hits, but nothing happened, and it doesn't look like we were waiting on any backend messages afterwards.

https://app.replay.io/recording/view-long-replay--1afe2dc1-646f-4403-941c-5f68c5e8e11a?point=139218459549484050787382377125518521&time=46900&hasFrames=true

hbenl:
The breakpoint timeline click is handled correctly: you clicked on the logpoint that you were already paused at.
The real issue is that the UI didn't show any frames or scopes for the current (initial) pause. This happened because when we try to create the pause, that region hasn't been loaded yet, so the `Session.createPause` command fails, hence `Pause.createWaiter` is rejected and `Pause.getFrames()` fails when it tries to `await this.createWaiter` (and there's no UI yet showing that fetching the frames failed).
Has the handling of loaded regions in `Session.createPause` changed recently? Otherwise I'd be surprised that we haven't noticed this so far, it should happen quite often when you open a replay link that has a `point` parameter for a pause with frames.

jcmorrow:
Got another good replay of this: https://app.replay.io/recording/cant-create-pause-on-start--74ce924b-abc3-44ac-b9ec-3c65b02b52f6

jaril:
Bumped this to next week's sprint!

jcmorrow:
Closed by https://github.com/replayio/devtools/pull/7009

PR #6963, opened 2022-05-26
Don't try to create initial pauses in unloaded regions

jcmorrow:
We used to have a simpler pause world before focus and turbo replays. Now, we have to do things like handle pauses that try to get created in regions that it is not possible to create a pause in (at the moment). Upon loading, we could do a few things:
- We could not try to pause in unloaded regions, and just go to the end instead
- We could try to load the requested region
- I think the latter is better, but the former is a *lot* easier 😅 

Related to #6548 , though it's not a perfect solve.

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/fetchScopes.js
+++ b/src/devtools/client/debugger/src/actions/pause/fetchScopes.js
@@ -14,7 +14,7 @@ export function fetchScopes(cx) {
       return;
     }
 
-    const scopes = dispatch({
+    dispatch({
       type: "ADD_SCOPES",
       cx,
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/actions/pause/fetchScopes.js
```diff
@@ -14,7 +14,7 @@ export function fetchScopes(cx) {
       return;
     }
 
-    const scopes = dispatch({
```
jcmorrow, line 17: Just cleaning up some random cruft.

markerikson (approved):

PR #6969, opened 2022-05-26
Add tests and testRuns views to the library

jaril:
This PR does a few things:
- Adds a library test run/test runs view as a dropdown option where the filter is (that's locked behind a pref!)
- Adds a test view for showing a grouping of recordings by the test (test path, if we're being specific).
- Adds a test runs view for showing a grouping of recording by the test run.
- Adds a click through action for both views so that you can drill into each test/test run

### DIFF
--- a/src/ui/components/Library/ViewerHeader.tsx
+++ b/src/ui/components/Library/ViewerHeader.tsx
@@ -1,14 +1,101 @@
-import React from "react";
+import React, { useContext } from "react";
+import { useSelector } from "react-redux";
+import { getWorkspaceId } from "ui/reducers/app";
+import { Recording } from "ui/types";
+import { PrimaryButton, SecondaryButton } from "../shared/Button";
+import BatchActionDropdown from "./BatchActionDropdown";
 import styles from "./Library.module.css";
+import TeamTrialEnd from "./TeamTrialEnd";
+import { LibraryContext } from "./useFilters";
+
+function ViewerHeaderActions({
+  isEditing,
+  setIsEditing,
+  setSelectedIds,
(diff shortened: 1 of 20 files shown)


src/ui/components/Library/ViewerRouter.tsx
```diff
@@ -65,14 +65,14 @@ function NonPendingTeamLibrary({ currentWorkspaceId }: ViewerRouterProps) {
...
 
-  return (
-    <Viewer
-      recordings={recordings}
-      workspaceName={
-        workspace.logo ? <Base64Image src={workspace.logo} className="max-h-12" /> : workspace.name
-      }
-    />
+  const workspaceName = (
+    <div className="flex flex-row space-x-2">
```
ryanjduffy, line 69: If we're adding it, let's add `items-center` too so everything will be aligned reasonably

src/ui/components/Library/ViewerHeader.tsx
```diff
@@ -1,22 +1,109 @@
...
+      </div>
     </div>
   );
 }
 
 export function ViewerHeaderLeft({
   children,
 }: {
-  children: React.ReactChild | React.ReactChild[];
+  children: React.ReactChild | (React.ReactChild | null)[];
```
ryanjduffy, line 106: `React.ReactNode` is probably the better type here since it includes all of these

ryanjduffy (approved):
TBH, this is too big of a PR for me to review well without spending a lot of time getting context. It seems to be okay from what I tried and most of the new code is behind a pref so I'm approving.

PR #6973, opened 2022-05-26
Do not force repainting upon evaluations

jcmorrow:
Right now, we force a repaint after every evaluation, and we run an
evaluation on every token hover when paused. This can result in a ton of
protocol messages when just moving around the editor. While those are
not necessarily a problem, it does not seem like doing them is
accomplishing anything as far as I can tell. I'm guessing that after
*some* evaluations it *might* be useful to repaint, but if that's the
case, we should figure out how to pass that flag into `evaluate`, rather
than always doing it.

See https://www.notion.so/replayio/How-we-use-Pauses-057fc6b63ace4095bd19926cec135596 for more detail/discussion

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -38,6 +38,7 @@ import {
   findAnnotationsResult,
   getHitCountsParameters,
 } from "@replayio/protocol";
+import { features } from "ui/utils/prefs";
 import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
 
@@ -726,8 +727,12 @@ class _ThreadFront {
     } else if (rv.exception) {
       rv.exception = new ValueFront(pause, rv.exception);
     }
-    const { repaint } = await import("protocol/graphics");
-    repaint(true);
+
+    if (features.repaintEvaluations) {
+      const { repaint } = await import("protocol/graphics");
+      repaint(true);
+    }
+
     return rv;
   }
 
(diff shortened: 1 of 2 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -731,8 +731,6 @@ class _ThreadFront {
     } else if (rv.exception) {
       rv.exception = new ValueFront(pause, rv.exception);
     }
-    const { repaint } = await import("protocol/graphics");
-    repaint(true);
```
jasonLaster, line 735: Could you wrap this in a feature flag?

```js
if (app.prefs.repaintEvaluations) {
    const { repaint } = await import("protocol/graphics");
    repaint(true);
}
```

no need to expose the pref in settings, but nice to have

jasonLaster (approved):

packages/protocol/thread/thread.ts
jcmorrow: sure!

jcmorrow: Done.

packages/protocol/thread/thread.ts
```diff
@@ -38,6 +38,7 @@ import {
   findAnnotationsResult,
   getHitCountsParameters,
 } from "@replayio/protocol";
+import { features } from "ui/utils/prefs";
```
bvaughn, line 41: We should avoid adding any imports from `packages/protocol` to things that are outside of `packages/protocol`. Maybe we need a lint rule for this.

PR #6977, opened 2022-05-26
Extract "Replay app" logic files from `protocol` folder

markerikson:
- Moved `logpoint`, `event-listeners`, and `find-tests` files out
  from the `protocol` folder, because they really represent specific
  Replay app functionality, not "generic client" logic
- Removed "framework listeners" function from `NodeFront`, and
  added a `WeakMap` cache to `event-listeners.ts`


markerikson:
Technically same caching as before, I think - that was `NodeFront` trying to keep a promise around to hand back if asked again.  I just moved that to be outside the class.

markerikson:
Confirmed framework event listeners are still showing up in preview.

PR #6979, opened 2022-05-26
Remove protocol reference to test harness

bvaughn:
Continuing the work in #6955  and related to #6977.

This PR removes another reference (this time to `test/harness`) from the new `packages/protocol` folder.

This is a soft blocker to the proof of concept app for #6932.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -203,8 +203,6 @@ class _ThreadFront {
   private annotationWaiters: Map<string, Promise<findAnnotationsResult>> = new Map();
   private annotationCallbacks: Map<string, ((annotations: Annotation[]) => void)[]> = new Map();
 
-  testName: string | undefined;
-
   // added by EventEmitter.decorate(ThreadFront)
   eventListeners!: Map<ThreadFrontEvent, ((value?: any) => void)[]>;
   on!: (name: ThreadFrontEvent, handler: (value?: any) => void) => void;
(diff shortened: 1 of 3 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -266,19 +264,8 @@ class _ThreadFront {
...
-    this.ensureCurrentPause();
-
-    if (this.testName) {
-      await gToolbox.selectTool("debugger");
-      window.Test = await import("test/harness");
-      const script = document.createElement("script");
-      script.src = `/test/scripts/${this.testName}`;
-      document.head.appendChild(script);
-    }
-  }
```
bvaughn, line 278: Replay devtools _e2e test code_ definitely didn't belong in the protocol package.

src/ui/utils/devtools-toolbox.ts
```diff
@@ -57,6 +58,18 @@ export class DevToolsToolbox {
...
+    if (testName) {
+      await gToolbox.selectTool("debugger");
+
+      window.Test = await import("test/harness");
+
+      const script = document.createElement("script");
+      script.src = `/test/scripts/${testName}`;
+
+      document.head.appendChild(script);
+    }
```
bvaughn, line 71: I'm not sure this is the best place for it to live _either_ but since this is the place that calls `threadFront.initializeToolbox()` I moved it here for now. At least it runs at the same time as before.

Baby steps!

packages/protocol/thread/pause.ts
```diff
@@ -24,7 +24,7 @@ import { NodeFront } from "./node";
 import { RuleFront } from "./rule";
 import { StyleFront } from "./style";
 import { StyleSheetFront } from "./styleSheet";
-import type { ThreadFront as ThreadFrontType } from "./thread";
+import { ThreadFront as ThreadFrontType } from "./thread";
```
markerikson, line 27: :hand: technically this ought to stay as an `import type` to avoid the actual circular import if possible.

markerikson (approved):

packages/protocol/thread/pause.ts
bvaughn: Ahhh, fair. My Next app was saying it didn't know how to handle that type (heh) of import but in the name of keeping this PR focused, I'll remove that change.

PR #6980, opened 2022-05-26
Add standalone analysis processing implementation and update Redux DevTools sanitization

markerikson:
This PR:

- Adds a new `createAnalysis` internal API that allows more flexibility in inspecting the results of each step than the existing `AnalysisManager` class
- Removes an apparently obsolete `showInConsole` flag from `logpoint.ts`
- Updates the `breakpoints` reducer to track normalized analysis status entries and mappings from a location to those entries
- updates `actions/logpoint` to dispatch actions that track the status of a given analysis entry in Redux.
- Updates the Redux DevTools sanitization settings:
  - skip showing several actions that are high volume and low value ("protocolMessages", "hovered line number", etc)
  - Reworks the state sanitization behavior to only strip out `ValueFront` entries so we can see changes to areas like `state.messages` and `state.breakpoints.analyses`, but caches those sanitized versions to avoid recalculating unless needed.  (Note that having `ValueFront` objects in the state shown by the DevTools causes a lot of "`NotAllowed`" messages to be printed in the console when the DevTools tries to serialize them)

As of right now, there _shouldn't_ be any behavior changes.  The original logpoint + analysis behavior is still in place.  We're just saving additional data into the `state.breakpoints` reducer.

The next steps after this will be to migrate the breakpoint-related components to TS, and then start changing where we read location hit data for a given breakpoint.

An example state update in our Redux DevTools:

image


packages/protocol/socket.ts
```diff
@@ -225,6 +236,28 @@ export function addEventListener<M extends EventMethods>(
...
+      if (callbacks) {
+        callbacks.onResults(results);
+      } else {
+        handler({ analysisId, results });
+      }
+    });
+    return;
+  }
   if (gEventListeners.has(event)) {
     throw new Error("Duplicate event listener: " + event);
```
jasonLaster, line 262: if this is not adding value, we can delete it too.

src/devtools/client/debugger/src/client/commands.ts
```diff
@@ -172,18 +172,11 @@ function runAnalysis(location: SourceLocation, options: InitialBreakpointOptions
...
-    setLogpoint(
-      logGroupId!,
-      { sourceId, line, column: column! },
-      logValue!,
-      condition!,
-      showInConsole
-    );
+    setLogpoint(logGroupId!, { sourceId, line, column: column! }, logValue!, condition!);
   } else {
-    setLogpointByURL(logGroupId!, sourceUrl!, line, column!, logValue!, condition!, showInConsole);
```
jasonLaster, line 186: wonder how much of this code we could kill?

packages/protocol/thread/analysis.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+    params.sessionId
+  );
+  const points: PointDescription[] = [];
+  const results: AnalysisEntry[] = [];
+  gAnalysisCallbacks.set(analysisId, {
+    onEvent: receivedPoints => {
+      points.push(...receivedPoints);
+    },
+    onCreate: console.log,
```
jasonLaster, line 46: lol

packages/protocol/thread/analysis.ts
```diff
@@ -0,0 +1,91 @@
...
+    onResults: receivedResults => {
+      results.push(...receivedResults);
+    },
+  });
+  return {
+    analysisId,
+    async findPoints() {
+      try {
+        await sendMessage("Analysis.findAnalysisPoints", { analysisId }, params.sessionId);
```
jasonLaster, line 55: This could be `Analysis.findAnalysisPoints()`

src/ui/reducers/analyses.ts
```diff
@@ -0,0 +1,151 @@
...
+  analysisResult,
+  Location,
+  PointDescription,
+  AnalysisEntry,
+} from "@replayio/protocol";
+import { createEntityAdapter, createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+import { AnalysisError } from "protocol/thread/analysis";
+
+enum AnalysisStatus {
```
jasonLaster, line 12: I like where this is headed, but wonder if Analysis should be a first class reducer or if the state should belong to a LogPoint and other application concepts.

This is what Josh and I drew up earlier as a reference
https://gist.github.com/jasonLaster/78fa5b41b7a9c5e9f7986182a60735ed

jasonLaster (commented):
💃

packages/protocol/thread/analysis.ts
markerikson: I BLAME JOSH!

(this was largely copy-pasted from his initial POC)

packages/protocol/thread/analysis.ts
```diff
@@ -0,0 +1,92 @@
...
+        };
+      } catch {
+        return {
+          points,
+          error: AnalysisError.TooManyPointsToFind,
+        };
+      }
+    },
+    async addLocation(location: Location) {
+      await sendMessage("Analysis.addLocation", { analysisId, location }, params.sessionId);
```
jcmorrow, line 69: Samesies for all of these. And yes, I know, I wrote this code 😬

src/ui/setup/store.ts
```diff
@@ -88,26 +106,55 @@ const sanitizeStateForDevtools = <S>(state: S) => {
...
+      draft.preview.preview = OMITTED;
     }
   });
 
   return sanitizedState as S;
 };
 
 const sanitizeActionForDevTools = <A extends AnyAction>(action: A) => {
-  // Actions may contain `ValueFront` objects
+  // Actions may contain `ValueFront` objectsaction);
```
jcmorrow, line 135: looks like an accidental paste here

jcmorrow (approved):
I would be happy to have this in 😄 

On top of it we could:

- Stop running analyses that we know are going to error because "too many points" and instead tell the user to focus harder
- Actually not run analyses outside of the focus window
- Show actual error messages like "10,000+ hits" when there are *actually* that many hits, and just "Error" when we get other errors.

packages/protocol/thread/analysis.ts
```diff
@@ -0,0 +1,92 @@
+import { AnalysisEntry, Location, PointDescription } from "@replayio/protocol";
+import { AnalysisParams } from "../analysisManager";
+import { gAnalysisCallbacks, sendMessage } from "../socket";
+
+export interface Analysis {
+  analysisId: string;
+  addLocation: (location: Location) => Promise<void>;
```
jcmorrow, line 7: For posterity, I'm going to note that soon `addPoints` will probably join this type.

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -16,23 +18,80 @@ import type { Breakpoint, SourceLocation } from "./types";
...
 
 type LocationWithoutColumn = Omit<Location, "column">;
+
+export enum AnalysisStatus {
+  // Happy path
+  Created = "Created",
+  LoadingPoints = "LoadingPoints",
+  PointsRetrieved = "PointsRetrieved",
+  LoadingResults = "LoadingResults",
+  Completed = "Completed",
```
jcmorrow, line 28: Also for posterity, I'm going to note that while this set of Status's make sense, it's not the only set of options. For instance, I could picture having separate statuses for things like "PointsFound" and "Completed" and "WasReleased" and "Errored", because those things are not all mutually exclusive. Anyways, that's a problem for future us - those folks really know what they are doing.

jcmorrow (approved):

packages/protocol/socket.ts
```diff
@@ -224,6 +235,28 @@ export function addEventListener<M extends EventMethods>(
...
 ) {
+  if (event === "Analysis.analysisPoints") {
+    gEventListeners.set(event, ({ analysisId, points }: analysisPoints) => {
+      const callbacks = gAnalysisCallbacks.get(analysisId);
+      if (callbacks) {
+        callbacks.onEvent(points);
+      } else {
+        handler({ analysisId, points });
+      }
+    });
```
bvaughn, line 246: Why is this an if/else? Shouldn't we always call the `handler` _in addition to_ any callback that was registered? This seems unexpected to me.

packages/protocol/thread/analysis.ts
bvaughn: I've been wondering why we sometimes use one and sometimes the other.

PR #6981, opened 2022-05-26
Removed ui/utils/prefs import from protocol/threads

bvaughn:
Follow on to #6979 and #6973.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -38,7 +38,6 @@ import {
   findAnnotationsResult,
   getHitCountsParameters,
 } from "@replayio/protocol";
-import { features } from "ui/utils/prefs";
 import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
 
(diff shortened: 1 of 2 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -118,6 +117,12 @@ declare global {
   }
 }
 
+// Experimental feature
+let repaintAfterEvaluations: boolean = false;
+export function setRepaintAfterEvaluations(value: boolean): void {
+  repaintAfterEvaluations = value;
+}
```
bvaughn, line 124: This pattern is a tiny bit awkward I guess but at least it doesn't leak references to outside packages?

pages/_app.tsx
```diff
@@ -134,6 +136,11 @@ if (isMock()) {
   configureMockEnvironmentForTesting();
 }
 
+// Expose app feature flags to the protocol through an app-agnostic API.
+if (features.repaintEvaluations) {
+  setRepaintAfterEvaluations(true);
+}
```
bvaughn, line 142: I assume this (during app initialization) is as good a place as any to set this logic?

markerikson: I'd probably lean towards putting it into `devtools.ts` with the rest of the recording-specific setup, but it'll work either way

markerikson (approved):

pages/_app.tsx
bvaughn: Thanks! 👍🏼 

Looking at `devtools.ts` I don't see a place that stands out to me as better for it but I don't feel strongly at all if we move it there (or somewhere else)

PR #6985, opened 2022-05-27
Remove external imports from protocol graphics

bvaughn:
Follow up to #6955 and a bunch of other cleanup PRs.

This PR **removes the last of the external references** from the new "protocol" package.
- [x] 7008ab6: Moved `precacheScreenshots` thunk out of the `protocol/graphics` package and into `ui/actions/timeline`.
- [x] b2a3406: Replaced various `store.dispatch` calls in `protocol/graphics` with optional event handlers (added in `devtools` setup).

Commits are atomic and so they can be reviewed individually if that's easier.

I think we should still invest in rearchitecting that package, but this at least unblocks a proof of concept for #6932.

### DIFF
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -15,8 +15,6 @@ import { getVideoNode } from "./videoNode";
 
 const MINIMUM_VIDEO_CONTENT = 5000;
 
-const { features } = require("ui/utils/prefs");
-
 // Temporary experimental feature flag
 let syncVideoPlaybackExperimentalFlag: boolean = false;
 export function setSyncVideoPlaybackExperimentalFlag(value: boolean): void {
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -2,11 +2,40 @@
 import { TimeStampedPoint, MouseEvent, ScreenShot, PaintPoint } from "@replayio/protocol";
 import { decode } from "base64-arraybuffer";
 import ResizeObserverPolyfill from "resize-observer-polyfill";
-import { UIStore, UIThunkAction } from "ui/actions";
-import { setCanvas, setEventsForType, setVideoUrl } from "ui/reducers/app";
-import { pointsReceived, setPlaybackPrecachedTime, setPlaybackStalled } from "ui/reducers/timeline";
-import { getPlaybackPrecachedTime, getRecordingDuration } from "ui/reducers/timeline";
+
+// TypeScript types only ///////////////////////////////////////////////////////////////////////////////////////
+
+import { UIStore } from "ui/actions";
 import { Canvas } from "ui/state/app";
+
+// Application events //////////////////////////////////////////////////////////////////////////////////////////
+
+// setupGraphics adds this to onRefreshGraphics() callback which gets called by refreshGraphics()
+// We could inject an external callback for this or emit an event.
+import { setCanvas } from "ui/reducers/app";
+
+// Gets called as part of client.Session.addMouseEventsListener() to notify of "mousedown" events
+// We could inject a "handleMouseDown" callback for this or emit a "mousedown" event.
+import { setEventsForType } from "ui/reducers/app";
+
+// Called by VideoPlayer.createUrl() in response to a new video fragment being appended.
+// We could inject a "handleUrl" callback for this or emit a "url" event.
+import { setVideoUrl } from "ui/reducers/app";
+
+// Called whenever new points are received.
+// Currently this means client.Graphics.addPaintPointsListener or client.Session.addMouseEventsListener
+// We could inject a "handlePoints" callback for this or emit a "points" event.
+import { pointsReceived } from "ui/reducers/timeline";
+
+// Called by the repaint() method to indicate playback stalled (timeout) or completed (after async code).
+// We could inject a "handlePlaybackStatus" callback for this or emit a "playback" event.
+import { setPlaybackStalled } from "ui/reducers/timeline";
+
+// Called by ThreadFront "paused" event handler (after some other things).
+// We could inject a "precacheScreenshots" callback for this or emit a "precacheScreenshots" event.
+import { precacheScreenshots } from "ui/actions/timeline";
+
+/////////////////////////////////////////////////////////////////////////////////////////////////////
 
 import { DownloadCancelledError, ScreenshotCache } from "./screenshot-cache";
 import { ThreadFront } from "./thread";
@@ -29,6 +58,7 @@ declare global {
 }
 
 export const screenshotCache = new ScreenshotCache();
+
 const repaintedScreenshots: Map<string, ScreenShot> = new Map();
 
 interface Timed {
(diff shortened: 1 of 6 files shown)


src/ui/actions/timeline.ts
```diff
@@ -337,7 +341,7 @@ function playback(startTime: number, endTime: number): UIThunkAction {
     const shouldContinuePlayback = () => getPlayback(getState());
     prepareNextGraphics();
 
-    Video.play();
+    Video.play(currentTime);
```
bvaughn, line 344: Is this value the same as what `protocol/graphics` was using before?

https://github.com/RecordReplay/devtools/blob/ae52cb0d709e2a1b480697d8afa9243b05c60f8a/packages/protocol/graphics.ts#L232

src/ui/components/Video.tsx
```diff
@@ -75,7 +75,7 @@ function Video({
     if (playback) {
       refreshGraphics();
       VideoPlayer.seek(currentTime);
-      VideoPlayer.play();
+      VideoPlayer.play(currentTime);
```
bvaughn, line 78: Is this value the same as what `protocol/graphics` was using before?

https://github.com/RecordReplay/devtools/blob/ae52cb0d709e2a1b480697d8afa9243b05c60f8a/packages/protocol/graphics.ts#L232

packages/protocol/graphics.ts
```diff
@@ -652,63 +663,48 @@ export async function getFirstMeaningfulPaint(limit: number = 10) {
...
-// precache this many milliseconds
-const precacheTime = 5000;
-// startTime of the currently running precacheScreenshots() call
-let precacheStartTime = -1;
+////////////////////////////////////////////////////////////////////////////////////////////////
+// The callbacks below allow external code to observe specific events.
+// This design pattern is a short term strategy to remove external imports from this package.
+//
+// TODO We should revisit this as part of a larger architectural redesign (#6932).
+////////////////////////////////////////////////////////////////////////////////////////////////
```
bvaughn, line 671: I laugh in the face of your 80-character limits 😈

jasonLaster: 🤣

packages/protocol/graphics.ts
```diff
@@ -652,63 +663,48 @@ export async function getFirstMeaningfulPaint(limit: number = 10) {
...
+  scale: number;
+  top: number;
+  width: number;
+}
 
-      const paintHash = gPaintPoints[index].paintHash;
-      if (!screenshotCache.hasScreenshot(paintHash)) {
-        const graphicsPromise = getGraphicsAtTime(time, true);
+let onMouseDownEvents: (events: MouseEvent[]) => void;
+export function setMouseDownEventsCallback(callback: typeof onMouseDownEvents): void {
```
jasonLaster, line 683: agreed that this is temporary, but it might feel a bit better to expose a simpler event listener interface through `addListener`, `emitEvent`.

```js
function addListener(event, callback) {
  gCallbacks[event]?.push(callback);
}
function emitEvent(event, ...args) {
  for (const callback in gCallbacks[event]) {
    callback(...args)
  }
}
```

src/ui/actions/timeline.ts
```diff
@@ -707,3 +711,64 @@ export function toggleFocusMode(): UIThunkAction {
     }
   };
 }
+
+const PRECACHE_DURATION: number = 5000;
+
+let precacheStartTime: number = -1;
+
+export function precacheScreenshots(startTime: number): UIThunkAction {
```
jasonLaster, line 719: smart to move this out of protocol

src/ui/actions/timeline.ts
jasonLaster: It looks like it could be different. It'd be easier to say if there were a replay :)

packages/protocol/graphics.ts
markerikson: FWIW we have at least two existing `EventEmitter` classes in this codebase.  One is in `devtools/`, plain JS, and more extensive behavior.  The other is in `ui/utils/`, is TS, and same basic API but simpler internals.

(but neither of them gives fully typed events, which is why I was debating pulling in a _third_ implementation at some point...)

markerikson (commented):
Changes seem reasonable as a whole, other than Jason's comments

packages/protocol/graphics.ts
markerikson: That's a modified "Hungarian Notation" syntax (a concept that goes _allll_ the way back to old-school Microsoft C/C++ days):

- `g` for global variables
- `m` for class members

etc

bvaughn: Fair enough but...why are we using it? These concepts aren't any more "global" than any of our other module level exports :D

bvaughn: TBH I thought the "g" had something to do with "graphics" bc' it was in that file 😆

jcmorrow: You'll see that notation used a lot more in the backend. I suspect the "how did it get that way" has to do with who originally wrote it and whether or not they preferred that style, but does not have a lot of firm reasons beyond that.

bvaughn: Thanks for the background info! 😄

bvaughn: > agreed that this is temporary, but it might feel a bit better to expose a simpler event listener interface through `addListener`, `emitEvent`.

I'm not opposed to this, but I've gotten bogged down in TypeScript generics trying to write a strongly typed event dispatcher API and I'd rather punt on that as far as this PR goes.

markerikson: Yeah. Early in my career I used it because I was writing C++/Java and that's how the example projects I learned from wrote things.  Later on I dropped it because it didn't seem as useful (and especially as I got into JS and more FP-inspired styles).

jcmorrow (approved):

packages/protocol/graphics.ts
```diff
@@ -39,7 +33,7 @@ interface Timed {
 
 // Given a sorted array of items with "time" properties, find the index of
 // the most recent item at or preceding a given time.
-function mostRecentIndex<T extends Timed>(array: T[], time: number): number | undefined {
+export function mostRecentIndex<T extends Timed>(array: T[], time: number): number | undefined {
```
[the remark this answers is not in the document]
bvaughn: Happy to see this change made in a follow up PR. Strongly prefer to limit the complexity and the number of things I'm changing in this PR :)

packages/protocol/graphics.ts
jcmorrow: Also: https://baymard.com/blog/line-length-readability

bvaughn: I don't know how applicable that article is to code/comments. Reading code/comments seems like a pretty different task than reading a website's "body text" to me.

If we think it _is_ applicable then even 80 chars is too long too and we should aim for 50-75.

src/ui/actions/timeline.ts
bvaughn: Okay for now I just matched the pre-existing behavior for this callsite.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -163,13 +173,35 @@ export default async function DevTools(store: AppStore) {
...
+  });
+  setPointsReceivedCallback((points: TimeStampedPoint[]) => {
+    store.dispatch(pointsReceived(points));
+  });
+  setRefreshGraphicsCallback((canvas: Canvas) => {
+    store.dispatch(setCanvas(canvas));
+  });
+  setVideoUrlCallback((url: string) => {
+    store.dispatch(setVideoUrl(url));
+  });
```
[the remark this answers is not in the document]
jasonLaster: sounds good

PR #6987, opened 2022-05-27
[WIP] Use npm ci 

yeion7:
Github recommends to use `npm ci` in continuous integration environments

This is a test to validate if it works

### DIFF
--- a/.github/actions/mockTest/action.yml
+++ b/.github/actions/mockTest/action.yml
@@ -3,8 +3,13 @@ description: "Run mock tests"
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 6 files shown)


ryanjduffy:
@yeion7 - Thanks for the PR! We recently reworked some of these to use https://github.com/bahmutov/npm-install to get out of the box caching and use `npm ci`. Wish I had seen this PR before then and we could have collaborated on it.

Nonetheless, there could be a few improvements here so can you rebase your branch and see what shakes out. Definitely will appreciate the help getting our CI runs as efficient as possible.

PR #6989, opened 2022-05-27
Features and prefs now emit change events

bvaughn:
How are we supposed to actually observe changes to features/prefs? The observer code seems to _exist_ but it doesn't appear to be wired up to work.

### DIFF
--- a/src/devtools/client/shared/prefs.js
+++ b/src/devtools/client/shared/prefs.js
@@ -82,9 +82,11 @@ function get(cache, prefType, prefsRoot, prefName) {
  * @param string prefName
  * @param any value
  */
-function set(cache, prefType, prefsRoot, prefName, value) {
+function set(self, cache, prefType, prefsRoot, prefName, value) {
   Services.prefs["set" + prefType + "Pref"]([prefsRoot, prefName].join("."), value);
   cache.set(prefName, value);
+
+  self.emit(`${prefName}-changed`, value);
 }
 
 /**
@@ -130,7 +132,7 @@ function map(
 
   Object.defineProperty(self, accessorName, {
     get: () => serializer.in(get(cache, prefType, prefsRoot, prefName)),
-    set: e => set(cache, prefType, prefsRoot, prefName, serializer.out(e)),
+    set: e => set(self, cache, prefType, prefsRoot, prefName, serializer.out(e)),
   });
 }
 
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -145,6 +145,14 @@ if (features.videoPlayback) {
   setSyncVideoPlaybackExperimentalFlag(true);
 }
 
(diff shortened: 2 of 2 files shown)


src/devtools/client/shared/prefs.js
```diff
@@ -82,9 +82,11 @@ function get(cache, prefType, prefsRoot, prefName) {
  * @param string prefName
  * @param any value
  */
-function set(cache, prefType, prefsRoot, prefName, value) {
+function set(self, cache, prefType, prefsRoot, prefName, value) {
   Services.prefs["set" + prefType + "Pref"]([prefsRoot, prefName].join("."), value);
   cache.set(prefName, value);
+
+  self.emit(`${prefName}-changed`, value);
```
bvaughn, line 89: This is what the docs say it's supposed to do, right? Am I missing something?
https://github.com/RecordReplay/devtools/blob/ec4ccc7f06e265764a6b3730a925bc17e294fcf8/src/devtools/client/shared/prefs.js#L24-L28

You'd think that `Services` would do this, because it calls a method called "notify":
https://github.com/RecordReplay/devtools/blob/ec4ccc7f06e265764a6b3730a925bc17e294fcf8/src/devtools/shared/services.js#L231-L237

But I don't know what `parent.notify` does. It doesn't seem to emit an event that I know how to subscribe to.
https://github.com/RecordReplay/devtools/blob/ec4ccc7f06e265764a6b3730a925bc17e294fcf8/src/devtools/shared/services.js#L270-L280

bvaughn: There appears to be an observer API but I do not see how the features object (or it's `PrefsHelper` really) passes anything through to it. It doesn't look like it's all the way wired up.

bvaughn:
@jaril Kind of an open question. Noticed things seemed to be disconnected so I opened this PR to come back to it but haven't yet. It's probably fine to sit for a while. Doesn't seem super pressing.

jaril:
👍 will check back in a ~week!

src/devtools/client/shared/prefs.js
jcmorrow: I'm not sure if this is relevant, but I use `addObserver` in both `useFeature` and `useStringPref`, and both of those *do* work (or at least, if you change the value via the console, you will see a component that uses that hook re-render in response to the observer going off).

bvaughn: That's interesting! I wonder why I wasn't seeing it work.

bvaughn: Ah, I think maybe I wasn't doing the key -> "full key" conversion step you were doing.

bvaughn:
Going to close this for now. Seems like there _is_ a way to observe these values based on `useStringPref` and `useFeature` hooks.

PR #6992, opened 2022-05-29
Architecture proof of concept

bvaughn:
### Related to the architecture proposal #6932

This branch explores a different architectural direction for Replay devtools. The branch attempts to illustrate the following:
- [x] What it would be like to write and maintain React views using this state management approach (see `packages/bvaughn-architecture-demo/components/console/MessagesList.tsx`).
- [x] The added UX benefits that React concurrent mode can provide if we go all-in on Suspense and concurrent compatibility.
- [x] How easy it can be to write tests for React components using this pattern.

---

Here is a walk through of the new design and implementation: https://www.loom.com/share/a8d5fa276cfb4f0181f07378bfa9636e

If we took this idea as far as I'd like to take it, our state architecture might look something like this:

State architecture diagram

---

For the purpose of this branch, I'm focusing on one data type: messages. Message are an interesting case study, because
1. Their data comes from both the backend (messages stored as part of the recording) and global state (user-defined log points added in memory).
5. Despite being in-memory, log points sometimes also require analysis to be run on the backend to evaluate complex expressions.
6. Focus range impacts which messages we show both in terms of remote filtering ([`findMessagesInRange`](https://static.replay.io/protocol/tot/Console/#method-findMessagesInRange)) and in-memory filtering (for long points or in the case of no overflow where we are able to fetch all recording messages in memory).
7. Additional client side filtering logic may also be applied (e.g. show warnings or errors).

I think the above constraints make messages a good smoke test candidate for this new architecture. If we can make messages work efficiently and with a nice API, then we can probably do so for other things as well.

---

To run the demo app after checking out this branch:
```sh
npm install
cd packages/bvaughn-architecture-demo/
npm run dev

# View a public Replay using a URL like:
# http://localhost:3000/?recordingId=48ca3af8-8493-4297-922c-d0ddf6cf44e6
```

---

Note that I think https://github.com/facebook/react/issues/24634 is probably a blocker for this issue since certain types of errors can cause hot cycles (in DEV with Next.js). Hopefully this issue will be resolved soon on the React side or on the Next side.


packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -0,0 +1,80 @@
...
+        time: startTime,
+      },
+      end: {
+        point: endPoint,
+        time: endTime,
+      },
+    };
+  }
+
+  const { countAfter, countBefore, didOverflow, messages } = getMessages(sessionId, focusMode);
```
bvaughn, line 44: Here's another key part of the proposal: Values that come from the backend (like messages) use Suspense patterns to make them easy to consume from the view's perspective. (Beside the scenes, if `getMessages` needs to fetch/refetch data, it can just _suspend_ to do that.)

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  // Memoized selector that joins log points and messages and filters by criteria (e.g. type)
+  // Note that we are intentionally not storing derived values like this in state.
+  const filteredMessages = useFilteredMessages(messages, {
+    ...levelFlags,
+    filterByText,
```
bvaughn, line 50: Here's another key part of the proposal: _Derived values_ (like filtered messages) are just memoized selectors. We don't need to store them in parallel "state" (or Redux).

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  // This component only needs to render a pending UI when a focus changes,
+  // because this might require an async backend request.
+  // Filter text changes are always processed synchronously by useFilteredMessages(),
+  // so dimming the UI would just cause a short flicker which we can avoid.
+  const isTransitionPending = isFocusTransitionPending;
+
+  return (
+    <div className={isTransitionPending ? styles.ContainerPending : styles.Container}>
```
bvaughn, line 60: A nice side benefit of building our UI on top of Suspense and [transitions](https://reactjs.org/docs/hooks-reference.html#usetransition) is that it's super easy to tweak the UI slightly (like dimming it) to indicate that the data being shown may be a little stale and we're updating it.

packages/bvaughn-architecture-demo/pages/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }),
+    [deferredFilterByText, filterByText, isTransitionPending, levelFlags, updateFilters]
+  );
+
+  // TODO Once we have a client implementation to interface with Replay backend,
+  // the app can inject a wrapper one that also reports cache hits and misses to this UI in a debug panel.
+
+  return (
+    <FocusContext.Provider value={focusContext}>
+      <ConsoleFiltersContext.Provider value={consoleFiltersContext}>
```
bvaughn, line 110: This component uses React primitives (like `useState` and context) to setup and share "global" state with other parts of the app. This gives us a nice point to integrate with other React APIs (like [transitions](https://reactjs.org/docs/hooks-reference.html#usetransition)).

packages/bvaughn-architecture-demo/src/contexts.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  // Filter by text is about to be updated as part of a transition;
+  // UI that consumes the focus for Suspense purposes may wish want reflect the temporary pending state.
+  isTransitionPending: boolean;
+
+  // Types of console messages to include (or filter out).
+  levelFlags: ConsoleLevelFlags;
+
+  update: (filterByText: string, levelFlags: ConsoleLevelFlags) => void;
+};
```
bvaughn, line 29: These contexts are nice for sharing state between different parts of the app (without "props drilling"). I guess they also give us a nice _super lightweight_ way to statically render our UI if we want to for visual testing (but I haven't thought about this use case much yet).

packages/bvaughn-architecture-demo/src/utils/suspense.ts
```diff
@@ -0,0 +1,68 @@
...
+      });
+    },
+  };
+
+  return wakeable;
+}
+
+// Helper function to read from multiple Suspense caches in parallel.
+// This method will re-throw any thrown value, but only after also calling subsequent caches.
+export function suspendInParallel(callbacks: Function[]): void {
```
bvaughn, line 54: We could use this as kind of a micro-optimization to fetch points in parallel. I didn't use it for now (because I didn't want to make "suspensey" React component code harder to read).

packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
```diff
@@ -0,0 +1,64 @@
...
+
+function createMap(): TimeToRecordMap {
+  return new Map();
+}
+
+function getRecordMap(): TimeToRecordMap {
+  return getCacheForType(createMap);
+}
+
+export function getClosestPointForTime(time: number, sessionId: SessionId): ExecutionPoint {
```
bvaughn, line 21: This is more of a typical Suspense cache implementation. No extra business logic like soft focus, just a map of key (time) to value (execution point) that we build over time– _suspending_ and loading information from the backend when we need to.

packages/bvaughn-architecture-demo/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+export type Record<T> = PendingRecord | ResolvedRecord<T> | RejectedRecord;
+
+// This type defines the subset of the Promise API that React uses (the .then method to add success/error callbacks).
+// You can use a Promise for this, but Promises have a downside (the microtask queue).
+// You can also create your own "thennable" if you want to support synchronous resolution/rejection.
+export interface Thennable {
+  then(onFulfill: () => any, onReject: () => any): void | Thennable;
+}
+
+export interface Wakeable extends Thennable {
```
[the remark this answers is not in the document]
bvaughn: Yes. You could use a native `Promise` instead of "wakeable". The _type_ doesn't matter. React doesn't do anything with the value the thing _resolves to_. The resolution is just a signal to React to try renderin again.

package.json
```diff
@@ -124,7 +124,7 @@
     "@storybook/react": "^6.4.21",
     "@tailwindcss/forms": "^0.5.0",
     "@testing-library/jest-dom": "^5.16.3",
-    "@testing-library/react": "^12.1.4",
+    "@testing-library/react": "^13.3.0",
```
bvaughn, line 127: Upgrading this fixed the `createRoot` vs `render` warning and our tests all still passed so...

packages/bvaughn-architecture-demo/components/console/Focuser.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        rangeForDisplay: null,
+      },
+      sessionContext: {
+        duration: 60_000,
+      },
+    });
+
+    expect(await screen.queryByText("Focus off")).toBeInTheDocument();
+    expect(await screen.queryByText("0:00 – 1:00")).toBeInTheDocument();
+  });
```
bvaughn, line 21: This is an example of the approach we might use to render a small part of our app to test in isolation. The `renderFocused` helper method could be shared between all tests. It accepts optional params which can be used to define a test scenario.

packages/shared/utils/testing.ts
```diff
@@ -0,0 +1,187 @@
...
+} from "@replayio/protocol";
+
+export const DEFAULT_SOURCE_ID = "fake-source-id";
+export const DEFAULT_SOURCE_URL = "fake-source-url";
+
+let uidCounter = 0;
+
+export type MessageTuple = ["Console.newMessage", Message];
+
+export function createConsoleMessage({
```
bvaughn, line 28: I moved these methods into the shared package so the demo app tests can use them too.

bvaughn:
After chatting with Jason, I'm going to go land this prototype for now. (It doesn't impact our devtools app any.) I'd love to get others playing with it and poking at it (and maybe coming up with their own POCs built on top of it).

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const { filterByText, levelFlags } = useContext(ConsoleFiltersContext);
+  const { range, isTransitionPending: isFocusTransitionPending } = useContext(FocusContext);
+
+  let focusRange = null;
+  if (range !== null) {
+    const [startTime, endTime] = range;
+
+    const startPoint = getClosestPointForTime(replayClient, startTime);
+    const endPoint = getClosestPointForTime(replayClient, endTime);
```
Andarist, line 30: q: if I understand correctly - this is using a simple **waterfall** fetch on render, right? That is certainly fine, especially for demo purposes. I just wonder what's the idiomatic/canonical answer for removing the waterfall but I guess the answer is just to preload both of those and that can be done in parallel so the potential "render waterfall" here wouldn't quite matter. Is that correct understanding of those patterns?

Andarist: I guess that, partially, you have also mentioned the waterfall concern [here](https://github.com/RecordReplay/devtools/pull/6992/files#r885643829) - but based on that helper's API I'm not actually sure how to use it 😅 

It has to be called from render as that is the only place from which one can suspend, right? But I'm surprised that it doesn't return any values cause I've imagined that it could be used like this:
```js
const [startPoint, endPoint] = suspendInParallel([
  () => getClosestPointForTime(replayClient, startTime),
  () => getClosestPointForTime(replayClient, endTime)
])
```

It also seems to "leak" suspense to renders whereas usually it's promoted as something completely hidden from the call site (which is cool!). So perhaps I'm misunderstanding completely how this can be used.

However, I'm still curious if the preloading logic would remove the need for such a helper altogether cause it seems to me that it would but again, I feel like I might miss some details here.

packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
```diff
@@ -0,0 +1,66 @@
...
+// For example, when we get Paints they have a corresponding point (and time) which could be pre-loaded here.
+
+type TimeToRecordMap = Map<number, Record<ExecutionPoint>>;
+
+function createMap(): TimeToRecordMap {
+  return new Map();
+}
+
+function getRecordMap(): TimeToRecordMap {
+  return getCacheForType(createMap);
```
Andarist, line 18: It would be very interesting to learn about the advantages of this React-owned cache. What it can do that the user land's implementation can't implement? Or what it automates that a user land implementation would have to craft on their own?

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
tjallingt: The `suspendInParallel` helper could help but doesn't (currently) return the values https://github.com/RecordReplay/devtools/pull/6992#discussion_r885643829

```js
const [startPoint, endPoint] = suspendInParallel([
  getClosestPointForTime(replayClient, startTime),
  getClosestPointForTime(replayClient, endTime),
]);
```

I find it a bit tricky to spot the "suspenders", and by extension its tricky to see where to apply this helper.

packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
bvaughn: I don't know if this will be the final API for React Suspense caches. It's the one React DevTools has been using for a year or so now but it may change before public release.

React owning the underlying cache (map) allows it to manage invalidation (e.g. invalidate this bit of data and re-fetch it at a lower priority) without preventing it from re-rendering a high priority update in the meanwhile. If we used a global cache, it would be unable to do the high-priority update because the global cache would now be missing data.

In Replay's case, this doesn't matter much because we don't need to invalidate this cache for example. It grows over time but once added to it, data is never "wrong" or "outdated".

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
bvaughn: I have some small follow up changes I'm going to make this morning (actually woke up a few times thinking about them last night 😭) and I'll toss this one in as well 😁

bvaughn: TBH the only real snag on `suspendInParallel` is that I don't know how to manage the return type (to be anything other than `any[]`)

Andarist: Oh, I'm your guy then :wink:

```ts
type AnyFunction = (...args: any[]) => any;

declare function suspendInParallel<T extends AnyFunction[]>(
  readers: [...T]
): { [K in keyof T]: ReturnType<Extract<T[K], AnyFunction>> };
```

Note that you might need to cast the final result within the implementation itself because TS won't be able to verify that it matches this annotation. For public consumers this should work great though.

Andarist: Perhaps I should explain this weirdo...

1. TS doesnt really have a tuple constraint (which is super annoying) so we need to use an array constraint on the generic `T extends AnyFunction[]`
2. we need to make it to actually narrow the inline arrays used as arguments here to be of a tuple type and using `[...T]` is a neat trick to do it
3. then we proceed to the return type which is a mapped type - it allows us to iterate over the type per property. And that is why we had to make it infer those array arguments as tuples. We want to "unwrap" types per position. So we just iterate over keys of `T` and then do some type-level manipulation on a specific element (`T[K]`)
4. It should be enough to just use `ReturnType<T[K]>` here and in TS 4.7 it actually is enough... but that's a new improvement. That extra `Extract` just allows us to "cast" this at the generic level and force TS to accept `T[K]` as a type argument passed to `ReturnType` (as it only allows functions as type arguments)

bvaughn: Oh hell yes! Thank you for the type info and the explanation :)

PR #6993, opened 2022-05-29
Add protocol playground

jasonLaster:
I'm switching directions a bit. Instead of creating a page that stress tests a recording or multiple recordings, I'll be working on a general purpose protocol playground.

This will serve a couple of purposes:
1. Make it easy to run a pre-defined fuzzer program on a recording
2. Make it easy to share example scripts that use the protocol
3. Make it easy to visualize requests/responses from the protocol using the viewer

Progress
- [x] Create a new playground page
- [x] Create a session with a logged in user 
- [ ] Prompt users to login if they haven't already
- [x] Show the viewer
- [ ] Show an editor 
- [ ] Show the fuzzer and other pre-defined scripts
- [ ] Save a script in the url

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -68,6 +68,16 @@ module.exports = {
 
   webpack: (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
     const entry = config.entry;
+
+    // Disable NextJS normal WASM loading pipeline so onigasm can load properly
+    config.module.rules.push({
+      test: /\.wasm$/,
+      type: "asset/resource",
+      generator: {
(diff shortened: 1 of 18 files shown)


jcmorrow:
Looks like this build is failing, but I'd love to take a look at what you've got going on!

jasonLaster:
@jcmorrow would love your help getting the Editor integrated

jaril:
@jasonLaster bumping this. This sounds like it could be a project. Thoughts on next steps?

jasonLaster:
This is a side project and probably should not be on the roadmap.

Issue #6988, opened 2022-05-27
Comments no longer linking to correct source


ceceliacreates:
**Description:**
Recorded this replay on Monday and added comments. Now, the comments aren't linked to the correct source in all cases. Sometimes it opens the minified code (not sure if even accurate file location) with no source maps toggle, other time it doesn't open a file at all.

Console logs are still linked correctly, and the source files do exist in the Source Explorer.

Potentially unrelated, but the comment notification also appears briefly then disappears.

These comments worked earlier this week, so likely a recent regression.

Link to original replay: https://app.replay.io/recording/subscribe-to-updates-not-working--3afe0c21-954e-4df9-b960-a4e79630b7f5
**Steps to Reproduce:**

**Expected Result:**

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):
* 
https://app.replay.io/recording/replay-comments-not-linked-to-source-anymore--c3e3e1ae-f478-4541-81c1-d3c7ec0db31b

jcmorrow:
Closed with https://github.com/RecordReplay/backend/issues/5767 and  #6994

PR #6994, opened 2022-05-30
Workaround for changing sourceIds (#6988)

hbenl:


markerikson (approved):

jasonLaster:
Can we hold off on landing this fix on the frontend until we better understand what the problem is on the backend? There are bigger architectural concerns if the frontend does not trust sourceIds to be consistent and I'd rather not go down start a slippery slope.

jcmorrow:
I'm all for a deeper understanding, but I also think we should get this in. We are now at a point where comments have been created with both "versions" of sourceIDs. That would have been a nice world to avoid, but now that we are here, we are probably going to need to do this on the front-end for the rest of time (or until we expire all recordings with the old source IDs or something like that).

jcmorrow (approved):

loganfsmyth (approved):

jasonLaster:
👍 approved

PR #7008, opened 2022-05-31
Allow replay.prod in CSP

jazzdan:
This is necessary to enable easy testing of backend services in pre-prod


pages/_document.tsx
```diff
@@ -13,7 +13,7 @@ const csp = (props: any) => {
   const hash = cspHashOf(NextScript.getInlineScriptSource(props));
   return [
     `default-src 'self'`,
-    `connect-src 'self' https://api.replay.io wss://dispatch.replay.io https://telemetry.replay.io https://webreplay.us.auth0.com https://api-js.mixpanel.com https://*.sentry.io https://*.intercom.io wss://*.intercom.io https://*.launchdarkly.com https://*.logrocket.io https://*.lr-ingest.io https://*.logrocket.com https://*.lr-in.com https://api.stripe.com ${
+    `connect-src 'self' https://api.replay.io wss://dispatch.replay.io https://telemetry.replay.io ws://*.replay.prod http://*.replay.prod https://webreplay.us.auth0.com https://api-js.mixpanel.com https://*.sentry.io https://*.intercom.io wss://*.intercom.io https://*.launchdarkly.com https://*.logrocket.io https://*.lr-ingest.io https://*.logrocket.com https://*.lr-in.com https://api.stripe.com ${
```
ryanjduffy, line 16: I just added `http://*.replay.prod` to the isDev block (in #7004) below but forgot about `ws:`. At the moment, I don't think there's anywhere where `NODE_ENV` is "production" that would use `.replay.prod` so I think adding `ws://` to L20 below makes sense. WDYT?

jazzdan: Isn't `NODE_ENV` production for app.replay.io?

What I want to be able to do is, from the backend deploy screen craft a URL that the deploying person can click on to be brought to their pre-prod service. Since we don't have a stable staging environment for the front end, I think I need the production deploy to be able to talk to staging URLs.

But it's possible that I'm missing something!

ryanjduffy: I think it would work for some things but not all. The graphql api is only set via env variable (I think) so that requires a deploy time config.

We do have a staging env on vercel that we could configure to use preproduction rather than prod. That was the plan for it so perhaps we should make that change. Bit more to think through there but could be a good change.

jazzdan: Yeah there's a layering here of history that's resulting in several different behaviors:

1) In the beginning there was only the dispatcher. We allowed the dispatcher URL to be changed on the frontend to make testing the backend easier.
2) Then we added the graphql api (api.replay.io). We decided to make this a compile time variable because being able to craft a URL that anyone can change was pretty scary.
3) Then we added CSP rules, effectively restricting 1) and making it so that 2) isn't necessary and could just be a runtime variable, and the list of allowed URLs becomes a compile time variable.

I think at this point I'd be in favor of making all of the URLs runtime variables, and changing the CSPs depending on a compile time variable. But that might be more than we want to do right now.

I really like the idea of standardizing all of this under a frontend staging environment. Do you know where that is configured?

jazzdan: To close out this conversation: Ryan and I decided to go with this runtime variable approach since we'll need it to easily support ephemeral environments.

PR #7009, opened 2022-06-01
Record failed pause events

jcmorrow:
This has taken a long time, partially because there was a lot of complexity around how pauses get created, partially because I went off on side tangents to simplify things, and partially because I was getting distracted. Anyways, this is the second part to fixing https://github.com/RecordReplay/devtools/issues/6548 and https://github.com/RecordReplay/customer-support/issues/86. Both of those issues were from devtools trying to create a pause/precache stuff that was not loaded. And both of them failed two ways:

- First off, we could have known that those calls were not a good idea and not made them
- When they failed, we should have told the user there was a failure

As of https://github.com/RecordReplay/devtools/commit/85e046b8d08bfd7f915d65b08f65a6cfe7d36824 and https://github.com/RecordReplay/devtools/commit/a303c248836722afde84d135a423a379115d373c we should not be making requests for pauses that either cannot be created or will cause an excess of data to load outside of the currently focused window, and this PR will make sure that there are real error states in the pause panel when we have failed to pause.


src/devtools/client/debugger/src/utils/timings.js
```diff
@@ -6,6 +6,8 @@
 
 import zip from "lodash/zip";
 
+// Is this ever used?
```
jcmorrow, line 9: Left a comment so I could ask this at review - is this one of those "it's used by the test harness" situations?

src/devtools/client/debugger/src/actions/pause/fetchScopes.js
```diff
@@ -7,7 +7,7 @@
 import { getSelectedFrame, getFrameScope } from "../../selectors";
 import { PROMISE } from "ui/setup/redux/middleware/promise";
 
-export function fetchScopes(cx) {
+export function fetchScopes() {
```
markerikson, line 10: :speech_balloon: I'm still shaky on and we do and don't need this passed through

src/devtools/client/debugger/src/actions/pause/paused.js
```diff
@@ -12,17 +12,28 @@ import { trackEvent } from "ui/utils/telemetry";
...
   return async function (dispatch, getState, { client }) {
     let frames;
     try {
-      frames = await client.getFrames();
-    } catch (e) {}
-    dispatch({ type: "FETCHED_FRAMES", frames, cx });
+      frames = await client.getFrames(pause);
+      dispatch({ type: "FETCHED_FRAMES", frames, pauseId: pause.id });
+    } catch (e) {
+      console.error({ e });
```
markerikson, line 30: :hand: do we want to leave the `console.error` in?

src/devtools/client/debugger/src/actions/pause/paused.js
```diff
@@ -47,32 +62,45 @@ function fetchAsyncFrames(cx) {
...
+
+
+    const pause = ThreadFront.ensurePause(executionPoint, time);
+
+    dispatch({ type: "PAUSED", executionPoint, time, id: pause.id });
+
+    try {
+      await pause.createWaiter;
+    } catch (e) {
+      console.error(e);
```
markerikson, line 79: :hand: Ditto

src/devtools/client/debugger/src/actions/pause/paused.js
```diff
@@ -47,32 +62,45 @@ function fetchAsyncFrames(cx) {
...
+      dispatch(failedToCreatePause(pause.id));
+    }
+
+    console.log({ pause, waiter: pause.createWaiter });
 
-    await dispatch(fetchFrames(cx));
+    await dispatch(fetchFrames(pause));
 
-    const frame = getSelectedFrame(getState());
+    const frame = getSelectedFrame(state);
```
markerikson, line 87: :hand: Bit of a difference in semantics with this change. We went from reading the state _right now_, to reusing the state from _earlier_ at the start of the thunk, before we awaited anything. May not be critical, but _could_ be meaningful.  Just checking to see if that's what we want here.

src/devtools/client/debugger/src/actions/pause/paused.js
```diff
@@ -47,32 +62,45 @@ function fetchAsyncFrames(cx) {
...
+    dispatch({ type: "PAUSED", executionPoint, time, id: pause.id });
+
+    try {
+      await pause.createWaiter;
+    } catch (e) {
+      console.error(e);
+      dispatch(failedToCreatePause(pause.id));
+    }
+
+    console.log({ pause, waiter: pause.createWaiter });
```
markerikson, line 83: :hand: More logging

markerikson (commented):
Nothing blocking-blocking, but some stray bits of logging, and a couple questions on intended behavior.

src/devtools/client/debugger/src/actions/pause/paused.js
jcmorrow: I kind of like it personally.

src/devtools/client/debugger/src/actions/pause/paused.js
jcmorrow: This is something that *really* bugs me about this code. There's a lot of

```
setLocalState()
dispatch(loadAdditionalState())
...
// in the thunk
readTheLocalState()
fetchAdditionalDataBasedOnLocalState()
...
```

It all just assumes that you can dispatch actions and the right thing will happen because you mutated some state right before dispatching. It would be a *lot* clearer and easier to follow if it was like `dispatch(loadAdditionalState(localState))`. Anyways, all that to say that yes, this *is* an intentional change. Right before we dispatch `paused` we set `ThreadFront.currentExecutionPoint` and `ThreadFront.currentTime`, and then when we call dispatch `fetchFrames` we tell `ThreadFront` to go create a pause for its `currentExecutionPoint` and `currentTime` (and we hope those are the same as they were at the beginning of the action). I don't think it's wise to change this all in one fell swoop, but I can at least try and make it behave a little more predictably 🤷

src/devtools/client/debugger/src/actions/pause/fetchScopes.js
markerikson: It's only used by the "context middleware", as a central checking point.  (I think the idea is that some actions may be irrelevant if the app has moved on to a new pause situation by the time a response returns.)

jcmorrow: ah yes, I had forgotten that some middleware was referencing it. I think we should have other methods for solving these problems.

src/devtools/client/debugger/src/actions/pause/paused.js
bvaughn: This pattern/abiguity is another type of thing that would be nice to address when we tackle the architecture.

src/devtools/client/debugger/src/utils/timings.js
bvaughn: Does not look like it's being referenced anywhere, nor is it writing any values to the global (😅 )

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/index.js
```diff
@@ -134,9 +161,13 @@ Frames.contextTypes = { l10n: PropTypes.object };
 
 const mapStateToProps = state => ({
   cx: getThreadContext(state),
+  currentTime: getCurrentTime(state),
   frames: getCallStackFrames(state),
   framesLoading: getFramesLoading(state),
+  framesErrored: getFramesErrored(state),
   frameworkGroupingOn: getFrameworkGroupingState(state),
+  pauseErrored: state.pause.pauseErrored,
+  pauseLoading: state.pause.pauseLoading,
```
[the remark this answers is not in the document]
markerikson: FWIW I think we actually _overuse_ selectors a bit.

When I wrote a "Using Selectors" page for the Redux docs, I actually said "_don't_ create a separate selector for every single tiny field - you're basically recreating Java at that point" :)

https://redux.js.org/usage/deriving-data-selectors#balance-selector-usage

In this case I don't know if we have pre-written selectors for those fields or not, so I assume Josh just threw those lookups in here.

I'll also note that it's safer to do straight lookups if a file is TS... but this file is still JS :)

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -145,6 +169,39 @@ function update(state = createPauseState(), action: AnyAction) {
...
+      };
+      return {
+        ...state,
+        pauseErrored: true,
+        pauseLoading: false,
+        isWaitingOnBreak: false,
+        selectedFrameId: undefined,
+        frames: undefined,
+        frameScopes: {},
+        executionPoint,
```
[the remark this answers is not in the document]
jcmorrow: I have no idea why, this file was doing this already, and I was a little terrified that the assignment might be doing something that I did not understand.

src/devtools/client/debugger/src/actions/pause/paused.js
```diff
@@ -47,32 +62,42 @@ function fetchAsyncFrames(cx) {
...
+
+    const pause = ThreadFront.ensurePause(executionPoint, time);
+
+    dispatch({ type: "PAUSED", executionPoint, time, id: pause.id });
+
+    try {
+      await pause.createWaiter;
+    } catch (e) {
+      console.error(e);
+      dispatch(failedToCreatePause(pause.id));
```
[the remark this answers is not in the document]
jcmorrow: Good question. I kind of like that we will also mark frames as failing to be fetched, but we could also dispatch that immediately and then return, which I think would be good 👍

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/index.js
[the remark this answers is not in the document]
bvaughn: Totally makes sense. I agree it's super repetitive to write code that way. I was just asking why we did a mix of both in this case. (Was just wondering if there was a reason to do one over the other.)

PR #7028, opened 2022-06-01
Fix query names

jaril:
The GraphQL schema types action is currently failing [here](https://github.com/RecordReplay/devtools/runs/6690800567?check_suite_focus=true) because of this error:

```
CLIError: Error in "Loading queries for Unnamed Project": Error: ️️There are multiple definitions for the `GetTest` operation. Please rename or remove all operations with the duplicated name before continuing.
```

This fixes that query name.

### DIFF
--- a/src/graphql/GetTest.ts
+++ b/src/graphql/GetTest.ts
@@ -1,0 +1,41 @@
+/* tslint:disable */
+/* eslint-disable */
+// @generated
+// This file was automatically generated and should not be edited.
+
+// ====================================================
+// GraphQL query operation: GetTest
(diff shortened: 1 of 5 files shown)


jaril:
I'm still seeing the graphql error. Investigating…

jcmorrow (approved):
Yeah I don't know why this task keeps failing that way, with no error output. Your changes do fix my local run though, so I think you are good to go 👍

jcmorrow:
Oh, actually, you should also run `npm run gql` with the correct Hasura key set and then add the changes that generates

PR #7029, opened 2022-06-01
Don't show broken next and previous buttons

jcmorrow:
If there are no analysisPoints for a particular line, then we should not
offer to run forward to the next hit, or back to the previous hit. We
could go back to the old way of running an `Analysis.findPoints` for each
location that this button is held on, but that also seems like not a
great idea, given how performance heavy that can be. I do think it would
be reasonable to make a `Debugger.stepToLocation` method, but we don't
have that yet, so here we are.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/ToggleWidgetButton.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/ToggleWidgetButton.tsx
@@ -123,11 +123,11 @@ function QuickActions({
 
   let button;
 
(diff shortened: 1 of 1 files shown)


jcmorrow:
> Approved, but also curious about where we're at fetching analysis points on hover when the meta key is pressed?

We could do this, but I think it's a bad idea until we have a `Debugger.stepToLocation` or something similar. Otherwise we risk firing off expensive tasks without the ability to cancel them.

PR #7034, opened 2022-06-01
Restore approximate source scroll offset when changing tabs from Viewer to DevTools

### LINKED ISSUE #6996: Source jumps back to the top after switching to Viewer
**Description:**
When viewing a source in the DevTools tab, then switching to Viewer, the source does not maintain its current scroll position when returning to DevTools. It instead scrolls back to the top of the source.

**Steps to Reproduce:**

**Expected Result:**

Expectation is the Source would maintain its current scroll position when tabbing between DevTools and Viewer.

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):

Replay posted in internal Discord channel (potentially sensitive).

bvaughn:
Resolves #6996

Loom https://www.loom.com/share/f1cbfa0c90874527b0add8eecd6e233e


src/devtools/client/debugger/src/reducers/sources.ts
```diff
@@ -256,11 +250,31 @@ function update(state = initialSourcesState(), action: AnyAction) {
 
     case "SOURCES_LOADED":
       return { ...state, sourcesLoading: false };
+
+    case "SET_SELECTED_SOURCE_SCROLL_POSITION":
+      return {
+        ...state,
+        sourceScrollPositions: {
+          ...state.sourceScrollPositions,
+          [action.sourceId]: action.scrollPosition,
```
markerikson, line 259: :speech_balloon: yeah, really can't say I'm a fan of keeping scroll position in Redux, but 🤷‍♂️

bvaughn: Cool beans. Let's use it instead then :)

PR #7039, opened 2022-06-01
move network filters inline

### LINKED ISSUE #6692: The Network Monitor type filters should look like the console filters
We currently show the net monitor types as a dropdown, but we should probably show it as a sidebar similar to the console.

The biggest issue here is that the dropdown flips between going up/down based on height.

Screen Shot 2022-05-10 at 1 38 50 PM

souporserious:
This PR moves the network filter to display inline since the popover menu was causing issues flipping around.

closes #6692

https://user-images.githubusercontent.com/2762082/171515651-0471c320-9ad6-4c02-b500-ed62f3b68aec.mov

## Future Considerations

I'd love to have this animate the layout soon, especially for the hero video 😛

### DIFF
--- a/src/ui/components/NetworkMonitor/FilterLayout.tsx
+++ b/src/ui/components/NetworkMonitor/FilterLayout.tsx
@@ -21,11 +21,12 @@ export const FilterLayout = ({
 
   return (
     <>
-      <div className="bg-bodyBgcolor p-1">
-        <div className="flex items-center rounded-md bg-themeTextFieldBgcolor p-1">
+      <div className="flex items-center bg-bodyBgcolor p-1 pl-0">
+        <div className={classNames("pl-1", filterOpen && "basis-32")}>
           <button
             onClick={() => setFilterOpen(!filterOpen)}
-            className={classNames("mr-1 flex items-center", {
+            className={classNames("flex items-center", {
+              "mr-1": !filterOpen,
               "text-primaryAccent hover:text-primaryAccentHover focus:text-primaryAccentHover":
                 types.size > 0,
             })}
@@ -34,7 +35,8 @@ export const FilterLayout = ({
               filter_alt
             </MaterialIcon>
           </button>
-
+        </div>
+        <div className="flex flex-1 items-center rounded-md bg-themeTextFieldBgcolor p-1">
           <MaterialIcon iconSize="lg">search</MaterialIcon>
 
           <input
@@ -47,10 +49,7 @@ export const FilterLayout = ({
 
       <div className="flex min-h-0 flex-1">
         {filterOpen ? (
-          <div
-            className="flex basis-32 flex-col overflow-auto border-r border-splitter bg-bodyBgcolor px-2 py-0.5"
-            style={{ paddingLeft: "0.55rem" }}
-          >
+          <div className="flex basis-32 flex-col overflow-auto border-r border-splitter bg-bodyBgcolor px-1.5 py-0.5">
             {RequestTypeOptions.map(canonicalType => (
               <ToggleRow
                 key={canonicalType.label}


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -87,31 +87,32 @@ export const NetworkMonitor = ({
...
-                />
-              }
-              endPanel={
-                selectedRequest ? <RequestDetails cx={cx} request={selectedRequest} /> : null
-              }
-              splitterSize={2}
-              vert={vert}
-            />
+            <Filter types={types} toggleType={toggleType} table={table}>
+              <SplitBox
```
souporserious, line 91: Opted for composition here so when `Filter` is opened/closed it doesn't re-render this component.

jasonLaster (commented):
Will let @jaril and others review. but UI looks great

src/ui/components/NetworkMonitor/index.tsx
jaril: Nit: this confused me a bit since the Filter component is more like a wrapper. Is there another way to avoid the re-render?

src/ui/components/NetworkMonitor/Filter.tsx
```diff
@@ -0,0 +1,73 @@
+import classNames from "classnames";
+import { useState } from "react";
+
+import MaterialIcon from "../shared/MaterialIcon";
+import Icon from "../shared/Icon";
+
+import { CanonicalRequestType, RequestTypeOptions } from "./utils";
+
+export function Filter({
```
jaril, line 9: If we were to go with this approach, it'd be nice if this were called something else, and have a separate `Filter` (or `FilterDrawer`?) component.

src/ui/components/NetworkMonitor/Filter.tsx
```diff
@@ -0,0 +1,73 @@
...
+          })}
+        >
+          <MaterialIcon iconSize="lg" outlined={true}>
+            filter_alt
+          </MaterialIcon>
+        </button>
+
+        <MaterialIcon iconSize="lg">search</MaterialIcon>
+
+        <input
```
jaril, line 39: It'd be nice if this input got bumped so that it clears the expanded drawer.

Currently:
image
In the console:
image

src/ui/components/NetworkMonitor/index.tsx
souporserious: Yeah, this is a little weird with `children` prop and makes it harder to reason about. I like the drawer component idea you mention below, we had this at my previous job and called it `FilterLayout` which might work here. We can use slots that handle opening/closing the sidebar:

```
<FilterLayout
  sidebarWidth={224}
  sidebar={<div>Sidebar</div>}
  toolbar={<div>Toolbar</Box>}
  table={<div>Table</div>}
/>
```

src/ui/components/NetworkMonitor/Filter.tsx
souporserious: Good catch! Yeah, we should push this to more consistent. I'll get this cleaned up tomorrow!

souporserious:
@jaril ready for another look 😎. Thanks for the suggestions! I ended up ditching the icons since they were pretty busy and moved to checkboxes to better match the other panels.

image

jasonLaster:
Looking better. Can we also show the badge count on the right similar to how we show console event listener counts?

souporserious:
Oh good callout! It does look funny with that input inset now. I'll do another pass tomorrow 😎

jaril (approved):

jaril:
@ceceliacreates 🙋‍♂️ in case there are screenshots that need updating

souporserious:
I checked [this doc](https://docs.replay.io/docs/network-24631b78240749879d54fffcda94f098), not sure if there are other areas, but it seems there are a couple of spots that could be updated on that page. The changes aren't really noticeable since we don't explicitly show filters in docs so doesn't seem urgent.

PR #7041, opened 2022-06-02
Do not register every analysis as a logpoint

jcmorrow:
In https://github.com/replayio/devtools/commit/3c825b6a95d3b2d5ef6201135ca23c4a7a1c0445 we removed the concept of `showInConsole` with the thinking that "It's a logpoint, all logpoints get shown in the console". But in order for that to be true, we need to not use the logpoint machinery for things that... are not logpoints. Now, this has resulted in some duplication, but I believe it is duplication that is justified at this point in time. We are still figuring out what the best API for these primitive operations is, and if it turns out that we definitely have too much of a low-level API and we are always repeating the same steps, we can extract those, but if we insist on never duplicating code, it's hard to tell what responsibilities are shared and which just happen to be next to each other but get turned on and off by passing in flags to functions (which, in general, I find to be a little code-smelly).

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
@@ -1,23 +1,32 @@
-import {
-  runAnalysisOnLine,
-  updateHoveredLineNumber,
-} from "devtools/client/debugger/src/actions/breakpoints/index";
+import { updateHoveredLineNumber } from "devtools/client/debugger/src/actions/breakpoints/index";
 import { setBreakpointHitCounts } from "devtools/client/debugger/src/actions/sources";
 import { minBy } from "lodash";
+import { AnalysisParams } from "protocol/analysisManager";
+import { Analysis, AnalysisError, createAnalysis } from "protocol/thread/analysis";
 import React, { useRef, useState, useEffect, ReactNode } from "react";
 import { useDispatch, useSelector } from "react-redux";
+import { UIThunkAction } from "ui/actions";
+import { saveAnalysisError } from "ui/actions/logpoint";
 import { KeyModifiers } from "ui/components/KeyModifiers";
 import MaterialIcon from "ui/components/shared/MaterialIcon";
 import hooks from "ui/hooks";
 import { Nag } from "ui/hooks/users";
 import { selectors } from "ui/reducers";
-import { setHoveredLineNumberLocation } from "ui/reducers/app";
-import { AnalysisError, AnalysisPayload } from "ui/state/app";
+import { setAnalysisPoints, setHoveredLineNumberLocation } from "ui/reducers/app";
+import { AnalysisPayload } from "ui/state/app";
 import { prefs, features } from "ui/utils/prefs";
 import { trackEvent } from "ui/utils/telemetry";
 import { shouldShowNag } from "ui/utils/user";
 
 import { getHitCountsForSelectedSource, getSelectedSource } from "../../reducers/sources";
+import {
+  analysisCreated,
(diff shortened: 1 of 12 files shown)


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -27,7 +36,10 @@ function getTextAndWarning(analysisPoints?: AnalysisPayload, analysisPointsCount
   if (analysisPoints?.error) {
     return {
       showWarning: false,
-      text: analysisPoints.error === AnalysisError.TooManyPoints ? "10k+ hits" : "Error",
+      text:
+        (analysisPoints.error as AnalysisError) === AnalysisError.TooManyPointsToFind
```
jcmorrow, line 40: For some reason, typescript thinks this things type is `AnalysisError.TooManyPointsToRun | AnalysisError.Unknown`, and so it thinks there is no overlap here. I'm sure it's doing some type inference somewhere, but it's unclear where?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -29,15 +30,19 @@ export function PanelStatus({
   if (!isIndexed || !analysisPoints) {
     status = "Loading";
   } else if (analysisPoints.error) {
-    status = analysisPoints.error === AnalysisError.TooManyPoints ? "10k+ hits" : "Error";
-  } else if (analysisPoints.data.length == 0) {
+    status =
+      (analysisPoints.error as AnalysisError) === AnalysisError.TooManyPointsToFind ||
```
jcmorrow, line 34: Similarly, typescript thinks this things type is `AnalysisError.TooManyPointsToFind | AnalysisError.Unknown`, and so it thinks there is no overlap here. I'm sure it's doing some type inference somewhere, but it's unclear where?

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
bvaughn: Seems a little bit risky that we have hard-coded strings like "10k+ hits" on the frontend. Maybe this number is unlikely to ever change on the backend?

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -71,6 +83,92 @@ function Wrapper({
...
+function runAnalysisOnLine(line: number): UIThunkAction {
+  return async (dispatch, getState, { ThreadFront }) => {
+    // A lot of this logic is reused from logpoint.ts
+    // It is possible to instead repurpose that function here, but increasingly
+    // I want logpoint.ts to be... for logpoints. Part of what has made our
+    // analysis code such a mess is trying to repurpose an analysis runner to
+    // work for many different devtools specific workflows. It would be better
+    // if many components used the primitives from `protocol` instead. If it
+    // turns out that those primitives are getting used the same way all the
+    // time, maybe it's time to add some new things to the `protocol` folder.
```
bvaughn, line 95: I don't disagree with the overall sentiment (that DRY isn't always a good choice) but I hope our long-term goal is for our _components_ not to have this kind of logic in them at all.

jcmorrow:
> Seems a little bit risky that we have hard-coded strings like "10k+ hits" on the frontend. Maybe this number is unlikely to ever change on the backend?

1000%. This has bitten us many, many times when something other than "TooManyPointsToFind" goes wrong and we just throw up a "10k+" error that is clearly wrong.

PR #7045, opened 2022-06-02
Allow points to be undefined

jcmorrow:
I made a change last night that allows `analysisPoints.data` to be
undefined rather than being an empty array. I still think this makes
sense, because there is a difference between "we got those points and
there were none" vs. "we hit an error and never retrieved those points".
But there were a couple of places where we needed to check for presence
now that I missed.

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
@@ -30,22 +30,24 @@ function BreakpointNavigation({
       seek(point.point, point.time, true);
     }
   };
-  const isEmpty = analysisPoints && (analysisPoints.error || analysisPoints.data.length == 0);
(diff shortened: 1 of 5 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.js
```diff
@@ -30,22 +30,24 @@ function BreakpointNavigation({
       seek(point.point, point.time, true);
     }
   };
-  const isEmpty = analysisPoints && (analysisPoints.error || analysisPoints.data.length == 0);
+  const points = analysisPoints?.data || [];
```
bvaughn, line 33: It's unfortunately how many untyped files we have in our codebase (combined with kind of minimal e2e test coverage).

bvaughn (approved):
Only somewhat relevant here, but I think we should use `null` values for cases like this (e.g. too many points) because that's more _meaningful_ than `undefined` (which could mean anything like even a mistyped attribute)

PR #7051, opened 2022-06-02
[DRAFT] Use new analysis entries and merged points results in selectors and UI

markerikson:
This PR:

- Rewrites `getAnalysisPointsForLocation` selector and usage sites
  - Replaced the existing `getAnalysisPointsForLocation` selector with a new implementation that reads from the analysis entries in `state.breakpoints`. It's also memoized, with a cache size of 50 results by default.
  - All usage sites of the existing selector have been updated to use
the new one instead.

The state in `state.app.analyses` has been commented out to confirm it's not being used, but hasn't been cleaned up yet.

I've thrown together a quick CodeSandbox at https://codesandbox.io/s/sad-burnell-uon2bh?file=/src/App.tsx that has a 205-iteration loop and a 10005-iteration loop for use in testing, and a replay at https://app.replay.io/recording/analysis-hits-size-checks--43f10c43-fccc-4519-b6fe-0cb95f03568c that uses that sandbox.

I'm seeing that setting an evaluation on the 205-iteration loop leaves us with a bunch of "Loading..." messages in the console, so those aren't getting cleaned up properly if we bail out of running the full analysis.


src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -371,3 +381,120 @@ export function getLogpointsForSource(state: UIState, sourceId: string) {
   const breakpoints = getBreakpointsList(state);
   return breakpoints.filter(bp => bp.location.sourceId === sourceId).filter(bp => isLogpoint(bp));
 }
+
+export type LocationAnalysisPoints = {
```
jcmorrow, line 385: I'm tempted to call this `LocationAnalysisResult`, since it also includes a potential error, but YMMV.

jcmorrow: I take it back. I think the nicest API to have here would be:

```
LocationAnalysisStatus {
	data: PointDescription[];
	error: AnalysisError | undefined;
	status: AnalysisStatus; // probably like "Running", "Errored", or "Completed"?
}
```

But maybe that's a follow-on.

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -371,3 +381,120 @@ export function getLogpointsForSource(state: UIState, sourceId: string) {
...
+
+    const allPoints = matchingEntries.map(entry => entry.points!).flat();
+    const uniquePoints = uniqBy(allPoints, item => item.point);
+    uniquePoints.sort((a, b) => compareNumericStrings(a.point, b.point));
+
+    const finalPoints = focusRegion ? filterToFocusRegion(uniquePoints, focusRegion) : uniquePoints;
+
+    // Meanwhile, we're going to use the _latest_ analysis run's status
+    // for display purposes.
+    const latestAnalysisEntry = analyses.entities[mappingEntry.currentAnalysis!];
```
jcmorrow, line 473: I think it would also be *a-okay* to have `getAnalysisStatusForLocation` and `getAnalysisPointsForLocation`, but again, could be a follow-on.

jcmorrow (commented):
There's obviously more to do, but this is a great start. I'll play around with the preview branch and then approve if it feels solid.

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -371,3 +381,120 @@ export function getLogpointsForSource(state: UIState, sourceId: string) {
...
+      const analysisEntry = analyses.entities[analysisId];
+      // TODO Double-check undefined vs empty string conditions here
+      if (!analysisEntry || (analysisEntry.condition ?? "") !== condition) {
+        return;
+      }
+
+      // TODO Are we sure about the error statuses here?
+      const hasPoints = [
+        // Successful queries
+        AnalysisStatus.PointsRetrieved,
```
jcmorrow, line 446: I would also probably call this `PointsFound`

jcmorrow (commented):
One thing that surprises me from the preview branch: when the condition is `false` I still end up getting points on the hit timeline?

src/devtools/client/debugger/src/reducers/breakpoints.ts
jcmorrow: https://app.replay.io/recording/false-condition-does-not-hide-logpoints--f74ebd50-d961-4c96-b996-5711a4112ed1?point=80805119884749729135803403855724710&time=31659.124300593958&hasFrames=true

jcmorrow: I wonder if we should keep the concept of an `analysisRequested` action, which clears the `currentAnalysis` for that location?

src/ui/actions/logpoint.ts
```diff
@@ -368,11 +341,10 @@ async function setMultiSourceLogpoint(
       );
 
       showLogpointsResult(logGroupId, results);
+    } else {
+      removeLogpoint(logGroupId);
     }
 
-    // TODO Remove this and redo Redux logic
-    saveLogpointHits(points, analysisResults, locations, condition);
```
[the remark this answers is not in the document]
markerikson: wait, no, this was only dispatching the now-dead analysis actions. hmm...

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -20,28 +20,32 @@ export function PanelStatus({
...
 }) {
   const isIndexed = useSelector(getIsIndexed);
   const time = useSelector(getCurrentTime);
   let status = "";
 
-  const points = analysisPoints?.data;
+  const points = analysisPoints?.data || [];
   const error = analysisPoints?.error;
 
   if (!isIndexed || !analysisPoints) {
```
jcmorrow, line 33: And then this line should change to be just `if (!points)`

src/ui/actions/logpoint.ts
```diff
@@ -262,7 +243,13 @@ async function setMultiSourceLogpoint(
       if (!points.error) {
         showPrimitiveLogpoints(logGroupId, points.data || [], primitiveFronts);
       }
-      return;
+      // If we're only displaying only primitives, we can bail out if there's no condition or the condition request is done
+      if (
+        !condition ||
+        (points.status === AnalysisStatus.Completed && points.condition === condition)
+      ) {
+        return;
```
jcmorrow, line 251: In the case that we have not seen this combo of locatoin & condition before, I would expect that `points` would not exist (because `getAnalysisPointsForLocation(store.getState(), locations[0], condition)` would return `undefined` and thus we would never get into this block in the first place because of the check on line 246. What am I missing?

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
```diff
@@ -20,28 +20,32 @@ export function PanelStatus({
...
-  analysisPoints: AnalysisPayload;
+  analysisPoints?: LocationAnalysisSummary;
   prefixBadge: PrefixBadge;
 }) {
   const isIndexed = useSelector(getIsIndexed);
   const time = useSelector(getCurrentTime);
   let status = "";
 
-  const points = analysisPoints?.data;
+  const points = analysisPoints?.data || [];
```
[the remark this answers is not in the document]
markerikson: This is for the panel UI, not the loading messages in the console logs

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/PanelStatus.tsx
markerikson: Lemme try this out

src/ui/actions/logpoint.ts
markerikson: This was specifically for later hits to this block.  In particular:

- Add a print statement. Initial message is just primitives, like `"someFunc", 123`.  Don't have to request an analysis.
- Add a condition.  We need to request an analysis.
- Change the condition. The latest analysis _did_ complete and has points.... _but_ the condition changed, so we need to run a _new_ analysis.  What I was seeing was that prior to adding the "if condition changed" check, we were bailing out early because the last analysis we ran did succeed, it was just for a condition that was now different.

jcmorrow: So `getAnalysisPointsForLocation(store.getState(), locations[0], condition)` ignores the condition and just returns the most recently run analysis points?

Issue #6913, opened 2022-05-23
Global Loader 1.6


jonbell-lot23:
Global Loader 1.5 is a big step forward, and here are some small revisions we've discussed for 1.6

**Ready to be discussed + worked on:**

- [x] #6849
- [x] #7064 ([design-discussion](https://github.com/RecordReplay/design/issues/65))
- [x] #7053 ([design-discussion](https://github.com/RecordReplay/design/issues/65))
- [x] #7060 ([design-discussion](https://github.com/RecordReplay/design/issues/64))

<details>
  <summary>Old stuff no longer in scope</summary>
  
Other:
- [ ] ~~Buffer information when you hover on the play button ([design-discussion](https://github.com/RecordReplay/design/issues/66))~~
    * ~~Also hovering over the timeline -- buffering versus processing (Jon recommends holding off on this)~~
- [ ] ~~Make sure we're only using the term "video only" rather than "unloaded" in the UI (might be redundant)~~
- [ ] ~~Add context menu to the timeline (not recommended yet)~~
</details>

bvaughn:
> Cursor shouldn't be a text input cursor on timestamp (keep standard pointer)

Just to clarify here, the text cursor _is_ actually the standard cursor for browsers like Safari (which I believe you use) when hovering over text like our timestamp. We can override it by using CSS to disable text selection (`user-select: none`) or we could override the cursor itself (`cursor: default`) if you'd like to though. 😄

jonbell-lot23:
> > * [ ]  Cursor shouldn't be a text input cursor on timestamp (keep standard pointer)
> 
> Just to clarify here, the text cursor _is_ actually the standard cursor for browsers like Safari (which I believe you use) when hovering over text like our timestamp. We can override it by using CSS to disable text selection (`user-select: none`) or we could override the cursor itself (`cursor: default`) if you'd like to though. 😄

Makes sense. Yup, that's what would be best, thanks! (And then of course when the precision editor is up, we'd use the text input cursor again)

jonbell-lot23:
Update: each of these design items have a POV attached. See links above for Loom walkthroughs and screenshots.

bvaughn:
@jonbell-lot23 What does "Surface unloaded data" mean?

Sorry if we've talked about this and I've just forgotten, but I'm not sure what that means in terms of actionable work.

jonbell-lot23:
You are right to be confused! That was a bite I forgot to flesh out.

[...]

PR #7060, opened 2022-06-03
Add loading stalled pop-out with link to support channel

bvaughn:
Relates to #6913 and https://github.com/replayio/design/issues/64

Loom: https://www.loom.com/share/e92ccf7b7629404e9d2ac10f0436a9ff

Modal popout screenshot

cc @jonbell-lot23 for wording

### DIFF
--- a/src/ui/components/Timeline/Capsule.module.css
+++ b/src/ui/components/Timeline/Capsule.module.css
@@ -4,6 +4,7 @@
   align-items: stretch;
   column-gap: 1px;
   height: 2rem;
+  position: relative;
 }
 
 .LeftSideLoading,
@@ -48,4 +49,22 @@
 .CircularProgressbar {
   width: 1.5rem;
   height: 1.5rem;
+}
+
+.SlowLoadingPopout {
+  position: absolute;
+  bottom: calc(100% + 0.25rem);
+  left: 0;
+  background: var(--modal-bgcolor);
+  border: 1px solid var(--modal-border);
+  color: var(--theme-text-color-strong);
+  padding: 0.5rem;
+  border-radius: 0.5rem;
+  font-size: var(--theme-body-font-size);
+  text-align: center;
+}
+
+.SlowLoadingLink {
+  text-decoration: underline;
+  color: var(--source-link-color);
 }
--- a/src/ui/components/Timeline/Capsule.tsx
+++ b/src/ui/components/Timeline/Capsule.tsx
@@ -13,15 +13,15 @@ import styles from "./Capsule.module.css";
 import { EditFocusButton } from "./EditFocusButton";
 import FocusInputs from "./FocusInputs";
 
-const SHOW_SLOW_LOADING_POPOUT_AFTER_DELAY = 1000;
+const SHOW_SLOW_LOADING_POP_OUT_AFTER_DELAY = 1000;
 
 export default function Capsule({
   setShowLoadingProgress,
 }: {
   setShowLoadingProgress: React.Dispatch<React.SetStateAction<boolean>>;
 }) {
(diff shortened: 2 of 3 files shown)


src/ui/components/Timeline/Capsule.module.css
```diff
@@ -49,3 +50,21 @@
   width: 1.5rem;
   height: 1.5rem;
 }
+
+.SlowLoadingPopout {
```
jcmorrow, line 54: Is the idea behind CamelCase for the class names that they will be the exact same token as the component they are styling?

src/ui/hooks/useModalDismissSignal.ts
```diff
@@ -18,7 +19,7 @@ export default function useModalDismissSignal(
     };
 
     const handleDocumentClick = (event: MouseEvent) => {
-      if (modalRef.current !== null && !modalRef.current.contains(event.target as Node)) {
+      if (!element.contains(event.target as Node)) {
```
jcmorrow, line 22: 👍

src/ui/components/Timeline/Capsule.tsx
```diff
@@ -55,13 +59,82 @@ function LoadingState({
...
+      }, SHOW_SLOW_LOADING_POPOUT_AFTER_DELAY);
+    };
+
+    const clearPendingTimeout = () => {
+      if (timeoutID !== null) {
+        clearTimeout(timeoutID);
+      }
+    };
+
+    div.addEventListener("click", clearPendingTimeout);
```
jcmorrow, line 97: I'm having a hard time picturing this interaction. Which element is this click handler on and which part of the interaction does it handle?

jcmorrow (approved):
Couple of questions out of curiosity, but it looks good ⚡

src/ui/components/Timeline/Capsule.tsx
bvaughn: Basically I'm setting a hover timeout (as per Jon's request) that we shouldn't show the popout _immediately_ but we should "wait a sec". So I'm just cleaning up the timeout if the user explicitly clicks to show it (or if the thing unmounts)

jcmorrow: Ah, gotchya, in that case, what do you think about putting this in the `onClick` handler on line 112 where we show the popout explicitly??

bvaughn: I think that makes sense, except that we'd then need to store the timeout ID in a ref so it could be referenced by the React click handler (since we're inside of an effect now).

TBH it's just a little optimization. Setting the value to true when it's already true will bailout anyway so if it's this line of code is a snag, I can just remove it entirely :)

jcmorrow: Nah, that makes sense. I definitely would not want to add a ref just to move a line of code around for readability, since that would probably add more cognitive load than it would relieve. All good! 🚢

bvaughn: I appreciate that 👍🏼 

I'm going to walk back my claim about it being an unneccessary optimization. It's true that React would be able to bailout without rendering if we were setting an already-true value to true, _but_ we'd have a potential edge case bug if a user hovered, clicked (to show) and then clicked again (to hide) before the timer fired– we'd end up re-showing the popout.

So I'm going to just leave it in place 😁

jonbell-lot23:
Thanks for the mention @bvaughn ! This is one of those interactions that I needed to feel to understand if it was a good idea or not. (Jason had a similar comment about tooltips versus popovers etc)

I'm pleased to report that the Loom looked totally natural to me! Thanks for this. I'm sure we'll update the copy over time, but this is a great start.

bvaughn:
You are welcome!

PR #7064, opened 2022-06-03
Timeline playback/selection can go outside of focus region

bvaughn:
# This is a draft

- [x] Allow users to click (or scrub) outside of the focus region on the timeline
- [x] Allow playback to start (or continue) outside of the focus region
- [ ] Don't try to create a pause point if a breakpoint is set and focus region is changed so that the breakpoint is no longer within the focus window (and/or loaded region)
  - [ ] Customize callstack message in this case as well


src/devtools/client/webconsole/reducers/messages.ts
```diff
@@ -438,13 +438,6 @@ function setVisibleMessages(messagesState: MessageState, forceTimestampSort = fa
       filtered[message.level as FilterCountKeys] = filtered[message.level as FilterCountKeys] + 1;
     }
   }
-  console.groupCollapsed(
-    `%csetVisibleMessages() (hide ${filtered})`,
-    "color: red; font-weight: bold;"
-  );
-  console.log("all messages:", messagesState.messages.entities);
-  console.log("visible:", messagesToShow);
-  console.groupEnd();
```
bvaughn, line 447: These shameful lines were added by me 😬

src/ui/actions/comments.ts
```diff
@@ -203,23 +204,18 @@ export function createLabels(
...
+
     if (item.sourceLocation) {
-      cx = selectors.getThreadContext(getState());
-      dispatch(selectLocation(cx, item.sourceLocation));
+      const regions = getLoadedRegions(getState());
+      const isTimeInLoadedRegion = regions !== null && isTimeInRegions(item.time, regions.loaded);
+      if (isTimeInLoadedRegion) {
+        context = selectors.getThreadContext(getState());
+        dispatch(selectLocation(context, item.sourceLocation));
+      }
```
jasonLaster, line 218: hmm, maybe it's safe to jump to the source location

jasonLaster (commented):
some small tweaks

jasonLaster (commented):
small tweaks

bvaughn:
Jason and I reviewed this one in realtime over VC, but I want to hold off landing anything until Monday morning. (Doesn't seem worth risking any weekend downtime if I overlooked something.)

jaril:
Not sure if this is intended behavior, but noticed that you _could_ still pause at a point that's outside the focus region: https://www.loom.com/share/e0063a4255a2460995fc6397184441dd

https://devtools-git-design-issue-65-recordreplay.vercel.app/recording/hello-world-tina--5944d4f8-f7c7-4a5e-9bb6-ad93e1fd756f?point=324518553658608888380883902398526&time=154&hasFrames=true#

bvaughn:
Here's a demo of the console message stack frame fix (e121d0c) which also fixes the broken e2e test:
https://www.loom.com/share/ef27658832ab430a92e5bce4a4baf19d

bvaughn:
@jaril I don't think this is correct, given that Jon's design issue explicitly included mocks for what the call stack and scopes panels should look like when you're paused outside of a loaded region (which in most cases means outside of the focus window).

jaril:
@ceceliacreates this might require a doc change 👋

PR #7066, opened 2022-06-03
Use separate DB's and don't query so much

jcmorrow:
Maybe fixes https://github.com/replayio/customer-support/issues/108, see that issue for more discussion, but the gist is:

Chrome's indexedDB can be slow, and we `await` a bunch of reads from it when we start up the store. Our theory is that these compete with each other, and slow down all tabs from loading. I suspect it would be even *better* if we just wrote to indexedDB way, way less, because as @gideonred has pointed out - none of the features that we use it to provide are necessities.

### DIFF
--- a/src/devtools/shared/async-storage.ts
+++ b/src/devtools/shared/async-storage.ts
@@ -44,16 +44,16 @@
  *
  */
 
-const DBNAME = "devtools-async-storage";
 const DBVERSION = 1;
 const STORENAME = "keyvaluepairs";
 
 const withStore = (
+  dbName: string,
   type: IDBTransactionMode,
   onsuccess: (store: IDBObjectStore, db: IDBDatabase) => void,
   onerror: () => void
 ) => {
-  const dbConn = indexedDB.open(DBNAME, DBVERSION);
+  const dbConn = indexedDB.open(dbName, DBVERSION);
 
   dbConn.onerror = () => {
     onerror();
@@ -72,9 +72,10 @@ const withStore = (
   };
 };
 
-export const getItem = (itemKey: string): Promise<any> => {
+export const getItem = (dbName: string, itemKey: string): Promise<any> => {
   return new Promise((resolve, reject) => {
     withStore(
+      dbName,
       "readonly",
       (store, db) => {
         store.transaction.oncomplete = () => {
@@ -96,9 +97,10 @@ export const getItem = (itemKey: string): Promise<any> => {
   });
 };
 
-export const setItem = (itemKey: string, value: any): Promise<void> => {
+export const setItem = (dbName: string, itemKey: string, value: any): Promise<void> => {
   return new Promise((resolve, reject) => {
     withStore(
+      dbName,
       "readwrite",
       (store, db) => {
         store.transaction.oncomplete = () => {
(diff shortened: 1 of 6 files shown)


src/devtools/shared/async-storage.ts
```diff
@@ -44,16 +44,16 @@
  *
  */
 
-const DBNAME = "devtools-async-storage";
 const DBVERSION = 1;
 const STORENAME = "keyvaluepairs";
 
 const withStore = (
+  dbName: string,
```
jcmorrow, line 51: Now each replay's `session` object will get its own little DB.

jcmorrow: `prefs` is still shared.

src/devtools/shared/async-store-helper.js
```diff
@@ -29,11 +29,11 @@ export function asyncStoreHelper(root, mappings) {
   Object.keys(mappings).map(key =>
     Object.defineProperty(store, key, {
       async get() {
-        const value = await asyncStorage.getItem(`${root}.${getMappingKey(key)}`);
```
jcmorrow, line 32: I thought this was clever 😁

src/ui/setup/index.ts
```diff
@@ -51,100 +50,79 @@ const getDefaultSelectedPrimaryPanel = (session: any, recording?: Recording) =>
...
   return syncInitialMessageState({
     ...overrides,
-    filters: { ...(await getInitialFiltersState()), ...filters },
+    filters: { ...getInitialFiltersState(session), ...filters },
   });
 };
 
 export async function bootstrapApp() {
+  const recordingId = getRecordingId();
+  const session = recordingId ? (await getReplaySession(recordingId)) : null;
```
jcmorrow, line 118: we now only get this once, rather than `await`ing it like 5 times.

gideonred (approved):
Hopefully we can all move on, one day, from noSQL to noDB.

src/devtools/shared/async-storage.ts
```diff
@@ -44,16 +44,16 @@
  *
  */
 
-const DBNAME = "devtools-async-storage";
 const DBVERSION = 1;
 const STORENAME = "keyvaluepairs";
```
jcmorrow, line 48: Also... this seems like an abuse of IndexedDB? We only have one "item" and it's a giant blob? That seems bad.

jcmorrow: AFAICT, this is like having a postgres table where you just have one column called `data` and it's a JSON column 😵‍💫

jcmorrow:
I'm.... pretty sure I made it *worse* somehow? 😬

src/ui/setup/prefs.ts
```diff
@@ -155,7 +155,7 @@ export async function isLocalNagDismissed(nag: LocalNag) {
     return;
   }
 
-  const replaySessions = await getReplaySessions();
+  const replaySessions = await getReplaySession(recordingId);
   const replaySession = replaySessions[recordingId];
```
bvaughn, line 159: This looks a little suspicious now:
```js
const replaySessions = await getReplaySession(recordingId);
const replaySession = replaySessions[recordingId];
```

Given the names and params, I would expect
```js
const replaySession = await getReplaySession(recordingId);
```

jcmorrow: Yep, you are right

src/devtools/shared/async-storage.ts
markerikson: @jcmorrow this goes back to what i was saying about `state.pendingBreakpoints` - it ends up being a deserialization of _alllll_ breakpoints for _alllll_ recordings :)

jcmorrow:
Closing in favor of https://github.com/replayio/devtools/pull/7069

PR #7068, opened 2022-06-03
Show logpoints even if there are more than 200

jcmorrow:
But less than 1000 because that is a lot of console messages to scroll
through : )


markerikson (approved):

src/ui/actions/logpoint.ts
```diff
@@ -104,7 +104,7 @@ async function showPrimitiveLogpoints(
   pointDescriptions: PointDescription[],
   values: ValueFront[]
 ) {
-  if (!LogpointHandlers.onResult || pointDescriptions.length >= 200) {
+  if (!LogpointHandlers.onResult || pointDescriptions.length >= 1000) {
```
jasonLaster, line 107: I think we should have a constant somewhere for this

PR #7089, opened 2022-06-07
Fix stepping-05 and object_preview-05 e2e tests

hbenl:
We enabled a new pretty-printer in the backend,
so some line numbers in the tests needed to be adjusted.
(and I removed an unused parameter in a `Test.warpToMessage()` call)


jcmorrow (approved):

jaril:
@hbenl the two tests seems to still be failing. or is it because the changed test files aren't the ones being run in this action?

https://github.com/replayio/devtools/actions/runs/2454770473

Issue #6919, opened 2022-05-23
Failing test: breakpoints-07


jaril:
Replay: https://app.replay.io/recording/replay-of-devtools-m3yk8crmh-recordreplayvercelapp--258d37cf-db69-43b9-91fa-202c99f2be1b

There seems to be a spike of breakpoint-07 test failures ([see here](https://ui.honeycomb.io/replay/datasets/backend/result/Hp3Ese95Vtv/a/3fveKN7U4W1?useStackedGraphs)). @hbenl I think you might have been looking at this already — can you let me know in case you've made more progress on it than I have?

hbenl:
I did look at it briefly and noticed that the backend responds with an internal error to the `Debugger.setBreakpoint` commands, so this is probably a backend issue.

jaril:
https://github.com/RecordReplay/backend/pull/5727 has been merged but it looks like the most recent breakpoints-07 test runs are still failing.

jaril:
Linking most recent issue: https://github.com/RecordReplay/backend/issues/5736.

jaril:
It looks like this has stopped becoming a problem as of last night, not seeing instances of the same error popping up in honeycomb anymore.

https://ui.honeycomb.io/replay/datasets/backend/result/ENB5fBCveDb/a/3fveKN7U4W1?useStackedGraphs

Will check again later today and if that's still the case, will be closing it.

jaril:
Not seeing this _as often_, but keeping it open to keep an eye on it.

jaril:
This is happening again: https://ui.honeycomb.io/replay/datasets/backend/result/4nSpxnJbYHr/a/3fveKN7U4W1?useStackedGraphs

hbenl:
Here's the recording of a recent failure: https://app.replay.io/recording/be410907-4775-4864-9d9d-343e21d7a2f4
There's a logpoint with 2 hits but only one corresponding console message:
![Screenshot from 2022-06-07 16-17-22](https://user-images.githubusercontent.com/16060807/172403775-850d8dc9-bc82-4141-8655-996cd3f50c1e.png)
![Screenshot from 2022-06-07 16-19-42](https://user-images.githubusercontent.com/16060807/172403981-15d20738-0cc3-4463-94e9-5ce8dbd5e83e.png)

PR #7093, opened 2022-06-07
Update the locations of received analysis points (#6919)

hbenl:
The new analysis code didn't update the locations of received points, which led to failed assertions, which led to the breakpoints-07 e2e test failing.
(We may have multiple identical sources in the recording but want to show them as one source to the user, so we pick one of the identical sources and update all incoming locations to reference that source)


hbenl:
There's a subsequent failure in the test which I'll look into tomorrow.

jcmorrow (approved):
I think it would be nice if we could move away from mutating backend results as they come in, and also having little things like this sprinkled throughout the code, but for right now, I think this is ok 👍

markerikson (approved):

Issue #7094, opened 2022-06-07
Network Monitor Stack Trace panel never loads


souporserious:
**Description:**
The Stack Trace panel never loads.

**Steps to Reproduce:**
Navigate to a POST request in the Network Monitor and view the Stack Trace panel which never seems to load.

**Expected Result:**
It shows the correct Stack Trace at the time of the POST.

https://app.replay.io/recording/network-monitor-stack-trace-never-loads--53f4da67-69bc-48cc-b0b3-a02d22dc5b5c

jcmorrow:
@souporserious could you make the original Replay public? That might help with diagnosing

souporserious:
Ah, yes! Here it is https://app.replay.io/recording/overboard--e5575afa-4719-4df8-ae45-d0639ae46b15

PR #7096, opened 2022-06-08
Fix the network monitor stack trace

jcmorrow:
Two things were off here:

1. We were passing in `framesLoading={true}` which resulted in that component always displaying `Loading...`. This made sense before https://github.com/replayio/devtools/commit/9175df71c382bdb228b601cbd7b1a875e3d11803. But now we don't need to pass that anymore. 
2. When selecting a frame from this frame stack, there is no guarantee that we are already paused there (because you can open a request's details without warping to it), so we need to both seek to this request, and then select the frame.

Fixes https://github.com/replayio/devtools/issues/7094

### DIFF
--- a/src/ui/actions/network.ts
+++ b/src/ui/actions/network.ts
@@ -4,10 +4,13 @@ import {
   responseBodyData,
   RequestId,
   requestBodyData,
+  Frame,
 } from "@replayio/protocol";
 import { createFrame } from "devtools/client/debugger/src/client/create";
+import { Context } from "devtools/client/debugger/src/reducers/pause";
+import { RequestSummary } from "ui/components/NetworkMonitor/utils";
 import { getLoadedRegions } from "ui/reducers/app";
-import { getSummaryById } from "ui/reducers/network";
+import { getRequestById, getSummaryById } from "ui/reducers/network";
 import { isPointInRegions } from "ui/utils/timeline";
 
 import { UIThunkAction } from ".";
@@ -90,7 +93,7 @@ export function hideRequestDetails() {
 export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
   return async (dispatch, getState, { ThreadFront }) => {
     const state = getState();
-    const request = state.network.requests.find(request => request.id === requestId);
+    const request = getRequestById(state, requestId);
     const loadedRegions = getLoadedRegions(state);
 
     // Don't select a request that's not within a loaded region.
@@ -124,5 +127,24 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
       requestId,
       type: "SHOW_REQUEST_DETAILS",
     });
+  };
+}
+
+export function seekToRequestFrame(
+  request: RequestSummary,
+  frame: Frame,
+  cx: Context
+): UIThunkAction {
+  return async (dispatch, getState, { ThreadFront }) => {
+    const state = getState();
+    const loadedRegions = getLoadedRegions(state);
+    const point = request.point;
+
+    // Don't select a request that's not within a loaded region.
+    if (!request || !loadedRegions || !isPointInRegions(loadedRegions.loaded, point.point)) {
+      return;
+    }
+
+    ThreadFront.timeWarp(point.point, point.time, true, frame);
   };
 }
--- a/src/ui/components/NetworkMonitor/RequestDetails.tsx
+++ b/src/ui/components/NetworkMonitor/RequestDetails.tsx
@@ -5,8 +5,7 @@ import sortBy from "lodash/sortBy";
 import { WiredFrame } from "protocol/thread/pause";
 import React, { FC, ReactNode, useEffect, useMemo, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { actions } from "ui/actions";
-import { hideRequestDetails } from "ui/actions/network";
+import { hideRequestDetails, seekToRequestFrame } from "ui/actions/network";
 import { useFeature } from "ui/hooks/settings";
 import { getLoadedRegions } from "ui/reducers/app";
 import { getFormattedFrames } from "ui/reducers/network";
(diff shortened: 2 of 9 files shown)


src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
 }
+
+export function seekToRequestFrame(
+  requestId: RequestId,
+  frameId: string,
+  cx: Context
+): UIThunkAction {
+  return async (dispatch, getState, { ThreadFront }) => {
+    const state = getState();
+    const request = state.network.requests.find(request => request.id === requestId);
```
hbenl, line 139: a selector for this would be nice, especially since there is already another place where we fetch a request from redux.

src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+
+    const requestSummary = getSummaryById(state, requestId);
+    if (!requestSummary) {
+      console.error(`Could not find summary for request "${requestId}"`);
+      return;
+    }
+
+    const timeStampedPoint = requestSummary.point;
+    const pause = ThreadFront.ensurePause(timeStampedPoint.point, timeStampedPoint.time);
+    ThreadFront.timeWarpToPause(pause);
```
hbenl, line 155: I'd expect this to fail because `pause.hasFrames` isn't set yet: https://github.com/replayio/devtools/blob/39e705382386b7460bc115e679c8ca4f9667a005/packages/protocol/thread/thread.ts#L350-L357
I'd use `ThreadFront.timeWarp(point, time, true)` followed by `ThreadFront.getFrames()` instead of `ThreadFront.ensurePause(point, time)` and then `ThreadFront.timeWarpToPause(pause)` followed by `pause.getFrames()`

src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+    const frames = (await pause.getFrames())?.filter(Boolean) || [];
```
hbenl, line 156: That `.filter(Boolean)` shouldn't be necessary, any reason why you added this? If `getFrames()` returns falsy frames, that would be a bug that should be fixed.

src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+    const formattedFrames = await Promise.all(frames?.map((frame, i) => createFrame(frame, i)));
+    const frame = formattedFrames.find(f => f?.id === frameId);
```
hbenl, line 158: Nit: we could format only the frame that we're interested in.

src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+    console.log({ pause, frames, frameId, request });
```
hbenl, line 159: remove

src/ui/actions/network.ts
```diff
@@ -126,3 +128,38 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+    if (!frame) {
+      console.error(`Request frame not found: ${frameId}`);
```
hbenl, line 161: If we found no frame then we shouldn't go on to try to select it.

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -144,15 +144,25 @@ const Cookies = ({ request }: { request: RequestSummary }) => {
...
+  cx: any;
+  frames: WiredFrame[];
+  request: RequestSummary;
+}) => {
   const dispatch = useDispatch();
-  const selectFrame = (cx: any, frame: WiredFrame) => dispatch(actions.selectFrame(cx, frame));
-
+  const selectFrame = async (cx: any, frame: any) => {
+    console.log({ frame });
+    dispatch(seekToRequestFrame(request.id, frame.id, cx));
```
hbenl, line 159: Why not pass in the `RequestSummary` that we have instead of extracting its `id` and then fetching it from redux in the thunk?

src/ui/components/NetworkMonitor/RequestDetails.tsx
```diff
@@ -144,15 +144,25 @@ const Cookies = ({ request }: { request: RequestSummary }) => {
...
+}: {
+  cx: any;
+  frames: WiredFrame[];
+  request: RequestSummary;
+}) => {
   const dispatch = useDispatch();
-  const selectFrame = (cx: any, frame: WiredFrame) => dispatch(actions.selectFrame(cx, frame));
-
+  const selectFrame = async (cx: any, frame: any) => {
+    console.log({ frame });
```
hbenl, line 158: remove

src/ui/actions/network.ts
jcmorrow: I think TS was typing this as `(Frame | undefined)[]`

hbenl: No it types it as `WiredFrame[] | undefined`

src/ui/actions/network.ts
```diff
@@ -126,3 +129,31 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+
+    ThreadFront.timeWarp(point.point, point.time, true);
+    const frames = (await ThreadFront.getFrames()) || [];
+    const formattedFrames = await Promise.all(frames?.map((frame, i) => createFrame(frame, i)));
+    const frame = formattedFrames.find(f => f?.id === frameId);
+    if (!frame) {
+      console.error(`Request frame not found: ${frameId}`);
+      return;
+    }
+    await (new Promise<void>((resolve) => setTimeout(resolve, 50)));
```
jcmorrow, line 156: @hbenl help! I'm having timing issues where because I create a pause the editor changes location, but then I immediately select a source, and for some reason they are getting processed in the wrong order. Doing this solves it but... it feels pretty gross. Is there any way to say "Pause here" without also saying "and go to the editor location of the pause"?

jcmorrow: sorry @souporserious I want to get this one thing taken care of and then I'll get this out!

souporserious: No worries at all! Just using the preview branch to get the stack trace for storyboarding 😎 thank you again for tackling this so quick!

jcmorrow: OK, I think I took care of this, I now allow `timeWarp` to take a frame to go to, rather than automatically going to the top frame of the acquired pause (but we will still go there if no frame is provided).

hbenl: That sounds like the right solution to me.

PR #7098, opened 2022-06-08
Fix the breakpoints-07 e2e test

hbenl:


hbenl:
This fix is just a workaround, the real issue seems to be #7109.

hbenl:
Superseded by #7141

PR #7099, opened 2022-06-08
Architecture exploration: GraphQL

bvaughn:
This PR explores what a Suspense-cache backed GraphQL system might look like. The initial commit adds comments and replies, but is pretty barebones. GraphQL messaging is done via the Fetch API rather than a library. Longer term, we may want to explore a technology like Relay.

The PR also explores the idea of storing _pending comments_ (or replies) in GraphQL as well rather than in-memory. This is something that I think we should do in `main` – although the exact UX (when to save as private vs public) is still TBD.

Here is a Loom walk through of the changes explored here:
https://www.loom.com/share/2b8b587a738f4913af2371613935da0d


bvaughn:
I'm going to go ahead and merge this without review for now since it's self-contained. Want to unblock continued iterating on this branch. Doesn't really require in-depth review at the moment although I'd love to talk about the general exploration with others.

Issue #7090, opened 2022-06-07
Focus Window edge cases


jasonLaster:
Saw two edge cases in this [replay](https://app.replay.io/recording/focus-window-console-messages--d2886ce7-bc4a-4d45-9ed6-7d94556f0372?point=96706529039814627722125151026806784&time=35368.20751761942&hasFrames=false)
1. we say there are too many messages, but dont show any messages in the console
2. the focus window is set to the end, but we say there might be messages after the window

bvaughn:
Happy to look at this but at least for the moment it's not in the current sprint– so I assume it should not be prioritized over other things.

bvaughn:
From what I can see, I think #7101 is the only "fix" here.

However I think we should probably probably also make a UX tweak:
> Perhaps we could change the wording in this case to say something more like "_There are at least X messages outside of your debugging window_"

bvaughn:
#7102 is an optional messaging tweak that _I think_ is a small improvement to the edge case I described above.

PR #7101, opened 2022-06-08
Update filtered logs message

bvaughn:
#7090

Don't show filtered logs before/after zoom regions that are at the start/end of a recording.

https://www.loom.com/share/d6725e1fa4734f449aa965c753a6a4b9


markerikson (approved):

bvaughn:
Hm 🥲 e2e test `breakpoints-02` seems flaky (sometimes fails, sometimes passes, locally) so...

Issue #7032, opened 2022-06-01
If part of the current focus window gets unloaded, we should update the focus window


jcmorrow:
Right now if part of the focus window gets unloaded we will reflect that visually on the timeline, which is great, but we won't update the bounds of the current focus window to reflect it, which means that the user can still seek outside of the loaded regions.

I cannot easily show a recording of this because I can't execute the unload command in the console, but here are repro steps:

- Open a recording
- Focus on the first 2-5 seconds
- Put the playhead in the first second
- Run this in the console: `await app.sendMessage("Session.unloadRegion", {region: {begin: 0, end: 1000}})`
- The first second should unload
- Observe that you can still seek on the timeline to the unloaded area

jaril:
@jcmorrow inferred from soft focus being a soft yes for being done by tomorrow as a signal to add this to the current sprint. Lmk if I got that wrong!

jaril:
@bvaughn mind picking this up?

bvaughn:
> Right now if part of the focus window gets unloaded we will reflect that visually on the timeline, which is great, but we won't update the bounds of the current focus window to reflect it, which means that the user can still seek outside of the loaded regions.

I don't think I see the connection you're making here @jcmorrow. We now explicitly allow you to seek outside of loaded regions (as #6913: Global Loader 1.6) so even if we resized the focus window to fit unloaded regions, it would still be possible to do that.

bvaughn:
Here's a Loom showing what I mean:
https://www.loom.com/share/86716b2dd31f4c08a2a0b45e5eb9fe66?from_recorder=1&focus_title=1

I think this is working as intended now (as of #7064).

bvaughn:
Going to close this as "working as designed" then. Josh please re-open if I'm mistaken!

jcmorrow:
Oh, ha, funny. Yeah, it allowed clicking outside of the window back when that *wasn't* supposed to be allowed, but now that we've changed the rules, it's fine.

PR #7103, opened 2022-06-08
Only re-run breakpoint analyses if the latest analysis result doesn't have enough data

### LINKED ISSUE #6851: Rerun analyses upon some focus/loaded region changes
### Context

- Right now, analyses don't get re-run upon loading or focus window changes.
- We need to rerun analyses for a location when the *loaded* regions change *and*:
  - The location overflowed while finding points
  - The *loaded* regions that we last ran this location's analysis for have changed such that they are not a subset of previously run analyses. (i.e. there are potentially new points to be found).
- This logic is identical to the logic described in #6915 *except* that we care about the loadedRegion windows of *all* past successful analysis runs for that location, not just the most recent successful run.

A subtask of https://github.com/RecordReplay/devtools/issues/6591

---

PREVIOUS CONTEXT FOR POSTERITY

- This is probably not going to be *that* common of a requirement, because, for example, if an analysis gets run *once* at the findPoints level, we will know the correct points for the entire recording and every possible focus region of the recording, and if the analysis has run successfully, we will also know the evaluation at those points, so we will not need to requery until the user changes the expression)
  - A note about points vs location: right now we always run analyses by adding the location of the print statement, but we *could* add a subset of points when we know there will be too many to run a full analysis. This is not a top priority, but it will be a requirement for whatever comes out of https://github.com/RecordReplay/design/issues/36 and should be the kind of operation that our changes will support in the future.
- (In a lot of ways, this is going to line up with other work that we're doing under the umbrella of #6590 , because the loading and focus transitions that would cause analyses to re-run, will also trigger other data refetching)

markerikson:
This PR:

- Starts tracking `focusRegion` in analysis entries, for later comparisons
- Extracts a `MAX_POINTS_FOR_FULL_ANALYSIS` constants so we stop hardcoding `200` all over the place
- Updates `syncFocusRegion` to check whether the latest analysis result for each breakpoint already has the data we need for the new focus region

This should tentatively fix #6851 .  

We may want to improve this to also check against earlier analysis entries as well, but this is at least a step in the right direction.

### DIFF
--- a/packages/protocol/thread/analysis.ts
+++ b/packages/protocol/thread/analysis.ts
@@ -3,6 +3,8 @@ import { ThreadFront } from "protocol/thread";
 import { AnalysisParams } from "../analysisManager";
 import { client, gAnalysisCallbacks } from "../socket";
 
+export const MAX_POINTS_FOR_FULL_ANALYSIS = 200;
+
 export interface Analysis {
   analysisId: string;
   addLocation: (location: Location) => Promise<void>;
--- a/src/devtools/client/debugger/src/reducers/breakpoints.ts
+++ b/src/devtools/client/debugger/src/reducers/breakpoints.ts
@@ -25,6 +25,7 @@ import { getLocationKey, isMatchingLocation, isLogpoint } from "../utils/breakpo
 
 import { getSelectedSource } from "./sources";
 import type { Breakpoint, SourceLocation } from "./types";
+import { FocusRegion } from "ui/state/timeline";
 export type { Breakpoint } from "./types";
 
 type LocationWithoutColumn = Omit<Location, "column">;
@@ -53,6 +54,7 @@ export type AnalysisRequest = {
   id: string;
   location: Location;
   condition?: string;
+  focusRegion: FocusRegion | null;
   points: PointDescription[] | undefined;
   results: AnalysisEntry[] | undefined;
   status: AnalysisStatus;
@@ -170,15 +172,21 @@ const breakpointsSlice = createSlice({
 
     analysisCreated(
       state,
-      action: PayloadAction<{ analysisId: string; location: Location; condition?: string }>
+      action: PayloadAction<{
+        analysisId: string;
+        location: Location;
+        condition?: string;
+        focusRegion: FocusRegion | null;
+      }>
     ) {
-      const { analysisId, location, condition } = action.payload;
+      const { analysisId, location, condition, focusRegion } = action.payload;
 
       analysesAdapter.addOne(state.analyses, {
         error: undefined,
         id: analysisId,
         location,
         condition,
+        focusRegion,
         points: undefined,
         results: undefined,
         status: AnalysisStatus.Created,
--- a/src/devtools/client/debugger/src/reducers/breakpoints.ts
+++ b/src/devtools/client/debugger/src/reducers/breakpoints.ts
@@ -9,7 +9,12 @@ import {
   PayloadAction,
   EntityState,
 } from "@reduxjs/toolkit";
-import { Location, PointDescription, AnalysisEntry } from "@replayio/protocol";
+import {
+  Location,
+  PointDescription,
+  AnalysisEntry,
+  TimeStampedPointRange,
+} from "@replayio/protocol";
 import uniqBy from "lodash/uniqBy";
 import type { Context } from "devtools/client/debugger/src/reducers/pause";
 import string from "devtools/packages/devtools-reps/reps/string";
@@ -54,7 +59,7 @@ export type AnalysisRequest = {
   id: string;
   location: Location;
   condition?: string;
-  focusRegion: FocusRegion | null;
+  timeRange: TimeStampedPointRange | null;
   points: PointDescription[] | undefined;
   results: AnalysisEntry[] | undefined;
   status: AnalysisStatus;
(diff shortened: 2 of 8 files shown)


packages/protocol/thread/analysis.ts
```diff
@@ -3,6 +3,8 @@ import { ThreadFront } from "protocol/thread";
 import { AnalysisParams } from "../analysisManager";
 import { gAnalysisCallbacks, sendMessage } from "../socket";
 
+export const MAX_POINTS_FOR_FULL_ANALYSIS = 200;
```
jcmorrow, line 6: 💯

src/ui/actions/logpoint.ts
```diff
@@ -276,7 +281,7 @@ async function setMultiSourceLogpoint(
     analysis = await createAnalysis(params);
     const { analysisId } = analysis;
 
-    store.dispatch(analysisCreated({ analysisId, location: locations[0], condition }));
+    store.dispatch(analysisCreated({ analysisId, location: locations[0], condition, focusRegion }));
```
jcmorrow, line 284: What do you think about explicitly passing `params.range` here? I think for the purposes of caching in the future, we will not care about it being a `focusRegion` in particular, and instead will want a `TimeStampedPointRange`, since that is a more generic type.

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -170,15 +172,21 @@ const breakpointsSlice = createSlice({
 
     analysisCreated(
       state,
-      action: PayloadAction<{ analysisId: string; location: Location; condition?: string }>
+      action: PayloadAction<{
+        analysisId: string;
+        location: Location;
+        condition?: string;
+        focusRegion: FocusRegion | null;
```
jcmorrow, line 179: I think this should be a `TimeStampedPointRange`, not a `FocusRegion`. `FocusRegion` is a concept that belongs to the DevTools UI. `TimeStampedPointRange` is a protocol-level type, which I think we should prefer when possible.

src/ui/actions/timeline.ts
```diff
@@ -605,6 +619,59 @@ export function setFocusRegionStartTime(startTime: number, sync: boolean): UIThu
...
+  focusRegion: FocusRegion
+): boolean => {
+  // Copy-pasted-tweaked comment from actions/messages.ts:
+  // Soft Focus: The frontend only needs to refetch data if:
+  // 1. The most recent time it requested data "overflowed" (too many hits to send them all), or
+  // 2. The new focus region is outside of the most recent region we ran breakpoint analysis for.
+  //
+  // There are two things to note about the second bullet point above:
+  // 1. When devtools is first opened, there is no focused region.
+  //    This is equivalent to focusing on the entire timeline, so we often won't need to refetch messages when focusing for the first time.
```
jcmorrow, line 634: This is actually not quite the same as console messages. When we first fetch console messages we call `findMessages` (not `findMessagesInRange`) which will attempt to find console messages throughout the recording *regardless of what is attempting to load*. With something like messages we can do this because it's static data that gets pulled out by basic data retrieval. However, by default I believe we attempt to load something like the last minute of the recording. So an initial analysis will run for that portion, not the entire recording, and we need to record that. Another way of saying this is: we care not about the focus region, but about the intersection of the focus region and the loading regions.

jcmorrow (commented):
I think this is close if we ignore the case where part of the recording is unloaded, but I'm not sure that we *should* ignore that part. This is definitely an edge case, but here's what can happen on production right now:
I'm focusing on roughly the first four seconds of the recording, during which there are 80 hits:
CleanShot 2022-06-09 at 11 24 42@2x

I intentionally unload the first second of the recording:
CleanShot 2022-06-09 at 11 25 02@2x

Now, notice that the breakpoint timeline still thinks that the first second is loaded and focused (which is a slightly separate bug). But the analysis also did not rerun, because the *focus* window did not change, but the loaded regions (and thus the area that would be analyzed currently) has changed. That counter should say 32 instead of 80. One way to fix this is to re-open #7032 . Another would be to store another window that we care about - the overlap of the focus and loaded windows. Right now, I think #7032 is probably the way to go, but we still care about this concept for the case where there is no focus window and only a loading window (which is the case on initial load of a large recording).

src/ui/actions/logpoint.ts
bvaughn: This might also transition more smoothly to when/if we eventually re-add the concept of zoom region in the UI. Then the "range" could just be focus (if there is one) or zoom (if there is one) or null.

jcmorrow (approved):

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -420,6 +434,54 @@ const customAnalysisResultComparator = (
...
+
+  const analysisOverflowed =
+    status === AnalysisStatus.PointsRetrieved && points.length > MAX_POINTS_FOR_FULL_ANALYSIS;
+
+  const analysisLoaded = [AnalysisStatus.PointsRetrieved, AnalysisStatus.Completed].includes(
+    status
+  );
+
+  // TODO Drop this conversion when we convert FocusRegion to TimeStampedPointRange
+  const rangeAsFocusRegion: UnsafeFocusRegion = {
```
jcmorrow, line 464: btw - there is a function `rangeForFocusRegion` specifically made so you don't have to do this by hand 😄

PR #7106, opened 2022-06-08
github: cache all npm installs to improve CI performance

jazzdan:


jasonLaster (approved):

.github/actions/test/action.yml
```diff
@@ -3,8 +3,7 @@ description: "Run end-to-end tests. Note that this action assumes that Node and
 runs:
   using: "composite"
   steps:
-    - name: Install dependencies
-      run: npm install
+    - uses: bahmutov/npm-install@v1
```
jcmorrow, line 6: What do you think about using something a little more official like the GitHub action for node?  https://github.com/actions/setup-node#caching-global-packages-data

jasonLaster:
NOTE the official version does not cache node_modules

jcmorrow:
> NOTE the official version does not cache node_modules

Huh... That seems odd? Oh well, I guess we go with this then 👍

hbenl:
We could run into the issue that `npm install` fails sporadically without `--force` again, but I think it's worth a try.

PR #7107, opened 2022-06-09
Show unloaded regions on the breakpoint timeline

jcmorrow:
Prior to this change, we would only show *unfocused* regions on the
breakpoint timeline, but unloaded regions would appear as if they were
both loaded and focused. Now, we look at the intersection of the loaded
and focused windows.

We still show hits in the unloaded region, which is unfortunate, but I
think the real solution there is not to do filtering, but rather to
adjust the currently focused window when a region gets unloaded.

Before:

CleanShot 2022-06-09 at 11 25 02@2x

After:

CleanShot 2022-06-09 at 11 56 25@2x

### DIFF
--- a/src/ui/components/Timeline/UnloadedRegions.tsx
+++ b/src/ui/components/Timeline/UnloadedRegions.tsx
@@ -6,6 +6,8 @@ import { getFocusRegion, getZoomRegion } from "ui/reducers/timeline";
 import {
   endTimeForFocusRegion,
   getVisiblePosition,
+  overlap,
+  rangeForFocusRegion,
   startTimeForFocusRegion,
 } from "ui/utils/timeline";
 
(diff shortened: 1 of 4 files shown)


jasonLaster:
Got it. side note, when you say "focus region" do you mean "focus window" find using the word region in multiple places confusing

markerikson (approved):

jcmorrow:
> Got it. side note, when you say "focus region" do you mean "focus window" find using the word region in multiple places confusing

yeah, sadly it is `FocusRegion` in the code 😭 

So I've been using that language just to be consistent, but I agree, it's a window, not a `Region`.

src/ui/components/Timeline/UnloadedRegions.tsx
```diff
@@ -29,8 +31,9 @@ export const UnloadedRegions: FC = () => {
   let beginTime = begin.time;
   let endTime = end.time;
   if (focusRegion) {
-    beginTime = startTimeForFocusRegion(focusRegion);
-    endTime = endTimeForFocusRegion(focusRegion);
+    const focusedAndLoaded = overlap([rangeForFocusRegion(focusRegion)], loadedRegions.loading)[0];
+    beginTime = focusedAndLoaded.begin.time;
+    endTime = focusedAndLoaded.end.time;
```
bvaughn, line 36: I think this might be a potential bug scenario. If `overlap` doesn't return at least (exactly?) one item, `focusedAndLoaded` will be undefined.

bvaughn: I think [this recording](https://app.replay.io/recording/hit-counts-are-off-typeerror--8759e652-2916-4453-9510-59615c9389e1) shows that scenario, where the times passed in are 12-48s and 50-110s (which don't overlap).

bvaughn: Potential fix in #7202 – though I haven't thoroughly tested it.

PR #7114, opened 2022-06-09
Make Soft Focus the Default

jcmorrow:
I've been saying for the past week that we basically already shipped soft focus, but not technically. I'd like to clarify what I meant by that, because it's confusing. Let's start from the beginning-ish.

I started [a document](https://www.notion.so/replayio/Loading-Region-Problems-b8dc0d03c49840f8b504e967c7ea5778) on April 26th describing an underlying cause for an entire family of problems that I was seeing in the devtools UI. Around that same time, we were also seeing a *lot* of crazy behavior with regions loading and unloading (I'm gonna come back to this). The big takeaway from that document was that we needed methods for focusing at slices more specific than regions. We did a lot of work to support this.

- We added range parameters to backend endpoints, so that front-end requests could send the range they were interested in, rather than always addressing the entirety of all loaded regions
- We reworked how we run, monitor, and store the results of analyses generally
- We built two completely new backend endpoints for converting from times to TimeStampedPoints
- We built a thunk which observed the focus mode, and decided when it should attempt to reload some resources (starting with console messages, and then analyses) based on the newly chosen window and previous fetches

I'd like to pause at that last one. In many ways, that was the ultimate goal at the time I wrote that document. It solves all "edge-piece" and "overflow" problems, and while we can go even farther in the direction of efficiency, I'm quite happy with the soundness of the foundation we've laid, and soundness was the motivator for this project.

We shipped the first version of everything up above one week ago, and sent it out with the newsletter. However, remember how I said that back when I was first scoping this project we were having a lot of trouble with loading and unloading regions? Well, one of the things that I realized early on was that if we correctly handled edge-pieces, it no longer mattered how big the edge pieces were. And if it does not matter how big the edge pieces are, then there's no reason to forcibly unload parts of the recording in order to exclude them. Back when unloading and loading were causing massive headaches both for us and for users, this seemed like a huge win, so I made it the final goal of Soft Focus - be able to move your focus window around, without ever having to unload a region.

My first shot at it was kind of wonky (you'll see just how wonky in a second), and I didn't love it. And while we were busy getting everything ready so we *could* turn this flag on, the rest of the team shipped Turbo Replay, and all of a sudden, it did not matter that much if we forcibly loaded or unloaded regions. It's still a nice performance win I think, for users who are moving their focus window around a lot, and more importantly, we corrected a ton of soundness problems on the way here.

---

OK, so back to this change.

The final solution is probably best explained by this comment I wrote earlier today to explain a type (if you've been deep in the weeds on this stuff already a lot of this will sound familiar):

>  You might be wondering why there are multiple `begin`s and `end`s for a single FocusRegion. Well, this is an artifact of the difference between world-time, which is essentially continuous (and thus, linearly interpolable), and execution time, which is both discrete (in the sense that even though there may be many points "between" points 100 and 200, there may not be any points that are valid for operations like Pausing) and irregularly spaced (the recording might have many valid points between the times 100-150ms, but no valid points between 150-500ms).
> 
> It is very convenient to accept user input in the form of times. The user can click somewhere on the timeline (which is really just a visual interpretation of the continuous, evenly-spaced number line underlying it), and we can try and use that to filter things down to that time.
> 
> And that is where the trouble starts.
> 
>  In order to accurately filter resources that are in memory, or to specify a range to a backend endpoint which accepts ranges, we need to bound our range by *points*, not just *times*. But the user did not give us a point, they only gave us a time. So what should we do? Well, we try to convert from our time to a point. However, because points are discrete and unevenly spaced, this is not an exact translation, it introduces a small amount of error (usually on the order of <100ms, but occasionally more). Also, to be as precise in the conversion as possible we need to hit a backend endpoint (getPointsBoundingTime), which is not horribly expensive, but is significantly more expensive than a simple lookup, for instance.
> 
>  All of this causes problems, especially when the user is dragging around a focus window. This is because:
>  1) The little errors start to add up. If you "snap" the continuous values of
>  the timeline to the discrete values of the pointline, you introduce a little bit of error. Then, maybe the user drags the line another pixel, and you introduce another little error, and so on and so forth and, well screens have many pixels so this can get out of hand quickly.
>  2) A difference of less than 100ms might not be super noticeable on the timeline, but as the errors get larger (or if the recording is quite short, for instance) it will start to be clear to the user that we are not actually putting them at quite the time that they asked for. And, well, we have very good reasons for doing that, but they are very hard to explain (for instance, we might have to subject them to reading something like this comment).
>  3) If we try and do an exact lookup every time the focus window changes while selecting it, we will end up sending a *ton* of protocol messages.
> 
>  The solution I have come up with so far is that we store *two* windows. One window is used for display. It is what we show on timelines, and it is also what we use as the inputs to our time -> point transformation *after* the user has chosen their focus window.
> 
>  The other time is a proper TimeStampedPointRange, which we use to do things like filter resources by point and interact with backend endpoints which accept ranges.

With the above solution, we can ship Soft Focus (in the sense that there does not need to be any forced unloading) without degrading the experience of picking a focus window or having to expose the messiness of the above to the user. Next, will be actually utilizing `Session.getPointsBoundingTime` to be more specific in our mapping, but even now, we are pretty darn specific most of the time, so I wanted to get this in rather than waiting for that optimization piece to be done.

As for the changes this actually makes:

- Deletes the `Soft Focus` flag from experimental settings
- Changes `startTimeForFocusRegion` to `displayedStartForFocusRegion`
- Likewise for `endTimeForFocusRegion` to `displayedEndForFocusRegion`
- Displays the `displayed` time values when choosing a window, rather than the mapped-points' time values

That's it! Not a ton of changes, but building on a much larger set over the past 6 weeks 🏗️ 

Fixes https://linear.app/replay/issue/FE-160/soft-focus

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -1,6 +1,6 @@
 version: 0.1
 cli:
-  version: 0.12.1-beta
+  version: 0.13.0-beta
 lint:
   enabled:
     - actionlint@1.6.8
--- a/src/devtools/client/debugger/src/actions/source-actors.ts
+++ b/src/devtools/client/debugger/src/actions/source-actors.ts
@@ -5,7 +5,7 @@
 import type { UIThunkAction } from "ui/actions";
 import { getFocusRegion } from "ui/reducers/timeline";
 import { PROMISE } from "ui/setup/redux/middleware/promise";
-import { endTimeForFocusRegion, beginTimeForFocusRegion } from "ui/utils/timeline";
+import { displayedEndForFocusRegion, displayedBeginForFocusRegion } from "ui/utils/timeline";
 
 import {
   getSourceActor,
@@ -93,10 +93,10 @@ export const loadSourceActorBreakpointHitCounts = memoizeableAction(
         id,
         Math.floor(lineNumber / MAX_LINE_HITS_TO_FETCH) * MAX_LINE_HITS_TO_FETCH,
         state.timeline.focusRegion
-          ? beginTimeForFocusRegion(state.timeline.focusRegion)
+          ? displayedBeginForFocusRegion(state.timeline.focusRegion)
           : "no_focus_start",
         state.timeline.focusRegion
-          ? endTimeForFocusRegion(state.timeline.focusRegion)
+          ? displayedEndForFocusRegion(state.timeline.focusRegion)
           : "no_focus_end",
       ].join("-");
       return key;
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -29,7 +29,7 @@ import type { AppDispatch } from "ui/setup/store";
 import type { UIState } from "ui/state";
 import { isVisible } from "ui/utils/dom";
 import { convertPointToTime } from "ui/utils/time";
-import { endTimeForFocusRegion, beginTimeForFocusRegion } from "ui/utils/timeline";
+import { displayedEndForFocusRegion, displayedBeginForFocusRegion } from "ui/utils/timeline";
 
 import ConsoleLoadingBar from "./ConsoleLoadingBar";
 import styles from "./ConsoleOutput.module.css";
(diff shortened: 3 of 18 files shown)


jcmorrow:
Sorry I meant to make this a draft, it's not *quiiiiite* ready yet.

jcmorrow:
Sits on top of #7135

PR #7118, opened 2022-06-09
Console overlay buttons compare time if no paused execution points

### LINKED ISSUE #7108: Missing forward/rewind buttons in the console
See https://discord.com/channels/779097926135054346/984197127565627422
Here are 2 recordings of the same e2e test (with the same underlying example recording and the same devtools preview deploy):
https://app.replay.io/recording/1cd332cc-7bd5-43a7-90ed-b48a18a7ecf8 - this one was made in CI, where the test failed
https://app.replay.io/recording/3db5ca49-de5e-4488-82bc-32e2f8787aeb - this one was made manually, here the test succeeds
Note that you can't *see* that the buttons are missing because this is an e2e test recording. I was unable to reproduce the missing buttons while recording.
The recordings contain comments explaining what went wrong in the first recording and how it works in the second.

bvaughn:
The console uses the paused execution point to determine whether to show a "fast-forward" or "rewind" label. There sometimes is no paused execution point though, in which case the UI now falls back to comparing the message's time to the app state currentTime.

Resolves #7108

### DIFF
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -19,7 +19,12 @@ import { ContextMenu } from "ui/components/ContextMenu";
 import { Dropdown, DropdownItem } from "ui/components/Library/LibraryDropdown";
 import Icon from "ui/components/shared/Icon";
 import { selectors } from "ui/reducers";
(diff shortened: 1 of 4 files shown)


markerikson (approved):
Seems reasonable to me

jaril (approved):

PR #7122, opened 2022-06-09
Move logpoint UI above editor line

bvaughn:
This more closely matches what Chrome devtools does _and_ (I think) brings the UI into alignment with how the feature actually works.

Here's a Loom walk through: https://www.loom.com/share/726fe462e65c408ab76a9b4366b31b9b

---

# How do things currently work?

## In Replay

image

My mental model of this based on the way the UI presents is:
```js
const canShowCount = lastFetchedFocusRange === null && !lastFetchDidOverflow;
console.log(`canShowCount? ${canShowCount} focus window? ${lastFetchedFocusRange === null}`);
```

## In Chrome

image

But my intuition from the way the Chrome UI presents things is:
```js
console.log('...'); // I wouldn't expect to be able to log _reactDOM here
var _reactDOM = _interopRequireDefault(require("react-dom"));
```

---

Fixes BAC-1673

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Widget.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Widget.tsx
@@ -22,6 +22,7 @@ export default function Widget({ location, children, editor, insertAt }: WidgetP
     }
     const editorLine = toEditorLine(location.line || 0);
     const _widget = editor.codeMirror.addLineWidget(editorLine, node, {
+      above: true,
       insertAt,
     });
 


souporserious:
I've done the logging on the same line quite a few times, but I don't think moving the panel above is the right call here. Like you mentioned with column breakpoints I can see us solving it like this:

1. We keep the UI as is even though it's a little wonky and wait until column breakpoints are supported.
2. Add a UI to show and control column breakpoints which will make it more apparent that we can't execute anything after that point.
3. Allow jumping to the column after where you are trying to log if it is available on that line which gives us an opportunity to educate you on this feature and not block what you're trying to do.

bvaughn:
I don't think column breakpoints fixes this, to be honest. I've seen a few people mention it in the pipeline as though it fixes this but I think the fundamental problem is still there: we read top to bottom. Program execution goes from top to bottom. Showing something _under_ a thing implies that it happens _after_ a thing. I think that's a fundamental problem with our log point UI.

jonbell-lot23:
Adding link to our Discord thread:
https://discord.com/channels/779097926135054346/984530777771503626/984530895946014730

I don't think column breakpoints *solve* the issue but they definitely become part of the holistic solution we come up with. So as we mentioned in the thread, I think the key is

1) Everyone turn on column BPs
2) Using column BPs as a foundational part of the interface, figure out how to make this work best

Also worth pointing out that we could end up in a 51/49 space, where there are tradeoffs either way, and not one that's inherently better. Personally, the "people read top to bottom" argument has always resonated pretty hard for my UX brain ... but I know there are a lot of tradeoffs in play, because this feature is more complicated than just reading a paragraph of text from top to bottom.

bvaughn:
FWIW here's what things look like with column breakpoints enabled:

### Current
Screen Shot 2022-06-09 at 4 12 38 PM

### Branch
Screen Shot 2022-06-09 at 4 11 58 PM

IMO above is still an improvement _but_ that's subjective.

PR #7130, opened 2022-06-09
Feature flag the print statement panel

jasonLaster:
Moves https://github.com/replayio/devtools/pull/7122 to a feature flag so we can continue exploring options for communicating where the expression is evaluated. 

Positioning the panel above or below does not solve this problem it just suggests that it comes before the end of the line. Additionally, placing the panel above the line and not adjusting the add/close button is currently awkward.

Lastly, this is a significant enough change, that it would be nice to package it up with Travis' work and update the docs and share a changelog post when it lands


bvaughn (approved):
Personally I think this is too buried (it's not exposed in the UI, right? only in the console?) and we're preserving the "broken" behavior by default 😁  but the change seems okay if we want to roll the change out slowly.

PR #7135, opened 2022-06-10
Change `start` to `begin`

jcmorrow:
This more closely matches the protocol language, and it felt like a good time to make the change since I'm also about to make some changes to focus operations for soft focus.


packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -27,19 +27,19 @@ export default function MessagesList() {
 
   let focusRange = null;
   if (range !== null) {
-    const [startTime, endTime] = range;
+    const [beginTime, endTime] = range;
```
bvaughn, line 30: omg :heart_eyes: thank you

bvaughn (approved):
:heart_eyes:

PR #7140, opened 2022-06-10
Update GraphQL types (FE-232)

hbenl:
@bvaughn We forgot to run `npm run gql` to regenerate the typescript types from the GraphQL schema after the last GraphQL deploy.


bvaughn:
Oh! Thanks.

PR #7144, opened 2022-06-10
FE-183: Debounce evaluations in editor hover and console eager evaluation

markerikson:
This PR:

- Migrates `actions/preview` and `Preview/index` to TS
- Debounces the evaluation on editor token hover by 300ms
- Debounces console eager evaluation by 300ms


src/devtools/client/webconsole/utils/autocomplete-eager.ts
```diff
@@ -64,9 +65,13 @@ async function eagerEvaluateExpression(
 
 export function useEagerEvaluateExpression() {
   const frame = useSelector(getSelectedFrame);
-  const callback = useCallback(
-    (expression: string) =>
-      frame ? eagerEvaluateExpression(expression, frame.asyncIndex, frame.protocolId) : null,
+  const callback = useMemo(
+    () => (expression: string) => {
+      if (frame) {
```
jcmorrow, line 70: This is the tiniest nit in the entire world, but I think I would reorder these to make the check read like a guard clause:

```
if (!frame) {
  return null;
}

return eagerEvaluateExpression(expression, frame.asyncIndex, frame.protocolId);
```

Either way is fine, I just feel like conforming to really common patterns like guard clauses can make the cognitive load a little lighter while reading.

src/devtools/client/debugger/src/reducers/preview.ts
```diff
@@ -8,8 +8,13 @@ import type { UIState } from "ui/state";
 
 import type { Preview } from "./types";
 
+// TODO Are we really putting more DOM nodes in state?
```
jcmorrow, line 11: Ew, are we?

jcmorrow (commented):
I'm gonna wait to play with the preview to 👍 but here's my comments

src/devtools/client/debugger/src/components/Editor/Preview/index.tsx
```diff
@@ -34,40 +52,54 @@ class Preview extends PureComponent {
...
+
+  startPreview = debounce((target: HTMLElement, tokenPos: SourceLocation) => {
     const { cx, editor, updatePreview } = this.props;
 
-    if (cx.isPaused && !this.state.selecting) {
+    // Double-check status after timer runs
+    if (cx?.isPaused && !this.state.selecting) {
       updatePreview(cx, target, tokenPos, editor.codeMirror);
     }
+  }, 300);
```
jcmorrow, line 80: I think this is too high. I don't want it to trigger for symbols that the user's cursor is *literally* just passing over, but I also think that it should feel as instant as possible. I would try lowering this to like 50 or 100.

jcmorrow (approved):
I think we should crank the timeout down on the hover target. I'm fine with erring on the side of firing off a *couple* of extra analyses in order to make it feel less laggy, as long as we are not literally firing off on every single symbol that the cursor touches 👍

PR #7152, opened 2022-06-13
Add a "Default to Dev Tools" experimental option

jcmorrow:
### Issues
- We want to have an option for people to default to opening all Replays to dev mode
- We had a `prefs.viewMode` option... but it was not used anywhere, and it was a `string` value, which did not play as nicely with the API I wanted for this flag (boolean)
- We still had `codeHeatMaps` options, even though we've stopped those from doing anything recently
- `RecordingRow` was adding an action prop `loadReplayPrefs`, which was the only thing it was getting from redux, and that prop was not used, so that component does not need to be connected
- The prefs file, which has many arbitrarily sorted lists, had things in pretty haphazard order. I think sorting lines alphabetically is not *always* the right call, but in this place, I think it's a good fit

### Changes
- Add a `Default to Devtools Mode` flag and UI in the experimental preferences section
- Delete `prefs.viewMode`
- Delete `loadReplayPrefs`
- Disconnect `RecordingRow`
- Delete `codeHeatMaps`
- Sort lists of values in the prefs file

### DIFF
--- a/src/ui/components/Library/RecordingRow.tsx
+++ b/src/ui/components/Library/RecordingRow.tsx
@@ -61,20 +61,20 @@ function RowWrapper({
   );
 }
 
-type RecordingRowProps = PropsFromRedux & {
-  recording: Recording;
-  isEditing: boolean;
-  selected: boolean;
+type RecordingRowProps = {
   addSelectedId: (recordingId: RecordingId) => void;
+  isEditing: boolean;
+  recording: Recording;
   removeSelectedId: (recordingId: RecordingId) => void;
+  selected: boolean;
 };
 
 function RecordingRow({
-  recording,
-  isEditing,
-  selected,
-  removeSelectedId,
   addSelectedId,
+  isEditing,
+  recording,
+  removeSelectedId,
+  selected,
 }: RecordingRowProps) {
   const { permissions, loading: permissionsLoading } = useGetUserPermissions(recording);
   const allowSelecting =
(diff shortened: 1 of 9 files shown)


src/ui/reducers/layout.ts
```diff
@@ -2,12 +2,13 @@ import { sourcesDisplayed } from "devtools/client/debugger/src/reducers/ui";
...
 import { UIState } from "ui/state";
 import { LayoutState } from "ui/state/layout";
+import { features, prefs } from "ui/utils/prefs";
 
 export const syncInitialLayoutState: LayoutState = {
   consoleFilterDrawerExpanded: true,
   showCommandPalette: false,
   selectedPrimaryPanel: "events",
-  viewMode: "non-dev",
+  viewMode: features.defaultToDevMode ? "dev" : "non-dev",
```
bvaughn, line 11: I wonder if this prop should just be a boolean too?

src/ui/components/shared/UserSettingsModal/ExperimentalSettings.tsx
```diff
@@ -44,9 +44,9 @@ const EXPERIMENTAL_SETTINGS: ExperimentalSetting[] = [
     key: "enableBreakpointPanelAutocomplete",
   },
   {
-    label: "Code Heatmaps",
-    description: "Calculate hit counts for editor files all at once",
-    key: "codeHeatMaps",
+    label: "Default to Dev Mode",
```
jasonLaster, line 47: perhaps: `Default to Dev Tools`

markerikson (approved):

jasonLaster (approved):
This is fine for now, but long term, feels like something that should live under settings > appearances

https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=12822%3A103430
Screen Shot 2022-06-13 at 11 13 11 AM

src/ui/components/Library/RecordingRow.tsx
```diff
@@ -188,11 +188,9 @@ export function ItemScreenshot({ recordingId }: { recordingId: RecordingId }) {
...
-      <div>{screenData && <img className="object-contain w-full h-9" src={screenData} />}</div>
+      <div>{screenData && <img className="h-9 w-full object-contain" src={screenData} />}</div>
     </Redacted>
   );
 }
 
-const connector = connect(() => ({}), { loadReplayPrefs: actions.loadReplayPrefs });
-type PropsFromRedux = ConnectedProps<typeof connector>;
-export default connector(RecordingRow);
+export default RecordingRow;
```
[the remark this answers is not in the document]
jcmorrow: FWIW, *all* tailwind changes in this request were done by the linter thing we have, so I am in the same boat as you! ⛵

src/ui/reducers/layout.ts
jcmorrow: I'm down for this.

jcmorrow: Actually, upon looking into this, it does not quite feel right to me. I think for the purposes of having a value that backs a UI checkbox, using a `boolean` is reasonable, but if we push this change down to the actual state, then we are making it harder to make a change that might add a third-view (not that this is an immediate concern, but just thinking of the future dev who might have to back these changes out if that happened).

jcmorrow: Oh, lol, I totally did not realize that Jason's comment was in response to the original comment here. I don't like that GH shows replies as separate comments when you are just looking at the "conversation" tab

PR #7153, opened 2022-06-13
Add up and down arrow key shortcuts to net monitor

jcmorrow:
This way you can easily select the next or previous request.

Also, this makes it so that if we take a while to load request or response bodies, we will show "Loading..." instead of just not showing anything (which feels like a bug).

### DIFF
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -1,5 +1,6 @@
 import { getThreadContext } from "devtools/client/debugger/src/selectors";
 import SplitBox from "devtools/client/shared/components/splitter/SplitBox";
+import { findIndex } from "lodash";
 import React, { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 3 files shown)


src/ui/components/NetworkMonitor/index.tsx
```diff
@@ -81,8 +82,16 @@ export const NetworkMonitor = ({
...
         let selectedRequest;
+        let previousRequestId = null;
+        let nextRequestId = null;
         if (selectedRequestId) {
-          selectedRequest = data.find(request => request.id === selectedRequestId);
+          selectedRequest = data.find((request, i) => {
+            if (request.id === selectedRequestId) {
+              previousRequestId = i > 0 ? data[i - 1].id : null;
+              nextRequestId = i + 1 < data.length ? data[i + 1].id : null;
+              return true;
```
jcmorrow, line 92: this is kind of ugly, open to rephrasings

markerikson (approved):

bvaughn (approved):

PR #7154, opened 2022-06-13
Fix logout to return to login page

ryanjduffy:
## Issue

Logging out of replay redirected to `http://localhost:3000`. This was cause both by a bad default configuration in auth0 and the absence of `returnTo` in the `logout()` call.

## Resolution

* Update Auth0 config to prioritize the production app for a logout without a `returnTo`
* This PR to add a `returnTo` pointing at the login page for logout actions

### DIFF
--- a/pages/login.tsx
+++ b/pages/login.tsx
@@ -12,5 +12,5 @@ export default function LoginPage() {
     dispatch(clearExpectedError());
   }, [dispatch]);
 
-  return <Login returnToPath={"" + router.query.returnTo} />;
+  return <Login returnToPath={"" + (router.query.returnTo || "/")} />;
 }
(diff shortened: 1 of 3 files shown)


pages/login.tsx
```diff
@@ -12,5 +12,5 @@ export default function LoginPage() {
     dispatch(clearExpectedError());
   }, [dispatch]);
 
-  return <Login returnToPath={"" + router.query.returnTo} />;
+  return <Login returnToPath={"" + (router.query.returnTo || "/")} />;
```
ryanjduffy, line 15: This eliminates the requirement that `returnTo` be provided on the query string. Without the fallback `returnTo` was stringified to `"undefined"` by concatenating it with `""`.

jasonLaster (approved):

PR #7156, opened 2022-06-13
Move default mode to settings

jcmorrow:
CleanShot 2022-06-13 at 21 53 56@2x


jcmorrow:
@ceceliacreates I did not ping you on this one earlier (which I guess might be for the best since we're changing it again so soon), but I guess this might the kind of thing we would document somewhere?

PR #7157, opened 2022-06-13
More print statements polish

souporserious:
WIP


packages/components/PrefixBadgePicker/Picker.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+        if (!child.props.id) {
+          const name = typeof child.type === "string" ? child.type : child.type.name;
+          throw Error(`Immediate Picker child "${name}" must pass an "id" prop.`);
+        }
+
+        const isToggle = child.props.id === "toggle";
+        const showToggle = value === undefined && isToggle;
+        const isChildActive = value === child.props.id || showToggle;
+        const wasChildActive = previousId.current === child.props.id;
```
souporserious, line 81: All of this state is gross and I want to clean it up in the near future. It is handling the active state for the toggle and other children to display styles properly.

packages/components/PrefixBadgePicker/PrefixBadgePicker.tsx
```diff
@@ -29,153 +21,57 @@ export function PrefixBadgePicker({
...
-              if (state === "closed") {
-                setState("opening");
-              }
-            }}
-            onMouseUp={() => {
-              if (state === "opened") {
-                handleSelect();
-              } else {
-                setState("opened");
-              }
```
souporserious, line 100: This mouse down interaction was causing issues with the click interaction and keyboard so I'd like to revisit this and probably use a library so it's accessible.

packages/components/PrefixBadgePicker/PrefixBadgePicker.tsx
```diff
@@ -29,153 +21,57 @@ export function PrefixBadgePicker({
...
-        className={
-          name === "unicorn" ? styles.UnicornBadge : classNames(styles.ColorBadge, styles[name])
-        }
+    <motion.button {...props} className={classNames(styles.PrefixBadge, props.className)}>
+      <motion.span
+        layout
+        variants={{
+          opened: { width: "0.5rem", height: "0.1rem" },
+          closed: { width: "0.3rem", height: "0.3rem" },
+        }}
```
souporserious, line 72: Variants are a pattern I want to explore in our design system since it works well with composition like they're being used here. It's a simple idea of a parent (Picker in this case) being able to set which variants are active and then cascading down so children can describe their state at that time. This is similar to the classnames package, but would be a first-class feature of our design system and would include JS animations when needed.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
```diff
@@ -7,33 +7,21 @@
...
-.breakpoint-panel-wrapper > :first-child {
-  border-top-left-radius: 4px;
-  border-top-right-radius: 4px;
-}
-
-.breakpoint-panel-wrapper > :last-child {
-  border-bottom-left-radius: 4px;
-  border-bottom-right-radius: 4px;
-}
-
```
souporserious, line 19: @jonbell-lot23 not sure if you know if there was a reason we did this and not overflow hidden on the container, but it fixes the background spilling out of the container edges. The only reason I'd see a problem is if we had a popover menu nested inside, but we should be able to just render that at the end of the document to avoid any issues there.

packages/components/AddCommentButton/AddCommentButton.tsx
```diff
@@ -25,22 +25,29 @@ export const AddCommentButton = forwardRef<HTMLButtonElement, AddCommentButtonPr
...
     const [isHovered, setIsHovered] = useState(false);
-    const isOpened = isHovered;
+    const hoverRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+    const maybeHover = () => {
+      hoverRef.current = setTimeout(() => setIsHovered(true), 300);
+    };
+    const clearHover = () => {
+      hoverRef.current && clearTimeout(hoverRef.current);
+      setIsHovered(false);
+    };
```
bvaughn, line 35: If we decide to move forward with this pattern– maybe we could wrap it in a `useDebouncedHover` custom hook?

Although something feels a bit off about adding delay for pop-outs like this (to me). I totally trust your judgement on the interaction though.

packages/components/PrefixBadgePicker/Picker.tsx
bvaughn: Agreed. To me, this feels really complicated to read and reason about compared to the UI before before animation was added to this component.

Do you think we'll be adding this kind of interaction in many places? If we do, I hope we abstract it away somehow. Would love to hear your thoughts about this somehow– how we can do this sort of thing in a maintainable way.

bvaughn (approved):
> Lastly, the `PrefixBadgePicker` is a nice interaction I'd like to explore for other areas, but is deceivingly complex since we're adding another layer on top of our current CSS so I'm going to hold off on anything like this again until we can address our design system.

Strong agree here. Building on top of our current shaky foundation to add something more complex like animations feels worrisome to me.

---

Note this branch has a small regression:
https://www.loom.com/share/52dcdb66568f49a0b21eb0b22f951b93

---

I think this branch/component is kind of an exploration, which is neat. The visual animation/effect feels really smooth. I have some concerns about the increased code complexity though. If possible, I'd like for us all to talk about how we're going to manage that going forward before we add more animations like this to the app.

jasonLaster:
Love this attention to detail.

packages/components/PrefixBadgePicker/Picker.tsx
souporserious: I don't see it happening in many places so these spots _should_ be better isolated, but as the team and codebase grows I know that can easily get out of hand. I envision us getting our design system in place which will take all of this into consideration and keep the complexity wrapped in lower-level components that can communicate efficiently when used in the application.

packages/components/AddCommentButton/AddCommentButton.tsx
souporserious: Sounds good! Yeah, I don't personally like the large slide-out label and might push for this to just be a tooltip with some of the other designs I've been exploring here.

souporserious:
Thanks for the review and comments, @bvaughn! 

> If possible, I'd like for us all to talk about how we're going to manage that going forward before we add more animations like this to the app.

Yeah, once the dust settles on the new marketing site I will shift my focus here since I don't want this to be a blocker for us to add delightful animations and I think we can keep most of the UI interaction and animation complexity wrapped up in our system components to keep from spilling into application code.

bvaughn:
> Note this branch has a small regression:
> https://www.loom.com/share/52dcdb66568f49a0b21eb0b22f951b93

@souporserious Did you catch this comment in my review?

I didn't see a commit after my comment with a fix, but maybe I missed it?

souporserious:
Oops, I missed that, my bad! Working on a fix right now.

jaril:
@ceceliacreates user facing change 🙋‍♂️

PR #7158, opened 2022-06-13
Proof of Concept prototype: use RTK Query for fetching Replay Protocol data to show Pause stack frames

markerikson:
This PR adds another proof-of-concept prototype to the `packages` folder.  This prototype uses RTK Query to manage data fetching of source entries, source text + line hit counts, hit points, and pause frame data.  It also uses CodeMirror 6 to show the highlighted source text.

- Set up another POC prototype project under the `packages` folder, with the usual Next+Redux store setup
- Added RTK Query to that Redux store
- Defined RTKQ endpoints for fetching the list of sources, source text + line hits by `sourceId`, `PointDescription` hits by `Location`, and pause data by `PointDescription`
- Added selection of a source entry
- Added CodeMirror 6
- Figured out how to handle click events in CM and convert screen cords to a line+column
- Found the nearest breakable location to that line+column
- Queried the hit times for that spot
- Added selection of a hit time
- Queried the pause data for a selected hit time
- Added display of the frame function names + locations

UI screenshot:

image

Loom showing the demo in action briefly:

https://www.loom.com/share/fced012fcce64c28a55ab7d132ce2a4f

**Note** While this is a POC, I _did_ end up upgrading the main app to React-Redux v8.  Per discussion with Jason, we actually do want to go ahead and do this in the next day or two anyway - there's no reason not to be on the latest React-Redux version.  But, I'd want to hold off on merging this PR until that's complete.


markerikson:
Bah, TS compilation issues.  I'll fix those tomorrow.

markerikson:
Ready for review. Since it's a separate prototype, it _should_ be generally safe to merge. The only real concern is the package management shenanigans with NPM.

packages/markerikson-stack-client-prototype/src/app/api.ts
```diff
@@ -0,0 +1,147 @@
...
+  node_modules: newSource[];
+  other: newSource[];
+}
+
+const reIsJsSourceFile = /(js|ts)x?$/;
+
+const CACHE_DATA_PERMANENTLY = 10 * 365 * 24 * 60 * 60;
+
+export const api = createApi({
+  baseQuery: fakeBaseQuery(),
```
jcmorrow, line 27: I've used RTK Query for interacting with REST API's, and it is *awesome*. We definitely get *some* of the awesomeness by coercing it to work with the websocket client, but it does feel like an awkward fit to me.

jcmorrow (approved):
I don't have a ton to mention here. I really, really like RTK Query for some things. This has convinced me that it is *possible* to use it with our websocket setup, but it has not convinced me that we *should*. (I'm also not saying that I think we shouldn't, it just doesn't seem like a home run to me).

packages/markerikson-stack-client-prototype/src/features/sources/SourceContent.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [selectedLocation, setSelectedLocation] = useState<Pick<
+    Location,
+    "line" | "column"
+  > | null>(null);
+  const dispatch = useAppDispatch();
+  const selectedSourceId = useAppSelector(state => state.sources.selectedSourceId);
+  const selectedPoint = useAppSelector(state => state.sources.selectedPoint);
+
+  const { currentData: sourceText } = useGetSourceTextQuery(selectedSourceId ?? skipToken);
+  const { currentData: sourceHits } = useGetSourceHitCountsQuery(selectedSourceId ?? skipToken);
```
bvaughn, line 27: The concept of a `skipToken` makes sense, but I think these being hooks (and so needing to follow the rules of hooks) unfortunately makes the API a little awkward. If this data were coming from a Suspense cache, we could just conditionally read, e.g.
```js
const sourceText = selectedSourceId ? getSourceText(selectedSourceId) : null
```
This feels a little more ergonomic to me ^ but it's a small note.

packages/markerikson-stack-client-prototype/src/app/api.ts
bvaughn: Agreed. This bit feels a little awkward to me as well. Not a _huge_ concern though.

bvaughn (approved):
Couple of high-level thoughts!

Some things I really like about this PR:
1. The Loom tech walk through was very helpful, so thank you. I'd like to see us do more of this.
2. The demo itself looks really slick! and the code in the React components is pretty minimal which is nice.
3. Containing all of the backend/protocol code in a place that the UI doesn't directly call or interact with is a big win. (Side question: How would this work with testing? Could we stub this stuff out? Would be interesting to see some example unit tests for this approach.)
4. The cache lifetime control mechanism seems like a nice optimization option.

The biggest reservations I have about this is that the `useBlahQuery` hooks won't fit as nicely into React concepts like _suspense_ and _transitions_ as using Suspense+cache primitives would. I worry if we go this route, we may end up locking ourselves out of some really nice UI/UX polish that React could provide out of the box. Ultimately I think we could get the API that components interact with (the code we spend most of our time iterating on) to be roughly equivalent, either approach we take. The biggest things I'm thinking about then are:
1. How well do these hooks work with _suspense_ and _transition_ APIs?
2. How easy is it for us to write automated tests (including stubbing data).

packages/markerikson-stack-client-prototype/src/app/api.ts
```diff
@@ -0,0 +1,147 @@
...
+
+interface SourceGroups {
+  src: newSource[];
+  node_modules: newSource[];
+  other: newSource[];
+}
+
+const reIsJsSourceFile = /(js|ts)x?$/;
+
+const CACHE_DATA_PERMANENTLY = 10 * 365 * 24 * 60 * 60;
```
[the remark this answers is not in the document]
markerikson: Per https://redux-toolkit.js.org/rtk-query/api/createApi#keepunuseddatafor-1 , the value is in seconds, so I was going for "10 years".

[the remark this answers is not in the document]
ryota-murakami: @markerikson Oh, it's not `ms`, I'm idiots 😅

markerikson:
Answered some of the questions in today's meeting, but summarizing here:

- RTKQ doesn't have Suspense support yet, because A) we just haven't built it, and B) the current React team (Dan in particular) keeps telling everyone "it's alpha, it may change, don't use this yet": https://twitter.com/dan_abramov/status/1482020154187866125
- There _is_ a POC PR to add Suspense support to RTKQ, but the author is still experimenting trying to figure out potential API designs: https://github.com/reduxjs/redux-toolkit/pull/2245
- RTKQ testing in general is "mock the network layer using `msw` and then test connected components with a real store".   In this case, we can't use `msw`, because our protocol client is a wrapper over websockets, not plain HTTP. So, that's kind of an open question.
- There shouldn't be any "race condition" concerns with the cleanup behavior in that `getPause` endpoint. When the "subscription" reference count for a cache entry goes to 0, RTKQ starts a timer (60 seconds by default). If no other components resume asking for that data before the timer expires, RTKQ will evict the cache entry, and _that_ is when the `cacheEntryRemoved` promise will resolve.  In other words, toggling back and forth between two selected timestamps will just swap the reference counter and each time we'll still have the previous value in cache already.
- Per the "using RTKQ for non-REST data is awkward": I'd sorta disagree with this, because A) this is actually the reason why the `queryFn` option exists, and B) we'd still have to write those client SDK calls somewhere anyway (_and_ manage the loading state ourselves).

My biggest potential concerns with the overall approach here are:

- While the endpoints themselves are defined outside React, some of the orchestration here is happening in a component (in terms of "fetch A, then B, then C").
- We may need to fetch more related data, and I haven't tried doing that yet
- Testing and websockets
- We've talked about trying to put a lot of the backend interactions into a UI/app-agnostic "protocol" package with a a caching layer.  The RTKQ stuff would be firmly in the "app-specific logic" layer instead.  So, somewhat at odds with the idea of doing that.  (On the flip side, just because _this_ demo uses the `@replayio/protocol` package to do the fetching doesn't mean a final implementation has to - if the "protocol package" exposes async logic, we could certainly use that here instead.)

packages/markerikson-stack-client-prototype/src/app/api.ts
phryneas: You could also use an identity function or just skip the `baseQuery` here (although TS would not be happy) - the benefit of the `fakeBaseQuery` is that it allows you to specify a custom error type.

phryneas:
> The biggest reservations I have about this is that the `useBlahQuery` hooks won't fit as nicely into React concepts like _suspense_ and _transitions_ as using Suspense+cache primitives would. I worry if we go this route, we may end up locking ourselves out of some really nice UI/UX polish that React could provide out of the box. Ultimately I think we could get the API that components interact with (the code we spend most of our time iterating on) to be roughly equivalent, either approach we take. The biggest things I'm thinking about then are:
> 
>     1. How well do these hooks work with _suspense_ and _transition_ APIs?

As the React team doesn't want us to do suspense in a fetching library yet, (or for any fetching on the client so far), this is a bit of a question. We currently have a PR open with a suggested solution that I quite like: https://github.com/reduxjs/redux-toolkit/pull/2245

Essentially instead of

```js
const result1 = use1Query(args)
const result2 = use2Query(args)
```

you would do
```js
const [result1, result2] = useSuspendAll (
  use1Query(args),
  use2Query(args)
)
```
This way you could both trigger the fetching from component render (which would introduce a waterfall, but only on component level, not multiple times per component) - or you could start a query before render by doing a `dispatch(api.endpoints.1.invoke(args)` and then this would simply be the component to actually suspend.

How well this works with transition aps I can't tell because honestly to this point, a lot of the "transition" stuff is still quite arcane to me. But if there were a lot of benefit to gain I'm sure we'd figure a way out.

packages/markerikson-stack-client-prototype/src/app/api.ts
phryneas: You can use `queryFn` either way, `fakeBaseQuery` just throws an error if you accidentally use `query` instead of `queryFn` and allows you to specify an error type for TS.

bvaughn:
> sing a hook here instead of "just" a `cache.read()` has a significant benefit:

Interesting angle! I don't think I agree, but I'd be interested in talking about it more to better understand. In many cases, a Suspense cache can be backed by a `WeakMap` – which doesn't require any explicit cleanup in the client. If a `WeakMap` doesn't work, or if you want to keep _some_ values cached longer, an LRU (like `lru-cache`) can also make a decent alternative. Knowledge about the React tree, and which components need data, isn't really needed for these to work.

React has the concept of a _cache component_ which can be mounted at a certain location in the tree if you want to (a) mask an outer cache with an inner value (so you can invalidate one without the other) _or_ if you want to have a cache evicted when a subtree is unmounted. I don't think we need either of those capabilities in the Replay case though, since recording data is kind of "global" in that an app will only ever load one recording per session.

I think it's nice to get away from the "no unconditional calls" rule of hooks, so we don't have to execute code unnecessarily. A data-loading hook needs effects and state to work– and so we're still accruing some overhead even in the case where we have nothing to actually read/load. (Admittedly, the performance cost of this is probably minimal! I think the ergonomics of the API is nice too though– being able to conditionally call a `read` method and not having to pass a special token to a hook.)

markerikson:
FWIW, ever since I first saw a demo of a `MySuspenseCache.read(someId)` API, I've felt that it _ought_ to be possible to implement one of those with Redux.

I'm like 98% sure that RTKQ's `api.util.getRunningOperationPromise()` API is basically what we'd need to build one:

https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#getrunningoperationpromise

honestly I may try playing with that idea in the next phase of this prototype just to see what happens!

phryneas:
> Interesting angle! I don't think I agree, but I'd be interested in talking about it more to better understand. In many cases, a Suspense cache can be backed by a WeakMap – which doesn't require any explicit cleanup in the client. If a WeakMap doesn't work, or if you want to keep some values cached longer, an LRU (like lru-cache) can also make a decent alternative. Knowledge about the React tree, and which components need data, isn't really needed for these to work.

How would the WeakMap work though? You would need a stable reference to the argument - but it the argument is accumulated within the component - maybe from different props, maybe with a calculation in-between. That stable reference doesn't seem easily achievable.
Especially since you cannot even use `useMemo` or `useState` since on a suspend before a first successful mount, state & memoized values (apparently) get thrown out of the window, as the whole component seems to be discarded.

As for an LRU cache: what if I have a LRU of 100 and 101 components mounted that need to access that data? To be honest, it seems a bit like guesswork.

What we do in RTK Query is essentially reference counting - each component "subscribes" to a cache entry, so we always know which cache entries are used.

That knowledge is also important beyond simple cache collection, for cache invalidation. Imagine a REST api where we send a POST request to change user 5. Now, we would have to refetch all cache entries that contain related data. We might have quite a lot of those cache entries though - and half of them might not be used at all at the moment. So we only want to refetch those that are currently in use, while discarding the others (they were stale anyways and can be fetched again whenever they are actually needed).

bvaughn:
> How would the WeakMap work though? You would need a stable reference to the argument

Yes. In my experience, stable references are often possible. If they are not, then an LRU cache could be used instead (if you _need_ the functionality– maybe you don't!).

I get why RTK would use reference counting, although reference counting can have pitfalls in this scenario. For example, if a component renders (and reads from the cache) but never mounts– then there will be no explicit cleanup. To my knowledge, the best "work around" for this is to use a timer to check for things-that-rendered-but-never-mounted.

Sorry for the slow response but I have trouble keeping up with a long-form text discussion like this.

phryneas:
> I get why RTK would use reference counting, although reference counting can have pitfalls in this scenario. For example, if a component renders (and reads from the cache) but never mounts– then there will be no explicit cleanup. To my knowledge, the best "work around" for this is to use a timer to check for things-that-rendered-but-never-mounted.

The idea for this is right now that when using suspense, a component starts a fetch during render (assuming one is not running yet), but immediately unsubscribes from it. 
As cache entries have a lifetime of 60 seconds per default, that is plenty of time for the fetch to fulfill, the component to finally render and a `useEffect` to actually do a lasting subscription.

bvaughn:
It sounds like this conversation may have reached a good conclusion 😄 If RTK is using a timer based approach for working around the lack of an explicit GC signal, then it's probably doing all it can (at least for now).

I think reference counting is probably a heavier solution than is needed in many cases– (I've used `WeakMap` as a Suspense cache before with no troubles)– but I can imagine it works well in some cases.

Revisiting my initial concerns:
> 1. How well do these hooks work with suspense and transition APIs?
> 2. How easy is it for us to write automated tests (including stubbing data).

I think some example tests could help address the second one.

Mark pointed out above that it's not really fair to expect RTK to have an answer for the first question since the APIs are still unstable. That's fair! I have a mental model for how to tap into those APIs using a React Suspense cache and `useTransition` though, so I think my concern is also still valid when consideration other new approaches.

I think this was a good discussion 😄 Thanks for sharing your thoughts and perspective! Maybe we can talk more as the APIs become more stable.

PR #7160, opened 2022-06-14
Ensure scopes are loaded before rendering them (FE-239)

hbenl:


src/devtools/client/debugger/src/components/SecondaryPanes/Scopes.js
```diff
@@ -36,9 +36,17 @@ class Scopes extends PureComponent {
...
+    this.updateScopes(why, selectedFrame, frameScopes);
+  }
+
+  async updateScopes(why, selectedFrame, frameScopes) {
+    const scopes = getScopes(why, selectedFrame, frameScopes);
+    if (scopes) {
+      const scopesToLoad = scopes.filter(scope => scope.type === "value" && !scope.loaded);
+      await Promise.all(scopesToLoad.map(scope => scope.contents.load()));
+      this.setState({ scopes });
+    }
```
bvaughn, line 49: I think this is fine for now, but this (side effect during React's _render phase_) is the sort of thing that we should move to using a Suspense cache for in the future. I think a blocker to that though would be making `getScopes` and `getScope` return stable values rather than new ones each time.

bvaughn (approved):

src/devtools/client/debugger/src/components/SecondaryPanes/Scopes.js
hbenl: Yeah, the way that loading objects to be displayed in `ObjectInspector` works is really hard to follow and error-prone. For me the main hurdle to refactoring this stuff was the tree component that the `ObjectInspector` uses, it doesn't fit the `ObjectInspector` requirements very well.

bvaughn: Makes sense. Hopefully we'll re-implement both the data loading and object inspector parts in the next few months.

PR #7174, opened 2022-06-14
Test suites polish

jasonLaster:
* Update selected color
* add test row padding
* hide unknown branch
* Selecting a replay will always go to a replay
* Hide the header left view when viewing runs/tests
* show runs/tests in filter dropdown

Screen Shot 2022-06-14 at 4 30 16 PM


jaril:
Continued in #7180.

PR #7175, opened 2022-06-14
Add rainbow picker to console context menu

bvaughn:
Things to sort out:
- [ ] Using `key` to reset derived state of `PrefixBadgePicker` picker fixed out-of-sync issue but causes the picker opening interaction to be awkward. More info [on this Loom](https://discord.com/channels/@me/974438142125953034/986364129797349378).

### DIFF
--- a/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
+++ b/src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
@@ -2,8 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import { Dispatch } from "@reduxjs/toolkit";
-import { SourceLocation } from "devtools/client/debugger/src/reducers/types";
+import classNames from "classnames";
+import { badges } from "components/PrefixBadgePicker/PrefixBadgePicker";
+import { PrefixBadge, SourceLocation } from "devtools/client/debugger/src/reducers/types";
 import { MessageContainer } from "devtools/client/webconsole/components/Output/MessageContainer";
 import { StateContext } from "devtools/client/webconsole/components/Search";
 import constants from "devtools/client/webconsole/constants";
@@ -17,7 +18,7 @@ import {
   setFocusRegionBeginTime,
 } from "ui/actions/timeline";
 import { ContextMenu } from "ui/components/ContextMenu";
-import { Dropdown, DropdownItem } from "ui/components/Library/LibraryDropdown";
+import { Dropdown, DropdownDivider, DropdownItem } from "ui/components/Library/LibraryDropdown";
 import Icon from "ui/components/shared/Icon";
 import { selectors } from "ui/reducers";
 import {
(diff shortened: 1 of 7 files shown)


src/ui/components/PrefixBadge.tsx
```diff
@@ -20,6 +20,7 @@ export default function PrefixBadgeButton({ breakpoint }: { breakpoint: Breakpoi
 
   return (
     <PrefixBadgePicker
+      key={breakpoint.options.prefixBadge}
```
bvaughn, line 23: Without this, external changes to the badge color (aka ones made from the context menu) won't be rendered and the two will seem out of sync.

Note that this change currently _breaks_ the click-to-open interaction for this picker because it had a pre-existing issue where clicking unexpectedly _cleared_ the selected badge (which now means a new `key` and resetting its is-open internal state).

src/ui/components/Library/LibraryDropdown.tsx
```diff
@@ -35,7 +35,7 @@ export function Dropdown({
 }: {
   children: React.ReactNode;
   menuItemsClassName?: string;
-  widthClass?: "w-56" | "w-64" | "w-80";
+  widthClass?: "" | "w-56" | "w-64" | "w-80";
```
bvaughn, line 38: Not sure why we wanted bigger-than-necessary menus but the mockups for this context menu seem to want it to tight-fit-content so...

src/ui/components/PrefixBadge.tsx
souporserious: Makes sense, key is a nice escape hatch here! 😅 I didn't look too much into how we could use derived state so we're not relying on local badge state which I carried over from the previous implementation.

bvaughn: IMO there are some valid use cases for derived state, but [many times it's not necessary](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html). I _think_ this is one of those times (not necessary) but I didn't want to assume too much before talking to you about why it was the way it was.

packages/components/PrefixBadgePicker/Picker.tsx
```diff
@@ -97,7 +97,7 @@ export function Picker<Values extends any>({
             transitioning.current = true;
           }
 
-          const nextActiveId = isToggle ? undefined : isChildActive ? undefined : child.props.id;
+          const nextActiveId = isToggle ? undefined : child.props.id;
```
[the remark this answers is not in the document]
jcmorrow: *and* gets rid of a nested ternary so bonus points.

souporserious (approved):
I think the key change is now blowing away the out animation, but something we can look into in a follow-up PR. Otherwise, LGTM! :shipit:

jcmorrow (approved):
Left one tiny comment, LGTM 🌈

src/devtools/client/webconsole/components/Output/ConsoleOutput.tsx
```diff
@@ -464,3 +505,5 @@ function TrimmedMessageCountRow({ position }: { position: "before" | "after" })
     );
   }
 }
+
+function noop() {}
```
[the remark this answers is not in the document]
bvaughn: Fair!

We could add an `emptyFunction` export somewhere, and update all of the places that e.g. default to empty function prop values (e.g. `onCreate: () => {}`) to use it too.

Let's do that in a focused follow up PR though?

jaril:
@ceceliacreates user facing change 🙋‍♂️

PR #7176, opened 2022-06-14
Improve error handling for auth failures from the Replay browser

ryanjduffy:
## Issue

If authentication fails for any reason from the Replay browser, the user ends up on a generic browser error screen with a 500 error

## Resolution

* Add some error context from the auth0 side for at least the unexpect IdP scenario (the most common case right now)
* Show a error dialog with a sign out button so the user can try something to fix the auth problem


jcmorrow (approved):
I don't have a perfect understanding of this code, but I also don't see anything that is obviously wrong.

ryanjduffy:
@jcmorrow - found a semi-hacky way to logout the user from the app and auth0 so they can try again on auth failure. Thanks for pushing on it!

PR #6860, opened 2022-05-19
[DRAFT] Try adding a "line heatmaps" feature to our editor

markerikson:
This is an initial attempt to add a "line heatmaps" feature per #6856 .

So far I've figured out how to add another gutter to CodeMirror and insert text on each line's gutter.

However, I'm stuck in that our line hover tooltips are showing up _behind_ the new gutter, and I can't immediately figure out how to get them to pop out over top.

Rough appearance so far:

image

If I can get the hover tooltips sorted out, I then ought to be able to loop over each line, create a `SourceLocation` entry, query the Redux store for any hit counts on that line, do some data munging, and display the hits on each line.

PR #7179, opened 2022-06-14
Show inline hit counts in editor

bvaughn:
Loom walk through: https://www.loom.com/share/d53059cc0b754727a0755fa456fe9a5b

Screen Shot 2022-06-15 at 10 49 20 AM

Screen Shot 2022-06-15 at 10 49 27 AM

Builds on top of #6860


jonbell-lot23:
Hello! Items I'm interested to explore:

* What's the default state that all users see by default?
* What's the entry point for turning this on? (you're using the footer, we could think about other places)
* Should it be set globally or per Replay?

Here's my thinking:

* I'd love to show a color representation of heat (such as shades of blue for example) by default
* I wonder if we could just click the blue gutter to toggle text/no text
* Personally, I think this should be global. Mark mentioned wanting this on at all times, and I'd be the same. It would be a bummer to have to re-deploy this every single time you try to use Replay. So I'd make it global.

Open to thoughts! This is just some thinking to kick things off.

bvaughn:
@jonbell-lot23 

> What's the default state that all users see by default?

The feature is _off_ by default.

> What's the entry point for turning this on? (you're using the footer, we could think about other places)

The little toggle switch in the footer.

> Should it be set globally or per Replay?

It is currently _global_.

> I'd love to show a color representation of heat (such as shades of blue for example) by default
> I wonder if we could just click the blue gutter to toggle text/no text

This might be something that would be nice to poll some engineers about. Both myself and Mark feel that the _actual hit count numbers_ are important information, and a color-only heat map would be less helpful. (After all, that's kind of what we already have with the light gray vs dark gray line numbers:

Screen Shot 2022-06-14 at 7 50 18 PM

I'm also a bit wary of overly complicating the UX this by letting users toggle it off, on, and on-with-numbers. I think we should pick the default (of what "on" means) and then give users a simple on/off switch.

I wonder how we can gather feedback on this in a lightweight way? We _could_ build both versions and allow you to turn them on off via the console I guess.

src/devtools/client/debugger/src/components/Editor/StaticTooltip.tsx
```diff
@@ -11,12 +11,14 @@ type StaticTooltipProps = {
...
+  const { value: inlineHitCounts } = useFeature("inlineHitCounts");
 
   return ReactDOM.createPortal(
     <div
       className={classNames(
-        "absolute z-50 flex flex-row space-x-px transform -right-1 translate-x-full bottom-4 mb-0.5 pointer-events-none",
-        enableLargeText && "bottom-6"
+        "pointer-events-none absolute bottom-4 z-50 mb-0.5 flex translate-x-full transform flex-row space-x-px",
+        enableLargeText && "bottom-6",
+        inlineHitCounts ? "-right-9" : "-right-1"
```
bvaughn, line 21: This is gross 😄 but it was already gross and kind of out of scope of this PR so 🤷🏼

jonbell-lot23:
@bvaughn ah, I'm asking at a higher level like "what does success look like" not "what does your current PR do."

So specifically I'm not asking where you're turning it on and off, or how you expect the defaults to work. I am doing design thinking to describe how I think it could work best.

>Both myself and Mark feel that the actual hit count numbers are important information, and a color-only heat map would be less helpful.

Color only is not on the table, I'm thinking through the default state. I am proposing a more useful default state because right now it's binary. The heatmap gives a scannable hint ("oh there's more going on here than I thought, I should investigate") while also providing an affordance that someone could click on to discover your feature. So this thinking attempts to pull the feature forward even more, by emphasising the default state.

Having a little switch on the footer is fine, but there are situations where we use that space, and we need to make sure they're not colliding. I also think the colors are helpful ambient information when first looking into a file.

bvaughn:
@jonbell-lot23

Fair. I thought you were asking for clarity about how it was currently implemented.

> Having a little switch on the footer is fine, but there are situations where we use that space, and we need to make sure they're not colliding.

Could you elaborate? We sometimes have a source map toggle there, and a link. I imagine we need to be able to overflow there though regardless.

jonbell-lot23:
> Could you elaborate? We sometimes have a source map toggle there, and a link. I imagine we need to be able to overflow there though regardless.

Yeah, the link is the main thing I want to make sure looks right. I vaguely remember other things going down there but I'm probably misremembering.

> RE: the gradient – what would the thresholds be? and the colors?
> 
> Would the thresholds be static (e.g. 1-5 hits, 6-25 hits, etc) or would they be relative (gradient from cool to hot based on the lowest to highest hit count for the current file)?

I'm sure reasonable people can disagree on this, and I don't feel strongly, but I'd personally make it relative. 

--

Process-wise, here's what I think we should do:

1. Once we feel good about this (including link alignment), we ship behind a flag
2. We get more data from the team about what we think about the gradient idea
3. Based on that feedback, we turn it on for everyone

That way we don't block your PR and we can get some more ideas from the team. These comments were just meant as ideas, I hope they're not coming off as proscriptive mandates from the jerky design team!

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -0,0 +1,84 @@
...
+        const className = hitCount > 0 ? styles.HitsBadge : styles.NoHitsBadge;
+        const innerHTML = hitCount > 0 ? `${hitCount}x` : "&nbsp;";
+        const title = hitCount > 0 ? `${hitCount} hits` : "";
+
+        const markerNode = document.createElement("div");
+        markerNode.className = className;
+        markerNode.innerHTML = innerHTML;
+        markerNode.title = title;
+
+        doc.setGutterMarker(lineHandle, "hit-markers", markerNode);
```
jasonLaster, line 56: Nice

jasonLaster (commented):
Code looks good! 

The more interesting question is around visuals + defaults. 

1. I like that hit counts are off by default. They're nice to have when you want them, but are an opt-in experience.
2. I don't think hit counts has earned a prominent position in the footer yet. We can add it to three other spots (editor context menu, command palette, and settings > appearances). We see folks check appearances in LogRocket so it would be discoverable
3. I'm curious to see what Jon comes up with on the visual front. For me, the key is that it remains in the background (soft), but is available for quickly scanning. 
4. I would like to try and improve our empty lines component to also cover not-hit lines as well
5. It would be nice to explore color coding our gutter based on hits as well. Not as a replacement, but on top of

https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=12336%3A94645

bvaughn:
Ready for review

bvaughn:
> 2. I don't think hit counts has earned a prominent position in the footer yet, but we can add it to three other spots (editor context menu, command palette, and settings > appearances). We see folks check appearances in LogRocket so it would be discoverable

I removed the footer and added it to the Appearance settings instead. Adding a _global_ setting to the context menu felt wrong, because those options are per-file. Adding it to he command palette felt odd because of the way that code is imperatively structured. (The label would have to say "Toggle line hit counts" rather than show/hide.)

> 4. I would like to try and improve our empty lines component to also cover not-hit lines as well

I don't think I understand what you're suggesting here so I left it for now.

markerikson (approved):
Love it. :ship: :ship: :shipit:

PR #7185, opened 2022-06-15
Adjust hit count badge

jasonLaster:
Screen Shot 2022-06-15 at 12 50 48 PM


src/ui/components/NetworkMonitor/content.ts
```diff
@@ -59,6 +59,11 @@ export const shouldTryAndTurnIntoText = (contentType: string): boolean => {
   if (contentType.startsWith("text")) {
     return true;
   }
+
+  if (contentType == "unknown") {
+    return true;
```
bvaughn, line 64: What is this doing?

PR #7188, opened 2022-06-16
Reload hitCounts when they get cleared due to focus

jcmorrow:
When the focus region changes, we clear out the hitCounts because they
are no longer reliable. However, the LineHitCounts component was not
refetching them when they got cleared, so they would not update as focus
changed. This seems like the kind of thing which might be addressed well
with one of the new architecture proposals? Until then, this is a slight
modification to the existing hackiness.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
@@ -27,8 +27,10 @@ function LineHitCounts({ cm }: Props) {
 
   useLayoutEffect(() => {
     // HACK Make sure we load hit count metadata; normally this is done in response to a mouseover event.
-    dispatch(setBreakpointHitCounts(sourceId, 0));
-  }, [dispatch, sourceId]);
+    if (hitCounts === null) {
+      dispatch(setBreakpointHitCounts(sourceId, 0));
+    }
+  }, [dispatch, sourceId, hitCounts]);
 
   useLayoutEffect(() => {
     if (!cm) {


src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -27,8 +27,10 @@ function LineHitCounts({ cm }: Props) {
 
   useLayoutEffect(() => {
     // HACK Make sure we load hit count metadata; normally this is done in response to a mouseover event.
-    dispatch(setBreakpointHitCounts(sourceId, 0));
-  }, [dispatch, sourceId]);
+    if (hitCounts === null) {
+      dispatch(setBreakpointHitCounts(sourceId, 0));
+    }
+  }, [dispatch, sourceId, hitCounts]);
```
bvaughn, line 33: I am confused about why this change would do what you said in chat (refresh hit counts when the focus region changes).

If `hitCounts` has already changed, then what needs to be re-fetched? I thought the problem was stale `hitCounts` values?

bvaughn (approved):
Accepting because I trust you, but I don't understand the fix.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
jcmorrow: So, backing up for a second so this change will make more sense. The way this works right now (it's a little rou goldberg esque) is that the `focusRegion` changes, then the reducers that holds `hitCounts` sees that and clears all values. Then, the *next* time a user hovers a line, the `hitCounts` get refetched because they are currently empty but something has requested them.

When you made this component, you correctly figured out that for this to work something has to trigger the initial query because otherwise `hitCounts` will remain `null`. But now, when the `focusRegion` changes the hitCounts get cleared still, but nothing does the re-requesting (because neither dispatch nor sourceId has changed) so right now line 31 will only run on first mount). This change basically says:

> Rather than only doing this hacky load on first mount, do it *any time* the `hitCounts` value changes to `null`.

I know it's convoluted, I wish it were less so. But in that absence of some less convoluted loading scheme, does that explanation make sense?

bvaughn: > The way that this refetching works in `LineNumberToolTip` right now... is that the `focusRegion` changes, then the reducers that holds `hitCounts` sees that and clears all values. Then, the *next* time a user hovers a line, the `hitCounts` get refetched because they are currently empty but something has requested them.

Ah, I wasn't aware of that bit!

That _is_ a bit convoluted! Thanks for explaining 😁

PR #7190, opened 2022-06-16
Send protocol timings to MixPanel

jcmorrow:
I know we have a lot of this timing in honeycomb, but it's nice to be able to tie it in with MixPanel stuff as well.

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -29,7 +29,7 @@ import type { ExpectedError, UnexpectedError } from "ui/state/app";
 import { Recording } from "ui/types";
 import { extractGraphQLError } from "ui/utils/apolloClient";
 import LogRocket from "ui/utils/logrocket";
-import { endMixpanelSession } from "ui/utils/mixpanel";
+import { endMixpanelSession, timeMixpanelEvent, trackMixpanelEvent } from "ui/utils/mixpanel";
 import { features, prefs } from "ui/utils/prefs";
 import { registerRecording, trackEvent } from "ui/utils/telemetry";
 import tokenManager from "ui/utils/tokenManager";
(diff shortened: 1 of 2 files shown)


jcmorrow:
> BTW Mixpanel does use session_start to indicate a session starting for our case.
> 
> 
> 
> image
> 
> 
> 
> This was done because we used to track each discrete action (login > open source file > hover on line > add breakpoint). And the default session start setting (which is what you're describing) was getting in the way of having a clear picture of that flow. That was ~6mos ago and is probably worth revisiting now.

Fair. What I was seeing is a lot of "Session start" followed by events called "session_start". Maybe something has changed in what MixPanel is expecting?

jaril:
> > BTW Mixpanel does use session_start to indicate a session starting for our case.
> > image
> > This was done because we used to track each discrete action (login > open source file > hover on line > add breakpoint). And the default session start setting (which is what you're describing) was getting in the way of having a clear picture of that flow. That was ~6mos ago and is probably worth revisiting now.
> 
> Fair. What I was seeing is a lot of "Session start" followed by events called "session_start". Maybe something has changed in what MixPanel is expecting?

Let's revisit when you're NOT in PTO :P

jaril:
> So yeah, this is what I see when I look at `Flows`:
> 
> CleanShot 2022-06-28 at 15 29 31@2x
> 
> It's a little weird that we are sending 2nd, 3rd, and 4th `session_start` events, but it's less problematic given that we can filter them out.

Ah yes. Last time I ran into this I didn't get to dig as far as I would've liked to. If you happen to figure out why, let me know :)

jcmorrow:
I don't think we need this now that we 🐝 (Honeycomb) all the things.

PR #7192, opened 2022-06-16
Try using Yarn 3 as a package manager

markerikson:
This PR:

- Switches package installation from NPM 8 to Yarn 3
- Switches the GH Actions setup to use Yarn, with `actions/cache@v3` to cache `node_modules` and `./yarn/cache`
- Updates all references of `npm run` that I could find to just be `yarn`, since Yarn will look for run script names
- Updates the Storybook config to find `@recordreplay/accordion` correctly and compile the `protocol` package
- Updates the Next config to compile the `protocol` package.  (This _may_ not be necessary now - it was definitely failing to compile the TS files early in the branch)
- Adds some missing peer deps like `@tiptap/core`

Compare with the PNPM PR at #7194 .

It may be worth extracting the shared behavior into its own action at some point to remove the copy-pasta setup.


markerikson:
Still a bit of E2E test flakiness, but I think the primary setup here is working.

To try this out:

```
git fetch
git checkout feature/FE-242-yarn-workspaces
yarn install
yarn dev
```

souporserious:
Tried this out and all seems to be working well for me! 😎

jasonLaster (approved):

PR #7193, opened 2022-06-16
Test suite polish

jaril:
Fixes FE-267.


jasonLaster:
* Summary should show the correct count
* Right sidebar summary should stack vertically
* format test-path in filter bar

Jon follow up designs
* the selected row should be white instead of grey
* the pass fail badge should look better (green on blue is washed out)
* long branch names should be truncated
* UI should be more mobile responsive 
* consider dropping less important values like pull_request (trigger)



Screen Shot 2022-06-16 at 2 39 05 PM

PR #7199, opened 2022-06-17
Decode filter

jasonLaster:
We'll probably quite a bit on the frontend query parsing. Today, the goal here is to support values that start with `'`, `"`, and `[`

https://www.loom.com/share/0e0a06563cdb476188f8fa856d06f7cf

### Before

Screen Shot 2022-06-16 at 8 50 07 PM

### After

Screen Shot 2022-06-16 at 8 48 11 PM

### DIFF
--- a/src/ui/components/Library/useFilters.ts
+++ b/src/ui/components/Library/useFilters.ts
@@ -25,6 +25,60 @@ export const LibraryContext = createContext<LibraryContextType>({
   preview: null,
 });
 
+type Surrounder = {
+  [open: string]: string;
+};
+
+const SURROUNDERS: Surrounder = {
+  "'": "'",
+  '"': '"',
+  "[": "]",
+};
+
+const peekUntil = (input: string, until: string): string | void => {
+  return input.match(new RegExp(`^(.*?${until})`))?.[0];
+};
+
+const peekNextWord = (input: string): string => {
+  return input.match(new RegExp(`^(\S+)`))?.[0] ?? "";
+};
+
+export function encodeFilter(str: string) {
+  let encodedStr = "";
+  while (str.length > 0) {
+    const [match] = str.match(/^(.*?:)/) || [];
+    if (match) {
+      encodedStr += match;
+      str = str.substring(match.length);
+
+      const nextChar = str.substring(0, 1);
+      if (Object.keys(SURROUNDERS).includes(nextChar)) {
+        const closingChar = SURROUNDERS[nextChar];
+        const valueMatch = peekUntil(str.substring(1), closingChar);
+
+        if (valueMatch) {
+          encodedStr += encodeURIComponent(`${nextChar}${valueMatch}`);
+          str = str.substring(valueMatch.length + 1);
+        } else {
+          encodedStr += str.substring(1);
+          str = "";
+        }
+      } else {
+        const nextWord = peekNextWord(str);
+        // const [nextWord] = str.match(/^(\w+)/) || [];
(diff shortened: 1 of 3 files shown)


src/ui/components/Library/useFilters.ts
```diff
@@ -25,6 +25,40 @@ export const LibraryContext = createContext<LibraryContextType>({
...
+      } else {
+        const [nextWord] = str.match(/^(\w+)/) || [];
+        encodedStr += nextWord;
+        str = str.substring(nextWord.length);
+      }
+    } else {
+      encodedStr += str;
+      str = "";
+    }
+  }
```
jasonLaster, line 57: this logic is pretty silly, but i dont mind writing parsers like this when it is backed by tests and self contained.

src/ui/components/Library/useFilters.ts
```diff
@@ -25,6 +25,40 @@ export const LibraryContext = createContext<LibraryContextType>({
...
+  let encodedStr = "";
+  while (str.length > 0) {
+    const [match] = str.match(/^(.*?:)/) || [];
+    if (match) {
+      encodedStr += match;
+      str = str.substring(match.length);
+
+      const nextChar = str.substring(0, 1);
+      if (["'", '"', "["].includes(nextChar)) {
+        const closingChar = nextChar == "[" ? "]" : nextChar;
```
jcmorrow, line 38: This is cute, but I think personally I would rather have an explicit object like

```
type Surrounder = {
  open: string;
  close: string;
};
const SURROUNDERS: Surrounder[] = [
  { open: "'", close: "'" },
  { open: '"', close: '"' },
  { open: "[", close: "]" },
];
```

src/ui/components/Library/useFilters.ts
```diff
@@ -25,6 +25,40 @@ export const LibraryContext = createContext<LibraryContextType>({
...
+        const [valueMatch] = str.substring(1).match(new RegExp(`^(.*?${closingChar})`)) || [];
```
jcmorrow, line 39: It's common to do this a lot in parser, but read regex sucks. What about writing a small `peekUntil` function, which takes string to peek and a character to peek until?

```
const peekUntil = (input: string, until: string): RegExpMatchArray | null => {
  return input.match(new RegExp(`^(.*?${until})`));
};
```

jcmorrow: Could be used on line 31 as well.

src/ui/components/Library/useFilters.ts
```diff
@@ -25,6 +25,40 @@ export const LibraryContext = createContext<LibraryContextType>({
...
+
+        if (valueMatch) {
+          encodedStr += encodeURIComponent(`${nextChar}${valueMatch}`);
+          str = str.substring(valueMatch.length + 1);
+        } else {
+          encodedStr += str.substring(1);
+          str = "";
+        }
+      } else {
+        const [nextWord] = str.match(/^(\w+)/) || [];
```
jcmorrow, line 49: are you sure you want a word boundary here? For instance, is `this-token` a valid next word? What about `a_wild_🦄_appears`?

I would be tempted to instead add a `BOUNDARY_TOKENS` const for whatever you consider the things that delimit a "word" and then do a `peekNextWord` similar to `peekUntil` above.

jcmorrow (commented):
I freakin love parsers.

jasonLaster:
Love this feedback. I added a bunch of these, although this would still qualify as experimental :)

PR #7202, opened 2022-06-17
Fixed potentially undefined value bug

bvaughn:
For consideration.

### DIFF
--- a/src/ui/components/Timeline/UnloadedRegions.tsx
+++ b/src/ui/components/Timeline/UnloadedRegions.tsx
@@ -31,9 +31,12 @@ export const UnloadedRegions: FC = () => {
   let beginTime = begin.time;
   let endTime = end.time;
   if (focusRegion) {
-    const focusedAndLoaded = overlap([rangeForFocusRegion(focusRegion)], loadedRegions.loading)[0];
-    beginTime = focusedAndLoaded.begin.time;
-    endTime = focusedAndLoaded.end.time;
+    const overlappingRegions = overlap([rangeForFocusRegion(focusRegion)], loadedRegions.loading);
+    if (overlappingRegions.length > 0) {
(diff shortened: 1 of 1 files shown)


src/ui/components/Timeline/UnloadedRegions.tsx
```diff
@@ -31,9 +31,12 @@ export const UnloadedRegions: FC = () => {
   let beginTime = begin.time;
   let endTime = end.time;
   if (focusRegion) {
-    const focusedAndLoaded = overlap([rangeForFocusRegion(focusRegion)], loadedRegions.loading)[0];
-    beginTime = focusedAndLoaded.begin.time;
-    endTime = focusedAndLoaded.end.time;
+    const overlappingRegions = overlap([rangeForFocusRegion(focusRegion)], loadedRegions.loading);
+    if (overlappingRegions.length > 0) {
```
bvaughn, line 35: I _think_ that– logically– only potential lengths we can have here are 0 and 1, but since the `overlap` method accepts two arrays and returns an array, it feels a little odd to hard-code that 😄

jcmorrow: I'm curious what case we would be seeing *no* overlap between our focus window and the loading regions? That actually seems like the root cause (and a really bad bug that would break everything).

jcmorrow: Oh I see you mentioned an example in https://github.com/replayio/devtools/pull/7107#discussion_r900112955. I'm going to look at that, because if we are "focused" on a completely unloaded region then something is super broken.

jcmorrow: Ah, it's because this uses the focus window as it is getting dragged, during which we have not yet sent `loadRegion` commands. OK, then this is a reasonable change I think.

PR #7204, opened 2022-06-17
add colors to line count heat map

souporserious:
Quick one hour spike on adding colors to line count heat maps because I want this in time for our videos 😛 

## Screenshot

image

## Future Considerations

- Not sure how in CodeMirror to set the gutter width to be smaller for hit counts, @bvaughn any ideas here?
- @jonbell-lot23 would love your thoughts here for what colors we should go with.
- Need to add the collapsed default version without numbers


src/devtools/client/debugger/src/components/Editor/Editor.css
```diff
@@ -55,6 +55,11 @@ html[dir="rtl"] .editor-mount {
   color: var(--grey-50);
 }
 
+/** Pull the breakpoint badge above the hit count badge **/
+.CodeMirror-linenumber {
+  z-index: 5;
+}
```
bvaughn, line 61: Nice 👍🏼

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -22,8 +24,21 @@ function LineHitCounts({ cm }: Props) {
...
+  const hitCountColorMap = useMemo(
+    () =>
+      new Map(
+        uniqueHitCounts.map(hitCount => [
+          hitCount,
+          Math.min(1, Math.max(0, Math.log(hitCount) / Math.log(maxHitCount))) || 0,
+        ])
+      ),
+    [uniqueHitCounts, maxHitCount]
+  );
```
bvaughn, line 41: This memoization looks broken. We're always re-creating `uniqueHitCounts` so `useMemo` will always be bypassed. (Need to move the `Set` creation into the memo function too.)

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -22,8 +24,21 @@ function LineHitCounts({ cm }: Props) {
...
+  const uniqueHitCounts = Array.from(
+    new Set(hitCounts ? hitCounts.map(hitCount => hitCount.hits) : [])
+  );
+  const maxHitCount = Math.max(...uniqueHitCounts);
+  const hitCountColorMap = useMemo(
+    () =>
+      new Map(
+        uniqueHitCounts.map(hitCount => [
+          hitCount,
+          Math.min(1, Math.max(0, Math.log(hitCount) / Math.log(maxHitCount))) || 0,
```
bvaughn, line 37: Totally no big deal! But I am a little curious for your thinking about log vs cube root here! In my experience, log maybe over-emphasizes lower amounts _by a bit_.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
souporserious: Oops, yeah, I'll fix this up!

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
souporserious: Yeah, I'm not sure, here is an example of both:

Log            |  Cube Root
:-------------------------:|:-------------------------:
image | image

I probably don't have enough space between the color lightness for the cube root example which is making it hard to discern any difference. Seeing it now, it doesn't really make sense as a monochromatic scheme either and seems worth interpolating between different hues.

Is it apparent the color is mapped to each source file and not the whole project? Like when moving between files and a 4 count hit will having varying color, is that disruptive? Sounds like something we could test amongst ourselves and see how it feels.

souporserious: Yeah, looks to be a little better using different hues with cube root:

image

bvaughn: I think the advantage of cube root won't be obvious until you're looking at a file with hundreds/thousands of line hits. The difference between 1 and 4 isn't big, so maybe it's actually good that we don't style it super differently? That being said, I'm not sure I understand the difference I'm seeing above. I'd still expect the lowest value to be the same color?

> Is it apparent the color is mapped to each source file and not the whole project? Like when moving between files and a 4 count hit will having varying color, is that disruptive? Sounds like something we could test amongst ourselves and see how it feels.

This is a good question. I think we'll have to try and see how it feels.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
```diff
@@ -4,12 +4,11 @@
   justify-content: center;
 
   width: 100%;
```
bvaughn, line 6: Wil

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
```diff
@@ -3,17 +3,15 @@
   align-items: center;
   justify-content: center;
 
-  width: 26px;
-  height: var(--theme-code-font-size);
-  margin: 2px;
-  padding-top: 1px;
+  width: 100%;
+  height: 16px;
```
bvaughn, line 7: This was using a CSS var before because we have a "large text size" experimental setting. I don't think a numeric value like 16px is going to work for that.

src/devtools/client/debugger/src/components/Editor/LineHitCountsToggle.tsx
```diff
@@ -0,0 +1,21 @@
...
+function LineHitCountsToggle() {
+  const { value } = useFeature("hitCounts");
+
+  if (!value) {
+    return null;
+  }
+
+  return (
+    <label className="flex items-center space-x-1 pt-0.5 pl-3">
+      <Toggle enabled={true} setEnabled={() => {}} />
```
souporserious, line 15: TODO: I need to add this state to the Redux store.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
souporserious: Oops, forgot about that, I'll get this added back!

jasonLaster:
@souporserious have you thought about hover states? i.e. when I hover over a line?

src/devtools/client/debugger/src/components/Editor/LineHitCountsToggle.tsx
bvaughn: We re-visiting the toggle?

bvaughn: I had it at first but we decided to remove it.

Looks like this maybe shouldn't have been committed? since it's not actually being used?

bvaughn:
Did you mean to merge this? :)

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
```diff
@@ -3,19 +3,26 @@
   align-items: center;
   justify-content: center;
 
-  width: 26px;
-  height: var(--theme-code-font-size);
-  margin: 2px;
-  padding-top: 1px;
+  width: 100%;
+  height: calc(var(--theme-code-font-size) * 1.4);
```
bvaughn, line 7: Ooo, why 1.4? That seems...kind of random 😁

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
```diff
@@ -3,19 +3,26 @@
...
+  position: relative;
+  top: -0.25px;
```
bvaughn, line 9: What's this? 😅

src/devtools/client/debugger/src/components/Editor/LineHitCountsToggle.tsx
bvaughn: Ooops looks like this was already removed and I was commenting on a stale thing. My bad!

souporserious:
Jason merged yesterday when we were on a call 😄

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
souporserious: Yeah, I should have put a comment here. It's sort of random, it's to remove the borders between each line background. Hopefully, we can look into a better way to handle this through CodeMirror.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
souporserious: Same as above, it's making sure the background covers a larger height so there isn't any space between the markers. We then need to pull things back up a little to accommodate for the larger height 😅

PR #7205, opened 2022-06-17
Proof of concept: create an RTKQ-based React Suspense cache for the sources list

markerikson:
This PR:

- Adds a **HIGHLY EXPERIMENTAL POC** (as in "look I tried to understand Brian's code and hacked on this for a couple hours and it amazingly seems to actually do something useful 🤷‍♂️ ") React Suspense cache built on top of RTK Query's `getRunningOperationPromise` API
- Updates the `<SourcesList>` component to read from that cache, and wraps it in a `<Suspense>` + loader (with an artificial delay added to the RTKQ API endpoint just to see that it's showing the fallback)
- Turns on `strict: true` in the `tsconfig.json` for both prototype folders, because **APPARENTLY WE DIDN'T HAVE TS STRICT MODE ON AND THIS WAS BREAKING THINGS ARGHGHHHHGGH!**

### DIFF
--- a/packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
+++ b/packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
@@ -4,7 +4,7 @@ import type { AppStore } from "../../app/store";
 import { api, SourceGroups } from "../../app/api";
 
 /*
- * This entire file is ENTIRELY WIP PROOF OF CONCEPT!
+ * This entire file is _mostly_ WIP PROOF OF CONCEPT!
 
   I've been staring at Brian's `CommentsCache` and `MessagesCache`
   trying to understand the data flow, and reverse-engineer how to 
@@ -14,7 +14,8 @@ import { api, SourceGroups } from "../../app/api";
 
   - The `unstable_createCacheForType` API, which seems to accept a factory function
     that holds the wrapper for a piece of data (such as a `new Map()`, or an object
-    with some field that has the actual value)
+    with some field that has the actual value).  Note: per Brian, that's not needed 
+    for this cache implementation, since we don't need to invalidate the data later.
   - React's undocumented but semi-known "throw a promise" mechanism, where we throw
     a consistent promise reference if the data isn't yet available.  
     (Note: the promise throwing behavior may still be subject to change? per Dan, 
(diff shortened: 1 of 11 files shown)


bvaughn:
About to review this change set _but_ this change would be nice to extract into its own PR:

* Turns on strict: true in the tsconfig.json for both prototype folders, because APPARENTLY WE DIDN'T HAVE TS STRICT MODE ON AND THIS WAS BREAKING THINGS ARGHGHHHHGGH!

markerikson:
@bvaughn sure, filed #7236 with the `tsconfig` change and I can rebase this after that's in

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
```diff
@@ -0,0 +1,118 @@
...
+
+function createSourceGroupsRecord(): SourceGroupsResult {
+  return {
+    groups: null,
+  };
+}
+
+export function getSourceGroups(): SourceGroups {
+  // Create or read the cache for this section of the UI (????)
+  const sourceGroupsRecord = getCacheForType(createSourceGroupsRecord);
```
bvaughn, line 59: Since there's only one set of sources– one thing we need to cache per recording– and only one recording per app session– we don't really need to use `getCacheForType`. We could literally just use a module-level variable to track our state.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // data is the last known good request result we have tracked - or if none has been tracked yet the last good result for the current args
+  let data = sourcesEntry!.data;
+
+  const hasData = data !== undefined;
+
+  // isFetching = true any time a request is in flight
+  const isFetching = sourcesEntry!.isLoading;
+  // isLoading = true only when loading while no data is present yet (initial load with no data in the cache)
+  const isLoading = !hasData && isFetching;
```
bvaughn, line 91: It's not clear to me why `sourcesEntry!.isLoading` isn't sufficient for this. RTK should reset that field to `false` once it has data, no?

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
```diff
@@ -0,0 +1,118 @@
...
+  // isSuccess = true when data is present
+  const isSuccess = sourcesEntry!.isSuccess || (isFetching && hasData);
```
bvaughn, line 93: Wait, `isFetching` and `hasData` coexist? I'm confused. Doesn't data normally come after fetching completes?

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  if (isFetching || sourcesEntry!.isUninitialized) {
+    console.log("Getting API promise and throwing");
+    const promise = api.util.getRunningOperationPromise("getSources", undefined)!;
+    promise.then(result => {
```
bvaughn, line 98: I think if the component were to re-request this data before the promise finishes, you'd end up adding multiple `.then()` callbacks, wouldn't you?

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
bvaughn: React's (current, experimental) `getCacheForType` API provides some additional functionality _that we don't need_ in this case:
1. Ability to _invalidate_ the cache (when data changes) as part of a low-priority update – without blocking high-priority updates that might come up.
2. Ability to mask a higher level cache with a lower level one. (For example, a higher level cache may be a friends list with only a few user attributes, and a lower level cache might be a user's profile details.)

In our case (1) sources are never invalidated and (2) the recording and its sources are global for an entire session.

bvaughn: In that sense, this cache is more like the React DevTools "hook names" cache:
https://github.com/facebook/react/blob/cf665c4b73a28b034c8173f4d929205fb12d2d2e/packages/react-devtools-shared/src/hookNamesCache.js#L64-L68

packages/markerikson-stack-client-prototype/src/pages/index.tsx
```diff
@@ -1,12 +1,19 @@
...
 import { useContext } from "react";
 
+import Loader from "../components/Loader";
 import { SourcesList } from "../features/sources/SourcesList";
 import { SourceContent } from "../features/sources/SourceContent";
 
+import { store } from "../app/store";
+import { setStore } from "../features/sources/sourcesCache";
+
+setStore(store);
```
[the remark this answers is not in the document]
markerikson: Potentially, yeah, although this is technically a pattern we do recommend for cases where non-React-related modules need access to the store:

https://redux.js.org/faq/code-structure#how-can-i-use-the-redux-store-in-non-component-files

In theory, having it get injected from a root file should mean no import order issues.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
markerikson: This is part of where I really don't know what I'm doing with Suspense Caches :)  I was attempting to mimic what I saw over in your prototype folder.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
markerikson: RTKQ differentiates between "loading", which is only true during the initial setup, and "fetching", which can go between false/true many times if you change the parameters being passed into a given query hook:

In this case, we're doing the work outside of any hook, so I actually copy-pasted this logic from inside of RTKQ's hooks code.  It may not be totally necessary here, I'd have to think about it a bit more.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
markerikson: See above comment - for the hooks the parameter can change, and that's where I pulled this code from.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
markerikson: Hmm. _Can_ that happen? Because the component is suspended and we're waiting for this exact promise to resolve before it will render again.

packages/markerikson-stack-client-prototype/src/pages/index.tsx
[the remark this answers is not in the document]
bvaughn: I think it's good to avoid globals as much as possible.

Even injecting it from the (application) "root" could lead to awkward cases when writing tests for a component, for example. If we don't require/render the root, but the component uses the hook– then it's going to run into an undefined value error.

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
bvaughn: Yeah totally makes sense. Was just sharing some thoughts as I read this but maybe it would be better for us to hop on a VC and talk about it?

packages/markerikson-stack-client-prototype/src/features/sources/sourcesCache.ts
bvaughn: Yes. React might try re-rendering in some scenarios. This is why it's important that the _thing that's thrown_ is stable, for example. (That would also be true of any event handlers you added to that thing.)

markerikson: Gotcha.  On the other hand, in this particular scenario, is that actually a problem?

The `.then` is solely for the purpose of mutating the record with the received value, so if it runs more than once it's a no-op.  Also, we're throwing the original promise, not the `.then` promise.

(I was also having a bit of trouble trying to translate the sync "wakeable" logic in your example over to an async promise, which is how I ended up with this setup.)

bvaughn: That's fair.

Unless I'm misunderstanding something, I think this cache could be simplified to the following:
```ts
import { unstable_getCacheForType as getCacheForType } from "react";

import { api, SourceGroups } from "../../app/api";
import type { AppStore } from "../../app/store";

let hasPromiseListenerBeenAttached = false;
let sourceGroups: SourceGroups | null;

const selectSourceGroupsEntry = api.endpoints.getSources.select();

export function getSourceGroups(store: AppStore): SourceGroups {
  if (sourceGroups !== null) {
    return sourceGroups;
  }

  let sourcesEntry: ReturnType<typeof selectSourceGroupsEntry> | undefined =
    selectSourceGroupsEntry(store.getState());

  if (!sourcesEntry || sourcesEntry.isUninitialized) {
    store.dispatch(api.endpoints.getSources.initiate());
    sourcesEntry = selectSourceGroupsEntry(store.getState());
  }

  const promise = api.util.getRunningOperationPromise("getSources", undefined)!;

  if (!hasPromiseListenerBeenAttached) {
    hasPromiseListenerBeenAttached = true;
    promise.then(result => {
      sourceGroups = result.data as SourceGroups;
    });
  }
  
  throw promise;
}
```

markerikson:
Some updates:

- Rebased this on `main` to pick up all the Yarn changes
- Reworked the API + client setup so that the client gets injected into the API file by `<Initializer>`
- Removed the `setStore()` from the cache file and am now passing it in to `getSourceGroups(store)`
- Replaced my original more complicated implementation with Brian's simpler implementation, since the sources list is a one-time fetch

PR #7207, opened 2022-06-17
Add routing for test run links

jaril:


src/ui/components/Library/Content/TestRuns/TestRunList.tsx
```diff
@@ -1,34 +1,42 @@
...
+    if (!loading && testRuns && !preview) {
+      let testRunId;
+
+      // Check that we have a test run that corresponds with the provided test run id
+      // from the URL. If not, use the first test run.
+      if (initialTestRunId && testRuns.some(run => run.id === initialTestRunId)) {
+        testRunId = initialTestRunId;
+      } else {
+        testRunId = testRuns[0].id;
+      }
```
ryanjduffy, line 25: This points to needing to lift state. We shouldn't have an `initialTestRunId` and a `testRunId`. There's only one state value, the test run id, and it's either set via the URL or by selecting it.

ryanjduffy (approved):

PR #7208, opened 2022-06-18
Show line hit counts by default

jasonLaster:
This uses some dark magic to highlight the gutter based on their hit counts. 

Todo list
- [ ] Update the hit-badge to use the new border based treatment
- [ ] remove the right gutter border
- [ ] adjust the breakpoint badge 

Screen Shot 2022-06-17 at 6 25 34 PM


src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -1,52 +1,74 @@
...
+      }\n
+      .cm-s-mozilla .hit-count-${count}:hover .CodeMirror-linenumber {
+        color: ${textHoverColor};
+      }\n
+      
+      .hits-badge-${count} {
+        color: ${foregroundColor};
+        background-color: ${backgroundColor};
+      }\n
+      `;
```
bvaughn, line 49: I'm concerned about this approach (if you're considering landing it).

Nested selectors can get pretty tricky as the number of style rules grow, and this approach to styling is outside of the half a dozen other approaches we're already using 😅

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -1,52 +1,74 @@
...
     // HACK Make sure we load hit count metadata; normally this is done in response to a mouseover event.
     if (hitCounts === null) {
       dispatch(setBreakpointHitCounts(sourceId, 0));
     }
+
+    let styles = "";
+    for (const [
+      count,
+      { foregroundColor, backgroundColor, textColor, textHoverColor },
+    ] of hitCountColor) {
```
bvaughn, line 36: This is a new effect dependency. It would need to be specified in the dependencies array.

src/devtools/client/debugger/src/utils/editor/hit-counts.ts
```diff
@@ -0,0 +1,47 @@
...
+    new Set(hitCounts ? hitCounts.map(hitCount => hitCount.hits) : [])
+  );
+  const maxHitCount = Math.max(...uniqueHitCounts);
+
+  const map = new Map(
+    uniqueHitCounts.map(hitCount => {
+      const avg = Math.min(1, Math.max(0, Math.log(hitCount) / Math.log(maxHitCount))) || 0;
+      const backgroundColor = interpolateLab(hitBadge[0], hitBadge[1])(avg);
+      const foregroundColor =
+        getLuminance(backgroundColor) > 0.5 ? "rgba(0,0,0,0.97)" : "rgba(255,255,255,.99)";
```
bvaughn, line 24: I don't understand the very specific but also kind of unexpected values of `rgba(0,0,0,0.97)` and `rgba(255,255,255,.99)`. Why .97 and .99?

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
jasonLaster: I'm not sure how many options codemirror will give us. 

With that said, this is just me prototyping to see how this ui feels.

src/devtools/client/debugger/src/utils/editor/hit-counts.ts
jasonLaster: yeah - i'm not sure I follow either. this came from elsewhere

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
bvaughn: Cool, cool 😁

src/devtools/client/debugger/src/utils/editor/hit-counts.ts
bvaughn: I think we might as well go with `#fff` and `#000` then 😁

PR #7209, opened 2022-06-18
Fix plus button loading bug

jasonLaster:
Fixes a small bug where we accidentally show loading if there aren't any found hits


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -118,29 +118,22 @@ function LineNumberTooltip({ editor, keyModifiers }: Props) {
     }
   }, [hits]);
 
-  if (
```
jasonLaster, line 121: this is no longer needed

bvaughn (approved):

PR #7211, opened 2022-06-18
Small timeline fixes

jasonLaster:
This fixes a couple of timeline interactions 

1. clicking on an unloaded region will now seek to the nearest point (similar to playing the video and pausing
2. hovering on an unloaded region will show unloaded in the tooltip
3. the timeline-container now acts as a click target (not just a hover target)

code
1. DownloadCancelled is no longer logged
2. getPointNearTime is moved to the threadfront

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -17,10 +17,12 @@ import {
   timeIsBeyondKnownPaints,
   screenshotCache,
 } from "protocol/graphics";
+import { DownloadCancelledError } from "protocol/screenshot-cache";
 import { ThreadFront } from "protocol/thread";
 import { Pause } from "protocol/thread/pause";
 import { PauseEventArgs } from "protocol/thread/thread";
 import { assert, waitForTime } from "protocol/utils";
+
 import { getFirstComment } from "ui/hooks/comments/comments";
 import {
   getCurrentTime,
@@ -192,7 +194,9 @@ export function setTimelineToTime(time: number | null, updateGraphics = true): U
 
       paintGraphics(screen, mouse);
     } catch (error) {
-      console.error(error);
+      if (!(error instanceof DownloadCancelledError)) {
+        console.error(error);
+      }
     }
   };
 }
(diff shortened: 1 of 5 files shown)


src/ui/actions/timeline.ts
```diff
@@ -245,15 +249,15 @@ export function seekToTime(targetTime: number): UIThunkAction {
...
-      // more exact method.
-      dispatch(seek(pointNearTime.point, targetTime, false));
+    let bestPoint = nearestEvent;
+    try {
+      const pointNearTime = await ThreadFront.getPointNearTime(targetTime);
+      if (Math.abs(pointNearTime.time - targetTime) < Math.abs(nearestEvent.time - targetTime)) {
+        bestPoint = pointNearTime;
+      }
+    } catch (e) {
+    } finally {
```
bvaughn, line 259: I think `finally` is only needed if you re-throw the error, but it's also fine either way 😄

src/ui/components/Timeline/Tooltip.tsx
```diff
@@ -29,8 +38,16 @@ export default function Tooltip({ timelineWidth }: { timelineWidth: number }) {
...
+    focusRegion &&
+    (displayedBeginForFocusRegion(focusRegion) > hoverTime ||
+      displayedEndForFocusRegion(focusRegion) < hoverTime);
+
   const timestamp = getTimestamp(hoverTime);
-  const message = isHoveredOnNonLoadingRegion ? `${timestamp} (Unloaded)` : timestamp;
+  const message =
+    isHoveredOnNonLoadingRegion || isHoveredOnUnFocusedRegion
+      ? `${timestamp} (Unloaded)`
+      : timestamp;
```
bvaughn, line 50: Nice 👍🏼

src/ui/components/Timeline/index.tsx
```diff
@@ -97,17 +97,13 @@ export default function Timeline() {
...
         >
           <div className="progress-bar-stack">
             <ProtocolTimeline />
-            <div
-              className="progress-bar"
-              ref={progressBarRef}
-              onClick={onClick}
-              onMouseMove={onMouseMove}
-            >
+            <div className="progress-bar" ref={progressBarRef} onMouseMove={onMouseMove}>
```
bvaughn, line 106: Feels weird to me to move the click handler but not the mousemove handler. Why?

bvaughn (approved):
Think this looks okay but left a question.

src/ui/components/Timeline/index.tsx
jasonLaster: oversight. mouseMove would be nice too

PR #7212, opened 2022-06-18
Restrict uses of the client and sendMessage to the protocol directory

jasonLaster:
This PR restricts uses of protocol/socket's client and sendMessage to the protocol directory. This is a nice win because it reduces the number of ways in which the application can talk to the protocol.

NOTE: this would not have been possible a couple of months ago before Brian and Mark started refactoring our protocol

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,6 +42,11 @@
             "name": "react-redux",
             "importNames": ["useDispatch", "useSelector"],
             "message": "Please import 'useAppDispatch/useAppSelector' from 'ui/setup/hooks' instead."
+          },
+          {
+            "name": "protocol/socket",
+            "importNames": ["client", "sendMessage"],
+            "message": "Please use ThreadFront."
           }
         ]
       }
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -14,8 +14,10 @@ import {
   FrameId,
   Location,
   MappedLocation,
+  keyboardEvents,
   Message,
   missingRegions,
+  navigationEvents,
   newSource,
   ObjectId,
   PauseDescription,
@@ -38,6 +40,8 @@ import {
   findAnnotationsResult,
   getHitCountsParameters,
   Frame,
+  PointRange,
+  TimeRange,
 } from "@replayio/protocol";
 import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
@@ -331,6 +335,10 @@ class _ThreadFront {
       }
       return this.annotationWaiters.get(kind)!;
     }
+  }
+
+  setAccessToken(accessToken: string) {
+    return client.Authentication.setAccessToken({ accessToken });
   }
 
   getRecordingTarget(): Promise<RecordingTarget> {
@@ -652,6 +660,10 @@ class _ThreadFront {
       : this.currentPause;
   }
 
+  loadRegion(region: TimeRange) {
+    return client.Session.loadRegion({ region }, ThreadFront.sessionId!);
+  }
+
   async loadAsyncParentFrames() {
     await this.ensureAllSources();
     const basePause = this.lastAsyncPause();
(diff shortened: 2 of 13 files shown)


.eslintrc.json
```diff
@@ -42,6 +42,11 @@
             "name": "react-redux",
             "importNames": ["useDispatch", "useSelector"],
             "message": "Please import 'useAppDispatch/useAppSelector' from 'ui/setup/hooks' instead."
+          },
+          {
+            "name": "protocol/socket",
+            "importNames": ["client", "sendMessage"],
+            "message": "Please use ThreadFront."
```
bvaughn, line 49: 😍

packages/protocol/thread/thread.ts
```diff
@@ -883,6 +899,23 @@ class _ThreadFront {
...
   }
 
+  findMessagesInRange(range: PointRange) {
+    return client.Console.findMessagesInRange({ range }, ThreadFront.sessionId!);
+  }
+
+  async findKeyboardEvents(onKeyboardEvents: (events: keyboardEvents) => void) {
+    const keyboardEventsLoaded = client.Session.findKeyboardEvents({}, this.sessionId!);
+    client.Session.addKeyboardEventsListener(onKeyboardEvents);
+    return keyboardEventsLoaded
```
bvaughn, line 909: The order of these two should probably be reversed (add the handler before calling `findKeyboardEvents`)
```suggestion
    client.Session.addKeyboardEventsListener(onKeyboardEvents);
    return client.Session.findKeyboardEvents({}, this.sessionId!);
```
Might not matter in this case but I think it's a good general practice to follow (so you don't miss any potentially synchronous events).

packages/protocol/thread/thread.ts
```diff
@@ -883,6 +899,23 @@ class _ThreadFront {
...
+  }
+
+  async findNavigationEvents(onNavigationEvents: (events: navigationEvents) => void) {
+    const navigationEventsLoaded = client.Session.findNavigationEvents({}, this.sessionId!);
+    client.Session.addNavigationEventsListener(onNavigationEvents);
+    return navigationEventsLoaded
```
bvaughn, line 915: Same as above (add the handler before calling `findNavigationEvents`)
```suggestion
    client.Session.addNavigationEventsListener(onNavigationEvents);
    return client.Session.findNavigationEvents({}, this.sessionId!);
```

packages/protocol/thread/thread.ts
bvaughn: PS Surprised the linter didn't complain about the missing semicolon on line 909

pages/recording/[id]/sourcemap/[sourceId].tsx
```diff
@@ -1,5 +1,6 @@
 import { useRouter } from "next/router";
 import React, { useEffect, useState } from "react";
+// eslint-disable-next-line no-restricted-imports
 import { initSocket, client } from "protocol/socket";
```
Comment (bvaughn) [inline pages/recording/[id]/sourcemap/[sourceId].tsx:4]: This use is still okay?

src/ui/actions/app.ts
```diff
@@ -161,8 +160,8 @@ function onKeyboardEvents(events: KeyboardEvent[], store: UIStore) {
   });
 }
 
-function onNavigationEvents(events: ReplayNavigationEvent[], store: UIStore) {
-  const currentNavEvents: ReplayNavigationEvent[] = events.map(e => ({ ...e, kind: "navigation" }));
+function onNavigationEvents(events: navigationEvents, store: UIStore) {
```
bvaughn, line 163: Oh that was bizarre before.

bvaughn (approved):
Nice! Couple of small nits.

pages/recording/[id]/sourcemap/[sourceId].tsx
Comment (jasonLaster) [reply pages/recording/[id]/sourcemap/[sourceId].tsx:4]: The sourcemap visualizer is its own beast.

src/ui/actions/app.ts
jasonLaster: a tad

packages/protocol/thread/thread.ts
jasonLaster: good call

.eslintrc.json
JessicaSachs: Yeah, okay, that's nice. Def stealing this as a tip.

PR #7213, opened 2022-06-18
Stop using protocol/socket log

jasonLaster:
This is a small cleanup. I'm pretty sure we're no longer using log since we can log the protocol directly.


bvaughn (approved):
Guessing you didn't rebase this branch? I think I see some changes from the other PR in it.

Changes look good though

PR #7218, opened 2022-06-19
Show the current team as the title

jasonLaster:
There might be a more idiomatic way to do this

Screen Shot 2022-06-18 at 9 02 01 PM


jaril (approved):
There might me but don't want to spend too much time on it. The new routing work should should make this easier.

bvaughn:
Cleaning up old/sale PRs. Closing this one as it seems abandoned. We can re-open and review if you decide to pick it back up, though 😄

PR #7219, opened 2022-06-19
Test results view

jasonLaster:
This is a fairly simple and unopinionated view. The goal is to help users see the latest results for a give test.

Today, I'm using the filter dropdown in the top left, but I'm not sure if we'll want to use that going forward. That is more of a crutch to get the UI out the door.

https://www.loom.com/share/b01a81b8f1714c4fb471dd49ca415d47
Screen Shot 2022-06-18 at 9 53 04 PM


jasonLaster:
Pushed a couple of other WIP ideas as conversation starters. will move them to their own pr

PR #7221, opened 2022-06-19
Fix hitCounts discrepency with print statements Fixes FE-289

jasonLaster:
This Fixes FE-289 by fetching the hit counts for each of the corresponding sources which is what we do for analysis as well.

Munging the data this way is certainly gross and we probably can do better. I initially tried following what logpoint.ts does with multiSourceLogPoint, but the reducer got annoying because `o2414` could not be found.

This works and I'd like to land the hot fix and then try and improve it as a high priority follow up.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -498,11 +498,35 @@ class _ThreadFront {
     range: { beginPoint: ExecutionPoint; endPoint: ExecutionPoint } | null
   ) {
     assert(this.sessionId, "no sessionId");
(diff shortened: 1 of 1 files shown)


markerikson (approved):

jcmorrow:
The part that strikes me as unfortunate about this is that we already do a bunch of "same source" mappings on the backend. This logic is ugly, not because of the way it's written, but just because source and source map logic is complicated. At the least though, we should centralize that complexity in one place. Accounting for it on the frontend *and* the backend makes it really easy for little bugs like this to sneak in.

PR #7222, opened 2022-06-19
Teach loadRegion to first unload surrounding regions

jasonLaster:
Fixes BAC-1813

This PR rolls back some of the changes in #7114

The rationale for this change is two fold
1. the backend will not load a new region if it is at capacity
2. asking the backend to unload a region is cheap because it no longer does unless a controller disconnects.

I imagine that we'll get smarter over time, but this seems like a fine solution for the time being

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -678,6 +678,19 @@ class _ThreadFront {
       : this.currentPause;
   }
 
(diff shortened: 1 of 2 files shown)


jcmorrow:
> the backend will not load a new region if it is at capacity

Iiiinteresting. I did not know this!

PR #7194, opened 2022-06-16
[DRAFT] Try using PNPM as a package manager

markerikson:
This PR:

- Switches package installation from NPM 8 to PNPM
- Switches the GH Actions setup to use PNPM, with cache setup

Compare with the Yarn 3 PR at #7192

PR #7239, opened 2022-06-21
[DRAFT] Further PNPM workspaces exploration

markerikson:
### DO NOT MERGE!

This PR is an additional set of changes on top of the PNPM workspace setup in #7194 .  The goals are:

- See how CI caching works and check install/build times as multiple pushes happen
- Try using PNPM for some package management scenarios (top-level dep, nested dep, new subpackage) and see how the changes look

Since this is all unrelated to the actual app work, it's a separate PR and we should close this without merging - we don't want any of these changes in the base PR.


markerikson:
I'm making an arbitrary decision that we're going to go with Yarn just because I'm more familiar with it. Closing.

PR #7241, opened 2022-06-22
Move ReplayClient and ReplayClientContext to packages/shared

bvaughn:
Step 1 to trying the new Object Inspector out in the Replay app is to move the `ReplayClient` to a shared location. This also cleans up some of the redundant code between Mark's and my prototypes.


packages/bvaughn-architecture-demo/components/Initializer.tsx
```diff
@@ -3,7 +3,8 @@
 
 import { ReactNode, useContext, useEffect, useRef, useState } from "react";
 
-import { ReplayClientContext } from "../src/contexts/ReplayClientContext";
+import { ReplayClientContext } from "../../shared/client/ReplayClientContext";
```
markerikson, line 6: :speech_balloon: fyi, declaring `shared` as a dep and installing it would let us drop the `"../../"` prefix, although that may work better once we switch package managers

markerikson (approved):

packages/bvaughn-architecture-demo/components/Initializer.tsx
bvaughn: I had trouble getting this to work. I'll try again!

PR #7245, opened 2022-06-22
Mark prototype uses ReplayClientContext

bvaughn:
Shuffling things around to remove some more redundancy.


bvaughn:
Trying to figure out how we're going to use this context in _the real app_ still (without trying to create two WebSockets and two sessions etc) and wanted to get our prototype apps both using the client+context in the same way first at least.

markerikson (approved):

bvaughn:
Merged by way of #7247

PR #7250, opened 2022-06-22
Use a subscription instead of polling/refetching comments and sessions

hbenl:
To test this you need to run a local GraphQL API using Replayio/backend#5990 and point the app to that API by setting the URLs in `.env`, e.g.
```
NEXT_PUBLIC_API_URL=http://localhost:8081/v1/graphql
NEXT_PUBLIC_API_SUBSCRIPTION_URL=ws://localhost:8081/v1/graphql
```

### DIFF
--- a/src/ui/actions/comments.ts
+++ b/src/ui/actions/comments.ts
@@ -48,7 +48,6 @@ export function createComment(
 
     const response = await mutate({
       mutation: ADD_COMMENT_MUTATION,
-      refetchQueries: ["GetComments"],
       variables: {
         input: {
           content: "",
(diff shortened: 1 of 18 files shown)


src/ui/actions/comments.ts
```diff
@@ -48,7 +48,7 @@ export function createComment(
 
     const response = await mutate({
       mutation: ADD_COMMENT_MUTATION,
-      refetchQueries: ["GetComments"],
+      // refetchQueries: ["GetComments"],
```
jasonLaster, line 51: Love this line

src/ui/hooks/recordings.ts
```diff
@@ -158,20 +158,38 @@ export async function getRecording(recordingId: RecordingId) {
   return convertRecording(result.data?.recording);
 }
 
+let subscribed = false;
```
jasonLaster, line 161: could this be a `useState`

markerikson:
Note that this will need to be rebased against `main` and updated to use Yarn instead

jaril (approved):
Seems to work well — LGTM!

PR #7261, opened 2022-06-23
Nicer "copy link" experience

jonbell-lot23:
Here's a Loom showing the difference:
https://www.loom.com/share/68cc08d77a3d40dc9af8b0cb6b749d5f

### DIFF
--- a/src/ui/components/Library/Overview/RunSummary.tsx
+++ b/src/ui/components/Library/Overview/RunSummary.tsx
@@ -1,4 +1,4 @@
-import { useContext } from "react";
+import { useContext, useState, useRef } from "react";
 import { useSelector } from "react-redux";
 import { getWorkspaceId } from "ui/actions/app";
 import MaterialIcon from "ui/components/shared/MaterialIcon";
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):
Nice work! What do you think about also showing the copy button when you hover on the title. I think this is pretty common.

jonbell-lot23:
> Nice work! What do you think about also showing the copy button when you hover on the title? I think this is pretty common.

Totally -- do you mean doing the blue highlight when hovering on the title, or do you mean hiding the icon until you mouse over the title?

I considered the latter and decided against it because it's not discoverable, but I think the former sounds good. Only thing is I need to figure out how to do it :)

jasonLaster:
@jonbell-lot23 we should probably step back and think through whether a copy button makes sense in this use case

1. it was never in any designs and i'm pretty sure Jaril added it primarily for testing purposes ([link](https://github.com/replayio/devtools/pull/7234))
2. the copy button currently copies the url for the test run, but that should already be the URL in the url bar. This is not currently the case, but I consider that a bug.
3. We do not currently have a copy button for internal replay links. If we were to do want to expose something like that it would typically be through an actions dropdown with a label explaining the action.

jonbell-lot23:
Sounds good, easy to delete when it's not needed for testing!

PR #7263, opened 2022-06-23
Use react-window for Console auto complete typeahead

bvaughn:
This commit fixes one of the outstanding performance problems with the console type-ahead feature by using [react-window](https://react-window.vercel.app/) to render the list of items.

This PR also fixes a couple of bugs that existed before:
* Clicking on an item selects _the item you clicked_ rather than the one that happened to be keyboard focused.
* If you scroll to an item, and then key up/down, the selected item scrolls back into view.

Loom walk through:
https://www.loom.com/share/07aba245febe41fdac2a036983639398

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "react-tooltip": "^4.2.21",
     "react-transition-group": "^4.4.2",
     "react-virtualized-auto-sizer": "^1.0.6",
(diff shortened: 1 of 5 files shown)


jasonLaster:
Love this and the Loom w/ the perf explanation

markerikson (approved):

PR #7264, opened 2022-06-23
Fix Next compile error in bvaughn prototype

bvaughn:
@markerikson Any idea why this change is necessary now?


markerikson:
Not immediately sure, no.  I would assume it's related to something about how Yarn is installing workspace subpackages, but not sure beyond that.

PR #7272, opened 2022-06-24
Persist hit counts mode

jasonLaster:
Screen Shot 2022-06-24 at 6 54 06 AM


src/devtools/client/debugger/src/components/Editor/Gutter.tsx
```diff
@@ -4,15 +4,10 @@ import LineHitCounts from "./LineHitCounts";
 import ToggleWidgetButton from "./ToggleWidgetButton";
 
 export default function Gutter({ editor }: any) {
-  const [isLineHitCountsCollapsed, setIsLineHitCountsCollapsed] = useState(false);
```
jasonLaster, line 7: I moved the state into redux so it would be easier to synchronize with prefs. Most of the changes below are there because of the move to redux.

PR #7278, opened 2022-06-24
Automated, reproducible testing architecture

bvaughn:
One of the goals of the new Replay devtools architecture was to enable test– whether automated or component UI testing– to be more light weight. This goal influenced the design of the `ReplayClientInterface`– which provides a production-ready client by default, but allows for overrides using React's Context API in the case of tests.

This PR explores how we might build on top of this concept to create mock/stub data for testing purposes– essentially running the client with "real" data, but without having to talk to a production backend.

Here is a Loom walk through of the idea: https://www.loom.com/share/0bf6819ffd9b43de9bc353a0d5893c8c

## The recorder
The "recorder" client can be used to wrap around the real client and record inputs and outputs for a session. This data is exposed via the `logEntries` array.
```js
function Routing({ Component, pageProps }: AppProps) {
  const replayClientRecorder = useMemo(
    () => createReplayClientRecorder(client),
    [client]
  );

  return (
    <ReplayClientContext.Provider value={replayClientRecorder}>
      <ReplayDevToolsApp />
    </ReplayClientContext.Provider>
  );
}
```

Each time a message is sent through the Replay client _recorder_, it will print a message with the exact JSON configuration string needed to create a matching _player_ instance.

## The player
The "player" client can be used to playback recorded log entries for testing purposes. (We _could_ even add the concept of delay into the async methods, but for now they just resolve immediately.)
```js
// This exact string (with the log entries JSON) can be copy-pasted from the console after running the recorder.
const replayClientPlayer = createReplayClientPlayer(logEntries);

function Routing({ Component, pageProps }: AppProps) {
  return (
    <ReplayClientContext.Provider value={replayClientPlayer}>
      <ReplayDevToolsApp />
    </ReplayClientContext.Provider>
  );
}
```

---

This recorder can be seen by checking out the branch, running the test app, and then opening http://localhost:3000/tests/console/basic

A new recording can be captured by appending `&record` to the test app's URL.


bvaughn:
For discussion!

jcmorrow:
This is so cool!

bvaughn:
Thank you!

PR #7279, opened 2022-06-24
Clean up the experimental settings tab

jaril:
Fix FE-340.

image
image


jasonLaster:
The label for the OI should be

> Preview objects with the new inspector

PR #7280, opened 2022-06-25
GitHub actions that run Jest and Playwright architecture tests

bvaughn:

PR #7286, opened 2022-06-27
GitHub actions to run Playwright screenshot tests in Docker

bvaughn:
Builds on top of #7280 but runs Playwright tests with Earthly to avoid OS rendering differences.

Getting this working was kind of a nightmare, because:
1. If Playwright fails, Earthly will stop (before saving "test-result" artifacts).
2. Even if we temporarily ignore the test failure, and save the "test-results" artifacts folder, Earthly does not complete the copy action if its process exits with an error code (see https://github.com/earthly/earthly/issues/988#issuecomment-873433240).

The work arounds for the above issues are:
1. Return a error code 0 regardless of Playwright's exit code, but write a special marker file `playwright_error` to signal that Playwright railed (see `packages/bvaughn-architecture-demo/playwright/run_tests_for_earthly.js`)
2. Check for the `playwright_error` marker file and save Playwright's "test-results" folder to the host file system if its present (see `packages/bvaughn-architecture-demo/playwright/Earthfile`)
3. Check for the "test-results" folder and infer a failure, ultimately failing the GitHub Workflow when it's present (see `.github/workflows/bvaughn-playwright.yml`)


packages/bvaughn-architecture-demo/playwright/Earthfile
```diff
@@ -0,0 +1,19 @@
...
+
+playwright-ci:
+    FROM +install
+    RUN yarn test
+
+playwright-local:
+    FROM +install
+    RUN yarn test --update-snapshots
+    # TODO Change this structure
+    SAVE ARTIFACT ./snapshots AS LOCAL ./snapshots
```
jasonLaster, line 19: very cool

jasonLaster (approved):
Excited to see this land! ✈️

packages/bvaughn-architecture-demo/playwright/Earthfile
bvaughn: Haha, it _would be_ if I could get GitHub to run `earthly` ! 😆  Thanks though

bvaughn:
Okay! I think we're good to go.

packages/bvaughn-architecture-demo/playwright/Earthfile
bvaughn: Hm. This seems to exit too early.

PR #7288, opened 2022-06-28
More docs on sources

hbenl:
I must admit that these docs still make my own head spin but that's because the relationships between sources (original/generated, preferred/alternate and corresponding (i.e. identical) sources) and how those relationships relate to each other are complicated.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -1,50 +1,34 @@
-...tl;dr of how things work in general and some basic structure
+# Replay DevTools Codebase Concepts and Key Terms
 
-# `ExecutionPoint`
+This document includes some incomplete but still very useful info about a number of key concepts used inside the Replay DevTools client and backend codebases, as well as a few of the component areas in the client.
 
-An [`ExecutionPoint`](https://static.replay.io/protocol/tot/Recording/#type-ExecutionPoint) is a numeric string
-that represents a point in time in the recording.
-`ExecutionPoint`s for earlier points in time are smaller numbers than the ones for later points in time.
-However, it is not possible to compute the time (i.e. the milliseconds elapsed since the beginning of the recording)
-from an `ExecutionPoint`, so that information is often transmitted together with an `ExecutionPoint`
-in a [`TimeStampedPoint`](https://static.replay.io/protocol/tot/Recording/#type-TimeStampedPoint).
-For more information on `ExecutionPoint`s see https://github.com/RecordReplay/backend/blob/master/docs/execution-points.md
+## Regions and Time Ranges
 
-# Regions
+The Replay backend has to load portions of the recording into memory, and allows the client to request that the backend pause the recording at specific points in time.
 
-A region is a time range in the recording, usually described by a
+### Regions
+
+A **region** is a time range in the recording, usually described by a
 [`TimeStampedPointRange`](https://static.replay.io/protocol/tot/Recording/#type-TimeStampedPointRange).
+
 When the backend loads a recording, it divides it into regions that can be unloaded individually to free up resources.
 Regions can be unloaded spontaneously by the backend to alleviate memory pressure, but the frontend can also tell the backend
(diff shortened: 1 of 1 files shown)


docs/concepts.md
```diff
@@ -21,7 +21,8 @@ the frontend requests that all regions that do not intersect with the selected t
...
+When a recording is opened, the frontend loads all javascript sources in that recording using [`Debugger.findSources`](https://static.replay.io/protocol/tot/Debugger/#method-findSources) and stores them in the [`ThreadFront`](../src/protocol/thread/thread.ts). Sourcemaps are handled in the backend, so the sources also include original sources from sourcemaps. Furthermore, the backend automatically adds pretty-printed versions of minified sources and extracted inline scripts from html pages. The [`SourceKind`](https://static.replay.io/protocol/tot/Debugger/#type-SourceKind) specifies where a source came from.
+The [`newSource`](https://static.replay.io/protocol/tot/Debugger/#event-newSource) events from the backend include the list of generated sources for every original source. This information is used to pick the sources that are shown to the user: Replay always tries to show an original source and may offer a generated source as an alternate that the user can switch to using the toggle at the bottom of the editor.
```
jasonLaster, line 25: > events from the backend include the list of generated sources for every original source

Did you mean: events from the backend include the list of original sources for every generated source?

hbenl: No, see https://static.replay.io/protocol/tot/Debugger/#event-newSource. The `ThreadFront` creates the list of original sources for every generated one from that.

jcmorrow: Nope, they include a list of generated sources for original sources (at least that's how it works per the docs at https://static.replay.io/protocol/tot/Debugger/#event-newSource)

docs/concepts.md
```diff
@@ -40,9 +41,13 @@ This is what a typical source graph for one source file loaded by the browser ma
...
+
+The selection of preferred and alternate sources is implemented in `ThreadFront._chooseSourceId()`. It takes an array of sourceIds that usually come from a `MappedLocation`, removes those of minified sources (i.e. those for which a pretty-printed version exists) and inline scripts (if the corresponding html page is among the sources) and categorizes each source as "original" if it is from a sourcemap or a pretty-printed version of a source from a sourcemap and "generated" otherwise. It assumes that there is at most one "generated" source according to this categorization. If there is at least one "original" source, it picks that as the preferred source and the "generated" source (if any) as the alternate. Otherwise there should be exactly one source which is "generated" and that is the preferred source.
```
jcmorrow, line 50: This was a @jasonLaster suggestion, but I agree, it might be helpful to centralize a list of terms at the top of this documentation. The terms that are non-obvious IMO are:

- Original source: This is what the developer typed into their editor, before it was run through any toolchains
- Generated source: The output of running an original source through a toolchain (might be minified, transpiled, bundled, or otherwise transformed)

- Corresponding sources: a file which was loaded multiple times during a single recording, from the same URL, with the same content. For example, because the user refreshed the page.
- Alternate sources: this one is slightly tricky to express concisely, I think. It's a group of sources which map onto each other. For example, the "point" of execution might correctly be identified as having several different locations. One could be an original source (which the browser was unaware of at runtime), another could be the generated source (which the browser *did* run but which can be mapped onto the original) and another might be a pretty-printed file (which didn't exist when the browser was running, and is instead the product of us doing our own pretty-printing post-recording).
- Preferred source: A source from a group of alternate sources which is considered the "best" choice as a canonical representation of that source. Often this will be an original if one is available. If it's not, pretty-printed is preferred, and last would be the generated, potentially still minified source.
- Preferred source: the "best" (by our reckoning) source for a group of alternate sources

hbenl: I'll try to add that, the problem is that most of these terms are used with slightly different meanings in different places. But that just makes it more important to add an explanation.

jcmorrow: I've put expanded versions of these definitions into the PR itself, but I'm realizing now that my definitions actually conflict with how we do things right now.

PR #7296, opened 2022-06-29
Better Library Routing 

jaril:
Fixes FE-332, FE-336 

# Motivation

This PR changes the way we navigate between teams in the library. 

Prior to this, we would obscure routing details and store it in Redux. That made it behave as if it we're taking advantage of client-side routing, but without some of the key benefits such as:
- An accurate link in the URL bar for linking to any team
- A history stack for back/forward navigation

This was workable up until now — navigation in the Library was minimal and mostly consisted of opening a workspace and finding a recent replay that you recorded. However, as we support test suites, those two benefits became crucial for common use cases such as: 
1) sharing a link to a test run that you're viewing to a team member, or 
2) investigating a test result A from a test run, then navigating back to the test run and investigating test result B.

# Routing implementation

Read here: https://www.notion.so/replayio/202206300642-Library-Routing-a8e4f5f5da9c475882ef3cca3c4005b8

# Change notes

## User-facing
- Going to `app.replay.io` still routes you to your last selected workspace, except it will now reflect in the URL like so: `app.replay.io/team/<team-id>/<view>`
- Non-test teams default to `/recordings`, while test teams default to `/runs`. Non-test teams can only have the `recordings` view, while test teams can only have the `runs`/`results` view
- Only show a view option (e.g. "Show Test Runs", "Show Test Results") in the dropdown if the user is in a test team
- Filtering is disabled for Test Results
- Changed the invitation URL from `/team/invitation/[code]` to `/team/invitation?code=[code]`
- Handled having an invalid team ID in the URL by redirecting you to your library
- Changed the entire test run result row to be no longer clickable (ask me about nested links). Only the Play button (which redirects you to the replay) or test name (which redirects you to a filtered view of test results for that test path)

## Not user-facing
- Implemented the [new routing conventions](https://www.notion.so/replayio/202206300642-Library-Routing-a8e4f5f5da9c475882ef3cca3c4005b8)
- Removed the `workspaceId` from the Redux store, and store it in the URL instead
- Removed the `LibraryContext` where we previously would store the `view` and `filter`. Instead, view is in `ViewContext` and filter is in `FilterContext`
- Tucked nag handling in the library into its own component `LibraryNags` that takes care of its own side effects

# Post-deploy
- [ ] Modify the billing links in Courier to follow the format: `app.replay.io/team/{teamId}?settings=billing`

## Follow up tasks (in order of importance)
_These are things that aren't already in Linear but I don't consider to be blocking:_
- This currently breaks the onboarding flow of accepting an invitation -> being redirected to that team. Fix that.
- Fix back to library interaction from a recording of a test
- Write up onboarding flow URLs that we currently support
- Redo PendingTeamScreen because ew
- Separate your library into its own route if possible
- Loading states — planted a bunch of todos with `LibrarySpinner`


src/ui/components/Library/Team/View/Recordings/Header/BatchActionDropdown.tsx
```diff
@@ -2,18 +2,17 @@ import { RecordingId } from "@replayio/protocol";
 import React, { useState } from "react";
 import hooks from "ui/hooks";
 import { connect, ConnectedProps } from "react-redux";
-import * as selectors from "ui/reducers/app";
-import { UIState } from "ui/state";
 import { WorkspaceId } from "ui/state/app";
-import { Dropdown, DropdownDivider, DropdownItem } from "./LibraryDropdown";
-import { DisabledButton, getButtonClasses } from "../shared/Button";
-import MaterialIcon from "../shared/MaterialIcon";
+import { Dropdown, DropdownItem } from "../../../../LibraryDropdown";
```
jasonLaster, line 6: probably want to use absolute paths

src/ui/components/Library/Team/View/Recordings/Header/BatchActionDropdown.tsx
```diff
@@ -138,8 +137,6 @@ function BatchActionDropdown({
   );
 }
 
-const connector = connect((state: UIState) => ({
-  currentWorkspaceId: selectors.getWorkspaceId(state),
-}));
+const connector = connect(() => ({}));
```
jasonLaster, line 140: can we drop redux here?

src/ui/components/Library/Team/View/Recordings/PendingTeam/PendingTeamPrompt.tsx
```diff
@@ -77,11 +77,6 @@ function PendingTeamPrompt({ workspace, setWorkspaceId }: PendingTeamPromptProps
   );
 }
 
-const connector = connect(
-  (state: UIState) => ({
-    currentWorkspaceId: selectors.getWorkspaceId(state),
-  }),
-  { setWorkspaceId: actions.setWorkspaceId }
-);
+const connector = connect((state: UIState) => ({}));
```
jasonLaster, line 80: same

src/ui/components/Library/Team/View/Recordings/RecordingListItem/MoveRecordingMenu.tsx
```diff
@@ -50,8 +49,6 @@ function MoveRecordingMenu({
   );
 }
 
-const connector = connect((state: UIState) => ({
-  currentWorkspaceId: selectors.getWorkspaceId(state),
-}));
+const connector = connect(() => ({}));
```
jasonLaster, line 52: ditto

jasonLaster (commented):
Nice!

jasonLaster:
Couple of minor details
- The normal libary filter dropdown "7 days" etc does not do anything
- The test library filter dropdown does not seem helpful so should probably be hidden

jaril:
## To test
- nags
- onboarding
- deleting team/leaving team
- accepting invite/declining invite
- dark mode

ryanjduffy:
So I didn't really do a code review because it's not really feasible at this size. I did read your doc and the PR description and they are great. Thanks so much for writing all of that up. Also, this is just a huge task and kudos for getting it as far as you have.

I did some user testing and here's what I found:

* This is weird. Would like to improve this route:
  image
* Navigating into team settings when a run is selected closes the run. Since settings is a modal, the underlying UI state should remain the same
* similar, if you select a test to see all the runs for it and open settings, it reverts back to the list of runs
* these both work correctly for User settings so the bug is limited to team settings
* The drop down options for "last 7 days" and "node recordings" seem to do nothing
* this is probably unrelated but there isn't any library management UI for test runs or tests so you can't delete them from the library
* The /settings/billing sub-path for teams doesn't work. It's used by courier to link users back to the billing page if their sub is expiring

Happy to re-test and/or discuss more as you need!

jaril:
> * This is weird. Would like to improve this route:
>   image

Yeah got painted into a corner there. Jon and I were talking about how that should just be `app.replay.io/my-library`. Will likely do that as a follow up.

> * Navigating into team settings when a run is selected closes the run. Since settings is a modal, the underlying UI state should remain the same
> * similar, if you select a test to see all the runs for it and open settings, it reverts back to the list of runs
> * these both work correctly for User settings so the bug is limited to team settings

Good catch. We have a link within a link that was doing a number on that interaction. Fixed it!

> * The drop down options for "last 7 days" and "node recordings" seem to do nothing
> * this is probably unrelated but there isn't any library management UI for test runs or tests so you can't delete them from the library

Correct. For now, I didn't see a reason for why a user might want to delete a run (or the recordings for a particular run, for that matter). Happy to revisit if we see a reason for it.

> * The /settings/billing sub-path for teams doesn't work. It's used by courier to link users back to the billing page if their sub is expiring

Forgot to note this — that still exists, but it's now handled as a query parameter. So once this gets in, we'll have to switch the courier links to `app.replay.io/team/[team-id]?settings=billing`. That's handled [here](src/ui/components/Library/Team/ParamHandlers/SettingsHandler.tsx).

ryanjduffy:
> The drop down options for "last 7 days" and "node recordings" seem to do nothing

This is still broken but otherwise I think we can give it a try.

ryanjduffy (approved):

jaril:
> > The drop down options for "last 7 days" and "node recordings" seem to do nothing
> 
> This is still broken but otherwise I think we can give it a try.

Wait I didn't merge that branch. I'll fix it before hitting merge. Thanks!!

PR #7297, opened 2022-06-30
Fix SourcemapToggle and SourcemapVisualizerLink

hbenl:
Both SourcemapToggle and SourcemapVisualizerLink need to find the alternate source to switch to but the logic for that was broken.

In this recording the SourcemapToggle should say "No sourcemaps found" because the recording doesn't contain sourcemaps:
Original recording: https://app.replay.io/recording/9e0910b1-43a4-478b-ba99-b6dbf9ba37a5
Recording showing the bug (from FE-143): https://app.replay.io/recording/use-sourcemap-click-bug--fed936ac-f1f9-4447-bbc4-f39911dda46f
Here's another recording with the same bug: https://app.replay.io/recording/minified2--82ae7e80-692d-4b0c-8d03-23292a96c3ff

In this recording the SourcemapToggle says "No sourcemaps found" for "exceptions_bundle.js" even though there is a sourcemap:
Original recording: https://app.replay.io/recording/replay-of-localhost8080--244a43ab-a6b7-4d3d-85c0-1684439ed280
Recording showing the bug: https://app.replay.io/recording/why-no-sourcemap-found--444cc3eb-1a3f-4494-8f09-e8c3d99ea358

I've rebuilt that logic, now it explicitly follows the links in the source(map) graph to find the alternate source.
Note that if we're paused in the selected source then we don't need that logic as we can rely on the alternate source that we get from the `MappedLocation` of the paused frame.

Fixes FE-143
Fixes FE-251

### DIFF
--- a/src/devtools/client/debugger/src/utils/sourceVisualizations.ts
+++ b/src/devtools/client/debugger/src/utils/sourceVisualizations.ts
@@ -1,6 +1,7 @@
 import type { SourceId } from "@replayio/protocol";
 import sortBy from "lodash/sortBy";
 import { ThreadFront } from "protocol/thread";
+import { assert } from "protocol/utils";
 
 import type { Source } from "../reducers/sources";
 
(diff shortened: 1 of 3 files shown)


markerikson (approved):

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -92,23 +77,28 @@ export function getUniqueAlternateSourceId(sourceId: string): {
   why?: "no-sourcemap" | "not-unique";
 } {
   const generatedSourceIds = ThreadFront.getGeneratedSourceIds(sourceId);
-  if (!generatedSourceIds?.length) {
-    return { why: "no-sourcemap" };
-  }
+  const nonPrettyPrintedSourceId =
```
jcmorrow, line 80: Only somewhat related: is it possible for both generated and original sources to have pretty-printed versions? If so, it seems like things get confusing?

hbenl: Yes, that's possible.

jcmorrow: That seems like something that the idea of "only one unique alternative source" does not seem to play nicely with. Does that seem correct to you?

hbenl: No, the logic for determining the "unique alternative source" basically conflates minified sources with their pretty-printed versions and then looks at "the other" source (because when you ignore pretty-printing, you're left with 2 sources again) and returns its pretty-printed version.

hbenl: I don't understand why it shouldn't work in that case. Maybe we should VC on that later.

jcmorrow: Yeah, I agree that there's nothing wrong with this implementation. I think (correct me if I'm wrong) that if we have 4 sources:

Page 1 (3)

And we view `o1` and we ask for it's alternate we will get `pp1`. Is that right? And then if we ask for the alternate of that we will get `ppo1`. It's not a bug in your implementation, I think it's just a limitation of the idea of "there can only be one alternative" when in fact a source was actually the middle step in a series of transformations.

hbenl: > And we view o1 and we ask for it's alternate we will get pp1. Is that right? And then if we ask for the alternate of that we will get ppo1.

Yes, that is exactly what would happen. BUT in this scenario you should never be able to see `o1`, only `ppo1`. If you ever manage to open `o1`, that would be a bug!

PR #7298, opened 2022-06-30
Attempt to make the frontend more resilient to pause timing

jasonLaster:
Changes
1.  general cleanup 
2. move pauseRequestedAt up so it always occurs
3. `isCurrentTimeInLoadedRegion` to `isPointInLoadedRegion` to focus on the new pause point

Hypothesis
1. isPaused will now be set if we receive a pause request
2. rewinding will succeed even if we are not finished loading

Confidnce:
low because if we are still loading regions, the pause request for the new rewind will also fail

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/commands.js
+++ b/src/devtools/client/debugger/src/actions/pause/commands.js
@@ -2,22 +2,12 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-//
-
 import { getResumePoint, getFramePositions } from "../../selectors";
 import { getLoadedRegions } from "ui/reducers/app";
 import { PROMISE } from "ui/setup/redux/middleware/promise";
-import { recordEvent } from "../../utils/telemetry";
 
 import { setFramePositions } from "./setFramePositions";
 
-/**
- * Debugger commands like stepOver, stepIn, stepUp
- *
- * @param string $0.type
- * @memberof actions/pause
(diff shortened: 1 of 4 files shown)


markerikson (approved):

src/ui/reducers/app.ts
```diff
@@ -370,3 +370,18 @@ export const isCurrentTimeInLoadedRegion = createSelector(
...
   }
 );
+
+export const isPointInLoadedRegion = createSelector(
+  getLoadedRegions,
+  (state: UIState, executionPoint: string) => executionPoint,
+  (regions: LoadedRegions | null, executionPoint: string) => {
+    return (
+      regions !== null &&
+      regions.loaded.some(
```
jcmorrow, line 380: I think we actually care about `loading` here, not `loaded`. I'm fairly sure that f we request a pause at a time where the backend is still loading it's fine, it will just take a while to get created.

src/devtools/client/debugger/src/actions/pause/commands.js
```diff
@@ -67,59 +45,26 @@ export function stepOver(cx) {
...
- * @returns {Function} {@link command}
- */
-export function reverseStepOver(cx) {
-  return dispatch => {
-    if (cx.isPaused) {
-      return dispatch(command(cx, "reverseStepOver"));
-    }
-  };
+export function rewind(cx) {
+  return dispatch => dispatch(command(cx, "rewind"));
```
jcmorrow, line 69: It seems odd to both ensure that `isPaused` is true and also to delete the place where we check it?

PR #7299, opened 2022-07-01
[WIP] Fix global expression evaluation

jasonLaster:
This is a significantly larger change than is probably necessary, but I like it because it moves the logic into the protocol where it belongs

The simple fix is to use a default value of 0 for asyncIndex which is what we did in the [old code](https://github.com/replayio/devtools/commit/d7e0e0f98ad08226e0b71aed8d5e6b16aca982a1#diff-66115b293bcb7aa787d41d02330aed02e29aaf80c23ffe1f3c4d5fdef6ea43d0L107).

The other changes here are to minimize the touchpoints on thread front

### DIFF
--- a/packages/protocol/thread/value.ts
+++ b/packages/protocol/thread/value.ts
@@ -466,7 +466,8 @@ export class ValueFront {
 Object.setPrototypeOf(ValueFront.prototype, new Proxy({}, DisallowEverythingProxyHandler));
 
 export type PrimitiveValue = string | number | boolean | null | undefined;
-export function createPrimitiveValueFront(value: PrimitiveValue, pause: Pause | null = null) {
+export function createPrimitiveValueFront(value: PrimitiveValue) {
+  const pause = ThreadFront.ensureCurrentPause();
   return new ValueFront(pause, { value });
 }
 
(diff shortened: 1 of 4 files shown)


markerikson (approved):

packages/protocol/thread/value.ts
```diff
@@ -466,7 +466,8 @@ export class ValueFront {
 Object.setPrototypeOf(ValueFront.prototype, new Proxy({}, DisallowEverythingProxyHandler));
 
 export type PrimitiveValue = string | number | boolean | null | undefined;
-export function createPrimitiveValueFront(value: PrimitiveValue, pause: Pause | null = null) {
+export function createPrimitiveValueFront(value: PrimitiveValue) {
+  const pause = ThreadFront.ensureCurrentPause();
   return new ValueFront(pause, { value });
```
jasonLaster, line 471: small cleanup to keep ThreadFront out of the thunks as much as possible

src/devtools/client/webconsole/actions/input.ts
```diff
@@ -143,18 +126,8 @@ export function eagerEvalExpression(expression: string): UIThunkAction {
       return null;
     }
 
-    const selectedFrame = getSelectedFrame(getState());
-    if (!selectedFrame) {
-      return;
-    }
```
jasonLaster, line 149: removing the early returns is sufficient. Everything else is cleanup to remove ThreadFront from as many places as possible

PR #7300, opened 2022-07-01
Fix global evaluations in the terminal

jasonLaster:
A much more modest change compared to #7299 which accomplishes the same goals

### DIFF
--- a/public/test/scripts/console_eval.js
+++ b/public/test/scripts/console_eval.js
@@ -1,5 +1,11 @@
 Test.describe(`Test global console evaluation.`, async () => {
   await Test.selectConsole();
+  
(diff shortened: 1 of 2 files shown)


public/test/scripts/console_eval.js
```diff
@@ -1,5 +1,8 @@
 Test.describe(`Test global console evaluation.`, async () => {
   await Test.selectConsole();
+  await Test.app.actions.seekToTime(1570)
+  await Test.executeInConsole("333");
+
```
jasonLaster, line 5: Adding a small test for global evaluations

jasonLaster:
Merging because #7299 was approved and this is more minor. Also the 2 failures were flakes.

PR #7304, opened 2022-07-01
Fix switching to `Your Library`

jcmorrow:
Otherwise we still send the update request with the ID  `me` which 500's
GraphQL.

Recording of the original issue: https://app.replay.io/recording/switching-to-your-library-generates-500-request--98698b50-8bb8-4a3a-9769-5fda391b201a

### DIFF
--- a/src/ui/components/Library/Navigation/Navigation.tsx
+++ b/src/ui/components/Library/Navigation/Navigation.tsx
@@ -17,8 +17,8 @@ function Teams() {
 
   return (
     <>
-      {workspaces.map((w, i) => (
-        <TeamButton label={w.name} key={i} id={w.id} isTest={w.isTest} />
+      {workspaces.map(w => (
+        <TeamButton label={w.name} key={w.id} id={w.id} isTest={w.isTest} />
       ))}
     </>
   );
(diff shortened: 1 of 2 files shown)


src/ui/components/Library/Navigation/TeamButton.tsx
```diff
@@ -47,11 +47,11 @@ export function TeamButton({
         )}
         onClick={onClick}
       >
-        <span className="overflow-hidden whitespace-pre overflow-ellipsis">
+        <span className="overflow-hidden overflow-ellipsis whitespace-pre">
```
jcmorrow, line 50: These CSS changes were autofixes from `trunk`.

src/ui/components/Library/Navigation/Navigation.tsx
```diff
@@ -17,8 +17,8 @@ function Teams() {
 
   return (
     <>
-      {workspaces.map((w, i) => (
-        <TeamButton label={w.name} key={i} id={w.id} isTest={w.isTest} />
+      {workspaces.map(w => (
+        <TeamButton label={w.name} key={w.id} id={w.id} isTest={w.isTest} />
```
jcmorrow, line 21: This is probably not a big deal, but we should not use the index of the list item as the key. It's unlikely this list will change order, but if it does, using indices as keys will create bugs.

jaril:
Thanks!! Will be changing this route so that we don't run into these kinds of problems but this is a solid fix.

jaril (approved):

PR #7309, opened 2022-07-01
Implement tooling to generate code coverage reports for E2E tests

markerikson:
This PR:

- Updates our E2E test commands to ignore SSL errors due to self-signed certs
- Adds standard Playwright as a dep in the `test` folder
- Updates the test logic to accept an env var that says to use standard Playwright instead
- Adds logic that uses Playwright+Chromium's v8-based code coverage to write coverage data to disk, converts it to Istanbul format, and writes it out as an HTML coverage report
  - Logic is mostly ported from https://github.com/bgotink/playwright-coverage, with adjustments to convert the reporter and page wrapper logic to Jest
- Updates a number of tests that were checking for FF usage only, and tweaks the conditions to accept running with Chromium if E2E coverage mode is enabled

If I've done this right, all our existing E2E tests should continue to work the same way - the altered logic should only kick in if a couple additional env vars are set.

## Usage

This is set up to be run manually, locally, when we want to get some insights on coverage.

Setup roughly looks like:

- Run `yarn` inside our `test` folder to make sure that standard Playwright gets installed, including its Chromium browser
- In a shell, set:

```bash
export DONT_RECORD_TEST=true
export E2E_COVERAGE_DATA=true
export RECORD_REPLAY_TARGET=chromium
```

You may also need to download the driver library (either https://static.replay.io/downloads/macOS-recordreplay.so or https://static.replay.io/downloads/linux-recordreplay.so), drop that binary into the `devtools` folder, and set the paths to the driver and `replay-node`:

```bash
export RECORD_REPLAY_DRIVER="./linux-recordreplay.so"
export RECORD_REPLAY_NODE="/home/$USERNAME/.nvm/versions/node/v16.14.0/bin/replay-node"
```

Once all that's done, run:

```bash
yarn test-e2e some-test-regex-here
```

The matched E2E tests should run, and each test should generate a corresponding `./coverage/testCoverage/$TEST_NAME.coverage.json` file.

When the Jest test run finishes, it should then try to process _all_ coverage JSON files on disk (even ones that were there before this run), merge them together, and output an HTML coverage report in `./coverage/index.html`.

### Notes

Running the tests with standard Playwright Chromium has been kinda flaky, which is part of why I opted to keep all the coverage JSON files on disk and re-merge them all every time.  I found myself running a handful of tests at once, letting the coverage files get written, and building up the number of files available to calculate from as I went along.

Also, something about the merging process is having issues - the Jest process just _dies_ completely when it starts merging around 10+ coverage files.  It gets through most of them, and then exits silently.  It's possible that altering the merging logic would fix this.  Right now it merges all the v8-format files together, then converts it all to Istanbul format.  Perhaps doing the Istanbul conversion per-file first and then merging them would work better.

## Results

I'm generally seeing around 75% coverage by combining results from 8-10 tests together.  That only went up slightly as I added more tests, so this tells me that the bulk of our code always runs on startup, and smaller specific features get covered per test.

image

I'm attaching a zipped copy of the last code coverage report I generated - just unzip and double-click `index.html` to view.

[E2E example code coverage report.zip](https://github.com/replayio/devtools/files/9030721/E2E.example.code.coverage.report.zip)


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     /* Visit https://aka.ms/tsconfig.json to read more about this file */
     /* Basic Options */
     "incremental": true /* Enable incremental compilation */,
-    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "target": "es2015" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
```
bvaughn, line 6: Nice!

bvaughn (approved):
I skimmed it but a little tl;dr 😅 I trust you!

markerikson:
The one failing E2E test is stuck waiting for a screenshot to change, so I'm going to assume that's _not_ related.

PR #7318, opened 2022-07-04
Less flickery loading screen

jonbell-lot23:
We have a bunch of status messages that are better suited for the console. Otherwise they cause a lot of flickery jittery loading experience, which doesn't look great.

Different statuses that we were throwing:
* Starting a session...
* Authenticating...
* Finding recording...
* Loading timeline...

Now:
I've put an indeterminate loader in, like so:

image

--

The "this is taking a while" timer used to trigger after 2 seconds, leading to even more string flicker. I changed it to 9 seconds to be a true edgecase message.

### DIFF
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -5,6 +5,7 @@ import { UIState } from "ui/state";
 import { LoadingTips } from "./LoadingTips";
 import { BubbleViewportWrapper } from "./Viewport";
 import ReplayLogo from "./ReplayLogo";
+import Spinner from "../shared/Spinner";
 
 export function LoadingScreenTemplate({
   children,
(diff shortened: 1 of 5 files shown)


src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -55,8 +56,11 @@ function LoadingScreen({
 
   return (
     <LoadingScreenTemplate showTips={true}>
-      <span>{stalled ? "This is taking longer than usual..." : waitingForMessage}</span>
-    </LoadingScreenTemplate>
+    <div className="flex text-xs">
+      <span>{waitingForMessage}</span>
+      <span className="ml-2">{stalled ? "This is taking longer than usual..." : ""}</span>
```
jcmorrow, line 61: I'm surprised we want a small margin on the left here? Are we aligning it with something?

markerikson:
FWIW I'm with Josh on this one.  If we don't have enough loading work to merit a progress bar, I'd like to see descriptive text.  Maybe add the spinner to that, but keep the "Establishing a session...", etc.

jonbell-lot23:
Closing to propose another approach.

PR #7320, opened 2022-07-04
Small tweaks to loading page

jonbell-lot23:
* Standardised font size
* Changed timeout from 2 seconds to 9


jonbell-lot23:
Pushing live -- it's a tiny change and it's a holiday

PR #7321, opened 2022-07-05
Testsuites CSS alignment and polish

jonbell-lot23:
Our two views were using different styles so they looked different. Now they're aligned.

30s Loom: https://www.loom.com/share/5db51a04492b4c75a02e7cbbb9a9031b


jasonLaster (approved):
👍

PR #7324, opened 2022-07-05
Add hover styles and target new tabs

jonbell-lot23:
* When clicking a Replay from test suites, it opens a new tab (heavy or out of context links should target new tabs)
* We were using both blue and green to mean success, which was inconsistent. Standardised on green.
* Added hover states to the "view replay" play button


jonbell-lot23:
Pushing in order to not get into rebase heck

PR #7326, opened 2022-07-05
Experimental sources reducer

jcmorrow:
Changes
---
- Adds a very simply entity adapter for storing raw protocol `newSource`s
- Adds a more opinionated, detailed store for source data called `SourceDetails` (also has an entity adapter)
- Add sources to the `newSource` entity adapter as they come in from the backend
- Once all sources are in (we can tell because the request has resolved) dispatch one action which kicks off a single call to `newSourcesToCompleteSourceDetails`. Build the simple maps first - `correspondingId`s and `generatedId`s (except for prettyPrinted sources)
- Then fix up canonical references (by type)
- A word on canonical IDs: I’ve made it so that the `canonicalId` for a source is always the non-pretty-printed version. I don’t feel super strongly about this. To me it just makes sense to split the world into `canonical/non-canonical` and `pretty-printed/unpretty-printed`
- Bundled files are still complicated! You can tell if a file is a bundle if it’s `generatedFrom` array has more than a single element. In that case, a sourceId is not enough to jump to a canonical file. You need a location. I think in the future when toggling between original and generated for bundled ID’s we should have some concept of a shallow `location` stack which we can traverse backwards so that we can make sure we are returning to the proper original source.
- We don’t currently link `eval`ed snippets as generated properly, we need to fix this on the backend.
- We don’t currently include `contentHash` for all sourceIds, but I think we should strive to (at the very least for pretty-printed ones, but also maybe for others that might not have it?)
- I’m not sure what an `other` source is or what to do with it
- Generally speaking, when asking questions of a source, like getting the generatedTo or generatedFrom, or getting corresponding ID’s, we should use the canonical source

Next Steps
---
- Use this in some places!

### DIFF
--- a/src/devtools/client/debugger/src/client/index.ts
+++ b/src/devtools/client/debugger/src/client/index.ts
@@ -5,6 +5,7 @@
 import { ThreadFront } from "protocol/thread";
 import { bindActionCreators } from "redux";
 import type { UIStore } from "ui/actions";
+import { addSource, allSourcesReceived } from "ui/reducers/sources";
 
 import actions from "../actions";
 import { initialBreakpointsState } from "../reducers/breakpoints";
@@ -32,8 +33,9 @@ type $FixTypeLater = any;
 
 async function setupDebugger() {
   const sourceInfos: $FixTypeLater[] = [];
-  // @ts-expect-error `sourceMapURL` doesn't exist?
-  await ThreadFront.findSources(({ sourceId, url, sourceMapURL }) =>
+  await ThreadFront.findSources(newSource => {
+    // @ts-expect-error `sourceMapURL` doesn't exist?
+    const { sourceId, url, sourceMapURL } = newSource;
     sourceInfos.push({
       type: "generated",
       data: prepareSourcePayload({
@@ -41,10 +43,15 @@ async function setupDebugger() {
         url,
         sourceMapURL,
       }),
-    })
-  );
+    });
+
+    // Can also probably just shove these into an array right here and dispatch
+    // them all at the end.
+    store.dispatch(addSource(newSource));
+  });
   await store.dispatch(actions.newQueuedSources(sourceInfos));
-  store.dispatch({ type: "SOURCES_LOADED" });
+
+  store.dispatch(allSourcesReceived());
 }
 
 export function bootstrap(_store: UIStore) {
--- a/src/ui/reducers/sources.ts
+++ b/src/ui/reducers/sources.ts
@@ -1,0 +1,71 @@
+import {
+  createEntityAdapter,
+  createSelector,
+  createSlice,
+  EntityState,
+  PayloadAction,
+} from "@reduxjs/toolkit";
+import { newSource, SourceKind } from "@replayio/protocol";
+import { getSelectedSourceId } from "devtools/client/debugger/src/selectors";
+import { UIState } from "ui/state";
+import { newSourcesToCompleteSourceDetails } from "ui/utils/sources";
+
+export interface SourceDetails {
+  canonicalId: string;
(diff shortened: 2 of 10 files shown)


src/ui/setup/store.ts
```diff
@@ -42,6 +43,7 @@ let reducers = {
   layout: layoutReducer,
   messages: messagesReducer,
   protocolMessages: protocolMessages,
+  newSources: sources,
```
markerikson, line 46: :speech_balloon: heh, I _think_ this is meant to be "this is the new sources reducer, different than the existing `sources` reducer", but it's amusing that we have the type name `newSource` from the backend API :)

src/ui/utils/graph.ts
```diff
@@ -0,0 +1,37 @@
+export interface Graph {
+  addNode(node: string): void;
+  connectNode(from: string, to: string): void;
+  from: (node: string) => string[];
+  to: (node: string) => string[];
+}
+
+const newGraph = (name: string = "Unnamed"): Graph & { inspect: () => any } => {
```
markerikson, line 8: :speech_balloon: very minimal - I like it!

src/ui/utils/sources.test.ts
```diff
@@ -0,0 +1,292 @@
+import { newSourcesToCompleteSourceDetails } from "./sources";
+
+describe("newSourcesToCompleteSourceDetails", () => {
```
markerikson, line 3: :speech_balloon: yay tests!

src/ui/utils/sources.ts
```diff
@@ -0,0 +1,107 @@
...
+      current = nextNode;
+      nextNode = canonical.from(current)?.[0];
+    }
+
+    return current;
+  };
+
+  const generated = newGraph("generated");
+  newSources.forEach((source: newSource) => {
+    if (corresponding[keyForSource(source)] === undefined) {
```
markerikson, line 54: :hand: tiny nit - can we just do the `keyForSource(source)` lookup once?

markerikson (approved):
I love this! Let's get it in now.

src/ui/reducers/sources.ts
```diff
@@ -0,0 +1,71 @@
...
+  PayloadAction,
+} from "@reduxjs/toolkit";
+import { newSource, SourceKind } from "@replayio/protocol";
+import { getSelectedSourceId } from "devtools/client/debugger/src/selectors";
+import { UIState } from "ui/state";
+import { newSourcesToCompleteSourceDetails } from "ui/utils/sources";
+
+export interface SourceDetails {
+  canonicalId: string;
+  contentHash: string | undefined;
```
markerikson, line 15: :hand: All these `| undefined` fields could probably be `field?: whatever` instead

src/devtools/client/debugger/src/client/index.ts
```diff
@@ -32,19 +33,27 @@ type $FixTypeLater = any;
...
+
+    // Can also probably just shove these into an array right here and dispatch
+    // them all at the end.
+    store.dispatch(addSource(newSource));
+  });
   await store.dispatch(actions.newQueuedSources(sourceInfos));
   store.dispatch({ type: "SOURCES_LOADED" });
+
+  // This fires off the coercing of sources into a more workable form
+  store.dispatch(allSourcesReceived());
```
markerikson, line 56: :hand: We could key off the existing `"SOURCES_LOADED"` action instead

jcmorrow: I went slightly the other way and changed the place that was listening for that action to listen for `sources/allSourcesReceived` instead, but same effect 👍

PR #7327, opened 2022-07-05
Make test run view load faster

jaril:
Backend piece is in: https://github.com/replayio/backend/pull/6060


jasonLaster (approved):
Love how this simplifies things too

PR #7330, opened 2022-07-05
Migrate more sources/pause files to TS

markerikson:
This PR:

- Migrates all files under `actions/{pause|sources|breakpoints}` to TS
- Removes a bunch of selector typedefs in `selectors/index.ts` that were masking the _real_ definitions now that those selectors really are in TS already
- Removes several `.d.ts` files that were already dead because their corresponding sources were previously converted to TS

Note that the one remaining `getContext` typedef causes errors if I remove it - there's a mismatch between the type of `state.pause.cx` and how that's being passed around.  That needs to be sorted out at some point.


markerikson:
PR now has 3 test failures, but I don't think any of them are "real":

- Mock tests failed with an Apollo "can't find the recording" error
- One E2E shard succeeded, but marked as failed afterwards due to a GH actions issue
- GraphQL types failed again

src/devtools/client/debugger/src/components/PrimaryPanes/index.tsx
```diff
@@ -60,7 +60,6 @@ const mapStateToProps = (state: UIState) => {
   return {
     cx: getContext(state),
     selectedTab: getSelectedPrimaryPaneTab(state),
-    sourceSearchOn: getActiveSearch(state) === "source",
```
markerikson, line 63: :speech_balloon: TS pointed out that a comparison with this field's possible values would never match, and it looks like this prop isn't being used any more

src/devtools/client/webconsole/actions/index.ts
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export { toggleTimestamps } from "devtools/client/webconsole/reducers/ui";
```
markerikson, line 5: :speech_balloon: If you compare vs the old `index.js` file, I pretty heavily rewrote the import/export syntax here. In theory that could cause problems, but I checked for the places that did actually reference this file and updated them.

jcmorrow (approved):
I've played with the preview branch and read through the changes, they seem ok although it's hard to be 100% sure. But I'm pretty comfortable merging this and then just keeping our 👀 peeled for things that look suspicious.

src/devtools/client/debugger/src/actions/pause/commands.ts
```diff
@@ -2,69 +2,74 @@
...
+export function stepIn(cx: Context): UIThunkAction {
   return dispatch => {
     if (cx.isPaused) {
       return dispatch(command(cx, "stepIn"));
     }
   };
 }
 
-export function stepOver(cx) {
+export function stepOver(cx: Context): UIThunkAction {
```
bvaughn, line 45: I wonder why `e` and `cx` are so popular in our codebase

PR #7333, opened 2022-07-06
add test info to replay info panels

jasonLaster:
Small tweak to make it easier to see which test you're looking at
* show test name in title
* update info panel
* hide trial date 

### before
Screen Shot 2022-07-05 at 5 48 21 PM

### after
Screen Shot 2022-07-05 at 5 47 57 PM

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -49,6 +49,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
     setModal("privacy");
   };
 
+  const isTest = recording.metadata.test;
   return (
     <div className="flex-column flex flex items-center overflow-hidden border-splitter bg-bodyBgcolor">
       <div className="my-1.5 flex w-full cursor-default flex-col self-stretch overflow-hidden px-1.5 pb-0 text-xs">
@@ -73,7 +74,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
           ) : null}
         </div>
         <div className="group">
-          {recording.url ? (
+          {!isTest && recording.url ? (
             <Row>
               <Icon
                 filename="external"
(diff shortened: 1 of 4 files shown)


src/ui/components/Header/Header.tsx
```diff
@@ -122,6 +122,11 @@ function HeaderTitle({
     }
   }, [editing, hasTitle, recording.title]);
 
+  const testName = recording.metadata?.test?.title;
```
jaril, line 125: Let's not do this. I don't like how this introduces a different concept of title based on what kind of replay you're looking at.

I agree that `Replay of devtools--3vku99qej` is **not** what you want to see on that upper left side. But we should fix that on the test adapter side — it should have been saving the title on these replays as "Replay of <insert-test-name-here>".

src/ui/components/Header/RecordingTrialEnd.tsx
```diff
@@ -8,7 +8,15 @@ export function RecordingTrialEnd() {
...
-  if (loading || !workspace || !inUnpaidFreeTrial(workspace) || workspace.hasPaymentMethod) {
+  const metadata = recording?.metadata;
+
+  if (
+    loading ||
+    !workspace ||
+    metadata?.test ||
+    !inUnpaidFreeTrial(workspace) ||
+    workspace.hasPaymentMethod
+  ) {
```
jaril, line 19: Let's change this to check the workspace instead of the recording. So don't show the trial if it's a test workspace, i.e. `workspace.isTest === true`.

src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -90,6 +91,40 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
             </Row>
           ) : null}
         </div>
+        <div className="group">
```
jaril, line 94: Will make sure to kick off a discussion tomorrow around how the experience of debugging a replay should look like if you're specifically debugging replay of a test. For now, this should do the trick.

jaril (changes requested):

src/ui/components/Header/Header.tsx
jasonLaster: Happy to remove this IF statement when it is obsolete, but for now i think we should use this work around

jaril:
Continuing in https://github.com/replayio/devtools/pull/7359

PR #7337, opened 2022-07-06
Fix the performance of the experimental sources reducer

hbenl:
Fixes FE-385


bvaughn:
I think this commit broke `main` for @markerikson and I. Every recording I try to open now throws:
Screen Shot 2022-07-06 at 5 03 10 PM
.

hbenl:
@bvaughn You'll have to add `NEXT_PUBLIC_API_SUBSCRIPTION_URL=wss://api.replay.io/v1/graphql` to your `.env` file.

PR #7338, opened 2022-07-06
Architecture: Updating test fixture data

bvaughn:
Loom overview: https://www.loom.com/share/eb0ec3c931e64611aa492083ed1e237f

- [x] Add automated process for updating (or re-generating) test fixture data.


packages/shared/client/createReplayClientRecorder.ts
```diff
@@ -37,16 +63,22 @@ export default function createReplayClientRecorder(
...
-        logEntries.push(entry);
+        if (findMatch(logEntries, prop, args) === null) {
+          logEntries.push(entry);
+        }
 
         // Unwrap Promise values
         if (result != null && typeof result.then === "function") {
           entry.isAsync = true;
 
+          // flagPendingClientRequest();
```
markerikson, line 74: :hand: dead code?

markerikson (approved):
Seems good enough :)

packages/shared/client/createReplayClientRecorder.ts
bvaughn: I'm not _totally_ convinced we won't need it so for now I've just commented it out.

PR #7348, opened 2022-07-07
New console

bvaughn:
- [x] Fix edge case in `Wakeable` implementation (and add tests)
- [x] Fetch and display event types filters beside the Console
- [ ] Wire up event type filters to add logs to console


bvaughn:
I have one more chunk of wiring up to do for the event-type toggles but I'm going to tackle that in a separate PR.

PR #7354, opened 2022-07-08
Notify attached callbacks from within `Wakeable#then` in case of a settled wakeable

Andarist:
Based on my investigation and [the drawn conclusions](https://github.com/facebook/react/issues/24864#issuecomment-1178716544) I think this should resolve the problem with some suspended components not being correctly resumed sometimes.


bvaughn:
Cleaning up old/sale PRs. This part of the codebase no longer exists 😄

bvaughn:
Sorry I missed this PR initially though FWIW. Don't even remember seeing it!

PR #7356, opened 2022-07-08
Fixed how `thrownValue` is being rethrown in `Wakeable#resolve`/`Wakeable#reject`

Andarist:
The current implementation is almost synonymous with not having a catch block at all - it just throws immediately after calling a single callback (it ignores thrown nulls but that's probably irrelevant here).

I think that my patch matches closer the original intention - we call all callbacks and only throw the last thrown value after we are done with the callbacks.


bvaughn:
Cleaning up old/sale PRs. This part of the codebase no longer exists, so~

PR #7360, opened 2022-07-08
Fix contentHashes for pretty-printed sources

jcmorrow:
This is an adaption of the logic which currently lives at
https://github.com/replayio/devtools/commit/3aff939984#diff-f4d7f2c6fc1b8c391f96cdd4964dd6009ad0ff580a6f602c99f6e83f18a455a2R1312-R1334
to the new sources adapter.


jcmorrow:
@markerikson I'm not sure how much copying and pasting you have to do when I update stuff in here, but just FYI 😄

markerikson (approved):

PR #7363, opened 2022-07-08
Fix result encoding for test workflow

ryanjduffy:


ceceliacreates (approved):

ryanjduffy:
Ugh ... created from a fork so the comments didn't work. Recreating

PR #7369, opened 2022-07-09
Small speedup so that we no longer wait on sources to load

jasonLaster:
Was looking into why the video sometimes takes awhile to initially load and found this small perf win.

~My current theory is that we try to fetch the graphic before the region the pause is in has loaded and fail. This would explain why if you click somewhere else you do see the video.~ Josh, reminded me that the backend will block until the region is loaded, so that's probably not it, but we could look at the session logs and see if they give us reasons why a graphics call might otherwise fail

### DIFF
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -246,14 +246,7 @@ export async function repaint(force = false) {
   }, 500);
 
   const { mouse } = await getGraphicsAtTime(ThreadFront.currentTime);
-  const point = ThreadFront.currentPoint;
-  await ThreadFront.ensureAllSources();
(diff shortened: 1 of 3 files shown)


jaril:
image

hbenl (changes requested):
This is unsafe: `ThreadFront.ensureCurrentPause()` will ultimately trigger `Pause.addData()` which ultimately triggers `ThreadFront.getCorrespondingSourceIds()` which asserts that all sources have been loaded.
But there's no reason to wait for sources before painting graphics, we just need to do the `await ThreadFront.ensureAllSources()` in the `createWaiter` (before `this.addData(...)`.

hbenl (approved):

PR #7370, opened 2022-07-09
Automatically parameterize the focus window

jcmorrow:
- and restore it when loading

https://www.loom.com/share/52e1b04ccd724865b2373c48ab9b3263

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/paused.ts
+++ b/src/devtools/client/debugger/src/actions/pause/paused.ts
@@ -11,7 +11,10 @@ import { getSelectedFrame, getThreadContext, getSelectedLocation } from "../../s
 import { fetchScopes } from "./fetchScopes";
 import { setFramePositions } from "./setFramePositions";
 import { trackEvent } from "ui/utils/telemetry";
-import { isPointInLoadedRegion } from "ui/reducers/app";
+import { isPointInLoadingRegion } from "ui/reducers/app";
+import { setFocusRegion } from "ui/actions/timeline";
+import { TimeStampedPoint } from "@replayio/protocol";
+import maxBy from "lodash/maxBy";
 
 // How many times to fetch an async set of parent frames.
 const MaxAsyncFrames = 5;
@@ -79,8 +82,8 @@ export function paused({
   return async function (dispatch, getState, { ThreadFront }) {
     dispatch(pauseRequestedAt(executionPoint));
 
-    if (!isPointInLoadedRegion(getState(), executionPoint)) {
-      return;
+    if (!isPointInLoadingRegion(getState(), executionPoint)) {
+      dispatch(failedToCreatePause(executionPoint, ""));
     }
 
     trackEvent("paused");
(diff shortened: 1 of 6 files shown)


src/devtools/client/debugger/src/actions/pause/paused.ts
```diff
@@ -79,8 +82,8 @@ export function paused({
   return async function (dispatch, getState, { ThreadFront }) {
     dispatch(pauseRequestedAt(executionPoint));
 
-    if (!isPointInLoadedRegion(getState(), executionPoint)) {
-      return;
+    if (!isPointInLoadingRegion(getState(), executionPoint)) {
+      dispatch(failedToCreatePause(executionPoint, ""));
```
jcmorrow, line 86: If we've ended up in a state where we are truly not going to load this point, we should not leave it marked as "loading"

src/ui/actions/session.ts
```diff
@@ -270,8 +269,6 @@ function onLoadingFinished(): UIThunkAction {
       await ThreadFront.waitForSession();
       await ThreadFront.initializedWaiter.promise;
       await ThreadFront.ensureAllSources();
-
-      ThreadFront.ensureCurrentPause();
```
jcmorrow, line 274: AFAICT, this is not needed, since we `warp` in `jumpToIniitalPausePoint`? Also, it was getting called before we were done wrangling the focus window, which was causing more `failed to create pause, region has been unloaded` errors.

src/ui/utils/environment.ts
```diff
@@ -128,7 +131,9 @@ export function updateUrlWithParams(params: Record<string, string>) {
   const url = new URL(window.location.toString());
 
   Object.entries(params).forEach(([key, value]) => {
-    url.searchParams.set(key, value);
+    if (value !== undefined && value !== null) {
```
jcmorrow, line 134: It's important to get this check right. `point=0` is a valid point, even though `0` is falsey. But we (almost certainly) never want to put the values `undefined` or `null` directly into a URL.

jcmorrow (commented):
Forgive my raspy loom! My throat has been struggling today

jcmorrow:
Welp, looks like I have displeased the E2E gods

src/ui/actions/timeline.ts
```diff
@@ -184,12 +200,22 @@ function updatePausePointParams({
...
   time: number;
   hasFrames: boolean;
+  focusRegion: FocusRegion | null;
 }) {
-  const params = { point, time: `${time}`, hasFrames: `${hasFrames}` };
+  const params: { point: string; time: string; hasFrames: string; focusRegion?: string } = {
+    point,
+    time: `${time}`,
+    hasFrames: `${hasFrames}`,
+    focusRegion: undefined,
```
jaril, line 214: I don't think it's the time to sound alarms yet, but we probably should take a look at how/if it's worth simplifying this when we get to parameter #5

jcmorrow: 🤷‍♂️

markerikson (approved):

PR #7371, opened 2022-07-09
Run `trunk check` and accept autofixes

jcmorrow:
Hold the line is generally great, but I've been getting a lot of warnings from trunk *locally* for these little issues, so I figured it would be worth just fixing all of the autofixable ones at once.


jasonLaster:
Smart!

PR #7372, opened 2022-07-09
Do not ignore mixpanel types

jcmorrow:
We added mixpanel types so that we would get some extra tsc assistance on things like spelling, deduplicating, etc. If we find that we are regularly ignoring these instead of adding type definitions, then it might be time to rip the types out. Personally, I think they add something, so I opted to keep them.


markerikson:
Yeah, I think I threw that in some of my early TS conversion work, but have since been adding them to the union type.

PR #7387, opened 2022-07-13
Trunk fixups and upgrade

jcmorrow:
Upgrade trunk and fix up the couple of issues that VSCode is always complaining about.


jazzdan (approved):

test/src/runTest.js
```diff
@@ -146,7 +146,7 @@ async function runTest(state, test, exampleRecordingId, target, saveFixture) {
     testPath += "&longTimeout=1";
   }
 
-  let success, why, recordingId;
+  let, recordingId, success, why;
```
jaril, line 149: That's not right is it?

jaril (approved):

test/src/runTest.js
jcmorrow: lol, thank *you*

PR #7380, opened 2022-07-13
Port breakable lines and breakpoint positions over to new sources infrastructure

jcmorrow:
Another chip off the ole https://github.com/replayio/devtools/pull/7375.

The Location shape munging that we do could *really* benefit from :

- Strong types
- Utility functions to convert between them
- Tests for those utility functions

If we had those things, it probably would have saved me several hours over the course of the past 2 days alone.

PR #7388, opened 2022-07-13
Fixup of location types

jcmorrow:
It can be really painful to make changes in places that expect Locations
in the codebase right now, because we have several different type
definitions for a Location. I've tried to cut down on the number of
places where we use definitions other than the Protocol's definition.


src/devtools/client/debugger/src/actions/breakpoints/modify.ts
```diff
@@ -190,39 +187,11 @@ export function _removeBreakpoint(
   };
 }
 
-export function removeBreakpointAtGeneratedLocation(
```
jcmorrow, line 193: Seems like this was unused.

src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -63,28 +63,21 @@ function checkSelectedSource(cx: Context, sourceId: string): UIThunkAction {
...
 
     const source = getSource(state, sourceId);
 
-    if (!source || !source.url) {
+    if (!source) {
       return;
     }
 
-    const pendingUrl = pendingLocation.url;
-    const rawPendingUrl = getRawSourceURL(pendingUrl);
```
jcmorrow, line 77: AFAICT this no longer does anything because our URL's don't have `:formatted` appended to the end? (AFAIK)

src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -63,28 +63,21 @@ function checkSelectedSource(cx: Context, sourceId: string): UIThunkAction {
...
-
-    if (rawPendingUrl === source.url) {
+    // Source IDs are not stable across Replays (or even across Replay sessions)
+    // so we compare via URL
+    if (pendingLocation.url === source.url) {
```
jcmorrow, line 78: This is the kind of place where `stableSourceId` would be nice. In truth, you can't know if a given file with the same URL actually contains this location. It could have totally different content, or it might be pretty-printed vs. not. If we had `kind` and `contentHash` as well, we would have true guarantees.

src/devtools/client/debugger/src/reducers/types.ts
```diff
@@ -195,9 +172,9 @@ export type BreakpointResult = {
  * @static
  */
 export type PendingBreakpoint = {
-  readonly location: PendingLocation;
+  readonly location: Location & { sourceUrl?: string };
```
jcmorrow, line 175: This type gets used enough that I thought about pulling out a name for it, but I decided against it for now.

src/devtools/client/debugger/src/reducers/sources.ts
```diff
@@ -74,7 +66,10 @@ export interface SourcesState {
   breakableLines: Record<string, number[]>;
   epoch: number;
   selectedLocation?: Location | null;
-  pendingSelectedLocation?: Location;
+  // Source IDs are not stable across Replays (or even across Replay sessions)
+  // so we also add a URL, to be sure that we can identify this source the next
+  // time we see it.
+  pendingSelectedLocation?: Location & { url: string };
```
jcmorrow, line 72: This type is *almost* a SourceLocation, but not quite.

markerikson (approved):

jcmorrow:
I ended up having to roll this into #7380 , because both changes alone broke some subtle things like pending selected location and pending breakpoints.

PR #7389, opened 2022-07-14
Fix analysis edge case

jcmorrow:
This is a super weird one. I still don't 100% understand how this particular bug resulted in this exact UI error, but the fix makes complete sense to me, and the bug seems to be gone.


jcmorrow:
@markerikson mind-bending to read this analysis code again 😆

jasonLaster (approved):
Nice

PR #7390, opened 2022-07-14
use pause.isLoaded() instead of ensureAllSources

jasonLaster:
I'm not sure if this is a good idea, but I think it could be a potential follow on cleanup to #7369 because we now know that we can reliably load a pause after sources are available

### DIFF
--- a/packages/protocol/thread/pause.ts
+++ b/packages/protocol/thread/pause.ts
@@ -190,6 +190,10 @@ export class Pause {
     datas.forEach(d => this._updateDataFronts(d));
   }
 
+  ensureLoaded() {
+    return this.createWaiter;
+  }
+
   // we're cheating Typescript here: the objects that we add are of type Frame/Scope/ObjectDescription
   // but we pretend they are of type WiredFrame/WiredScope/WiredObject. These objects will be changed
   // in _updateDataFronts() so that they have the correct type afterwards.
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -626,7 +626,7 @@ class _ThreadFront {
     return pause;
   }
 
-  ensureCurrentPause() {
+  getCurrentPause() {
     if (!this.currentPause) {
       this.currentPause = this.ensurePause(this.currentPoint, this.currentTime);
     }
@@ -655,16 +655,13 @@ class _ThreadFront {
       return [];
     }
 
-    await this.ensureAllSources();
-    this.ensureCurrentPause();
-    return await this.currentPause!.getFrames();
+    return await this.getCurrentPause().getFrames();
   }
 
   lastAsyncPause() {
-    this.ensureCurrentPause();
     return this.asyncPauses.length
       ? this.asyncPauses[this.asyncPauses.length - 1]
-      : this.currentPause;
+      : this.getCurrentPause();
   }
 
   async loadRegion(region: TimeRange, endTime: number) {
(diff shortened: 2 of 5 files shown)


jcmorrow:
If it's not too much of a hassle, I'd like to get confirmation from @bhackett1024 that https://linear.app/replay/issue/FE-341/replay-not-pausing-at-point-in-url-params#comment-7f568699 is or is not still happening before we keep messing with initial pause stuff. Getting this stuff right is hard, but getting it right when we're all trying different stuff at the same time feels *extra* hard. What do you think?

packages/protocol/thread/thread.ts
```diff
@@ -662,7 +659,7 @@ class _ThreadFront {
   }
 
   async loadAsyncParentFrames() {
-    await this.ensureAllSources();
+    await this.getCurrentPause().isLoaded();
```
jcmorrow, line 662: I will say, if we're going to call `get/ensureCurrentPause` I much prefer getting this Promise back. Otherwise, we call it but it's not async and it doesn't actually ensure anything (which I'm guessing is why the name changed 😉)

jcmorrow (commented):
Actually, after reading through it, I don't think it will interfere with the other changes too much. The one thing I don't know about is what happens if we call `ThreadFront.getCurrentPause` before we warp to the correct point in this line https://github.com/replayio/devtools/pull/7370/files#diff-0b3b6041213ad0d8eec8fbfb5111600076ab4ab1178b839e802d76a4e8e44689L118

packages/protocol/thread/thread.ts
```diff
@@ -960,42 +953,36 @@ class _ThreadFront {
...
     }
-    const pause = this.currentPause;
-    await this.ensureAllSources();
-    this.ensureCurrentPause();
-    const nodeBounds = await this.currentPause!.getMouseTarget(x, y, nodeIds);
-    return pause == this.currentPause ? nodeBounds : null;
+    const pause = this.getCurrentPause();
+    await pause.isLoaded();
+    const nodeBounds = await pause.getMouseTarget(x, y, nodeIds);
+    return pause == pause ? nodeBounds : null;
```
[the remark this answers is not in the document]
hbenl: This should be `pause === this.currentPause`

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -127,7 +127,7 @@ class ReplayWall implements Wall {
         case "startInspectingNative": {
           this.initializePicker();
 
-          ThreadFront.ensureCurrentPause();
+          ThreadFront.getCurrentPause();
           await ThreadFront.currentPause!.createWaiter;
```
hbenl, line 131: `await ThreadFront.getCurrentPause().isLoaded();`

packages/protocol/thread/pause.ts
```diff
@@ -305,7 +309,6 @@ export class Pause {
   }
 
   async getFrames() {
-    assert(this.createWaiter, "no createWaiter");
```
hbenl, line 308: I'd like to keep these assertions

src/devtools/client/inspector/markup/markup.ts
```diff
@@ -96,7 +96,7 @@ class MarkupView {
     this.isLoadingPostponed = false;
 
     await ThreadFront.ensureAllSources();
-    ThreadFront.ensureCurrentPause();
+    ThreadFront.getCurrentPause();
     const pause = ThreadFront.currentPause;
```
hbenl, line 100: `const pause = ThreadFront.getCurrentPause();`

packages/protocol/thread/pause.ts
```diff
@@ -190,6 +190,10 @@ export class Pause {
     datas.forEach(d => this._updateDataFronts(d));
   }
 
+  isLoaded() {
```
hbenl, line 193: Can we change the name of this method? `isLoaded` suggests it checks if the pause is loaded and returns a boolean, so someone is bound to do `if (pause.isLoaded) {...}` eventually.

hbenl (approved):

PR #7399, opened 2022-07-16
Memoize possible breakpoints

jasonLaster:
Not the most elegant solution, but avoids recomputing `getBreakableLinesForSource` tons of times.


jasonLaster:
landing this fix because DevTools is unusable with it

PR #7401, opened 2022-07-17
Fix Hit Counts for selected source FE-430

jasonLaster:
Fixes an issue we were seeing after the new hitCounts reducer where we weren't including all the hitcounts for a selected source.


markerikson (approved):
Seems like it could work. We may want to put some more thought into how we store and fetch these.

jcmorrow:
Pretty sure this was resolved in one of the other PR's from Sunday

PR #7402, opened 2022-07-17
Use a memoized selector for fetching breakable lines

jcmorrow:
- I wouldn't mind storing these in the store just to avoid potential
  cache misses, but this also works
- For hitCounts, we were not sending the line number to the selector, so
  it was not always grabbing the right line segments. We should make a
  selector that merges those, but again, this works for now.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
@@ -13,7 +13,7 @@ import { shouldShowNag } from "ui/utils/user";
 import { getSelectedSource } from "../../reducers/sources";
 
 import StaticTooltip from "./StaticTooltip";
-import { fetchHitCounts, getHitCountsForSelectedSource } from "ui/reducers/hitCounts";
+import { fetchHitCounts, getHitCountsForSource } from "ui/reducers/hitCounts";
 
 export const AWESOME_BACKGROUND = `linear-gradient(116.71deg, #FF2F86 21.74%, #EC275D 83.58%), linear-gradient(133.71deg, #01ACFD 3.31%, #F155FF 106.39%, #F477F8 157.93%, #F33685 212.38%), #007AFF`;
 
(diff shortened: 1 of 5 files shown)


src/ui/reducers/possibleBreakpoints.ts
```diff
@@ -120,17 +125,21 @@ export const getPossibleBreakpointsForSelectedSource = (state: UIState): Locatio
...
+  (possibleBreakpointsSlice, sourceId: string) => {
+    const possibleBreakpoints = possibleBreakpointsSlice.entities[sourceId];
+    if (!possibleBreakpoints?.possibleBreakpoints) {
+      return null;
+    }
+    return computeBreakableLines(possibleBreakpoints.possibleBreakpoints);
+  }
+);
 
-const computeBreakableLines = memoizeLast((possibleBreakpoints: Location[]) => {
```
jcmorrow, line 131: Gonna be honest - I don't understand why this was not working.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -65,10 +65,12 @@ function LineNumberTooltip({ editor, keyModifiers }: Props) {
   const [targetNode, setTargetNode] = useState<HTMLElement | null>(null);
   const lastHoveredLineNumber = useRef<number | null>(null);
   const isMetaActive = keyModifiers.meta;
-
-  const hitCounts = useAppSelector(getHitCountsForSelectedSource);
   const source = useAppSelector(getSelectedSource);
 
+  const hitCounts = useAppSelector(state =>
+    getHitCountsForSource(state, source!.id, lastHoveredLineNumber.current || 0)
```
markerikson, line 71: :speech_balloon: The React team would probably yell at us for reading from a ref while rendering, but 🤷‍♂️

src/ui/reducers/possibleBreakpoints.ts
```diff
@@ -120,17 +125,21 @@ export const getPossibleBreakpointsForSelectedSource = (state: UIState): Locatio
...
-    adapterSelectors
-      .selectById(state, sourceId)
-      ?.possibleBreakpoints?.map(location => location.line)
-  );
-};
+export const getBreakableLinesForSource = createSelector(
+  (state: UIState) => state.possibleBreakpoints,
+  (_state: UIState, sourceId: string): string => sourceId,
+  (possibleBreakpointsSlice, sourceId: string) => {
+    const possibleBreakpoints = possibleBreakpointsSlice.entities[sourceId];
```
markerikson, line 132: :hand: realistically we could actually move this line into line 129 - we have both `state` and `sourceId` available.

We _could_ also give this selector a larger cache size, although I'm not sure if it's absolutely necessary given that we really just use this for the selected source.

markerikson (approved):

PR #7403, opened 2022-07-17
Aggregate hitCounts for a source and focus region

jcmorrow:
And also don't compute breakable lines so much

### DIFF
--- a/src/ui/reducers/possibleBreakpoints.ts
+++ b/src/ui/reducers/possibleBreakpoints.ts
@@ -25,6 +25,7 @@ export interface PossibleBreakpoints {
   error?: string;
   id: string;
   possibleBreakpoints?: Location[];
+  breakableLines?: number[];
   status: LoadingStatus;
 }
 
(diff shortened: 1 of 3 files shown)


packages/protocol/thread/possibleBreakpoints.test.ts
```diff
@@ -1,3 +1,5 @@
+const { performance } = require("perf_hooks");
```
markerikson, line 1: :hand: Accidental import, should be removed

src/ui/reducers/possibleBreakpoints.ts
```diff
@@ -125,37 +127,20 @@ export const getPossibleBreakpointsForSelectedSource = (state: UIState): Locatio
...
 
 export const getBreakableLinesForSelectedSource = (state: UIState) => {
   const sourceId = state.sources.selectedLocation?.sourceId;
   if (!sourceId) {
-    return null;
-  }
-  const possibleBreakpoints = adapterSelectors.selectById(state, sourceId)?.possibleBreakpoints;
-  if (!possibleBreakpoints) {
-    return null;
+    return [];
```
markerikson, line 141: :hand: Move this out so it's a constant declared _outside_ the selector, like `const NO_BREAKPOINTS = []`, and then do `return NO_BREAKPOINTS` inside - we actually saw this same problem causing extra re-renders yesterday.

markerikson (changes requested):

markerikson (approved):

PR #7409, opened 2022-07-19
Speed up Quick Open FE-145 FE-199

jasonLaster:
Fixes a couple of issues FE-145 FE-199

1. show sources loading
2. memoize format sources
3. improve debounce
4. dont fetch global functions unecessarily

### DIFF
--- a/src/devtools/client/debugger/src/actions/quick-open.ts
+++ b/src/devtools/client/debugger/src/actions/quick-open.ts
@@ -11,8 +11,6 @@ import {
   closeQuickOpen,
   getQuickOpenEnabled,
 } from "../reducers/quick-open";
-import { setViewMode } from "ui/actions/layout";
-import { getViewMode } from "ui/reducers/layout";
 
 // TODO This is a workaround solely to allow dispatching without errors.
 // Additionally, the Redux team recommends against action type unions.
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/components/QuickOpenModal.tsx
```diff
@@ -508,4 +479,40 @@ export class QuickOpenModal extends Component<PropsFromRedux, QOMState> {
   }
 }
 
+function mapStateToProps(state: UIState) {
+  const selectedSource = getSelectedSource(state)!;
+  const tabs = getTabs(state);
+  console.log(`symbols loading: ${isGlobalFunctionsLoading(state)}`);
```
markerikson, line 485: :hand: debug logging

src/devtools/client/debugger/src/actions/quick-open.ts
```diff
@@ -26,7 +24,9 @@ export { setQuickOpenQuery, closeQuickOpen };
 
 export function openQuickOpen(query = "", project = false): UIThunkAction {
   return dispatch => {
-    dispatch(loadGlobalFunctions());
+    if (project) {
```
jcmorrow, line 27: Can we also change this variable name to `functionSearch` or something else that reflects why we would pass this flag?

jcmorrow (approved):

src/devtools/client/debugger/src/actions/quick-open.ts
jcmorrow: Nevermind! This is differentiating between global function search and project function search, so the name kind of makes sense 🤷

PR #7411, opened 2022-07-19
Kill symbols

jasonLaster:
This removes the frontend logic for parsing files and creating symbols.

It does remove some functionality
* classes
* hasTypes
* hasJSX

but i think it's worth it in favor of standardizing on the backend and removing the frontend complexity

Fix FE-446.


markerikson:
Superseded by #7469 .

Note that dropping Babel ends up breaking too many parts of the existing logic, so #7469 just rewrites the thunks and reducer logic around doing parsing and saving results.

If/when we improve the backend to provide additional details like function enclosure line ranges, we can look again at dropping Babel.

PR #7418, opened 2022-07-19
New Console experimental integration

bvaughn:
Loom overview:
https://www.loom.com/share/e0ca04227b364b50856876162d3cd3f3

- [x] Track down "_This part of the recording has been unloaded_" error that sometimes gets thrown by `getPointNearTime` if you slide the focus window around. For some reason, the new architecture test app never triggers this error but the legacy devtools _does_ when the new Console feature is enabled. Why?

### DIFF
--- a/src/ui/components/SecondaryToolbox/NewConsole.tsx
+++ b/src/ui/components/SecondaryToolbox/NewConsole.tsx
@@ -1,0 +1,388 @@
+import {
+  ExecutionPoint,
+  MappedLocation,
+  PauseId,
+  Value as ProtocolValue,
+} from "@replayio/protocol";
+import NewConsole from "bvaughn-architecture-demo/components/console";
+import { SearchContext } from "bvaughn-architecture-demo/components/console/SearchContext";
+import { FocusContext } from "bvaughn-architecture-demo/src/contexts/FocusContext";
+import { InspectorContext } from "@bvaughn/src/contexts/InspectorContext";
+import { Point, PointsContext } from "bvaughn-architecture-demo/src/contexts/PointsContext";
+import { TerminalContext, TerminalExpression } from "@bvaughn/src/contexts/TerminalContext";
+import {
+  TimelineContext,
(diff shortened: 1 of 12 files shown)


src/ui/components/SecondaryToolbox/NewConsole.tsx
```diff
@@ -0,0 +1,388 @@
...
+        return region.begin.time <= focusBegin && region.end.time >= focusEnd;
+      });
+      if (isLoaded) {
+        startTransition(() => {
+          setDeferredFocusRegion(focusRegion);
+        });
+      } else {
+        timeoutId = setTimeout(updateFocusRegionOnceLoaded, 250);
+      }
+    };
```
bvaughn, line 146: This is kind of unfortunate, but without it– the new Console might _suspend_ while asking the backend for the nearest point to a time and get an "unloaded region" error back.

src/ui/components/SecondaryToolbox/NewConsole.tsx
```diff
@@ -0,0 +1,388 @@
...
+  const executionPoint = useAppSelector(getCurrentPoint);
+
+  const update = useCallback(
+    (time: number, executionPoint: ExecutionPoint, pauseId: PauseId) => {
+      dispatch(seek(executionPoint, time, false /* hasFrames */, pauseId));
+    },
+    [dispatch]
+  );
+
+  useLayoutEffect(() => {
```
jaril, line 362: What's the reason for using `useLayoutEffect` here?

jaril (approved):

src/ui/components/SecondaryToolbox/NewConsole.tsx
bvaughn: It's probably not strictly _necessary_ but it avoids a browser paint in between where the new Console and the legacy app might reflect different points. (React flushes layout effects, and any scheduled updates, synchronously– before returning to the browser to paint).

PR #7426, opened 2022-07-21
Editor search results adjustments v1 (baby step)

jonbell-lot23:
We have a lot of other ideas in store for this feature, but for starters we're going to make the color a little more obvious.

Was:
image

Now:
image


jcmorrow (approved):
💖

PR #7428, opened 2022-07-21
Improve highlighting in sources dark theme 

jonbell-lot23:
Here's a 5 second loom showing the difference:
https://www.loom.com/share/a52bfffc5a534024a607decd32b58b77

Before: we were hovering with a text input cursor while using a very faint background color
Now: pointer cursor and more clear background color


jaril (approved):
👍 Needs a lint pass (`npm run lint:fix`) but should be good!

PR #7430, opened 2022-07-21
Search results highlighting

jonbell-lot23:
Previously, we only showed the current search result. With this change, we're showing all results but using a different color for the one you're currently on. This aligns us to VS Code and other tools that do a good job with this scenario.

Here's a 30 second Loom explaining it:
https://www.loom.com/share/526564f29a3549adbaa6afc9a04e4118


jcmorrow (approved):
Seems like a step in the right direction 👍

PR #7435, opened 2022-07-22
Migrate Editor-related content to TS

markerikson:
This PR:

- Migrates effectively all Editor components to TS
- Migrates all Editor-related logic (search, CodeMirror setup, etc) to TS
- Updates our Redux store setup to sanitize the new sources contents from the store to help Redux DevTools perf

This is a precursor to swapping over where we store source contents, as a lot of these files depend on the current (legacy) sources reducer.  Converting these to TS now will make the final swap to the new sources reducer much more straightforward.


src/devtools/client/debugger/src/actions/file-search.ts
```diff
@@ -15,17 +15,11 @@ import {
...
   clearSearch,
-  // @ts-expect-error
   find,
-  // @ts-expect-error
   findNext,
-  // @ts-expect-error
   findPrev,
-  // @ts-expect-error
   removeOverlay,
-  // @ts-expect-error
```
jcmorrow, line 28: This is nice to see 👼

src/devtools/client/debugger/src/actions/sources/blackbox.ts
```diff
@@ -33,7 +33,7 @@ async function blackboxActors(
   return { isBlackBoxed: !isBlackBoxed };
 }
 
-export function toggleBlackBox(cx: Context, source: Source): UIThunkAction {
+export function toggleBlackBox(cx: Context, source: Source | SourceWithContent): UIThunkAction {
```
jcmorrow, line 36: Do you think this will stay typed as `Source | SourceWithContent`? Or is that a temporary type until we get to phase II?

src/devtools/client/debugger/src/components/App.tsx
```diff
@@ -5,45 +5,51 @@
...
 import { getSelectedPanel } from "ui/reducers/layout";
-import { getToolboxLayout } from "ui/reducers/layout";
 import { useGetUserSettings } from "ui/hooks/settings";
 
 import KeyShortcuts from "devtools/client/shared/key-shortcuts";
 
 import { EditorPane } from "./Editor/EditorPane";
 
-class Debugger extends Component {
+const mapStateToProps = (state: UIState) => ({
```
jcmorrow, line 22: Hmmmmm I see we are pulling `connect` code *up* here. I think most of our files have this towards the bottom, and in fact I've actually seen folks pulling these *down* to match other places. I am used to seeing them at the bottom, but I don't care that much as long as we follow *something* consistently. (And obviously the number of connect statements should be going down over time because hooks exist).

src/devtools/client/debugger/src/components/App.tsx
```diff
@@ -5,45 +5,51 @@
...
+});
+
+type PropsFromRedux = ConnectedProps<typeof connector>;
+type DebuggerProps = PropsFromRedux & { wrapper: HTMLDivElement };
+
+class Debugger extends Component<DebuggerProps> {
+  shortcuts = new KeyShortcuts({ window, target: this.props.wrapper });
+
   getChildContext = () => {
+    // @ts-expect-error L10n must be a global?
```
jcmorrow, line 38: Pretty sure it's actually dead?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Breakpoint.tsx
```diff
@@ -98,6 +98,7 @@ class Breakpoint extends PureComponent<BreakpointProps> {
     const doc = getDocument(sourceId);
 
     resizeBreakpointGutter(editor.codeMirror);
+    // @ts-expect-error method doesn't exist on Doc
```
jcmorrow, line 101: This feels like a good sign that we have this typed incorrectly 🤔 . Is it supposed to be `Editor`?

src/devtools/client/debugger/src/components/Editor/EditorMenu.tsx
```diff
@@ -20,45 +22,50 @@ import {
...
-    );
+  showMenu(props: FinalEMProps) {
+    const { cx, selectedSource, editorActions, alternateSource, contextMenu: event } = props;
 
     showMenu(
       event,
       editorMenuItems({
         cx,
         editorActions,
+        // @ts-expect-error Source/SourceWithContent mismatch  but will be changed shortly
```
jcmorrow, line 66: Ah, sounds like this answers my earlier question.

src/devtools/client/debugger/src/components/Editor/HighlightLine.tsx
```diff
@@ -96,11 +123,12 @@ export class HighlightLine extends Component {
     }
 
     const doc = getDocument(sourceId);
+    // @ts-expect-error method doesn't exist
```
jcmorrow, line 126: again, that is *strange*

src/devtools/client/debugger/src/components/Editor/SearchBar.tsx
```diff
@@ -67,24 +161,30 @@ class SearchBar extends Component {
...
 
-    shortcuts.on(searchAgainShortcut, e => this.traverseResults(e, false));
+    shortcuts.on(searchAgainShortcut, (e: React.KeyboardEvent) => this.traverseResults(e, false));
   }
 
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate() {
+    // @ts-expect-error string refs ugh!
     if (this.refs.resultList && this.refs.resultList.refs) {
+      // @ts-expect-error string refs ugh!
```
jcmorrow, line 182: Not gonna lie I'm scared of this file 👻

src/devtools/client/debugger/src/components/Editor/Tab.tsx
```diff
@@ -37,14 +39,53 @@ import classnames from "classnames";
...
+  onDragOver: React.DragEventHandler<HTMLDivElement>;
+  onDragStart: React.DragEventHandler<HTMLDivElement>;
+  onDragEnd: React.DragEventHandler<HTMLDivElement>;
+}
+
+const mapStateToProps = (state: UIState, { source }: TabProps) => {
+  const selectedSource = getSelectedSource(state);
+
+  return {
+    cx: getContext(state),
```
jcmorrow, line 53: Ugh.

src/devtools/client/debugger/src/components/Editor/index.tsx
```diff
@@ -464,40 +520,9 @@ class Editor extends PureComponent {
   }
 }
 
+// @ts-expect-error context shenanigans
```
jcmorrow, line 523: Makes me a little sad. Since we never use context I wonder if it would be reasonable to make Context's type `CurrentContextType | {}` and just start passing `{}` everywhere until we actually pull it out. Probably a bad idea.

src/ui/actions/comments.ts
```diff
@@ -166,6 +166,7 @@ export function createLabels(
     if (snippet) {
       await waitForEditor();
       const CodeMirror = getCodeMirror();
+      // @ts-expect-error runMode doesn't exist on CodeMirror
```
jcmorrow, line 169: 🤨

jcmorrow (commented):
I think I'm generally ok with this? I hate to be that team member but we should probably either stick to bottom connect statements or at least make it clear that we are going with a "leave it where you found it" approach just to reduce diff noise back and forth. I don't think getting the codemirror types right is the most important thing in the world if the code runs, but it's very strange to me that the types we have don't line up with the function calls in the working code...

src/devtools/client/debugger/src/components/App.tsx
markerikson: Yeah, personally I've always put the `connect` call itself at the bottom, but the `mapState/mapDispatch` definitions _above_ the component, so you can read downwards and see what props it's getting.

Beyond that, there's a couple issues here:

- The `ConnectedProps<T>` usage requires splitting the `connect` call into `const connector = connect(mapState, mapDispatch)`, and then inferring `type PropsFromRedux = ConnectedProps<typeof connector>`
- _Technically_ you can define the props types down at the bottom of the file and use them earlier, but that does feel really awkward to me. I would expect to see the props types defined first _before_ they're used in the component

src/devtools/client/debugger/src/components/Editor/index.tsx
markerikson: The issue here is TS arguing "you defined a class above here, but you didn't say it would have a `.contextTypes` field attached!"

(the real solution is "DON'T USE LEGACY CONTEXT!" :)  )

markerikson:
The Node E2E test failures are unrelated - the backend isn't returning recording info.  This looks good otherwise (including a hand-check of editor search bar behavior, which I touched)

PR #7441, opened 2022-07-23
Add experimental recording cache

jasonLaster:
Adds the recording cache as an experimental setting that users can enable

Screen Shot 2022-07-23 at 3 40 43 PM


src/ui/setup/helpers.ts
```diff
@@ -55,9 +57,11 @@ export async function setupAppHelper(store: UIStore) {
     asyncStore,
     triggerEvent,
     replaySession,
+    client,
     sendMessage: (cmd, args = {}, pauseId) =>
       sendMessage(cmd, args, window.sessionId, pauseId as any),
 
+    releaseSession: () => client.Recording.releaseSession({ sessionId: window.sessionId }),
```
jasonLaster, line 64: not needed for this PR, but helpful for testing

PR #7446, opened 2022-07-25
Socket supports multiple listeners per event type

bvaughn:
I didn't touch the _Analysis_ code because I don't fully understand why it's written the way it is. Ideally, we wouldn't need to special case specific event types like this but maybe @markerikson or @hbenl will have more context.

In support of #7418 and FE-337.


jasonLaster:
👍 Happy to see us updating Socket to support multiple listeners. Wonder if we could update Analysis as well...

PR #7448, opened 2022-07-25
Show GraphQL errors when filtering recordings

jcmorrow:
Pairs with https://github.com/replayio/backend/pull/6107 to make sure
that we will display GraphQL errors that we get back when filtering
recordings.

https://user-images.githubusercontent.com/5903784/180859234-1dbcb8e3-f57f-4918-a9b5-0caf666a8d5e.mp4

CC'ing @jonbell-lot23 to see if he can suggest any good treatments for the error message.

### DIFF
--- a/src/ui/components/Library/Team/View/Recordings/RecordingsPage.tsx
+++ b/src/ui/components/Library/Team/View/Recordings/RecordingsPage.tsx
@@ -24,7 +24,7 @@ function MyRecordingsPage({ team }: { team: typeof MY_LIBRARY_TEAM }) {
 
   if (loading || !recordings) {
     return (
-      <div className="flex flex-col flex-grow p-4 overflow-hidden">
+      <div className="flex flex-grow flex-col overflow-hidden p-4">
         <LibrarySpinner />
       </div>
     );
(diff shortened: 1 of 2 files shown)


jcmorrow:
@jaril not sure if this will end up being necessary, I think the changes in the backend PR to get things not to be reported to Sentry for bad user input is the most important part, but I figured I would throw up an example of what we *could* do as well.

jasonLaster:
will defer to @jonbell-lot23 for the treatment, but a disabled state in the search box would probably work. also - shouldnt be on key press which is weird

jonbell-lot23:
Hey team! Just chatted with @jcmorrow real quick about this and here's the design POV:

* It's possible to type garbage into our search field and get zero results (e.g. "foo:03rufknlsdflk" would say zero results)
* This is a similar pattern -- there's garbage, there are zero results, so we should show zero results

So I don't think we do anything special here. That said, if we think this is a common state because links will frequently be broken for some reason, I could be persuaded to go further here. But this seems like small potatoes so it doesn't warrant a whole custom error message.

PR #7451, opened 2022-07-25
Switch all source + source contents usage to the new sources reducer

markerikson:
This PR:

- Ports the logic Josh wrote in #7375 for fetching source contents over to the new sources reducer
- Switches the `loadSourcText` calls over to the new thunk
- Temporarily disables all blackboxing logic (to be re-enabled in a follow-up PR in the next couple days), as the old `Source` type had `source.isBlackBoxed` and that's used in a bunch of places
- **Has a _massive_ commit to switch _alllll_ usages of the legacy `Source` and `SourceWithContents`/resource types over to the new `SourceDetails/SourceContent` types**
- Converts `utils/quick-open.js` to TS
- Tweaks the logic in `formatSources` to try to list and open the "canonical" source for the selected item
- Removes the legacy sources reducer from the store setup
- Deletes the legacy source text thunk file
- Tweaks the `notAllowed()` util (used in some proxy handlers to indicate unauthorized delayed access) to print some details on _what_ is being accessed and where

This PR may be easier to read commit-by-commit.


src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -195,20 +206,26 @@ export function newOriginalSources(sourceInfo: SourceData[]): UIThunkAction<Prom
...
   };
 }
-export function newGeneratedSources(sourceInfo: SourceData[]): UIThunkAction<Promise<Source[]>> {
+export function newGeneratedSources(
+  sourceInfo: SourceData[]
+): UIThunkAction<Promise<SourceDetails[]>> {
   return async (dispatch, getState, { ThreadFront }) => {
     const resultIds: string[] = [];
+
+    // TODO Rewrite "check pending" handling
```
jcmorrow, line 227: I suspect that this code will end up getting deleted eventually, but we need to replace it with something else to do the job it was doing (select the previously selected location and recreate previous breakpoints once sources are loaded).

src/ui/reducers/sources.ts
```diff
@@ -1,34 +1,103 @@
...
   prettyPrintedFrom?: string;
   url?: string;
+  // TODO stableId: string;
+}
+
+/**
+ * Both `Source` and `SourceDetails` have `{id, url?}`,
+ * and we usually only need those fields in string manipulation functions.
+ */
+export interface MiniSource {
```
jcmorrow, line 39: 🤏  Thiiiiiiiiiis small.

src/ui/reducers/sources.ts
```diff
@@ -41,31 +110,262 @@ const sourcesSlice = createSlice({
...
+export const getUniqueUrlForSource = (state: UIState, sourceId: string) => {
+  const sourceDetails = getSourceDetails(state, sourceId);
+  if (!sourceDetails || !sourceDetails.url) {
+    return null;
+  }
+  if (state.experimentalSources.sourcesByUrl[sourceDetails.url].length > 1) {
+    // I was going to put the query string back here...
+    // But I'm not sure we're actually *removing* query strings in the first
+    // place yet!
+    // TODO @jcmorrow - actually remove query strings, then add them back here.
```
jcmorrow, line 281: 😬

src/ui/reducers/sources.ts
```diff
@@ -41,31 +110,262 @@ const sourcesSlice = createSlice({
...
+    // presence
```
jcmorrow, line 282: this line seems like it should be ✂️

src/ui/reducers/sources.ts
```diff
@@ -41,31 +110,262 @@ const sourcesSlice = createSlice({
...
+}
+
+export const getPreviousPersistedLocation = (state: UIState) =>
+  state.experimentalSources.persistedSelectedLocation;
+
+/*
+export const getStableLocationForLocation = (
+  state: UIState,
+  location: Location
+): StableLocation => {
```
jcmorrow, line 334: FWIW, this is what I was using for holding pendingLocations. YMMV.

packages/protocol/utils.ts
```diff
@@ -77,8 +77,8 @@ export function binarySearch(start: number, end: number, callback: (mid: number)
   return start;
 }
 
-function NotAllowed() {
-  console.error("Not allowed");
+function NotAllowed(reason = "") {
+  console.trace(`Not allowed${reason ? ` (${reason})` : ""}`);
```
jcmorrow, line 81: I like these changes.

jcmorrow (commented):
I have no idea what I was doing making a PR that was twice as big as this. I must have been fever-coding. Anyways - let's make sure we get the pending breakpoints and pending selected location stuff back in, and then I think this is probably good to go!

packages/protocol/utils.ts
```diff
@@ -89,22 +89,28 @@ export const DisallowEverythingProxyHandler: ProxyHandler<object> = {
   //  NotAllowed();
   //},
   //set() { NotAllowed(); },
-  get() {
+  get(target, p) {
+    if (
+      ["Symbol(immer-state)", "Symbol(immer-draftable)", "isReactComponent"].includes(p.toString())
+    ) {
+      return;
+    }
```
jasonLaster, line 97: what's this doing?

src/devtools/client/debugger/src/actions/sources/symbols.ts
```diff
@@ -24,14 +24,14 @@ async function doSetSymbols(source: Source, thunkArgs: { dispatch: AppDispatch }
...
     if (!symbols || symbols.loading) {
       return null;
     }
 
     return fulfilled(symbols);
   },
-  createKey: ({ source }) => source.id,
+  createKey: ({ source }: { source: SourceDetails }) => source.id,
   action: ({ source }, thunkArgs) => doSetSymbols(source, thunkArgs),
 });
```
jasonLaster, line 37: cant wait to kill symbols
https://linear.app/replay/issue/FE-446/start-using-symbols-from-the-server

src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -317,3 +327,69 @@ function checkNewSources(cx: Context, sources: SourceDetails[]): UIThunkAction {
     return sources;
   };
 }
+
+// Delay adding these until the store is created
+export const setupSourcesListeners = (startAppListening: AppStartListening) => {
+  // When sources are received, we want to check for an existing
+  // selected location, and open that automatically.
+  // Also, we
```
jcmorrow, line 335: Don't leave me hanging!

src/devtools/client/debugger/src/actions/sources/symbols.ts
markerikson: Yeah, this may be the last remaining use of `memoizableAction` - I'd have to see what else is left.

src/ui/reducers/sources.ts
```diff
@@ -328,6 +328,27 @@ export function getHasSiblingOfSameName(state: UIState, source: MiniSource) {
   return state.experimentalSources.sourcesByUrl[source.url]?.length > 0;
 }
 
+export const getCanonicalSourceFromEntities = (
+  detailsEntities: Dictionary<SourceDetails>,
+  sd: SourceDetails
+) => {
+  const canonicalSource = detailsEntities[sd.canonicalId];
+  return canonicalSource ?? sd;
```
jcmorrow, line 336: We might want to `console.error` or something if this isn't here. Unless I'm missing a case, that seems like it would mean something was really broken.

src/ui/reducers/sources.ts
```diff
@@ -328,6 +328,27 @@ export function getHasSiblingOfSameName(state: UIState, source: MiniSource) {
...
+};
+
+export const getCanonicalSource = (state: UIState, sd: SourceDetails) => {
+  return getCanonicalSourceFromEntities(state.experimentalSources.sourceDetails.entities, sd);
+};
+
+export const getCanonicalSourceForUrl = (state: UIState, url: string) => {
+  const sd = getSourceByUrl(state, url);
+  if (!sd) {
```
jcmorrow, line 345: I would also maybe add logging here. It's a bug if we're trying to open a source for an invalid URL.

src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -317,3 +327,69 @@ function checkNewSources(cx: Context, sources: SourceDetails[]): UIThunkAction {
...
+        }
+      }
+
+      // TODO Re-enable blackboxing - this is a no-op for now
+      dispatch(restoreBlackBoxedSources(cx, sources));
+
+      // There may have been some breakpoints / print statements persisted as well.
+      // Reload those if possible.
+      for (const source of canonicalTabSources) {
+        dispatch(checkPendingBreakpoints(cx, source.id));
```
jcmorrow, line 391: So this whole method is still the same? That's nice.

src/ui/utils/thunk.ts
```diff
@@ -5,3 +5,5 @@ export interface ThunkExtraArgs {
   client: typeof clientCommands;
   ThreadFront: typeof ThreadFront;
 }
+
+export const extraThunkArgs = {} as ThunkExtraArgs;
```
jcmorrow, line 9: I feel like there's a story here?

jcmorrow (approved):
I haven't actually gotten to play with the preview branch yet because of the TSC thing, but pending the preview branch working I think this looks great 👍

src/ui/utils/thunk.ts
markerikson: Yeah, the listener middleware has the same ability to accept an `extra` argument as the thunk middleware, and I briefly thought I'd need to access `ThreadFront.recordingId` in that listener.  

The `extraThunkArgs` object _was_ defined in `store.ts`, so I couldn't import it over to `listenerMiddleware.ts`.  So, I extracted it over to the thunk file, which is where the TS `ThunkExtraArgs` type is defined anyway, and now the listener middleware file can safely import that one object and pass it into `createListenerMiddleware({extra: extraThunkArgs})`.

(I ended up _not_ needing `ThreadFront.recordingId` after all, but hey, no reason not to have it available!)

jcmorrow:
This feels good on the preview - let's see how it does on E2E's after the rebase!

src/devtools/client/debugger/src/actions/sources/newSources.ts
markerikson: Yeah, I was debating embedding the logic in here directly, but figured I could leave it as-is

PR #7455, opened 2022-07-26
Fix edge case for undefined value

bvaughn:
Follow on for #7431

### DIFF
--- a/packages/bvaughn-architecture-demo/src/utils/protocol.ts
+++ b/packages/bvaughn-architecture-demo/src/utils/protocol.ts
@@ -182,11 +182,12 @@ export function protocolValueToClientValue(
   }
 
   // (BAC-1986) Additional edge case where an Object Property is a symbol but has no corresponding object value.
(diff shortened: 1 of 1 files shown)


packages/bvaughn-architecture-demo/src/utils/protocol.ts
```diff
@@ -182,11 +182,12 @@ export function protocolValueToClientValue(
...
 
   // (BAC-1986) Additional edge case where an Object Property is a symbol but has no corresponding object value.
+  // Apparently this scenario means that the property key itself is a symbol and the value is undefined.
   if (isProtocolProperty(protocolValue)) {
     if (protocolValue.isSymbol === true) {
       return {
         name,
-        preview: `${(protocolValue as ProtocolProperty).name}`,
+        preview: "undefined",
         type: "symbol",
```
loganfsmyth, line 191: We should chat at some point, because this isn't quite right. For the `undefined` case, it should be `type: "undefined"` too. This should really be handled as part of the `keys` filter above.

PR #7456, opened 2022-07-26
Address DEV warnings

bvaughn:
So far as I can tell from my spot testing, this does not result in any user-facing behavioral change.

### DIFF
--- a/src/devtools/client/debugger/src/components/App.tsx
+++ b/src/devtools/client/debugger/src/components/App.tsx
@@ -34,6 +34,11 @@ type DebuggerProps = PropsFromRedux & { wrapper: HTMLDivElement };
 class Debugger extends Component<DebuggerProps> {
   shortcuts = new KeyShortcuts({ window, target: this.props.wrapper });
 
+  static childContextTypes = {
+    l10n: PropTypes.any,
+    shortcuts: PropTypes.object,
+  };
+
   getChildContext = () => {
     // @ts-expect-error L10n must be a global?
     return { shortcuts: this.shortcuts, l10n: L10N };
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/App.tsx
```diff
@@ -34,6 +34,11 @@ type DebuggerProps = PropsFromRedux & { wrapper: HTMLDivElement };
 class Debugger extends Component<DebuggerProps> {
   shortcuts = new KeyShortcuts({ window, target: this.props.wrapper });
 
+  static childContextTypes = {
```
markerikson, line 37: :hand: Just add `static` in front of the other `childContextTypes` field below, which is what I was _trying_ to do when I moved that inside the class.  (or move it back out and do `Debugger.childContextTypes = ` after the definition, which is how I was going to fix this myself.)

src/ui/actions/session.ts
```diff
@@ -165,31 +166,51 @@ export function createSocket(
 
       const loadPoint = new URL(window.location.href).searchParams.get("point") || undefined;
 
+      const queuedActions: Array<Action | UIThunkAction> = [];
+      let flushTimeoutId: NodeJS.Timeout | null = null;
+
+      function flushQueuedActions() {
+        flushTimeoutId = null;
+
+        const actions = queuedActions.splice(0);
```
markerikson, line 175: :question: I'm confused by the `.splice(0)`. What's that doing?

src/devtools/client/debugger/src/components/App.tsx
bvaughn: Oh fuck. I ... missed the other one 😮‍💨 It's been a day

src/ui/actions/session.ts
bvaughn: Grabbing all of the current actions _and_ clearing out the queue so we can add more actions to it later.

```js
const actions = queuedActions.splice(0);

// actions now has all of the items queuedActions used to have
// queuedActions is now empty
```

markerikson: Ah. I probably woulda just replaced the existing array or something :)

PR #7459, opened 2022-07-26
disable `incremental` TypeScript feature

EndangeredMassa:
We (at Vercel) were looking into the TypeScript failure [mentioned on Twitter](https://twitter.com/brian_d_vaughn/status/1551936568465694722).

---

This seems to be caused by an error in TypeScript itself related to the `incremental: true` setting. We were able to reproduce it locally using this repo, but only some of the time. It is a flakey issue.

Turning the `incremental` setting off or deleting the `tsconfig.tsbuildinfo` file seems to mitigate the problem.

---

We suggest:

* turning off `incremental` in the `tsconfig.json` seems to mitigate the issue (which is what this PR does)
* try upgrading TypeScript to see if that resolves the issue. It looks like there have been some merged PRs that address issues with this setting in TypeScript itself since the project version of 4.6.3. ([Example](https://github.com/microsoft/TypeScript/issues/49527))
* if the issue persists, let us know, but you may also want to report the issue to TypeScript

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig.json to read more about this file */
     /* Basic Options */
-    "incremental": true /* Enable incremental compilation */,
+    "incremental": false /* Enable incremental compilation */,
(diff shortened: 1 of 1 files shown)


bvaughn (approved):

markerikson:
Thank you!

FWIW I'm also up for us bumping to the latest TS version and leaving `incremental` on for a while, and seeing if that makes a difference.

PR #7463, opened 2022-07-27
Fix broken node tests

jcmorrow:
If this works for this one I'll do the rest

I know we have a thing for this already (`example-recordings.json`), but I kind of want to make it clear that this is a jerry-rigged patch that we should not run with, and I feel like just checking that file in might mean that we stay reliant on this for longer. (Also, anyone's workflow that might be reliant on `test/example-recordings.json` being in `.gitignore` would break)

### DIFF
--- a/test/e2e/getExample.ts
+++ b/test/e2e/getExample.ts
@@ -5,6 +5,7 @@ import config from "./config";
 import { recordNodeExample } from "./recordNode";
 import { recordPlaywright, uploadLastRecording } from "./recordPlaywright";
 import { waitUntilMessage } from "./utils";
+const manifest = require("../manifest");
 
 const exampleFile = path.resolve(__dirname, "../../test/example-recordings.json");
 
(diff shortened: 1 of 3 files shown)


bhackett1024:
Why can't we make a new recording, like we've done in the past and do for other tests?  Using an existing recording means we aren't doing true end-to-end testing of the recording / debugging experience and will allow for things to break without us detecting that.

jcmorrow:
> Why can't we make a new recording, like we've done in the past and do for other tests? Using an existing recording means we aren't doing true end-to-end testing of the recording / debugging experience and will allow for things to break without us detecting that.

@bhackett1024 Yeah, I agree, I don't intend for this to be permanent. Recording node on mac is broken right now, but that's not a front-end problem. Our test suite should not be perma-failing because of an error in a completely different repo.

bhackett1024:
Is there an issue filed for fixing mac node recording?  I haven't seen anything.  That's a pretty serious problem we need to get on top of, and if recording node on mac is broken I'd rather that the tests fail than us not knowing about it at all.

jcmorrow:
@bhackett1024 IDK, I'm running with what I heard, which is that node recording is not working right now, and it's not helpful for the front-end tests to fail because of that 🤷

bhackett1024:
OK, thanks for filing an issue for this.  That needs to be the first step when things break, so we don't lose track of problems like this.  Instead of this workaround, can we just disable the tests until the node problem is fixed?

jcmorrow:
>  can we just disable the tests until the node problem is fixed?

Then the runs might pass, but it seems like the tests not running at all might be worse than the tests always failing?

bhackett1024:
The tests not running is OK as long as we are keeping track of the problem and can get it fixed quickly.  Disabling tests which are broken is nice in a situation like this because it's a temporary situation and disabling the tests requires less effort than working around the problem.

FWIW the problem might be fixed now, I did a new node release and triggered a run here: https://github.com/replayio/devtools/actions/runs/2749539826

bhackett1024:
It looks like things got fixed by doing the node release (see also https://linear.app/replay/issue/BAC-2012/all-node-recordings-are-failing-on-e2e-tests).

jcmorrow:
Great, thanks @bhackett1024 !

PR #7466, opened 2022-07-28
New Console UI/UX regressions

bvaughn:
Issues with the current point indicator (the red line)
- [x] Should be at top not bottom.
- [x] Should show between messages when there is no exact match.

Context menu issues
- [x] Looks bad on light theme.
- [x] Badge colors weren't toggle-able for Jason (not writing to Redux).

Misc issues
- [x] Hide toggle for errors and warnings with no stack.
- [x] Badge circle too big in legacy app.
- [x] Sometimes toggling warnings/errors on the left side doesn't show them in the log list.


bvaughn:
Let's land these changes incrementally. This is a good batch of them and I can follow up with the rest.

PR #7467, opened 2022-07-28
If the user is already logged in, redirect

jcmorrow:
Immediately, without asking them to sign in again. This comes up when
the Replay browser is authing to app.replay.io. It has a query string to
return to some other path, but then it opens Chrome to auth, and Chrome
then reopens Replay browser, but with a new tab on the Home page. The
old tab is still there, pointed at `/login?returnTo=asdklfjkdlsjkl`, but
if you refresh it looks like you *still* need to log in, even though it
clearly says "Rec" and not "Sign in" in the browser. If we land on that
login page, but we have a user, we should just redirect them to the
`returnTo` path.

I'm gonna need a thorough @ryanjduffy check on this logic though because this stuff always confuses the crap out of me.

### DIFF
--- a/src/ui/components/shared/Login/Login.tsx
+++ b/src/ui/components/shared/Login/Login.tsx
@@ -15,6 +15,7 @@ const isOSX = Services.appinfo.OS === "Darwin";
 import { PrimaryLgButton } from "../Button";
 import { OnboardingContentWrapper, OnboardingModalContainer } from "../Onboarding";
 import { GetConnection, GetConnectionVariables } from "graphql/GetConnection";
+import { useRouter } from "next/router";
 
 enum LoginReferrer {
   default = "default",
--- a/pages/login.tsx
+++ b/pages/login.tsx
@@ -3,14 +3,20 @@ import React, { useEffect } from "react";
 import { useAppDispatch } from "ui/setup/hooks";
 import Login from "ui/components/shared/Login/Login";
 import { clearExpectedError } from "ui/reducers/app";
+import useAuth0 from "ui/utils/useAuth0";
 
 export default function LoginPage() {
   const router = useRouter();
+  const { user } = useAuth0();
   const dispatch = useAppDispatch();
+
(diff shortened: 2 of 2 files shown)


src/ui/components/shared/Login/Login.tsx
```diff
@@ -209,13 +210,18 @@ export function LoginLink({
...
-  const { loginWithRedirect, error } = useAuth0();
+  const { loginWithRedirect, error, user } = useAuth0();
+  const router = useRouter();
   const [sso, setSSO] = useState(false);
 
   if (returnToPath.startsWith("/login")) {
     returnToPath = "/";
   }
 
+  if (user) {
```
ryanjduffy, line 221: We should add `&& returnToPath` here since it could be unset and the redirect wouldn't make much sense

src/ui/components/shared/Login/Login.tsx
```diff
@@ -95,7 +96,7 @@ function LoginMessaging() {
   // time that they have opened the browser.
   if (referrer === LoginReferrer.firstBrowserOpen) {
     return (
-      <div className="text-base text-center">
+      <div className="text-center text-base">
```
ryanjduffy, line 99: I'm team base-before-center but w/e ;)

ryanjduffy (approved):
I haven't tested this but here are the two things that come to mind:

* If the user is authenticated and manually navigates to the login page, they should probably be able to initiate a login and not be automatically redirected somewhere. This could be a follow up since it's pretty edge case-y.
* I think the logic should be in `/pages/login` rather than the component. Feels like the right place to manage routing logic rather than component land.

ryanjduffy:
@jcmorrow  checking back if this is ready to land?

jcmorrow:
> @jcmorrow  checking back if this is ready to land?

Whoops I absolutely thought I merged this last week 🤔

jcmorrow:
Lol, I hit this yesterday and was like "dangit, I thought I fixed that!"

PR #7472, opened 2022-07-30
Fix O(N^2) look up for hitcounts

jasonLaster:
This fixes an issue we were seeing for 100k line files where looking up the hitcounts for a source location was O(N^2)

### Before
Screen Shot 2022-07-29 at 10 44 11 PM

### After
(can't find it in the profile)

### DIFF
--- a/src/ui/reducers/index.ts
+++ b/src/ui/reducers/index.ts
@@ -10,6 +10,7 @@ import consoleReducers from "devtools/client/webconsole/reducers";
 import * as consoleSelectors from "devtools/client/webconsole/selectors";
 import * as debuggerSelectors from "devtools/client/debugger/src/selectors";
 import * as inspectorReducers from "devtools/client/inspector/reducers";
+import * as hitCountsSelectors from "./hitCounts";
 import { selectors as possibleBreakpointsSelectors } from "./possibleBreakpoints";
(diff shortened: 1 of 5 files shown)


src/ui/reducers/index.ts
```diff
@@ -32,6 +33,7 @@ export const selectors = {
   ...consoleSelectors,
   ...debuggerSelectors,
   ...eventListenerBreakpointsSelectors,
+  ...hitCountsSelectors,
```
jasonLaster, line 36: this is nice to have because it lets me evaluate `app.selectors.getHitCountsForSource` in the console

markerikson: :speech_balloon: The one issue here is that this also ends up including the Redux action creators, which aren't actually selectors :)

Granted we've had that happen with other files too, but going forward I'd like to stick with what I did over in the new `ui/reducers/sources` file and explicitly define an object with the right selectors inside, which can then be imported and spread.

markerikson (approved):

src/ui/reducers/index.ts
jasonLaster: @markerikson that works for me. Just love having an `app.selectors`

PR #7475, opened 2022-08-01
Upgrade to CodeMiror 5.65.7

jasonLaster:


jasonLaster:
This does not fix FE-491, but could still be good to land
https://devtools-epwg66mlb-recordreplay.vercel.app/recording/flicker-switching-sources--c86f662d-f91d-4d36-9cf3-e9a369df577e?point=64254673646623787651720665135841280&time=24975&hasFrames=false

PR #7483, opened 2022-08-02
Always import material icon fonts

jaril:
Fixes FE-408.

The issue appears to be that we were loading the local Material Icons, which in Lenz's case (for some reason), did not have the symbols that we were using.

This changes it so that we always import the material icon fonts.

### DIFF
--- a/public/fonts/material_icons/material_icons.css
+++ b/public/fonts/material_icons/material_icons.css
@@ -3,8 +3,7 @@
   font-style: normal;
   font-weight: 400;
   font-display: block;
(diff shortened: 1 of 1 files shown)


jcmorrow (approved):
I don't feel too strongly about this either way. If someone has the latest version of the font installed, I could see them being annoyed that we still download a copy, but if someone has an *old* version of the font installed (which I suspect is the case here), it feels like a bug *not* to download a new copy. Given how bandwidth heavy Replay is, I supposed erring on the side of correctness at the expense of one extra font's worth of bandwidth is probably OK (plus, caching).

PR #7492, opened 2022-08-03
Attempt to avoid API requests outside of loaded regions

bvaughn:
Note that this is currently not 100% reliable because time is fuzzier than execution point.


packages/bvaughn-architecture-demo/components/console/Input.tsx
```diff
@@ -43,6 +32,27 @@ export default function Input() {
...
   }, [searchState.visible]);
 
+  // If we are not currently paused at an explicit point, find the nearest point and pause there.
+  let pauseId = explicitPauseId;
+  let frameId: FrameId | null = null;
+  if (pauseId === null) {
+    const isLoaded =
+      loadedRegions !== null &&
+      executionPoint !== null &&
+      isPointInRegions(executionPoint, loadedRegions.loaded);
```
bvaughn, line 42: Conceptually, we need to avoid making API requests outside of loaded regions to avoid `CommandErrors`. I don't know if a better way to do that than this approach (subscribing to the loaded regions and checking to see if the request is within them).

**Unfortunately** time is fuzzier than execution point, and it's possible that we _think_ we're inside of a loaded region when we actually aren't.

I don't know how to work around this at the moment.

src/ui/components/SecondaryToolbox/NewConsole.tsx
```diff
@@ -139,29 +139,9 @@ function FocusContextReduxAdapter({ children }: PropsWithChildren) {
...
-        });
-      } else {
-        timeoutId = setTimeout(updateFocusRegionOnceLoaded, 250);
-      }
-    };
-
-    let timeoutId = setTimeout(updateFocusRegionOnceLoaded, 250);
-    return () => {
-      clearTimeout(timeoutId);
-    };
```
bvaughn, line 164: This should not be necessary. We should be able to load _some_ of our UI without being in a loaded region. Only certain API requests require this.

Also this check wasn't 100% reliable before – for the same reason mentioned above.

packages/shared/client/ReplayClient.ts
```diff
@@ -60,6 +68,19 @@ export class ReplayClient implements ReplayClientInterface {
...
+  }
+
+  addEventListener(type: ReplayClientEvents, handler: Function): void {
+    if (!this._eventHandlers.has(type)) {
+      this._eventHandlers.set(type, []);
+    }
+
+    const handlers = this._eventHandlers.get(type)!;
+    handlers.push(handler);
+  }
```
bvaughn, line 82: Ugh, this approach (events) doesn't work with the Proxy client approach.

src/ui/components/SecondaryToolbox/NewConsole.tsx
bvaughn: Josh says this is the not the right values to be reading. There is another selector that returns both time and execution point.

bvaughn: Can I put points in the Focus context instead (and update the new app to do what `Timeline` does in terms of eagerly reading these)?

bvaughn: Related commit: https://github.com/replayio/devtools/commit/aa62c1e8a1aac5cef19453014894023735f05df2

packages/shared/client/ReplayClient.ts
bvaughn: Commit 73041cb implements the last workaround mentioned above.

bvaughn:
I'm going to land this as-is because it helps clean things up. Then I'll follow up on the remaining comment with a new PR.

PR #7493, opened 2022-08-03
Remove MarkupView

jasonLaster:
* Removes MarkupView class
  * Moves callbacks like `expandNode` to thunks
  * Removes `loadNewDocument` in favor of subscribing a thunk to the ThreadFront directly
  * moves selection to a global, but will likely kill in a subsequent pr

Note: this was mostly working and then i just kept going in favor of having something where we could 🔫 MarkupView

https://www.loom.com/share/efa3811a15034eefbb072d15ac4e5570

Fixes FE-509


markerikson:
Additional changes on top of Jason's original draft PR:

- The `Selection` class now creates a singleton instance in its own module, and _every_ file that cares about reading or writing to `selection.nodeFront` now directly does `import { selection } from "whatever/selection"`. This eliminates the need to set `window.gSelection` as a global, or grab it off of `inspector`. It's a singleton, let's treat it like one.  (Prior to this PR, a single instance was created inside of `Inspector`.)
- Rewrote `reducers/markup` as a `createSlice` + `createEntityAdapter`
  - Moved the types defined in `state/markup.ts` over into the reducer slice file, as that's the standard location for types
- Re-implemented the "handle reloading the markup nodes content" logic as an RTK listener that waits for the `"PAUSED"` action to be dispatched. 
- Updated the "highlight this hovered node" logic to keep the current node ID as a local module-scoped variable. No reason for this to be in Redux.
- Fixed a mutation bug in `NodeAttributes` caused by sorting an array
- Re-added `toggleNodeExpanded` as a thunk, as that logic is needed in a couple places
- Rewrote `client/index.ts` to subscribe to `ThreadFront.on("paused" / "resumed")` and deleted the old `client/events.js` file where that logic used to live (which was doing overly-abstracted things with binding action creators and cutesy object key mapping anyway)

PR #7496, opened 2022-08-04
Batch onPaints and replaySession lookups

jasonLaster:
This has two nice perf wins:
1. batch onPaints which caused some render thrashing
2. cache the replaySessions lookup which was showing up on each redux update


markerikson:
Picked this up and added more changes:

- Keyboard events are now all dispatched in a single action
- Debounced React annotations too
- Mousedown events seem to come in multiple times but with duplicate values. Debounced that to only dispatch the last set.

PR #7497, opened 2022-08-04
New Console focus improvements

bvaughn:
- [x] ~~Disable Object Inspector inspection for objects in unloaded regions.~~ Make the new Console filter messages in unloaded regions _entirely_.
- [x] Add in-memory time-to-execution-point caching (and lookup) to better mimic how legacy app handles lookups within unloaded regions (and avoid potential "unloaded region" errors).
- [x] Update the prototype shell to show loading+loaded regions and to remotely load focus regions (to more closely mirror the real app).


packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -80,44 +83,49 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
...
-          isFocused={loggable === currentSearchResult}
-          terminalExpression={loggable}
-        />
-      );
-    } else {
-      throw Error("Unsupported loggable type");
+    const isLoaded =
+      loadedRegions !== null &&
+      isPointInRegions(getLoggableExecutionPoint(loggable), loadedRegions.loaded);
+    if (isLoaded) {
```
bvaughn, line 89: The new Console will deviate from the old Console here, by hiding loggables that aren't in loaded regions. This means that for longer recordings, the new Console _may_ render fewer things than the old Console.

PR #7503, opened 2022-08-05
Delete unused "Changes" folder and remove other dead code

markerikson:
This PR:

- Deletes the unused `changes` folder from the Inspector code
- Nukes several more files in `/inspector`
- Moves a few types and init values from `inspector/state/index` to `inspector/reducers/index`
  - Also adds a couple specific imports of `inspector/prefs`, because that is a stateful init file and moving that around changed import ordering such that other files threw errors when trying to read prefs that hadn't been initialized yet


src/devtools/server/actors/utils/event-breakpoints.js
```diff
@@ -2,7 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-"use strict";
+// Side-effectful import, needed to initialize prefs before read
+require("devtools/client/inspector/prefs");
```
jasonLaster, line 6: nice find. would be nice to consolidate at some point

jasonLaster (approved):
🔫

PR #7507, opened 2022-08-08
Fix 1 of 2 problems with react_devtools e2e test

bvaughn:
Looks like the e2e test was _actually working_ but the final selector was failing probably due to a minor whitespace issue (at least from what I'm seeing locally).

The page showed the text:
> Mounting your React application...
>
> Try picking a different point on the timeline.

The test selector was waiting to match:
```js
textContent === "Mounting your React application...Try picking a different point on the timeline."
```

Looking at the `textContent` though, it now contains _a lot more text_ (that's just been hidden by the Offscreen API):
```
'ExceptionsErrors1WarningsLogs4Common EventsOther EventsTimer7Hide Node ModulesShow TimestampsRewindApp.js:15 Initial listRewindApp.js:22 Added an entryCommentadd_commentApp.js:26 Removed an entryFast-forwardApp.js:28 Error: "Baz"e App.js:28c runtime.js:63_invoke runtime.js:294x runtime.js:119r asyncToGenerator.js:3u asyncToGenerator.js:25promise callback*r asyncToGenerator.js:13u asyncToGenerator.js:25promise callback*r asyncToGenerator.js:13u asyncToGenerator.js:25promise callback*r asyncToGenerator.js:13u asyncToGenerator.js:25l asyncToGenerator.js:32l asyncToGenerator.js:21e App.js:11p App.js:11p App.js:33React6Webpack12Fast-forwardApp.js:31 ExampleFinished​x ​xxxxxxxxxx ​xxxxxxxxxx ​Mounting your React application...Try picking a different point on the timeline.'
```

The fix is just to make the selector do a substring comparison.

I dug into the failure here:
https://www.loom.com/share/8fa12146d376433bbcd9ca9d554669a8?t=378

### DIFF
--- a/public/test/scripts/react_devtools.js
+++ b/public/test/scripts/react_devtools.js
@@ -71,7 +71,7 @@ Test.describe(`Test React DevTools.`, async () => {
 
   await Test.app.actions.seekToTime(0);
   await Test.waitUntil(
-    () => document.querySelector(".secondary-toolbox-content").textContent === "Mounting your React application...Try picking a different point on the timeline.",
+    () => document.querySelector(".secondary-toolbox-content").textContent.includes("Try picking a different point on the timeline"),
     { waitingFor: 'ReactDevTools to say "Try picking a different point"' },
   );
 });


markerikson (approved):

bvaughn:
Interesting!

CI still fails this test (even though it passes for me locally, after this change): http://localhost:8080/recording/reactdevtoolsjs--c1540c15-64ad-4b54-a740-bf0abd14c16a

jasonLaster (approved):

hbenl (commented):
This doesn't seem to fix anything: I've made a manual recording of the react_devtools test without this PR and it succeeded: https://app.replay.io/recording/reactdevtools-local-prod--1e235571-1aa5-4020-8515-1d138ca950a6
I've added comments to this recording and the failed CI recording (https://app.replay.io/recording/reactdevtoolsjs--c1540c15-64ad-4b54-a740-bf0abd14c16a) showing that in the successful recording, `createReactDevTools()` is called 3 times, which is what I would expect because we call it whenever we jump to a point with react annotations and we jump to 3 such points in the test. But in the failed recording, `createReactDevTools()` is only called twice (and even the connector for the `ReactDevtoolsPanel` isn't called for any other point!).
The problem may be the error "Timed out while inspecting element 6." that the react devtools frontend throws because after that error the connector isn't called anymore. That would mean that the react_devtools e2e test is failing due to poor browser performance in CI.

bvaughn:
> This doesn't seem to fix anything

@hbenl [I agree.](https://github.com/replayio/devtools/pull/7507#issuecomment-1208485264)

I think this change might be a necessary _part_ of the fix though, based on my own querying of the DOM when the test was hung, which I describe in the PR description.

I was originally worried that React might be dropping updates (maybe ones from Redux?) to _offscreen_ trees. When the React tab is not selected, the `Offscreen` component _hides_ it and then _re-shows it_ if you later select it. At that point, any pending updates that were scheduled should be run but didn't seem to be when I had first looked into this issue a couple of weeks ago.

Was leaving this PR open until I could dig in further. If you have a lead and would like to dig in though I'd love the help.

hbenl:
> I think this change might be a necessary part of the fix though

@bvaughn You're right, I hadn't looked at my manual recording properly, it was also failing (at a later point in the test).
So we have 2 issues that make this test fail: one of them is fixed by this PR and the other one is that react doesn't rerender `ReactDevtoolsPanel` (and doesn't even call its connector) even though `currentPoint` changed in the redux store.
My theory is still that it's the error from the react devtools frontend that breaks rerendering, although I have no idea how that is possible and how to investigate it.
Thinking about this some more I realized that the error from the react devtools frontend is probably _not_ related to the browser's performance: the react devtools frontend runs in the browser, but it times out waiting for a reply from the react devtools backend, which runs in the replay backend. So maybe this perf problem is sporadic and it was just a coincidence that it failed in this CI run and succeeded when running locally.

hbenl (approved):

bvaughn:
> My theory is still that it's the error from the react devtools frontend that breaks rerendering, although I have no idea how that is possible and how to investigate it.

Hmm, I don't think that matches what I saw when I looked a couple of weeks ago. I _think_ I saw that our `ReactDevTools` wrapper component was never getting rendered after the Redux selector scheduled an update while the "React" tab was hidden (offscreen). Could be misremembering though.

PR #7508, opened 2022-08-08
Toggle theme shortcut now ALT+Shift+T

bvaughn:
Hopefully this will avoid the number of times it's accidentally toggled while focus hasn't yet changed to the comment input.

### DIFF
--- a/src/ui/components/KeyboardShortcuts.tsx
+++ b/src/ui/components/KeyboardShortcuts.tsx
@@ -154,7 +154,7 @@ function KeyboardShortcuts({
 
       // Should be ignored when an editable element is focused
       "Shift+C": addComment,
-      "Shift+T": toggleTheme,
+      "Alt+Shift+T": toggleTheme,
       "Shift+F": toggleEditFocusMode,
 
       // Quick Open-related toggles


jonbell-lot23 (commented):
We do two things:

1. Make this a user setting
2. Allow it to be keyboard accessible

We have some work to do:

1. Our comment box should never capture keystrokes -- this isn't just a shift-T issue, it will happen with other shortcuts today and in the future. So that should be its own task.

Here's what I propose here:

1. Don't kill the shortcut outright, just make it harder to press. I propose shift-option-T. I use this feature constantly, and need it for the sort of work I'm doing, where I do indeed need to toggle back and forth quickly. (On that note, we currently have dark/light theme bugs slip in, so we should be trying to do these tests more)

bvaughn:
> Make this a user setting

To be clear, this is already a user setting. There has always been a theme preference (light, dark, or system).

jonbell-lot23:
> > Make this a user setting
> 
> To be clear, this is already a user setting. There has always been a theme preference (light, dark, or system).

Correct, that was listed under the header "What we do" to differentiate it from "We have some work to do" and "What I propose here."

PR #7509, opened 2022-08-08
new console design cleanup

souporserious:
First cleanup pass for the new console that moves the checkboxes to a new shared `Checkbox` component and styles the console layout similar to the recent updates we did for the old console. In a follow-up pass I'll work on moving the rewind/fast-forward button into a shared component.

closes FE-520

## Design System Beginnings

This PR also starts a simple design system under `packages/design` starting with `Checkbox` and `Badge` components. This is where we can continue to push new components and move over old components. I'll work on getting a README and simple docs setup for this in a follow-up PR.


markerikson:
Toss in a couple before/after screenshots?

souporserious:
@markerikson my bad, I should have marked as draft, getting Earthly working locally so I should have a screenshot diff here soon!

packages/bvaughn-architecture-demo/components/console/filters/EventType.tsx
```diff
@@ -3,6 +3,8 @@ import { ConsoleFiltersContext } from "@bvaughn/src/contexts/ConsoleFiltersConte
 import { Event } from "@bvaughn/src/suspense/EventsCache";
 import { useContext } from "react";
 
+import Checkbox from "../../../../../src/ui/components/shared/Forms/Checkbox";
```
souporserious, line 6: I was struggling getting alias paths working, but I need to move the `Checkbox` component up into the `packages` directory anyways which should clean this up.

src/ui/components/shared/Forms/Checkbox.tsx
```diff
@@ -4,12 +4,14 @@ import React from "react";
...
   id,
   checked,
+  disabled,
   onChange,
   className,
 }: React.HTMLProps<HTMLInputElement>) {
   return (
     <input
       type="checkbox"
+      disabled={disabled}
```
bvaughn, line 14: Btw I don't think this component is used in the new code at all. The change seems fine though, so long as it was intentional.

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
```diff
@@ -67,23 +71,27 @@ export default function ConsoleRoot({
...
-                </Offscreen>
-                <div className={styles.MessageColumn}>
-                  <ErrorBoundary>
-                    <Suspense fallback={<Loader />}>
-                      <MessagesList ref={messageListRef} />
-                    </Suspense>
-                  </ErrorBoundary>
 
-                  {terminalInput}
+              <FilterText />
```
bvaughn, line 75: This change looks like it might be a regression? The `DeleteTerminalExpressionButton` should be far-right aligned.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.module.css
```diff
@@ -29,7 +29,9 @@
 
 .Checkbox {
   margin: 0;
-  margin-right: 0.5ch;
+  margin-right: 0.3rem;
```
bvaughn, line 32: Can we keep our grid in increments of 0.25 (or 0.125 in a pinch)?

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
```diff
@@ -5,6 +5,7 @@ import { getMessages } from "@bvaughn/src/suspense/MessagesCache";
 import camelCase from "lodash/camelCase";
 import React, { Suspense, useContext, useMemo } from "react";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
+import Checkbox from "../../../../../src/ui/components/shared/Forms/Checkbox";
```
bvaughn, line 8: This won't work. Surprised it didn't fail to compile.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
```diff
@@ -114,12 +115,11 @@ function Toggle({
   const id = `FilterToggle-${camelCase(label)}`;
   return (
     <div className={styles.Filter}>
-      <input
-        className={styles.Checkbox}
+      <Checkbox
+        className="mr-1"
```
bvaughn, line 119: Let's not use any global class names.

bvaughn (changes requested):

src/ui/components/shared/Forms/Checkbox.tsx
souporserious: Yeah, this was the old `Checkbox` I was trying to import. I moved things to a new one, but will leave this since it doesn't harm anything.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.module.css
souporserious: Yeah, for sure! Updated things to use local styles that use these increments.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
souporserious: So I added a new Checkbox component under `packages/components`, but it's failing because NextJS can't handle importing CSS outside of the directory: 

image

Wondering if I should move it under your package for now and then we can share later?

packages/bvaughn-architecture-demo/components/console/filters/EventType.tsx
```diff
@@ -3,7 +3,7 @@ import { ConsoleFiltersContext } from "@bvaughn/src/contexts/ConsoleFiltersConte
 import { Event } from "@bvaughn/src/suspense/EventsCache";
 import { useContext } from "react";
 
-import Checkbox from "../../../../../src/ui/components/shared/Forms/Checkbox";
+import { Checkbox } from "../../../../components";
```
bvaughn: This import still won't work because it's importing from _outside of the `packages/bvaughn-architecture-demo` directory.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
```diff
@@ -5,7 +5,7 @@ import { getMessages } from "@bvaughn/src/suspense/MessagesCache";
 import camelCase from "lodash/camelCase";
 import React, { Suspense, useContext, useMemo } from "react";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
-import Checkbox from "../../../../../src/ui/components/shared/Forms/Checkbox";
+import { Checkbox } from "../../../../components";
```
bvaughn: Likewise.

packages/bvaughn-architecture-demo/components/console/filters/EventType.tsx
bvaughn: Just to be clear: the only package that `packages/bvaughn-architecture-demo` can import from (other than itself) is `packages/shared`.

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
souporserious: Using CSS Grid to place things 😁, but please let me know if that broke something! I think it should all be good from what I tested.

bvaughn: Ah, I see. Okay!

packages/bvaughn-architecture-demo/components/Icon.tsx
```diff
@@ -62,7 +62,7 @@ export default function Icon({
...
...C4.38711 4.97773 3.95592 5.41148 3.65306 5.97613C3.35021 6.54078 3.19878 7.21322 3.19878 7.99346V14.0917C3.19878 14.8719 3.35277 15.5443 3.66076 16.109C3.97389 16.6736 4.40251 17.1074 4.94662 17.4102C5.49587 17.7131 6.12725 17.8645 6.84076 17.8645H7.05635V20.0821C7.05635 20.3592 7.12821 20.58 7.27194 20.7442C7.4208 20.9085 7.621 20.9906 7.87252 20.9906ZM8.03422 11.0426C8.03422 10.7756 8.11891 10.5575 8.28831 10.3881C8.46284 10.2136 8.67843 10.1263 8.93509 10.1263H10.9062V8.15516C10.9062 7.8985 10.9909 7.68547 11.1603 7.51608C11.3297 7.34155 11.5453 7.25429 11.8071 7.25429C12.0689 7.25429 12.2845 7.34155 12.4539 7.51608C12.6284 7.68547 12.7157 7.8985 12.7157 8.15516V10.1263H14.6945C14.9512 10.1263 15.1642 10.2136 15.3336 10.3881C15.503 10.5575 15.5877 10.7756 15.5877 11.0426C15.5877 11.2992 15.503 11.5148 15.3336 11.6893C15.1642 11.8587 14.9512 11.9434 14.6945 11.9434H12.7157V13.9146C12.7157 14.1712 12.6284 14.3842 12.4539 14.5536C12.2845 14.723 12.0689 14.8077 11.8071 14.8077C11.5453 14.8077 11.3297 14.723 11.1603 14.5536C10.9909 14.3842 10.9062 14.1712 10.9062 13.9146V11.9434H8.93509C8.67843 11.9434 8.46284 11.8587 8.28831 11.6893C8.11891 11.5148 8.03422 11.2992 8.03422 11.0426Z";
```
bvaughn, line 65: Ooo, I'm curious where this SVG came from. It doesn't look like it quite fits the 24x24 view box as tightly as the others:
Screen Shot 2022-08-10 at 10 34 22 AM

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
```diff
@@ -42,8 +42,12 @@ export default function ConsoleRoot({
             messageListRef={messageListRef}
             showSearchInputByDefault={showSearchInputByDefault}
           >
-            <div className={styles.ConsoleRoot} data-test-id="ConsoleRoot">
-              <div className={styles.TopRow}>
+            <div
+              className={styles.ConsoleRoot}
+              data-menu-open={isMenuOpen}
```
bvaughn, line 47: I'm curious why you used a data attribute here rather than a different `className`? It's not a big deal, but the latter would be more consistent with how the other styles work in the package.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.tsx
```diff
@@ -72,7 +72,10 @@ export default function MessageHoverButton({
           ref={ref}
         >
           <Icon className={styles.AddCommentButtonIcon} type="comment" />
-          <span className={isHovered ? styles.LabelHovered : styles.Label}> Add comment </span>
+          <span className={isHovered ? styles.LabelHovered : styles.Label}>
+            {" "}
+            Add comment &nbsp;
```
bvaughn, line 77: This feels a little hacky. Why two spaces on the right side?

Screen Shot 2022-08-10 at 10 50 35 AM

If you're aiming to horizontally balance the padding on the right side with the icon on the left, can we just use `padding-right`?

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
```diff
@@ -119,17 +121,14 @@ function Toggle({
   const id = `FilterToggle-${camelCase(label)}`;
   return (
     <div className={styles.Filter}>
-      <input
-        className={styles.Checkbox}
+      <Checkbox
+        data-test-id={id}
+        label={label}
+        secondaryLabel={count !== null && count > 0 && count}
```
bvaughn, line 127: A `secondaryLabel` attribute feels very not-checkbox-like.

I think we should make the `label` prop accept a `ReactNode` (which includes strings) and then if we want to have a string + badge, that can just be passed in as the `label`.

packages/design/Checkbox.tsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  return label ? (
+    <label className={classNames(styles.Label, className)} {...props}>
+      <input {...checkboxProps} className={styles.Input} />
+      <span>{label}</span>
+      {secondaryLabel ? <Badge label={secondaryLabel} /> : null}
+    </label>
+  ) : (
+    <input {...checkboxProps} {...props} className={classNames(styles.Input, className)} />
```
bvaughn, line 39: IIR, the reason I was using the `htmlFor` attribute was that the text didn't truncate with "..." nicely if the input and display text were both inside of a `<label>`. I assume you've got that sorted out though so cool 😄

packages/bvaughn-architecture-demo/pages/variables.css
```diff
@@ -102,6 +107,11 @@
   --color-secondary-button: #ffffff;
   --color-warning: #6c5914;
 
+  --color-primary-background: #ffffff;
```
bvaughn, line 110: nit
```suggestion
  --color-primary-background: var(--background-color-contrast-1);
```

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
```diff
@@ -29,6 +31,7 @@
 .ConsoleMessageHoverButtonIcon {
   width: 1.25rem;
   height: 1.25rem;
+  padding: 0.0625rem;
```
bvaughn, line 34: This looks sketchy.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
```diff
@@ -1,11 +1,13 @@
 .AddCommentButton,
 .ConsoleMessageHoverButton {
+  font-family: "Inter", sans-serif;
+  font-weight: 600;
```
bvaughn, line 4: Ooo, hmm. Can we avoid using one-off `font-family` and `font-weigh` styles like this? Seems like we should have _one_ uniform font for components (well one for mono and one for not-mono).

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
```diff
@@ -1,11 +1,13 @@
...
   padding: 0 0.125rem;
   position: absolute;
   left: 0.25rem;
   top: 0.25rem;
   height: 1.5rem;
-  border-radius: 0.25rem;
+  border-radius: 3rem;
```
bvaughn, line 10: I'm not a fan of this change. I think the button looked significantly better before.

### Before
image

### After
image

The new layout feels cramped and vertical alignment looks a bit off. (To be fair it looks a bit misaligned in both of them, but it feels less noticeable in the less cramped before style.)

packages/design/Checkbox.tsx
bvaughn: Actually this looks like a regression.
Kapture 2022-08-10 at 10 53 32

packages/bvaughn-architecture-demo/components/Icon.tsx
souporserious: This is our current add comment icon. I tightened it up a little bit on the bounds, but I'll be going through all of our icons after this ships so it will be changing here soon. I want to normalize all of our icons in design and dev and get things down to one component and system.

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
souporserious: Mostly because it felt awkward to use BEM style syntax in CSS modules, but moved it to a simple class name so things are consistent 😎

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.tsx
souporserious: Yeah, this was weird, I was having trouble getting padding to work yesterday, but got things fixed up.

packages/bvaughn-architecture-demo/components/Icon.tsx
```diff
@@ -62,7 +62,7 @@ export default function Icon({
...
...47359 3 4.68875 3.18357 4.03223 3.55071C3.38168 3.91173 2.88033 4.42879 2.5282 5.10188C2.17607 5.77497 2 6.57656 2 7.50665V14.776C2 15.7061 2.17905 16.5077 2.53715 17.1808C2.90122 17.8539 3.39958 18.371 4.03223 18.732C4.67084 19.093 5.40495 19.2735 6.23456 19.2735H6.48523V21.9169C6.48523 22.2474 6.56879 22.5105 6.7359 22.7063C6.90898 22.9021 7.14175 23 7.4342 23ZM7.6222 11.1413C7.6222 10.8232 7.72068 10.5631 7.91764 10.3612C8.12056 10.1531 8.37123 10.0491 8.66965 10.0491H10.9615V7.6994C10.9615 7.39345 11.06 7.13951 11.2569 6.93759C11.4539 6.72954 11.7046 6.62552 12.009 6.62552C12.3133 6.62552 12.564 6.72954 12.761 6.93759C12.9639 7.13951 13.0654 7.39345 13.0654 7.6994V10.0491H15.3662C15.6646 10.0491 15.9123 10.1531 16.1092 10.3612C16.3062 10.5631 16.4047 10.8232 16.4047 11.1413C16.4047 11.4473 16.3062 11.7043 16.1092 11.9123C15.9123 12.1143 15.6646 12.2152 15.3662 12.2152H13.0654V14.5649C13.0654 14.8709 12.9639 15.1248 12.761 15.3268C12.564 15.5287 12.3133 15.6296 12.009 15.6296C11.7046 15.6296 11.4539 15.5287 11.2569 15.3268C11.06 15.1248 10.9615 14.8709 10.9615 14.5649V12.2152H8.66965C8.37123 12.2152 8.12056 12.1143 7.91764 11.9123C7.72068 11.7043 7.6222 11.4473 7.6222 11.1413Z";
```
bvaughn, line 65: Thanks! I think this is a better fit, although there's still a bit of vertical misalignment:
Screen Shot 2022-08-10 at 1 39 02 PM

Mostly mentioning as an FYI.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
souporserious: Can you expand more on why it doesn't feel checkbox-like?

The `label` prop currently accepts a `ReactNode`. This takes care of some the styling concerns around placing the secondary child. As always in API design I struggled a little with a name here. We could potentially use `addon` and `element` nomenclature similar to [ChakraUI](https://chakra-ui.com/docs/components/input/usage#left-and-right-addons) and other libraries like Bootstrap.

I'd like to tackle this in a follow-up though so it doesn't block anything if that's cool 😇

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
bvaughn: Not sure if you missed this comment but this still seems like a regression to me.

packages/bvaughn-architecture-demo/components/Icon.tsx
bvaughn: Sure, sure. No problem with the shape of the icon at all. My concern was that it doesn't "fit" the view box very well, which means it's going to look odd (smaller) relative to the other icons

packages/design/Checkbox.tsx
souporserious: Oops, yeah I'll get this working again! The span should be able to handle that.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
souporserious: Half of 0.125 😄 I'll work on getting a spacing scale in so these values are at least tied to a system. We'll need these smaller values though since we're dealing a with pretty condensed UI.

packages/bvaughn-architecture-demo/pages/variables.css
souporserious: 👍 I'll be in here soon for more cleanup. I want to do a full pass on colors and get things sorted, probably after icon cleanup.

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
bvaughn: Conceptually, checkboxes have attributes like: label, checked, change, disabled, etc.

The idea of "secondary label" is a thing that doesn't really have anything to do with a checkbox.

Just to be clear, all I'm asking is that we change:
```js
type CheckboxProps = {
  label: ReactNode;
  // ...
}
```

And then you can use it like:
```js
<Checkbox
  data-test-id={id}
  label={
    <>
      {label}
      {count !== null && <span className="...">{count}</span> > 0 && count}
  </>
  }
```

And we don't _leak_ in any non-standard concepts like "secondary label".

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
souporserious: Yeah, more design systems work 😅 would like to tackle this in a follow-up if that's ok.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
souporserious: Just pushed up a revision of this that's a middle ground and matches what we currently do. I'd like the design to be as similar as possible to our current aesthetic for now and we can follow-up and make tweaks afterwards.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
bvaughn: Can we just remove the font change for now then until we're ready to tackle the large font design stuff?

I know I'm being picky but I'd rather not leak one-off font families into the new CSS.

packages/bvaughn-architecture-demo/components/Icon.tsx
souporserious: Ah geeze lol fixed! 😊

bvaughn: Thanks 😆

packages/bvaughn-architecture-demo/components/console/filters/FilterToggles.tsx
souporserious: Yeah, totally get the name can be confusing. I renamed `secondaryLabel` to `addon` for now, we could use `children`, but that feels a little weird to me. Going the `label` prop as the only route won't work, or at least will require managing layout each time which is what the `addon` prop handles.

bvaughn: > Going the `label` prop as the only route won't work, or at least will require managing layout each time which is what the `addon` prop handles.

Could you explain?

My concern is that _the concept_ "primary" and "secondary" labels is _specific_ to the Console filters. It's not something a generic checkbox component should concern itself with.

To me, the main question is: What would you expect to happen if you clicked on the count (the "secondary label"?)
* If you'd expect nothing, then it shouldn't be passed to the checkbox at all.
* If you'd expect the checkbox to be toggled, then it should just be part of the "label".

Making the `label` prop a `ReactNode` (or using `children` – same thing really) leaves the flexibility of how to style the custom stuff (like the category counts) to the caller, and hopefully prevents us from adding more one-off props in the future.

In this case, it also gives us the flexibility to decide that the label count should truncate (with "...") but the count shouldn't. I think that's also kind of context specific.

packages/bvaughn-architecture-demo/components/console/MessageHoverButton.module.css
souporserious: No worries! I set to a sans-serif for now so it is at least closer to our other buttons and not monospace, but we can clean this up and do a proper pass on fonts.

bvaughn (approved):

PR #7511, opened 2022-08-08
Thread.listenForLoadChanges only calls Session.listenForLoadChanges once

bvaughn:
So long as `Thread.listenForLoadChanges` exists (and seems to be the preferred way to add listeners for `client.Session.addLoadedRegionsListener`) then it should avoid calling `client.Session.listenForLoadChanges` unnecessarily and it should avoid triggering previously added listeners each time a new listener is added.

---

```js
app.threadFront.listenForLoadChanges((...args) => {
  console.log("load changes 1", ...args);
});

setTimeout(() => {
  app.threadFront.listenForLoadChanges((...args) => {
    console.log("load changes 2", ...args);
  });
}, 1000);

setTimeout(() => {
  app.threadFront.listenForLoadChanges((...args) => {
    console.log("load changes 3", ...args);
  });
}, 2000);
```

### Before
```
load changes 1 {loaded: Array(1), loading: Array(1), indexed: Array(1)}

load changes 1 {loaded: Array(1), loading: Array(1), indexed: Array(1)}
load changes 2 {loaded: Array(1), loading: Array(1), indexed: Array(1)}

load changes 1 {loaded: Array(1), loading: Array(1), indexed: Array(1)}
load changes 2 {loaded: Array(1), loading: Array(1), indexed: Array(1)}
load changes 3 {loaded: Array(1), loading: Array(1), indexed: Array(1)}
```

### After
```
load changes 1 {loaded: Array(1), loading: Array(1), indexed: Array(1)}

load changes 2 {loaded: Array(1), loading: Array(1), indexed: Array(1)}

load changes 3 {loaded: Array(1), loading: Array(1), indexed: Array(1)}
```

---

Maybe it's debatable how much value `ThreadFront` is providing in cases like this, but as long as it exists and seems to be the preferred way to interact with APIs like this...I guess this is a net positive change?

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -38,7 +38,6 @@ import {
   responseBodyData,
   requestBodyData,
   findAnnotationsResult,
-  getHitCountsParameters,
   Frame,
   PointRange,
   TimeRange,
@@ -312,9 +311,7 @@ class _ThreadFront {
       client.Session.addLoadedRegionsListener((loadedRegions: LoadedRegions) => {
         this._mostRecentLoadedRegions = loadedRegions;
 
-        if (loadedRegions.loading) {
-          this.loadingHasBegun.resolve();
-        }
+        this.loadingHasBegun.resolve();
 
         this._loadedRegionsListeners.forEach(callback => callback(loadedRegions));
       });
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -29,7 +29,7 @@ import {
   SourceLocation,
   TimeStamp,
   unprocessedRegions,
-  loadedRegions,
+  loadedRegions as LoadedRegions,
   SameLineSourceLocations,
   RequestEventInfo,
   RequestInfo,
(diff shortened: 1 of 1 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -296,17 +297,34 @@ class _ThreadFront {
...
+
+    if (!this._listeningForLoadChanges) {
+      this._listeningForLoadChanges = true;
+
+      const sessionId = await this.waitForSession();
+
+      client.Session.addLoadedRegionsListener((loadedRegions: LoadedRegions) => {
+        this._mostRecentLoadedRegions = loadedRegions;
+
+        if (loadedRegions.loading) {
```
loganfsmyth, line 315: I realize this is existing logic, but just to mention it...I think this is always `true`, since `loading` is just an array.

loganfsmyth (approved):

markerikson (approved):

packages/protocol/thread/thread.ts
bvaughn: Ooo ok good point! I'll clean that up too then while I'm in here.

PR #7512, opened 2022-08-08
Socket.removeEventListener() supports (optional) explicit handler param

bvaughn:
The generated `@replayio/protocol` client will soon pass a callback param to both `addEventListener` and `removeEventListener` params to enable multiple listeners to be attached per event type. The `Socket` class in the frontend should take advantage of this then and only remove all listeners for a type in the event that no callback param is included.

Depending on the timing of this PR (and BAC-2052) we might be able to update the `@replayio/protocol` dependency to `0.31.1` and remove the undefined param check.

### DIFF
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -290,13 +290,38 @@ export function addEventListener<M extends EventMethods>(
   }
 }
 
-// TODO This method should have the same signature as addEventListener.
-// It should only remove a specific event handler.
-export function removeEventListener<M extends EventMethods>(event: M) {
-  gEventListeners.delete(event);
+export function removeEventListener<M extends EventMethods>(
+  event: M,
+  handler?: (params: EventParams<M>) => void
+) {
+  switch (event) {
+    case "Analysis.analysisPoints":
+    case "Analysis.analysisResult": {
+      gEventListeners.delete(event);
+      break;
+    }
+    default: {
+      if (handler != null) {
+        const handlers = gEventToCallbacksMap.get(event);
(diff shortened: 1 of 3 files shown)


packages/protocol/socket.ts
```diff
@@ -290,13 +290,38 @@ export function addEventListener<M extends EventMethods>(
...
+export function removeEventListener<M extends EventMethods>(
+  event: M,
+  handler?: (params: EventParams<M>) => void
+) {
+  switch (event) {
+    case "Analysis.analysisPoints":
+    case "Analysis.analysisResult": {
+      gEventListeners.delete(event);
+      break;
+    }
```
bvaughn, line 302: I'm kind of afraid to touch the special casing around analysis stuff because I haven't dug in enough to understand it, so for now I'm leaving its behavior as-is.

packages/protocol/socket.ts
```diff
@@ -290,13 +290,38 @@ export function addEventListener<M extends EventMethods>(
...
+        if (handlers == null || handlers.length === 0) {
+          gEventListeners.delete(event);
+        }
+      } else {
+        gEventListeners.delete(event);
+
+        if (gEventToCallbacksMap.has(event)) {
+          gEventToCallbacksMap.delete(event);
+        }
+      }
```
bvaughn, line 322: Once we update the `@replayio/protocol` package, we can remove this fallback path and mark the param as required.

markerikson (approved):

hbenl (approved):

PR #7513, opened 2022-08-08
Kill `RulesView`

markerikson:
This PR:

- Rewrites `reducers/rules` to use RTK
  - Rewrote reducer to use `createSlice`
  - Moved rules state typedefs into the reducer
  - Deleted dead actions
- Removes dead fields from `RulesView` and `ElementStyle`
- Actually kills `RulesView`, using the same patterns as the other inspector conversions
  - selection listening is now done in `inspector/rules/actions/rules`
  - that file loads the element style data and dispatches actions


jasonLaster (approved):
This is much simpler!

PR #7514, opened 2022-08-09
Optimize sources processing perf on start

markerikson:
This PR:

- Speeds up the processing in `newSourcesToCompleteSourceDetails`:
  - Used object destructuring and manual lookup tables instead of Lodash
  - Changed `.forEach` to be `for...of` loops
  - Created lookup tables in a single loop instead of multiples
  - Used some `Object.assign` instead of spreads
- Speeds up `SourcesTree`'s `mapState`, by using an RTK `selectTotal` selector instead of calling `Object.values(sources).length` every time
- Consolidates the two `addSources` and `allSourcesReceived` actions that we were dispatching back-to-back, and removes storing the original source data in Redux (which we never read or process anywhere ever again anyway)

Based on local perf profiling, what I'm seeing is that the overall time to execute `setupDebugger` dropped from about 675ms to about 210ms.  More specifically, `newSourcesToCompleteSourceDetails` dropped from about 240ms to 105ms, and `SourcesTree.mapState` dropped from 250ms to, uh... small enough I can't see it in the profiler any more :)

### DIFF
--- a/src/ui/utils/sources.ts
+++ b/src/ui/utils/sources.ts
@@ -1,8 +1,5 @@
 import { EntityId } from "@reduxjs/toolkit";
 import { newSource, SourceKind } from "@replayio/protocol";
-import groupBy from "lodash/groupBy";
-import omit from "lodash/omit";
-import keyBy from "lodash/keyBy";
 import { SourceDetails } from "ui/reducers/sources";
 import newGraph from "./graph";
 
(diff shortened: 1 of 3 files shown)


src/ui/utils/sources.ts
```diff
@@ -86,36 +100,40 @@ export const newSourcesToCompleteSourceDetails = (
...
-      generatedFrom: generated.to(source.sourceId) || [],
-      canonicalId: findCanonicalId(source.sourceId),
-    });
-  });
+  }
+
+  for (let source of newSources) {
+    const { sourceId, generatedSourceIds, ...remainingFields } = source;
+
+    returnValue[sourceId] = fullSourceDetails(
```
hbenl, line 124: Why don't we get rid of `fullSourceDetails()`? All it seems to do is fill in some default values which we could do directly here.

hbenl (approved):

src/ui/utils/sources.ts
markerikson: sure, I'll do that now

markerikson: Hah, ironically all of those _are_ being filled in already :)

PR #7520, opened 2022-08-09
DO NOT MERGE

bvaughn:
Test removing Offscreen component to see if it fixes React DevTools e2e test


bvaughn:
Sure enough this "fixes" the test but that doesn't tell much really.

PR #7525, opened 2022-08-09
Kill `Inspector`!!!

markerikson:
This PR:

- Converts the `Popup` component (used for hover previews in the editor) to TS
- Removes all remaining references to `inspector` and `window.gInspector` variables
- Actually kills the `inspector.ts` file! 🎉 
- Adds a `max-height` to the `Popup` CSS to force it to scroll if the preview size is large


src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -268,10 +268,11 @@ export function selectNode(nodeId: string, reason?: SelectionReason): UIThunkAct
     const nodeFront = ThreadFront.currentPause?.getNodeFront(nodeId);
     if (nodeFront) {
       Highlighter.highlight(nodeFront, 1000);
-      // HACK This is ugly, but we lazy-load the component and also try to use `window.gInspector` in places.
+      // HACK This is ugly, but we lazy-load the component.
       // So, ensure it's loaded, _then_ use this global
```
bvaughn, line 272:
```js
// So, ensure it's loaded, _then_ use this global
```
This comment isn't needed anymore, is it?

bvaughn (approved):
Just skimmed

src/devtools/client/inspector/markup/actions/markup.ts
markerikson: Probably not, lemme tweak that

PR #7527, opened 2022-08-09
Kill multiple unused chunks of code from `devtools/client/shared`

markerikson:
This PR:

- Nukes a bunch more dead files/folders from `devtools/client/shared`


jasonLaster (approved):
# 🔫

PR #7530, opened 2022-08-09
Standardise testsuite icons and make them more accessible for color-blindness

jonbell-lot23:
Addresses issue https://linear.app/replay/issue/DES-47/successfail-icons-in-test-suites-are-too-similar-bad-for-accessibility which was brought up by @jazzdan .

See also: [Discord thread where we discussed some ideas](https://discord.com/channels/779097926135054346/1006674024950140928/1006674027445747832)

--

image

Previously, we were mixing "outline" and "solid" styles between success and failure. The most notable example of this is how we were using the same play icon for success/fail, which can cause accessibility issues. By simply making success solid and failure outlined, it means we're not only using colour to distinguish between the two.

Now it looks like this, which is more consistent and easier to tell apart:

image

--

**FAQ**

Q: Why make success solid and failure outlined? Could we reverse them?
A: We could, but in my experiments it felt better this way. I like the idea that something isn't fully there (outline visual) when failing, then on success it's a fully rendered icon (solid).  But the key issue is making sure they look different in more ways than just color, and this PR achieves that.

Q: Why were we only using color to distinguish these icons?
A: It wasn't an issue in the primary view because we were using grouping. (Everything failing is on the top, and everything succeeding is down below.) But there are other views (such as the Test Results view) that will have different sorting, so this makes the visual language more robust and clear for different scenarios we'll be adding.


bvaughn:
Hey just a heads up that this commit broke the `main` branch. Hold off on merging if you ever see failures in either of the following CI tasks:
* Lint / Trunk Check runner
* Typecheck / typecheck

src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
```diff
@@ -7,23 +7,30 @@ import styles from "../../../../Library.module.css";
...
-      outlined
-      className={
-        passed
-          ? "text-green-500 group-hover:text-green-700"
-          : "text-red-500 group-hover:text-red-700"
-      }
-    >
-      play_circle
-    </MaterialIcon>
+    <Link href={`/recording/${recordingId}`}>
```
bvaughn, line 10: This variable is undefined.

jonbell-lot23: Yeah, I'm not sure what happened but it appears to have been a messy merge/collision with Jaril code in the same hour?

I spent time trying to unwrap it and I'm a little confused about the state of things. I closed this PR before this comment but it's possible it's still causing trouble somewhere? If so a) sorry b) feel free to fix/roll back.

bvaughn: Hm, not sure. The commit in this PR looks like it was a little broken (there was no `recordingId` variable) but at least it was an easy fix :)

PR #7532, opened 2022-08-10
Source reducer exploration

hbenl:


hbenl:
Fixes FE-552

hbenl:
Fixes FE-540

packages/protocol/thread/thread.ts
```diff
@@ -46,11 +46,13 @@ import {
 } from "@replayio/protocol";
 import groupBy from "lodash/groupBy";
 import uniqueId from "lodash/uniqueId";
+import { getAlternateSourceId, getPreferredSourceId } from "ui/reducers/sources";
```
markerikson, line 49: :hand: conceptually, this ends up dragging in a bunch of the "app" code.

Could we switch this up to injecting these via setter into `ThreadFront` as the app starts?

I know it's a bit silly, but it'd be nice to maintain that separation for now.

markerikson: That said: I've also just now realized that `ThreadFront` is used over in Brian's prototype, which has _no_ Redux store at all. So, that prototype is going to completely break because of this PR.

markerikson (commented):
At least as _intended_, the changes seem okay.

**BUT**... per discussion in Discord, we've realized that this actually breaks Brian's prototype package entirely.

`packages/shared` relies on `packages/protocol`, which in turn is now both importing part of the main app's Redux store logic, and also needing a Redux store with appropriate setup to actually exist.

The import chain is an issue, but so is the implicit dependence on a valid Redux store, especially since Brian's prototype doesn't have _any_ Redux store at all.

I think it's best to split this up and try to land _just_ the `sources` reducer changes by themselves (new sources processing, selectors, etc).

Then we should figure out how to rip _all_ sources handling out of `ThreadFront` entirely and put it somewhere else.

src/ui/utils/sources.test.ts
```diff
@@ -94,27 +94,27 @@ describe("newSourcesToCompleteSourceDetails", () => {
       ])
     ).toEqual({
       "1": {
-        canonicalId: "1",
         contentHash: "contentHash#1",
         correspondingSourceIds: ["1"],
         generated: [],
-        generatedFrom: [],
+        generatedFrom: ["pp1"],
```
jcmorrow, line 100: Personally, I really don't like this. `generatedFrom` should mean that we have a sourcemap linking source `pp1` with the generated source `1`, and that is how we computed the content of `pp1`. Generated and original sources only make sense in the context of source-mapping, and that is not what pretty-printing is (or, if it was, it would be correct to call the pretty-printed file the `generated`, not the `original`). I'm not going to block on it, because I am not doing much with this code these days, but I find this a major source of confusion. I think if we continue to completely reverse the meanings of these terms in some contexts we will also continue to completely confuse the daylight out of every single person who ever tries to understand the code.

src/ui/utils/sources.ts
```diff
@@ -1,118 +1,63 @@
...
-      current = nextNode;
-      nextNode = canonical.from(current)?.[0];
-    }
-
-    return current;
-  };
-
-  const generated = newGraph();
-
-  // We'll iterate through the whole list once, and perform three setup operations per source:
```
jcmorrow, line 34: This code is incredibly confusing, even for me, and I wrote it. I think it's fine if we rework it to be more efficient, but not at the cost of removing all explanations of how it works. I would like to understand:

1) Are these changes more performant? How much more?
2) What did we change about how we process sources?
3) Can we leave some explanation about the trickier parts of this code? It can live in the commit, or in separate documentation, or inline, but I think future us will need it somewhere, at some point.

src/ui/utils/sources.ts
```diff
@@ -1,118 +1,63 @@
...
   const keyForSource = (source: newSource): string => {
     return `${source.kind}:${source.url!}:${contentHashForSource(source)}`;
   };
 
   for (let source of newSources) {
-    const key = keyForSource(source);
-    if (corresponding[key] === undefined) {
-      corresponding[key] = [];
+    newSourcesById[source.sourceId] = source;
+  }
```
jcmorrow, line 32: If we don't need to do the corresponding check I think it would be easier to read this as
`const newSourcesById = groupBy(newSources, "sourceId)` but I don't feel strongly about it.

jcmorrow: Failing making that change we should at least have line 30 be `for (const ...` rather than `for (let ...`

packages/protocol/thread/thread.ts
hbenl: > Could we switch this up to injecting these via setter into ThreadFront as the app starts?

Done.

src/ui/utils/sources.test.ts
hbenl: We have 3 different reasons for original <-> generated links between sources: sourcemaps, pretty-printing and inline script extraction. We could store these 3 types of links separately but in our code we mostly use these links to determine which source is preferred for display - separating the links would make these parts of our code (slightly) more complicated while I can't think of a single place where it would help make the code simpler.
If we want to avoid confusion I think we should use different terms for these links but not make working with them in the code harder.

src/ui/utils/sources.ts
hbenl: `groupBy` would create a `Dictionary<newSource[]>`, but we know that there is only one source per sourceId, so I prefer not using that (though I don't feel strongly about it either).
I'll replace all `for (let ...` with `for (const ...`.

src/ui/utils/sources.ts
hbenl: 1. I didn't rewrite this to be more performant but to be less confusing. I think it's rather straightforward now.
2. no more `canonicalId` and the links between pretty-printed and minified sources are not removed from `generated`/`generatedFrom` anymore
3. I've added a few comments and could add more if you tell me which parts you find confusing

src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
```diff
@@ -89,7 +89,7 @@ describe("useConsoleSearch", () => {
     expect(lastCommitedSearchState?.results.map(result => result.id)).toMatchInlineSnapshot(`
       Array [
         "1",
-        "3",
+        "2",
```
markerikson, line 92: :question: How did this snapshot change?

markerikson (approved):
Looks pretty good overall!

src/devtools/client/webconsole/components/Search/useConsoleSearch.test.tsx
hbenl: The vscode jest extension offered to update this "inline snapshot".
The test and the hook that it tests are a complete mystery to me, far too much React magic. I could only guess from the name that it has to do with searching in the replay console so I did some manual tests and didn't notice any differences in behavior.
@bvaughn could you have a look if this snapshot update makes sense?

bvaughn: I'd be happy to look, but this doesn't feel like a very good hand off. Maybe next time we can verify that an unexpected change to snapshot values wasn't a regression before we land the change.

bvaughn: Just verified that the committed search result ids still look correct, given the message contents. (The same messages are still being matched, just with slightly different ids).

I think something about this change impacted the ordering of, or id generation for, "newSource" messages in the Redux store. (That's all the ids are.)

PR #7538, opened 2022-08-11
Fix react_devtools e2e test

bvaughn:
The core of this fix was replacing the Redux `connect()` HOC component with newer Redux selector hooks.

I think what was happening with the older HOC essentially boils down to how React Redux and the new `Offscreen` API interact:
1. When React _hides_ an offscreen tree, it runs effect cleanup functions.
1. Redux `connect()` HOC unsubscribes from Store updates when hidden (effect cleanup functions).
1. Redux `connect()` HOC doesn't correctly handle checking for missed updates when shown (effects remounted).

I talk through why this is in this lengthy Loom: https://www.loom.com/share/3d60c14466184b03acc9af7b80e9eff8

### DIFF
--- a/public/test/scripts/react_devtools.js
+++ b/public/test/scripts/react_devtools.js
@@ -24,8 +24,6 @@ Test.describe(`Test React DevTools.`, async () => {
   await Test.warpToMessage("Added an entry");
 
   await Test.selectReactDevTools();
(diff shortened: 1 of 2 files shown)


hbenl (approved):

bvaughn:
Going to land this so CI is passing. Happy to make follow up changes if requested!

PR #7540, opened 2022-08-11
Log multiple Pause IDs per Execution Point to Sentry

bvaughn:
Conceptually, an execution point can have multiple pauses in order to allow forking for analysis with side effects. Practically, I don't think the devtools wants to support this (at least for now). To get a sense of how often execution points end up with more than one pause ID _unexpectedly_, this PR logs that event to Sentry as an exception.

This PR also updates the `TimelineContext` to remove the _requirement_ for a `pauseId` to be eagerly sent when seeking to a new point. This means that we can wait to request the pause _after_ seeking, and (usually? always?) re-use a cached pause id at that point. Locally testing seems to confirm that this is true.


packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
```diff
@@ -172,3 +176,25 @@ async function fetchPauseId(
...
+    executionPointToPauseIdSetMap.set(executionPoint, new Set([pauseId]));
+  } else {
+    set.add(pauseId);
+
+    const pauseIds = Array.from(set).join(",");
+    const error = new Error(
+      `Duplicate pause ids found for point "${executionPoint}" (${pauseIds})`
+    );
+
+    captureException(error);
```
bvaughn, line 196: I'm not very familiar with Sentry so this might be the wrong way to approach this problem. Feedback welcome!

bvaughn:
Was mostly looking for someone to double check the Sentry logging.

jasonLaster (approved):
Sentry looks good

bvaughn:
Made the mistake of rebasing and `main` branch is broken 😢

PR #7542, opened 2022-08-11
Migrate `pause` reducer to RTK and replace `promise` middleware with thunks

markerikson:
This PR:

- Migrates the `context` middleware to TS
  - Moved `validateContext` into `middleware/context`
- Migrates the `pause` reducer to `createSlice`
  - I removed some cases that were no longer relevant, like `"BREAK_ON_NEXT" / "CONNECT"`
- Replaced all usages of the Redux `promise` middleware with uses of `createAsyncThunk` instead
  - Moved most pause thunks into `reducers/pause`
  - Inlined `getFrames/getAsyncFrames` from `client/commands` into their respective thunks
  - Moved the `if (type === "rewind")` check from the `command` thunk into the reducer logic for `executeCommand.pending`
- Deleted the "debugger step" functions in `client/commands`, as they just wrapped `ThreadFront`, and the `executeCommand` thunk already has access to `ThreadFront` and can call it directly
- Deleted logic for tracking "previously expanded scopes", as that is part of the old Object Inspector and about to be replaced

This PR got way out of hand. The original intent was just to replace the `promise` middleware with `createAsyncThunk`, but all the remaining usages were directly tied to the `pause` reducer. That led me to rewrite the reducer in RTK, and I ended up moving most of the related thunks into the reducer file.

At the same time, Jason and I identified several functions in `client/commands` that were now irrelevant, such as the "debugger step" commands that just wrapped `ThreadFront[commandName]`, and also that I could inline the logic from `getFrames/getAsyncFrames/getScopes` into their respective thunks.

It _looks_ like this is all working, but we'll see from the tests :)


markerikson:
3 test failures, all very plausible (two `stepping` tests, and `console_async_eval`).

I'll investigate this tomorrow!

jasonLaster (approved):

PR #7550, opened 2022-08-13
Standardize Playwright test suite

jasonLaster:
- [x] Figure out why CI is failing (during Replay Chromium download step?)
- [x] Get Node and Browser recording generation process working
- [x] Hook up new e2e tests to CI (GitHub action?)
- [x] Document processes of...
  - [x] Running tests (locally)
  - [x] Debugging tests (locally)
  - [x] Writing new tests
  - [x] Regenerating test recordings (locally)

---

This takes what we've learned about playwright test and puts it to work creating what should the kind of test suite they'd recommend we write.

1. The tests use Playwright locators and commands
2. The tests are run with Playwright Test, not jest or our custom script
3. The tests can be run without Replay
4. The tests navigate to an already recorded example. We can choose to record the examples in the frontend CI or not. At this point, I'd prefer only re-recording them as needed. QA Wolf has used the same recordings for 10 months and just recently re-recorded them
5. Starts creating shared helpers that other tests can re-use

https://app.replay.io/recording/stepping-01--42f247e2-6630-4ea3-8a64-4cd5072644d4


jasonLaster:
Here's a fun insight. We can probably port harness mostly 1.1 to helpers which will let us switch over fairly quickly.

This means we wont really be testing a DOM based approach to testing, but we will get a bunch of other things which I like

1. we switch over which means we can drop the old Jest + customer runners
2. we wont regress on stability. 
3. we can experiment re-writing specific helpers with a DOM based approach

packages/e2e-tests/helpers.ts
```diff
@@ -0,0 +1,310 @@
...
+  await msg.hover();
+  const warpButton = msg.locator(".rewind .button") || msg.locator(".fast-forward");
+
+  await warpButton.hover();
+  await warpButton.click();
+
+  await waitForPaused(page, line);
+}
+
+export async function executeInConsole() {}
```
bvaughn, line 58: lol

packages/e2e-tests/tests/object_preview-01.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  msg = await waitForConsoleMessage(page, "Promise {  }");
+
+  // Promise contents aren't currently available in chromium.
+
+  await checkMessageObjectContents(page, msg, ['"pending"'], []);
+
+  await executeInConsole(page, "Promise.resolve({ a: 1 })");
+  msg = await waitForConsoleMessage(page, "Promise {  }");
+
+  await checkMessageObjectContents(page, msg, ['"fulfilled"', "a: 1"], ["<value>"]);
```
bvaughn, line 130: What's the difference between the 3rd and 4th parameter here?

bvaughn:
Just rebased this and pushed an update. @markerikson, @hbenl, @souporserious, and I are going to chip in on finishing this effort up. We'll just create PRs that point to this branch.

packages/e2e-tests/package.json
```diff
@@ -8,16 +8,17 @@
...
     "test:install": "playwright install",
     "ts-node": "ts-node --project ../../tsconfig.json"
   },
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "@playwright/test": "^1.25.1",
-    "@replayio/playwright": "^0.2.24",
-    "playwright": "^1.25.1",
+    "@playwright/test": "1.19",
```
bvaughn, line 19: @jaril This doesn't seem optimal. Why'd we do this?

packages/e2e-tests/playwright.config.ts
```diff
@@ -1,11 +1,12 @@
 import { PlaywrightTestConfig, devices } from "@playwright/test";
+import { devices as replayDevices } from "@replayio/playwright";
 
 const { DEBUG, SLOW_MO } = process.env;
 
 const config: PlaywrightTestConfig = {
   use: {
     headless: !DEBUG,
-    browserName: "chromium",
+    // browserName: "chromium",
```
bvaughn, line 9: nit: We can delete this?

src/ui/components/shared/PortalDropdown.tsx
```diff
@@ -49,6 +49,7 @@ export default function PortalDropdown(props: PortalDropdownProps) {
         className={`expand-dropdown flex w-full ${buttonStyle}`}
         disabled={props.disabled}
         id="portal-dropdown-button"
+        data-test-id="consoleDockButton"
```
bvaughn, line 52: Why is this test id on a component called `PortalDropdown` that's used in a few places (like Workspace members, etc)

Mistake?

PR #7552, opened 2022-08-15
Replace `connect` with hooks in 3 components

Andarist:
I was playing with https://github.com/reduxjs/react-redux/issues/1940 and I've started looking into the codebase here. Since I've never actually used "modern Redux" in an app I've figured out that it would be a nice exercise to migrate a few components to hooks.

Feel free to do whatever u want with this PR 😉 cc @markerikson

### DIFF
--- a/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
+++ b/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
@@ -1,9 +1,8 @@
-import React, { ChangeEvent, useEffect, useState } from "react";
-import { connect, ConnectedProps } from "react-redux";
+import React, { ChangeEvent, useState } from "react";
 import * as actions from "ui/actions/app";
 import hooks from "ui/hooks";
+import { useAppSelector, useAppDispatch } from "ui/setup/hooks";
 import * as selectors from "ui/reducers/app";
-import { UIState } from "ui/state";
 import { WorkspaceUser } from "ui/types";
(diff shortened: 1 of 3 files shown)


src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
```diff
@@ -221,24 +220,24 @@ const settings: Settings<
...
   const { userId: localUserId } = hooks.useGetUserId();
 
-  useEffect(() => {
-    if (view) {
-      const views: Record<string, string> = {
-        billing: "Billing",
-        members: "Team Members",
-        api: "API Keys",
-      };
-      setTab(views[view]);
```
Andarist, line 238: This was trying to derive the local state from the selected state - so I've just refactored this to use the actual selected state right from the start

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
```diff
@@ -130,7 +129,6 @@ const settings: Settings<
     settings?: any;
     isAdmin: boolean;
     workspaceId: string;
-    hideModal: PropsFromRedux["hideModal"];
```
Andarist, line 133: If I read this correctly... only `TeamMembers` were actually using `hideModal` so I've just moved it directly to it and cleaned this up from the parent components

markerikson:
Thanks! I'll merge this probably early next week

PR #7556, opened 2022-08-15
Migrate Scopes and Frames components and logic to TS

markerikson:
This PR:

- Migrates all of the "Scopes" and "Frames components and logic to TS
- Removes a couple uses of `_.find` and replaces them with `Array.find`


markerikson:
Not sure what's up with that failing `highlighter` test. It says it's stuck waiting for the highlighter overlay to be in the DOM, but I can see that on the preview branch highlighting definitely works when hovering over nodes in the "Elements" panel and the "Breadcrumbs" section.

PR #7559, opened 2022-08-15
Nuke even more dead code

markerikson:
This PR:

- Nukes a ton more dead code in various util files


src/devtools/shared/contextmenu.js
```diff
@@ -10,55 +10,6 @@ function inToolbox() {
   }
 }
```
jasonLaster, line 11: surprised we can't kill this entire file

jasonLaster (approved):

src/devtools/shared/contextmenu.js
markerikson: It's still being used in a couple places when I checked

PR #7566, opened 2022-08-17
Improve object pre-caching for legacy and new architecture interop

bvaughn:
This commit addresses some cases where we were _missing_ some pre-cached data thanks to timing (the connector being further down inside of the tree and subscribing inside of a passive effect).

It also fixes a possible issue where our _data cloning_ (something we have to do because of `ValueFront` mutations) wasn't properly handling circular references for values like `Pause`.


packages/protocol/thread/pause.ts
```diff
@@ -211,6 +207,15 @@ export class Pause {
...
     datas.forEach(d => this._updateDataFronts(d));
+
+    const pauseId = this.pauseId!;
+    const point = this.point!;
+
+    if (pauseDataHandlers.length > 0) {
+      datas.forEach(data => {
+        pauseDataHandlers.forEach(handler => handler(pauseId, point, data));
+      });
+    }
```
bvaughn, line 218: This (`addData`) is a better place to call pre-cache because more places may call this than just the `create` and `instantiate` methods.

pages/recording/[[...id]].tsx
```diff
@@ -185,6 +184,7 @@ function RecordingPage({
     return (
       <>
         {head}
+        <ObjectPreviewSuspenseCacheAdapter />
```
Comment (bvaughn) [inline pages/recording/[[...id]].tsx:187]: Feels goofy to have to hoist this component all the way to the top of the page. We could probably get away with moving it _a little lower_ but I don't want to miss values. Alternately we could move it outside of React entirely and into our module initialization path.

src/devtools/client/debugger/src/components/SecondaryPanes/ObjectPreviewSuspenseCacheAdapter.tsx
```diff
@@ -21,7 +36,9 @@ export default function ObjectPreviewSuspenseCacheAdapter() {
       if (objects) {
         // Be sure to clone object data before pre-caching it.
         // Otherwise ValueFronts might deeply mutate it and change its structure.
-        preCacheObjects(pauseId, JSON.parse(JSON.stringify(objects)));
+        // Because object data may contain circular references, we need to also filter out empty entries.
+        const cloned = JSON.parse(JSON.stringify(objects, createReplacer())).filter(Boolean);
+        preCacheObjects(pauseId, cloned);
```
bvaughn, line 41: This is gross. I wish we didn't have to be so heavy handed, but `ValueFront` sucks.

jasonLaster: This is one of the grossest things i've seen in awhile. Can't remember the last time I used the 2nd param to `stringify`

bvaughn: I can 🤓 

https://github.com/replayio/devtools/blob/127c09c5dfb1c2989d1d2522bf2bc7b83cd4bcc3/packages/shared/client/encoder.ts#L13-L27

jasonLaster: oh god - you are the replacer!

packages/protocol/thread/pause.ts
```diff
@@ -209,6 +205,17 @@ export class Pause {
...
+    const pauseId = this.pauseId!;
+    const point = this.point!;
+
+    // Event handlers must be called before processing the data.
+    // Calling _updateDataFronts() will add ValueFronts and mutate the underlying data.
+    if (pauseDataHandlers.length > 0) {
+      datas.forEach(data => {
+        pauseDataHandlers.forEach(handler => handler(pauseId, point, data));
+      });
+    }
```
bvaughn, line 217: This (`addData`) is a better place to call pre-cache because more places may call this than just the `create` and `instantiate` methods.

src/devtools/client/debugger/src/components/SecondaryPanes/ObjectPreviewSuspenseCacheAdapter.tsx
```diff
@@ -2,16 +2,31 @@ import { preCacheObjects } from "@bvaughn/src/suspense/ObjectPreviews";
...
+  return function replacer(_: string, value: any): any {
+    if (typeof value === "object" && value !== null) {
+      if (cache.has(value)) {
+        return undefined;
+      }
+      cache.add(value);
+    }
+    return value;
+  };
+}
```
bvaughn, line 19: We need to **deeply** clone data because of `ValueFront` mutations (😔) so `JSON.stringify` can be helpful. Our data _might_ have circular references though, so we need this replacer.

**TODO** Verify if this is still the case or if it was only because of the `ValueFront` mutations.

packages/bvaughn-architecture-demo/components/inspector/useClientValue.ts
```diff
@@ -1,11 +1,24 @@
...
+  const client = useContext(ReplayClientContext);
+
+  const objectId = protocolValue.object;
+  if (objectId != null && getObject(pauseId, objectId) === null) {
+    // If we are converting an object value, protocolValueToClientValue() will require preview data.
+    // Typically we will have already cached this data from a previous request.
+    // There may be edge cases where this is not true though (see BAC-2095).
+    // As a workaround, just-in-time fetch the data before calling protocolValueToClientValue().
+    getObjectWithPreview(client, pauseId, objectId);
+  }
```
bvaughn, line 20: This is unfortunate but seems to be necessary (at least for now, see BAC-2095)

src/devtools/client/debugger/src/components/SecondaryPanes/ObjectPreviewSuspenseCacheAdapter.tsx
bvaughn: I think this is safe to remove now actually.

bvaughn:
Mark and I did a verbal review of this while on Hangouts so I'm going to land it once CI is happy.

PR #7568, opened 2022-08-17
Filter timestamped points by point, not time

jcmorrow:
We were filtering analysis points by displayed begin and end times. That
does not work. The displayed begin and end times for a focus region
don't necessarily reflect the true begin and end times for that region.
Since we pass a point range for hitCounts, sometimes analysis and hitCounts
disagree. Now, we might have another class of bug: it's theoretically
possible (though hopefully pretty unlikely!) to have "hits" that are
outside of the displayed timeline of a breakpoint. If we see that
happening, we can rethink some of this, but I think it will not happen
*nearly* as often as us miscounting hits because of edge pieces.

### DIFF
--- a/src/ui/utils/timeline.test.ts
+++ b/src/ui/utils/timeline.test.ts
@@ -10,7 +10,7 @@ import {
   overlap,
 } from "./timeline";
 
-const point = (time: number) => ({ time, point: "" });
+const point = (time: number) => ({ time, point: `${time}` });
 const focusRegion = (from: number, to: number): FocusRegion => ({
   begin: point(from),
   beginTime: from,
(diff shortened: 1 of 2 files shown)


jcmorrow:
This will fix https://linear.app/replay/issue/BAC-2082/source-hit-points-and-log-point-panel-counts-disagree

markerikson (approved):

bvaughn:
> Can we use execution points instead of time?

I was also wondering this.

My understanding, after being burned by time several times, is that it's really only useful to _show_ to a user but never to sort or filter with.

jcmorrow:
Sure, we can also filter with points, I think in this case, since I'm sure that I'm comparing timestamps **that are attached to timestamp points** I'm not too worried about it, but it's a reasonable rule to stick to.

jcmorrow:
Made that change!

bvaughn (approved):

PR #7571, opened 2022-08-17
Bump packages and remove unused packages

markerikson:
This PR:

- Bumps React to the latest experimental build ( `0.0.0-experimental-1e5245df8-20220817` ) to help possibly fix issues with `connect` + `<Offscreen>`
- Bumps RTK to 1.8.4
- Changes all remaining references to `import from "redux"` to be `from "@reduxjs/toolkit"` instead
- Removes the `ResizeObserverPolyfill` package
- Removes a bunch of unused packages
- Rewrites `<AddCommentButton>` to use CSS modules, as it was the _only_ component in the codebase actually using `styled-components`
  - note: the Redux DevTools Extension code apparently _also_ relies on `styled-components`. The RDT integration is disabled atm at the display level, but turns out that _was_ technically still being `import()`ed.  I've commented that out as well so that Webpack doesn't process it.
- Removes `styled-components` as a dependency
- Updates TS to 4.7.4


bvaughn (approved):

souporserious:
Woot, thank you for cleaning up the `AddCommentButton` component! That was the first task I did before we decided on what styling solution we were going with 😅

PR #7578, opened 2022-08-18
Fix ThreadFront for the architecture demo

hbenl:
The architecture demo doesn't set `ThreadFront.sourcesSelectors` but it uses the `AnalysisManager` which asks for corresponding sourceIds. So we need to ensure that `ThreadFront.getCorrespondingSourceIds()` can work without `ThreadFront.sourcesSelectors`.


hbenl:
Fixes FE-586

bvaughn (approved):

PR #7586, opened 2022-08-18
Enable the new Console and Object Inspector by default for all users

bvaughn:
I think the easiest way to default this to being on for all users is to just rename the preference and change its default to on rather than off.

Note that this PR also updates a lot of our e2e test helper functions and selectors to use data attributes rather than CSS class names. **This is not a backwards compatible change.** If we revert the new console, we will need to revert (or disable) these tests also.

### DIFF
--- a/src/ui/utils/prefs.ts
+++ b/src/ui/utils/prefs.ts
@@ -23,7 +23,7 @@ pref("devtools.hitCounts", "hide-counts");
 pref("devtools.features.columnBreakpoints", false);
 pref("devtools.features.commentAttachments", false);
 pref("devtools.features.enableLargeText", false);
-pref("devtools.features.enableNewComponentArchitecture", false);
+pref("devtools.features.disableNewComponentArchitecture", false);
 pref("devtools.features.logProtocol", false);
 pref("devtools.features.logProtocolEvents", false);
 pref("devtools.features.originalClassNames", false);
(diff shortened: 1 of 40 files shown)


src/ui/utils/prefs.ts
```diff
@@ -23,7 +23,7 @@ pref("devtools.hitCounts", "hide-counts");
 pref("devtools.features.columnBreakpoints", false);
 pref("devtools.features.commentAttachments", false);
 pref("devtools.features.enableLargeText", false);
-pref("devtools.features.enableNewComponentArchitecture", false);
+pref("devtools.features.disableNewComponentArchitecture", false);
```
bvaughn, line 26: Initializing _disable_ to `false` means all users will get the new console by default.

bvaughn:
Haha! Looks like this breaks a bunch of our e2e tests. Oh, yay.

jasonLaster (approved):
Looks good to me. 

For what it is worth, we usually keep the flag the way it is and have it be enabled by default, but this is fine too.

bvaughn:
> For what it is worth, we usually keep the flag the way it is and have it be enabled by default, but this is fine too.

Interesting! I'd considered doing that, but it wouldn't have rolled the feature out to any users who have modified their settings (presumably)

bvaughn:
I dunno like...maybe they all pass now.

bvaughn:
Playwright is being sketchy but they all pass locally so I'm going to just assume it's flakiness.

bvaughn:
This PR is ready to be merged, but I'm nervous about FE-611. I don't know a solution given our current constraints (the protocol library design). I don't feel good about merging this until we either accept the downside explicitly or have a plan for addressing it that's better than "delete the old Console eventually'

bvaughn:
Not sure what's up with the failing QA Wolf test, but it's logging in and out of the Library page, so it has nothing to do with this PR.

PR #7587, opened 2022-08-18
Prevent Earthly from skipping Playwright tests due to caching

bvaughn:
Am I crazy? Is there a better way to do this? 😅

### DIFF
--- a/packages/bvaughn-architecture-demo/playwright/.gitignore
+++ b/packages/bvaughn-architecture-demo/playwright/.gitignore
@@ -1,5 +1,6 @@
 .editorconfig
 .yarn
+cache_buster
 playwright-report*
 test-results
 tests/*.tmp
--- a/packages/bvaughn-architecture-demo/playwright/Earthfile
+++ b/packages/bvaughn-architecture-demo/playwright/Earthfile
@@ -7,6 +7,7 @@ install:
     WORKDIR /playwright
     RUN yarn install
     RUN yarn playwright:install
+    RUN node bust_earthly_cache.js
     EXPOSE 3000
 
 playwright:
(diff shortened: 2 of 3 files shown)


bvaughn:
No worries 😆 Wouldn't `--push` invalidate _all_ caching? It's nice to get _some_ caching. I just don't want the contents of the tests themselves to be cached.

I don't have a very solid mental model for how this works.

jazzdan:
So `--push` should only prevent caching on `RUN`s that have `--push` applied. All other steps would still be cached

bvaughn:
Isn't that basically the same thing? If we always have to `--push` to ensure we don't get false positives (b'c of caching causing steps to be skipped), and nothing is cached when `--push` is used, then wouldn't we always bypass caching?

The thing I was trying to do with this PR was to only cache-bust the tests themselves.

bvaughn:
Actually it looks like `--push` doesn't work anyway
Screen Shot 2022-08-24 at 2 53 32 PM

jazzdan:
Yeah you're right.

I think the fundamental problem here is that earthly doesn't know anything about the code being tested, what we actually want is to bust the test result cache when any of the code that goes in to running `yarn dev` changes but because all of that is happening outside of earthly there's no way for it to do that.

jazzdan (approved):

jazzdan:
I _think_ there's a way to achieve earthly knowing about all of the code that goes in to `yarn dev` but each time I've tried to do it I've been thwarted by yarn workspaces weirdness. Maybe we could pair on it at some point? :)

bvaughn:
Gotcha. That's fair. Thanks a bunch for the discussion!

PR #7591, opened 2022-08-22
Remove ThreadFront source methods

hbenl:


packages/protocol/thread/thread.ts
```diff
@@ -553,25 +475,19 @@ class _ThreadFront {
...
-            const location = { sourceId, line, column };
-            const { breakpointId } = await client.Debugger.setBreakpoint(
-              { location, condition },
-              this.sessionId!
-            );
-            if (breakpointId) {
-              this.breakpoints.set(breakpointId, { location });
-            }
-          })
+        await ThreadFront.getBreakpointPositionsCompressed(sourceId);
```
markerikson, line 482: :hand: shouldn't it be `this.` instead of `ThreadFront.`?

markerikson (approved):

packages/protocol/thread/thread.ts
hbenl: Agreed, although I didn't actually change that (and there are several other instances of this in `ThreadFront`)

markerikson:
BOOM!

PR #7599, opened 2022-08-23
Expose `ReplayClient` to thunks and move methods from `ThreadFront`

markerikson:
This PR:

- Alters the startup sequence to pull in `ReplayClient` and expose it to thunks as part of `thunkArgs.extra`
- Moves multiple methods from `ThreadFront` to `ReplayClient`

The methods I moved just wrapped the low-level `client` and called `client.Domain.doSomething(this.sessionId)`, so they didn't _need_ to be in `ThreadFront` specifically.

A couple of the methods had equivalents in `ReplayClient` already, like `loadRegions`.


packages/bvaughn-architecture-demo/src/utils/testing.tsx
```diff
@@ -194,6 +195,12 @@ export function createMockReplayClient() {
...
     removeEventListener: jest.fn(),
     runAnalysis: jest.fn().mockImplementation(async () => []),
-    searchSources: jest.fn().mockImplementation(async () => []),
+    searchSources: jest.fn().mockImplementation(async () => {}),
+    searchFunctions: jest.fn().mockImplementation(async () => {}),
+    getAnnotationKinds: jest.fn().mockImplementation(async () => []),
+    getResponseBody: jest.fn(),
+    getRequestBody: jest.fn(),
+    findKeyboardEvents: jest.fn().mockImplementation(async () => []),
+    findNavigationEvents: jest.fn().mockImplementation(async () => []),
```
bvaughn, line 204: Small nit but these were alpha sorted before and I find them easier to read/scan quickly that way.

src/ui/actions/network.ts
```diff
@@ -123,11 +123,20 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+  */
+
     if (requestSummary.hasResponseBody) {
-      ThreadFront.fetchResponseBody(requestId);
+      replayClient.getResponseBody(requestId);
     }
     if (requestSummary.hasRequestBody) {
-      ThreadFront.fetchRequestBody(requestId);
+      replayClient.getRequestBody(requestId);
     }
```
bvaughn, line 140: This is our change to re-think these awkward/confusing patterns. (Or defer re-thinking them to later, but let's not copy them over as-is.)

bvaughn (changes requested):

src/ui/setup/dynamic/devtools.ts
```diff
@@ -152,6 +153,15 @@ export default async function DevTools(store: AppStore) {
...
   window.app.prefs = window.app.prefs ?? {};
 
+  if (!replayClient) {
+    // This _should_ have been passed in from `RecordingPage`.
+    // But, some test code also calls `setupDevtools()`, and those
+    // test files don't call `useReplayClient` anywhere.
+    // Grab the usual instance manually so we can pass it to thunks.
+    const replayClientModule = await import("shared/client/ReplayClientContext");
+    replayClient = replayClientModule.replayClient;
+  }
```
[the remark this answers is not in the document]
markerikson: `setupDevtools()` is called in some of the unit-ish tests, and I was trying to make sure that at least _some_ instance of `replayClient` was available in case that logic ended up trying to access it.  It may not be 100% necessary - I can try running the tests without it and see what happens.

src/ui/actions/network.ts
markerikson: I'm not sure how much of this can be "re-thought" atm.

The backend behavior is what really defines the calls here, as the bodies come back via events instead of direct responses.

I agree I don't _like_ this pattern at all. But we can't change how the protocol API returns these values, so the logic for triggering the requests and handling the events is going to be essentially the same regardless of where the event handling lives in the code.

bvaughn: > I agree I don't like this pattern at all. But we can't change how the protocol API returns these values

I don't think we should try to change the protocol at this point. I agree that's a big lift.

> so the logic for triggering the requests and handling the events is going to be essentially the same regardless of where the event handling lives in the code.

If we move this functionality into `ReplayClient` – the methods should do something meaningful (like collect and return the values). If there happens to be other listeners attached (in the legacy code) that also observes this information, that's okay.

I feel really strongly that we shouldn't copy over non-obvious patterns like this to the new code though.

bvaughn: Honestly I don't think it's worth us removing this stuff from `ThreadFront` at all if we're just copying it as-is into `ReplayClient`. I'd rather just leave it in TF>

bvaughn: But literally moving code verbatim from one place, to another, doesn't buy us anything.

src/ui/hooks/useReplayClient.ts
```diff
@@ -63,4 +63,6 @@ export default function useReplayClient() {
       }
     };
   }, []);
+
+  return replayClient;
```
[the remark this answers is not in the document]
markerikson: Eh, it's still only being called in one place - I just moved it up to the actual Next page level so it can be passed into the setup logic.  I don't think we have a reason to call it anywhere else.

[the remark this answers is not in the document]
bvaughn: I know. I agree that we don't. My concern here is that adding a return value _implies_ that it's meant to be the way (or at least _a valid way_) to get the client– and it's not. It's specifically only meant to _configure_ the client. Does that clarify my concern..?

src/ui/actions/network.ts
```diff
@@ -123,11 +123,22 @@ export function selectAndFetchRequest(requestId: RequestId): UIThunkAction {
...
+  */
+
     if (requestSummary.hasResponseBody) {
-      ThreadFront.fetchResponseBody(requestId);
+      protocolClient.Network.getResponseBody({ id: requestId, range: { end: 5e9 } }, sessionId);
     }
     if (requestSummary.hasRequestBody) {
-      ThreadFront.fetchRequestBody(requestId);
+      protocolClient.Network.getRequestBody({ id: requestId, range: { end: 5e9 } }, sessionId);
     }
```
bvaughn, line 142: Thank you 🙇🏼

src/ui/setup/dynamic/devtools.ts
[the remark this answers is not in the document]
bvaughn: I'd be curious to know what you found here.

bvaughn (approved):
I'd prefer the change to `useReplayClient` be reverted but otherwise this looks good 👍🏼

packages/shared/client/ReplayClient.ts
```diff
@@ -138,11 +149,19 @@ export class ReplayClient implements ReplayClientInterface {
...
 
     this._sessionId = sessionId;
+    this.sessionWaiter.resolve(sessionId);
     this._threadFront.setSessionId(sessionId);
 
     return sessionId;
   }
 
+  async findKeyboardEvents(onKeyboardEvents: (events: keyboardEvents) => void) {
+    const sessionId = this.getSessionIdThrows();
```
hbenl, line 159: Since we're in an async function I think we should await the `sessionWaiter`, that would make this method (slightly) more robust. Same for all other uses of `getSessionIdThrows`.

hbenl (approved):

packages/shared/client/ReplayClient.ts
markerikson: The only place this is called is inside of `setupApp()` in `ui/actions/app`, and that already does an explicit `ThreadFront.waitForSession().then()`.

src/ui/setup/dynamic/devtools.ts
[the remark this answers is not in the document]
markerikson: Just tried removing that, and sure enough, several tests fail:

```
 FAIL  src/devtools/client/webconsole/components/__tests__/WebConsoleApp.test.tsx (12.992 s)
  ● Web Console UI › Can render a list of messages and filter it

    TypeError: Cannot read properties of undefined (reading 'findKeyboardEvents')

      73 |
      74 |     replayClient
    > 75 |       .findKeyboardEvents(({ events }) => onKeyboardEvents(events, store))
         |        ^
      76 |       .then(() => {
      77 |         store.dispatch(loadReceivedKeyboardEvents());
      78 |       });

      at findKeyboardEvents (src/ui/actions/app.ts:75:8)
```

packages/shared/client/ReplayClient.ts
hbenl: I'd still prefer the more robust version, so that we don't have to worry about it when calls to this method or any of the other methods using `getSessionIdThrows()` are moved around.

bvaughn: Maybe we should just remove the session waiter.

PR #7600, opened 2022-08-23
fix frames not spanning full grid

souporserious:
Fixes FE-595 where the frame wasn't spanning the full width of the grid. I ran the snapshot tests and nothing changed 🎉


bvaughn (approved):
You rock 🥰

PR #7602, opened 2022-08-24
Always wait for the pauseId before sending a message (FE-598)

hbenl:


packages/protocol/thread/pause.ts
```diff
@@ -395,12 +395,13 @@ export class Pause {
...
 
-  sendMessage<P, R>(
+  async sendMessage<P, R>(
     method: (parameters: P, sessionId?: SessionId, pauseId?: PauseId) => R,
     params: P
   ) {
+    await this.pauseIdWaiter.promise;
     assert(this.pauseId, "pauseId not set before sending a message");
-    return method(params, this.sessionId, this.pauseId);
+    return await method(params, this.sessionId, this.pauseId);
```
bvaughn, line 404: Does this `await` actually do anything? I think returning another async method (from within an async method) is the same but maybe I'm wrong and I'm curious 😁

bvaughn (approved):

packages/protocol/thread/pause.ts
hbenl: Yes, it's the same, I just prefer to do this explicitly.

PR #7605, opened 2022-08-24
Don't try to release an analysis if we failed to create one (FE-606)

hbenl:


src/ui/actions/logpoint.ts
```diff
@@ -383,7 +383,9 @@ async function setMultiSourceLogpoint(
     // Rather than running *this* analysis, create a *new* analysis which only has
     // the found points which fall *inside* of our current focusRegion, and run *that*.
   } finally {
-    await analysis!.releaseAnalysis();
+    if (analysis!) {
+      await analysis.releaseAnalysis();
+    }
```
hbenl, line 388: why not just `await analysis?.releaseAnalysis()`? Because then Typescript complains "Variable 'analysis' is used before being assigned."

src/ui/actions/logpoint.ts
```diff
@@ -383,7 +383,9 @@ async function setMultiSourceLogpoint(
     // Rather than running *this* analysis, create a *new* analysis which only has
     // the found points which fall *inside* of our current focusRegion, and run *that*.
   } finally {
-    await analysis!.releaseAnalysis();
+    if (analysis!) {
```
bvaughn, line 386:
```suggestion
    if (analysis) {
```

bvaughn (approved):

src/ui/actions/logpoint.ts
hbenl: That would lead to the same complaint ("Variable 'analysis' is used before being assigned.") from Typescript

bvaughn: I thought the `!` operator in TypeScript was just a way to assert that the value was not null or undefined. I didn't think it would even affect the error you're seeing.

What if you explicitly initialize `analysis = undefined` above?

hbenl: > I didn't think it would even affect the error you're seeing.

Yes, this surprised me as well.

> What if you explicitly initialize analysis = undefined above?

If I do that then I also need to add `| undefined` to its type and then I need to add `!` in line 311.

bvaughn: > If I do that then I also need to add | undefined to its type and then I need to add ! in line 311.

I wouldn't expect that to be true. (Did you see it?) TypeScript is pretty smart about seeing if you've initialized a variable within the current scope.

hbenl: Yes, I did see it, and was also surprised by it.

PR #3861, opened 2021-10-05
React devtools shutdown

jasonLaster:
This PR adds some error handling for the unsupported bridge protocol logic which was responsible for showing the warning modal.

1. when we do not have the supported bridge protocol version we close react devtools
2. when we shutdown we log a telemetry event
3. i added a check in the secondary toolbox so we no longer render the react devtools panel if we don't have the react components to show

PR #7606, opened 2022-08-24
Handle undefined response from RDT backend (FE-607)

hbenl:


src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -172,7 +172,9 @@ class ReplayWall implements Wall {
 
     if (response.returned) {
       const result: any = await response.returned.getJSON();
-      this._listener?.({ event: result.event, payload: result.data });
+      if (result) {
+        this._listener?.({ event: result.event, payload: result.data });
+      }
       return result;
```
bvaughn, line 178: The whole `sendRequest` method looks weird to me. It returns _either_ a `result` (which might be undefined) _or_ `response.returned` (which also might be undefined).

I'm not sure what it's supposed to be doing? I think maybe @jasonLaster would be better suited to review this as the author of the code originally (#3861)?

hbenl: That `return response.returned` doesn't make sense (it's only executed if `response.returned` is falsy) so I just removed it.

hbenl: And I don't think @jasonLaster has more insights on this, so I'm just gonna land it.

bvaughn: I dunno. Seems like a sketchy function that's exactly the sort of thing we should invest in reviewing.

Jason wrote the function originally. Why wouldn't he have insight?

jasonLaster: going to go on record and say i do not have additional insights... sorry to disappoint

PR #7611, opened 2022-08-24
Wait for access token during devtools setup

jasonLaster:
See this replay which shows how setAccessToken is called in the token manager logic, but could be too late if `getToken` takes awhile

Fix SCS-17


ryanjduffy (approved):
I don't think does any harm but I'm not convinced it'll change how this works. I tried adding a synthetic delay in the `fetchToken` logic but that didn't cause the bug to repro (without this fix) so ... /shrug?

PR #7613, opened 2022-08-24
Migrate more components and utils to TS

markerikson:
This PR:

- Migrates a bunch more component and util files to TS


bvaughn (approved):
I assume the failure in `breakpoints-03.js` is just because of flakiness?

markerikson:
Seems that way - it passed on re-run

PR #7618, opened 2022-08-25
Add recording id to protocol log bug reports

bvaughn:


bvaughn:
Super discouraging to see the Playwright tests have gotten flaky. They were so solid. I wonder what changed.

PR #7623, opened 2022-08-26
Treat application/x-www-form-urlencoded as textish

jaril:


src/ui/components/NetworkMonitor/content.ts
```diff
@@ -51,6 +51,7 @@ const TEXTISH_CONTENT_TYPES = [
   "application/xhtml+xml",
   "application/xml",
   "image/svg+xml",
+  "application/x-www-form-urlencoded",
```
jcmorrow, line 54: Can we sort this array? 🙃

jcmorrow (approved):

PR #7628, opened 2022-08-29
Better timeout message

jonbell-lot23:
Follow-up to https://github.com/replayio/devtools/pull/7537

The old alert was just long enough to have a single word floating on the second line. This one is more succinct.

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -110,7 +110,7 @@ export function getDisconnectionError(): UnexpectedError {
   endMixpanelSession("disconnected");
   return {
     action: "refresh",
-    content: "Replays disconnect after 15 minutes to reduce server load.",
+    content: "This replay timed out to reduce server load.",
     message: "Ready when you are!",
   };
 }


src/ui/actions/session.ts
```diff
@@ -110,7 +110,7 @@ export function getDisconnectionError(): UnexpectedError {
   endMixpanelSession("disconnected");
   return {
     action: "refresh",
-    content: "Replays disconnect after 15 minutes to reduce server load.",
+    content: "This replay timed out to reduce server load.",
```
jasonLaster, line 113: Can we say this in a friendly way... the previous copy wasn't that approachable either, but i think we could be better

jonbell-lot23: We both agree with making this page as friendly as possible. But the most tone/friendliness is going to come from the upcoming illustration, along with the "Ready when you are!" header and the call to action, which can more friendly with something like "Dive in!"

The other issue is the length of the copy itself, which is important to make as clear and concise as possible.

[Google Material UX guidelines: be concise](https://material.io/design/communication/writing.html#principles)
[Microsoft UX guidelines: The message [should be as] short as possible, but no shorter](https://docs.microsoft.com/en-us/windows/win32/uxguide/mess-error)
[Adobe: Clarity is key](https://xd.adobe.com/ideas/process/information-architecture/error-message-design-ux/)
[UXWriting Hub: Keep language clear and concise](https://uxwritinghub.com/error-message-examples/)
[UXWriting Hub: Microcopy is not marketing copy](https://uxwritinghub.com/what-is-microcopy/)
[PRSA: Sentences that are eight words or less are understood better than longer sentences](http://prsay.prsa.org/2009/01/14/how-to-make-your-copy-more-readable-make-sentences-shorter/)

We're going to improve this page's design and overall tone significantly when we address the visual design. In the meantime, I recommend being clear and concise so it doesn't wrap to the second line and is as glanceable as possible. Even adding a little "Please try again!" is redundant with the header and makes the visual less easy to scan.

jasonLaster: Sounds good

PR #7632, opened 2022-08-29
Fixed logic bug in display of overflowing Object/Array properties

bvaughn:


packages/bvaughn-architecture-demo/components/inspector/PropertiesRenderer.tsx
```diff
@@ -32,14 +32,16 @@ export default function PropertiesRenderer({
...
+  const { className, objectId } = object;
 
   // If we have an ObjectPreview already, use it.
   // If we just have an Object, then Suspend while we fetch preview data.
-  if (preview == null || preview.overflow) {
+  if (object.preview == null || object.preview.overflow) {
     object = getObjectWithPreview(client, pauseId, objectId, true);
   }
 
+  const { preview } = object;
```
jcmorrow, line 43: It's very annoying that I can't comment on lines that didn't change - but just to make sure I am following here, we had a situation where line 51 was basically referencing a stale (or potentially missing) `preview`, is that right?

jasonLaster (approved):

packages/bvaughn-architecture-demo/components/inspector/PropertiesRenderer.tsx
```diff
@@ -113,7 +118,7 @@ export default function PropertiesRenderer({
       ))}
 
       <Suspense fallback={<Loader />}>
-        {properties.map((property, index) => (
+        {properties.slice(propertiesIndex).map((property, index) => (
```
jcmorrow, line 121: I don't think I understand what changed here exactly, I actually can't even tell if it's related to the change up above?

jcmorrow (approved):
Relative to the context I have with this code (which is basically 0) it looks good (and complicated 😆 ) to me.

packages/bvaughn-architecture-demo/components/inspector/PropertiesRenderer.tsx
bvaughn: Yup ☹️

packages/bvaughn-architecture-demo/components/inspector/PropertiesRenderer.tsx
bvaughn: It's unrelated, but I had a side effect in this code before (was calling `array.splice` instead of `array.slice`). Overall this was a some pretty clowny code on my park.

PR #7633, opened 2022-08-29
Rewrite EventListeners logic to remove `Node/ValueFront` and fix pause behavior

markerikson:
This PR:

- Fixes a timing edge case where the `paused()` thunk would dispatch the `"pause/paused"` action before we actually received a pause ID back from the server. The thunk now waits until that pause ID is received, so that the Redux state can save a copy of the pause ID.
- Exposes several of the new Suspense Object Cache fetch methods as `thunkExtraArgs.objectCache`
- Rewrites all of the EventListeners logic to remove `Node/ValueFront` usage:
  - The `<EventListenersApp>` component now looks at the `selectedNodeId` in the Redux state to decide when to update, instead of listening to `Selection.on("new-node-front")`
  - The component now suspends to make sure we have the DOM node preview data before continuing
  - The `<XHTMLNode>` accepts that preview object for display instead of a `NodeFront`
  - The logic for fetching both plain and "framework" event listeners is now in a thunk
  - The event listener fetching logic relies on the low-level `protocolClient` API and the object cache's "fetch this item by ID" behavior, instead of `Node/ValueFront` methods
  - Fixed up the React "framework listener" detection so that it works with React 17/18, as the internal special hidden field name changed with 17.0
  - Adds either a "framework" icon (currently just the React logo) or a "JS" icon to the treeview event names, and slightly ups the font size of the event names (both for visibility and to align the icons better)

Example of a button with an `onClick` handler:

image

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/paused.ts
+++ b/src/devtools/client/debugger/src/actions/pause/paused.ts
@@ -49,6 +49,8 @@ export function paused({
     // @ts-expect-error optional time mismatch
     const pause = ThreadFront.ensurePause(executionPoint, time);
 
+    await pause.createWaiter;
+
     dispatch(pausedAction({ executionPoint, time, id: pause.pauseId!, frame }));
 
     const cx = getThreadContext(getState());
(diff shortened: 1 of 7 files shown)


markerikson:
Hmm. 2 failing tests, in both pushes. But, the `object-preview` failure is from backend errors, and I don't see an obvious error in the `doc_exceptions_bundle` recording?

PR #7634, opened 2022-08-29
Hit points and log points auto-fetch corresponding locations

bvaughn:
Not sure how this stuff works entirely, but basically I looked for places in the new Console that we're running analysis – and I added code to auto-map corresponding locations.

Re-testing on the recording that was broken as per FE-587 shows log points now, so maybe it's working?

Screen Shot 2022-08-29 at 4 18 55 PM


packages/shared/client/ReplayClient.ts
```diff
@@ -582,6 +601,16 @@ export class ReplayClient implements ReplayClientInterface {
     });
   }
 
+  _getCorrespondingLocations(location: Location): Location[] {
+    const { column, line, sourceId } = location;
+    const sourceIds = this._threadFront.getCorrespondingSourceIds(sourceId);
```
markerikson, line 606: :hand: After Holger's sources refactoring, the current implementation of `ThreadFront.getCorrespondingSourceIds` is just:

```ts
  getCorrespondingSourceIds = (sourceId: SourceId) => [sourceId];
```

To recap: "corresponding" sources are sources that have 100% identical contents. The `corresponding` array in a `SourceDetails` object will always have at least one entry - the ID of the `SourceDetails` itself.  It _may_ have others too (ie from fast refresh, etc).

Given that, I'm actually not sure that `ReplayClient._getCorrespondingLocations` is doing anything useful, and I'm confused why you're seeing an improvement in behavior.

I guess the other change is from passing _one_ location to possibly many?  Perhaps it's a switch from accidentally pointing to a different source ID, to at least including the right one somewhere in the array?

bvaughn: > :hand: After Holger's sources refactoring, the current implementation of `ThreadFront.getCorrespondingSourceIds` is just:
>
> ```ts
>   getCorrespondingSourceIds = (sourceId: SourceId) => [sourceId];
> ```

What! but...why is it working?

That seemed to be what the old console was doing! Otherwise my PR does...nothing, right?

bvaughn: What would the purpose of these lines be then?
https://github.com/replayio/devtools/blob/badcae03c231cc26983011ca7767e61db1d74844/src/ui/actions/logpoint.ts#L245-L247

bvaughn: And what's the purpose of this?
https://github.com/replayio/devtools/blob/badcae03c231cc26983011ca7767e61db1d74844/packages/protocol/thread/thread.ts#L784-L787

markerikson: The logic for checking corresponding source IDs _did_ use to live in `ThreadFront`, but after Holger's PR, it doesn't, really.

We may want to double-check with @hbenl what the right behavior should be for a couple of these spots.

bvaughn: Yeah I think my main point is that...it seems like the cleanup is incomplete if this method isn't doing anything, because even other places within `thread.ts` are still calling it like it has a purpose.

bvaughn:
Tell you what– I'm going to land this, because for whatever reason– it improves the situation. Once @hbenl is back, I want to pick his brain about our thread above and what a better approach would be.

packages/shared/client/ReplayClient.ts
hbenl: Sorry, I should have added a comment for `ThreadFront.getCorrespondingSourceIds()`: this method is overwritten [here](https://github.com/replayio/devtools/blob/205e0e2810079e355f5897953ff0e6cf48c7ee43/src/ui/setup/index.ts#L133-L135), connecting it to the sources in the redux store. So the initial method definition is just the fallback when there is no information on corresponding sources.

hbenl: > I wonder if we could use a slightly more discoverable pattern here

Yes, we should. We also wanted to move the corresponding sources stuff out of the ThreadFront into a separate module.

> I guess on the one hand, that explains why this change seemed to improve things.

Agreed, this PR is doing the right thing.

PR #7639, opened 2022-08-30
New OI should allow jump-to-function-def

bvaughn:
Just needed to inject a callback via the `InspectorContext`. I didn't realize this was desirable for the OI _and_ the Console.


markerikson (approved):

bvaughn:
Looks like the preview branch is crashing when I open previews. That's unexpected. No crashes locally (nor on production, thankfully). Going to try try a Vercel re-build.

bvaughn:
Looks like the previous import was inadvertently causing some of code-mirror to be pulled in, and it _wasn't_ server compatible. So I moved things around and hopefully that will address it. Let's see if the new deploy is better.

PR #7641, opened 2022-08-30
Deleted old Console [WIP]

bvaughn:
How much code can I delete? We will see...

PR #7643, opened 2022-08-31
Delete old console and event listeners and event-listeners

bvaughn:
Builds on top of #7641 and deletes additional things:
- [x] Event types state
  - I'm not 100% about this one because the `Matches` UI component was referencing this Redux node, _but_ I'm not sure it was ever being used. (I didn't see anything actually setting the `enableEventLink` setting to true– so maybe it was just an old orphaned thing?)
- [x] Analysis logic in `src/ui/actions/logpoint`
   - This overlaps with the logic in the new Console. We don't need to run extra analysis. We should just share data between the old and new architecture.
- [x] Fixed a couple of bugs (see https://github.com/replayio/devtools/pull/7643#discussion_r969826546, https://github.com/replayio/devtools/pull/7643#pullrequestreview-1106084427, https://github.com/replayio/devtools/pull/7643#discussion_r969822136)


markerikson:
Per Jaril:

> Matches was a (very) hacky implementation of connecting a user event (eg mouse click) to the click handler
> Idea is if there was a click in the code, you'd want to be one click away from the react onclick callback
> but it was difficult to pull off for many reasons 🥲
> something something react handled events in a way that made it complicated to just get the location for that callback (i think)
> and then how that's handled was different across react versions

markerikson (approved):
I am in favor of going ahead and merging this PR instead of #7641 .

Reviewed both, they seem good to me.

LET'S KILL THAT CONSOLE!

bvaughn:
PR #7651 will help unblock this.

packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
```diff
@@ -73,6 +98,31 @@ export function getHitPointsForLocation(
...
+  condition: string | null,
+  focusRange: TimeStampedPointRange | null
+): Promise<HitPointsAndStatusTuple> {
+  try {
+    return getHitPointsForLocation(client, location, condition, focusRange);
+  } catch (errorOrPromise) {
+    if (
+      errorOrPromise != null &&
+      typeof errorOrPromise === "object" &&
+      typeof errorOrPromise.hasOwnProperty("then")
```
eddyw, line 116: 👀 the condition will be true with any object even when there is no `then` own property because `typeof` operator always returns a string
```suggestion
      errorOrPromise.hasOwnProperty("then")
```

bvaughn: Nice catch! Thanks :)

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -53,35 +48,71 @@ const connector = connect(
...
+  const filteredHitPoints = useMemo(() => {
+    return hitPoints
+      ? hitPoints.filter(hitPoint => {
+          return (
+            hitPoint.time >= unsafeFocusRegionForSuspense.beginTime &&
+            hitPoint.time <= unsafeFocusRegionForSuspense.endTime
+          );
+        })
+      : null;
+  }, [hitPoints, unsafeFocusRegionForSuspense]);
```
bvaughn, line 107: This section re-uses our new Suspense caches to load hit points data. Then we have a secondary pass to filter the hit points by the user-visible focus region. This is necessary, unfortunately, because we often ask for a focus region that's bigger than the time the user requested.

src/ui/components/PrefixBadge.tsx
```diff
@@ -15,7 +15,6 @@ export default function PrefixBadgeButton({ breakpoint }: { breakpoint: Breakpoi
 
   return (
     <PrefixBadgePicker
-      key={breakpoint.options.prefixBadge}
```
bvaughn, line 18: I'm not sure why we were doing this before, but it was breaking the prefix badge close animation. Now it works again.

src/ui/components/Timeline/UnloadedRegions.tsx
```diff
@@ -31,7 +27,21 @@ export const UnloadedRegions: FC = () => {
...
+    const userVisibleTimeStampedPointRange: TimeStampedPointRange = {
+      begin: {
+        point: unsafe.begin.point,
+        time: unsafe.beginTime,
+      },
+      end: {
+        point: unsafe.end.point,
+        time: unsafe.endTime,
+      },
+    };
```
bvaughn, line 42: This fixes an issue on the `main` branch where our mini-timeline showed a _larger_ loaded region than the user had focused on, which was confusing.

Before:
Screen Shot 2022-09-13 at 12 08 21 PM

After:
Screen Shot 2022-09-13 at 12 08 42 PM

bvaughn:
I think this is ready for a final review 😄

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
```diff
@@ -94,20 +91,22 @@ function BreakpointTimeline({
...
     const time = Math.ceil(beginTime + (endTime - beginTime) * clickPosition);
     setHoveredTime(time);
   };
-  const onMouseLeave = (e: React.MouseEvent) => setHoveredTime(null);
+  const onMouseLeave = () => setHoveredTime(null);
 
   const hoverPercent = `${getVisiblePosition({ time: hoveredTime, zoom: zoomRegion }) * 100}%`;
   const percent = getVisiblePosition({ time: currentTime, zoom: zoomRegion }) * 100;
 
+  // TODO [bvaughn] UnloadedRegions should be loaded + focused
```
jasonLaster, line 108: is this todo okay?

src/ui/actions/find-tests.ts
```diff
@@ -73,7 +71,6 @@ return [{
 // Mapper extracting the "name" and "message" properties from a thrown exception
 // for use in test failure messages.
 const JestExceptionMapper = `
-${Helpers}
```
jasonLaster, line 76: why is it okay to remove these? Seems unrelated

src/ui/components/PrefixBadge.tsx
jasonLaster: odd

src/ui/actions/find-tests.ts
markerikson: Good catch - the `Helpers` string was:

```ts
export const Helpers = `
  const finalData = { frames: [], scopes: [], objects: [] };
  function addPauseData({ frames, scopes, objects }) {
    finalData.frames.push(...(frames || []));
    finalData.scopes.push(...(scopes || []));
    finalData.objects.push(...(objects || []));
  }
  function getTopFrame() {
    const { frame, data } = sendCommand("Pause.getTopFrame");
    addPauseData(data);
    return finalData.frames.find(f => f.frameId == frame);
  }
`;
```

Note that it adds `const finalData`.

I see a reference to `finalData` in one of the Jest test-finding mappers:

```ts
const JestExceptionMapper = `
${Helpers}
const { point, time, pauseId } = input;
const { frameId, location } = getTopFrame();
const { exception, data: exceptionData } = sendCommand("Pause.getExceptionValue");
addPauseData(exceptionData);
return [{
  key: point,
  value: { time, pauseId, location, exception, data: finalData },
}];
`;
```

So we should re-add `Helpers` somewhere, and restore those lines here.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimeline.tsx
bvaughn: Ooops. Forgot to delete this comment. Yeah, I already did that. It's what was fixed here: https://github.com/replayio/devtools/pull/7643/files/606ba4950b4c24a83cf748ef3af266aa06b71afd#diff-c9da310bb17d439da080e71962563ba5a48b4740dd69f5243d3f8c33fff2f8d7

src/ui/actions/find-tests.ts
bvaughn: It's related. This is in `src/ui/actions` so it's Redux stuff, which this PR is deleting a bunch of. In this case, "Helpers" was referring to an export from `actions/logpoint` which was deleted in this PR :)

markerikson: Yeah, what I'm seeing is that the `JestExceptionMapper` string requires one of the variables created by that `Helper` string, so we need to bring it back.

bvaughn:
Cool. Good catch. Thanks.

Fixed that issue and rebased. I think we should be unblocked to land today, assuming CI is happy.

bvaughn:
The failing Playwright test is a false negative, and has been fixed by #7744. Otherwise things look green. I think we're good to land this.

PR #7644, opened 2022-08-31
Fix e2e tests

bvaughn:


bvaughn:
d060ab3 should fix `doc_exceptions_bundle.html`

PR #7648, opened 2022-08-31
Allow analysis to fail

jasonLaster:
Should fix the crash we were seeing with invalid location. Of course, we should also still have a backend command error that we want to hunt down to explain why we were sending the bad location as well

### DIFF
--- a/packages/protocol/analysisManager.ts
+++ b/packages/protocol/analysisManager.ts
@@ -160,6 +160,9 @@ class AnalysisManager {
         !handler.onAnalysisResult || client.Analysis.runAnalysis({ analysisId }, sessionId),
         !handler.onAnalysisPoints || client.Analysis.findAnalysisPoints({ analysisId }, sessionId),
       ]);
+    } catch (e) {
+      this.onAnalysisError({ analysisId, error: e instanceof Error ? e.message : "Unknown Error" });
+      console.error(e);
     } finally {
       this.handlers.delete(analysisId);
       await client.Analysis.releaseAnalysis({ analysisId }, sessionId);
--- a/src/ui/setup/helpers.ts
+++ b/src/ui/setup/helpers.ts
@@ -1,6 +1,7 @@
 import { UIStore } from "ui/actions";
 import { getRecordingId } from "ui/utils/recording";
 import { prefs, features, asyncStore } from "ui/utils/prefs";
+import { asyncStore as debuggerAsyncStore } from "devtools/client/debugger/src/utils/prefs";
 
 // eslint-disable-next-line no-restricted-imports
 import { triggerEvent, sendMessage, client } from "protocol/socket";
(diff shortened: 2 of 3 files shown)


src/ui/setup/helpers.ts
```diff
@@ -55,6 +57,7 @@ export async function setupAppHelper(store: UIStore) {
     prefs,
     features,
     asyncStore,
+    debuggerAsyncStore,
```
bvaughn, line 60: This seems unrelated to this PR. What is it for?

packages/protocol/analysisManager.ts
```diff
@@ -160,6 +160,9 @@ class AnalysisManager {
         !handler.onAnalysisResult || client.Analysis.runAnalysis({ analysisId }, sessionId),
         !handler.onAnalysisPoints || client.Analysis.findAnalysisPoints({ analysisId }, sessionId),
       ]);
+    } catch (e) {
+      // log the error if the analysis fails
```
bvaughn, line 164: nit This comment is kind of useless 😁

packages/protocol/analysisManager.ts
```diff
@@ -160,6 +160,9 @@ class AnalysisManager {
...
+      console.error(e);
```
bvaughn, line 165: Shouldn't this be calling `onAnalysisError` somehow? Otherwise won't it just hang forever?

bvaughn (commented):
Not sure I understand this fix.

packages/protocol/analysisManager.ts
jasonLaster: lol - so true...

packages/protocol/analysisManager.ts
jasonLaster: Not sure, i think `onAnalysisError` is reserved for the response failing, as opposed to the request not being created... subtle difference.

bvaughn: Hm, interesting. So what's the expected behavior from the POV of the thing calling the API? It will just await the result and then it will need to handle the fact that neither complete nor error were called?

jasonLaster: Honestly, I gave this 10 minutes of thought earlier. My sole goal was don't crash. We should probably take another stab tomorrow

jcmorrow: If the analysis fails, I think we should call `onAnalysisError`, regardless of what the error is.

jcmorrow (approved):

src/ui/setup/helpers.ts
bvaughn: Was this change intentional to land? @jcmorrow

jcmorrow: Oh, I totally forgot that this was even in here. It seems unrelated. I'll PR it back out.

PR #7649, opened 2022-08-31
Consolidate uses of `Highlighter.highlight()` and node picker logic

markerikson:
This PR:

- Consolidates all previous outstanding uses of `Highlighter.highlight/unhighlight()`, so that there's only _one_ file that calls those: `actions/markup.ts`. All other uses now call `dispatch(highlightNode(nodeId))`.
- Rewrites the `<NodePicker>` component to consolidate implementation logic, as it had exactly duplicate behavior with the `ui/utils/nodePicker` vanilla JS class version:
  - Rewrote it to be a function component with hooks instead of a connected class component
  - It now instantiates an instance of the `NodePicker` JS class to do all of the `mousemove` event listeners and "which node is underneath the mouse?" logic, instead of duplicating that
  - It still exposes a `window.gNodePicker` global variable for use in `test/harness.ts` (which we should get rid of), but that's now an object with two methods instead of the React class component `this` instance

These are prep steps towards rewriting the highlighting implementation entirely to no longer be a singleton that relies on `NodeFront`s.

### DIFF
--- a/src/devtools/client/inspector/boxmodel/components/BoxModel.tsx
+++ b/src/devtools/client/inspector/boxmodel/components/BoxModel.tsx
@@ -5,7 +5,7 @@
 "use strict";
 
 import React from "react";
-import { BoxModelMain } from "./BoxModelMain";
+import BoxModelMain from "./BoxModelMain";
 import { BoxModelProperties } from "./BoxModelProperties";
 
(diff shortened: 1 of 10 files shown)


markerikson:
Confirmed highlighting still works in all cases.

PR #7659, opened 2022-09-02
Fix the fallback logic for inconsistent sourceIds (FE-671)

hbenl:
This used to throw when no source to display for the given URL was found.

### DIFF
--- a/src/devtools/client/debugger/src/actions/sources/select.ts
+++ b/src/devtools/client/debugger/src/actions/sources/select.ts
@@ -132,9 +132,10 @@ export function selectLocation(
     if (location.sourceUrl && location.sourceUrl !== source?.url) {
       await ThreadFront.ensureAllSources();
       const state = getState();
-      const sourceId = getSourceToDisplayForUrl(state, location.sourceUrl)!.id;
-      source = getSourceDetails(state, sourceId);
-      location = { ...location, sourceId };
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/actions/sources/select.ts
```diff
@@ -132,9 +132,10 @@ export function selectLocation(
     if (location.sourceUrl && location.sourceUrl !== source?.url) {
       await ThreadFront.ensureAllSources();
       const state = getState();
-      const sourceId = getSourceToDisplayForUrl(state, location.sourceUrl)!.id;
-      source = getSourceDetails(state, sourceId);
-      location = { ...location, sourceId };
+      source = getSourceToDisplayForUrl(state, location.sourceUrl);
+      if (source) {
+        location = { ...location, sourceId: source.id };
```
bvaughn, line 137: Is the `!source` case correct? We'll end up calling...
```js
    dispatch(locationSelected({ location, source }));
```
...with a `PartialLocation` rather than a `Location`. That seems like it might cause trouble in other places?

hbenl: No, we'll end up calling `return dispatch(clearSelectedLocation());`

bvaughn (approved):
I don't really have a lot of context here. I guess this fix is a good change?

PR #7662, opened 2022-09-02
consolidate design system components

souporserious:
When I first started design system efforts I was putting things in a `components` package. This PR moves `AddCommentButton` and `PrefixBadgePicker` into the `design` package and removes the other components as they were relying on Tailwind. I have some ideas for how we can handle Grid/Column/Row components in the future.


jasonLaster (approved):

bvaughn:
I think this broke frontend e2e tests (locally and on CI, from what I can tell).
Screen Shot 2022-09-08 at 9 50 45 AM

bvaughn:
FYI I am reverting this via #7699 for now.

PR #7663, opened 2022-09-02
Better error message for outdated Replays

jonbell-lot23:
Was:
"This error has been fixed in an updated version of Replay. Please try upgrading Replay and trying a new recording."

Now:
"This recording is no longer available because we have updated Replay. Please try recording a new replay."

To an end user, this isn't an "error" that is fixed by "upgrading Replay." It's more like a 404, where it's just not going to work, and if you didn't record the original replay, there's nothing you can do. So this makes the error more clear for this case.


jonbell-lot23:
Fixes https://linear.app/replay/issue/FE-687/session-died-crash

PR #7664, opened 2022-09-03
Remove debugger async store

jcmorrow:
See #7648 for context.


jasonLaster:
@jcmorrow I'd like to keep this because it makes it easier to see what the pending breakpoints are and debug some of these issues 

`await app.debuggerAsyncStore.pendingBreakpoints`

jcmorrow:
> @jcmorrow I'd like to keep this because it makes it easier to see what the pending breakpoints are and debug some of these issues
> 
> `await app.debuggerAsyncStore.pendingBreakpoints`

Ah, cool, as long as it has a purpose 👍

PR #7666, opened 2022-09-03
Refine the focus window as we receive points

jcmorrow:
I'm not sure how this particular feature never got in. I remember
thinking last time I was working on focus stuff that we should do it,
but I guess I never actually got around to doing it. The idea is
relatively simple:

Sometimes the user asks for a focus region in which our known timeline
is sparse. We do our best to get them an accurate window, but it's tough
because points are few and far between. But there is *good* news there,
which is that we can't display anything incorrectly, because we don't
know about anything in the sparse ranges (that's why they are sparse).

But, then the trouble starts. The user has chosen a window, and they
start doing things like settings breakpoints. We are filling in our
sparse timeline! But sometimes those points actually fall *between* the
displayed end of the focus region and real end of the window for which
we are currently making requests. When this happens, you get
funny-looking results, like analysis points hanging out in regions that
look unloaded. The solution is actually relatively simple. When we are
focused and we discover new points, we check to see if any of those
points could be used to make a more refined focus window. If so, we just
update our focus window. No additional interactions with the backend are
required, and if we have a really dense analysis, that's fine, because
it just means we will get even *more* accurate information about the
point <-> time relationship in that region.

### DIFF
--- a/packages/protocol/analysisManager.ts
+++ b/packages/protocol/analysisManager.ts
@@ -1,3 +1,4 @@
+import { preCacheExecutionPointForTime } from "bvaughn-architecture-demo/src/suspense/PointsCache";
 import {
   addEventHandlerEntryPointsParameters,
   addFunctionEntryPointsParameters,
@@ -10,6 +11,7 @@ import {
   ExecutionPoint,
   PointDescription,
   PointRange,
+  TimeStampedPoint,
 } from "@replayio/protocol";
 import { addEventListener } from "protocol/socket";
 import { ThreadFront } from "protocol/thread";
@@ -85,6 +87,11 @@ export interface AnalysisHandler<T> {
 
 // When running analyses in batches, limit on the points to use in each batch.
 const MaxPointsPerBatch = MAX_POINTS_FOR_FULL_ANALYSIS;
+
+let onPointsReceived: (points: TimeStampedPoint[]) => void;
+export function setPointsReceivedCallback(callback: typeof onPointsReceived): void {
+  onPointsReceived = callback;
+}
 
 class AnalysisManager {
   private handlers = new Map<AnalysisId, AnalysisHandler<any>>();
@@ -208,9 +215,17 @@ class AnalysisManager {
   };
 
   private readonly onAnalysisPoints = ({ analysisId, points }: analysisPoints) => {
+    if (typeof onPointsReceived === "function") {
+      onPointsReceived(points);
+    }
     const handler = this.handlers.get(analysisId);
+    points.forEach(point => {
+      preCacheExecutionPointForTime(point);
+      if (handler != null && typeof handler.onAnalysisPoints === "function") {
+        ThreadFront.updateMappedLocation(point.frame);
+      }
+    });
     if (handler != null && typeof handler.onAnalysisPoints === "function") {
-      points.forEach(point => ThreadFront.updateMappedLocation(point.frame));
       handler.onAnalysisPoints(points);
     }
   };
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -34,7 +34,7 @@ import {
   getZoomRegion,
   getShowFocusModeControls,
   setPlaybackPrecachedTime,
-  pointsReceived,
+  pointsReceivedThunk,
 } from "ui/reducers/timeline";
 import { FocusRegion, HoveredItem } from "ui/state/timeline";
 import {
@@ -100,7 +100,7 @@ export async function setupTimeline(store: UIStore) {
 export function jumpToInitialPausePoint(): UIThunkAction {
   return async (dispatch, getState, { ThreadFront, replayClient }) => {
     const endpoint = await replayClient.getSessionEndpoint(replayClient.getSessionId()!);
-    dispatch(pointsReceived([endpoint]));
+    dispatch(pointsReceivedThunk([endpoint]));
     let { point, time } = endpoint;
 
     const state = getState();
@@ -198,7 +198,7 @@ export function setTimelineToTime(time: number | null, updateGraphics = true): U
   };
 }
 
-function updatePausePointParams({
+export function updatePausePointParams({
   point,
   time,
   hasFrames,
@@ -234,6 +234,7 @@ export function seek(
       ThreadFront.timeWarpToPause(pause);
     } else {
       const regions = getLoadedRegions(getState());
+      const focusRegion = getFocusRegion(getState());
       const isTimeInLoadedRegion = regions !== null && isTimeInRegions(time, regions.loaded);
       if (isTimeInLoadedRegion) {
         ThreadFront.timeWarp(point, time, hasFrames);
@@ -242,6 +243,7 @@ export function seek(
         // In this case the best we can do is update the current time and the "video" frame.
         dispatch(setTimelineState({ currentTime: time }));
         dispatch(setTimelineToTime(time, true));
+        updatePausePointParams({ point, time, hasFrames, focusRegion });
       }
     }
     return true;
--- a/src/ui/reducers/timeline.ts
+++ b/src/ui/reducers/timeline.ts
@@ -1,14 +1,14 @@
 import { TimeStampedPoint } from "@replayio/protocol";
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { UIState } from "ui/state";
-import { FocusRegion, HoveredItem, TimelineState, ZoomRegion } from "ui/state/timeline";
 import {
   displayedEndForFocusRegion,
   displayedBeginForFocusRegion,
   mergeSortedPointLists,
 } from "ui/utils/timeline";
+import { FocusRegion, HoveredItem, TimelineState, UnsafeFocusRegion } from "ui/state/timeline";
 import sortBy from "lodash/sortBy";
-import maxBy from "lodash/maxBy";
+import { UIThunkAction } from "ui/actions";
 
 function initialTimelineState(): TimelineState {
   return {
@@ -71,6 +71,12 @@ const timelineSlice = createSlice({
   },
 });
 
+// If abs(A - B) > EPSILON, then A and B are considered different.
+// If abs(A - B) < EPSILON, then A and B are considered equal.
+const EPSILON = 0.0001;
+// returns true is A is less than B, and the difference is greater than EPSILON
+const lessThan = (a: number, b: number) => b - a > EPSILON;
+
 export const {
   allPaintsReceived,
   setHoveredItem,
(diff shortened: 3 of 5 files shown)


src/ui/actions/timeline.ts
```diff
@@ -125,7 +126,9 @@ export function jumpToInitialPausePoint(): UIThunkAction {
       hasFrames = initialPausePoint.hasFrames;
       time = initialPausePoint.time;
     }
-    ThreadFront.timeWarp(point, time, hasFrames);
+    if (isPointInLoadingRegion(state, point)) {
```
jasonLaster, line 129: what's this if statement doing?

src/ui/reducers/timeline.ts
```diff
@@ -84,6 +91,32 @@ export const {
...
+export const pointsReceivedThunk = (points: TimeStampedPoint[]): UIThunkAction => {
+  return (dispatch, getState) => {
+    const state = getState() as UIState;
+    dispatch(pointsReceived(points));
+    const focusRegion = getFocusRegion(state) as UnsafeFocusRegion | null;
+    if (!focusRegion) {
+      return;
+    }
+    const betterFocusStart = points.find(
+      p => BigInt(p.point) > BigInt(focusRegion.begin.point) && p.time < focusRegion.beginTime
```
jasonLaster, line 103: this violates the monotonic nature of points and time which is super weird

jasonLaster (approved):

src/ui/reducers/timeline.ts
bvaughn: I...think...this should be...
```js
p => BigInt(p.point) < BigInt(focusRegion.begin.point) && p.time >= focusRegion.beginTime
```

src/ui/reducers/timeline.ts
```diff
@@ -84,6 +91,32 @@ export const {
...
+    );
+    const betterFocusEnd = points.find(
+      p => BigInt(p.point) < BigInt(focusRegion.end.point) && p.time > focusRegion.endTime
```
bvaughn, line 106: This also seems weird.

jcmorrow:
I think it's very reasonable that there is confusion around what this code does, so I'll take this as a good prompt to draw some pictures!

jcmorrow:
This might be a terrible explanation, and if so I am sorry 😆 
Random Drawings

src/ui/actions/timeline.ts
jcmorrow: If we are not actually loading the point that this would take us to, we should not try and time warp there, it will just cause a protocol error. If the URL also contained a focus window, then the point will be inside the range and we'll be all good 👍

jcmorrow: This is a little unrelated to the focus window stuff, but I spotted it and figured while I was in here I would be a little more defensive about pause creation.

src/ui/reducers/timeline.ts
bvaughn: Prior to seeing @jcmorrow's diagram, I thought this code was looking to _expand_ the focus region slightly– to include points that are within the user-defined _time_ but outside of the previous points boundaries we knew about.

I now understand this code to be looking to narrow down the points of the focus window itself so that it more tightly fits the user's specified time– so the logic check makes sense  (although it could use an inline comment about its purpose I think).

That being said, should we also add code for the case I mentioned above? Let's say the user selected a range of 500-1,000 ms, and we _had_ a local execution point for 500 ms– but later we happened to get another _earlier point_ that was also at 500ms. I guess this wouldn't happen in this reducer, since we scoped our analysis by the later cutoff point? (This stuff feels confusion-prone, given that we operate with execution points while presenting _time_.)

bvaughn (approved):
Approved but please add some kind of inline comment to the focus region refinement code.

src/ui/reducers/timeline.ts
jcmorrow: @bvaughn I am not sure I understand the second part? Like, if the user has said they want to load starting at millisecond 550.5 and our nearest point is 550.1, but we find a new point at 550.4... then we should use it? The window can get arbitrarily small, but as long as the relationship between point and time is a monotonic function the logic is always correct.

jcmorrow: I suppose there could be some kind of floating point imprecision at some level of arbitrary representation, but if we are in that far, then I don't think changing the focus window should matter? The one exception to this might be if the user has specified a *point* they want to use, and because of a combination of the refinement logic + floating point imprecision we move their point. In general, we should probably have a "the user asked for this point" flag, so that we don't do these gymnastics.

bvaughn: > Like, if the user has said they want to load starting at millisecond 550.5 and our nearest point is 550.1, but we find a new point at 550.4

No. Different ms values would be handled by the client already, since we have guards to make sure the range is _at least as big as_ the one you defined.

I'm asking about ... imprecision of time stamps. I guess it's not super common, but multiple calls to get system time can return the same value– even though those calls have different execution points. I'm asking about _that_ case.

jcmorrow: I've added an epsilon for this case.

bvaughn:
@jcmorrow I have to run soon for my brother's birthday dinner, but I'm happy to help here. Want to maybe pair on this tomorrow morning?

jcmorrow:
Yeah that sounds good!

bvaughn:
Great! Want to toss something on the calendar?

I tried adding something tomorrow but your first available time didn't appear to be until 6pm which is a bit later than I hope to be working since it's the last day I'll be in VA with family before heading back.

jcmorrow:
OK, I think this is ready for one more look @bvaughn! (I don't know why lint is failing though)

bvaughn:
> I don't know why lint is failing though

I think #7721 was merged with a TSC failure. I wonder if that could be related (since Trunk/Lint are now using TS lint rules too.)

packages/protocol/analysisManager.ts
```diff
@@ -1,3 +1,4 @@
+import { preCacheExecutionPointForTime } from "@bvaughn/src/suspense/PointsCache";
```
bvaughn, line 1: I don't know if this package knows what `@bvaughn` is. This import should be:
```suggestion
import { preCacheExecutionPointForTime } from "bvaughn-architecture-demo/src/suspense/PointsCache";
```

src/ui/reducers/timeline.ts
```diff
@@ -71,6 +71,8 @@ const timelineSlice = createSlice({
   },
 });
 
+const EPSILON = 0.0001;
```
bvaughn, line 74: What is this?

I don't think this word has a standalone meaning that provides enough context into what it is.

src/ui/reducers/timeline.ts
bvaughn: TBH I don't understand how that fixes this.

src/ui/reducers/timeline.ts
```diff
@@ -82,6 +84,46 @@ export const {
...
+      return;
+    }
+    // If we have just received points that we did not know about, those points
+    // might represent a better fit for the user-requested focus window than
+    // whatever points we were using before. If so, we will narrow the focus
+    // region down farther to the best-found points. However, in the case that
+    // we might be mucking about with time beyond the precision of floating
+    // point numbers, we should just chill, we are close enough to the user
+    // specified boundary (hence the epsilon check).
+    // See https://github.com/replayio/devtools/pull/7666 for more info.
```
bvaughn, line 102: Why? Is it not true that we might filter out logs/events/network requests that are technically within the time range the user selected just because they're really close to a point we already knew about. I don't understand why this threshold is a good thing.

bvaughn (approved):
TBH I don't understand the epsilon thing. Maybe it's a good idea and I'm missing something but it seems like the potential edge case I raised still exists and I don't see how the epsilon helps (or why it's even related conceptually TBH).

Stamping approve because I generally think this PR is a good thing, just not sure about that part.

src/ui/reducers/timeline.ts
markerikson: Could maybe rename this to be a bit clearer, but "Epsilon" is a standard term for "amount of tolerance when comparing floats":

- https://en.wikipedia.org/wiki/Machine_epsilon

jcmorrow: Yeah, that's the meaning I was using, but you're right, it's not obvious to people who have not done floating point stuff much.

src/ui/reducers/timeline.ts
jcmorrow: > I'm asking about ... imprecision of time stamps. I guess it's not super common, but multiple calls to get system time can return the same value– even though those calls have different execution points. I'm asking about that case.

If I ask for the system time and I get back `1,000` and then in a few nanoseconds I ask again and I get `1,000` again, then it seems fine to include both of those points in any window that includes either of them? We are trying to fit sub-millisecond precision to a draggable UI, which even on a short replay probably can't be any more precise than a couple of milliseconds, so I don't see what the harm is in choosing the "wrong" nanosecond point to bound that imprecise window by? Besides, this is the case where not only did that weird thing with the system clock happen, but also that was the *exact* time we registered the mouse as picking when dragging around the focus window. Maybe I'm missing some further scenario?

jcmorrow: But to explain the epsilon thinking further, consider this scenario (I'll annotate points as `ms:point` for brevity:

1. The user indicates with their mouse that they want their window to end at `100.5`.
2. We don't have a point at `100.5`, but we have one at `110.1:123`, so we use that as our point.
3. The user runs an analysis and we end up finding a new point at `101.099:120`. I think we all agree that we should use that point as our new endpoint.
4. The user runs a further analysis, and we find a new point at `100.500099:117`. What do we do? Well, that number is A) a better fit for the window and B) significantly different from our current choice. Therefore, we should use that as our new fit. We are now within epsilon of the requested window.
5. Another points comes in, this time at `100.500095:118`. Now, notice, our `point` part actually went *up*, even though our time value is still technically better. Now we might end up *expanding* the focus window, and later, if another of these cases happens, we could end up shrinking it further again, causing something a little like z-fighting in graphics (https://en.wikipedia.org/wiki/Z-fighting). If we calculate the amount that our new window would be better by we get  `100.500099 - 100.500095`, or `0.000003999999989900971`. That value is smaller than epsilon. We can not really any longer say whether something is a better fit, because we're so close that we are approaching floating point precision (not really, actually, I think we're still several orders of magnitude off, but we are *definitely* approaching the limits of how precise a human hand can drag a mouse, and anyhow we definitely *could* reach the limits to floating point precision, so we might as well call it here rather than at `0.000001`).

Another way to think of epsilon is "What if we multiplied the number by `1/epsilon` and then `Math.floor`ed it so we could just compute in integers?".

Anyways, it sounds like the case that you are thinking of (the system clock gives the same time number for multiple points) would be fine to just leave both points in the window, and this would also accomplish that.

src/ui/reducers/timeline.ts
```diff
@@ -71,7 +71,10 @@ const timelineSlice = createSlice({
   },
 });
 
+// If abs(A - B) > EPSILON, then A and B are considered different.
+// If abs(A - B) < EPSILON, then A and B are considered equal.
 const EPSILON = 0.0001;
+const lessThan = (a: number, b: number) => b - a > EPSILON;
```
bvaughn, line 77: I feel kind of stupid but why is this named "less than" ? isn't it actually checking if the delta is greater than ?

bvaughn (approved):

src/ui/reducers/timeline.ts
bvaughn: I'm still kind of not understanding TBH

src/ui/reducers/timeline.ts
jcmorrow: It's checking if A is less than B, taking epsilon into account. So it will also return false if B is greater than A but the difference is larger than epsilon.

jcmorrow: would it be better to name it like `definitelyLessThan`?

bvaughn: I think the thing that confused me is that it's checking if the delta between a and b is greater than a minimal threshold.

I don't care. It doesn't matter. I was just confused by the name.

The thing I care about is the question above.

PR #7668, opened 2022-09-04
Return the generated source ID until all sources have come in

jcmorrow:
Right now it's pretty rare that console messages will come in before
sources, but it's possible, and when that happens we get bombarded with
`unknown source` errors as we try to find preferred locations without
having source details loaded yet. By returning the generated location
until we actually load sources, we avoid those errors. There might be an
even better way to solve this problem, but this one seems to work well
in my tests (I just threw a fake await in before actually processing the
sources).

### DIFF
--- a/src/ui/reducers/sources.ts
+++ b/src/ui/reducers/sources.ts
@@ -359,6 +359,9 @@ export function getPreferredLocation(
   locations: MappedLocation,
   preferredGeneratedSources: Set<string>
 ) {
+  if (!state.sources.allSourcesReceived) {
+    return locations[0];
+  }
   const sourceId = getPreferredSourceId(
     getSourceDetailsEntities(state),
     locations.map(l => l.sourceId),


jasonLaster:
Curious what @hbenl and @markerikson think 

Two thoughts that come to mind:
1. what does this mean when we get the original source later?
2. are there ways to make this safer? Fetch messages after we have sources? Wait for the source before dispatching the action...

hbenl:
I think this is a bad idea because users will see different results in their sessions depending on the timing. Instead, we should wait for the sources to be loaded before processing the console messages (we used to do that but I guess the new console doesn't do it yet).

jcmorrow:
> I think this is a bad idea because users will see different results in their sessions depending on the timing. Instead, we should wait for the sources to be loaded before processing the console messages (we used to do that but I guess the new console doesn't do it yet).

I understand the concern. FWIW, with this patch the console messages load but the right hand side where the source would be says `Loading...` until the sources are fully processed, which I think is better than the console messages not appearing at all until the sources are in. So I think it would be great if we could keep that (even if we express it another way than this, because that's not what this code really says, and I don't understand why that's the effect it has, it's just a happy coincidence).

markerikson:
Okay, I think I see what's going on.

In `<MessageRenderer>`, we have:

```ts
            {frame?.location?.length > 0 && <Source locations={frame.location} />}
          </Suspense>
```

so each `<Source>` is individually suspendable.

Inside `<Source>`:

```ts
  const location = client.getPreferredLocation(locations);
  if (location == null) {
    return null;
  }

  const source = getSource(client, location.sourceId);
  if (source == null) {
    return null;
  }
```

Hmm. Based on that, it should be bailing out _before_ we get to the `getSource()` call that suspends.

Lemme poke at this a bit more.

markerikson:
Okay, after making a replay at https://app.replay.io/recording/new-console-preferred-sources-loading-with-location-bailout-fix--9699f9ee-28fc-44f9-8744-dea868cfe7df and looking through it a bit, I'm pretty sure these components re-render later.  So, yeah, they might start with an incorrect `sourceId`, but it'll get corrected when we do have the real data available.

Let's get this in as a bandaid and move on.

bvaughn:
I am worried about the number of "bandaids" we're landing each week

PR #7670, opened 2022-09-04
Show the comments panel

jasonLaster:
This fixes FE-691, but might have upset the Console redux purity gods....


jasonLaster (commented):
Looks much better

packages/bvaughn-architecture-demo/src/contexts/InspectorContext.ts
```diff
@@ -18,6 +20,7 @@ type InspectHTMLElement = (
 export type InspectorContextType = {
```
bvaughn, line 20: This change (injecting a param) seems like the right one 👍🏼 

I think maybe we should _rename_ this context at this point though. It's not just 'inspecting' things anymore. It's kind of just a generic adapter between legacy and new architectures.

PR #7671, opened 2022-09-04
Move the error boundary up a level

jasonLaster:
Tested this by adding this error

```diff
diff --git a/packages/shared/client/ReplayClient.ts b/packages/shared/client/ReplayClient.ts
index 99b46ee11..8eefeba16 100644
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -415,6 +415,7 @@ export class ReplayClient implements ReplayClientInterface {
       }
     }

+    throw new Error("oops");
     return [collectedHitPoints, status];
   }
```

Screen Shot 2022-09-04 at 11 19 24 AM

### DIFF
--- a/packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
+++ b/packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
@@ -3,6 +3,7 @@
   text-align: center;
   background-color: var(--background-color-error);
   color: var(--color-default);
+  grid-column: 2;
 }
 
 .Logo {
--- a/packages/bvaughn-architecture-demo/components/ErrorBoundary.tsx
+++ b/packages/bvaughn-architecture-demo/components/ErrorBoundary.tsx
@@ -27,7 +27,8 @@ export default class ErrorBoundary extends Component<PropsWithChildren<{}>, Erro
           <ReplayLogo />
           <div className={styles.Header}>Our apologies!</div>
           <div className={styles.Message}>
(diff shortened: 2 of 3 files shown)


packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
```diff
@@ -3,6 +3,7 @@
   text-align: center;
   background-color: var(--background-color-error);
   color: var(--color-default);
+  grid-column: 2;
```
jcmorrow, line 6: What's up here?

jcmorrow (commented):
Makes sense to me I think?

packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
jasonLaster: this fixes a styling issue

bvaughn: I don't think this was the right place to add the style. The `ErrorBoundary` component is a generic component. It might be shown in a lot of different areas, so styling it for a specific instance of a grid seems like it will cause problems in other places.

I think a better solution here would probably be to make the fallback UI configurable in some way (even just via some optional prop like `fallbackClassName`).

bvaughn: I went to make this change and the `grid-column` style doesn't actually seem necessary.

PR #7672, opened 2022-09-05
Await pause

jasonLaster:
There are several places where we create a pause and try to then use it only to learn that it could not be created.

In many of these places we first await `ensureLoaded`, but in those case it could fail silently as well. What I'm proposing is that we have `ensurePause` return the `createAwaiter` so that when we have the pause we know it is legit.

This will for example help us with async call stacks where we need the pause, but the async frame could be outside the loaded region


src/devtools/client/debugger/src/actions/pause/paused.ts
```diff
@@ -47,16 +47,13 @@ export function paused({
...
-    dispatch(pausedAction({ executionPoint, time, id: pause.pauseId!, frame }));
-
+    const pause = await ThreadFront.ensurePause(executionPoint, time);
     const cx = getThreadContext(getState());
 
+    dispatch(pausedAction({ executionPoint, time, id: pause.pauseId!, frame }));
     try {
-      await pause.createWaiter;
+      // @ts-expect-error optional time mismatch
+      await ThreadFront.ensurePause(executionPoint, time);
```
jasonLaster, line 56: i'm pretty unsure about this

PR #7676, opened 2022-09-06
Temporary fix for the new console and corresponding sources (FE-702)

hbenl:
This is just a temporary fix, we should discuss how we want to handle corresponding sources in the new architecture.

### DIFF
--- a/src/ui/utils/preferredLocation.ts
+++ b/src/ui/utils/preferredLocation.ts
@@ -1,7 +1,10 @@
 import { MappedLocation } from "@replayio/protocol";
 import { ThreadFront } from "protocol/thread";
 import { UIStore } from "ui/actions";
-import { getPreferredLocation as getPreferredLocationSelector } from "ui/reducers/sources";
+import {
+  getPreferredLocation as getPreferredLocationSelector,
+  getCorrespondingSourceIds,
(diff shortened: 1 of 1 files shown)


src/ui/utils/preferredLocation.ts
```diff
@@ -1,18 +1,33 @@
...
+    return;
+  }
+  const state = store.getState();
+  // TODO [hbenl] another hack: the new console doesn't update sourceIds in locations
+  // to their first corresponding sourceId, which getPreferredLocationSelector
+  // subsequently complains about
+  const correspondingLocations = locations.map(location => ({
+    ...location,
+    sourceId: getCorrespondingSourceIds(state, location.sourceId)[0],
+  }));
```
bvaughn, line 26: What's needed to make this _actually_ happen?

There was a Linear issue that got closed by a temporary fix but do we have a non-temporary issue open that describes what needs to be done?

hbenl: I wanted to discuss in one of our frontend meetings how we handle corresponding sources in the new architecture. Personally I think we should do what the old architecture did: update all locations as soon as we receive them from the backend, but I'm open to other solutions.

PR #7677, opened 2022-09-06
Don't throw on Analysis errors (FE-690)

hbenl:


markerikson:
I think this seems reasonable conceptually - what's the net behavior difference at runtime?

markerikson (approved):

jcmorrow:
> I think this seems reasonable conceptually - what's the net behavior difference at runtime?


@markerikson I think this is to make it less likely that the console error boundary gets triggered and causes the whole thing to turn into "Oops, sorry, something crashed"

markerikson:
Sure, that seems like a good thing :)

bvaughn:
What's the UX of this change? (What actually happens in the Console?)

Would have probably been nice to see a video or something.

PR #7678, opened 2022-09-06
Rewrite box model highlighting logic as a simpler React component

markerikson:
This PR:

- Clears out all DOM node data whenever the user hits "play", since we're no longer paused
- Reimplements the core box model highlighting overlay behavior as a much simpler React component (reusing the "quads" calculation logic and CSS transform behavior)
- Sorta gets the "highlighting infobar" behavior working, but leaves it disabled for now
  - This shows the `div#id.className` contents, with colored text for each segment
  - The legacy implementation _was_ trying to update the position and contents of this infobar, but somehow the CSS layout was leaving it buried underneath the rest of the preview area and invisible
  - I got it displaying, but the layout was only working for "on top of" the highlighted node, and the styling is ugly. So, leaving it as a separate component that isn't being rendered at all, and we can revisit that later
- Rewrites the logic that fetches box model data to call the `protocolClient` directly instead of `NodeFront` methods
- Deletes the legacy highlighting implementation entirely


src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -28,9 +31,9 @@ import {
 
 import { paused } from "devtools/client/debugger/src/reducers/pause";
 
-import Highlighter from "highlighter/highlighter";
 import NodeConstants from "devtools/shared/dom-node-constants";
 import { features } from "devtools/client/inspector/prefs";
+import { clearTimeout } from "timers";
```
hbenl, line 36: This overrides the browser's `clearTimeout()` function (I guess that VS Code automatically added this import although this is not what we want).

hbenl (approved):
LGTM except the one import that needs to be removed.

src/devtools/client/inspector/markup/actions/markup.ts
markerikson: Whoops, yup, will fix!

PR #7680, opened 2022-09-06
Turn query cache off by default

jcmorrow:
See conversation here: https://github.com/replayio/backend/pull/6299

Also: keep the new architecture on by default *in tests*, because some of them permafail when it's turned off.

### DIFF
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -91,6 +91,7 @@ if (typeof window !== "undefined") {
 export type ExperimentalSettings = {
   listenForMetrics: boolean;
   disableCache?: boolean;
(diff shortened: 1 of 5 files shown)


bhackett1024:
The query cache needs to be turned off in the backend.  If the frontend sets experimental settings by default that are different from what the backend uses when replaying newly created recordings, it will greatly slow down the loading experience for new recordings.

jcmorrow:
@bhackett1024 that's fine, but we still need a flag to be able to test it ourselves, thanks.

markerikson (approved):

markerikson (approved):

PR #7684, opened 2022-09-06
Simpler default editor view

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-73/our-default-editor-screen-should-be-more-like-vs-code

* Removed colored background
* Added VS Code-style default editor
* Synced with Travis

This is a down payment. We can work with the visuals in a follow-up ticket to make it more polished.

image


jcmorrow:
@jonbell-lot23 feel free to totally disregard this comment, but I wonder if it would be helpful to add like a link to Replay's docs or some other "get inspired to use this tool" entry point. A thing that keeps ringing through my mind was the user feedback from the call last week: "it looks like a debugger, so I just defaulted to my normal flow using a debugger". I've been trying to think of how we can center the `it's really not just another debugger, it's a new way to ask questions about a program's behavior` point of view.

jonbell-lot23:
> @jonbell-lot23 feel free to totally disregard this comment, but I wonder if it would be helpful to add like a link to Replay's docs or some other "get inspired to use this tool" entry point. A thing that keeps ringing through my mind was the user feedback from the call last week: "it looks like a debugger, so I just defaulted to my normal flow using a debugger". I've been trying to think of how we can center the `it's really not just another debugger, it's a new way to ask questions about a program's behavior` point of view.

Great feedback and I totally agree. Here was my first rough sketch:

image

souporserious (approved):
Woot, love going back the simple route! 😌 I agree with Josh and think it'd be nice to add some more links to help people understand Replay better and something we can think about with the NUX project moving forward.

PR #7690, opened 2022-09-07
Don't show logpoints outside focus, to avoid analysis errors

markerikson:
This PR:

- Updates the new console "loggables" rendering to skip showing `<LogPointRenderer>` instances that are outside of the current focus region. This avoids showing "Analysis error" messages due to analysis requests for non-focused points in time.

Not sure if this is the _ideal_ solution, but I can now drag a focus region back and forth and see the logpoints appear and disappear, recalculating themselves as needed, with no "Analysis error" messages happening.

Loom:

https://www.loom.com/share/a94915c2ebd54e9f9ef95f9726bfbaf6

### DIFF
--- a/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
@@ -9,7 +9,7 @@ import {
   isTerminalExpression,
   isUncaughtException,
 } from "@bvaughn/src/utils/loggables";
-import { isExecutionPointsLessThan } from "@bvaughn/src/utils/time";
+import { isExecutionPointsLessThan, isExecutionPointsWithinRange } from "@bvaughn/src/utils/time";
 import { ForwardedRef, forwardRef, MutableRefObject, ReactNode, useContext, useMemo } from "react";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import Icon from "../Icon";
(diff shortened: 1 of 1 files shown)


packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -98,14 +98,28 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
...
+        // (since they have analysis that results in errors otherwise ).
+        // TODO Is this _too_ restrictive?
+
+        const isFocused =
+          focusRange === null ||
+          isExecutionPointsWithinRange(
+            loggableExecutionPoint,
+            focusRange.begin.point,
+            focusRange.end.point
+          );
```
bvaughn, line 111: Why would we only need to do this for log points?

If there is an issue where we are showing things outside of the focused region, wouldn't that apply to _all types of messages_ (in the same way as the `isLoaded` check above does) ?

bvaughn: > Why would we only need to do this for log points?

Answer: It looks like other log types are being pre-filtered by the `LoggablesContextRoot`, e.g.
```js
// Trim eventLogs and logPoints by focusRange.
// Messages will have already been filtered from the backend.
const focusedEventLogs = useMemo<EventLog[]>(() => {
  if (focusRange === null) {
    return eventLogs;
  } else {
    return eventLogs.filter(eventLog =>
      isExecutionPointsWithinRange(eventLog.point, focusRange.begin.point, focusRange.end.point)
    );
  }
}, [eventLogs, focusRange]);
```

This is also expected to occur for the log points:
```
      const [hitPoints, status] = getHitPointsForLocation(
        client,
        point.location,
        point.condition,
        focusRange
      );
```

But for some reason that filtering doesn't seem to be working correctly for us in all scenarios? I think we should dig into that.

bvaughn: The code that's running here is:
```js
const [executionPoints, status] = await client.getHitPointsForLocation(
  focusRange,
  location,
  condition
);
```

So why is it giving us back analysis points that are outside of the focus region?

bvaughn: This is the problem:
https://github.com/replayio/devtools/blob/84892c90085993a053f80ba399ca7e8705fb01b9/packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts#L48-L54

This "key" string will just be "...:[Object]:[Object]" for any focus range.

PR #7691, opened 2022-09-07
Allow last paint when all paint points have loaded

jcmorrow:
I wrote this code back in https://github.com/replayio/devtools/commit/c7ce7d4d9b723c533e7a2c74391f27f3e4fefd38, but it turns out that `getGraphicsAtTime` will *also* bail if we are past the last paint, so it wasn't actually doing anything 😢


jasonLaster (approved):
Gosh - i'd love a replay for this

jcmorrow:
> Gosh - i'd love a replay for this

Here ya go https://app.replay.io/recording/repaint-at-the-end-without-pause-repainting--f7fea0ff-31b6-4f42-a7b2-08bb99196e3d

PR #7696, opened 2022-09-08
Show SourcemapToggle and SourcemapVisualizerLink for the current position in a source (FE-444)

hbenl:
Previously we could only show the toggle and link if we were either paused in the source (in which case we got the alternate source from the selected frame's `MappedLocation`) or if there was a unique alternate source. However, this didn't work if the source was a bundle.
This PR adds another strategy for this case: we try to find an alternate source by asking the backend for a `MappedLocation` for the first breakable position after the currently selected position in the source.


src/devtools/client/debugger/src/components/Editor/SourcemapToggle.tsx
```diff
@@ -1,73 +1,72 @@
...
-    const result = getUniqueAlternateSourceId(selectedSource, sourcesById);
-    alternateSourceId = result.sourceId;
-    if (!alternateSourceId && result.why === "not-unique") {
-      return null;
-    }
+export default function SourcemapToggle({ cursorPosition }: { cursorPosition: CursorPosition }) {
+  const selectedSource = useAppSelector(getSelectedSource);
+  const alternateSource = useAppSelector(getAlternateSource);
+  const sourcesById = useAppSelector(getSourceDetailsEntities);
+  const { result: alternateSourceIdResult } = useAsync(getAlternateSourceId, [
```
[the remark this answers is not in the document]
bvaughn: Agreed with the hesitation to add another external dependency.

Suspense could definitely be used for this! The biggest question would be: How many places would need to _suspend_ to fetch this data?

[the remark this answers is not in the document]
hbenl: I don't know how to do this with Suspense but if you can show me, that'd be awesome.

hbenl:
@bvaughn @markerikson This PR now uses suspense caches instead of react-async-hook

packages/bvaughn-architecture-demo/src/suspense/GenericCache.ts
```diff
@@ -0,0 +1,54 @@
...
+    wakeable: Wakeable<TValue>,
+    ...args: TParams
+  ) {
+    try {
+      const value = await fetchValue(...args);
+
+      record.status = STATUS_RESOLVED;
+      record.value = value;
+
+      wakeable.resolve(record.value);
```
bvaughn, line 21: Super tiny nit: You have an unnecessary property access here (`record.value`) since `value` is in scope right above.

packages/bvaughn-architecture-demo/src/suspense/GenericCache.ts
```diff
@@ -0,0 +1,54 @@
+import { Record, STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Wakeable } from "./types";
+import { createWakeable } from "../utils/suspense";
+
+export function createGenericCache<TParams extends Array<any>, TValue>(
```
bvaughn, line 4: This is clever!

Kind of a nit, but can we rename this file to `createGenericCache.ts` since it's a cache-factory and not actually a cache itself?

packages/bvaughn-architecture-demo/src/suspense/MappedLocationCache.ts
```diff
@@ -0,0 +1,14 @@
...
+import { ReplayClientInterface } from "shared/client/types";
+import { createGenericCache } from "./GenericCache";
+
+export const { getValue: getMappedLocation } = createGenericCache<
+  [ReplayClientInterface, ProtocolLocation],
+  ProtocolMappedLocation
+>(
+  (client, location) => client.getMappedLocation(location),
+  (client, location) => `${location.sourceId}:${location.line}:${location.column}`
+);
```
bvaughn, line 14: This is really nice and succinct. I like it.

packages/bvaughn-architecture-demo/src/suspense/GenericCache.ts
```diff
@@ -0,0 +1,54 @@
...
+        fetchAndStoreValue(record, record.value, ...args);
+      }
+
+      if (record.status === STATUS_RESOLVED) {
+        return record.value;
+      } else {
+        throw record.value;
+      }
+    },
+  };
```
bvaughn, line 53: Maybe we could simplify here and just have this return the `getValue` function? Unless you think there are other things we will be returning soon.

packages/bvaughn-architecture-demo/src/utils/testing.tsx
```diff
@@ -201,6 +201,8 @@ export function createMockReplayClient() {
       contentType: "text/javascript",
     })),
     getSourceHitCounts: jest.fn().mockImplementation(async () => new Map()),
+    getBreakpointPositions: jest.fn().mockImplementation(async () => []),
```
bvaughn, line 204: Super tiny nit: Small preference to preserve alpha-formatting in places like this because it makes merge conflicts easier to manage.

packages/shared/client/types.ts
```diff
@@ -107,6 +109,11 @@ export interface ReplayClientInterface {
   getSessionId(): SessionId | null;
   getSourceContents(sourceId: SourceId): Promise<{ contents: string; contentType: ContentType }>;
   getSourceHitCounts(sourceId: SourceId): Promise<Map<number, LineHits>>;
+  getBreakpointPositions(
+    sourceId: SourceId,
+    range?: { start: SourceLocation; end: SourceLocation }
+  ): Promise<SameLineSourceLocations[]>;
+  getMappedLocation(location: Location): Promise<MappedLocation>;
```
bvaughn, line 116: Same small preference about alpha-sorting.

bvaughn (approved):
I didn't smoke test this, and I don't really have enough context to understand what's going on in `src/devtools/client/debugger/src/utils/sourceVisualizations` without spending a lot of time digging in– but the UI parts and the Suspense parts of this PR look really nice! 👍🏼 

Asked for a couple of minor changes but nothing major 😁

packages/bvaughn-architecture-demo/src/suspense/GenericCache.ts
hbenl: I was thinking we may add `getValueAsync(...args: TParams): Promise<TValue>` and/or `getValueIfAlreadyLoaded(...args: TParams): TValue | undefined`.

bvaughn:
I _think_ this PR may have introduced some calls to methods that might _suspend_ outside of render.

For example: `getSourcemapVisualizerURL` calls `getSourceToVisualize` which calls `getAlternateSourceIdForPosition` which calls `getBreakpointPositions` which calls `_getBreakpointPositions` which _suspends_.

What calls `getSourcemapVisualizerURL` then?
* ✅ `SourcemapVisualizerLink` calls it inside of a `Suspense` boundary.
* ❌ The `EditorMenu` component calls `editorMenuItems` which calls `sourceMapItem` (inside of the `devtools/client/debugger/src/components/Editor/menus/editor` module) which calls it. This might cause the whole `Editor` to _suspend_, and that doesn't seem to be wrapped in a `Suspense` boundary.

It also seems like the call to `Debugger.getPossibleBreakpoints` sometimes throws an "Internal error" (which may be unexpected, see BAC-2207) that doesn't get handled by either of the call sites above, which means it can cause the whole app to get unmounted.

If that error is expected (probably even if it's not) we should wrap error boundaries around these two places to ensure this crash doesn't happen.

hbenl:
`getSourceToVisualize` will only suspend when its last 2 arguments are truthy, which is not the case when it's called by the `EditorMenu`component.
But this should be more obvious: I will append `Suspense` to all names of functions that can suspend and split this function into one that won't suspend (and doesn't expect the last 2 arguments) and one that can (with `Suspense` in its name).

PR #7697, opened 2022-09-08
Fix bad focus range serialization in PointsCache that was causing bad analysis results

bvaughn:
Follow up to #7690


packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
```diff
@@ -51,7 +51,9 @@ export function getHitPointsForLocation(
   // without any overflow, and then later fetch for a focus range. Right now we re-fetch in this case.
 
   const locationKey = `${location.sourceId}:${location.line}:${location.column}:${condition}`;
-  const key = focusRange ? `${locationKey}:${focusRange.begin}:${focusRange.end}` : locationKey;
+  const key = focusRange
+    ? `${locationKey}:${focusRange.begin.point}:${focusRange.end.point}`
+    : locationKey;
```
bvaughn, line 56: This was causing the problem. We were incorrectly reusing a cached set of hit points for the previous focus range because I thought this was a `PointRange` rather than a `TimeStampedPointRange` 😞

markerikson (approved):

PR #7698, opened 2022-09-08
Always create a new Pause in ThreadFront.instantiatePause()

hbenl:
Currently we try to not create multiple pauses for the same point, but I realized that this is unsafe if we receive a pauseId from the backend for a console message or a logpoint result: The message or logpoint result may contain objectIds from that pause, but there can be no guarantee that an objectId will reference the same object in a different pause for the same point.
This also fixes FE-679, where not creating a new pause for a console message also means that the sourceIds in its frames are not updated to the first corresponding sourceId.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -459,11 +459,7 @@ class _ThreadFront {
     hasFrames: boolean,
     data: PauseData = {}
   ) {
-    let pause = Pause.getByPoint(point);
-    if (pause) {
-      return pause;
-    }
-    pause = new Pause(this);
+    const pause = new Pause(this);
     pause.instantiate(pauseId, point, time, hasFrames, data);
     return pause;
   }


markerikson:
This may be worth discussing further with Brian, specifically since we _are_ trying to avoid having multiple Pause instances per execution point.

bvaughn:
> Currently we try to not create multiple pauses for the same point, but I realized that this is unsafe if we receive a pauseId from the backend for a console message or a logpoint result: The message or logpoint result may contain objectIds from that pause, but **there can be no guarantee that an objectId will reference the same object in a different pause for the same point.**

This is exactly why we try to avoid creating more than one Pause per point.

Seems like this PR would invalidate _most_ of the frontend object preview caching?

bvaughn (changes requested):
Preferably, let's _not_ do this, at least not without talking more.

bvaughn:
> > This is exactly why we try to avoid creating more than one Pause per point.
>
> Huh? But the way it works now means that we may get an objectId from a logpoint result and then try to fetch the object preview using a different pause that was created earlier for the same point - and that may give us the preview of a different object.

We should fix _that_ though, shouldn't we? (We shouldn't ask for the preview data _at a new pause_.)

> > Seems like this PR would invalidate most of the frontend object preview caching?
>
> It would mean that we'll have to cache object previews once for each pause, even if the pauses are for the same point. But I don't see how we can avoid that.

The point is that we're _pre-caching_ a lot of data now for a given point+pause pair. We _have_ to cache object previews at the pause+point level (not just the point level) because of the unstable object ids as you mention. So if we change to always creating a new point, we'll _always_ invalidate any data we may have already cached for that point.

markerikson:
After _much_ discussion, some takeaways:

- This is driven by trying to make sure that `Location`s in frame data point to the correctly-updated source IDs
- we mutate those `Location`s when the frames are received and processed by the `Pause` instance
- but there are apparently times we do end up with multiple `Pause` instances for a given `ExecutionPoint` and we want to make sure the duplicated frame objects also get processed to have correct locations

We need to do more investigation and thinking about this, but that should wait until we've deleted the old console (hopefully Monday?), and can see exactly what code is still remaining.

It's possible that there's few enough spots that _use_ the frame data and locations that we could actually deal with the "corresponding source" lookups at the UI layer instead (ie, <10 places, rather than "hundreds").

hbenl:
Closing this since we're removing `Pause` anyway.

PR #7700, opened 2022-09-08
Re-enable new console as default again

markerikson:
Re-enables the new console as default now that we've landed several fixes.

Putting this up now so it's here when we're ready to turn this back on.

See #7674 for the commit that turned it off.


bvaughn (approved):

bvaughn:
Any reason not to land this?

markerikson:
Looks like this is obsolete now :)

PR #7702, opened 2022-09-08
Revert "Revert PR 7662"

souporserious:
This PR is for the original changes in #7662 which were reverted due to the console tests still pointing to old code.

Reverts replayio/devtools#7699


bvaughn (approved):
Looks like CI is happy 👍🏼 Thanks!

PR #7703, opened 2022-09-08
Disable console eval if the user isn't a "Developer" role

markerikson:
This PR:

- Reworks the `<JSTerm>` logic to disable the evaluation input and show a "Disabled...." message if the user does not have the "Developer" role, instead of letting them type something in and then show a "nope you can't do that!" message

image


bvaughn (approved):
Seems like a more user-friendly approach.

PR #7704, opened 2022-09-08
Fixed an overflow/scrolling issue in the preview popup

bvaughn:


src/devtools/client/debugger/src/components/Editor/Preview/NewObjectInspector.module.css
```diff
@@ -2,10 +2,12 @@
   color: var(--color-default);
   display: flex;
   flex-direction: column;
+  flex-wrap: wrap;
+  max-height: 250px !important;
```
bvaughn, line 6: This sucks, but the `.popup .preview-popup` style was overriding this because of having a greater specificity and so...for now, until we rip out the old Console, this is fine.

PR #7707, opened 2022-09-09
Update ESlint rules

bvaughn:
A couple of recent bug fixes (#7701, #7697) could have been better detected by TypeScript if we leveraged some of the TypeScript-specific ESLint rules. This PR enables them (and verifies that they would have caught the original bugs)!

Bug from #7697
Screen Shot 2022-09-09 at 7 52 30 AM
ESLint error message:
Screen Shot 2022-09-09 at 7 51 25 AM

Bug from #7701
Screen Shot 2022-09-09 at 7 51 56 AM
ESLint error message:
Screen Shot 2022-09-09 at 7 52 12 AM

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,10 +15,13 @@
       "jsx": true
     },
     "ecmaVersion": 2018,
+    "project": ["tsconfig.json"],
     "sourceType": "module"
   },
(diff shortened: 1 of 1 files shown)


bvaughn:
I think this is a pretty uncontroversial change, so I'm going to just go ahead and land it 😄 Will revert if anyone objects.

PR #7710, opened 2022-09-09
Temporarily disable repaint for Chromium

bvaughn:
I _think_ this is what FE-673 was calling for, but @jasonLaster can clarify 😄


jasonLaster (approved):
Looks good to me

PR #7713, opened 2022-09-11
Disable Network and Event Types for Chromium recordings

bvaughn:
Related to FE-673, FE-709, and FE-710.

I think that ideally, this type of capabilities metadata would be embedded into a recording by the runtime and passed along to the frontend via the protocol, but in lieu of that this PR adds the concept to the protocol package. Maybe we can build on top of this pattern for other runtime-specific capabilities so that we at least have only a single place to update these things as capabilities change?

Here is a screenshot of a Chromium recording with no Network tab or event types in the Console side filter:
Screen Shot 2022-09-11 at 11 03 55 AM

Note that I didn't update the old Console to respect the event types capabilities because it will soon be deleted (via #7641 or #7643).


markerikson:
I like this! Much cleaner checks than "`node` here, `chromium` there"

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -70,6 +74,7 @@ const PanelButtons: FC<PanelButtonsProps> = ({
   hasReduxAnnotations,
   toolboxLayout,
   hasGraphics,
```
jasonLaster, line 76: I bet we can also get rid of `hasGraphics`

jasonLaster (approved):
Looks great. This is a much better way to do it

src/ui/components/SecondaryToolbox/index.tsx
bvaughn: Good call.

PR #7720, opened 2022-09-12
Added an 'add team member' button to the library

jonbell-lot23:
Addresses this Linear ticket: https://linear.app/replay/issue/GTM-410/add-new-members-button-to-library-view
Here's a Loom: https://www.loom.com/share/68aee683d426415681962b7f8acab5fc


ryanjduffy:
Having trouble making a replay on my phone's hotspot but couple notes:

* The button shows up on the user's library so we should hide it there. Side effect of this is clicking it does nothing until you select a team and then the modal opens.
* There were a few times in which clicking the button showed the billing page instead of the team member page. I think we need to set the modal options as well to ensure we select the right tab

https://github.com/replayio/devtools/blob/main/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx#L232-L236

ryanjduffy (approved):

PR #7721, opened 2022-09-12
Adjust loading number based on paints seen

jcmorrow:
This is kind of a crappy solution to the problem discussed in https://linear.app/replay/issue/DES-82/sometimes-loading-appears-stuck-at-0percent

But... also maybe it's better than what we have right now?

https://user-images.githubusercontent.com/5903784/189777158-3f11e7dc-d0fd-4853-80c6-79a56b539c7a.mp4

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -163,6 +163,8 @@ export function createSocket(
       }
 
       const experimentalSettings: ExperimentalSettings = {
+        // Uncomment this to get a new session every time you refresh
+        // controllerKey: String(Math.floor(Math.random() * 10e8)),
         listenForMetrics: !!prefs.listenForMetrics,
         disableCache: !!prefs.disableCache || !!features.profileWorkerThreads,
         disableQueryCache: !features.enableQueryCache,
--- a/src/ui/reducers/timeline.ts
+++ b/src/ui/reducers/timeline.ts
@@ -2,11 +2,17 @@ import { TimeStampedPoint } from "@replayio/protocol";
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { UIState } from "ui/state";
 import { FocusRegion, HoveredItem, TimelineState, ZoomRegion } from "ui/state/timeline";
-import { displayedEndForFocusRegion, displayedBeginForFocusRegion } from "ui/utils/timeline";
+import {
+  displayedEndForFocusRegion,
+  displayedBeginForFocusRegion,
+  mergeSortedPointLists,
+} from "ui/utils/timeline";
 import sortBy from "lodash/sortBy";
+import maxBy from "lodash/maxBy";
 
 function initialTimelineState(): TimelineState {
   return {
+    allPaintsReceived: false,
     currentTime: 0,
     focusRegion: null,
     focusRegionBackup: null,
@@ -14,6 +20,7 @@ function initialTimelineState(): TimelineState {
     hoveredItem: null,
     playback: null,
     playbackPrecachedTime: 0,
+    paints: [{ time: 0, point: "0" }],
     points: [{ time: 0, point: "0" }],
     recordingDuration: null,
     shouldAnimate: true,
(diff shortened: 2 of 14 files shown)


jonbell-lot23:
Copied from Linear:

_So the worst case scenario is:

Nothing really populates on the screen

But in the corner you do see our turtle + percentage indicator, which should count up

It's possible this could be better, but I'd like to test more. Can we push live via some sort of experimental flag, or special querystring? I want to practice on a bunch of Replays to get a sense for it before it's fully live._

jcmorrow:
OK, this branch now has an experimental flag, and so could go out whenever without affecting people.

jcmorrow:
@bvaughn (for tomorrow) let me know how you feel about this. I kind of hate that it's not coordinated with #7732, but I'm also not sure exactly how much work it would take to get it there 🤔

src/ui/actions/session.ts
```diff
@@ -163,6 +163,8 @@ export function createSocket(
       }
 
       const experimentalSettings: ExperimentalSettings = {
+        // Uncomment this to get a new session every time you refresh
+        // controllerKey: String(Math.floor(Math.random() * 10e8)),
```
bvaughn, line 167: I wonder if it would be worth adding a URL flag that trips this? e.g. `?forceNewSession` or something?

src/ui/reducers/timeline.ts
```diff
@@ -96,6 +94,19 @@ export const getUnprocessedRegions = (state: UIState) => state.timeline.unproces
...
+  if (state.timeline.allPaintsReceived) {
+    return 1.0;
+  }
+  const maxPaint = maxBy(state.timeline.paints, p => BigInt(p.point));
+  const maxPoint = maxBy(state.timeline.points, p => BigInt(p.point));
+  if (!maxPoint || maxPoint.point === "0") {
+    return 0.0;
+  }
+  return (1.0 * (maxPaint?.time || 0)) / maxPoint.time;
+};
```
bvaughn, line 109: Two thoughts:

1. This selector feels pretty heavy, given that it calls `maxBy` twice each time it's run– which loops over the entire array of paints and points. Fortunately I don't think we need to call `maxBy` since those arrays are sorted. (Can't we just look at the last element in the array?)

2. Should `getBasicProcessingProgress` be taking paints into account even if `basicProcessingLoadingBar` is disabled? (Should it not fall back to the prior behavior of comparing the latest time in the known regions?)

src/ui/utils/prefs.ts
```diff
@@ -21,6 +21,8 @@ pref("devtools.consoleFilterDrawerExpanded", true);
 pref("devtools.hitCounts", "hide-counts");
 
 // app features
+pref("devtools.features.basicProcessingLoadingBar", false);
+pref("devtools.features.enableQueryCache", false);
```
bvaughn, line 25: Where did "enableQueryCache" come from? 🤨 Was it missing before? It doesn't seem related to this PR.

bvaughn (approved):

src/ui/utils/prefs.ts
jcmorrow: Huh. It was (and still is) also on line 30 🤔  Good catch!

src/ui/reducers/timeline.ts
jcmorrow: 1. Yes. Good point, duh! I will make that change.
2. I think either: 1) `getBasicProcessingProgress` is off, in which case this is never called, or else it is on, in which case we need paints to make any sort of reasonable estimate about how far basic processing has gotten.

src/ui/actions/session.ts
jcmorrow: Yeah, I have also toyed with this, but never *quite* found the right way to do it, the URL flag seems like a reasonable way to do it, though I might leave it off of this PR just because it's big enough as is, especially in combination with #7732

bvaughn:
Seems like this was merged with a TSC error?
```
src/ui/setup/dynamic/devtools.ts(314,5): error TS2322: Type '(...args: any[]) => FocusRegion | null | undefined' is not assignable to type 'SelectorWithoutStateArg<ActionCreatorWithoutPayload<string>> & SelectorWithoutStateArg<(state: UIState) => boolean> & ... 204 more ... & SelectorWithoutStateArg<...>'.
  Type '(...args: any[]) => FocusRegion | null | undefined' is not assignable to type 'SelectorWithoutStateArg<ActionCreatorWithoutPayload<string>>'.
    Type 'FocusRegion | null | undefined' is not assignable to type '{ payload: undefined; type: string; }'.
      Type 'undefined' is not assignable to type '{ payload: undefined; type: string; }'.
src/ui/setup/dynamic/devtools.ts(31[5](https://github.com/replayio/devtools/actions/runs/3065125050/jobs/4948906217#step:6:6),42): error TS255[6](https://github.com/replayio/devtools/actions/runs/3065125050/jobs/4948906217#step:6:7): A spread argument must either have a tuple type or be passed to a rest parameter.
Error: Process completed with exit code 2.
```

src/ui/utils/timeline.ts
```diff
@@ -309,3 +310,48 @@ export function filterToFocusRegion<T extends TimeStampedPoint>(
...
+
+function assertSorted(a: TimeStampedPoint[]) {
+  a.reduce(
+    (prev, curr) => {
+      assert(prev.time <= curr.time);
+      return curr;
+    },
+    { point: "0", time: 0 }
+  );
+}
```
bvaughn, line 322: I'm seeing this assertion throw errors for recording 436b7cf5-50f1-4e69-894d-a61343e128c2 but it's not really clear how to use that information.

bvaughn: I _think_ this assert actually uncovered a bad recording (see BAC-2339).

I'm going to change it for now though to (1) only run in DEV and (2) compare points rather than time (since those are more fine-grained anyway). Let's chat if you think this is a bad move!

src/ui/utils/timeline.test.ts
```diff
@@ -231,3 +232,32 @@ describe("filterToFocusRegion", () => {
     ]);
   });
 });
+
+describe("mergeSortedPointLists", () => {
+  fit("will correctly interleave overlapping ranges", () => {
```
bvaughn, line 237: We should probs add a line rule to prevent `fit` or `test.only` from being committed accidentally. I do this sometimes too.

PR #7722, opened 2022-09-13
Remove unnecessary async from Mixpanel functions

bvaughn:
We recently enabled the `@typescript-eslint/no-floating-promises` lint rule (#7714) which flagged a bunch of false-positives with `trackEvent` calls. Events are fire-and-forget, so we could just ignore those lint rule violations– except that the Mixpanel functions were unnecessarily flagged as async in the first place. This PR fixes that.


markerikson:
Beat me to it!

PR #7724, opened 2022-09-13
Rewrite React DevTools evaluations to not use `NodeFront/ValueFront`

markerikson:
This PR:

- Rewrites all of the `ThreadFront.evaluate()` calls in `ReactDevTools.tsx` to call `ThreadFront.evaluateNew()` instead.  This method does not return `ValueFront` instances in the data.
- Ports `ValueFront.getJSON()`, which converts a nested protocol data description into the equivalent POJO data, to be a standalone `getJSON` util that does not need `ValueFront`s

### DIFF
--- a/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
@@ -45,7 +45,8 @@ class ReplayWall implements Wall {
     private onShutdown: () => void,
     private highlightNode: (nodeId: string) => void,
     private unhighlightNode: () => void,
-    private replayClient: ReplayClientInterface
+    private replayClient: ReplayClientInterface,
+    private pauseId: string | undefined
   ) {}
 
   // called by the frontend to register a listener for receiving backend messages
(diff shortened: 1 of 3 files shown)


src/ui/utils/objectFetching.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+    case "null": {
+      return null;
+    }
+    case "undefined": {
+      return undefined;
+    }
+  }
+
+  throw new Error("Unexpected client value! " + JSON.stringify(clientObject));
```
markerikson, line 108: :speech_balloon: Left this in here for debugging purposes atm. I should probably turn this into a `console.error()` or something.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -260,20 +280,43 @@ function createReactDevTools(
...
-  // We should only load the DevTools module once we know which protocol version it requires.
-  // If we don't have a version yet, it probably means we're too early in the Replay session.
-  if (version >= 2) {
-    stateUpdaterCallback(await import("react-devtools-inline/frontend"));
-  } else if (version === 1) {
-    stateUpdaterCallback(await import("react-devtools-inline_4_17_0/frontend"));
+  if (response.returned?.object) {
+    // Unwrap the nested eval objects by asking the backend for contents
+    // of the nested fields: `{data: {version: 123}}`
+    const responseDataObject = await getObjectPropertyHelper(
```
markerikson, line 299: :speech_balloon: Now that I've got `getJSON()`, I _could_ change this to be one `getJSON()` call instead and access `obj.data.version`. I may do that as a follow-up.

bvaughn (approved):
I think this looks solid 👍🏼 

Might be nice to have some unit tests for `getJSON` at least, if they're feasible to write.

markerikson:
Yeah, lemme think about the testing bit once I've got these fronts removed

PR #7725, opened 2022-09-13
Optimize PointsCache

bvaughn:
Cache uses newer protocol API `Session.getPointsBoundingTime` which returns an array of the nearest points before and after a specific time. This enables us to do in-memory matching for times more efficiently, reducing the overall number of protocol requests we make.

This change added complexity to the cache, so I also added unit tests.


packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts
```diff
@@ -12,89 +13,84 @@ import { isExecutionPointsLessThan } from "../utils/time";
...
-    record.value = error;
+    if (rethrowError) {
+      throw error;
+    } else {
+      // TODO
+      // Note that we might be caching unloaded region errors by doing this.
+      // That's better than flooding the backend with repeated requests,
+      // but we also probably don't want to permanently cache this type of error.
+      // Maybe we should somehow make loaded regions part of the cache key?
+      timeToErrorMap.set(time, error);
```
bvaughn, line 88: TBH I'm not sure what we should do here. The frontend could guard against asking for a point/time outside of the loaded range, so this should never _actually_ happen, but if we miss that guard...it's possible.

markerikson (approved):

PR #7733, opened 2022-09-14
Clarified types for NewObjectInspector wrapper

bvaughn:


markerikson (approved):

markerikson:
My PR at #7726 actually deletes the editor file chunk of conversion entirely, so you can probably just drop that part here

bvaughn:
> My PR at https://github.com/replayio/devtools/pull/7726 actually deletes the editor file chunk of conversion entirely, so you can probably just drop that part here

Oh. Damn it. Sorry.

At least it should be an easy merge resolve. 😅

PR #7735, opened 2022-09-14
Merge getter values and properties when rendering object previews

bvaughn:
Addresses FE-718 and FE-725


jasonLaster (approved):

bvaughn:
Not sure why Playwright tests have started flaking for us :( Same test as failed on Mark's recent PR. In both cases, the tests pass locally for me on these branches.

PR #7737, opened 2022-09-14
Optimize CodeMirror hit counts gutter performance

markerikson:
Made multiple changes to how we update hit counts in the CodeMirror gutter:

- We precalculate `gutterWidth`, and only call `CM.setOptions()` to recreate the gutters whenever that _character_-based width changes
- `drawLines` now bails out early in many cases, and also skips updating lines that we've seen already with this hit count map
- We only create marker DOM nodes once per line
- We no longer dispatch `setHoveredLine(null)` _every_ time the mouse leaves a line. Instead, we call a debounced function, and cancel it in the "mouseenter" handler, since the most common behavior is to mouse from line X to X + 1.
- We also don't dispatch if the hovered line number is null, which is the case when the existing line is inactive / no hits
- We _do_ now refetch hit counts when the focus region changes, so that the hit counts update as soon as you close the focus bar without having to wait until you mouse over the editor again.

Also I tweaked the TS types to better refer to which `editor/sourceEditor` variables/types we're using here.

This is still _slightly_ draft-y. What's here should be a big improvement, but I want to do a couple more perf captures from this preview build to see how it behaves in comparison.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Gutter.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Gutter.tsx
@@ -1,11 +1,12 @@
 import LineHitCounts from "./LineHitCounts";
 import ToggleWidgetButton from "./ToggleWidgetButton";
+import type { SourceEditor } from "../../utils/editor/source-editor";
 
-export default function Gutter({ editor }: any) {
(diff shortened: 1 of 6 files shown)


jcmorrow (approved):
Some of this looks a little gnarly, but the preview branch scrolls so smooth, I say we ship it 🎨

PR #7738, opened 2022-09-14
Switch disable repainting boolean flag

jcmorrow:
Adding this concept was a really good idea, it will make adding checks like this a lot easier 👍 

Fixes https://linear.app/replay/issue/FE-734/do-we-still-request-pause-graphics-for-node-recordings

### DIFF
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -248,7 +248,7 @@ export function setupGraphics() {
 
 export async function repaint(force = false) {
   const recordingCapabilities = await ThreadFront.getRecordingCapabilities();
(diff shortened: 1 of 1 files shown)


markerikson (approved):

packages/protocol/graphics.ts
```diff
@@ -248,7 +248,7 @@ export function setupGraphics() {
 
 export async function repaint(force = false) {
   const recordingCapabilities = await ThreadFront.getRecordingCapabilities();
-  if (recordingCapabilities.supportsRepaintingGraphics) {
+  if (!recordingCapabilities.supportsRepaintingGraphics) {
     return;
```
bvaughn, line 252: Oof 😅

PR #7741, opened 2022-09-15
Add discord, email, and documentation links to error messages

jonbell-lot23:
Addresses [DES-33](https://linear.app/replay/issue/DES-33/add-a-forward-path-issues-discord-support-to-our-error-screens)

We're adding three links underneath our error messages that can help things feel less like a dead-end.

image

image

image

### DIFF
--- a/src/ui/components/shared/Dialog.tsx
+++ b/src/ui/components/shared/Dialog.tsx
@@ -6,14 +6,31 @@ export type DialogPropTypes = HTMLProps<HTMLDivElement>;
 
 export function Dialog({ children, className, ...props }: DialogPropTypes) {
   return (
+    <>
     <div
       {...props}
       className={classNames("dialog flex flex-col items-center", className)}
       role="dialog"
       style={{ animation: "linearFadeIn ease 200ms", width: 400 }}
     >
-      {children}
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/Dialog.tsx
```diff
@@ -6,14 +6,31 @@ export type DialogPropTypes = HTMLProps<HTMLDivElement>;
...
       role="dialog"
       style={{ animation: "linearFadeIn ease 200ms", width: 400 }}
     >
-      {children}
+      {children}      
+    </div>
+    <div>
+      <div className="grid grid-cols-1 gap-1 pt-4 text-xs text-gray-400 place-items-center">            
+            
+            <div>
```
jcmorrow, line 21: I think the linter is mad because of the empty line and the fact that this block is indented one extra step than it should be.

jcmorrow (approved):
😍 I love this. Do make sure the linter is happy before merging, but otherwise, looks great! (I am 99% certain those test failures have nothing to do with this, so feel free to ignore).

jonbell-lot23:
Fixed the code and ran the linter, so I think I'm good!

PR #7745, opened 2022-09-15
Implement Frames and Scopes panels using suspense caches

hbenl:


packages/protocol/thread/pause.ts
```diff
@@ -242,11 +248,20 @@ export class Pause {
       if (!this.frames.has(f.frameId)) {
         this.frames.set(f.frameId, f as WiredFrame);
       }
+      if (!this.rawFrames.has(f.frameId)) {
+        const rawFrame = cloneDeep(f);
```
markerikson, line 252: :speech_balloon_ I _hate_ this :(  fortunately it's only temporary

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -0,0 +1,163 @@
...
+            state,
+            ThreadFront.preferredGeneratedSources,
+            protocolFrame,
+            index,
+            asyncIndex
+          )
+        ) || null,
+        sources
+      ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
markerikson, line 85: :question: what dep are we avoiding a warning for here?

src/devtools/client/debugger/src/components/SecondaryPanes/NewScopes.tsx
```diff
@@ -0,0 +1,112 @@
...
+  //TODO title for Window scope
+  //TODO isCurrentTimeInLoadedRegion
+  //TODO isErrored
+  const pauseId = useAppSelector(getPauseId);
+  const frameId = useAppSelector(getSelectedFrameId);
+  if (!pauseId || !frameId) {
+    return null; //TODO?
+  }
+
+  const splitFrameId = frameId.split(":");
```
markerikson, line 28: :speech_balloon: purely stylistically, I'd lean towards `const [groupIndex, frameIndex] = frameId.split(":")` (would still need to convert to a number)

markerikson (commented):
Look promising so far!

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
```diff
@@ -1,12 +1,38 @@
+import { handleError } from "protocol/utils";
 import { Record, STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Wakeable } from "./types";
 import { createWakeable } from "../utils/suspense";
 
+export interface GenericCache<TParams extends Array<any>, TValue> {
+  getValueSuspense(...args: TParams): TValue;
+  getValueAsync(...args: TParams): Promise<TValue>;
+  getValueIfCached(...args: TParams): TValue | undefined;
```
bvaughn, line 8: nit: I think methods like this should return `null` for uncached objects. It's more explicit. (`undefined` can indicate an unhandled case since many collection types return that value for missing keys).

packages/bvaughn-architecture-demo/src/suspense/types.ts
```diff
@@ -23,7 +23,7 @@ export type Record<T> = PendingRecord<T> | ResolvedRecord<T> | RejectedRecord;
 // You can use a Promise for this, but Promises have a downside (the microtask queue).
 // You can also create your own "thennable" if you want to support synchronous resolution/rejection.
 export interface Thennable<T> {
-  then(onFulfill: (value: T) => any, onReject: () => any): void | Thennable<T>;
+  then(onFulfill: (value: T) => any, onReject?: (err: any) => any): void | Thennable<T>;
```
bvaughn, line 26: Why did you make `onReject` optional? Can we keep it required? (**Edit**: Is it because of the change to the `isThennable` method?)

Also small nit: Abbreviating `error` doesn't buy us anything 😁  (but admittedly I dislike arbitrary abbreviations in code)

packages/protocol/thread/pause.ts
bvaughn, line 252: Is this cloning step necessary because of the fronts and mutation?

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
bvaughn: This feels like a foot gun. If we use it to look at sources (or anything) then it's a dependency, isn't it?

Can we pre-extract sources and use _them_ as the dependency then?

Generally, bypassing the exhaustive deps rule seems like a bad idea to me.

src/ui/suspense/util.ts
```diff
@@ -0,0 +1,54 @@
...
+  const asyncParentPauseId = getAsyncParentPauseIdSuspense(pauseId, asyncIndex);
+  if (!asyncParentPauseId) {
+    return;
+  }
+  const frames = getFramesSuspense(asyncParentPauseId);
+  if (!frames) {
+    return;
+  }
+  return frames.slice(1);
+}
```
bvaughn, line 54: I'm curious about this "util"

The API looks like a Suspense cache, but the way it's written isn't like one. What's the motivation?

src/ui/suspense/frameCache.ts
```diff
@@ -0,0 +1,18 @@
...
+  getValueIfCached: getFramesIfCached,
+} = createGenericCache<[PauseId], Frame[] | undefined>(
+  async pauseId => {
+    const pause = Pause.getById(pauseId);
+    assert(pause, `no pause for ${pauseId}`);
+    const wiredFrames = await pause.getFrames();
+    return wiredFrames?.map(f => pause.rawFrames.get(f.frameId)!);
+  },
+  pauseId => pauseId
+);
```
bvaughn, line 18: I'm curious about these caches. Is there a reason they're in the main parent directory instead of in `@bvaughn/src/suspense` where we could add some test coverage with the new type of tests? Is it to avoid adding `ReplayClientInterface` methods for them or something else?

It's not a problem. I'm just curious.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const selectedFrame = getSelectedFrame(state);
+  const dispatch = useAppDispatch();
+
+  function toggleFrameworkGrouping() {
+    dispatch(setFrameworkGroupingAction(!frameworkGroupingOn));
+  }
+  function selectFrame(...args: Parameters<typeof selectFrameAction>) {
+    dispatch(selectFrameAction(...args));
+  }
+  //TODO copy async parent frames as well
```
bvaughn, line 48: Do we have a Linear issue filed for this follow up? (I'm not sure what it entails TBH.)

bvaughn (commented):
Mix of questions mostly 😁

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
hbenl: Personally I prefer to avoid `null` whenever possible because I don't like the code being littered with `null | undefined`. Plus this is essentially a `Map` lookup so returning `undefined` seems consistent to me.

packages/bvaughn-architecture-demo/src/suspense/types.ts
hbenl: `Promise.then()` doesn't require it and a `Promise` should be a `Thennable`. And yes, I noticed it when I changed `isThennable()` because then typescript complained about `createRecorder.ts` where `Thennable.then()` is only passed one argument.

packages/protocol/thread/pause.ts
hbenl: Yes.

src/ui/suspense/util.ts
hbenl: > Can we pre-extract sources and use them as the dependency then?

That would require changing the sources reducers and all its callers. But I agree that this would be preferable.

> The API looks like a Suspense cache, but the way it's written isn't like one. What's the motivation?

It works with existing caches and doesn't do any expensive computations so I preferred to do it this way. I could also simply put that logic inside the render methods but this seemed good to extract.

src/ui/suspense/frameCache.ts
hbenl: I put them there because they reference `ThreadFront` (for now) and I didn't want to introduce that dependency to `packages/bvaughn-architecture-demo`.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
hbenl: Just created FE-750

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
bvaughn: Why would code be `null | undefined` ? A `null` return type is explicit. You only need to check if null or not.

src/ui/suspense/frameCache.ts
bvaughn: Right. We shouldn't ever depend on TF in the new package, but on `shared/client`.

packages/bvaughn-architecture-demo/src/suspense/MappedLocationCache.ts
```diff
@@ -5,10 +5,11 @@ import {
 import { ReplayClientInterface } from "shared/client/types";
 import { createGenericCache } from "./createGenericCache";
 
-export const { getValue: getMappedLocation } = createGenericCache<
-  [ReplayClientInterface, ProtocolLocation],
-  ProtocolMappedLocation
->(
+export const {
+  getValueSuspense: getMappedLocationSuspense,
```
markerikson, line 9: :speech_balloon: huh, this is kinda similar to RTK's `createEntityAdapter` returning several premade selectors. Neat!

packages/bvaughn-architecture-demo/src/suspense/MappedLocationCache.ts
```diff
@@ -5,10 +5,11 @@ import {
...
+  getValueAsync: getMappedLocationAsync,
+  getValueIfCached: getMappedLocationIfCached,
+} = createGenericCache<[ReplayClientInterface, ProtocolLocation], ProtocolMappedLocation>(
```
markerikson, line 12: :speech_balloon: I like the use of generic args here

packages/protocol/thread/pause.ts
markerikson: BUT HOPEFULLY NOT FOR VERY LONG! :)

packages/protocol/thread/pause.ts
```diff
@@ -242,11 +248,20 @@ export class Pause {
...
+        this.ThreadFront.updateMappedLocation(rawFrame.location);
+        this.ThreadFront.updateMappedLocation(rawFrame.functionLocation);
```
markerikson, line 254: :speech_balloon: and just filed FE-756 to cover removing the need for `updateMappedLocation` entirely

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
markerikson: Inclined to agree with Brian on principle here. It's worth rearranging function signatures so we don't have to do this and can instead just pass in the data that's needed.

markerikson (commented):
Seems generally reasonable to me - really just need to fix up the deps warning to not pass `state` through.

PR #7747, opened 2022-09-15
Fixed Keyboard Shortcuts issues

bvaughn:
This PR fixes a couple of issues I noticed while looking at our keyboard shortcuts code (as part of FE-730):
* Add "reverse step over" shortcut, which was missing before.
* Fix incorrect "step out" command for Mac OS; it should be `Cmd+Shift+;` not `Cmd+Shift+:`.
* Fix a logic bug in command parsing that incorrectly parsed strings like "CommandOrControl".
* Target `CommandBar` (aka breakpoint shortcuts) on the document rather than the command bar itself. (How often is someone going to _focus on_ the little command bar?)

@markerikson Could you give this a test drive on Windows?

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/CommandBar.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/CommandBar.tsx
@@ -26,25 +26,27 @@ const isMacOS = appinfo.OS === "Darwin";
 
 // NOTE: the "resume" command will call either the resume
 // depending on whether or not the debugger is paused or running
-const COMMANDS = ["resume", "stepOver", "stepIn", "stepOut"] as const;
+const COMMANDS = ["resume", "reverseStepOver", "stepOver", "stepIn", "stepOut"] as const;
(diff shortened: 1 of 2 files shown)


markerikson:
Sure. Btw, why _do_ we have F11 set up as a shortcut? I get that may be a default in editors, but browsers treat that as "Show Full Screen". Does intercepting that actually work?

bvaughn:
> Sure. Btw, why do we have F11 set up as a shortcut? I get that may be a default in editors, but browsers treat that as "Show Full Screen". Does intercepting that actually work?

F11 _doesn't work_ if the browser is already full screen, for sure, because JS shouldn't be able to "trap" a user in full-screen. I don't know if it works when you're not already full screen or not. It does not for me either (on my Mac) because that key shows the desktop. I didn't add that key though so 🤷🏼 The Mac-specific keys _do work_ at least.

bvaughn:
To put a finer point on it: I believe this PR fixes several things, but maybe there are still some things that aren't great 😁

markerikson:
Yeah, might as well merge this

PR #7749, opened 2022-09-15
↑/↓ keys suggest last commands in the new console

bvaughn:
https://www.loom.com/share/356c3c0eb8404398a284e911c2c94ba8

### DIFF
--- a/src/devtools/client/webconsole/components/Input/JSTerm.tsx
+++ b/src/devtools/client/webconsole/components/Input/JSTerm.tsx
@@ -86,10 +86,18 @@ export default function JSTerm({
     } else if (e.key === Keys.ARROW_UP && editor.getCursor().line === 0) {
       e.preventDefault();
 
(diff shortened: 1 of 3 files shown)


src/devtools/client/webconsole/components/Input/JSTerm.tsx
```diff
@@ -86,10 +86,18 @@ export default function JSTerm({
...
+      let newIndex = null;
+      let newValue = null;
+
+      if (historyIndex === null) {
+        newIndex = terminalExpressionHistory.length - 1;
+        newValue = terminalExpressionHistory[newIndex];
+      } else {
+        newIndex = historyIndex - 1;
+        newValue = terminalExpressionHistory[newIndex];
+      }
```
bvaughn, line 99: This was the gist of the regression. I wasn't properly handling `null` state (the default/initial value). Ooof.

markerikson (approved):

PR #7753, opened 2022-09-15
Rewrite hit counts rendering to use a virtualized list for scroll perf

markerikson:
This PR:

- Entirely replaces the existing hit counts rendering logic to use a `react-window` virtualized list instead of manually inserting and updating DOM nodes in CodeMirror's gutter
- Adds null checks to a couple of line mouse event handlers (which I think only had issues because I was changing zoom levels, but doesn't hurt to check)

Up until now, we've shown hit counts by adding a "gutter" in CodeMirror. This adds an extra column on the left side, next to the line numbers gutter.  We then watched for incoming hit counts, and used `doc.eachLine(lower, upper)` to iterate over lines and manually insert/update DOM nodes with the right classnames and hit count text content.

This led to laggy scroll perf, as the process of updating those DOM nodes would cause CodeMirror to block the main thread, in prod builds, for around 200ms:

image

I've completely rewritten the implementation.  Now, we still create that gutter (so that CodeMirror adjusts the line start location),  but we no longer insert/update DOM nodes manually.  Instead, we render a `react-window` virtualized list positioned absolutely on top of the CM gutter, pull list item heights from CodeMirror's own line metadata, and update the list scroll position to match any scroll events in the editor overall.

This improves perf in several ways:

- We no longer do as much work to ask CM to iterate over lines
- A chunk of the time in that perf screenshot was DOM reflow. Since the list is absolutely positioned, it shouldn't be affecting other layout as much
- Since the list is virtualized, there aren't nearly as many DOM nodes being added and updated
- Even though there's work done to re-render and keep the list in sync any time you scroll in the editor, it should be fairly minimal most of the time (especially with the generous amount of "overscroll" items rendered above and below)

I've confirmed this works in Chrome and FF on Windows, with both normal and "large" editor font text sizes.

The main issue I ran into was fractional list item sizes (like 16.36667px) causing drift as the file got very large.  The current logic appears to have fixed those.  I also had issues with print statements early in a file causing misalignment when you scrolled far past them, because `react-window` couldn't take those abnormal list item sizes into account.  I've hacked around this by manually forcing it to scroll to those lines synchronously in a `useLayoutEffect` whenever the list of print statements changes, render those rows and cache their sizes, then jump back to the actual scroll position.  This does actually work :)

The net result: I can smoothly scroll the entire 271K lines of `built/local/run.js` in Andarist's recording https://app.replay.io/recording/if-thingisbeautiful-andand-thingisbiggerthanbreadbox-oror-thingisyellow--6ea8b10b-3e04-4475-be94-e29820424d7e?point=13954297841544188978501389905821696&time=11491&hasFrames=false , smoothly, with no hiccups, _in a local dev build_.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Editor.css
+++ b/src/devtools/client/debugger/src/components/Editor/Editor.css
@@ -230,4 +230,5 @@ separately stack the pause line visuals and the code. */
 
 .editor-wrapper.showLineHits .CodeMirror-gutters .hit-markers {
   width: 30px;
+  position: relative;
 }
--- a/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
+++ b/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
@@ -6,10 +6,8 @@
   align-items: center;
   justify-content: center;
 
+  /* Height will be managed by the virtualized list renderer */
   width: 100%;
-  /* height: 15px;
-  height: calc(var(--theme-code-font-size) * 1.4); */
-
   font-size: 8px;
   border-left: 5px solid var(--gutter-hit-count-background);
 
--- a/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
+++ b/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
@@ -7,7 +7,8 @@
   justify-content: center;
 
   width: 100%;
-  height: calc(var(--theme-code-font-size) * 1.4);
+  height: 15px;
+  /* height: calc(var(--theme-code-font-size) * 1.4); */
 
   font-size: 8px;
   border-left: 5px solid var(--gutter-hit-count-background);
--- a/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
+++ b/src/devtools/client/debugger/src/components/Editor/LineHitCounts.module.css
@@ -7,8 +7,8 @@
   justify-content: center;
 
   width: 100%;
-  height: 15px;
-  /* height: calc(var(--theme-code-font-size) * 1.4); */
+  /* height: 15px;
+  height: calc(var(--theme-code-font-size) * 1.4); */
 
   font-size: 8px;
   border-left: 5px solid var(--gutter-hit-count-background);
--- a/src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
@@ -1,4 +1,12 @@
-import React, { useState, useMemo, useLayoutEffect, useRef, useEffect } from "react";
+import React, {
+  useState,
+  useMemo,
+  useLayoutEffect,
+  useRef,
+  useEffect,
+  useCallback,
+  useReducer,
+} from "react";
 import { createPortal } from "react-dom";
 import { VariableSizeList as List } from "react-window";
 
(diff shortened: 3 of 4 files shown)


jasonLaster (commented):
All for trying something like this which could have a nice perf win. 

Will defer to @bvaughn though since he knows a thing or two about it

jasonLaster:
These seem like fixable issues
https://app.replay.io/recording/toggling-print-statements-does-not-adjust-the-gutter--f6497310-9cbd-42d0-ba14-1f4ba41da14e?point=98004603410387116657717896928046417&time=37003.635684647306&hasFrames=false

bvaughn:
> A chunk of the time in that perf screenshot was DOM reflow. Since the list is absolutely positioned, it shouldn't be affecting other layout as much

I haven't started reviewing the code yet, but reading this I found myself wondering how much [`contain`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) and maybe [`will-change`](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) could have helped with the reflow/layout issues (and if we tried these at any point).

bvaughn:
I guess I'm also a little surprised that our own windowing of gutter hit counts helped this much, given that Code Mirror _already_ windows its content. Admittedly it uses a different windowing component/library but that makes me wonder if we weren't doing something else that was exacerbating the issue.

markerikson:
I just ran a quick perf capture from this branch scrolling through that big TSC `run.js` file. I'm still seeing visible lag and jank, and the perf capture shows the same ~150ms blocking from CodeMirror iterating over lines and updating markers:

image

souporserious:
Amazing work here, Mark! I love creative solutions like this for performance issues.

I have few thoughts and notes after looking over production, this PR, as well as Brian's PR.

## Production performance

I tested all of these in production and they seem to be relatively the same _after_ initially scrolling. So if I've previously scrolled an area, it doesn't seem to drop performance much in all cases. I didn't look extensively, but it feels like we are trying to fetch data during scroll which is why it's faster to scroll back through previously visited lines.

In all of the cases, this PR seems to scroll new content in the fastest (strictly just physically testing scrolling, not profiling).

## Layout thrashing

As the performance profiles show, we are getting layout thrashing currently in production. I think this is mostly due to us calling `createElement` multiple times on every frame (great example of this are tooltips that are appended to the end of the document will cause a full reflow of all body styles 😱). Your PR showcases a fix for this well since our own additional virtualization mechanism seems to help the underlying issue of creating too many elements during scroll.

## Possible solution

Some concerns I have with this particular approach are it feels hacky that we're having to create an additional virtual list and then coordinate that with CodeMirror. I'm not super familiar with CM, and the docs are somewhat sparse on gutters from what I remember, but maybe there's even a first-class way we can handle this that avoids layout thrashing altogether.

A different clever solution here could be strictly using CM gutters and a data-attribute to display the hit count through CSS pseudo elements. This would have the benefit of not creating additional nodes or having to maintain a separate list. Combined with a RAF solution like what Brian did, it could be a nice bump in perf.

With all of that said, I know we want to ship a new editor implementation in the near future so I say we ship this if it feels better, especially if this code is going to be getting a major update soon anyways!

markerikson:
Unfortunately, after further testing, we've found some cases that I can't manage to get working.  In particular, adding a new print statement causes later hit counts boxes to be misaligned, because the item height was already cached.  Spent some time trying to get it to cooperate, with no luck.

Given that, I'm (sadly) going to have to close this.

On the flip side, I _have_ managed to find some more optimizations that will improve the existing CodeMirror-based hit counts gutter performance, and will have those up in a follow-up PR.

PR #7756, opened 2022-09-16
Batch Code Mirror updates into RAF

bvaughn:
Potential alternative to #7753 for discussion.

Basically this batches the updates we were making (in layout effects) that invalidated layout into animation frames. Seems like the impact of this change is comparable to the change in #7753 based on my anecdotal perf profiling?

Loom walk through: https://www.loom.com/share/f930fdb5c13b42acb2a0c92dd584b674

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
@@ -33,7 +33,7 @@ interface ELProps {
 type FinalELProps = PropsFromRedux & ELProps;
 
 class EmptyLines extends Component<FinalELProps> {
-  _rafId: number | null = null;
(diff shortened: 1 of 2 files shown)


markerikson:
I just ran a quick perf capture from this branch using the same big TSC `run.js` file I'd been checking against.  I can still see a bunch of visible lag and jank as I scroll, and the perf capture shows the same ~150ms CodeMirror line iteration chunks blocking the thread:

image

bvaughn:
Swapped RAF out for idle callback. Seems like it's working well?
Screen Shot 2022-09-16 at 1 02 20 PM

Wondering if you're still seeing chunks like the above.

bvaughn:
https://www.loom.com/share/2a266da80f684699993a543efad61e95

PR #7549, opened 2022-08-12
new console messages polish

souporserious:
Further refinement of messages to match previous console.

PR #7758, opened 2022-09-16
Removed current time indicator z-index

bvaughn:
PR #7549 added some z-indexes to things in the new Console that caused regressions for the context menu.

I think I've fixed the context menu issue without regressing on the other things #7549 was intended to fix, although the solution is more complex than I wish. Here is a walk-through video showing the console after this change:

https://www.loom.com/share/2a266da80f684699993a543efad61e95


bvaughn:
Ah, I think I get why this was being done. It was so the border from the error row was visible on both top and bottom (it's red, vs the normal not-red border).

Okay let me re-add z-index for these two things, _and_ the context menu.

jasonLaster (commented):
I believe we want to show the pause line at the beginning / end of the logs

souporserious:
@bvaughn yeah, exactly top/bottom! My bad, I should have documented that. We should probably do a z-index audit and get all of those into the design package.

bvaughn:
I need to sort out a small alignment issue with the top border and the current time. I want to add a -1px offset to the current time placement, except that it doesn't work well if it's at the top of the console. Need to sort that out then I'll push an update.

bvaughn:
https://www.loom.com/share/2a266da80f684699993a543efad61e95

bvaughn:
I am going to add playwright tests for this now.

PR #7759, opened 2022-09-17
Rewrote fixture logic for tests

bvaughn:
This PR dramatically changes the way fixture data is managed for Playwright tests.

# Background
Playwright tests run on every PR to help detect regressions in the devtools. In the pasts, these tests have sometimes failed due to temporary backend outages, (which produces a false positive for the PR). To avoid this, all communication with the backend API is done through an interface which (for testing purposes) is a `Proxy` that can [replay](https://github.com/replayio/devtools/blob/main/packages/shared/proxy/createPlayer.ts) that was previously [recorded](https://github.com/replayio/devtools/blob/main/packages/shared/proxy/createRecorder.ts) whenever requests are made to the API.

This recorded data must be updated each time a new tests is written as well as when changes are made to the protocol API methods.

# Motivations
My reasons for changing the way proxy data is recorded and managed is to simplify the following use cases:
1. Adding new tests without impacting existing fixtures
   * Also removing the requirement to mirror test actions in a special "setup" function to generate fixture data
2. Updating fixture data without manually editing anything 
   * Including adding the ability to update fixture data for individual tests
3. Remove the requirement for each test to have a dedicated test harness (which was often redundant)

To accomplish this, fixture data has been moved into a new directory (in `packages/bvaughn-architecture-demo/playwright/protocol-fixtures`), one file per test case. This data is loaded into the test shell at runtime via a new Next API route (`pages/api/data.ts`) which expected a `fixtureDataPath` query param computed from the test file and test case name (e.g. "/api/data?fixtureDataPath=source-and-console/should-gracefully-handle-invalid-remote-analysis.json").

# Remaining work
- [ ] Solve failing tests
   * Currently there is a test that fails with the recorded fixture data: "_should support setting focus range via the context menu_". This probably as to do with the original sequence of method calls and events misaligning with the static fixture data (which strips events and merges some of the method calls). My guess is that this results in a slightly different sequence of best-guess time-to-point mappings, which in turn impacts API method call parameters.

# Updating protocol fixture data

The process of updating fixture data starts as before:
```sh
# packages/bvaughn-architecture-demo/playwright
yarn test:update-fixture-data
```

The command above runs all tests (or tests selected with `test.only`). Upon successful completion, each test's fixture data will be updated:
Screen Shot 2022-09-16 at 8 41 15 PM

If a test fails, the updater will just print an error message (and leave the fixture data unmodified):
Screen Shot 2022-09-16 at 8 44 18 PM

If a test does something unexpected– like generating no fixture data, an error will be thrown:
Screen Shot 2022-09-16 at 8 51 57 PM

# Running tests
Tests are otherwise run as before, as described in the README.


bvaughn:
Ignore the temporary state of this PR. The JSON won't be on multiple lines if/when this lands.

bvaughn:
This is working well except for _one test_. I've recorded a Replay to dig in.

Here's the test in _record_ mode: [8ffb82c4-65ab-4d19-90cf-3ced00bbce6b](https://app.replay.io/recording/playwright-record-proxy--8ffb82c4-65ab-4d19-90cf-3ced00bbce6b)

And here's the test in _playback_ mode (where it fails): [f5da2ee0-eca6-4e6f-ac59-04531b1456f2](https://app.replay.io/recording/playwright-play-proxy--f5da2ee0-eca6-4e6f-ac59-04531b1456f2)

I think I know roughly what's the underlying cause but I want to go through and annotate it.

bvaughn:
From what I can tell, Playwright released the HAR functionality a little after I landed the initial version of this approach (or I would have started with it). I think the remaining blocking issue– which has to do with non-deterministic ids and timing (request timing) would likely still exist with Playwright's solution.

PR #7760, opened 2022-09-17
Show the pause line faster when stepping

jasonLaster:
This change starts updating our pause reducer so that we can use the returned target location as the pause preview location.

There are some other things we'll want to do here to ensure preview pause locations are set/cleared appropriately, but should be hot

https://www.loom.com/share/c4c5fceb42184e31bf9fc63081a30ae8

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
@@ -9,7 +9,6 @@ import React, { Component } from "react";
 import { connect, ConnectedProps } from "react-redux";
 import { selectors } from "ui/reducers";
 import { actions } from "ui/actions";
-import { previewLocationCleared } from "devtools/client/debugger/src/reducers/pause";
 
 import classnames from "classnames";
 import ReactTooltip from "react-tooltip";
@@ -106,7 +105,7 @@ class FrameTimeline extends Component<PropsFromRedux, FrameTimelineState> {
   };
 
   onMouseUp = (event: MouseEvent) => {
-    const { seek, clearPreviewPausedLocation } = this.props;
+    const { seek } = this.props;
 
     const progress = this.getProgress(event.clientX);
     const position = this.getPosition(progress);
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
```diff
@@ -106,15 +105,14 @@ class FrameTimeline extends Component<PropsFromRedux, FrameTimelineState> {
...
-    const { seek, clearPreviewPausedLocation } = this.props;
+    const { seek } = this.props;
 
     const progress = this.getProgress(event.clientX);
     const position = this.getPosition(progress);
     this.setState({ scrubbing: false });
 
     if (position) {
       seek(position.point, position.time, true);
-      clearPreviewPausedLocation();
```
jasonLaster, line 117: This fixes a jumping issue

https://www.loom.com/share/21fe570bdf2c4d48a1a6a3544d842a28

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -114,17 +114,26 @@ const initialState: PauseState = {
...
+    const { extra, getState } = thunkApi;
+    const { ThreadFront } = extra;
+    const state = getState();
+    const loadedRegions = getLoadedRegions(state)!;
+    const nextPoint = getResumePoint(state, command)!;
+
+    resp = await ThreadFront[command](nextPoint, loadedRegions);
 
-  await ThreadFront[command](nextPoint, loadedRegions);
+    location = getPreferredLocation(state, resp?.frame!, ThreadFront.preferredGeneratedSources);
```
jasonLaster, line 131: I copied this from another place, not sure if it is the way to go

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -315,18 +325,15 @@ const pauseSlice = createSlice({
...
     },
   },
   extraReducers: builder => {
     builder
       .addCase(executeCommandOperation.pending, (state, action) => {
         const { command } = action.meta.arg;
-
         state.command = command;
         state.lastCommand = command;
-        state.threadcx.pauseCounter++;
```
jasonLaster, line 329: this might be the part i'm least sure of. I removed a bunch of these incrementers `state.threadcx.pauseCounter` so that we only bumped it when we truly paused. I think that is safe, but it would be worth thinking about.

markerikson (approved):

PR #7763, opened 2022-09-17
Grow the second row of the `ConsoleRoot`'s correctly

Andarist:
I've noticed that the console's input was not stuck to the bottom of the page. I've inspected the DOM and this is was I've seen:

Screenshot 2022-09-17 at 07 48 22

As we might notice, this grid has only 2 rows. I've rechecked the children of [this component](https://github.com/replayio/devtools/blob/9ae4c4d6818b7c3f9d12970bff6d3b41bf449246/packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx#L44) to check if maybe there are some conditional rows or something but I can't see any - so I think that the fix is correct but maybe I'm missing something.

cc @bvaughn


packages/bvaughn-architecture-demo/components/console/ConsoleRoot.module.css
```diff
@@ -1,9 +1,7 @@
 .ConsoleRoot {
   display: grid;
   grid-template-columns: auto 2fr;
-  grid-template-rows: auto auto 1fr;
-  flex-direction: column;
-  flex: 1 1 auto;
+  grid-template-rows: auto 1fr;
```
Andarist, line 4: this is the fix

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.module.css
```diff
@@ -1,9 +1,7 @@
 .ConsoleRoot {
   display: grid;
   grid-template-columns: auto 2fr;
-  grid-template-rows: auto auto 1fr;
-  flex-direction: column;
-  flex: 1 1 auto;
```
Andarist, line 6: those were redundant as this container is a grid and not a flex

jasonLaster (commented):
Wait! You also understand css grid? You know everything

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.module.css
bvaughn: I think this is left over cruft from when it used to be a flex. Nice.

bvaughn (approved):
Yeah, I think you're right. Thanks!

bvaughn:
I think this change will likely invalidate our screenshot tests. Can we hold off landing it until after #7759 lands – then I'll own updating the screenshots.

bvaughn:
I'm going to regenerate screenshots. Hopefully I'll have the option to push to this branch. Worst case, I'll just copy the fix over to a new branch.

markerikson:
This appears to have broken the console messages list layout - it's not filling the width properly:

image

looks like it worked okay as of the previous merge in #7762 , so I'm going to revert this

bvaughn:
Thanks for the catch, mark!

bvaughn:
> This appears to have broken the console messages list layout - it's not filling the width properly

I'm very curious about how to repro this.

PR #7768, opened 2022-09-18
Treat too many points as an expected error

jasonLaster:


markerikson (approved):

src/devtools/client/debugger/src/actions/pause/setFramePositions.ts
```diff
@@ -16,7 +16,18 @@ export function setFramePositions(): UIThunkAction<Promise<void>> {
...
     }
 
-    const positions = await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    let positions;
+    try {
+      positions = await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    } catch (e: any) {
+      // "There are too many points to complete this operation"
+      // is expected if the frame is too long and should not be treated as an error.
+      if (e.code == 55) {
```
bvaughn, line 25: Follow up request since this has already landed: Could we avoid using the magic number 55 here and use the `ProtocolError.TooManyPoints` constant instead?

https://github.com/replayio/devtools/blob/6a5b39d64e0e6078125ecbe422a8a0664a8c5a58/src/ui/state/app.ts#L143-L145

PR #7774, opened 2022-09-19
Add naive event dispatcher support to Playwright test proxy

bvaughn:
I'm not sure this PR is worth landing. _It works_ but only for a naive event dispatcher case.


bvaughn:
#7776 is a better solution

PR #7775, opened 2022-09-19
Rewrite hit counts logic to improve perf and correctness

markerikson:
This PR:

- Starts with Brian's attempt to batch some hit counts updates using `requestIdleCallback`, and switches it back to his previous attempt to use `requestAnimationFrame` (which I found to be smoother)
- Makes some major updates to the hit counts rendering logic to improve perf:
  - Wrapped the `add/removeClass` lines in `cm.operation()`, which delays all CM-internal reflow calculations until after all of those individual operations are complete
  - **Reworked the line index handling to fix an off-by-one issue caused by calling `cm.getLineInfo(lineNumber)` (FE-748)**. CM uses 0-based indexing, but our `lineNumber` at that point was 1-indexed.  So, I think we were getting the right hit counts out of the map, but attaching the marker DOM node on the wrong line.
  - Improved behavior around when we bail out of updating the markers
- Improved perf around fetching hit counts for a hovered line:
  - Debounced the `fetchHitCountsForVisibleLines` method so we don't keep running it as you mouse around
  - Fixed the logic for calculating top/bottom lines so we don't go out of bounds and try to fetch lines `-100..0`
  - Deduplicated lines that could be in the same 100-line chunk, to minimize thunk dispatches
- Adds the correct `editor: SourceEditor` type across our editor UI components
- Updates "get breakable columns for a line" selection logic to use a normalized lookup table, which fixes a small hotspot I consistently saw in `useGetHitPointsForHoveredLine` and `getFirstBreakpoint`

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
@@ -33,7 +33,7 @@ interface ELProps {
 type FinalELProps = PropsFromRedux & ELProps;
 
 class EmptyLines extends Component<FinalELProps> {
-  _idCallbackId: number | null = null;
+  _rafId: number | null = null;
 
   componentDidMount() {
     this.disableEmptyLinesRaf();
@@ -44,8 +44,8 @@ class EmptyLines extends Component<FinalELProps> {
   }
 
   componentWillUnmount() {
-    if (this._idCallbackId) {
-      cancelIdleCallback(this._idCallbackId);
+    if (this._rafId) {
+      cancelAnimationFrame(this._rafId);
     }
 
     const { editor, lower, upper } = this.props;
@@ -58,12 +58,12 @@ class EmptyLines extends Component<FinalELProps> {
   }
 
   disableEmptyLinesRaf = () => {
-    if (this._idCallbackId) {
-      clearTimeout(this._idCallbackId);
+    if (this._rafId) {
+      cancelAnimationFrame(this._rafId);
     }
 
-    this._idCallbackId = requestIdleCallback(() => {
-      this._idCallbackId = null;
+    this._rafId = requestAnimationFrame(() => {
+      this._rafId = null;
       this.disableEmptyLines();
     });
   };
--- a/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
@@ -33,7 +33,7 @@ interface ELProps {
 type FinalELProps = PropsFromRedux & ELProps;
 
 class EmptyLines extends Component<FinalELProps> {
-  _rafId: number | null = null;
+  _idCallbackId: number | null = null;
 
   componentDidMount() {
     this.disableEmptyLinesRaf();
@@ -44,8 +44,8 @@ class EmptyLines extends Component<FinalELProps> {
   }
 
   componentWillUnmount() {
-    if (this._rafId) {
-      cancelAnimationFrame(this._rafId);
+    if (this._idCallbackId) {
+      cancelIdleCallback(this._idCallbackId);
     }
 
     const { editor, lower, upper } = this.props;
@@ -58,12 +58,12 @@ class EmptyLines extends Component<FinalELProps> {
   }
 
   disableEmptyLinesRaf = () => {
-    if (this._rafId) {
-      cancelAnimationFrame(this._rafId);
+    if (this._idCallbackId) {
+      clearTimeout(this._idCallbackId);
     }
 
-    this._rafId = requestAnimationFrame(() => {
-      this._rafId = null;
+    this._idCallbackId = requestIdleCallback(() => {
+      this._idCallbackId = null;
       this.disableEmptyLines();
     });
   };
--- a/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
@@ -33,15 +33,21 @@ interface ELProps {
 type FinalELProps = PropsFromRedux & ELProps;
 
 class EmptyLines extends Component<FinalELProps> {
+  _rafId: number | null = null;
+
   componentDidMount() {
-    this.disableEmptyLines();
+    this.disableEmptyLinesRaf();
   }
 
   componentDidUpdate() {
-    this.disableEmptyLines();
+    this.disableEmptyLinesRaf();
   }
 
   componentWillUnmount() {
+    if (this._rafId) {
+      cancelAnimationFrame(this._rafId);
+    }
+
     const { editor, lower, upper } = this.props;
 
     editor.codeMirror.operation(() => {
@@ -50,6 +56,17 @@ class EmptyLines extends Component<FinalELProps> {
       });
     });
   }
+
+  disableEmptyLinesRaf = () => {
+    if (this._rafId) {
+      cancelAnimationFrame(this._rafId);
+    }
+
+    this._rafId = requestAnimationFrame(() => {
+      this._rafId = null;
+      this.disableEmptyLines();
+    });
+  };
 
   disableEmptyLines() {
     const { breakableLines, editor, lower, upper } = this.props;
--- a/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
@@ -10,6 +10,11 @@ import { getBreakableLinesForSelectedSource } from "ui/reducers/possibleBreakpoi
 import { getBoundsForLineNumber } from "ui/reducers/hitCounts";
 import type { UIState } from "ui/state";
 
+import { calculateHitCountChunksForVisibleLines } from "devtools/client/debugger/src/utils/editor/lineHitCounts";
+import { editorItemActions } from "./menus/editor";
+
+import type { SourceEditor } from "../../utils/editor/source-editor";
+
 const mapStateToProps = (state: UIState) => {
   const breakableLines = getBreakableLinesForSelectedSource(state);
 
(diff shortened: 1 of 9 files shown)


src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -27,21 +28,27 @@ const connector = connect(mapStateToProps);
...
 
 interface ELProps {
-  editor: any;
+  editor: SourceEditor;
 }
 
 type FinalELProps = PropsFromRedux & ELProps;
 
 class EmptyLines extends Component<FinalELProps> {
+  _rafId: number | null = null;
```
jcmorrow, line 36: It took me a minute to realize `raf` was `requestAnimationFrame`. It's a mouthful, so I am not sure I want to change the variable name, and I realized it pretty quick, but thought I'd mention it in case anyone else has good suggestions for another name

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -167,89 +166,106 @@ function LineHitCounts({ sourceEditor }: Props) {
...
+          const currentCMLineNumber = editor.getLineNumber(lineHandle);
+          if (currentCMLineNumber === null) {
+            return;
+          }
 
-        // Skip this line if we've updated it with its current hit count already
-        if (updatedLineNumbers?.has(lineNumber)) {
-          return;
-        }
+          // CM uses 0-based indexing. Our hit counts are 1-indexed.
```
jcmorrow, line 182: It might be worth mentioning that our `Location` type is really where we get this from, and it's 1-indexed on lines, 0-indexed on columns. Or it might not 🙃

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -167,89 +166,106 @@ function LineHitCounts({ sourceEditor }: Props) {
...
+            return;
           }
-          className = styles[`HitsBadge${index + 1}`];
-        }
 
-        if (features.disableUnHitLines) {
+          updatedLineNumbers?.add(oneIndexedLineNumber);
+
+          // We use a gradient to indicate the "heat" (the number of hits).
+          // This absolute hit count values are relative, per file.
```
jcmorrow, line 194: `This` should probably be `These`.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -167,89 +166,106 @@ function LineHitCounts({ sourceEditor }: Props) {
...
           }
-          markerNode.textContent = hitsLabel;
-        }
-        markerNode.title = `${hitCount} hits`;
+
+          markerNode.className = className;
+          if (!isCollapsed) {
+            let hitsLabel = "";
+            if (hitCount > 0) {
+              hitsLabel = hitCount < 1000 ? `${hitCount}` : `${(hitCount / 1000).toFixed(1)}k`;
```
jcmorrow, line 239: this feels like it could be a nice helper function given how complex everything around already is, but I don't feel too strongly.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
```diff
@@ -167,89 +166,106 @@ function LineHitCounts({ sourceEditor }: Props) {
...
+          }
+          markerNode.title = `${hitCount} hits`;
+        });
       });
     };
 
-    editor.on("change", drawLines);
-    editor.on("swapDoc", drawLines);
+    let rafId: number | null = 0;
+    // let idCallbackId: number | null = null;
```
jcmorrow, line 249: What's up here?

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -96,13 +97,35 @@ function LineNumberTooltip({ editor, keyModifiers }: Props) {
     var rect = editor.codeMirror.getWrapperElement().getBoundingClientRect();
     var topVisibleLine = editor.codeMirror.lineAtHeight(rect.top, "window");
     var bottomVisibleLine = editor.codeMirror.lineAtHeight(rect.bottom, "window");
-    dispatch(fetchHitCounts(source!.id, editor.codeMirror.lineAtHeight()));
-    dispatch(fetchHitCounts(source!.id, topVisibleLine - 10));
-    dispatch(fetchHitCounts(source!.id, bottomVisibleLine + 10));
+
+    const viewport = editor.editor.getViewport();
+    const { from: topViewportLine, to: bottomViewportLine } = viewport;
+    const centerLine = ((bottomViewportLine - topViewportLine) / 2) | 0;
```
jcmorrow, line 103: Is this supposed to be `||` or did you mean `| 0`? 🙃

src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -33,15 +33,21 @@ interface ELProps {
...
```
jcmorrow, line 36: It took me a minute to realize `raf` was `requestAnimationFrame`. I'm not saying we should spell out `requestAnimationFrame`, because that is long, but thought I'd mention it in case you or anyone else has a better suggestion.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -96,13 +97,35 @@ function LineNumberTooltip({ editor, keyModifiers }: Props) {
...
+    );
+    const lineBoundsMap = new Map<number, { lower: number; upper: number }>();
+
+    // Deduplicate bounds chunks based on their `lower` line number
+    for (let entry of lineBounds) {
+      lineBoundsMap.set(entry.lower, entry);
+    }
+
+    // Now try to fetch hit counts for the unique bounds chunks
+    for (let bounds of lineBoundsMap.values()) {
```
jcmorrow, line 122: This feels like a lot of ceremony to `uniq` these values. Would it be clearer to just call `uniq`?

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -115,6 +138,10 @@ function LineNumberTooltip({ editor, keyModifiers }: Props) {
       }
     }, 10);
 
+    const debouncedFetchHitCountsForVisibleLines = debounce(() => {
+      fetchHitCountsForVisibleLines();
+    }, 100);
```
jcmorrow, line 143: I'm curious if you can feel the tenth of a second lag. I feel like maybe this should be pretty low, just to take care of the *really* fast scrolling case. Having it be responsive as you scroll feels really good if we can pull it off. Anyways - we can iterate on this as well.

src/devtools/client/debugger/src/components/SourceOutline/getOutlineSymbols.tsx
```diff
@@ -50,7 +53,7 @@ function addHitCountsToFunctions(
 
   return functions.map(functionSymbol => {
     const hitCount = getClosestHitCount(hitCountsByLine, functionSymbol.location);
-    return { ...functionSymbol, hits: hitCount?.hits };
+    return Object.assign({}, functionSymbol, { hits: hitCount?.hits }); // { ...functionSymbol, hits: hitCount?.hits };
```
jasonLaster, line 56: can this be cleaned up?

jcmorrow, line 56: what's up with this comment?

jcmorrow (approved):
Just the one thing about debounce times might be worth visiting, the rest is pretty nitty. I'm gonna play with the preview branch rn and see if my debounce comment holds up at all 🙂 

This is great!!! 🎉

src/devtools/client/debugger/src/components/SourceOutline/getOutlineSymbols.tsx
markerikson: This was actually intentional.

In an earlier perf capture I noticed a bunch of `_objectSpread()` instances taking up a few milliseconds here, and I was wanting to see if a manual `Object.assign()` would improve this.

I don't know _why_ that would be getting transpiled, tbh. It shouldn't be given our Babel settings.

But at worst this shouldn't actually hurt.

src/devtools/client/debugger/src/components/Editor/LineHitCounts.tsx
markerikson: Leftover, will delete

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
markerikson: Intentional - that's a "force this to be an integer value", in case we ended up with something like 37.5.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
markerikson: Hmm. Yeah, I can maybe try `uniqBy`?

jcmorrow:
Hmmmm... I'm hitting some issues on the preview branch where the lines never show up https://app.replay.io/recording/jotai-demo-forked-codesandbox--2623ca64-e499-49b8-ae6b-8c50a5c0b451

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
markerikson: I tried 10ms initially, and that wasn't enough time - this was still getting called a bunch. I wanted to make sure we wouldn't keep calling immediately as you move the mouse up and down lines.

markerikson:
Updated the PR to remove the early bailouts. Also, all three of `EmptyLines`, `LineHitCounts`, and `LineNumberTooltip` now use the same "figure out which 100-line chunks overlap the visible lines" logic.

I'm seeing some more CM iteration work than before this commit, but it's still fairly smooth, and better to have something that's consistent and correct _now_ so we can merge this, then check for more bailouts later.

PR #7777, opened 2022-09-20
Optimize EmptyLines perf by only drawing each range once

markerikson:
This PR:

- Further optimizes the `<EmptyLines>` component after yesterday's PR, by ensuring we only attempt to draw each 100-line range once
- Also rewrites it to be a function component

This eliminates the main additional source of CPU time I was seeing after I merged #7775


src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -1,109 +1,92 @@
...
   editor: SourceEditor;
 }
 
-type FinalELProps = PropsFromRedux & ELProps;
-
-class EmptyLines extends Component<FinalELProps> {
-  _animationFrameId: number | null = null;
+export default function EmptyLines({ editor }: ELProps) {
+  const breakableLines = useAppSelector(getBreakableLinesForSelectedSource);
+  const animationFrameId = useRef<number | null>(null);
```
bvaughn, line 23: This ref isn't necessary, is it? You only reference the values it holds inside of the effect, so you should just be able to use a local variable.

src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -1,109 +1,92 @@
...
-          if (breakableLines?.includes(line)) {
-            editor.codeMirror.removeLineClass(lineHandle, "line", "empty-line");
-          } else {
-            editor.codeMirror.addLineClass(lineHandle, "line", "empty-line");
-          }
-        });
+    return () => {
+      if (animationFrameId.current) {
+        // We _want_ to clean this up if it exists
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
bvaughn, line 81: I don't think this eslint comment will be necessary either if we use a local variable.

src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -1,109 +1,92 @@
...
+        for (let hitCountChunk of uniqueChunks) {
+          const { lower, upper } = hitCountChunk;
 
-  componentWillUnmount() {
-    if (this._animationFrameId) {
-      cancelAnimationFrame(this._animationFrameId);
+          if (drawnRanges.has(lower)) {
+            continue;
+          }
+          drawnRanges.add(lower);
```
bvaughn, line 57: Something about this feels fishy to me but maybe I'm misunderstanding it.

We're memoizing our line hit count work based on the lower bound of each slice, but the size of the slices (and so the lower bounds) depends on whatever size the viewport happens to be when we call `calculateHitCountChunksForVisibleLines`:
https://github.com/replayio/devtools/blob/ff1c0ab0a5172b4e64e1d062df27f45544365cef/src/devtools/client/debugger/src/utils/editor/lineHitCounts.ts#L5-L26

But we don't listen for any kind of "resize" event or account for this change in any other way that I can tell.

Am I misunderstanding?

markerikson: Empty Lines are different than Hit Counts.  (Admittedly I'm reusing a function that was written for hit counts originally, but it's not specific to the actual hit counts feature - I should probably rename it.)

We get the list of "all breakable lines", _once_ per file, and don't have to keep refetching that info.

markerikson: More details:

Overall, the goal is to check each line, _once_, and possibly apply an `"empty-lines"` class so that CM draws it grayed-out.

There's also no point in re-checking a given 100-line chunk once we've done it before - the list of possible breakable lines for a file won't change once we've fetched it.  So we can skip over a chunk after we know we've checked it once.

In general, we want to _try_ to check around the visible lines content area any time the user changes what lines they're looking at.  So, listening for `"scroll"` and `"change"`, etc, does that for us.

bvaughn: I don't understand.

We get these "chunks" by calling `calculateHitCountChunksForVisibleLines`:
```js
const uniqueChunks = calculateHitCountChunksForVisibleLines(editor);
```

Which looks like this:
https://github.com/replayio/devtools/blob/ff1c0ab0a5172b4e64e1d062df27f45544365cef/src/devtools/client/debugger/src/utils/editor/lineHitCounts.ts#L5-L26

So the chunks are computed around a "center line" which is based on the viewport size, as well as the "buffer" which is based on which lines happen to be visible at the top and bottom of the window.

If we only computed this once per session _I guess_ it would be okay, but this code is calling `calculateHitCountChunksForVisibleLines` each time the `disableEmptyLines` is called:
```js
const uniqueChunks = calculateHitCountChunksForVisibleLines(editor);
```

This seems like an unsafe "side effect" type thing that would invalidate our memoization, no? Like what if we bail out here:
https://github.com/replayio/devtools/blob/b09df9ada7276d0ce0b0b93ddbe6545c89d975f7/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx#L54-L57

Because we have a match already for that line number– but the "chunk size" is different (bigger0, and we end up not processing some lines because of it?

bvaughn: I didn't trace the code far enough to get to here:
https://github.com/replayio/devtools/blob/ff1c0ab0a5172b4e64e1d062df27f45544365cef/src/ui/reducers/hitCounts.ts#L80-L91

bvaughn (approved):

PR #7778, opened 2022-09-20
Switch default hit counts setting to "show-counts"

markerikson:
Per request in SCS-129  , this changes the pref for hit counts to "show numbers" by default.  

(If users have actually selected a setting themselves, that setting will be preserved.)

### DIFF
--- a/src/ui/utils/prefs.ts
+++ b/src/ui/utils/prefs.ts
@@ -18,7 +18,7 @@ pref("devtools.sidePanelSize", "240px");
 pref("devtools.theme", "system");
 pref("devtools.toolbox-size", "50%");
 pref("devtools.consoleFilterDrawerExpanded", true);
-pref("devtools.hitCounts", "hide-counts");
+pref("devtools.hitCounts", "show-counts");
 
 // app features
 pref("devtools.features.basicProcessingLoadingBar", false);


markerikson:
Checked this works by opening up https://devtools-git-feature-scs-129-default-hit-counts-recordreplay.vercel.app/recording/reduxts-csb-2--be1a4568-6c84-474c-a9eb-3a0491a792d8?point=16550446245188038797950505902407747&time=6216.022988505747&hasFrames=true in an anonymous window and verifying that hit counts show up immediately:

image

jasonLaster:
I'd prefer holding off on this for the time being

jonbell-lot23:
> I'd prefer holding off on this for the time being

We discussed in Discord, had a team brainstorm, registered various points of view, then I summarised the team ms decision by saying we'd move forward on this.

Jason, I know this is isn't your favourite, but we followed a process and without a "disagree and commit" from you here, this feels like a busted process.

jonbell-lot23:
(I left a comment above before seeing the Linear discussion, so I'll paste my Linear response here as well)

```
* We just have a difference of opinion, which is healthy
* We discussed as a team, which is healthy
* We came up with a plan in that meeting, even though we didn't have 100% consensus, which is healthy
* Several of us feel very strongly that is directly helps NUX. This is not a side issue, it's a direct response to our goal to improve NUX metrics.

Our little print statement plus sign says "Click here!" whereas this feature says "Here's a map of your document!" They do different things and each can benefit each other.
```

jonbell-lot23:
@markerikson @jasonLaster 

A few days ago Jason proposed putting this on hold in order to:

* Improve the print statement button (done)
* Not show the hit count tooltip when labels are shown (tbd)
* Improve the tooltip (not blocking)

Mark, as a follow-up to our chat today, can you not show the hit count tooltip when labels are shown? At that point, I think it's important to #biastowardsaction and #actwithurgency and get this live so we can start testing this.

jasonLaster:
> Not show the hit count tooltip when labels are shown (tbd)

I was actually making a different point. It's nice not to show the tooltip which is redundant, but from a NUX perspective the tooltip could be an important aspect of DES-86 and making adding print statements more discoverable.

jonbell-lot23:
I am copy/pasting what I wrote in Linear here:

Hey team! This issue has led to a lot of discussions in groups and 1:1s, so I want to summarise where we are and next steps:

We all agree this is a big change. People who want to use it in NUX say it's big in a good way. But those who are concerned about downstream effects say it's big … and in ways that might bite us.

We all agree that we need a lot of NUX exploration. I was on the side of "let's do this and then continue iterating!" Jason's POV is "this is all part of a unified experience, let's explore a bit more and understand how the pieces fit."

When we discussed this last week, we had crossed wires — some of us left the meeting thinking we were going to turn it on, others thought we agreed not to. That led to a misalignment.

Next steps

We're going to table this feature for now so design can explore further. I apologise for the lost cycles, but these discussions will help us as we continue iterating.

We may end up turning this on by default for NUX only, or maybe everyone, or maybe we just tell people about the setting. Stay tuned for more info as we iterate further.

markerikson:
I think this is OBE now that we have the new source viewer configured.

PR #7779, opened 2022-09-20
use css for panel width

souporserious:
Since we are now using a full width panel this moves from a JavaScript solution to a CSS-only solution.

## Future Considerations

I'm prototyping an approach [here](https://codesandbox.io/s/min-fill-panels-7ebbmw) to better shrink the Print Statement Panel so the controls are always visible.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
@@ -2,10 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-.breakpoint-panel-wrapper {
-  min-width: 300px;
-}
-
 .breakpoint-panel {
   margin: 4px 0;
   background: var(--theme-toolbar-background);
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
```diff
@@ -2,16 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-.breakpoint-panel-wrapper {
-  min-width: 300px;
-}
-
 .breakpoint-panel {
-  margin: 4px 0;
+  margin: 0.25rem 0 0.5rem -1px;
```
souporserious, line 6: The negative pixel here pulls the panel back slightly so it doesn't create a 1px space between the gutter and panel.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.css
```diff
@@ -2,16 +2,10 @@
...
   background: var(--theme-toolbar-background);
-  border: 1px solid var(--theme-splitter-color);
-  border-left: none;
-  border-right: none;
+  border: 1px solid var(--theme-toolbar-background);
```
souporserious, line 8: We add the same colored border as the background to align to other elements like the active paused line which uses borders.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointNavigation.css
```diff
@@ -1,7 +1,7 @@
 .breakpoint-navigation {
   display: flex;
   flex-direction: row;
-  /* background-color: var(--theme-toolbar-background); */
+  background-color: var(--theme-toolbar-background);
```
souporserious, line 4: Reverting from prototyping in #7739

PR #7782, opened 2022-09-20
Fix console layout

bvaughn:
Fix Console layout so the top portion of the console fills all available space and the bottom portion (with the JS Term input) stays snapped to the bottom.

Relates to #7763


packages/bvaughn-architecture-demo/components/console/ConsoleRoot.module.css
```diff
@@ -1,8 +1,7 @@
 .ConsoleRoot {
   display: grid;
   grid-template-columns: auto 2fr;
-  grid-template-rows: auto auto 1fr;
-  flex-direction: column;
```
bvaughn, line 5: This prop wasn't doing anything.

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.module.css
```diff
@@ -1,8 +1,7 @@
 .ConsoleRoot {
   display: grid;
   grid-template-columns: auto 2fr;
-  grid-template-rows: auto auto 1fr;
```
bvaughn, line 4: This prop was mistaken. There were only 2 rows.

PR #7787, opened 2022-09-21
Fixing dangling links

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-107/new-documentation-links-look-awkward-in-some-states

* Added the helpful links for error pages
* Removed the helpful links for the sourcemaps view
* ... but they'll also be removed anywhere else we don't specifically set them

### DIFF
--- a/src/ui/components/shared/Dialog.tsx
+++ b/src/ui/components/shared/Dialog.tsx
@@ -4,7 +4,7 @@ import ReplayLogo from "./ReplayLogo";
 
 export type DialogPropTypes = HTMLProps<HTMLDivElement>;
 
-export function Dialog({ children, className, ...props }: DialogPropTypes) {
+export function Dialog({ children, className, showFooterLinks, ...props }: DialogPropTypes) {
   return (
     <>
       <div
@@ -13,8 +13,10 @@ export function Dialog({ children, className, ...props }: DialogPropTypes) {
         role="dialog"
         style={{ animation: "linearFadeIn ease 200ms", width: 400 }}
       >
-        {children}
+        {children}                  
       </div>
+
+      {showFooterLinks ?            
       <div>
         <div className="grid grid-cols-1 gap-1 pt-4 text-xs text-gray-400 place-items-center">
           <div>
@@ -40,6 +42,7 @@ export function Dialog({ children, className, ...props }: DialogPropTypes) {
           </div>
         </div>
       </div>
+      : null }
     </>
   );
 }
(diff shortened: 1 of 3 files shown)


src/ui/components/shared/Dialog.tsx
```diff
@@ -4,7 +4,7 @@ import ReplayLogo from "./ReplayLogo";
 
 export type DialogPropTypes = HTMLProps<HTMLDivElement>;
 
-export function Dialog({ children, className, ...props }: DialogPropTypes) {
+export function Dialog({ children, className, showFooterLinks, ...props }: DialogPropTypes) {
```
jcmorrow, line 7: I suspect that you will need to add this to the type declaration on line 5 in order to make the typecheck pass.

src/ui/components/shared/Dialog.tsx
```diff
@@ -4,7 +4,7 @@ import ReplayLogo from "./ReplayLogo";
 
 export type DialogPropTypes = HTMLProps<HTMLDivElement>;
```
jcmorrow, line 5: This line should become 

```
export type DialogPropTypes = HTMLProps<HTMLDivElement> & { showFooterLinks?: boolean };
```

src/ui/components/shared/Dialog.tsx
```diff
@@ -15,31 +15,34 @@ export function Dialog({ children, className, ...props }: DialogPropTypes) {
...
-                </a>
-              </div>
-              <div className="text-gray-300">•</div>
-              <div>
-                <a className="hover:underline" href="mailto:support@replay.io">
-                  Email support
-                </a>
+
+      {showFooterLinks ? (
+        <div>
```
jcmorrow, line 20: I think this `div` tag and the `div` tag on line 22 are probably not doing anything, but I could be wrong.

src/ui/components/shared/Dialog.tsx
```diff
@@ -15,31 +15,34 @@ export function Dialog({ children, className, ...props }: DialogPropTypes) {
...
-                  Discord
-                </a>
-              </div>
-              <div className="text-gray-300">•</div>
-              <div>
-                <a className="hover:underline" href="mailto:support@replay.io">
-                  Email support
-                </a>
+
+      {showFooterLinks ? (
```
jcmorrow, line 19: Since we have made this an optional prop (by adding the `?` in the type declaration change I suggested on line 5) it would probably be clearer and more technically correct to do `!!showFooterLinks` just to make it clear that we don't know exactly what type it is, but we are just checking for a boolean.

jcmorrow (commented):
Looking good! Let me know if you want any clarification on any suggestions!

src/ui/components/shared/Dialog.tsx
jonbell-lot23: Huh, I hadn't seen !! before. Will throw that it.

jcmorrow (approved):

PR #7791, opened 2022-09-21
Set height in requestAnimationFrame for perf

jcmorrow:
`requestAnimationFrame` fires right after the DOM has just calculated layout anyways, so reading from dom rects is usually cheaper when done here, where no recalc is required. If we call `getBoundingClientRect` in render, we are causing style recalcs, which are expensive.

Before:
https://share.firefox.dev/3dzkSy0

CleanShot 2022-09-21 at 14 18 33

After:
https://share.firefox.dev/3Bxej73

I don't have a screenshot because `getBoundingClientRect` is only in 3 samples so it does not make it on to that list 🙂

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/ToggleWidgetButton.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/ToggleWidgetButton.tsx
@@ -18,7 +18,7 @@ import findLast from "lodash/findLast";
 import { compareNumericStrings } from "protocol/utils";
 import { getExecutionPoint } from "../../reducers/pause";
 import { seek } from "ui/actions/timeline";
-import { useFeature, useStringPref } from "ui/hooks/settings";
+import { useFeature } from "ui/hooks/settings";
 import useHitPointsForHoveredLocation from "ui/hooks/useHitPointsForHoveredLocation";
 
 const QuickActionButton: FC<{
@@ -89,13 +89,21 @@ function QuickActions({
   breakpoint?: Breakpoint;
   cx: any;
 }) {
+  const [height, setHeight] = useState(18);
+  useEffect(() => {
+    const animationFrameId = requestAnimationFrame(() =>
+      setHeight(targetNode.getBoundingClientRect().height)
+    );
+
+    return () => cancelAnimationFrame(animationFrameId);
+  }, [targetNode]);
+
   const isMetaActive = keyModifiers.meta;
   const isShiftActive = keyModifiers.shift;
   const dispatch = useAppDispatch();
   const executionPoint = useAppSelector(getExecutionPoint);
   const { nags } = hooks.useGetUserInfo();
   const showNag = shouldShowNag(nags, Nag.FIRST_BREAKPOINT_ADD);
-  const { height } = targetNode.getBoundingClientRect();
   const { value: enableLargeText } = useFeature("enableLargeText");
 
   const [hitPoints, hitPointStatus] = useHitPointsForHoveredLocation();


souporserious (approved):
LGTM :shipit: for future consideration (or now if you want), since we only use `height` for the top position below I think we can refactor this so `targetNode` and `height` are calculated using a ref to bypass React rendering entirely.

jcmorrow:
> LGTM :shipit: for future consideration (or now if you want), since we only use height for the top position below I think we can refactor this so targetNode and height are calculated using a ref to bypass React rendering entirely.

Yeah, I think I see what you are suggesting. I kind of like the fact that if `height` changes the whole thing will re-render though, it makes me a little more confident that we won't accidentally get a stale height if it were to change (for zoom level reasons, or whatever).

bvaughn (commented):
Moving this out of render was a great move! 👍🏼 

Is a RAF scheduled in a passive effect the right place to move it to (vs a layout effect)? I have no idea. It will mean that the button renders and is temporarily visible at one position and then updates.

If we did the measurement in a _layout effect_ (maybe from a RAF scheduled there?) we could avoid this temporarily visible state.

I don't know if this is a problem that we care about or not though :)

bvaughn:
Generally speaking, layout effects are for things that affect _layout_. Typical example is positioning a modal or a tooltip– something that you want to do _as part of layout_ but can't quite do with the current declarative APIs. React guarantees to run this type of effect (and any updates they might schedule) _before_ yielding to the browser to paint the update(s). This means that any intermediate state will never be user-visible.

Doing this sort of thing in a "passive" effect (`useEffect`) means that it's possible (likely) that people will see things render in one place (or at one size) and then flicker to a different one.

Doing performance-sensitive stuff in layout effects isn't great because it delays paint– _but_ sometimes this is what you want/need.

bvaughn:
It's maybe also worth noting that moving perf sensitive code into a passive effect (or even a RAF scheduled by a passive effect) isn't necessary a "fix" for perf issues either– because if something else is scheduling sync/high-pri updates, React will flush pending passive effects _before_ processing the update– so the bottleneck still happens. In that case, a more traditional "debounce" technique (using `setTimeout` or `requestIdleFrame`) may be more appropriate.

jcmorrow:
> Doing performance-sensitive stuff in layout effects isn't great because it delays paint– but sometimes this is what you want/need.

OK, I am updating my model for how all of this works in real time, let's see if I got warmer or colder:

- Previously, we were calling `getBoundingClientRect` in `render`, which is not ideal because:
  - it runs on every render
  - it might force the browser to recalculate layout, which can be expensive
- Moving it to `requestAnimationFrame` means that:
  - it will run after layout was just calculated, so it will be cheap, *but*, if that raf is scheduled from a passive effect then the whole component will render once (in the wrong spot), and then render again after height gets updated
  - this is performant, but means that some renders will be incorrect
- Moving it to `useLayoutEffect` will mean it runs:
  - right before painting, but also before layout is calculated?
  - this means that if the height does not change, then layout does not need to be recalculated and it is essentially the same as running from raf, but in the case that height *does* change, it will force that update before the component renders in the wrong spot
  - In some cases, that might be a performance bummer (if height were changing all the time), but in this case, where it should not update often, it's probably about the same performance-wise as the raf approach, but will give a stronger guarantee of not painting something in the wrong spot.

Is that right?

> It's maybe also worth noting that moving perf sensitive code into a passive effect (or even a RAF scheduled by a passive effect) isn't necessary a "fix" for perf issues either– because if something else is scheduling sync/high-pri updates, React will flush pending passive effects before processing the update– so the bottleneck still happens. In that case, a more traditional "debounce" technique (using setTimeout or requestIdleFrame) may be more appropriate.

I think I only half-followed this part, but I think that's ok!

jcmorrow:
I tried doing this in a straight `useLayoutEffect` and, while it does guarantee that we won't render a few pixels off, it also causes `getBoundingClientRect` to appear on the stack about 5% of the time: https://share.firefox.dev/3qWYGAS

I'm sure we could schedule raf from inside of `useLayoutEffect`, but then, if I understand correctly, we would not be getting the benefits of `useLayoutEffect`. Anyways, I'm glad to have dove in on this, happy to see anyone else's experiments and profiles of it, and *super* happy with how snappy our editor feels right now 😄

PR #7792, opened 2022-09-21
Don't try and evaluate with `pure` from node

jcmorrow:
I noticed this error coming up for node recordings.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -89,6 +89,7 @@ export type RecordingCapabilities = {
   supportsEventTypes: boolean;
   supportsNetworkRequests: boolean;
   supportsRepaintingGraphics: boolean;
+  supportsPureEvaluation: boolean;
 };
 
 // Target applications which can create recordings.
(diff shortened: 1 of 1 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -586,7 +592,8 @@ class _ThreadFront {
     const pause = this.pauseForAsyncIndex(asyncIndex);
     assert(pause, "no pause for asyncIndex");
 
-    const rv = await pause.evaluate(frameId, text, pure);
+    const abilities = await this.recordingCapabilitiesWaiter.promise;
+    const rv = await pause.evaluate(frameId, text, abilities.supportsPureEvaluation && pure);
```
markerikson, line 596: 🛑 FYI the `evaluate()` method is dead as of earlier today, and should be deleted in a follow-up PR - try applying this change to `evaluateNew()` instead

jcmorrow: Added to the new one as well! Thanks!

markerikson (approved):

bvaughn (commented):
Niiice! 😁

PR #7795, opened 2022-09-22
[Draft] Adding nag for devtools

jonbell-lot23:
@markerikson Thanks for your help today!

Here's what my code is doing so far:

* Checks if there should be a nag
* Allows you to click the banner or the toggle to move to devtools
* Removes the nag after going to devtools

Here's what I think still needs doing, and I was hoping you can help:

* We shouldn't show this banner to existing Replay users, just people brand new to Replay
* We shouldn't show share in NUX -- the two buttons look dumb next to each other, and [Hicks' Law](https://lawsofux.com/en/hicks-law/) teaches us it's a bad idea to have multiple buttons during NUX because it will reduce success rates

Can you take a look when you get a chance? Thanks!

### DIFF
--- a/src/ui/components/Header/Header.tsx
+++ b/src/ui/components/Header/Header.tsx
@@ -52,7 +52,7 @@ function Links({ recordingTarget }: Pick<PropsFromRedux, "recordingTarget">) {
   return (
     <div className={css.links}>
       <RecordingTrialEnd />
-      {isAuthenticated ? null : null}
+      {isAuthenticated ? <ShareButton /> : null}
       <Avatars recordingId={recordingId} />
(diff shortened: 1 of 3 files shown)


markerikson:
Okay, updated:

- Rebased the branch against `main` just so it's up to date
- Added a `shouldShowDevtoolsNag` function to encapsulate the logic
- Used that in both `<Links>` and `<ViewToggle>`
- Made "Share" conditional

image

jonbell-lot23:
Thanks very much @markerikson ! This is going to really help our users get rolling :)

I'm marking this as ready for review, can you review when you get a chance?

PR #7797, opened 2022-09-22
Rewrite CSS rule fetching/parsing logic to remove fronts

markerikson:
- Added caches for AppliedRules, ComputedStyles, and BoundingRects
  - Updated existing logic to use these caches
- Rewrote `Rule/Style/StyleSheetFront` to remove `Pause` fields
- Rewrote `ElementStyle` to read style and node data from caches

Note that this leaves the `Rule/Style/StyleSheetFront` classes in existence, because the existing parsing logic relies on those heavily.  However, they are now only used from within `ElementStyle` and the other classes in `inspector/rules/models`.

I opted not to move them in this PR, but in a follow-up cleanup task I'll move those three files over to that `rules/models` folder.

### DIFF
--- a/packages/protocol/thread/style.ts
+++ b/packages/protocol/thread/style.ts
@@ -1,9 +1,9 @@
-import { StyleDeclaration } from "@replayio/protocol";
+// TODO Move this into `inspector/rules/models` as cleanup
+
+import { StyleDeclaration, Object as ProtocolObject } from "@replayio/protocol";
 import { ELEMENT_STYLE } from "shared/constants";
 
-import { assert, DisallowEverythingProxyHandler } from "../utils";
-
-import { Pause, WiredObject } from "./pause";
+import { assert } from "../utils";
 
 // Manages interaction with a CSSStyleDeclaration.
 // StyleFront represents an inline element style.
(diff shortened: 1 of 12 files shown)


packages/protocol/thread/style.ts
```diff
@@ -1,9 +1,9 @@
-import { StyleDeclaration } from "@replayio/protocol";
-import { ELEMENT_STYLE } from "shared/constants";
+// TODO Move this into `inspector/rules/models` as cleanup
 
-import { assert, DisallowEverythingProxyHandler } from "../utils";
```
bvaughn, line 4: `DisallowEverythingProxyHandler` happy to see this one gone lol

src/ui/suspense/styleCaches.ts
```diff
@@ -0,0 +1,135 @@
...
+  getValueSuspense: getBoundingRectSuspense,
+  getValueAsync: getBoundingRectAsync,
+  getValueIfCached: getBoundingRectIfCached,
+} = createGenericCache<
+  [client: ProtocolClient, sessionId: string, pauseId: PauseId, nodeId: string],
+  DOMRect | undefined
+>(
+  async (client, sessionId, pauseId, nodeId) => {
+    try {
+      const { rect } = await client.DOM.getBoundingClientRect(
```
bvaughn, line 121: Using the protocol client directly (rather than the Replay client) limits our ability to write new Playwright style tests for this stuff. That's probably _fine_ though, just maybe something it would be nice to eventually have the option to do :)

bvaughn (approved):
Just skimmed it but it looks good

src/ui/suspense/styleCaches.ts
markerikson: Hmm. Last time I tried adding more methods to `ReplayClient`, you'd said you were trying to keep that minimal, so I was under the impression I should avoid doing that.  Also since the existing code just called `client.DOM.whatever()` directly, I was copying that logic.

bvaughn: I think you're referring to different kinds of methods. `ReplayClient` exists to proxy calls to the Replay protocol (specifically because this allows us to mock out data for e2e tests).

It's not a great place to add utility functions.

PR #7801, opened 2022-09-23
Route browser auth through the standard devtools auth flow

ryanjduffy:
The current browser auth flow directs the user through the Auth0 login UI which can mislead some users to try to use email / password to login. Instead, route them through our UI so we can have a consistent experience both in the app and from the browser.

https://user-images.githubusercontent.com/788456/191889584-25db2016-949b-4032-b559-c57d0be6d6be.mp4


markerikson (approved):
I KNOW NOTHING ABOUT AUTH

so...

LGTM!

PR #7805, opened 2022-09-26
Add async frames when copying the stacktrace to the clipboard (FE-750)

hbenl:


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -102,14 +103,20 @@ function FramesRenderer({
...
+          copyStackTrace={copyStackTrace}
+        />
       </Suspense>
     </>
   );
 }
 
 export default function Frames({ panel }: { panel: "debugger" | "console" | "webconsole" }) {
-  const sourcesLoading = useAppSelector(getSourcesLoading);
+  const state = useAppSelector(state => state);
```
markerikson, line 118: :stop_sign: **DON'T DO THIS!**  It will cause the component to re-render on _every single dispatched action!**

A better option would be to put that logic into a thunk, which has access to `getState`

markerikson (changes requested):

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
hbenl: D'oh! And the `FramesRenderer` component made the same mistake...

markerikson (approved):

PR #7806, opened 2022-09-26
Don't suppress e2e test failures (FE-772)

hbenl:


bvaughn (approved):
Need to run `npm run lint` on this before landing.

bvaughn:
This should probably be reverted since it's breaking all of the open PRs.

bvaughn:
Reverted (for now) in #7811. We can land again once it doesn't clobber CI?

PR #7807, opened 2022-09-26
E2E test harness fixes

hbenl:

PR #7809, opened 2022-09-26
Rewrite markup/selection logic to remove remaining `NodeFront` usages

markerikson:
This PR:

#### KILLS OFF THE LAST REMAINING USAGES OF `NodeFront` YAYYYYY!!!!

- Completely replaces all remaining usages of the `selection` singleton in favor of just tracking a selected node ID (and "reason" string) in Redux, which we were already partly doing
  - Deletes the dead selection logic entirely
- Adds a new `nodeCaches` file that has a cache capable of handling several types of node-related queries ("node", "parentNodes", "childNodes", "searchDOM", "querySelector")
- Rewrites `actions/markup` to replace all remaining `NodeFront`-based data fetching with equivalent node object cache logic
- Rewrites the DOM search input from a vanilla JS class to a React function component, as this was doing some selection/`NodeFront` calls for queries
- Tweaks the types for our global `window.app.actions` to fix thunk usages
- Removes the remaining `NodeFront` usages from `test/harness.ts`

At this point, **THERE SHOULD BE NO MORE USAGES OF `NodeFront` IN THIS CODEBASE WOOHOO!!!!!**

I'm going to save deleting everything else for a follow-up PR as soon as we merge this one.


markerikson:
Looking at the test failures: Holger just merged #7806 , which apparently unveiled a bunch of failures we _were_ suppressing, and has #7807 in progress.

Given that, I don't _think_ all these failures are actually due to this PR, and I'll need to coordinate with Holger to check that.

Also, Trunk itself appears to be dying, due to a TS-internal error while running the `no-floating-promises` rule:

```
        exit_status: exited
        exit_code: 2
        stdout: (none)
        stderr: |
  
          Oops! Something went wrong! :(
  
          ESLint: 8.13.0
  
          TypeError: Cannot read properties of undefined (reading 'flags')
          Occurred while linting /home/runner/work/devtools/devtools/src/ui/setup/dynamic/devtools.ts:322
          Rule: "@typescript-eslint/no-floating-promises"
              at Object.getCheckFlags (/home/runner/work/devtools/devtools/node_modules/typescript/lib/typescript.js:19291:23)
              at getTypeOfSymbol (/home/runner/work/devtools/devtools/node_modules/typescript/lib/typescript.js:55857:33)
              at getParameterCount (/home/runner/work/devtools/devtools/node_modules/typescript/lib/typescript.js:76200:32)
              at combineIntersectionParameters (/home/runner/work/devtools/devtools/node_modules/typescript/lib/typescript.js:71921:29)
```

That's certainly not _good_, but I don't _think_ I've added a bunch more lint issues.

bvaughn:
We should revert the PR from Holger in the meanwhile

packages/bvaughn-architecture-demo/src/suspense/ObjectPreviews.ts
```diff
@@ -222,7 +222,7 @@ export async function getObjectPropertyHelper(
   }
 }
 
-export function preCacheObjects(pauseId: PauseId, objects: Object[]): void {
+export function preCacheObjects(pauseId: PauseId, objects: Object[] = []): void {
```
bvaughn, line 225: I don't think we should make this change.

Not that it would cause a bug, but it seems to hint at a conceptual problem. (What does it mean for code to be calling `preCacheObjects` without even passing a parameter for the objects that are being cached?)

src/devtools/client/debugger/src/components/shared/InspectorContextReduxAdapter.tsx
```diff
@@ -80,7 +80,7 @@ export default function InspectorContextReduxAdapter({ children }: { children: R
               objectId = object.preview?.node?.parentNode;
             }
 
-            dispatch(openNodeInInspector(valueFront));
+            // dispatch(openNodeInInspector(valueFront));
```
bvaughn, line 83: Did you mean to delete this or did you accidentally commit the commented out version?

packages/bvaughn-architecture-demo/src/suspense/ObjectPreviews.ts
markerikson: Yeah, I made this change because the standard `PauseData` type contains `{objects?: []}`, and I was getting annoyed at having to provide a default every time I called `preCacheObjects()`.  I can live with `preCacheObjects(pauseData.objects ?? [])`, I guess.

src/devtools/client/debugger/src/components/shared/InspectorContextReduxAdapter.tsx
markerikson: Intentional. This function isn't actually hooked up at all right now anyway, and this was the _only_ function that calls `openNodeInInspector`.  Given that, and that my immediate goal is to rip out the `NodeFront` stuff, I just commented this out to let me keep going with cleanup.

packages/bvaughn-architecture-demo/src/suspense/ObjectPreviews.ts
bvaughn: Wouldn't the more semantic thing to do be
```js
if (pauseData.objects) {
  preCacheObjects(pauseData.objects);
}
```
There's no point to make a new array _and_ call a method when there's no data to cache.

markerikson: True, that would also work.

markerikson: Reverted this and used `if` statements instead

src/devtools/client/debugger/src/components/shared/InspectorContextReduxAdapter.tsx
markerikson: Restored this since it led to a linter "don't need `[dispatch]` deps" warning

markerikson:
PR has been rebased to pick up all the E2E test changes. Also fixed a race condition where we might add child data for a parent that isn't fully loaded yet.

PR #7803, opened 2022-09-23
really fix print statement button flickering

souporserious:
This attempts to fix print statement button flickering by only rendering one element to the DOM and then translating it based on the current hovered line. I went down a small rabbit hole to get max performance by using refs and bypassing React, but it was making things more complicated so sticking with React state for now.

## Before

https://user-images.githubusercontent.com/2762082/192340437-74d08ca1-fab9-44de-ae8b-b46b948dc844.mov

## After

https://user-images.githubusercontent.com/2762082/192340471-28ef5695-19a1-42c8-b5ca-274d913de446.mov

## Future Considerations

The tooltip still sticks out as a little erratic when hovering lines. I think we can clean this up with a small overhaul to merge the print statement button, hit counts gutter, and tooltip into one state.

PR #7790, opened 2022-09-21
Refactor breakpoint/log point architecture

bvaughn:
This is a work-in-progress! Do not review it!

- [x] Delete all existing breakpoint and log point actions/state/reducers
- [x] Add new _point_ state+context (from bvaughn architecture package)
   - [x] Find breakable lines (and columns)
   - [x] Wire up hover to the right column (not just the first one)
   - [x] Show +/- and breakpoint toggle buttons
   - [x] Make points panel popup work
   - [x] Make sure breaking points get added to the Points panel
   - [x] Make sure logging points get added to the Console
- [x] Points should be stored/restored via local storage
- [x] Stress test to make sure large replays or big sources work
   - [x] Handle "_too many locations to perform analysis_"
- [x] Address any lingering `TODO [FE-757]` comments
- [ ] Add new Playwright tests

### Follow up work
Re-land #7803 and #7810 once this is merged.

PR #7810, opened 2022-09-26
enable empty line hover

souporserious:
Follow-up to #7803 to reenable all lines for hovering.


bvaughn:
Travis and I chatted out of band about this and decided it was easiest to revert this temporarily and re-land it once https://github.com/replayio/devtools/pull/7790 lands. That PR changes so many things about the Toggle Widget Button component that it's hard to apply these changes on top of that. I can own that change if you all are too busy or I can pair etc.

Reverting in #7812

PR #7816, opened 2022-09-27
Set up Playwright Testing Library

markerikson:
This PR:

- Adds https://github.com/testing-library/playwright-testing-library to let us use the Testing Library DOM query APIs with Playwright
- Updates all tests and helper methods to pass through `screen` instead of `page`, as the `Screen` type is a superset of the `Page` type that adds all the Testing Library queries (such as `screen.getByTestId()` )

### DIFF
--- a/packages/e2e-tests/helpers.ts
+++ b/packages/e2e-tests/helpers.ts
@@ -1,5 +1,16 @@
-import { Page, Locator, expect } from "@playwright/test";
+import { Page, Locator, expect, test as base, PlaywrightTestArgs } from "@playwright/test";
 import { assert } from "protocol/utils";
+
+import {
+  locatorFixtures as fixtures,
+  LocatorFixtures as TestingLibraryFixtures,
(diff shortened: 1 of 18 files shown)


packages/e2e-tests/helpers.ts
```diff
@@ -1,64 +1,75 @@
...
 
-export async function openPauseInformation(page: Page) {
+export async function openPauseInformation(screen: Screen) {
   // Only click if it's not already open; clicking again will collapse the side bar.
-  const pane = await getBreakpointsPane(page);
+  const pane = await getBreakpointsPane(screen);
   const isVisible = await pane.isVisible();
   if (!isVisible) {
-    return page.locator('[data-test-name="ToolbarButton-PauseInformation"]').click();
+    return screen.locator('[data-test-name="ToolbarButton-PauseInformation"]').click();
```
bvaughn, line 46: I guess there is no `getByTestName` equivalent?

PR #7819, opened 2022-09-27
Get test stepping-01 working

markerikson:


packages/e2e-tests/helpers.ts
```diff
@@ -77,17 +79,23 @@ export async function togglePausePane(screen: Screen) {
 
 // Console
 
+function delay(timeout: number) {
```
bvaughn, line 82: Ooo, what's this for? Hopefully we don't end up needing to use time-based delays very often. That seems like it could lead to flaky.

markerikson: Ah, that was mostly left over. I was trying to get Playwright to pause right after the "type stuff into the terminal" step long enough for me to confirm that it was, in fact, actually typing something into the terminal :)

PR #7827, opened 2022-09-28
Test suit bvaughn

bvaughn:


packages/e2e-tests/helpers.ts
```diff
@@ -315,11 +315,23 @@ export async function closeSource(screen: Screen, url: string) {
...
+  }
+
+  // Otherwise find it in the sources tree.
   const pane = await getSourcesPane(screen);
 
   let foundSource = false;
 
   while (true) {
+    // Wait for the panel's HTML to settle (to reflect the new toggled state.)
+    await pane.innerHTML();
```
bvaughn, line 333: This seems to have been the issue that was breaking Mark's tests. I guess we weren't waiting for the HTML to update between opening one source and the next 😅

PR #7828, opened 2022-09-28
fix object preview test

souporserious:
Cleans up the first object preview test. Brian Vaughn and I chatted whether these tests were worth having anymore since the new console tests cover most of these situations already. Something worth noting is the `executeInConsole` helper is quite similar to `checkEvaluateInTopFrame` so we may want to consolidate this after the dust settles.

### DIFF
--- a/packages/e2e-tests/helpers.ts
+++ b/packages/e2e-tests/helpers.ts
@@ -149,7 +149,11 @@ export async function warpToMessage(screen: Screen, text: string, line?: number)
 }
 
 export async function executeInConsole(screen: Screen, text: string) {
-  // TODO [PR 7550]
+  const consoleTextArea = screen.locator('[data-test-id="ConsoleRoot"] textarea');
+
+  await consoleTextArea.focus();
+  await consoleTextArea.fill(text);
+  await consoleTextArea.press("Enter");
 }
 
 export async function checkMessageObjectContents(
--- a/packages/e2e-tests/helpers.ts
+++ b/packages/e2e-tests/helpers.ts
@@ -1,4 +1,4 @@
-import { expect, test as base, PlaywrightTestArgs, Locator } from "@playwright/test";
+import { expect, test as base, PlaywrightTestArgs, Locator, ElementHandle } from "@playwright/test";
 import { waitFor } from "@playwright-testing-library/test";
 import {
   locatorFixtures as fixtures,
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/helpers.ts
```diff
@@ -149,12 +149,16 @@ export async function warpToMessage(screen: Screen, text: string, line?: number)
...
+
+  await consoleTextArea.focus();
+  await consoleTextArea.fill(text);
+  await consoleTextArea.press("Enter");
 }
 
 export async function checkMessageObjectContents(
   screen: Screen,
-  message: HTMLElement,
+  message: ElementHandle<HTMLElement | SVGElement>,
```
bvaughn, line 161: Interesting! 😮

packages/e2e-tests/tests/object_preview-01.test.ts
```diff
@@ -3,137 +3,80 @@ import {
...
-    screen,
-    `
-        function f() {
-          throw Error("there");
-        }
-        f();
-      `
-  );
+  await executeInConsole(screen, 'function f() { throw Error("there"); }()');
   // FIXME the first line in this stack isn't right.
```
bvaughn, line 47: Is this FIXME still relevant? seems like it isn't.

bvaughn (approved):

packages/e2e-tests/tests/object_preview-01.test.ts
```diff
@@ -3,137 +3,80 @@ import {
...
 
   await executeInConsole(screen, "new Map([[1, {a:1}], [2, {b:2}]])");
-  msg = await waitForConsoleMessage(screen, "Map { 1 → {…}, 2 → {…} }");
-  await checkMessageObjectContents(
-    screen,
-    msg,
-    ["0: 1 → Object { a: 1 }", "1: 2 → Object { b: 2 }", "size: 2"],
-    ["<entries>"]
-  );
+  await waitForConsoleMessage(screen, "Map(2) {1 → {…}, 2 → {…}}");
```
bvaughn, line 63: We could probably just do e.g. "Map(2)" for this type of object, but no big deal 😁

packages/e2e-tests/helpers.ts
souporserious: Ah, this was from me trying to fill out this helper it's what playwright returns from `waitForSelector`.

packages/e2e-tests/tests/object_preview-01.test.ts
souporserious: I wasn't sure, deleting as the output seems correct.

PR #7837, opened 2022-09-29
fix print statement panel width calculation

souporserious:
Fixes the print statement panel width calculation by updating the panel width when hit counts are collapsed and using the proper left offset based on the actual line counts gutter width rather than a fixed width.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
@@ -22,6 +22,7 @@ import { Nag } from "ui/hooks/users";
 import { MAX_POINTS_FOR_FULL_ANALYSIS } from "protocol/thread/analysis";
 import { getFocusRegion } from "ui/reducers/timeline";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
+import { useStringPref } from "ui/hooks/settings";
 import { UnsafeFocusRegion } from "ui/state/timeline";
 import { getHitPointsForLocation } from "bvaughn-architecture-demo/src/suspense/PointsCache";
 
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -131,6 +138,10 @@ function Panel({
     };
   }, [editor]);
 
+  useEffect(() => {
```
markerikson, line 141: :hand: This doesn't need to be in an effect - we should be able to just calculate it directly while rendering based on the known variables.

markerikson:
I put up a PR with an alternate approach in #7841 - thoughts?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
bvaughn: Actually, reading DOM layout (global mutable value) isn't really safe to do during render because it will produce unpredictable results for the given input (props) based on whatever the global mutable state happens to be.

Maybe we could do this using `useSyncExternalStore` though?

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -22,13 +22,19 @@ import { Nag } from "ui/hooks/users";
 import { MAX_POINTS_FOR_FULL_ANALYSIS } from "protocol/thread/analysis";
 import { getFocusRegion } from "ui/reducers/timeline";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
+import { useStringPref } from "ui/hooks/settings";
 import { UnsafeFocusRegion } from "ui/state/timeline";
 import { getHitPointsForLocation } from "bvaughn-architecture-demo/src/suspense/PointsCache";
 
-const gutterOffset = 37;
+const lineNumberGutterAndBorderWidth = 31;
```
bvaughn, line 29: This looks pretty sketchy.

souporserious: Yeah, I was more focused on an immediate fix 😇 these values should ideally come from system variables, but since it's not being used anywhere else I wanted to keep it local and simple for now.

souporserious:
fixed in #7841 😎

PR #7841, opened 2022-09-29
Base print statement width on CSS vars updated from CM element widths

markerikson:
An alternative to #7837 .

In this PR, I'm just listening for CM's `"refresh"` event to indicate a resize, reading the actual client widths of  the main CM editor element and the gutter element, subtracting them plus a bit of buffer, and setting that as a CSS var to be used as the print statement max width.

This _appears_ to work correctly as I drag-resize the editor, dock it into the "Console" tabs, or change the hit counts options.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
@@ -153,7 +153,15 @@ function Panel({
 
   return (
     <Widget location={breakpoint!.location} editor={editor} insertAt={insertAt}>
-      <div onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
+      <div
+        onMouseEnter={onMouseEnter}
+        onMouseLeave={onMouseLeave}
+        style={{
+          position: "sticky",
+          left: "var(--codemirror-gutter-width)",
+          maxWidth: "var(--print-statement-max-width)",
+        }}
+      >
         <FirstEditNag editing={editing} />
         <div className={classnames("breakpoint-panel", { editing })}>
           <div className="flex items-center space-x-0.5 pt-2 pl-1 pr-4">
--- a/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
@@ -24,10 +24,11 @@ import { getFocusRegion } from "ui/reducers/timeline";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import { UnsafeFocusRegion } from "ui/state/timeline";
 import { getHitPointsForLocation } from "bvaughn-architecture-demo/src/suspense/PointsCache";
+import type { SourceEditor } from "devtools/client/debugger/src/utils/editor/source-editor";
 
 const gutterOffset = 37;
 
-function getPanelWidth({ editor }: { editor: $FixTypeLater }) {
+function getPanelWidth({ editor }: SourceEditor) {
   return editor.getScrollInfo().clientWidth - gutterOffset;
 }
 
@@ -43,12 +44,11 @@ const connector = connect(
   }
 );
 
-type $FixTypeLater = any;
 type PropsFromRedux = ConnectedProps<typeof connector>;
 
 type ExternalProps = {
   breakpoint?: Breakpoint;
-  editor: $FixTypeLater;
+  editor: SourceEditor;
   insertAt: number;
 };
 
@@ -66,7 +66,6 @@ function Panel({
 }: PanelProps) {
   const [editing, setEditing] = useState(false);
   const [showCondition, setShowCondition] = useState(Boolean(breakpoint!.options.condition)); // nosemgrep
-  const [width, setWidth] = useState(getPanelWidth(editor)); // nosemgrep
   const [inputToFocus, setInputToFocus] = useState<"condition" | "logValue">("logValue");
   const dismissNag = hooks.useDismissNag();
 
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
```diff
@@ -26,33 +29,58 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
...
-  render() {
-    const { cx, editor, columnBreakpoints, selectedSource } = this.props;
-
-    if (!selectedSource || columnBreakpoints.length === 0) {
-      return null;
-    }
-
-    let breakpoints;
-    // TODO Is this a safe thing to do while rendering?
-    editor.codeMirror.operation(() => {
```
markerikson, line 42: :speech_balloon: I really have no idea why this `cm.operation()` call was here in the first place. We're mapping over data.

Only thing I can figure is that maybe at some point early on we were inserting widgets directly into CM?

markerikson:
Here's how it looks on Chrome Windows atm. Notice the 10px right buffer away from the scrollbar:

image

and here's how it is with Replay Firefox and that stupid giant HiDPI scrollbar problem. interestingly, it seems to be _flush_ with the scrollbar, but at least everything's visible:

image

also note that the second screenshot has the splitter dragged left enough that there's horizontal scrolling in the editor, and the print statements are still the correct size.

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -43,12 +44,11 @@ const connector = connect(
   }
 );
 
-type $FixTypeLater = any;
 type PropsFromRedux = ConnectedProps<typeof connector>;
 
 type ExternalProps = {
   breakpoint?: Breakpoint;
-  editor: $FixTypeLater;
+  editor: SourceEditor;
```
bvaughn, line 51: Nice

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
```diff
@@ -163,16 +153,12 @@ function Panel({
 
   return (
     <Widget location={breakpoint!.location} editor={editor} insertAt={insertAt}>
-      <div
-        onMouseEnter={onMouseEnter}
-        onMouseLeave={onMouseLeave}
-        style={{ position: "sticky", left: gutterOffset, maxWidth: width }}
```
bvaughn, line 169: Oh that's interesting, we can just remove the "stick" style without it changing anything?

src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
bvaughn: Yeah, I had the same thought (and removed it) as part of the points refactor branch.

src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
```diff
@@ -26,33 +29,58 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
...
+    };
+
+    editor.editor.on("refresh", updateWidth);
+
+    updateWidth();
+
+    return () => {
+      editor.editor.off("refresh", updateWidth);
+    };
+  }, [editor, hitCountsMode]);
```
bvaughn, line 66: This looks fine. We could also maybe do it via uSES too but no preference which we use really.
https://github.com/replayio/devtools/pull/7837#discussion_r983410954

src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
```diff
@@ -26,33 +29,58 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
...
+      const gutter = editor.editor.getGutterElement();
+
+      const gutterWidth = gutter.getBoundingClientRect().width;
+      const scrollWidth = editorElement.getBoundingClientRect().width;
+
+      // Magic formula to determine scrollbar width: https://stackoverflow.com/a/60008044/62937
+      let scrollbarWidth = window.innerWidth - document.body.clientWidth;
+
+      // Fill the _visible_ width of the editor, minus scrollbar, minus a small bit of padding
+      const newWidth = scrollWidth - gutterWidth - scrollbarWidth - 10;
```
bvaughn, line 53: Why is 10 a good amount of "padding" ? 😅

bvaughn (approved):

src/devtools/client/debugger/src/components/Editor/Breakpoints/Panel/Panel.tsx
souporserious: Yeah, this will break the current functionality for sticking to the left side if the editor scroll area overflows.

souporserious (approved):
This seems like a much more robust approach for handling the print statement panel maxWidth calculation! FWIW I was trying to get a quick fix out that didn't touch too much while we wait for Brian's points refactor to land. I'm fine with either solution, but please note we'll be removing the sticky functionality which isn't necessarily bad.

markerikson:
Fixed the horizontal sticky behavior by:

```ts
        style={{
          position: "sticky",
          left: "var(--codemirror-gutter-width)",
          maxWidth: "var(--print-statement-max-width)",
        }}
```

and then setting the gutter width variable in the same effect:

image

src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
```diff
@@ -26,33 +29,59 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
...
-      ));
-    });
-    return <div>{breakpoints}</div>;
+function ColumnBreakpoints2({ editor }: CBProps) {
+  const cx = useAppSelector(getContext);
+  const selectedSource = useAppSelector(getSelectedSource);
+  const columnBreakpoints = useAppSelector(getVisibleColumnBreakpoints);
+  const { value: hitCountsMode } = useStringPref("hitCounts");
+
+  useEffect(() => {
```
davilima6, line 41: Would `useLayoutEffect` fit?

markerikson: _Could_ certainly use that here, but I'm not sure there's any real benefit. Almost all of the CSS var updates will occur later as the CodeMirror editor is resized.  The only difference that would make here is right after the initial render, and after the "Show Hit Counts" option is changed.

bvaughn: Agreed with Mark.

Since we're calling `updateWidth` to check for any updates we may have missed between render and mount (something we should be doing even with layout effect) then this is fine. There's a potential for a small visible flicker if the timing is just right (wrong?)

PR #7842, opened 2022-09-29
Migrate inspector-01 and inspector-03 e2e tests

hbenl:


packages/e2e-tests/helpers.ts
```diff
@@ -790,3 +800,110 @@ export async function clickSourceTreeNode(screen: Screen, node: string) {
...
+          }
+
+          const selectorNodes = (
+            propertyNode.nextSibling as HTMLElement
+          ).querySelectorAll<HTMLElement>(".rule-text");
+          return [...selectorNodes].map(selectorNode => {
+            const selector = (selectorNode.children[0] as HTMLElement).innerText;
+            const value = (selectorNode.children[1] as HTMLElement).innerText;
+            const label = (selectorNode.previousSibling as HTMLElement).innerText;
+            const previousChild = (selectorNode.previousSibling as HTMLElement).children[0];
```
bvaughn, line 874: Some of this method, like this section with [0] and [1] children, looks kind of brittle. I wonder if there's a way to rewrite this that isn't so dependent on the _order_ of children (which could easily change)

bvaughn:
Going to land this so I can merge 7835 too and take a pass at cleaning up our helper functions.

packages/e2e-tests/helpers.ts
hbenl: I'm sure this could be improved. I'm also not 100% certain that having all this logic inside an `evalute()` is optimal. But it's good enough for now (and we want to migrate the tests ASAP).

PR #7844, opened 2022-09-29
Migrate remaining inspector tests

hbenl:


packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -9,78 +9,131 @@ export async function activateInspectorTool(screen: Screen): Promise<void> {
...
-        );
-      },
-      { style, value, matchedSelectors: expectedSelectors }
-    );
-    expect(result).toBe(true);
+      );
+      return { selector, source, properties };
+    });
   });
 }
```
bvaughn, line 89: These functions are all pretty scary... sibling nodes, parent nodes– all traversing the structure the DOM _happens to be_.

I wonder if we could write the helper functions without being so tightly coupled to the DOM structure?

packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -9,78 +9,131 @@ export async function activateInspectorTool(screen: Screen): Promise<void> {
   await screen.queryByTestId("PanelButton-inspector").click();
 }
 
-// TODO [FE-626] Rewrite this helper to reduce complexity.
+export async function checkAppliedRules(screen: Screen, expected: any) {
```
bvaughn, line 12: This should be `page: Page` now, btw.

PR #7851, opened 2022-09-29
update autocomplete tests

souporserious:
Updates the autocomplete tests to use Playwright.

I'm not sure if something is broken here since our old tests expected `objectProperties` for each check when comparing autocomplete matches, but currently we only show only some of the results which include `constructor` and not all of the object properties available on prototype.

### DIFF
--- a/packages/e2e-tests/helpers/console-panel.ts
+++ b/packages/e2e-tests/helpers/console-panel.ts
@@ -49,6 +49,8 @@ export async function enableConsoleMessageType(
 }
 
 export async function focusConsoleTextArea(page: Page) {
+  await openConsolePanel(page);
+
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/helpers/console-panel.ts
```diff
@@ -61,7 +57,19 @@ export async function executeTerminalExpression(page: Page, text: string): Promi
...
+
   await textArea.focus();
+
+  return textArea;
+}
+
+export async function executeTerminalExpression(page: Page, text: string): Promise<void> {
+  debugPrint(`Executing terminal expression "${chalk.bold(text)}"`, "executeTerminalExpression");
+
+  await openConsolePanel(page);
```
bvaughn, line 69: This should be in `focusConsoleTextArea` shouldn't it?

PR #7856, opened 2022-09-30
e2e tests

bvaughn:
- [x] Fixed log points 01 test by adding missing `addBreakpoint` statement
- [x] Finish object preview 03 test


packages/e2e-tests/tests/logpoints-01.test.ts
```diff
@@ -50,6 +51,5 @@ test(`log-points appear in the correct order and allow time warping`, async ({ p
   await executeAndVerifyTerminalExpression(page, "number", 5);
   await reverseStepOverToLine(page, 19);
 
-  // The log-point acts like a breakpoint when resuming.
```
bvaughn, line 53: No it doesn't. This comment wasn't accurate before (on the old test that I copied it over from).

PR #7861, opened 2022-09-30
Remove children before adding a new editor

jcmorrow:
This is stupid, but https://discord.com/channels/779097926135054346/930844769213640794/1025382501809664111 is a really, really annoying bug, so Travis and I figured we would propose *something*.


bvaughn (approved):
If it works, it works.

Hope we'll rip this code out "soon" anyway 😁

PR #7863, opened 2022-09-30
update highlighter tests

souporserious:
Updates the highlighter tests to use Playwright.

### DIFF
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -22,10 +22,6 @@ export async function openDevToolsTab(page: Page) {
   }
 }
 
-export async function selectNodePicker(page: Page) {
-  await page.locator('[data-test-id="NodePicker"]').click();
-}
-
 export async function openViewerTab(page: Page) {
   return page.locator('[data-test-id="ViewToggle-Viewer"]').click();
 }
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -22,6 +22,10 @@ export async function openDevToolsTab(page: Page) {
   }
 }
 
+export async function selectNodePicker(page: Page) {
+  await page.locator('[data-test-id="NodePicker"]').click();
+}
+
 export async function openViewerTab(page: Page) {
   return page.locator('[data-test-id="ViewToggle-Viewer"]').click();
 }
--- a/packages/e2e-tests/tests/highlighter.test.ts
+++ b/packages/e2e-tests/tests/highlighter.test.ts
@@ -1,4 +1,4 @@
-import test from "@playwright/test";
+import test, { expect } from "@playwright/test";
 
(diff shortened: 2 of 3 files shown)


souporserious:
@hbenl would love your help here on Monday. We're getting different coordinates than in the current highlighter test. After looking at your recent work for the react devtools tests, I'm not entirely sure how to map `getBoundingClientRect` in this specific test.

packages/e2e-tests/helpers/index.ts
```diff
@@ -14,6 +14,10 @@ export async function openDevToolsTab(page: Page) {
   }
 }
 
+export async function selectNodePicker(page: Page) {
+  await page.locator('[data-test-id="NodePicker"]').click();
+}
```
bvaughn, line 19: nit: `helpers/index` is kind of just for top-level things.

This helper function should probably go in `helpers/elements-panel`. There's already a bit of code in there that works with that button that you could also replace 😄 
```js
const pickerButton = page.locator('[title="Select an element in the video to inspect it"]')!;
await pickerButton.click();
```

packages/e2e-tests/tests/highlighter.test.ts
```diff
@@ -0,0 +1,30 @@
...
+  await openConsolePanel(page);
+  await warpToMessage(page, "ExampleFinished");
+
+  await selectNodePicker(page);
+
+  await page.waitForSelector(`[data-value="myiframe"]`);
+
+  const markupNode = await page.locator(`[data-value="myiframe"]`);
+
+  await markupNode.click();
```
bvaughn, line 19: You could replace these three lines using the helper function in `helpers/elements-panel`:
```suggestion
  await selectElementsRowWithText(page, "myiframe");
```

packages/e2e-tests/tests/highlighter.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const highlighter = page.locator("#box-model-content");
+
+  await highlighter.waitFor();
+
+  const pathDefinition = await highlighter.getAttribute("d");
+
+  const pathDefinitionToCompare = `M10,48.400001525878906 L310,48.400001525878906 L310,198.39999389648438 L10,198.39999389648438`;
```
bvaughn, line 27: I wonder if there's a way we could avoid hard-coding SVG coordinates in this test?

We don't have to figure that out now, since it was already like this. Just thinking out loud.

packages/e2e-tests/helpers/index.ts
souporserious: Cleaned up! ✨

packages/e2e-tests/tests/highlighter.test.ts
souporserious: Yeah, Holger has a better approach here that he used for the react inspector we should be able to copy here in a follow-up. Left a TODO comment with a ticket attached 😎

bvaughn (approved):

PR #7870, opened 2022-10-01
Close console drawer by default

jcmorrow:
But add an experimental pref for defaulting to open, and either way, remember the previous state on a per-Replay basis.

Fixes https://linear.app/replay/issue/FE-797/the-console-drawer-should-be-closed-by-default

### DIFF
--- a/packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
@@ -8,6 +8,7 @@ import {
   unstable_Offscreen as Offscreen,
   useContext,
   useRef,
+  useLayoutEffect,
   useState,
 } from "react";
 import { TerminalContext } from "@bvaughn/src/contexts/TerminalContext";
@@ -43,8 +44,16 @@ export default function ConsoleRoot({
     `Replay:Console:MenuOpen:${recordingId}`,
     false
   );
+  const [menuValueHasBeenToggled, setMenuValueHasBeenToggled] = useState(false);
 
   const messageListRef = useRef<HTMLElement>(null);
+
+  // We default to having the console filters panel turned off, to minimize UI "busyness".
+  // _If_ it's off initially, we want to completely skip rendering it, which
+  // avoids making the "fetch events" calls during app startup to speed up loading.
+  // But, if it's ever been shown and toggled off, continue rendering it
+  // inside the `<Offscreen>` to preserve state.
+  const renderFilters = isMenuOpen || menuValueHasBeenToggled;
 
   return (
     <ConsoleContextMenuContextRoot>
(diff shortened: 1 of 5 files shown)


jonbell-lot23 (approved):
Looks fantastic. Our NUX metrics will thank you :)

markerikson:
I like it! Merge it now, or should we also add code to delay fetching event types?

jcmorrow:
> I like it! Merge it now, or should we also add code to delay fetching event types?

Oh, I actually thought that this was going to do that automatically, but it does not!! Let me see if that's an easy add-on.

markerikson:
Josh's updated commit skips rendering `<Filters>` at all, _any_ time the `isMenuOpen` flag is false.

But, that loses the value of the `<Offscreen>` wrapper.

What we really want is to skip rendering `<Filters>` _if the initial value is false_.  If the initial value is true, or the user has _toggled_ this at all, we still want to render it even if hidden.  

I've force-pushed an alternate commit that should do that.

jcmorrow:
Thanks Mark!

packages/bvaughn-architecture-demo/components/console/ConsoleRoot.tsx
```diff
@@ -23,24 +23,39 @@ import { LoggablesContextRoot } from "./LoggablesContext";
...
   const { clearMessages: clearConsoleEvaluations, messages: consoleEvaluations } =
     useContext(TerminalContext);
 
-  const [isMenuOpen, setIsMenuOpen] = useLocalStorage<boolean>("Replay:Console:MenuOpen", true);
+  const { recordingId } = useContext(SessionContext);
+  const [isMenuOpen, setIsMenuOpen] = useLocalStorage<boolean>(
+    `Replay:Console:MenuOpen:${recordingId}`,
+    showFiltersByDefault
+  );
+  const [menuValueHasBeenToggled, setMenuValueHasBeenToggled] = useState(false);
```
jcmorrow, line 48: I would argue that you could change this to `const [filtersHaveBeenShown, setFiltersHaveBeenShown] = useState(showFiltersByDefault)` and skip the `||` down on line 57.

jcmorrow: Eh, actually, this is fine 🙂  I considered proposing this change:

```
  const filtersHaveBeenRendered: boolean = useMemo(() => {
    return !!(isMenuOpen || filtersHaveBeenRendered);
  }, [isMenuOpen]);
```

Just because I feel like the `useState` hook introduces more cognitive load than the `useMemo` hook, but I think it's mostly just a matter of taste, and that's not a great reason to go around changing working code!

bvaughn:
I think this change broke most of the screenshot tests for the new console

PR #7879, opened 2022-10-03
Port remaining node tests

markerikson:
- Ported `node_worker`, `node_stepping`, and `node_object_preview`
- Rewrote the "find visible lines" logic to use actual DOM bounding boxes on the target side

### DIFF
--- a/packages/e2e-tests/helpers/console-panel.ts
+++ b/packages/e2e-tests/helpers/console-panel.ts
@@ -123,8 +123,9 @@ export function findConsoleMessage(
     ? `[data-test-message-type="${messageType}"]`
     : '[data-test-name="Message"]';
 
+  // Use single quotes because a couple messages have double quoted text displayed
   return expected
-    ? page.locator(`${attributeSelector}:has-text("${expected}")`)
+    ? page.locator(`${attributeSelector}:has-text('${expected}')`)
     : page.locator(`${attributeSelector}`);
 }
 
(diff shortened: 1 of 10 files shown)


packages/e2e-tests/helpers/console-panel.ts
```diff
@@ -123,8 +123,9 @@ export function findConsoleMessage(
     ? `[data-test-message-type="${messageType}"]`
     : '[data-test-name="Message"]';
 
+  // Use single quotes because a couple messages have double quoted text displayed
```
bvaughn, line 126: Wonder if we should throw if you pass a string with double quotes? (Or we could switch outer quote styles, depending?)

packages/e2e-tests/helpers/source-panel.ts
```diff
@@ -4,7 +4,7 @@ import chalk from "chalk";
 import { openDevToolsTab } from ".";
 import { openPauseInformationPanel } from "./pause-information-panel";
 import { openSource, openSourceExplorerPanel } from "./source-explorer-panel";
-import { clearTextArea, debugPrint, delay, forEach, waitFor } from "./utils";
+import { clearTextArea, debugPrint, delay, forEach, waitFor, mapLocators } from "./utils";
```
bvaughn, line 7: You don't actually use the new `mapLocators` function

bvaughn (approved):

packages/e2e-tests/helpers/source-panel.ts
markerikson: Yeah, I started to import it here, then realized that the logic that uses `Array.from()` in this file is subtly different - it's mapping over a subset of the locators, not all of them.  I can remove the import.

markerikson: On the other hand I could use that and slice later. That would work too.

PR #7892, opened 2022-10-03
Shard Playwright tests and use multiple workers

markerikson:


packages/e2e-tests/playwright.config.ts
```diff
@@ -14,6 +14,8 @@ const config: PlaywrightTestConfig = {
       height: 1024,
     },
   },
+  // Limit the number of workers on CI, use default locally
+  workers: CI ? 2 : undefined,
```
bvaughn, line 18: Stupid question maybe but why are we overriding and limiting to 2 for CI?

PR #7893, opened 2022-10-03
Add a little resilience to failed async frame protocol commands

ryanjduffy:
This is a precursor to proper error handling of these `getFrameSteps` calls which can fail with "too many points" errors. In testing this on the customer replay that was failing, this change at least allowed the call stack to show up and the stepper to work most of the time. There were a couple instances where they didn't but I'm not sure why.

TBH, I don't know the implications of "ignoring" these errors in devtools so I'm opening this as draft for feedback from those that know it better (cc: @markerikson , @bvaughn , @hbenl )

A replay of the "fix" in action:
https://app.replay.io/recording/replay-of-localhost8080--c4d8afe5-68b0-4cf9-8852-507a3dcfcb62

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/selectFrame.ts
+++ b/src/devtools/client/debugger/src/actions/pause/selectFrame.ts
@@ -9,6 +9,7 @@ import { selectLocation } from "../sources";
 import { fetchScopes, frameSelected } from "../../reducers/pause";
 import { setFramePositions } from "./setFramePositions";
 import { SourceLocation } from "../../reducers/types";
+import { isCommandError, ProtocolError } from "shared/utils/error";
 
 interface TempFrame {
   id: string;
@@ -32,7 +33,16 @@ export function selectFrame(cx: Context, frame: TempFrame): UIThunkAction {
 
     dispatch(frameSelected({ cx, frameId: frame.id }));
 
-    await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    try {
+      await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    } catch (e) {
+      // TODO [FE-795]: Communicate this to the user
+      if (isCommandError(e, ProtocolError.TooManyPoints)) {
+        console.error(e);
+        return;
+      }
+      throw e;
+    }
 
     dispatch(selectLocation(cx, frame.location));
     dispatch(setFramePositions());
--- a/src/ui/suspense/util.ts
+++ b/src/ui/suspense/util.ts
@@ -1,5 +1,6 @@
 import { Frame, PauseId } from "@replayio/protocol";
 import { ThreadFront } from "protocol/thread";
+import { isCommandError, ProtocolError } from "shared/utils/error";
 import { getFramesSuspense } from "./frameCache";
 import { getFrameStepsSuspense } from "./frameStepsCache";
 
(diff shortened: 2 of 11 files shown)


src/ui/suspense/util.ts
```diff
@@ -39,16 +39,22 @@ export function getAsyncParentFramesSuspense(
...
+    }
+    const frames = getFramesSuspense(asyncParentPauseId);
+    if (!frames) {
+      return;
+    }
+    return frames.slice(1);
+  } catch (e) {
+    // TODO [FE-795]: Communicate this to the user
+    console.error(e);
+    return [];
```
bvaughn, line 58: Looks like this would break the whole _suspense_ part of this function (which relies on throwing a promise-like object to signal data being loaded).

At a minimum you'd probably want to do something like this in the `catch`:
```js
} catch (errorOrPromise) {
  if (isThennable(errorOrPromise)) {
    throw errorOrPromise;
  } else {
    // ...
  }
}
```

but truthfully, I'm initially uneasy about this change to swallow errors.

jcmorrow: It would probably be best to have some kind of `Result` data type for places where we expect fetching might fail. Otherwise it's difficult to distinguish between "Fetch succeeded, no results" and "Fetch errored", which often leads to the user being stranded as to what to do next. Curious to hear others thoughts on what that might look like though!

ryanjduffy: Thanks for the feedback on the suspense side. That makes sense.

> but truthfully, I'm initially uneasy about this change to swallow errors.

Agree that it's far from ideal. This could be more surgical and only swallow the "too many points" error. Ultimately, this PR is trying to be a stopgap until we handle these error situations properly since this issue is blocking a customer.

jcmorrow:
I pushed a commit that I think might be a better short-term fix, lmk what y'all think!

src/devtools/client/debugger/src/actions/pause/selectFrame.ts
```diff
@@ -32,7 +33,16 @@ export function selectFrame(cx: Context, frame: TempFrame): UIThunkAction {
 
     dispatch(frameSelected({ cx, frameId: frame.id }));
 
-    await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    try {
+      await ThreadFront.getFrameSteps(frame.asyncIndex, frame.protocolId);
+    } catch (e: any) {
```
jcmorrow, line 38: is it better practice to do like `any | undefined | null` here and do a `e?.code` down below? I never know how to handle errors with typescript.

bvaughn: Technically you can throw `null` in JavaScript, so it's probably _strictly_ best to let TypeScript tell you what the default type of `error` is and follow its rules.

bvaughn (approved):
I'm worried about this change but I feel like there's a lot of pressure to land it.

I think I'm mostly worried about `selectFrame` and `setFramePositions` since these will fail silently and potentially make for a confusing UX. (I think the fallback of an empty array for `setFramePositions` seems pretty reasonable.)

src/devtools/client/debugger/src/actions/pause/selectFrame.ts
```diff
@@ -32,7 +33,16 @@ export function selectFrame(cx: Context, frame: TempFrame): UIThunkAction {
...
+      // TODO [FE-795]: Communicate this to the user
+      if (e.code === ProtocolError.TooManyPoints) {
+        console.error(e);
```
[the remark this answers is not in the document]
jcmorrow: Oh, no particular reason, I was just shuffling the original code around. I will add this elsewhere, it seems like a helpful thing to have.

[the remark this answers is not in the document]
bvaughn: Sure, agreed! I should have been more clear: We should probably _also_ log errors in `setFramePositions` since otherwise they would just fail silently.

src/devtools/client/debugger/src/actions/pause/selectFrame.ts
loganfsmyth: `unknown` is always preferable for types on bindings I think.

In my opinion the best way would be to declare a helper, e.g.
```
export function isProtocolError(e: unknown, code: ProtocolError): boolean {
  return e instanceof Error && e.name === "CommandError" && (e as any).code === code;
}
```
then do
```
} catch (e: unknown) {
  if (isProtocolError(e, ProtocolError.TooManyPoints)) {
```

loganfsmyth: An error subclass would be an option too, but I sort of forget the state of extending the native Error type in browsers, I think it works?

jcmorrow: haha, I literally just wrote this code! Good, I'm glad we're in agreement!

src/devtools/client/debugger/src/actions/pause/selectFrame.ts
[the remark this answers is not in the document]
jcmorrow: Done!

ryanjduffy:
@jcmorrow - should I move this to ready for review?

jcmorrow:
@ryanjduffy I think this is actually ready to go in, I just got distracted yesterday. Mind doing the rebase & merge?

jcmorrow (approved):

PR #7898, opened 2022-10-04
Fix the logpoints-04 test

hbenl:
The new version of the test contained an additional jump to a different console message compared with the old version.
Removing that additional jump made the test work for me.

### DIFF
--- a/packages/e2e-tests/tests/logpoints-04.test.ts
+++ b/packages/e2e-tests/tests/logpoints-04.test.ts
@@ -28,8 +28,6 @@ test(`should display exceptions in the console`, async ({ page }) => {
   await seekToConsoleMessage(page, messages.first());
   await waitForFrameTimeline(page, "100%");
 
-  messages = await findConsoleMessage(page, "number: 10", "exception");
-  await seekToConsoleMessage(page, messages.first());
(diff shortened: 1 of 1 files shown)


packages/e2e-tests/tests/logpoints-04.test.ts
```diff
@@ -28,8 +28,6 @@ test(`should display exceptions in the console`, async ({ page }) => {
   await seekToConsoleMessage(page, messages.first());
   await waitForFrameTimeline(page, "100%");
 
-  messages = await findConsoleMessage(page, "number: 10", "exception");
-  await seekToConsoleMessage(page, messages.first());
```
bvaughn, line 32: Ooo, nice catch! Although I _am_ curious why this reliably worked locally but not on CI.

bvaughn (approved):

packages/e2e-tests/tests/logpoints-04.test.ts
hbenl: My guess would be that we had a bug in our test helper so that `await seekToConsoleMessage(...)` didn't really wait for the app to jump to the new point and so the evaluation was done on the previous point.

PR #7907, opened 2022-10-04
add new useDismissNag hook

souporserious:
There was a recent regression when we merged the new console where the new user nag for navigating to the console never gets dismissed. This adds a new dismiss nag graphql mutation and `useDismissNag` hook for clearing these in the new architecture.

closes SCS-148

## Follow-up

Currently the nag does not disappear when navigating because the GraphQL mutation does not trigger a re-render for some reason. (tracking in FE-822)

### DIFF
--- a/packages/bvaughn-architecture-demo/src/hooks/useDismissNag.ts
+++ b/packages/bvaughn-architecture-demo/src/hooks/useDismissNag.ts
@@ -1,0 +1,25 @@
+import { useContext } from "react";
+
+import { GraphQLClientContext } from "../contexts/GraphQLClientContext";
+import { SessionContext } from "../contexts/SessionContext";
+import { Nag } from "../graphql/types";
+import { dismissNag } from "../graphql/User";
+
+export function useDismissNag() {
+  const { accessToken, currentUserInfo } = useContext(SessionContext);
+  const graphQLClient = useContext(GraphQLClientContext);
+
+  return (nag: Nag) => {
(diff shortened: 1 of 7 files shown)


packages/bvaughn-architecture-demo/src/hooks/useDismissNag.ts
```diff
@@ -0,0 +1,25 @@
...
+      return;
+    }
+
+    const { id, nags } = currentUserInfo;
+
+    if (!nags || nags.includes(nag) || !id) {
+      return;
+    }
+
+    dismissNag(graphQLClient, accessToken, nag);
```
bvaughn, line 23: I think _maybe_ the reason this works differently for the old console is that the mutation was configured to also refetch user info:
```js
const [dismissNag, { error }] = useMutation<DismissNag, DismissNagVariables>(DISMISS_NAG, {
  refetchQueries: ["GetUser"],
});
```

The new architecture stuff doesn't use Apollo, just plain GraphQL, so there's nothing that tells the legacy app (with Apollo) that it needs to refresh user info.

We can address this as part of the follow up issue though! 😄

bvaughn (approved):

packages/bvaughn-architecture-demo/src/hooks/useDismissNag.ts
souporserious: Oh gosh, that makes sense why it's not updating 🤦 sounds good!

PR #7912, opened 2022-10-05
Use the new Frames component in the NetworkMonitor

hbenl:
Preparation for the removal of the old Frames and Scopes components.

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
@@ -2,7 +2,6 @@ import { Suspense, useMemo } from "react";
 import { PauseId } from "@replayio/protocol";
 import { ThreadFront } from "protocol/thread";
 import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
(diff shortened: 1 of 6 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -31,38 +30,16 @@ function FramesRenderer({
...
 }: {
-  panel: "debugger" | "console" | "webconsole";
+  panel: CommonFrameComponentProps["panel"];
   pauseId: PauseId;
   asyncIndex?: number;
-  copyStackTrace: () => void;
+  getFullStackTrace: () => PauseFrame[];
 }) {
-  const cx = useAppSelector(getThreadContext);
+  const commonProps = useGetCommonFrameComponentProps(panel, getFullStackTrace);
```
bvaughn, line 40: Why do these props need to be fetched outside and passed into `PauseFrames` ? I think the API would be a bit simpler if `PauseFrames` required a `panel: CommonFrameComponentProps["panel"]` prop and fetched them itself.

bvaughn (approved):

PR #7914, opened 2022-10-05
Remove mouse targets from `Thread/Pause`

markerikson:
This PR:

- Rewrites the calls to `ThreadFront/Pause.loadMouseTargets()` to use a new bounding rects cache instead
- Deletes the legacy E2E test harness file! (and also the code that loaded the legacy E2E tests into the app for execution)


test/mock/scripts/unload-recording-01.ts
```diff
@@ -15,6 +14,49 @@ import {
 import { basicMessageHandlers, basicBindings } from "../src/handlers";
 import { Page } from "@recordreplay/playwright";
 
+export function waitForTime(ms: number, waitingFor?: string) {
```
markerikson, line 17: :speech_balloon: Moved these from the old `harness.ts` file, since this was the only other place that needed them

src/ui/actions/app.ts
```diff
@@ -189,12 +177,26 @@ export function setCanvas(canvas: Canvas): UIThunkAction {
...
+    const sessionId = state.app.sessionId;
+
+    if (!sessionId || !pauseId) {
+      return;
+    }
+
+    return getBoundingRectsAsync(protocolClient, sessionId, pauseId);
+  };
+}
+
```
bvaughn, line 193: I don't have the best understanding of the rules of what makes a Redux action or not, but this doesn't seem like it should be an "action" to me. It doesn't dispatch anything. It just uses the `getState` callback to run another async function (`getBoundingRectsAsync`).

I guess I would have expected this to be a custom hook or something.
```js
function useGetBoundingRectsAsync() {
  const pauseId = useAppSelector(getPauseId);
  const sessionId = useAppSelector(getSessionId);

  return useCallback(() => {
    if (!sessionId || !pauseId) {
      return null;
    }

    return getBoundingRectsAsync(protocolClient, sessionId, pauseId);
  }, [sessionId, pauseId]);
}
```

I'm curious about your thinking here!

bvaughn (approved):

markerikson:
Looks like there was a spurious failure but it succeeded on retry.

PR #7916, opened 2022-10-05
show console nav by default for new users

souporserious:
As an effort to get new users to interact with the console this PR adds an initial hovered state to the first console rewind/forward button until the user has successfully navigated through the console.

closes des-111

### DIFF
--- a/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
@@ -91,6 +91,8 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
       listItems.push(currentTimeIndicator);
     }
 
+    const isLastMessageInRange = index === loggables.length - 1;
+
     const loggableExecutionPoint = getLoggableExecutionPoint(loggable);
     const isLoaded =
       loadedRegions !== null && isPointInRegions(loggableExecutionPoint, loadedRegions.loaded);
@@ -112,6 +114,7 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
             index={index}
             isFocused={loggable === currentSearchResult}
             logPointInstance={loggable}
+            isLastMessageInRange={isLastMessageInRange}
           />
         );
       } else if (isProtocolMessage(loggable)) {
@@ -121,6 +124,7 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
             index={index}
             isFocused={loggable === currentSearchResult}
             message={loggable}
+            isLastMessageInRange={isLastMessageInRange}
           />
         );
       } else if (isTerminalExpression(loggable)) {
--- a/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
@@ -10,7 +10,7 @@ import {
   isTerminalExpression,
   isUncaughtException,
 } from "@bvaughn/src/utils/loggables";
-import { isExecutionPointsLessThan } from "@bvaughn/src/utils/time";
+import { isExecutionPointsLessThan, isExecutionPointsWithinRange } from "@bvaughn/src/utils/time";
 import {
   ForwardedRef,
   forwardRef,
@@ -47,6 +47,7 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
   const replayClient = useContext(ReplayClientContext);
   const [searchState] = useContext(ConsoleSearchContext);
   const { executionPoint: currentExecutionPoint } = useContext(TimelineContext);
+
   const loadedRegions = useLoadedRegions(replayClient);
 
   // The Console should render a line indicating the current execution point.
--- a/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
@@ -1,8 +1,6 @@
 import Icon from "@bvaughn/components/Icon";
 import { FocusContext } from "@bvaughn/src/contexts/FocusContext";
-import { SessionContext } from "@bvaughn/src/contexts/SessionContext";
 import { TimelineContext } from "@bvaughn/src/contexts/TimelineContext";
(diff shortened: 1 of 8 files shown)


bvaughn (changes requested):
I'd like some time to think about this change, but I'm swamped by bugs – so I'm just planting a flag here to hold off merging for a bit 🙇🏼

bvaughn:
Thanks! :)

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -86,6 +88,8 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
     const loggableExecutionPoint = getLoggableExecutionPoint(loggable);
     const isLoaded =
       loadedRegions !== null && isPointInRegions(loggableExecutionPoint, loadedRegions.loaded);
+    const includesFirstConsoleNag = currentUserInfo?.nags?.includes(Nag.FIRST_CONSOLE_NAVIGATE);
+    const initialIsHovered = Boolean(!includesFirstConsoleNag && index === 0);
```
bvaughn, line 92: Hmm. Item renderers already have an `index` prop, so a renderer can already know if it's the first item in the list. (Admittedly this wouldn't account for the case where you filter items in the list such that the actual first one doesn't get shown– but I don't think this PR is account for that case either? and I don't think we need to, to be honest.)

I wonder if renderers should check this condition directly?

```js
const showInitialConsoleHoverPrompt = currentUserInfo.onboarding.showInitialConsoleHoverPrompt;
const [isHovered, setIsHovered] = useState(showInitialConsoleHoverPrompt);
```

I don't think _all_ renderer types would need this check either. Maybe just `LogPointRenderer`?

If it's the first time they're using Replay, they won't have added a terminal expression or a log point yet– so those renderers seem unnecessary.

Event logs or uncaught exception logs renderers _could_ check this, although those require the user to actively ask to see the data– so maybe they wouldn't even need to? On the fence about it.

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
```diff
@@ -86,6 +88,8 @@ function MessagesList({ forwardedRef }: { forwardedRef: ForwardedRef<HTMLElement
     const loggableExecutionPoint = getLoggableExecutionPoint(loggable);
     const isLoaded =
       loadedRegions !== null && isPointInRegions(loggableExecutionPoint, loadedRegions.loaded);
+    const includesFirstConsoleNag = currentUserInfo?.nags?.includes(Nag.FIRST_CONSOLE_NAVIGATE);
```
bvaughn, line 91: Something about the way we model this stuff irks me.

Maybe a small part of it is that we call these things "nags" – which is an inherently negative term.

I think the bigger part though is the way we expose them as an array of strings that we do an `includes` check on. Even if that's how we model things in GraphQL, I think we could expose it to our views with a better API (e.g. `currentUserInfo.onboarding.showInitialConsoleHoverPrompt`)

bvaughn (commented):
Sorry for taking so long to review this.

This approach feels pretty heavy handed to me– that we'd add a _new property_ to every type of item renderer to handle this edge case.

Thinking about other was we might accomplish this... left some comments. What are your thoughts?

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
souporserious: Ah, yeah I think this was overkill adding to each renderer! I cleaned this up to only change the initial hover state of the `LogPointRenderer`.

Thinking through this more I think it will need to be a little smarter since it's checking purely index right now, but should be whatever the _first_ log point is. Assuming that shouldn't be hard to calculate though.

packages/bvaughn-architecture-demo/components/console/MessagesList.tsx
souporserious: Totally, I was going to copy over the utility we have right now for this so it's more explicit, but is probably worth investing whatever renaming we want to do now before this starts getting added to new code.

souporserious:
@bvaughn updated to only modify `LogPointRender` default hover state which is _much_ simpler now. Jon should be able to help with checking the onboarding flow here if needed. You'll need to reset your account state to get the nags to show up. We have a guide for this in Notion if you search for "onboarding audit".

bvaughn:
Cool beans. This is on my plate to land right after the points stuff. Thanks!

markerikson:
Looking at this, I've got a couple behavioral questions:

- the original design issue said we want the _last_ message to have this displayed, but the current implementation in the PR tries to do it for the _first_ message. Which is it?
- The PR implementation only does it for `<LogPointRenderer>`, but I think the intent is for it to be _any_ message.  Would like confirmation on that

From a technical perspective:

- Not immediately sure how to do the "hide when hovering others" behavior, unless we start throwing another context above the entire messages list and track "are we hovering anything in the list" to partially override the per-message-item hover behavior

Going to wait until I can talk with Jon about what the intended behavior is to tackle this one further

markerikson:
Jon's last direction in Linear was:

> - We'll show the bottom message by default. The goal is to always have something in view, so the top message might be scrolled out of view.
> - Yup, first/last message of any type. We just want people to understand that they can navigate without first having to hover to see the arrow.

We discussed scope and feasibility of implementation today:

> - Add this "default hovered on" behavior for the last message of type message or logpoint that is being rendered inside the current focus region
> - _Not_ try to hide that force-shown "Rewind" button if you start trying to hover over other messages in the console.  So, if you were to hover over message 3 in the list, you would end up with the normal hover-caused button on message 3 plus the force-shown button on the last message.  Trying to track "are we hovering anything?" and turning off the force-shown button would add too much complexity.

I've updated the PR accordingly.  That also includes scrolling to that last message if we're forcing display of the hover button (which overrides the "scroll to current timestamp marker" behavior).

markerikson:
update:

Jon and I realized after discussion that the "force-enable the rewind hover button" approach is really not ideal visually. Instead, we think an overlay over the entire console messages list (ie, the classic tutorial UI "SEE THIS THING HERE ---->" approach) would be better visually _and_ simpler to implement. Jon's going back to the drawing board and sketch that out

bvaughn:
Cleaning up old/sale PRs. Closing this one as it seems abandoned. We can re-open and review if you decide to pick it back up, though 😄

PR #7921, opened 2022-10-06
Don't set breakpoints on lines with no breakpoints

jcmorrow:
Right now you can click on the gutter to set a breakpoint on *any* line, not just lines with breakable points. It looks silly and it causes a backend error.


markerikson (approved):

src/devtools/client/debugger/src/actions/breakpoints/modify.ts
```diff
@@ -103,10 +103,14 @@ export function addBreakpoint(
     if (!column) {
       const lineBreakpoints = getPossibleBreakpointsForSource(getState(), sourceId);
       if (!lineBreakpoints) {
-        console.debug(`No possible breakpoints on line ${line} of source ${sourceId}`);
+        console.debug(`No possible breakpoints for source ${sourceId}`);
         return;
       }
       column = Math.min(...lineBreakpoints.filter(bp => bp.line === line).map(bp => bp.column));
+      if (!column || column === Infinity) {
```
bvaughn, line 110: I think this caused a regression (see FE-828)

`column === 0` will fail this

jcmorrow: Yep, shit, sorry, thanks Brian!

bvaughn: No worries! 😆 This one gets me all the time.

PR #7922, opened 2022-10-06
Don't mutate read-only object

jcmorrow:
Fixup to #7920


jcmorrow:
I'm seeing a few instances of this in Sentry, so going to push this fix.

PR #7925, opened 2022-10-06
Fix Console errors

bvaughn:
- Add handling for string literal "too many points" error returned by Analysis API (see BAC-2330)
  - Added tests
- Change sort assertions to only run in DEV and to sort by points rather than time (more accurate) to account for a bad recording (see BAC-2339)

### DIFF
--- a/src/ui/utils/timeline.ts
+++ b/src/ui/utils/timeline.ts
@@ -314,7 +314,9 @@ export function filterToFocusRegion<T extends TimeStampedPoint>(
 function assertSorted(a: TimeStampedPoint[]) {
   a.reduce(
     (prev, curr) => {
-      assert(prev.time <= curr.time);
+      // It would be cheaper to compare time but at least one recording has a bizarre time (see BAC-2339).
+      // Truthfully, since points are more fine-grained than time, we should be comparing them anyway.
+      assert(BigInt(prev.point) <= BigInt(curr.point));
       return curr;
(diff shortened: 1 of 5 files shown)


src/ui/utils/timeline.ts
```diff
@@ -314,7 +314,9 @@ export function filterToFocusRegion<T extends TimeStampedPoint>(
 function assertSorted(a: TimeStampedPoint[]) {
   a.reduce(
     (prev, curr) => {
-      assert(prev.time <= curr.time);
+      // It would be cheaper to compare time but at least one recording has a bizarre time (see BAC-2339).
+      // Truthfully, since points are more fine-grained than time, we should be comparing them anyway.
+      assert(BigInt(prev.point) <= BigInt(curr.point));
```
bvaughn, line 319: cc @jcmorrow

jcmorrow: Just to be clear: every time we see this it's a bug. It non-monotonic time will never be expected behavior.

bvaughn: @jcmorrow That makes sense. I filed a backend bug for this.
I think arguably we should still be comparing _points_ though (since they are finer grained than time, aka there might be duplicate times with different points)

We could compare _both_ but I can't uncomment this until the backend is fixed or the frontend will get wonky again.

PR #7930, opened 2022-10-06
No longer fetch frame steps while selecting a frame (fixes fe-768)

jasonLaster:
This feels like a regression because we're not using the results of `getFrameSteps` in `selectFrame` and there is no reason to block on `selectLocation`

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/selectFrame.ts
+++ b/src/devtools/client/debugger/src/actions/pause/selectFrame.ts
@@ -33,17 +33,6 @@ export function selectFrame(cx: Context, frame: TempFrame): UIThunkAction {
 
     dispatch(frameSelected({ cx, frameId: frame.id }));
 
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):
I don't have much context on this to know if this is safe or not. You're right that we don't use the data but I don't know if that means the call isn't necessary.

So, I'm approving mostly because I trust you know more about this than I and if it doesn't work, it's easy to revert.

jcmorrow:
Yeah, wow, I was curious and I traced that call all the way back to https://github.com/replayio/devtools/blame/ed24f311e1681734e1add519a5c6b319895b51cb/src/devtools/client/debugger/src/actions/pause/selectFrame.js#L39, which was just basically porting over a bunch of FF code I think? So, seems reasonable that it is probably not doing anything these days given how different the app is now!

PR #7932, opened 2022-10-07
Editor should scroll to line when comment is clicked

bvaughn:
Work in progress.

- [x] Rewrite main Editor component from class to function component
- [x] Rewrite `HighlightLine`, `HighlightLines`, and `LineHitCounts` to hooks (`useHighlightedLined` and `useLineHitCounts`)
- [x] Fix comments to always scroll to the associated line (in _most_ cases)


markerikson:
Remind me what the problem is?  Is this about comment clicks not scrolling?

bvaughn:
Yeah. They open the source but they don't scroll to the line anymore. Looking at the "should scroll" function, I realized it was never returning true– but we had this prop in Redux, kind of seemingly made for this (given when it's set/reset). Then again I have _no_ context for this stuff so maybe I'm misunderstanding 😄

bvaughn:
Ignore my 5 minute change. Looking at the Editor, I think it maybe needs a re-work. I think I'm seeing a few things we might want to revisit:
1. We're causing side effects (scrolling) from `componentWillReceiveProps`.
2. We're not correctly handling location changes in some cases (so we miss scrolling)

The code is also written in a kind of unusual way:
```js
  // Called by componentDidMount and UNSAFE_componentWillReceiveProps
  updateEditor(props: PropsFromRedux) {
    let editor = this.state.editor;

    if (!this.state.editor && props.selectedSource) {
      // This is odd; if it's not in state, we call a helper function that creates it and sets state–
      // but we do it during render.
      editor = this.setupEditor();
    }

    // Then we do more things with the editor
    // (that isn't even yet in state, since the previous render is still in progress)
```

markerikson:
Oh I am _totally_ on board with rewriting this to be a function component :)

bvaughn:
Okay. I'm going to spend a little time this morning seeing what that might look like. I made some good initial progress last night.

bvaughn:
Summary from 1:1 chat Mark and I just had:

* This branch doesn't _completely_ fix the bug reported in FE-791 but it _does_ fix a couple of cases of it (so it's better than production).
* I'm not aware of any _regressions_ this branch introduces, so it seems a net positive.
* It's scary to land a change like this before the weekend, so I'll land it first thing next week.

bvaughn:
I think this should be ready to land in the morning.

PR #7933, opened 2022-10-07
Remove frames and scopes from redux (FE-819)

hbenl:


src/ui/suspense/frameCache.ts
```diff
@@ -7,12 +7,16 @@ export const {
   getValueSuspense: getFramesSuspense,
   getValueAsync: getFramesAsync,
   getValueIfCached: getFramesIfCached,
-} = createGenericCache<[pauseId: PauseId], Frame[] | undefined>(
+  useGetValue: useGetFrames,
+} = createGenericCache<[pauseId: PauseId | undefined], Frame[] | undefined>(
   async pauseId => {
+    if (!pauseId) {
+      return undefined;
+    }
```
bvaughn, line 15: Under what condition would this method be called without a pause id? That feels suspicious to me.

We should just not call this method if there's no pause id.

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -199,30 +230,41 @@ function getAlternateSourceIdForPosition(
...
+    return { sourceId: alternateSourceId };
+  }
+
+  const uniqueAlternate = getUniqueAlternateSourceId(
+    selectedSource,
+    sourcesState.sourceDetails.entities
+  );
   if (uniqueAlternate.sourceId || uniqueAlternate.why === "no-sourcemap") {
     return uniqueAlternate;
   }
```
bvaughn, line 261: I'm not sure I understand how these lines make sense.

bvaughn (commented):
I'm feeling pretty under the weather today so I just skimmed it and left a question or two. Probably want @markerikson to review it.

src/ui/suspense/frameCache.ts
hbenl: I've added `useGetValue(...)` to the generic cache (which is used as `useGetFrames(...)` in `CommandBar.tsx`) - since that is a hook, we can't bail before calling it, even if calling it makes little sense in that case.

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
hbenl: If `getUniqueAlternateSourceId()` returns a `sourceId` or if it says that there is no sourcemap then we want to pass that on. If it says that there are multiple alternate sourceIds (`uniqueAlternate.why === "not-unique"`) then we want to carry on and call `getAlternateSourceIdForPositionSuspense()`

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
```diff
@@ -90,5 +98,26 @@ export function createGenericCache<TParams extends Array<any>, TValue>(
...
+    useGetValue(...args: TParams) {
+      const [counter, setCounter] = useState(0);
+      function triggerRendering() {
+        setCounter(counter + 1);
+      }
+
+      const record = getOrCreateRecord(...args);
+      switch (record.status) {
+        case STATUS_PENDING: {
+          record.value.then(triggerRendering, triggerRendering);
```
bvaughn, line 111: This looks side effecty. Rendering multiple times would end up attaching multiple `triggerRendering` callbacks.

src/ui/suspense/frameCache.ts
bvaughn: Why do we need `useGetValue` ?

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
hbenl: Yes, but they all resolve at the same time, so this should not trigger rerendering more than once.

src/ui/suspense/frameCache.ts
hbenl: We don't strictly need it but it's convenient in `CommandBar.tsx` and `Toolbar.tsx`: in both cases we want to enable some buttons or add a badge to one of them if there are frames for the current point. Doing this with a suspense mechanism would be less straightforward.

bvaughn: > Doing this with a suspense mechanism would be less straightforward.

Why?

I feel uneasy with us reinventing an alternative Suspense model here. Feels like we're complicating things unnecessary, and it introduces extra constraints like the "can't be called conditionally" thing you. mentioned above.

bvaughn: > Doing this with a suspense mechanism would be less straightforward.

I think maybe there's something here that's being misunderstood. Unless I'm missing some important context, I think Suspense should be a good fit for the command bar and toolbar components too. Maybe you can elaborate on what the complicated part is?

hbenl: Maybe I'm just blind. Imagine you're doing something like this:
```
function CommandBar({ pauseId }) {
  const { frames } = useGetFrames(pauseId);
  const disabled = !frames?.length;
  return (
    <Button stuff={...} disabled={disabled} />
    <Button stuff={...} />
    <Button stuff={...} disabled={disabled} />
  );
```
How would you do that with `getFramesSuspense()`? Maybe something like this:
```
function CommandBar({ pauseId }) {
  return (
    <Suspense fallback={<Button stuff={...} disabled={true} />}>
      <SuspendingButton pauseId={pauseId} stuff={...} />
    </Suspense>
    <Button stuff={...} />
    <Suspense fallback={<Button stuff={...} disabled={true} />}>
      <SuspendingButton pauseId={pauseId} stuff={...} />
    </Suspense>
}

function SuspendingButton({ pauseId, ...stuff }) {
  const frames = getFramesSuspense(pauseId);
  const disabled = !frames?.length;
  return <Button stuff={...stuff} disabled={disabled} />;
}
```
And that looks a lot less straightforward to me. But maybe you know a cleaner way to write this with suspense.

bvaughn: Ah, thanks for elaborating. The transition and deferred APIs can be used for this sort of thing.

```js
function CommandBar({ pauseId }) {
  const deferredPauseId = useDeferredValue(pauseId);
  const isPending = deferredPauseId !== pauseId;

  const { frames } = deferredPauseId != null ? getFramesSuspense(pauseId) : null;
  const hasFrames = !!frames?.length;

  const disabled = isPending || !hasFrames;

  return (
    <Button stuff={...} disabled={disabled} />
    <Button stuff={...} />
    <Button stuff={...} disabled={disabled} />
  );
```

hbenl: Interesting, thanks. I'll have to read up on `useDeferredValue()` later to understand this.

bvaughn: Would you be interested in pairing on that change? so we can avoid adding a Suspense-alternative like `useGetValue`?

hbenl: I have just pushed an improved hook that doesn't violate the rules and is better at avoiding unnecessary rerenders.
I've also experimented with your solution above but `useDeferredValue()` didn't work as I expected, in particular on the very first render `useDeferredValue(pauseId)` will return `pauseId`, so the component will suspend (instead of rendering disabled buttons).

bvaughn: > on the very first render `useDeferredValue(pauseId)` will return pauseId, so the component will suspend (instead of rendering disabled buttons).

Isn't that expected? If there's no data to show (even in an "outdated but this is updating" way) then it should suspend, no?

hbenl: > Isn't that expected?

It's not what *I* expected. The original implementation would always render disabled buttons while frames are being fetched and I wanted to keep that behavior. If we want to change it so that the buttons keep their old disabled state until the frames are fetched, transitions are obviously what we need. But I wanted to discuss how to implement the old behavior with the suspense cache.

bvaughn: Okay. Fair enough.

What I was trying to say is just that...that's how Suspense works. When content is loading the first time, React shows the fallback UI (which is generally some kind of spinner or "_loading_" text). Knowing this, you can design the UI to "fallback" however you'd like.

My point is that this isn't a bug or a problem. It's just how Suspense is meant to work.

hbenl: So would you agree that the `useGetValue()` hook is useful if we want to keep the old behavior?

bvaughn: I still think the `useGetValue` abstraction is unnecessary and that we should use Suspense as it works out of the box rather than work around it by adding extra state and effects, but I understand that we don't share that opinion.

I wrote the code snippet above under the assumption that we were talking about _updates_ (because those are a little more complicated until you understand how transitions work).

I'm probably not going to think about this PR too much until the points stuff is done and merged because it's kind of sucking up all of my mental energy at the moment. Maybe we can chat about this again on Monday?

bvaughn (approved):
I think the cleanup in this PR is a good thing to land.

As mentioned before, I don't personally like the `getValueIfCached` abstraction but I don't feel super strongly about it and I don't want to hold up the nice things in this PR for it either 😄 

I may post a follow up proposal for a pure Suspense alternative though.

bvaughn:
Not sure if @markerikson would like to review it before it lands?

bvaughn:
It looks like there were a bunch of failing Playwright tests for this. Hopefully they were false positives?

hbenl:
They passed locally and also on `main` after merging: https://github.com/replayio/devtools/actions/runs/3266009403

bvaughn:
Ooo, maybe this is another instance of the bot messing up :(

PR #7934, opened 2022-10-07
Revert "**REALLY KILL `NodeFront`, `ValueFront`, and dead `ThreadFront/Pause` logic!!!!**"

markerikson:
Reverts replayio/devtools#7915


markerikson:
Seems this was an over-reaction - the symptom I saw was source files failing to load, and it seems like the actual problem was old source IDs cached in `localStorage` for pending breakpoints.

PR #7940, opened 2022-10-08
Load recording when auth is injected by replay browser

ryanjduffy:
## Issue

When auth is passed from the Replay browser into the app when the app is on a recording page, the recording isn't loaded

## Resolution

* The `deferredState` promise in `tokenManager` was already resolved in the first load of the page so the token from the Replay browser wasn't available in `DevTools` to set as the access token to the backend. Resetting the token state when the external auth is injected resets that promise so it can be resolved again.
* The error modal needs to be cleared as well since it doesn't apply now that auth has occurred.


src/ui/actions/session.ts
```diff
@@ -62,6 +61,8 @@ export function getAccessibleRecording(
         if (expectedError) {
           dispatch(setExpectedError(expectedError));
           return null;
+        } else {
+          dispatch(selectors.clearExpectedError());
```
ryanjduffy, line 65: This seems suspect to me since it could be clearing other errors. Alternatives welcome

bvaughn, line 65: nit: Can we just import `clearExpectedError` (without the `selectors` wrapper):
```suggestion
          dispatch(clearExpectedError());
```
Not a fan of the `import * as blah` pattern we use in the codebase. I think it mostly exists for class components, to distinguish between the "connected" action and the imported one, but for function components this distinction is unnecessary 😄 

As for the "suspect" part – I guess you're right, since other things might call `setExpectedError`

Is there some way we could check the current error, and only clear it if it's the exact type this module sets?

bvaughn (commented):
I don't really have any context on this code. The `tokenManager` component looks kind of brittle, with the timeouts and global listeners but 🤷🏼 I guess that's been that way for a long time so it's probably okay.

src/ui/actions/session.ts
ryanjduffy: `tokenManager` is definitely up to be revisited sometime soon. It's messy at best.

I've added a new helper action to clear just the recording access errors which I think makes it more clear and safer.

jasonLaster (approved):
It'd be nice to have a replay of this flow as a reference

PR #7943, opened 2022-10-10
Add "copy cURL" to Network Monitor

jonbell-lot23:
Addresses https://linear.app/replay/issue/FE-826/network-monitor-copy-as-curl

image

image


jonbell-lot23:
There are typescript errors I need to work through, stay tuned

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -78,73 +81,100 @@ const RequestTable = ({
     }
   };
 
+  const copyAsCurl = () => {
+    setContextMenuData(null);
+    navigator.clipboard.writeText("curl '" + contextMenuData.row.values.url + "'");
```
jasonLaster, line 86: I think the curl also includes header information as well.

jasonLaster (commented):
Very cool

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -78,73 +81,100 @@ const RequestTable = ({
...
+
+    setShowCopied(true);
+    setTimeout(() => setShowCopied(false), 2700);
+    setTimeout(() => document.getElementById("showCopied").classList.remove("opacity-100"), 2000);
+    setTimeout(() => document.getElementById("showCopied").classList.add("opacity-0"), 2000);
```
bvaughn, line 91: The way this is written (a chain of `setTimeout` calls that attempt to line up with a CSS transition) seems kind of brittle. If it's the effect you're going for, I don't have a better suggestion, but maybe we can make it a little more obvious what it's doing?

```suggestion
let timeout = setTimeout(() => {
  document.getElementById("showCopied").classList.remove("opacity-100");
  document.getElementById("showCopied").classList.add("opacity-0");
  
  // The fade out transition lasts 700ms
  timeout = setTimeout(() => {
    setShowCopied(false);
  }, 700);
}, 2000);

return () => {
  clearTimeout(timeout);
}
```

Also we should probably clear the timeout on unmount (as shown above) so we don't risk a potential undefined access error.

jonbell-lot23: Thanks! All good suggestions.

jonbell-lot23:
This item was moved into the FE backlog for now.

PR #7947, opened 2022-10-11
Re-wrote Popover component to better handle resizing popup content

bvaughn:
The old `Popover` component did not properly align content in some cases, nor did it react to content that resized (e.g. data that loads async via Suspense). This commit rewrites that component and removes some old dead/unused styles.

Here is a before and after that illustrates the previous positioning problem:

| `main` branch (misaligned)  | rewrite (properly aligned) |
| :-- | :-- |
| Screen Shot 2022-10-11 at 3 59 23 PM | Screen Shot 2022-10-11 at 4 00 21 PM |

The new `Popover` component can be used for popovers that have tails (like the inspector that has a tail which points to a variable) or ones that do not (such as simple tooltips). At the moment, we only use it for the former but we could replace other instances with this as well.

Kapture 2022-10-11 at 14 34 21


bvaughn:
Note that the e2e test failure is a false negative:
> inspector-02: element picker and iframe behavior

Looks like the recording timed out loading the initial page. Ran it locally and it passed.

markerikson (approved):

src/devtools/client/debugger/src/components/shared/Popover.tsx
```diff
@@ -1,310 +1,138 @@
...
-import React, { Component } from "react";
-import ReactDOM from "react-dom";
-
-import classNames from "classnames";
-import BracketArrow from "./BracketArrow";
-import type { BracketArrowOrientation, BAProps } from "./BracketArrow";
-import SmartGap from "./SmartGap";
-
-interface PopoverProps {
-  type: "tooltip" | "popover";
```
bvaughn, line 15: The old "tooltip" type wan't being used, but it's kind of akin to `showTail={false}` in the new component so far as I **know.**

PR #7951, opened 2022-10-12
Refactor breakpoint/logpoint architecture

bvaughn:
Split into separate commits for easier review:
- [x] Delete old breakpoint Redux state
- [x] Modify components to use new points architecture


packages/e2e-tests/tests/focus_mode-01.test.ts
```diff
@@ -34,7 +34,7 @@ test("focus_mode-01: should filter messages as regions based on the active focus
   await setFocusRange(page, { startTimeString: "0:04", endTimeString: "0:05" });
 
   // Verify fewer messages
-  await verifyConsoleMessage(page, "Log point", "log-point", 11);
+  await verifyConsoleMessage(page, "Log point", "log-point", 10);
```
bvaughn, line 37: I'm not sure why this test had 11 before.

I only see 10 messages in the Console after focusing on 0:04-0:05– both in this branch and [in production](https://app.replay.io/recording/7e0c2f3e-ae52-4af2-8603-6234de4eda10).

bvaughn:
* `breakpoints-07` flaked once on this PR, then passed. Passes for me locally. IMO our bot shouldn't leave comments in this case, or it should at least comment clearly as "flake" not "fail" but 🤷🏼

markerikson:
Playing with this now.  Will paste in issues (and hopefully replays) as I see them.

First: the hover tooltip that should show "N hits" is missing the number in most cases:

- https://app.replay.io/recording/new-points-pr-no-hit-numbers-in-tooltip--80afea2c-fc7c-4ae8-8d27-947e511b7226

markerikson:
Hmm. I was scrolling through the giant `run.js` file in https://app.replay.io/recording/if-thingisbeautiful-andand-thingisbiggerthanbreadbox-oror-thingisyellow--6ea8b10b-3e04-4475-be94-e29820424d7e?point=13954297841544188978501389905821696&time=11491&hasFrames=false to get a sense of how this behaves with loading hit counts in large files.

With the protocol message log viewer active, I noted that this PR is making tons of `createAnalysis` requests every time we hover a new line.  However, I tried doing the same thing in prod, and I see we're triggering basically the same number of requests, so that's _not_ new behavior in this PR.

Either way that seems like something we should improve?

bvaughn:
> Either way that seems like something we should improve?

Agreed! Strong preference to do this as a follow up though, since it's not a regression.

bvaughn:
> First: the hover tooltip that should show "N hits" is missing the number in most cases:

This is unexpected to me, since that tooltip code is:
```js
if (hitsCounts == null || hoveredLineNode === null) {
  return <Loading hoveredLineNode={hoveredLineNode} />;
}

return (
  <StaticTooltip targetNode={hoveredLineNode}>
    <Wrapper showWarning={hitsCount! > 200}>
      {hitsCount} hit{hitsCount == 1 ? "" : "s"}
    </Wrapper>
  </StaticTooltip>
);
```

bvaughn:
RE: The large layout effect that has `points` as a dependency: baa5cc0

src/devtools/client/debugger/src/utils/prefs.js
```diff
@@ -158,17 +158,14 @@ export const features = new PrefsHelper("devtools.debugger.features", {
   jumpLine: ["Bool", "jump-line"],
 });
 
-export const asyncStore = asyncStoreHelper("debugger", {
-  pendingBreakpoints: ["pending-breakpoints", {}],
```
markerikson, line 162: :hand: This seems like an issue - do we actually still persist these anywhere?

src/devtools/client/debugger/src/components/Editor/useEditor.ts
```diff
@@ -316,17 +404,34 @@ export default function useEditor(
...
       codeMirrorWrapper.removeEventListener("mouseover", onCodeMirrorMouseOverLine);
     };
-  }, [dispatch, editor, setContextMenu]);
+  }, [
+    addPoint,
+    deletePoints,
+    editPoint,
+    dispatch,
+    editor,
+    points,
```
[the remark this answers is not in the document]
markerikson: Got resolved already

src/devtools/client/debugger/src/actions/sources/newSources.ts
```diff
@@ -1,73 +1,17 @@
...
-    }
-
-    for (const bp of pendingBreakpoints) {
-      const line = bp.location.line;
-      const column = bp.location.column;
-      dispatch(setRequestedBreakpoint({ column, line, sourceId }));
-    }
-
-    // load the source text if there is a pending breakpoint for it
-    const textPromise = dispatch(loadSourceText(source.id));
```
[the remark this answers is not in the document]
bvaughn: The concept of "pending" went away entirely because it's not necessary anymore.

Persistence is done using `useLocalStorage`:
https://github.com/replayio/devtools/blob/1b9bfd91ab2ba4d4a0667c7f21e96efe574c2e99/packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx#L40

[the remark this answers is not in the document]
bvaughn: It's built into the cache:
https://github.com/replayio/devtools/blob/1b9bfd91ab2ba4d4a0667c7f21e96efe574c2e99/packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx#L40

src/devtools/client/debugger/src/client/index.ts
```diff
@@ -7,21 +7,10 @@ import type { ThreadFront as TF } from "protocol/thread";
...
 
-import { initialBreakpointsState } from "../reducers/breakpoints";
-import { asyncStore, verifyPrefSchema } from "../utils/prefs";
+import { verifyPrefSchema } from "../utils/prefs";
 
 import { resumed, paused } from "../actions/pause";
 
-export async function loadInitialState() {
-  // @ts-expect-error missing `pendingBreakpoints` field
-  const pendingBreakpoints = await asyncStore.pendingBreakpoints;
```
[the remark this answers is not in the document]
bvaughn: It's built into the cache:
https://github.com/replayio/devtools/blob/1b9bfd91ab2ba4d4a0667c7f21e96efe574c2e99/packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx#L40

src/devtools/client/debugger/src/components/Editor/Breakpoints/useBreakpoints.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+
+      const lineNumber = point.location.line;
+      if (renderedLines.has(lineNumber)) {
+        return;
+      }
+
+      renderedLines.add(lineNumber);
+
+      const marker = makeMarker();
```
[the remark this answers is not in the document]
bvaughn: Yes.

This was kind of a conscious decision, since it provided a simple way for adding vs removing. (I wanted to avoid having to manually diff the previous and next `points` array to determine what needed to be added or removed.)

I _think_ it's probably fine in practice since there aren't many points, as you say.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -61,137 +52,74 @@ function Wrapper({
...
-export default function LineNumberTooltipWrapper(props: Props) {
-  return <LineNumberTooltip {...props} />;
-}
-
-function LineNumberTooltip({ editor, keyModifiers }: Props) {
-  const dispatch = useAppDispatch();
-  const [targetNode, setTargetNode] = useState<HTMLElement | null>(null);
-  const lastHoveredLineNumber = useRef<number | null>(null);
+function LineNumberTooltipSuspends({ keyModifiers }: Props) {
   const isMetaActive = keyModifiers.meta;
```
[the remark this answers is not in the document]
bvaughn: Sorry, it's not clear what you're asking about. Why do we care about...the modifiers?

The component renders null if `keyModifiers.meta` is true. I'm not sure why that's done 😄

src/devtools/client/debugger/src/components/Editor/useEditor.ts
```diff
@@ -219,22 +231,78 @@ export default function useEditor(
 
     const onCodeMirrorContextMenu = (_: any, __: number, ___: any, event: MouseEvent) => {};
 
-    const onCodeMirrorGutterClick = (_: any, line: number, __: any, event: MouseEvent) => {
-      const { button, ctrlKey, target } = event;
-      const { cx, selectedSource } = instancePropsRef.current;
+    const onCodeMirrorGutterClick = async (
+      cm: any,
```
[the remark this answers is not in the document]
bvaughn: Yeah, to be fair– our types are not very good RE: `SourceEditor`, `CodeMirror`, and `EditorWithDoc` – but this was true before, and I got burned out thinking about improving it with this PR. Worth a follow up.

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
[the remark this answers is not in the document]
markerikson: Yeah, I don't know why we actually do that behavior in the first place :)

src/devtools/client/debugger/src/components/Editor/useEditor.ts
```diff
@@ -219,22 +231,78 @@ export default function useEditor(
...
+      // ignore right clicks in the gutter
+      if ((clickEvent.ctrlKey && clickEvent.button === 0) || clickEvent.button === 2) {
+        return;
+      }
 
-      if ((ctrlKey && button === 0) || button === 2 || !selectedSource) {
-        // ignore right clicks in the gutter
+      // Don't add a breakpoint if the user clicked on something other than the gutter line number,
+      // e.g., the blank gutter space caused by adding a CodeMirror widget.
+      if (![...(clickEvent.target as HTMLElement).classList].includes("CodeMirror-linenumber")) {
```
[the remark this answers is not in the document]
bvaughn: Did you mean?
```js
(clickEvent.target as HTMLElement).classList.contains("CodeMirror-linenumber")
```
Sure. Seems fine.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointTimelinePoint.tsx
```diff
@@ -102,29 +102,26 @@ function BreakpointTimelinePoint({
...
-  (prevProps, nextProps) => {
-    function selectorChanged(selector: (props: BTPProps) => any) {
-      return selector(nextProps) !== selector(prevProps);
-    }
-
-    function hasChanged(key: keyof BTPProps) {
-      return nextProps[key] !== prevProps[key];
-    }
+const MemoizedBreakpointTimelinePoint = memo(BreakpointTimelinePoint, (prevProps, nextProps) => {
+  function selectorChanged(selector: (props: BTPProps) => any) {
```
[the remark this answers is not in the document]
bvaughn: Yeah. I didn't really touch this. Let's clean it up later.

src/devtools/client/debugger/src/reducers/pending-breakpoints.ts
```diff
@@ -1,96 +0,0 @@
...
-import {
-  createPendingBreakpoint as savePendingBreakpoint,
-  makePendingLocationId,
-} from "../utils/breakpoint";
-
-import { removeBreakpoint, removeBreakpoints, Breakpoint } from "./breakpoints";
-import { PendingBreakpoint, SourceLocation } from "./types";
-
-/**
- * A set of known breakpoints for all recordings, persisted to `indexedDB`
```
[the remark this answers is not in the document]
bvaughn: Mentioned elsewhere, but `useLocalStorage`.

src/devtools/client/debugger/src/utils/prefs.js
bvaughn: Yes. See all of the other replies 😆

src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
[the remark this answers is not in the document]
bvaughn: Happy to stamp a PR that removes it (as a follow up) 😄

src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx
```diff
@@ -41,24 +37,34 @@ export default function EmptyLines({ editor }: ELProps) {
...
+      const breakpointPositions = await getBreakpointPositionsAsync(
+        replayClient,
+        sourceId,
+        visibleLines
+      );
+      const breakableLines = breakpointPositions.map(({ line }) => line);
+
+      // For performance reasons, we should only modify lines (chunks of lines in this case) once.
+      // However if the breakpoint positions change, we need to re-draw things.
+      if (memoizedDrawnLines.breakpointPositions !== breakpointPositions) {
```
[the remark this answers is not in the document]
bvaughn: Hmm. Yes, but I think I'm probably over invalidating now that I look again, because this would also change when `visibleLines` changes and we don't actually need that. I'll relax it to be source ID Specific.

**Edit** – I remember now why I did this. It's. because this is how it was written before:
https://github.com/replayio/devtools/blob/1b9bfd91ab2ba4d4a0667c7f21e96efe574c2e99/src/devtools/client/debugger/src/components/Editor/EmptyLines.tsx#L25-L29

But I think you were also over-invalidating.

src/devtools/client/debugger/src/components/Editor/Breakpoints/useBreakpoints.ts
```diff
@@ -0,0 +1,113 @@
...
+      marker.addEventListener("click", event => {
+        if ((event.ctrlKey && event.button === 0) || event.button === 2) {
+          // Ignore right clicks
+          return;
+        }
+
+        event.stopPropagation();
+        event.preventDefault();
+
+        if (point.shouldLog) {
```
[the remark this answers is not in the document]
bvaughn: Yeah, fair. Would be a reasonable follow up. :)

markerikson:
One new concern: it looks like the switch from persisting `pendingBreakpoints` over to `${recordingId}::points` would presumably lose _all_ existing persisted breakpoints.  Is there any way we can convert/adapt those?

bvaughn:
> One new concern: it looks like the switch from persisting `pendingBreakpoints` over to `${recordingId}::points` would presumably lose all existing persisted breakpoints. Is there any way we can convert/adapt those?

Yes, it's technically possible, although we would need to:
1. Check `localStorage` and if empty–
1. Check the old storage location and if we find a match–
1. Convert the old points to the new format

My gut is that this is not worth it. It's a one time migration and the impact doesn't seem very large.

src/devtools/client/debugger/src/reducers/breakpoints.ts
```diff
@@ -1,443 +0,0 @@
...
-      action: PayloadAction<{
-        analysisId: string;
-        location: Location;
-        condition?: string;
-        timeRange: TimeStampedPointRange | null;
-      }>
-    ) {
-      const { analysisId, location, condition, timeRange } = action.payload;
-
-      analysesAdapter.addOne(state.analyses, {
```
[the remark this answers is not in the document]
bvaughn: It's been replaced by the new Suspense caches– primarily [`AnalysisCache`](https://github.com/replayio/devtools/blob/main/packages/bvaughn-architecture-demo/src/suspense/AnalysisCache.ts) and [`PointsCache`](https://github.com/replayio/devtools/blob/main/packages/bvaughn-architecture-demo/src/suspense/PointsCache.ts).

Do you have a specific piece of functionality you're wondering about?

packages/shared/client/ReplayClient.ts
```diff
@@ -580,7 +589,7 @@ export class ReplayClient implements ReplayClientInterface {
 
       hitCounts.set(location.line, {
         columnHits: [...previous.columnHits, columnHits],
-        hits: previous.hits + hits,
+        hits: Math.max(hits, previous.hits),
```
jasonLaster, line 592: was this what we were doing before?

bvaughn: Functionally, I'm not sure– but I think it's what makes the most sense. If we're combining all hits on a line to a single number in the gutter, I think it makes sense to show the highest number.

bvaughn: I think I see a problem on production related to this, in fact.

Looking at recording 768b3f13-0635-4dbc-8b8e-e3f53a70e405, and opening `bundle_input.js` – I see the backend return the following for `getHitCounts`:
```json
      {
        "location": {
          "sourceId": "ppo2-0-bf6b0aa22d03fb56ecb173a045d549ab9e5a9f5f0d646bc12cf33587db090dfe-34da5b56d657c786b881afc8211b9a9e4d692c26ff995edfe976f7fddced4e7d",
          "line": 1,
          "column": 15
        },
        "hits": 56
      }
```

My branch shows:
Screen Shot 2022-10-13 at 11 11 07 PM

Production shows:
Screen Shot 2022-10-13 at 11 11 13 PM

bvaughn:
`stepping-06` seems flaky, but `breakpoints-07` seems legit broken. I'll dig into this more in the morning.

packages/shared/client/ReplayClient.ts
jasonLaster: The rationale for summing the hit counts is that that's what you'll see when you add a print statement

jasonLaster:
Will spend some time testing / reviewing tomorrow, but just wanted to state the obvious. This change is awesome.

packages/bvaughn-architecture-demo/src/hooks/useBreakpointIdsFromServer.ts
```diff
@@ -53,10 +51,11 @@ export default function useBreakpointIdsFromServer(
         prevPoints.forEach(prevPoint => {
           const point = prevPoints.find(({ id }) => id === prevPoint.id);
```
hbenl, line 52: this should be `points.find(...)`, otherwise `point` will never be `undefined` and breakpoints will not be removed on the server.

packages/shared/client/ReplayClient.ts
```diff
@@ -114,17 +114,26 @@ export class ReplayClient implements ReplayClientInterface {
...
+      const { breakpointId } = await client.Debugger.setBreakpoint(
+        {
+          condition: point.condition || undefined,
+          location,
+        },
+        sessionId
+      );
+
+      breakpointIds.push(breakpointId);
+    }
```
hbenl, line 134: We should set the breakpoints in parallel:
```typescript
const breakpointIds = await Promise.all(
  this._getCorrespondingLocations(point.location).map(location =>
    client.Debugger.setBreakpoint(
      {
        condition: point.condition || undefined,
        location,
      },
      sessionId
    )
  )
);
```

hbenl (changes requested):

packages/bvaughn-architecture-demo/src/hooks/useBreakpointIdsFromServer.ts
bvaughn: lol yeah that's a silly mistake on my part. You're right.

packages/shared/client/ReplayClient.ts
bvaughn: Sure. That seems like a pretty minor optimization but I'm not opposed to it.

bvaughn:
I made a small test recording (324ad5fb-ce14-4222-8d6d-24fec92e5aff) to compare the corresponding sources behavior this morning (for line hit counts, breakpoints/logpoints, etc.)

src/ui/components/SecondaryToolbox/TerminalContextAdapter.tsx
```diff
@@ -0,0 +1,50 @@
...
+export default function TerminalContextController({ children }: PropsWithChildren) {
+  const [isPending, startTransition] = useTransition();
+  const [messages, setMessages] = useState<TerminalExpression[]>([]);
+
+  const idCounterRef = useRef(0);
+
+  const addMessage = useCallback((partialTerminalExpression: NewTerminalExpression) => {
+    // New expressions should be added in a transition because they suspend.
+    startTransition(() => {
+      startTransition(() => {
```
hbenl, line 24: Do we need two nested `startTransition()` calls here? If so, why?

bvaughn: No, def not. This looks like a rebase flub. Thanks!

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -170,9 +180,10 @@ function getAlternateSourceIdForPosition(
   client: ReplayClientInterface,
   source: SourceDetails,
   position: CursorPosition,
-  sourcesById: Dictionary<SourceDetails>
+  sourcesById: Dictionary<SourceDetails>,
+  sourceLocationRange: SourceLocationRange | null
 ) {
-  const lineLocations = getBreakpointPositionsSuspense(client, source.id, null);
+  const lineLocations = getBreakpointPositionsSuspense(client, source.id, sourceLocationRange);
```
hbenl, line 186: Requesting all breakpoint positions for the source (i.e. without a range) certainly wasn't optimal but I don't think we should use the currently *visible* source range here since we're interested in finding a breakable position after the currently *selected* position and that position may not be in the visible range. This has two downsides:
- if the currently selected position is below the currently visible range the SourcemapToggle will always say "The currently selected location is not mapped" even if that is not the case
- we rerender the SourcemapToggle every time the currently visible range changes

I think we should use a range starting with the currently selected line and ending maybe 10 lines below that.

hbenl (changes requested):

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
bvaughn: > I don't think we should use the currently visible source range here since we're interested in finding a breakable position after the currently selected position and that position may not be in the visible range

I don't understand what you mean by this.

> if the currently selected position is below the currently visible range the SourcemapToggle will always say "The currently selected location is not mapped" even if that is not the case...I think we should use a range starting with the currently selected line and ending maybe 10 lines below that.

Why would 10 lines help? If there's a concern that something could be selected but not within the view port, then wouldn't that still be possible?

Btw, I don't think we're ever passing the visible range param to these functions. (It's optional.)

PS. You don't need to keep marking "_requested changes_" 😄  This is a large refactor. I'm not going to merge it until everyone is onboard.

bvaughn: When I say: _I don't understand what you mean by this_

What I mean is: Can you describe a specific scenario that would cause a problem here?

hbenl: > I don't understand what you mean by this.

We look for the first entry in `lineLocations` whose `line` is greater or equal to `position.line`. If the range for which we requested the breakable positions doesn't include `position.line` then we won't find any and hence we can't get a `MappedLocation` (because that only works for breakable positions) and hence we can't find an alternate source.

> Why would 10 lines help? 

It's not about the 10 lines, it's about the range including `position.line`. Since we're only interested in breakable positions *after* that line, we don't need any earlier lines in the range but the next breakable position may be at a line after `position.line`, so the range needs to include some lines after `position.line`. We could use a range starting with `position.line` and ending with the last line of the source but that could become a perf problem (just like not using a range at all). I think it's reasonable if there are no breakable positions on the first 10 lines after `position.line` to say we can't find an alternate source.

> Can you describe a specific scenario that would cause a problem here?

Select a position in a bundle that you're not currently paused in and that can be mapped to an original source. As long as the selected position is visible, the SourcemapToggle will let you switch to that original source. But as soon as you scroll the source up so that the selected position is not visible anymore, the SourcemapToggle gets disabled and says "The currently selected location is not mapped". Interested in a replay?

bvaughn: > Since we're only interested in breakable positions _after_ that line

I don't understand this part. I thought we were specifically looking for other positions _on that same line_. Why would we look for ones on _after_ the specified line?

packages/shared/client/ReplayClient.ts
bvaughn: Thanks for explaining this. It set me off down a bit of a rabbit trail, but I learned:
* We merge hit counts for all corresponding sources
* We only count the hits for the first breakable column on any given line (for each source)
* We _have_ to _explicitly_ fetch breakable points for each source before querying their hit counts, or the backend will incorrectly say that we've asked about an "invalid location" (which I filed BAC-2370 about)

This branch and production now match in terms of the hit count labels displayed.

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
bvaughn: tl;dr I'm going to remove the visible lines range filtering from this Suspense cache. Looking at production, we're actually fetching this data _multiple times_ currently– once without any visible lines filtered, and once with them filtered (or once per chunk of visible lines).

So removing the filtering altogether will make things _better_. (We'll only fetch once and then read from memory.) It will also side-step this issue.

packages/bvaughn-architecture-demo/src/utils/time.ts
```diff
@@ -7,8 +7,13 @@ import { ExecutionPoint, PointRange, TimeStampedPointRange } from "@replayio/pro
...
 import prettyMilliseconds from "pretty-ms";
 
+const collator = new Intl.Collator("en", {
+  numeric: true,
+  sensitivity: "base",
+});
+
 export function compareExecutionPoints(a: ExecutionPoint, b: ExecutionPoint): number {
-  return a.localeCompare(b, undefined, { numeric: true });
+  return collator.compare(a, b);
```
bvaughn, line 16: Neat!

hbenl (approved):

PR #7959, opened 2022-10-14
Add chromium net monitor feature flag

jasonLaster:


src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -67,20 +68,24 @@ const PanelButtons: FC<PanelButtonsProps> = ({
   toolboxLayout,
   recordingCapabilities,
 }) => {
-  const { supportsNetworkRequests } = recordingCapabilities;
+  const { supportsNetworkRequests, supportsRepaintingGraphics } = recordingCapabilities;
+  const { value: chromiumNetMonitorEnabled } = useFeature("chromiumNetMonitor");
+
   return (
     <div className="panel-buttons theme-tab-font-size flex flex-row items-center overflow-hidden">
-      {supportsNetworkRequests && <NodePicker />}
```
bvaughn, line 73: 😮 Oh nice catch.

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -160,7 +166,7 @@ function SecondaryToolbox() {
         </div>
       </header>
       <Redacted className="secondary-toolbox-content bg-chrome text-xs">
-        {recordingCapabilities.supportsNetworkRequests && (
+        {(chromiumNetMonitorEnabled || recordingCapabilities.supportsNetworkRequests) && (
```
bvaughn, line 169: Hmm, this reads a bit confusing to me. So the `chromiumNetMonitorEnabled` is an experimental thing? Is it tied to the recording at all?

It seems like this might be less confusing if `recordingCapabilities.supportsNetworkRequests` just took this into consideration– but maybe that doesn't make sense for some reason?

bvaughn (approved):

src/ui/components/SecondaryToolbox/index.tsx
jasonLaster: That was my initial idea, but I wasn't comfortable with mixing user settings with the protocol threadfront...

bvaughn: That makes sense.

It's a little weird to see "not supported" but then we do it anyway– but it's probably fine 😁

PR #7960, opened 2022-10-15
New Source viewer built with CM's "Lezer" parser

bvaughn:
## New Source viewer built on top of the [Lezer parser](https://github.com/lezer-parser/)

It parses and renders large files _fast_ – and it incrementally fetches hit counts as you scroll. For example, a 16k line file:

https://user-images.githubusercontent.com/29597/196064528-1337bf4b-da55-41d8-8237-d19ccf946767.mp4

It supports log points and break points:

https://user-images.githubusercontent.com/29597/196064593-5ce41a0a-16ba-4039-bea5-997801edb2c5.mp4

It supports find-in-file:

https://user-images.githubusercontent.com/29597/196064637-93cc3659-7cc5-4944-8b72-571be404b9cb.mp4

It supports inspecting values for the current frame:

https://user-images.githubusercontent.com/29597/196064724-c083cc58-1570-4d82-a8fe-e2da0a9c1d54.mp4

I also added a simple CMD+O file open dialog just for fun:

https://user-images.githubusercontent.com/29597/196198034-3f261f95-21bb-4c0b-9ba6-322a1b3ec77c.mp4


packages/bvaughn-architecture-demo/src/suspense/TokenizerCache.ts
```diff
@@ -0,0 +1,69 @@
...
+async function tokenizer(code: string): Promise<Tree | null> {
+  const state = codeToState(code);
+  const tree = stateToTree(state);
+
+  return tree;
+}
+
+// TODO
+// Suspense can be async; we could move this to a Worker if it's slow.
+async function highlighter(code: string, tree: Tree): Promise<string[]> {
```
jasonLaster, line 35: we can also be smart and operate on a range of lines as well. This would probably work pretty well if we're already returning an array of lines as well.

This is only needed when we're trying to be lazy and not highlight the entire document.

markerikson: Heh, my main immediate question about the approach is here is that we'd need to start implementing a virtualized list ourselves :)

bvaughn: Yeah for sure. I know exactly how I'm going to window this 😄

bvaughn:
I'm going to land this since it's all inside of my architecture package. Excited to iterate on it more soon!

PR #7962, opened 2022-10-17
Use Line->Column maps of breakable positions for faster lookups

markerikson:
This PR:

- Updates the breakable positions cache to also create a `Map<number, SameLineSourceLocations>` as soon as breakable positions are received, and returns both the array and map as a tuple
- Updates the places where we were iterating over the _entire_ breakable positions array doing line number comparisons to instead look up specific line numbers out of the map
  - The "alternate source ID" logic is a bit trickier because it's trying to find "the first breakable line _after_ the cursor position" so it can translate that into an alternate source mapped location.  I think I came up with a viable approach there by starting from the cursor line number and checking the map until we find a line that has breakable positions
- Tweaks the logic in `EmptyLines.tsx` to only do the `cm.operation()` loop _if_ we know we haven't updated the current chunk of visible lines, and also use map lookups.
- Memoizes `<ProtocolChunk>` as that was showing up in perf captures while scrolling.  (Only affects us since it's a hidden feature, but might as well)

In local dev, the `EmptyLines` check in particular was consistently showing up as 2-3ms multiple times as I'd scroll in Andarist's 270K-line `run.js` file, and afterwards it's down in the 1 microsecond range.


src/devtools/client/debugger/src/components/Editor/ColumnBreakpoints.tsx
```diff
@@ -92,29 +100,19 @@ function PointsForRow({
...
   editor: SourceEditor;
   points: Point[];
 }) {
   const line = points[0]!.location.line;
 
-  // TODO BAC-2329
-  // The backend sometimes returns duplicate columns per line;
-  // In order to prevent the frontend from showing something weird, let's dedupe them here.
-  const breakpointsForLine = useMemo(() => {
-    const match = breakpoints.find(breakpoint => breakpoint.line === line);
```
bvaughn, line 105: This was probably a bit silly before. If the breakpoints returned by the backend _are_ pre-sorted (as the comment elsewhere says) we should have been using a binary search here rather than a linear one.

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -185,39 +185,46 @@ function getUniqueAlternateSourceId(
...
+  // so that we can translate that position into an alternate mapped location.
+  let firstBreakableLineAfterPosition: number | null = null;
+
+  // Grab the last line off the array. It's possible there might _not_ be any breakable lines.
+  const [lastBreakableLine] = breakablePositions.slice(-1);
+
+  // We have a Map with line numbers as keys. Rather than check the entire array,
+  // start from the position's line and check succeeding line numbers to find the
+  // next line number that has a breakable position.
+  for (let lineToCheck = position.line; lineToCheck <= lastBreakableLine?.line; lineToCheck++) {
```
bvaughn, line 209: I wonder if using a binary search lookup (within the sorted array) wouldn't make this slightly more intuitive to do in a performant way?

bvaughn (approved):
Seems like an improvement over the old lookup we were doing!

I wonder if just using a binary search would have been okay though, given the pre-sorted array?

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
markerikson: Possibly. On the other hand, the apparent intent from the existing code is to find the closest breakable line `>=` the position, and I would expect that most of the time that's either the actual line from the position or at most a handful of lines later.

bvaughn: I guess the real question then is: What are the perf characteristics of binary search with a sorted array vs creating a map and then looking things up via it?

hbenl (approved):

PR #7965, opened 2022-10-18
Frames cleanup

hbenl:
Follow-up to the removal of frames and scopes from redux:
- frames used to be referenced by their `asyncIndex` and `index`, which made working with them awkward now that the async stacktrace is not stored in redux anymore - the entire async stacktrace needed to be reconstructed every time we needed to find a frame. Now the frames are referenced by their `pauseId` and `frameId`.
- removed `framePositions` from redux


src/devtools/client/debugger/src/selectors/isSelectedFrameVisible.ts
```diff
@@ -6,15 +6,15 @@ import type { UIState } from "ui/state";
...
 
 /*
  * Checks to if the selected frame's source is currently
  * selected.
  */
-export function isSelectedFrameVisible(state: UIState) {
+export async function isSelectedFrameVisible(state: UIState) {
   const selectedLocation = getSelectedLocation(state);
-  const selectedFrame = getSelectedFrame(state);
+  const selectedFrame = await getSelectedFrameAsync(state);
```
bvaughn, line 17: Are async "selectors" common?

hbenl: Not really. Would you prefer to remove that "selector" and do the `getSelectedFrameId()` / `getPauseFrameAsync()` combo at each call site?

bvaughn: I think a helper function is fine. I just think it's confusing when we have things called "selectors" that do more than just synchronously extract values from Redux state– but @markerikson probably has a more informed opinion here.

markerikson: Yeah, we've recommended `selectX` as a preferred naming convention for selectors for the last few years, but this codebase was already using `getX` instead.

This _is_ becoming confusing now that we're also using `getSomeThingAsync` as a naming convention for our cache APIs...

PR #7969, opened 2022-10-18
earthly: create a devtools dev server image, run arch demo tests against it

jazzdan:


Earthfile
```diff
@@ -1,21 +1,33 @@
...
+
+  COPY package.json yarn.lock .
+  RUN yarn set version 3.2.1
+  COPY packages/protocol/package.json packages/protocol/package.json .
+  COPY packages/bvaughn-architecture-demo/package.json packages/bvaughn-architecture-demo/package.json .
+  COPY packages/design/package.json packages/design/package.json .
+  COPY packages/bvaughn-architecture-demo/playwright/package.json packages/bvaughn-architecture-demo/playwright/package.json .
+
+  COPY .yarnrc.yml .
+  COPY .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```
jazzdan, line 18: No idea if this is the right thing to do to get this yarn setup working in a container. @markerikson any ideas?

Earthfile
```diff
@@ -1,21 +1,33 @@
...
-    COPY postcss.config.js .
-    COPY tailwind.config.js .
-    COPY webpack.config.js .
-    ARG GIT_SHA
-    RUN REPLAY_BUILD_VISUALIZE=1 NODE_ENV=production ./node_modules/.bin/webpack --env REPLAY_RELEASE=$GIT_SHA
-    SAVE ARTIFACT dist /dist AS LOCAL ./dist
+# NOTE(dmiller): this needs to be 16._15_ otherwise playwright will fail to install, at least until we update our playwright version
+FROM node:16.15-bullseye-slim
 
-dist:
```
jazzdan, line 16: These targets are old and broken so I just deleted them.

jazzdan:
Oh, to run the tests with the new method we'd just run `earthly -P +test-architecture-demo`. The `-P` is important as it allows Earthly to run docker images inside of itself.

Earthfile
markerikson: Very likely, yeah. You'll also need to make sure you copy the actual Yarn "binary" file, currently `.yarn/release/yarn-3.2.1.cjs`

bvaughn (approved):
<i have no idea what i'm doing dog>

Would you be willing to also update update the instructions in [this README](https://github.com/replayio/devtools/blob/main/packages/bvaughn-architecture-demo/playwright/README.md)?

It's not clear to me from scanning this, but I think we'd lose two of the three targets that we depend on?

1. Run tests
2. Run tests and save a (visual) recording for debugging purposes
3. Run tests and update screenshot images (and copy them out to the host OS afterward)

jazzdan:
Whoops, sorry for the late response @bvaughn!

Yeah, this is incomplete in the ways that you described but I think it should be possible to enable those things as well. I'll circle back to that when I'm back from vacation on 10/31

bvaughn:
Sounds great! 🥳

jazzdan:
Okay there are now three top level earthly targets:

* `test-architecture-demo`: just runs the test
* `test-architecture-demo-save`: runs the tests and saves the screenshots to disk
* `test-architecture-demo-save-video`: runs the tests and outputs the videos

Note that I did this on about 1.5 hours of sleep so the names probably don't make sense and things can probably be cleaned up. I'll take another pass tomorrow :)

bvaughn:
Awesome! No hurry 😄

jazzdan:
Renamed some things. Now it looks like this:

* `+test-architecture-demo`: run tests
* `+test-architecture-demo-save-snapshots`: run tests, save new snapshots
* `+test-architecture-demo-save-video`: run tests, save videos

bvaughn:
Let me know whenever you're ready for any kind of review/testing!

jazzdan:
@bvaughn just sent you an invite!

PR #7971, opened 2022-10-18
Devtools test tweaks so that we can also run against the backend CI

loganfsmyth:


packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -215,23 +228,29 @@ async function makeReplayPublic(apiKey: string, recordingId: string) {
...
-async function waitUntilMessage(page: Page, message: string, timeout: number = 30_000) {
-  return await new Promise((resolve, reject) => {
-    let timer = setTimeout(reject, timeout);
+async function waitUntilMessage(
+  page: Page,
+  message: string,
+  timeout: number = 30_000
+): Promise<void> {
+  const gotMatch = await new Promise<boolean>(resolve => {
+    let timer = setTimeout(() => resolve(false), timeout);
```
bvaughn, line 237: I'd like to better understand what motivated this change in behavior because it's not completely obvious to me. Before, if the operation timed out, it would thrown– and no it doesn't.

Seems like that's a significant change that might cause false positives in places like

https://github.com/replayio/devtools/blob/798464e945a7dd0e0139ea5e6e067a84747ee9fa/test/src/getExample.js#L32-L37

https://github.com/replayio/devtools/blob/798464e945a7dd0e0139ea5e6e067a84747ee9fa/test/src/getExample.js#L65-L70

packages/e2e-tests/tests/breakpoints-05.test.ts
```diff
@@ -33,6 +34,12 @@ test(`breakpoints-05: Test interaction of breakpoints with debugger statements`,
     lineNumber: 8,
     url,
   });
+
+  // This slight delay ensures that the breakpoint removal has had time to
+  // be processed and synched to the server, otherwise we could get racy behavior
+  // with the next rewind call.
+  await delay(1_000);
```
bvaughn, line 41: If this is necessary, I think it should go into the `removeBreakpoint` helper rather than in a one-off test.

https://github.com/replayio/devtools/blob/798464e945a7dd0e0139ea5e6e067a84747ee9fa/packages/e2e-tests/helpers/source-panel.ts#L231-L255

public/test/examples/doc_rr_objects.html
```diff
@@ -33,7 +33,7 @@
   g[`a${i}`] = i;
 }
 var h = /abc/gi;
-var i = new Date();
+var i = new Date(1666130037374);
```
bvaughn, line 36: I'm curious about this change?

bvaughn (commented):
Hey @loganfsmyth! Thanks for improving this code! 😄

I wish I'd had a chance to take a look before it landed– I hope it's okay for me to ask a follow up question or to? 😄

packages/e2e-tests/scripts/save-examples.ts
loganfsmyth: It does still throw, it's just a few lines farther down: https://github.com/replayio/devtools/blob/798464e945a7dd0e0139ea5e6e067a84747ee9fa/packages/e2e-tests/scripts/save-examples.ts#L252

The main motivation is that before this was essentially `throw undefined;` which made it really really hard for me to figure out what was throwing when this failed.

public/test/examples/doc_rr_objects.html
loganfsmyth: The old fixture was a pain because the date value changed every time we re-recorded the example! This way the test fixture always produces the same values. See https://github.com/replayio/devtools/pull/7971/files#diff-b30a7a48057849c2194202ed696d2d9e394ab383acb8d5468c2b223d647bc6a4R28

packages/e2e-tests/tests/breakpoints-05.test.ts
loganfsmyth: sounds good!

public/test/examples/doc_rr_objects.html
bvaughn: That's fair! 😄 If

packages/e2e-tests/scripts/save-examples.ts
bvaughn: Ahh, fair enough! I overlooked that.

PR #7974, opened 2022-10-19
Fix display of generated scopes (FE-794)

hbenl:
- #7745 introduced FE-794: the decision whether to show original or generated scopes was made when the scopes were added to the cache (and couldn't be changed afterwards). Now we cache both original and generated scopes and pick the ones to show while rendering.
- moving `preferredGeneratedSources` into redux (FE-845) ensures that the scopes panel is rerendered after switching between original and alternate source.
- the `showAlternateSource()` thunk doesn't need to call the `paused()` thunk anymore: this was necessary in the past to recompute the frames and scopes stored in redux

### DIFF
--- a/src/devtools/client/debugger/src/actions/sources/select.ts
+++ b/src/devtools/client/debugger/src/actions/sources/select.ts
@@ -26,6 +26,7 @@ import {
   clearSelectedLocation,
   getSourceIdToDisplayForUrl,
   getSourceToDisplayForUrl,
+  preferSource,
 } from "ui/reducers/sources";
 import { getActiveSearch, getExecutionPoint, getThreadContext, getContext } from "../../selectors";
 import { createLocation } from "../../utils/location";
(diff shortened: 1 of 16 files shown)


src/devtools/client/debugger/src/actions/sources/select.ts
```diff
@@ -218,28 +219,13 @@ export function showAlternateSource(
...
-  return async (dispatch, getState, { ThreadFront }) => {
+  return async (dispatch, getState) => {
     const state = getState();
     if (getSourceDetails(state, oldSourceId)?.isSourceMapped) {
-      ThreadFront.preferSource(newSourceId, true);
-    } else {
-      ThreadFront.preferSource(oldSourceId, false);
-    }
-
-    let selectSourceByPausing = false;
```
markerikson, line 229: :question: what was this `selectSourceByPausing` thing?  Do we still care about that?

src/ui/components/shared/CodeEditor/useAutocomplete.tsx
```diff
@@ -57,24 +59,31 @@ async function getScopeMatches(
...
+      const topFrame = (await getFramesAsync(pauseId))?.[0];
+      if (!topFrame) {
+        return [];
       }
-      // TODO Hardcoding frame ID 0 probably isn't great
-      const frameScope = await getScopesAsync(pauseId, "0");
-      if (!frameScope) {
-        return [] as string[];
+      const frameScopes = pickScopes(
+        await getScopesAsync(pauseId, topFrame.frameId),
```
markerikson, line 75: 👍

markerikson (approved):

src/devtools/client/debugger/src/actions/sources/select.ts
hbenl: That's what I meant when I wrote
> the `showAlternateSource()` thunk doesn't need to call the `paused()` thunk anymore: this was necessary in the past to recompute the frames and scopes stored in redux

PR #7980, opened 2022-10-20
Remove frames from the network redux slice (FE-844)

hbenl:
- remove frames from the network redux slice (FE-844)
- for every pause involved in the async stacktrace: check if the pause is in a loaded region and add an `ErrorBoundary` in case loading frames for that pause fails for some other reason (FE-856)


bvaughn (approved):
Just skimmed it but 👍🏼

markerikson (approved):

PR #7981, opened 2022-10-20
Add suspense wrapper around new console (FE-848)

hbenl:
When opening a replay it can take a long time until the secondary toolbox appears. The reason is that the new console (in particular the `LoggablesContext`) suspends and the next suspense wrapper in the component tree is around the secondary toolbox.
With this PR at least the other panels are usable while the console is still suspended. We should probably figure out how to make the console show up earlier and be (partially) usable as well.


markerikson (approved):

bvaughn:
Hey @hbenl I think we said we'd tweak the UI before merging this?

hbenl:
> Hey @hbenl I think we said we'd tweak the UI before merging this?

Oh, sorry, I misunderstood.

PR #7982, opened 2022-10-20
Update comments createLabel() thunk to more robustly handle source URLs

bvaughn:
Fixes the issue reported.

I'm pretty tired today though so please double check my logic.

### DIFF
--- a/src/ui/actions/comments.ts
+++ b/src/ui/actions/comments.ts
@@ -155,20 +155,23 @@ export function createLabels(
   return async (dispatch, getState, { ThreadFront, replayClient }) => {
     await ThreadFront.ensureAllSources();
     const state = getState();
-    const sourceId = handleUnstableSourceIds(sourceLocation.sourceUrl, state);
 
-    if (sourceId) {
-      sourceLocation = {
-        ...sourceLocation,
-        sourceId,
-      };
-    }
+    const { line, sourceUrl } = sourceLocation;
 
-    const { sourceUrl, line } = sourceLocation;
+    let sourceId: string | null = sourceLocation.sourceId || null;
 
     const filename = sourceUrl ? getFilenameFromURL(sourceUrl) : "unknown source";
+
     if (!sourceId) {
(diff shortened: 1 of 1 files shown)


src/ui/actions/comments.ts
```diff
@@ -155,20 +155,23 @@ export function createLabels(
...
-      };
-    }
+    const { line, sourceUrl } = sourceLocation;
 
-    const { sourceUrl, line } = sourceLocation;
+    let sourceId: string | null = sourceLocation.sourceId || null;
 
     const filename = sourceUrl ? getFilenameFromURL(sourceUrl) : "unknown source";
+
     if (!sourceId) {
```
bvaughn, line 165: I'm not sure why we weren't already doing this check. Was there a good reason, or was this an oversight?

It seems analogous to this code that @jcmorrow recently changed:
https://github.com/replayio/devtools/blob/79d06aab168237d525f40c25d068f92ef7c16001/src/devtools/client/debugger/src/actions/sources/select.ts#L138-L142

Am I missing something?

markerikson: I haven't poked at this code myself, but the diff _seems_ valid...

bvaughn:
Should we land this then?

markerikson:
Sure, go ahead!

src/ui/actions/comments.ts
jcmorrow: I'm not 100% sure I understand the change here, but it seems worth pointing out that even if we have an ID that doesn't mean it's the *correct* ID.

bvaughn: Thanks for that added context, Josh. I'll make that change before this lands then.

bvaughn: I think I've adjusted things now as discussed.

bvaughn:
Huh. Lots of tests are failing (both new e2e tests and the old screenshot tests). I'm not sure if this is because the backend is being slow or what. I would not have expected this change to make things fail.

PR #7983, opened 2022-10-20
Implement import sorting via `import-sort` and Prettier

markerikson:
This PR:

- Sets up infrastructure for sorting imports via https://github.com/renke/import-sort and https://github.com/ggascoigne/prettier-plugin-import-sort
- Adds a custom `import-sort-order-replay` sub-package with custom logic for ordering our imports in a specific sequence and set of groupings
- Formats a few files for bikeshedding purposes so we can see what the effects look like

At a high level this is based on my own general preference for ordering/grouping:

1) built-ins (like Node's path)
2) Major libs (React, Redux)
3) smaller libs (Lodash)
4) app-wide absolute imports ( ui/some/thing )
5) relative imports
6) assets

I've set up some custom grouping based on our own current package structure.  Here's an example of the current output:

```ts
import React, { Component, ContextType } from "react";
import { ConnectedProps, connect } from "react-redux";
import PropTypes from "prop-types";

import debounce from "lodash/debounce";

import type { UIState } from "ui/state";

import type { Modifiers } from "devtools/client/debugger/src/reducers/file-search";
import type { Context } from "devtools/client/debugger/src/reducers/pause";
import { getSelectedSource, getSourceContent, isFulfilled } from "ui/reducers/sources";

import actions from "../../actions";
import {
  getActiveSearch,
  getContext,
  getFileSearchModifiers,
  getFileSearchQuery,
  getFileSearchResults,
} from "../../selectors";
import { removeOverlay } from "../../utils/editor";
import { scrollList } from "../../utils/result-list";
import AccessibleImage from "../shared/AccessibleImage";
import { CloseButton } from "../shared/Button";
import SearchInput from "../shared/SearchInput";
import ShortcutsContext from "./ShortcutsContext";
```

So, the main questions are:

- What order do we _want_ to have?
- How soon do we pull the trigger and reformat everything?

Cue the bikeshedding!


bvaughn:
> At a high level this is based on my own general preference for ordering/grouping:
> 
> 1. built-ins (like Node's path)
> 1. Major libs (React, Redux)
> 1. smaller libs (Lodash)
> 1. app-wide absolute imports ( ui/some/thing )
> 1. relative imports
> 1. assets

This feels way too fine grained for me. I suspect many files will end up look like

```js
import A from "React";

import B from "package";

import C from "ui/some/thing";

import D from "../foo";
```

Can we combine the first 3 at least? So we'd have something more like:
> 1. Built-ins and external libs
> 1. App-wide absolute imports ( ui/some/thing )
> 1. relative imports

I don't feel too strongly about the "assets" import being its own thing (is this always just CSS?).

bvaughn:
TBH I kind of wish we'd kill relative imports too. I hate "../blah" and I _really hate_ "../../blah".

hbenl:
I'd also prefer less separators in the imports. We could either use less groups (like Brian suggested) *or* we could keep Mark's groups and not have separators between all of them.

bvaughn:
Are there specific files we should review when considering this PR?

For instance, is this `Toolbar` file sorted by the newly proposed rules?
https://github.com/replayio/devtools/blob/74932c278663d9300e18a419cd7fad313a74193a/src/ui/components/Toolbar.tsx#L1-L18

The _groupings_ look right, but _sorting within the groupings_ seems off.

markerikson:
Yeah, all the actual source files changed in this PR are examples of what it would look like.

Looking at that example, it seems correct based on the rules I've got defined.  I'm still trying to put `ui/setup/hooks` and `ui/state` at the top of the "app-wide imports" group.

The current definition of that section is:

```ts
    // App-wide globals

    // import { useAppDispatch } from "ui/setup/hooks"
    // import type { UIState } from "ui/state";
    {
      match: isGlobalReduxStoreConfig,
    },

    // import { ThreadFront } from "protocol/thread"
    {
      match: isProtocolModule,
    },
    // import { PointsContext } from "bvaughn-architecture-demo/src/contexts/PointsContext";
    {
      match: isNewPrototypeModule,
      sort: moduleName(naturally),
    },
    // import { Icon } from "design/Icon";
    {
      match: isOtherInternalPackageModule,
      sort: moduleName(naturally),
    },

    // Main app source absolute imports

    // import Popover from "devtools/client/debugger/src/components/shared/Popover";
    {
      match: isMainSrcFolder,
      sort: moduleName(naturally),
    },
    {
      match: isOtherSrcFolder,
      sort: moduleName(naturally),
    },
```

bvaughn:
> Looking at that example, it seems correct based on the rules I've got defined

I don't understand. The imports are _not sorted_.

Just looking at the first two lines. C _should_ come before R, right?
https://github.com/replayio/devtools/blob/74932c278663d9300e18a419cd7fad313a74193a/src/ui/components/Toolbar.tsx#L1-L2

markerikson:
That's also intentional - I'm used to having React first, then Redux, then other libs:

```ts

    // Start with side-effectful imports
    // import "foo"
    { match: and(hasNoMember, isAbsoluteModule) },

    // import "./foo"
    {
      match: and(hasNoMember, isRelativeModule, not(isStylesModule), not(isImageModule)),
    },

    // Built-ins
    // import … from "fs";
    {
      match: isNodeModule,
      sort: moduleName(naturally),
      sortNamedMembers: alias(unicode),
    },

    // Third-party libs

    // import React from "react";
    {
      match: isReactEcosystemModule,
      sort: moduleName(reactComparator),
      sortNamedMembers: alias(unicode),
    },

    // import uniq from 'lodash/uniq';
    {
      match: and(isInstalledModule(__filename), not(isRelativeModule)),
      sort: moduleName(naturally),
      sortNamedMembers: alias(unicode),
    },
    { separator: true },
```

The sorting is defined in `packages/import-sort-style-replay/import-sort-style-replay.js`

markerikson:
In other words, I'm trying to do some grouping by a sense of "priority" in each group, not just strictly alphabetically

bvaughn:
> In other words, I'm trying to do some grouping by a sense of "priority" in each group, not just strictly alphabetically

Why though? Seems to me like alpha-sorting is the easiest to understand. Is there any benefit to adding some extra (arbitrary?) sort criteria?

markerikson:
Habit, mostly?

My general principle for sorting is "most global -> most local/specific".

I've always listed React and Redux as the first imports in any file I've written.  Conceptually, I see them as "most global".

Similarly, for the "app-wide imports" area, the Redux state type and hooks are "most global" for the whole app.

If we want to have each group sorted 100% alphabetically, I can figure out how to do that.  But this is what I've used on all the projects I've worked on.

bvaughn:
My preference is that we just sort things alphabetically.

I understand some basic grouping (e.g. relative vs absolute– or even workspace absolute vs NPM absolute) but sorting by "which is more popular" feels really arbitrary to me.

In terms picking my battles, I'd list this as a _preference_ more than a _requirement_ FWIW.

markerikson:
Okay.  I'll see if I can figure out how to do them alphabetically per group.

bvaughn (approved):
Spot checked a few files and this looks awesome. Thanks!

PR #7984, opened 2022-10-20
Rework selected location checks to fix accidental scrolls to top

markerikson:
Per [this bug repro replay](https://app.replay.io/recording/fe-860-scroll-position-resets--fb1fb2e7-76d7-48bb-9ae7-052a2cc45e1d?point=29206669838329388894545426312593751&time=12759.55458898265&hasFrames=true&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6IjIzMzY1MzM1ODY4MzI4MTQxNTU2MTA1NzQ2MjE4NzUwNjM0IiwidGltZSI6MTA3NTIsInNjcmVlblNob3RzIjpbeyJtaW1lVHlwZSI6ImltYWdlL2pwZWciLCJoYXNoIjoiMTQ5MDkwNDk5NiJ9XX0sImVuZCI6eyJwb2ludCI6IjM1MDQ4MDAzODA5Mjk3NTY4NDQwNDY5NzMwMzExMDIwMjU4IiwidGltZSI6MTQ3NzcsImtpbmQiOiJtb3VzZWRvd24iLCJjbGllbnRYIjo1MTUsImNsaWVudFkiOjYyfX0%253D), we were trying to reset the scroll position whenever `selectedLocation.line` was 0, as it is when you normally switch tabs.

Compare to [this earlier working replay pre-Editor-rewrite](https://app.replay.io/recording/fe-860-scroll-position-was-working--c64eeeca-0d65-47d8-a323-6380f7035427?point=25312447189885280135493704105328944&time=9176.440991735537&hasFrames=true&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6IjEzMzA1MjYwNzAyODE2MzAyNzI2NTcwNzg1MDY1OTI3OTI4IiwidGltZSI6NTI3Nywic2NyZWVuU2hvdHMiOlt7Im1pbWVUeXBlIjoiaW1hZ2UvanBlZyIsImhhc2giOiItMTEwMTc2NzE2MCJ9XX0sImVuZCI6eyJwb2ludCI6IjMzMTAwODkyNDc4NTEwMTk0NDgyNTA5NjI5Njc0NzQ5OTUyIiwidGltZSI6MTE3MTd9fQ%253D%253D), where we bailed out in that same situation.

Looks like the issue was swiching from a basic falsy `if (!selectedLocation.line)` to `selectedLocation.line != null)`.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/useEditor.ts
+++ b/src/devtools/client/debugger/src/components/Editor/useEditor.ts
@@ -453,10 +453,11 @@ function scrollToLocationHelper(
   if (selectedLocation) {
     let shouldScrollToLocation = false;
     if (
-      selectedSource != null &&
-      selectedSourceContent != null &&
-      selectedLocation?.line != null &&
-      selectedSourceContent?.value != null
+      selectedSource &&
+      selectedSourceContent &&
+      // Intentionally bail out if line is 0
+      selectedLocation?.line &&
+      selectedSourceContent?.value
     ) {
       const isFirstLoad = prevSelectedSourceContent?.value == null;
       const locationChanged = prevSelectedLocation !== selectedLocation;


src/devtools/client/debugger/src/components/Editor/useEditor.ts
```diff
@@ -453,10 +453,11 @@ function scrollToLocationHelper(
...
     let shouldScrollToLocation = false;
     if (
-      selectedSource != null &&
-      selectedSourceContent != null &&
-      selectedLocation?.line != null &&
-      selectedSourceContent?.value != null
+      selectedSource &&
+      selectedSourceContent &&
+      // Intentionally bail out if line is 0
+      selectedLocation?.line &&
```
bvaughn, line 459: Is this really what we alway want?

Seems like there may be cases where we _want_ to scroll the editor to line 0 (e.g. if there's a comment there, or a log, etc)

jcmorrow: I thought lines were 1-indexed 🤔 (or at least, I'm pretty sure they are in protocol land, not sure about the editor)

bvaughn: I've found and fixed some broken conditionals where we had `!value` and sometimes bailed unintentionally when `value` was 0. I think "locations" are generally _protocol_ thing, and the protocol uses 1-based line numbers and 0-based column indices for some reason 😞

Anyway, in this case– lines must have been coming in as 0 in some cases or Mark's change would not have affected anything...right?

I dunno. It's late.

markerikson: Yeah, what I saw in both pre-`useEditor` code and current code is that when I was switching tabs, `selectedLocation.line` was 0.  Not sure if that's _always_ the case, but it was at least in these cases.

jcmorrow: > Anyway, in this case– lines must have been coming in as 0 in some cases or Mark's change would not have affected anything...right?

I don't think that's what this sounds like:

> Yeah, what I saw in both pre-useEditor code and current code is that when I was switching tabs, selectedLocation.line was 0

When you are switching tabs you are not navigating to a protocol-provided location, probably somewhere we just hard-coded `{line: 0, column: 0}` as the default location while switching tabs, but actually that's an invalid location that should never exist.

bvaughn: > When you are switching tabs you are not navigating to a protocol-provided location, probably somewhere we just hard-coded `{line: 0, column: 0}` as the default location

Ah, if so then I think we should just store `null` as the _default_ rather than an invalid/fake location. That would make more sense (and maybe make this weird check for an invalid value not relevant)

bvaughn: > > Anyway, in this case– lines must have been coming in as 0 in some cases or Mark's change would not have affected anything...right?
>
> I don't think that's what this sounds like:

Not to belabor the point, but I think that's exactly what it sounded like was happening.

PR #7987, opened 2022-10-21
Use the selected frame's pause for console evaluations (FE-857)

hbenl:
- use the currently selected frame for evaluations
- fix the console_async_eval e2e test:
  - one check for an async evaluation expected a value from a different frame than the selected one
  - one check for a failing evaluation expected an error message that was shown because the selected `frameId` did not exist in the current pause (because it was from a different pause), now we have a different error message because the evaluated expression throws (which is what this check was supposed to test in the first place)
- fix the `executeAndVerifyTerminalExpression()` test helper: it always succeeded if the expected evaluation result was identical to (or a substring of) the evaluated expression
- fix the console_eval e2e test: this test always failed because it tries to evaluate a (global) expression in the initial pause at the very end of the recording but that always fails in the backend, `executeAndVerifyTerminalExpression()` missed these failures in the past

### DIFF
--- a/packages/bvaughn-architecture-demo/components/console/Input.tsx
+++ b/packages/bvaughn-architecture-demo/components/console/Input.tsx
@@ -1,7 +1,7 @@
+import { SelectedFrameContext } from "@bvaughn/src/contexts/SelectedFrameContext";
 import { TerminalContext } from "@bvaughn/src/contexts/TerminalContext";
 import { TimelineContext } from "@bvaughn/src/contexts/TimelineContext";
 import useCurrentPause from "@bvaughn/src/hooks/useCurrentPause";
-import { FrameId, PauseId } from "@replayio/protocol";
 import { useContext, useEffect, useRef } from "react";
 
 import Icon from "../Icon";
@@ -25,14 +25,10 @@ export default function Input() {
     searchStateVisibleRef.current = searchState.visible;
   }, [searchState.visible]);
 
-  const pause = useCurrentPause();
-
-  let frameId: FrameId | null = null;
-  let pauseId: PauseId | null = null;
-  if (pause) {
-    frameId = pause.data.frames?.[0]?.frameId ?? null;
-    pauseId = pause.pauseId;
-  }
+  const currentPause = useCurrentPause();
+  const { selectedPauseAndFrameId } = useContext(SelectedFrameContext);
+  const frameId = selectedPauseAndFrameId?.frameId ?? currentPause?.stack?.[0] ?? null;
+  const pauseId = selectedPauseAndFrameId?.pauseId ?? currentPause?.pauseId ?? null;
 
   const onKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
     switch (event.key) {
--- a/packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
+++ b/packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
@@ -1,0 +1,66 @@
+import { FrameId, PauseId } from "@replayio/protocol";
+import { createContext, PropsWithChildren, Suspense, useContext, useMemo, useState } from "react";
+import isEqual from "lodash/isEqual";
+import useLoadedRegions from "../hooks/useRegions";
+import { ReplayClientContext } from "shared/client/ReplayClientContext";
+import { TimelineContext } from "./TimelineContext";
+import { isPointInRegions } from "shared/utils/time";
+import { getPauseForExecutionPoint } from "../suspense/PauseCache";
+
+interface PauseAndFrameId {
+  pauseId: PauseId;
+  frameId: FrameId;
+}
+
+export interface SelectedFrameContextType {
+  selectedPauseAndFrameId: PauseAndFrameId | null;
+  setSelectedPauseAndFrameId(pauseAndFrameId: PauseAndFrameId | null): void;
+}
+
+export const SelectedFrameContext = createContext<SelectedFrameContextType>({
+  selectedPauseAndFrameId: null,
+  setSelectedPauseAndFrameId: () => {},
+});
+
+export function SelectedFrameContextRoot({ children }: PropsWithChildren<{}>) {
+  const [selectedPauseAndFrameId, setSelectedPauseAndFrameId] = useState<PauseAndFrameId | null>(
+    null
+  );
+  const context = useMemo(
+    () => ({ selectedPauseAndFrameId, setSelectedPauseAndFrameId }),
+    [selectedPauseAndFrameId, setSelectedPauseAndFrameId]
+  );
+  return <SelectedFrameContext.Provider value={context}>{children}</SelectedFrameContext.Provider>;
+}
+
+function SelectedFrameContextAdapter() {
+  const client = useContext(ReplayClientContext);
+  const loadedRegions = useLoadedRegions(client);
+  const { executionPoint } = useContext(TimelineContext);
+  const { selectedPauseAndFrameId, setSelectedPauseAndFrameId } = useContext(SelectedFrameContext);
+
+  const isLoaded = loadedRegions !== null && isPointInRegions(executionPoint, loadedRegions.loaded);
+
+  const pause = isLoaded ? getPauseForExecutionPoint(client, executionPoint) : undefined;
+
+  const pauseId = pause?.pauseId;
+  const frameId = pause?.stack?.[0] ?? null;
+
+  const pauseAndFrameId = pauseId && frameId ? { pauseId, frameId } : null;
+  if (!isEqual(pauseAndFrameId, selectedPauseAndFrameId)) {
+    setSelectedPauseAndFrameId(pauseAndFrameId);
+  }
+
+  return null;
+}
+
+export default function SelectedFrameContextWrapper({ children }: PropsWithChildren<{}>) {
+  return (
+    <SelectedFrameContextRoot>
+      <Suspense>
+        <SelectedFrameContextAdapter />
+      </Suspense>
+      {children}
+    </SelectedFrameContextRoot>
+  );
+}
--- a/packages/e2e-tests/helpers/console-panel.ts
+++ b/packages/e2e-tests/helpers/console-panel.ts
@@ -84,7 +84,7 @@ export async function executeAndVerifyTerminalExpression(
 ): Promise<void> {
   await openConsolePanel(page);
   await executeTerminalExpression(page, text);
-  await verifyConsoleMessage(page, expected, "terminal-expression");
+  await verifyEvaluationResult(page, expected);
   if (clearConsoleAfter) {
     await clearConsoleEvaluations(page);
   }
@@ -220,12 +220,32 @@ export async function verifyConsoleMessage(
   );
 
   const messages = await findConsoleMessage(page, expected, messageType);
+  await verifyExpectedCount(messages, expectedCount);
+}
 
+export async function verifyEvaluationResult(
+  page: Page,
+  expected: Expected,
+  expectedCount?: number
+) {
+  await debugPrint(
+    page,
+    `Verifying the presence of an evaluation result with text "${chalk.bold(expected)}"`,
+    "verifyConsoleMessage"
+  );
+
+  const messages = page.locator(
+    `[data-test-name="TerminalExpression-Result"]:has-text('${expected}')`
+  );
+  await verifyExpectedCount(messages, expectedCount);
+}
+
+export async function verifyExpectedCount(locator: Locator, expectedCount?: number) {
   if (expectedCount != null) {
     // Verify a specific number of messages
     await waitFor(
       async () => {
-        const count = await messages.count();
+        const count = await locator.count();
         if (count !== expectedCount) {
           throw `Expected ${expectedCount} messages, but found ${count}`;
         }
@@ -238,7 +258,7 @@ export async function verifyConsoleMessage(
     );
   } else {
     // Or just verify that there was at least one
-    const message = messages.first();
+    const message = locator.first();
     await message.waitFor();
   }
 }
--- a/packages/e2e-tests/helpers/timeline.ts
+++ b/packages/e2e-tests/helpers/timeline.ts
@@ -92,3 +92,8 @@ export async function setFocusRangeStartTime(page: Page, timeString: string): Pr
   await clearTimeInput(page, input);
   await input.fill(timeString);
 }
+
+export async function seekToPreviousScreenshot(page: Page) {
+  const timeline = page.locator(".progress-bar");
+  await timeline.press("ArrowLeft");
+}
--- a/src/ui/components/SecondaryToolbox/NewConsole.tsx
+++ b/src/ui/components/SecondaryToolbox/NewConsole.tsx
@@ -2,13 +2,13 @@ import NewConsole from "bvaughn-architecture-demo/components/console";
 import { ConsoleSearchContext } from "bvaughn-architecture-demo/components/console/ConsoleSearchContext";
 import { TerminalContext } from "bvaughn-architecture-demo/src/contexts/TerminalContext";
 import React, { KeyboardEvent, useContext } from "react";
-import { getPauseId, getSelectedFrameId } from "devtools/client/debugger/src/selectors";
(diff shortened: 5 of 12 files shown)


packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+    frameId: FrameId;
+  } | null;
+}
+
+export const SelectedFrameContext = createContext<SelectedFrameContextType>({
+  selectedFrameId: null,
+});
+
+export function SelectedFrameContextRoot({ children }: PropsWithChildren<{}>) {
+  const pause = useCurrentPause();
```
bvaughn, line 17: `useCurrentPause` _suspends_, so this change means that we might Suspend pretty much at the top of the DevTools tree. I'm a little wary of that. If we mess up an update (don't wrap it in a transition) then it could cause the entire app to unmount and remount.

Maybe we should add a guard here?

What if we replaced `useCurrentPause` with a new hook, `usePauseData`, and then we passed in a deferred pause to that hook here to avoid accidentally suspending during a high priority update?

hbenl: Thanks, I overlooked that (and will fix it). I think we should add `Suspense` to the names of all functions that can suspend.

bvaughn: Yeah, that's probably a good call. This hook should either be called `useCurrentPauseSuspends` to make it more obvious, or we should just remove the _suspending_ part.

Maybe we could split this hook into something smaller that _doesn't_ suspend but is still helpful, like...
```js
function useExecutionPointIsLoaded(executionPoint: ExecutionPoint): boolean {
  const loadedRegions = useLoadedRegions(client);

  const isLoaded = loadedRegions !== null && isPointInRegions(executionPoint, loadedRegions.loaded);
  return isLoaded;
}
```

Then you could decide whether to Suspend or not like
```js
const { executionPoint } = useContext(TimelineContext);
const isLoaded = useExecutionPointIsLoaded(executionPoint);
let pause = null;
if (isLoaded) {
  pause = getPauseForExecutionPoint(client, executionPoint);
}
```

hbenl: But `getPauseForExecutionPoint()` can still suspend in that case because we may not have a `pauseId` for that execution point yet (and have to call `Session.createPause` to get it).

hbenl: What if the context offered a suspending function `getSelectedFrameId()`, so it wouldn't suspend itself, only the components that use the function would...?

bvaughn: > But `getPauseForExecutionPoint()` can still suspend in that case because we may not have a `pauseId` for that execution point yet (and have to call `Session.createPause` to get it).

Yes. That's why I was suggesting that part _doesn't_ go into a hook. (So it's _clear_ when you choose to Suspend.)

Basically, I think having a hook that _suspends_ was probably not a good idea.

> What if the context offered a suspending function `getSelectedFrameId()`, so it wouldn't suspend itself, only the components that use the function would...?

I'm not sure what you're suggesting.

hbenl: > I'm not sure what you're suggesting.

What I did in this commit: https://github.com/replayio/devtools/pull/7987/commits/df483c26e594195efa3aa77c0fe10cac2e4a9b1c

packages/e2e-tests/helpers/timeline.ts
```diff
@@ -92,3 +92,8 @@ export async function setFocusRangeStartTime(page: Page, timeString: string): Pr
   await clearTimeInput(page, input);
   await input.fill(timeString);
 }
+
+export async function seekToPreviousScreenshot(page: Page) {
```
bvaughn, line 96: Why is this called "previous screenshot" ?

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const pauseId = pause?.pauseId;
+    const frameId = pause?.data.frames?.[0]?.frameId ?? null;
+
+    return pauseId && frameId ? { pauseId, frameId } : null;
+  }
+
+  const context = { getSelectedFrameId };
+
+  return <SelectedFrameContext.Provider value={context}>{children}</SelectedFrameContext.Provider>;
+}
```
bvaughn, line 39: I don't understand why this is a _context_. Contexts are stateful things– that you can use to set state from one component(s) and share it with another component(s) in a different part of the tree. This doesn't look like it has any state.

Couldn't `getSelectedFrameId` just be a regular hook (`useSelectedFrameId`)?

(I think it's also generally a good idea to memoize context values (with `useMemo` or `useCallback`) so that subscribers don't over-render.)

bvaughn (commented):
I think it might make sense to add a selected-frame context, but as a stateful thing (like our other contexts).

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
bvaughn: More succinctly, what I'm suggesting is:
* `SelectedFrameContext` has two values: `selectedFrameId: FrameId | null` and `setSelectedFrameId(frameId: FrameId | null)`.
  * New architecture components _read_ `context.selectedFrameId`
* `SelectedFrameContextAdapter` updates `SelectedFrameContext` when Redux values change (like our other adapters do).

packages/e2e-tests/helpers/timeline.ts
hbenl: It presses the left arrow key on the timeline, which seeks to the execution point of the last screenshot before the current execution point.

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
hbenl: > I don't understand why this is a context.

I'm using the context as a dependency injection mechanism: the `SelectedFrameContext` is used by components in the architecture demo package, which doesn't have access to the redux store that contains `selectedFrameId`. So in the architecture demo we use the `SelectedFrameContextRoot` from this file but in devtools we use `SelectedFrameContextAdapter`, which provides `selectedFrameId` from the redux store.

> I think it's also generally a good idea to memoize context values

Agreed, I just couldn't figure out how to memoize a *suspending* value.

> what I'm suggesting is: [...]

Sounds good, that would also fix the memoization issue.

hbenl:
I'm trying to figure out why some playwright tests from the architecture demo fail. I'm making recordings of failing tests but RUN-715 makes it practically impossible to work with them.

src/ui/components/SelectedFrameContextAdapter.tsx
```diff
@@ -0,0 +1,11 @@
+import { SelectedFrameContext } from "bvaughn-architecture-demo/src/contexts/SelectedFrameContext";
+import { getSelectedFrameId } from "devtools/client/debugger/src/selectors";
+import { PropsWithChildren, useContext } from "react";
+import { useAppSelector } from "ui/setup/hooks";
+
+export default function SelectedFrameContextAdapter({ children }: PropsWithChildren) {
+  const selectedFrameId = useAppSelector(getSelectedFrameId);
+  const { setSelectedFrameId } = useContext(SelectedFrameContext);
+  setSelectedFrameId(selectedFrameId);
```
bvaughn, line 9: React is bailing out on equality for us here, to avoid an infinite loop, but it would be nice if we conditionally called this instead, just to be clearer:

```suggestion
  const selectedFrameIdRedux = useAppSelector(getSelectedFrameId);
  const { selectedFrameId, setSelectedFrameId } = useContext(SelectedFrameContext);
  if (selectedFrameId !== selectedFrameIdRedux) {
    setSelectedFrameId(selectedFrameId);
  }
```

src/ui/components/SecondaryToolbox/NewConsole.tsx
```diff
@@ -85,9 +83,9 @@ function JSTermWrapper() {
     addMessage({
       expression,
       frameId: frameId || null,
-      pauseId,
-      point,
-      time,
+      pauseId: pause.pauseId,
+      point: pause.point!,
+      time: pause.time!,
```
bvaughn, line 88: Is it safe to assume we'll always have a `point` and `time` here?

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface PauseAndFrameId {
+  pauseId: PauseId;
+  frameId: FrameId;
+}
+
+export interface SelectedFrameContextType {
+  selectedFrameId: PauseAndFrameId | null;
+  setSelectedFrameId(pauseAndFrameId: PauseAndFrameId | null): void;
+}
```
bvaughn, line 17: I think we should split `selectedFrameId` into two values to avoid confusion:
```suggestion
export interface SelectedFrameContextType {
  frameId: FrameId | null;
  pauseId: PauseId | null;
  setSelectedFrame(
    pauseId: PauseId | null,
    frameId: FrameId | null,
  ): void;
}
```

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const pauseId = pause?.pauseId;
+  const frameId = pause?.stack?.[0] ?? null;
+
+  useEffect(() => {
+    const pauseAndFrameId = pauseId && frameId ? { pauseId, frameId } : null;
+    setSelectedFrameId(pauseAndFrameId);
+  }, [setSelectedFrameId, pauseId, frameId]);
+
+  return null;
+}
```
bvaughn, line 52: Looks like `SelectedFrameContextAdapter` _suspends_ (when it calls `getPauseForExecutionPoint`) but then it uses an effect to update context values.

If we keep the _suspending_ behavior, then it seems like the effect is unnecessary (and we should add a `Suspense` wrapper to avoid it suspending the context and its entire subtree).

But if we want to avoid suspending we could add/use an async "helper" to the pause cache to load the data (so the effect loads and updates the context).

What are your thoughts?

packages/bvaughn-architecture-demo/components/console/Input.tsx
```diff
@@ -25,14 +25,10 @@ export default function Input() {
...
-
-  let frameId: FrameId | null = null;
-  let pauseId: PauseId | null = null;
-  if (pause) {
-    frameId = pause.data.frames?.[0]?.frameId ?? null;
-    pauseId = pause.pauseId;
-  }
+  const currentPause = useCurrentPause();
+  const { selectedFrameId } = useContext(SelectedFrameContext);
+  const frameId = selectedFrameId?.frameId ?? currentPause?.stack?.[0] ?? null;
```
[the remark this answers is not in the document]
hbenl: Yes

src/ui/components/shared/CodeEditor/useAutocomplete.tsx
```diff
@@ -118,19 +118,17 @@ function useGetEvalMatches(expression: string) {
...
-
   const getEvalMatchesForSelectedFrame = useCallback(
     async (expression: string) => {
-      if (!pauseId) {
+      if (!selectedFrameId) {
         return null;
       }
       const fetchObject = async (objectId: string) => {
-        return getObjectWithPreviewHelper(replayClient, pauseId, objectId);
+        return getObjectWithPreviewHelper(replayClient, selectedFrameId.pauseId, objectId);
```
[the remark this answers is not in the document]
hbenl: I don't like `selectedFrame` either because it isn't a frame, only the IDs that are necessary to uniquely identify the frame. I've used `PauseAndFrameId` in a couple of places because I couldn't think of a better name.

src/ui/components/SecondaryToolbox/NewConsole.tsx
hbenl: Yes, those are only set to `null` briefly during the initialisation of the Pause, but by the time the Pause is added to `pausesById` (which is where we get this Pause from), they are always updated.

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
hbenl: I combined these values so that the type makes it clear that we either have a `pauseId` and a `frameId` or we have neither (because having just one of these IDs doesn't make sense).

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
hbenl: > If we keep the suspending behavior, then it seems like the effect is unnecessary

I added the effect because I thought render functions should not call a setter.

> and we should add a Suspense wrapper to avoid it suspending the context and its entire subtree

`SelectedFrameContextAdapter` doesn't take any `children`, so there is no subtree.

Why do you want to avoid suspending `SelectedFrameContextAdapter`?

src/ui/components/shared/CodeEditor/useAutocomplete.tsx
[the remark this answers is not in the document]
bvaughn: Then call it `selectedPauseAndFrameId` (or just store them separately as `selectedPauseId` and `selectedFrameId`?

Calling it `selectedFrameId` seems really confusing though because it's _not_ a frame id.

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
bvaughn: I don't think the way it is structured _is_ clear though. Maybe renaming the main thing could help.

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
bvaughn: Suspending during render just to call an update function in an effect seems awkward to me.

Functions can call a state updater during render. It's not _ideal_ because it restarts some of the render work, but it can be done. (That's what you're doing in the other `SelectedFrameContextAdapter` component.)

src/ui/components/shared/CodeEditor/useAutocomplete.tsx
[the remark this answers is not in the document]
hbenl: I've renamed it to `selectedPauseAndFrameId`

packages/bvaughn-architecture-demo/src/contexts/SelectedFrameContext.tsx
hbenl: > Functions can call a state updater during render.

I didn't know that, thanks.
So I removed the effect and also added a guard so that we only call `setSelectedPauseAndFrameId()` if it really changed.
And I added a Suspense wrapper because I realized that's necessary (otherwise it would suspend the entire subtree of the next Suspense wrapper, not just its own (empty) subtree).

packages/e2e-tests/helpers/console-panel.ts
```diff
@@ -220,12 +220,32 @@ export async function verifyConsoleMessage(
...
+    "verifyConsoleMessage"
+  );
+
+  const messages = page.locator(
+    `[data-test-name="TerminalExpression-Result"]:has-text('${expected}')`
+  );
+  await verifyExpectedCount(messages, expectedCount);
+}
 
+export async function verifyExpectedCount(locator: Locator, expectedCount?: number) {
```
bvaughn, line 243: nit: Seems odd to me that we have a method named `verifyExpectedCount` that takes an _optional_ `expectedCount` param. What does it mean to call this method without that param?

Seems like we should just adjust the call sight in `verifyEvaluationResult` to only call this if the count is passed in.

bvaughn (approved):

PR #7988, opened 2022-10-21
Source viewer

bvaughn:
This PR only touches things in the bvaughn architecture package– stuff related to the new source viewer.

- New points panel editing functionality
  - Typeahead suggestions
- Fixed some popup positioning issues


bvaughn:
Two of the three failing e2e tests pass for me locally. The one that doesn't (stacking) seems flaky 😞  because it fails on `main` for me locally as well.

```
  1) [chromium] › tests/stacking.test.ts:223:1 › stacking: Element highlighter selects the correct element when they overlap

    Error: Incorrect rules found for test case: r3c1

    - Expected  - 4
    + Received  + 1

    - Array [
    -   "element",
    -   "body > div",
    - ]
    + Array []
```

PR #8004, opened 2022-10-24
Fix hit count tooltip logic and clarify variable names

markerikson:
- Renamed `hitsCounts` to `hitCountsForAllLines`
- Renamed `hitsCount` to `hitsForHoveredLine`
- Fixed apparent typo with `if (hitsCount == null)` case
- Re-added `if (hitsForHoveredLine == null) return null` case

I checked the earlier implementation of the component and noted that:

- It used to show "Loading..." if the lookup table of _all_ hit counts for this file was null, whereas prod currently checks if `hitsCount` for _the hovered line_ is null-ish
- It used to bail out entirely if the hovered line was null-ish

I think the similarity between the variables `hitsCounts` and `hitsCount` caused a typo in the "Loading" comparison, and we lost the "bail out if no hits for line" case.

I _think_ this should fix the issue with the hit counts tooltip showing "Loading..." when it shouldn't have been.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
@@ -64,38 +64,40 @@ function LineNumberTooltipSuspends({ keyModifiers }: Props) {
   const replayClient = useContext(ReplayClientContext);
   const { range: focusRange } = useContext(FocusContext);
 
-  const hitsCounts =
+  const hitCountsForAllLines =
     source && visibleLines
       ? getSourceHitCounts(replayClient, source.id, visibleLines, focusRange)
       : null;
 
-  let hitsCount: number | null = null;
-  if (hoveredLineIndex != null && hitsCounts != null) {
+  let hitsForHoveredLine: number | null = null;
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/components/Editor/LineNumberTooltip.tsx
```diff
@@ -64,38 +64,40 @@ function LineNumberTooltipSuspends({ keyModifiers }: Props) {
...
+    trackEvent(hitsForHoveredLine ? "breakpoint.preview_has_hits" : "breakpoint.preview_no_hits");
+    trackEvent("breakpoint.preview_hits", { hitsCount: hitsForHoveredLine });
+  }, [hitsForHoveredLine]);
 
   if (isMetaActive) {
     return null;
   } else if (hoveredLineIndex === null || hoveredLineNode === null) {
     return null;
-  } else if (hitsCount == null) {
+  } else if (hitCountsForAllLines == null) {
```
markerikson, line 91: :speech_balloon: this should be the key fix here, and then re-added the "don't show anything if there's no hits for the line" check after that

bvaughn (approved):
👍🏼

PR #8005, opened 2022-10-24
Hide Network requests in unloaded regions

bvaughn:
This is what I propose:
* We continue filtering/hiding Network requests based _only_ on whether they are inside of outside of the user's _focus region_.
* We continue to _dim_ Network requests that are outside of the _loaded region_ (since you can't inspect them or interact with them quite as fully).
  * Side note: We may want to update our visual styling here so it stands out more from how we render Network requests that are _after the current execution point_. (We also dim those.)
* We add an explicit header footer with the number of requests filtered before/after.

https://user-images.githubusercontent.com/29597/199289378-4161f91c-eb7b-4657-a068-46cd9f4ff62e.mp4

Demo walk through:
https://www.loom.com/share/05322d15d2964f2ebec130f880f19513

### DIFF
--- a/src/ui/reducers/network.ts
+++ b/src/ui/reducers/network.ts
@@ -7,17 +7,16 @@ import {
 import sortBy from "lodash/sortBy";
 import sortedUniqBy from "lodash/sortedUniqBy";
 import { createSelector } from "reselect";
+import { isPointInRegions } from "shared/utils/time";
 import { NetworkAction } from "ui/actions/network";
 import { partialRequestsToCompleteSummaries } from "ui/components/NetworkMonitor/utils";
 import { UIState } from "ui/state";
 import {
+  displayedBeginForFocusRegion,
   displayedEndForFocusRegion,
   filterToFocusRegion,
-  displayedBeginForFocusRegion,
-  filterToLoadedRegions,
 } from "ui/utils/timeline";
 
-import { getLoadedRegions } from "./app";
 import { getFocusRegion } from "./timeline";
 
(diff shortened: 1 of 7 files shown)


src/ui/reducers/network.ts
```diff
@@ -105,10 +107,20 @@ export const getFocusedEvents = (state: UIState) => {
   return events.filter(e => e.time > beginTime && e.time <= endTime);
 };
 export const getFocusedRequests = (state: UIState) => {
```
markerikson, line 109: :hand: Hmm. Not sure what this was doing before, but this looks like it probably needs to be a memoized selector - right now it seems to be returning new array references every time, either from filtering or the empty array

bvaughn: Agreed. Let's memoize that (and `getFocusedEvents`) in a follow up commit though since it's not new behaiovr?

bvaughn: Nevermind. I'll just do it while I'm in there. Nbd.

markerikson (approved):

bvaughn:
Going to hold off merging this for now. @jonbell seems to have some thoughts/concerns that he wants to talk over first.

bvaughn:
After some discussion today, the goal for this PR/issue is to filter out (hide) Network requests that are outside of _either_ the focus region _or_ the loaded region(s).

bvaughn:
Tests seem to be timing out. Not sure if related to this branch or a backend issue.

PR #8010, opened 2022-10-25
Don't assume Message.data.frames is sorted (FE-867)

hbenl:


packages/bvaughn-architecture-demo/src/utils/messages.ts
```diff
@@ -14,14 +14,14 @@ export function isInNodeModules(message: ProtocolMessage): boolean {
   if (messageToNodeModuleCache.has(message)) {
     return messageToNodeModuleCache.get(message)!;
   } else {
-    let lastFrame = null;
-    if (message.data.frames) {
-      lastFrame = message.data.frames[message.data.frames.length - 1];
+    let topFrame = null;
+    if (message.stack?.length && message.data.frames?.length) {
+      topFrame = message.data.frames.find(frame => frame.frameId === message.stack![0]);
```
bvaughn, line 19: I wonder if we should make a helper function e.g. `getTopFrame(frames, stack)`

Repeating this code in two places isn't bad though I guess

bvaughn (approved):

PR #8021, opened 2022-10-27
Fix rendering of Maps and Sets without a preview (FE-809)

hbenl:
Maps and Sets for which no preview had been loaded yet were rendered as `Map (0)` / `Set (0)`, now they are rendered as `Map` / `Set`.
Furthermore, empty Maps/Sets now have the same styling applied as non-empty ones.


markerikson:
Probably need to update the screenshots for this.

Also, TS seems to be unhappy with some of these changes :)

hbenl:
@markerikson Thanks, PR has been updated.

PR #8022, opened 2022-10-27
Small UI tweaks to sizes/paddings/animations (to prepare for opt-in)

bvaughn:
- [x] Replaced REMs with EMs to make sizing more consistent between test harness and legacy app.
  - [x] Tweaked paddings and button/icon sizes accordingly.
- [x] Added point panel hover tooltip to show what time is being hovered over.
- [x] Update screenshot tests (currently blocked b'c of a Docker issue).


bvaughn:
Hmm, looks like `point-panel-default-badge-expected.png` didn't match when CI ran.

bvaughn:
Yeah, unfortunately this is still not landable. Tried restarting the Playwright tasks, but some of the screenshots are still not matching.

* tests/source-and-console.ts:153:1 › should handle too many points to find
* tests/source-and-console.ts:283:1 › should allow log point badge colors to be toggled

bvaughn:
Looks like the failing Playwright test is related to a regression introduced in #8029

PR #8024, opened 2022-10-28
Fix FocusRegion persistence (FE-868)

hbenl:
Fixes two issues that prevented the focus region from being persisted and reapplied:
- the `focusRegion` URL parameter was not set until the user jumped to a new point - now it is set whenever the focus region changes
- the initialization of the focus region from its URL parameter always created a focus region spanning the entire recording
  (it would still ask the backend to load the correct region, but the new region would be saved to the URL and so after reloading
  devtools it would "focus" on the entire recording)

Some cleanups:
- removed the `UnsafeFocusRegion` type
- most of the `TimeStampedPoint`s stored in `state.timeline.points` had additional properties that would then be copied into the focus region and from there into the URL - now they are cleaned up


markerikson (approved):
Seems okay to me, but I'm not sure what Brian had in mind with that `UnsafeFocusRegion` bit.

PR #8027, opened 2022-10-28
Add Suspense to the names of all suspending functions (FE-863)

hbenl:
Unlike what we said in the last architecture sync I have also added `Suspense` to functions in the `suspense` folder because I think otherwise it's too easy to overlook.


markerikson (approved):

markerikson:
Seems good, pending lint/test fixes

PR #8029, opened 2022-10-28
Actually fix problems with obsolete source IDs in persisted points

markerikson:
This PR:

- Swaps the mechanism of fetching sources to be done via Suspense cache and `ReplayClient`, instead of inside `ThreadFront`
  - The main app code that called `ThreadFront.findSources()` now calls `getSourcesHelper()` instead
  - Since a bunch of code relies on `await ThreadFront.ensureAllSources()`, that same setup logic now resolves the promise in `ThreadFront`
- Updates `useBreakpointIdsFromServer` to wait for sources before trying to apply persisted breakpoints, and skips over + removes any points whose source location is not in the set of actual source IDs
- Tosses in a `yarn.lock` diff that got missed earlier

### DIFF
--- a/packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
+++ b/packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
@@ -10,6 +10,7 @@ import {
   useState,
   useTransition,
 } from "react";
+import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import { Point, PointId } from "shared/client/types";
 
 import useBreakpointIdsFromServer from "../hooks/useBreakpointIdsFromServer";
@@ -40,6 +41,7 @@ export const PointsContext = createContext<PointsContextType>(null as any);
 
 export function PointsContextRoot({ children }: PropsWithChildren<{}>) {
   const { recordingId } = useContext(SessionContext);
+  const replayClient = useContext(ReplayClientContext);
 
   // Both high-pri state and transition state should be managed by useLocalStorage,
   // Else values from other tabs will only be synced to the high-pri state.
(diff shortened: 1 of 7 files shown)


bvaughn (approved):
I'm not sure I understand why moving the source loading into the Replay client was necessary vs the way it was being done before, but I trust your research there.

markerikson:
For reference:

Source fetching was _previously_ done inside of `ThreadFront.findSources()`, which also tied into the promise returned by `ThreadFront.ensureAllSources()`.

In order to make the `useBreakpointIdsFromServer` hook wait for sources to load / get the cached list of sources, I needed to keep `ThreadFront` out of the equation, since `useBreakpointIdsFromServer ` and `PointsContext` live over in the prototype folder and know nothing about `ThreadFront`.

I was actually pleasantly surprised at how little code actually had to change to make that work.

Besides, this also has the bonus architectural improvement benefits of being one less thing that `ReplayClient` _needs_ to use `ThreadFront` for, and one less responsibility _inside_ `ThreadFront`.

bvaughn:
I think the core of the fix was making the sync hook _wait_ until sources had loaded– not specifically keeping thread front uninvolved.

So I guess I'm not sure why the sync hook couldn't just wait for all sources to be loaded (or even just assume they already had been, since we load as part of boot strapping).

Not saying that change was bad. I just don't think I understood why _that bit_ was needed.

markerikson:
It's specifically the "wait for all sources to be loaded" that was the actual issue.

That knowledge previously existed in two places:

- `protocol/thread/thread.ts`, in `ThreadFront`
- `src/ui/reducers/sources.ts`, in the Redux store

`PointsContext.ts` is over in the `prototype` folder and doesn't know about either of those, and we've been intentionally keeping the prototype folder from importing anything in the main app.

So, by flipping where we do the fetching and where the knowledge of "do we have sources yet" lives, I was able to rework the hook to wait for the sources.  Plus, it needed _access_ to the entire list of sources as well, with the same problems architecturally.

bvaughn:
I think this change broke the new Source viewer– specifically some of the Playwright tests I added in #8022.

After this change, clicking "+" to add a new breakpoint just permanently sticks "loading.."

bvaughn:
Here's the fix: b8e10e2

Need to connect the Replay client and `ThreadFront` at a different place so they couldn't accidentally be out of sync.

PR #8023, opened 2022-10-28
Add streaming source cache API

bvaughn:
Sibling PR for https://github.com/replayio/backend/pull/6537

- [x] #8066
- [x] #8067
- [x] Use the new `Debugger.streamSourceContents` to load Source contents
- [x] Add new Suspense cache to incrementally parse contents as they stream in
- [x] Remove unused `maxChunkSize` param

PR #8031, opened 2022-10-30
Streaming source viewer spike

bvaughn:
Builds on top of #8023 (but with hard-coded source contents) to demo what an integration with the new streaming source API could look like.


bvaughn:
This branch has served its purpose.

PR #8033, opened 2022-10-31
Add opt-in experimental setting for the new Source viewer

bvaughn:
- [x] Add experimental opt-in flag
- [x] Source viewer should be integrated with...
  - [x] Source explorer
  - [x] CMD+O/CMD+G menu
  - [x] Console seek-to and view-source links
  - [x] Outline view
- [x] "Add comment" button in log Point panel should set the source code preview
- [x] Wire up "enable large text" preference


bvaughn:
I think this is ready for review.

markerikson:
Code in the PR seems generally fine.

Doing side-by-side comparisons between the existing editor and the new viewer, I'm seeing a couple actual bugs, and some visual differences that are more "what do we want it to be?"

- **Bugs**
  - Variable hover tooltips are not showing up when I am paused at a point
  - The default new print statement content string is the name of the file, when it _should_ be the nearest enclosing function name
  - When I switched the editor pref, the new viewer showed up okay, but after adding one print statement, hovering over any line showed me the "Continue to here" hover button instead of the "Add print statement" button
  - Viewer seems to have a horizontal scrollbar that is stuck permanently on, and showing ~95% scroll width.  Dragging the editor/viewer scroll pane back and forth resizes the editor, but the scrollbar never goes away.
- **Visual differences**
  - Some differences in syntax highlighting (both color, and which tokens)
  - Some slight font rendering differences
  - Hit count color bars are coming in with different gradients.  Same recording, same file, same lines, but they're showing up as "light" in the old editor and "dark" in the new one
  - New viewer hit count color bar is wider/thicker

Old:

image

New:

image

bvaughn:
### Bugs
- [ ] Variable hover tooltips are not showing up when I am paused at a point
- [ ] @markerikson: The default new print statement content string is the name of the file, when it should be the nearest enclosing function name
- [ ] @bvaughn: When I switched the editor pref, the new viewer showed up okay, but after adding one print statement, hovering over any line showed me the "Continue to here" hover button instead of the "Add print statement" button
- [ ] Viewer seems to have a horizontal scrollbar that is stuck permanently on, and showing ~95% scroll width. Dragging the editor/viewer scroll pane back and forth resizes the editor, but the scrollbar never goes away.

PR #8063, opened 2022-11-02
Enable tracking telemetry events in new codebase

markerikson:
This PR:

- Injects `trackEvent` from the main codebase into the new codebase via `SessionContext`
- Re-adds Mixpanel telemetry events for breakpoints that were lost during the "Points" refactor PR in #7951


jonbell-lot23:
I have nothing to add to the review, but thank you! <3

PR #8075, opened 2022-11-04
Remove lingering references to @bvaughn

bvaughn:
- [x] Deleted some references to `@bvaughn` in TS config files. These caused Code to auto-import using the `@bvaughn` prefix, which breaks import sorting.
- [x] Fixed typo in Storybook config file.
- [x] Updated couple of import statements that had slipped in with the `@bvaughn` prefix.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -32,7 +32,7 @@ module.exports = {
       "@recordreplay/accordion": path.resolve(__dirname, "../packages/accordion/index.tsx"),
       "third-party": path.resolve(__dirname, "../packages/third-party"),
       "packages/third-party": path.resolve(__dirname, "../packages/third-party"),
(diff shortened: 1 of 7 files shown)


.storybook/main.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
       "@recordreplay/accordion": path.resolve(__dirname, "../packages/accordion/index.tsx"),
       "third-party": path.resolve(__dirname, "../packages/third-party"),
       "packages/third-party": path.resolve(__dirname, "../packages/third-party"),
-      "@bvaughn": path.resolve(__dirname, "../packages/bvaugh-architecture-demo"),
```
bvaughn, line 35: This is why the `@bvaughn` prefix never worked for Storybook 🤣

markerikson (approved):

PR #8078, opened 2022-11-04
[DRAFT] Improve display of large strings via truncation

markerikson:
This PR:

- Attempts to improve display of large strings in the Object Inspector by changing the output based on whether the parent is collapsed or open:
  - When collapsed:
    - Replacing actual newlines with `"\n"` literal text
    - Inner-truncating the string to 100 characters (current replacement text: ` ...*truncated*... ` )
  - When open:
    - Going back to showing the string with actual line breaks
    - Showing an additional disclosure toggle that switches between still truncating at 250 characters, or showing the _entire_ string

The logic works, but there's a display issue when the string toggle is expanded - the object key on the left gets pushed down out of sight:

String toggle closed:

image

Expanded fully:

image
image


bvaughn:
Thanks for sharing this idea! 😄 Initial thought: showing an explicit "truncated" string seems odd to me.

Screen Shot 2022-11-04 at 2 46 43 PM

For value-strings (in an object/map/whatever) Chrome just shows all of the text on one line, truncated (without even a "...") with two buttons: "Show more" and "Copy"

markerikson:
Hmm.  Odd, both of those look different than what I was seeing in my own Chrome.  (I may be using a slightly older version, so perhaps that changed recently?)

Jon was the one who specifically suggested using inner truncation over in the discussion thread.

jonbell-lot23:
I propose this:

[...]

I could be convinced into this as a plan B:

...

But in my view, showing nothing at all feels broken. I can't explain why Chrome does it, but it doesn't seem ideal.

jonbell-lot23 (approved):

bvaughn:
I'm also not _opposed_ to inner truncation. I dislike "...\*truncated\*..." though. If we truncate, I think we should only use an ellipsis. I think that's standard enough that it shouldn't need further explanation.

That being said, I don't personally thing we have a super strong reason to diverge from what Chrome does here, and I think adding the truncation in the middle makes the UX of loading the rest more awkward. For example, if the thing you're looking for is in the middle, you'd have to scroll to the end to see the "load more" button, right? then scroll back?

markerikson:
Part of my concern there is that an ellipsis is identical to, say, a `{...someFields}` operator :)  so it doesn't particularly stand out.

jonbell-lot23:
Let’s push this out without truncation at all and we’ll see how that feels. Then truncation can be a later discussion with more real-world data.On 6/11/2022, at 3:25 AM, Mark Erikson ***@***.***> wrote:﻿
Part of my concern there is that an ellipsis is identical to, say, a {...someFields} operator :)  so it doesn't particularly stand out.

bvaughn:
@jonbell-lot23 So the incremental change you're asking for here is just to change the collapsed string view to replace line breaks with the text "\n" ?

bvaughn:
> Part of my concern there is that an ellipsis is identical to, say, a {...someFields} operator :) so it doesn't particularly stand out.

@markerikson Fair enough! I think though that when it comes _at the end_ of the string, beside a "load more" button, it's probably fairly intuitive. But maybe this is why Chrome doesn't show it?

jasonLaster:
love this conversation

markerikson:
I'm confused where we're landing on this :)

I think there's a few questions to answer atm:

- Do we want to "inner truncate" in the middle, or truncate at the end?
- What should the truncation text be?
- For the "show more" use case, do we want to use an expander toggle same as the `<Expandable>` component does now, or some kind of a "Show More" button?

bvaughn:
Maybe we should chat about this during the Design Sync today?

Here's my two cents though:
> Do we want to "inner truncate" in the middle, or truncate at the end?

Do what Chrome does. (Truncate at the end.)

> What should the truncation text be?

Do what Chrome does. (No truncation text.)

> For the "show more" use case, do we want to use an expander toggle same as the <Expandable> component does now, or some kind of a "Show More" button?

Do what Chrome does. (Inline show more button.)

markerikson:
Conclusions after talking with Jon:

- We'll go with end truncation
- We'll use `[...]` as the truncation text
- We'll add a "Show More" button, at the end, next to the truncation
- Jon will provide me with the Tailwind styles for a subtle-appearing button (ie, grays-ish rather than primary blue)

jonbell-lot23:
Two more details:

1. We need messaging for when the item is copied or pasted.
2. It'd be great to make that its own component

If you look into [this stalled PR](https://github.com/replayio/devtools/pull/7943/files), you can see the code I used to do this in our cURL functionality:


Here's the element itself:
```
<div id="showCopied" className="mb-1.5 flex shrink rounded-lg bg-black p-1.5 text-center text-white opacity-100 shadow-2xl transition-all duration-700 ease-in-out">Copied to clipboard</div></div>
```             

And here's Brian's suggestion for how I animate it:
```
  
  timeout = setTimeout(() => {
    setShowCopied(false);
  }, 700);
}, 2000);
return () => {
  clearTimeout(timeout);
}
```

bvaughn:
Do we have much of a concept of Toast style messages elsewhere in the App?

Seems like something that, ideally, we'd want to design for rather than tackle as one-offs.

jonbell-lot23:
Aw darn, I had two tabs open and posted my toast comments into the wrong one. Deleting.

markerikson:
Updated the behavior based on feedback.  It now looks like this in the console:

### Nested Object: Inline + Truncated

image

### Expanded Object: Truncated

image


and over in the Editor Preview popup, for comparison:

image


### Expanded Object: Expanded String (End)

image


This is going to make the Playwright screenshot tests fail. I know Brian has another PR up that updates those, so I'll hold off on regenerating the screenshots until that PR is merged.

bvaughn:
> This is going to make the Playwright screenshot tests fail. I know Brian has another PR up that updates those, so I'll hold off on regenerating the screenshots until that PR is merged.

Thanks for being considerate! But those PRs are just open for iterating with Jon. Not sure how soon they'll be ready to land. Please go for it.

bvaughn:
Oooh, as for this:
image


I thought we were going for **no wrapping** ?

markerikson:
Final resolution: the last tweak was replacing the Tailwind styling for the button with the equivalent CSS Modules definition.

PR #8079, opened 2022-11-04
Don't use PauseData directly (FE-872)

hbenl:
- moved FrameCache and ScopeCache to the new architecture package so that they're available everywhere
  - I've left `get(Frame|Scope)(Suspense|Async|IfCached)` methods in the old place that call the corresponding new methods - removing them and calling the new methods directly is straightforward but touches a lot of files so I'll do that in a follow-up PR
- rewrote the PauseCache
  - it offered PauseData, but that shouldn't be used directly anymore, now it only offers a `PauseId` for an `ExecutionPoint`
  - removed `trackExecutionPointPauseIds()` - pretty soon the PauseCache should be the only place where `createPause()` is called so this shouldn't be necessary anymore
- added `addValue()` to the generic cache, which is used to cache frames and scopes whenever we receive `PauseData` from the backend
- wrote a second version of `createGenericCache()` that takes an extra argument which is not used for computing the `cacheKey` and adding data to the cache
  - this is not really necessary but I disliked having to pass e.g. a `ReplayClient` to `getFramesIfCached()` even though it's not used
- removed objects, frames and scopes from `Pause`
- fixed the `ErrorBoundary` in the `FramesRenderer` component: when loading async frames failed and the user then switched to a different pause, the `ErrorBoundary` was not reset


packages/bvaughn-architecture-demo/components/console/Input.tsx
```diff
@@ -25,10 +29,18 @@ export default function Input() {
...
+  let pauseId: PauseId | null = useCurrentPauseIdSuspense();
+  let frameId: FrameId | null = null;
+  if (selectedPauseAndFrameId) {
+    pauseId = selectedPauseAndFrameId.pauseId;
+    frameId = selectedPauseAndFrameId.frameId;
+  } else {
+    if (pauseId) {
+      const frames = getFramesSuspense(replayClient, pauseId);
+      frameId = frames?.[0]?.frameId ?? null;
+    }
```
bvaughn, line 42: Seems like we would ideally _not_ suspend here until after we'd confirmed the pause and frame ids didn't match.
```suggestion
  let pauseId: PauseId | null = null;
  let frameId: FrameId | null = null;
  if (selectedPauseAndFrameId) {
    pauseId = selectedPauseAndFrameId.pauseId;
    frameId = selectedPauseAndFrameId.frameId;
  } else {
    pauseId = useCurrentPauseIdSuspense();
    if (pauseId) {
      const frames = getFramesSuspense(replayClient, pauseId);
      frameId = frames?.[0]?.frameId ?? null;
    }
```

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoCompleteList.tsx
```diff
@@ -1,11 +1,13 @@
-import { createPauseResult as Pause, Property, Scope } from "@replayio/protocol";
+import { PauseId, Property } from "@replayio/protocol";
 import { RefObject, useContext, useEffect, useMemo, useRef, useState } from "react";
 import { FixedSizeList as List } from "react-window";
 
 import Popup from "bvaughn-architecture-demo/components/Popup";
-import useCurrentPause from "bvaughn-architecture-demo/src/hooks/useCurrentPause";
+import useCurrentPauseIdSuspense from "bvaughn-architecture-demo/src/hooks/useCurrentPause";
```
bvaughn, line 6: Seems like we should rename `bvaughn-architecture-demo/src/hooks/useCurrentPause` (and the function it exports) rather than just renaming the imports? 

As it is, it would be pretty easy to add new imports _without_ the "suspense" suffix.

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoCompleteList.tsx
```diff
@@ -89,18 +91,17 @@ function AutoCompleteListInner({
   }, [expression]);
 
   let properties: Property[] | null = null;
-  let scopes: Scope[] | null = pause.data.scopes || null;
+  const frameId = getFramesSuspense(replayClient, pauseId)?.[0]?.frameId ?? null;
+  const scopes = frameId ? getScopesSuspense(replayClient, pauseId, frameId).generatedScopes : null;
```
bvaughn, line 95: Should this be pulling frame from the new `SelectedFrameContext` rather than just snagging the top one?

packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
```diff
@@ -1,214 +1,99 @@
...
-      `Point (${executionPoint}) has multiple pause ids (${firstPauseId}, ${newPauseId})`
-    );
-
-    // TODO Hook this up so that it fails our e2e tests if called.
-    captureException(error);
-
-    console.error(error);
+function updateMappedLocation(client: ReplayClientInterface, mappedLocation: MappedLocation) {
+  for (const location of mappedLocation) {
+    location.sourceId = client.getCorrespondingSourceIds(location.sourceId)[0];
```
bvaughn, line 97: I'm a little uneasy about _mutating_ data. That can lead to unpredictable/inconsistent results.

Can this function return a new location? Can we update `sortFramesAndUpdateLocations` return a new array of new location objects (rather than a new array of mutated location objects)?

packages/bvaughn-architecture-demo/src/hooks/useCurrentPause.ts
```diff
@@ -1,14 +1,14 @@
...
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import { isPointInRegions } from "shared/utils/time";
 
 import { TimelineContext } from "../contexts/TimelineContext";
-import { getPauseForExecutionPointSuspense } from "../suspense/PauseCache";
+import { getPauseIdForExecutionPointSuspense } from "../suspense/PauseCache";
 import useLoadedRegions from "./useRegions";
 
-export default function useCurrentPause(): Pause | null {
+export default function useCurrentPauseIdSuspense(): PauseId | null {
```
bvaughn, line 11: Can we rename this file to match the function name? 😄

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
```diff
@@ -9,6 +9,14 @@ export interface GenericCache<TParams extends Array<any>, TValue> {
   getValueSuspense(...args: TParams): TValue;
   getValueAsync(...args: TParams): Promise<TValue>;
   getValueIfCached(...args: TParams): { value: TValue } | undefined;
+  addValue(value: TValue, ...args: TParams): void;
+}
+
+export interface GenericCache2<TExtra, TParams extends Array<any>, TValue> {
```
bvaughn, line 15: `GenericCache2` doesn't seem like a great name. (The suffix "2" doesn't really provide any hint as to what may be different.)

Can we just update `GenericCache` to make the extra param _optional_ somehow (and only have _one_ of these things?)

packages/protocol/thread/pause.ts
```diff
@@ -149,75 +129,20 @@ export class Pause {
...
-            scope: id,
-          });
-          this.addData(data);
-        }
-        const scope = this.scopes.get(id);
-        assert(scope, `scope not found (scope ID: ${id})`);
-        return scope;
-      })
-    );
-  }
```
bvaughn, line 219: Nice! 🥳

packages/protocol/thread/thread.ts
```diff
@@ -502,14 +502,6 @@ class _ThreadFront {
     return pause;
   }
 
-  async getFrames() {
-    if (!this.currentPointHasFrames) {
-      return [];
-    }
-
-    return await this.getCurrentPause().getFrames();
-  }
```
bvaughn, line 511: Nice! 🥳

packages/shared/client/ReplayClient.ts
```diff
@@ -159,6 +162,10 @@ export class ReplayClient implements ReplayClientInterface {
     return response;
   }
 
+  async ensureSourcesLoaded(): Promise<void> {
+    await this._threadFront.ensureAllSources();
+  }
```
bvaughn, line 167: Can we name this something _more explicit_ like `waitForLoadingSources`?

To me, the name "ensure loaded" implies it will:
1.  Check if the thing is already loaded
2. If not, check if it's still loading (and wait)
3. If not, start loading it (and wait)

It doesn't do that 3rd thing though.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -94,7 +94,10 @@ function FramesRenderer({
     <>
       {asyncSeparator}
       <PauseFrames pauseId={pauseId} frames={frames} panel={panel} />
-      <ErrorBoundary fallback={<div className="pane-info empty">Error loading frames</div>}>
+      <ErrorBoundary
+        key={pauseId}
```
bvaughn, line 98: This seems like a good call 👍🏼

src/ui/actions/comments.ts
```diff
@@ -228,9 +234,14 @@ export function seekToComment(item: Comment | Reply): UIThunkAction {
...
   ThreadFront: typeof ThreadFrontType,
+  replayClient: ReplayClientInterface,
   getState: () => UIState
 ) {
-  const frame = (await ThreadFront.currentPause?.getFrames())?.[0];
+  const pauseId = getPauseId(getState());
+  if (!pauseId) {
+    return;
+  }
+  const frame = (await getFramesAsync(replayClient, pauseId))?.[0];
```
bvaughn, line 244: Is this (assuming the first frame) safe/correct?

src/ui/suspense/scopeCache.ts
```diff
@@ -1,37 +1,24 @@
-import { FrameId, MappedLocation, PauseId, Scope, SourceId } from "@replayio/protocol";
+import { FrameId, PauseId, Scope, SourceId } from "@replayio/protocol";
 
-import { createGenericCache } from "bvaughn-architecture-demo/src/suspense/createGenericCache";
-import { Pause, ThreadFront } from "protocol/thread";
-import { assert } from "protocol/utils";
+import {
+  FrameScopes,
+  getScopesAsync as getScopesAsyncNew,
+  getScopesIfCached as getScopesIfCachedNew,
```
bvaughn, line 6: Eee 😬 How long will we have "getScopesIfCached" and "getScopesIfCachedNew"? That feels pretty confusing.

src/ui/suspense/scopeCache.ts
```diff
@@ -1,37 +1,24 @@
...
+  return getScopesAsyncNew(replayClient, pauseId, frameId);
+}
 
-    return { frameLocation: frame.location, generatedScopes, originalScopes };
-  },
-  (pauseId, frameId) => `${pauseId}:${frameId}`
-);
+export function getScopesIfCached(pauseId: PauseId, frameId: FrameId) {
+  return getScopesIfCachedNew(pauseId, frameId);
+}
```
bvaughn, line 21: Why is this necessary?

src/ui/suspense/scopeCache.ts
```diff
@@ -1,37 +1,24 @@
...
-import { createGenericCache } from "bvaughn-architecture-demo/src/suspense/createGenericCache";
-import { Pause, ThreadFront } from "protocol/thread";
-import { assert } from "protocol/utils";
+import {
+  FrameScopes,
+  getScopesAsync as getScopesAsyncNew,
+  getScopesIfCached as getScopesIfCachedNew,
+  getScopesSuspense as getScopesSuspenseNew,
+} from "bvaughn-architecture-demo/src/suspense/ScopeCache";
+import { replayClient } from "shared/client/ReplayClientContext";
```
bvaughn, line 9: Same comment as above about importing this directly.

src/ui/suspense/frameCache.ts
```diff
@@ -1,28 +1,28 @@
...
+  getFramesAsync as getFramesAsyncNew,
+  getFramesIfCached as getFramesIfCachedNew,
+  getFramesSuspense as getFramesSuspenseNew,
+} from "bvaughn-architecture-demo/src/suspense/FrameCache";
 import { createFrame } from "devtools/client/debugger/src/client/create";
 import { PauseAndFrameId } from "devtools/client/debugger/src/reducers/pause";
 import { formatCallStackFrames } from "devtools/client/debugger/src/selectors/getCallStackFrames";
-import { Pause } from "protocol/thread/pause";
-import { assert } from "protocol/utils";
+import { replayClient } from "shared/client/ReplayClientContext";
```
bvaughn, line 12: We shouldn't be importing and using `replayClient` directly. This value _should always_ come from context, or it will break things (e.g. our Playwright test proxy logic)

src/ui/components/SecondaryToolbox/TimelineContextAdapter.tsx
```diff
@@ -38,7 +38,7 @@ export default function TimelineContextAdapter({ children }: PropsWithChildren)
 
       // Pre-cache Pause data (required by legacy app code) before calling seek().
       // The new Console doesn't load this data but the old one requires it.
-      const cachedPauseId = getCachedPauseIdForExecutionPoint(executionPoint);
+      const cachedPauseId = getPauseIdForExecutionPointIfCached(executionPoint)?.value;
```
bvaughn, line 41: Why does this have to call `.value`? That seems wrong. Is this a typo?

src/ui/suspense/frameCache.ts
bvaughn, line 12: Same comment as other places.

packages/bvaughn-architecture-demo/components/console/Input.tsx
hbenl: `useCurrentPauseIdSuspense()` is a hook and would be called conditionally with this suggestion. So I removed it and do the hook part unconditionally and the suspending part conditionally now.

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
hbenl: > Can we just update GenericCache to make the extra param optional somehow (and only have one of these things?)

I'd love to but I don't know how (since you can't have an optional parameter *and* rest parameters).

src/ui/actions/comments.ts
hbenl: I think so. This is used when adding comments to the screenshot or in the logpoint panel - in both cases I think the comment should be associated with the top frame (and if there is no top frame we bail out).

src/ui/suspense/scopeCache.ts
hbenl: Not for long! This is what I meant when I wrote
> I've left `get(Frame|Scope)(Suspense|Async|IfCached)` methods in the old place that call the corresponding new methods - removing them and calling the new methods directly is straightforward but touches a lot of files so I'll do that in a follow-up PR

src/ui/suspense/scopeCache.ts
hbenl: It's not so I removed it.

src/ui/suspense/frameCache.ts
hbenl: This will be removed in the follow-up PR I mentioned above.

hbenl: This will be removed in the follow-up PR I mentioned above.

src/ui/components/SecondaryToolbox/TimelineContextAdapter.tsx
hbenl: This is correct - `getPauseIdForExecutionPointIfCached()` does not return the `pauseId` directly but rather an object containing it. Why? Because the cache may contain `undefined` values (*this* cache doesn't but the frame cache does, for example) and we need to distinguish between a cache hit with an `undefined` value and a cache miss.

hbenl:
I'll have a look at the failing tests next.

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
bvaughn: You could change the parameter for a given cache to be an object that has some optional attributes and some required attributes.

hbenl: But then we'd be losing some typing information: for example `getFramesSuspense()` needs a `ReplayClient` but `getFramesIfCached()` doesn't, but we'd have an optional `ReplayClient` attribute for both of these functions instead.

src/ui/components/SecondaryToolbox/TimelineContextAdapter.tsx
bvaughn: > the cache may contain `undefined` values

Interesting. Feels funny to me that `getValueIfCached` has a different return type/shape than `getValueSuspense` and `getValueAsync` although I see where you're coming from.

I'm curious– are there any caches where this distinction _does_ matter? or is it more of a conceptual one?

Either way, we can chat more about this later. Doesn't seem like anything to get hung up on for this PR 😄

packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
bvaughn: Interesting.

I don't love having two APIs that are basically the same but for this detail but we can always iterate more on it in future PRs 😁

src/ui/suspense/frameCache.ts
hbenl: I mostly split this into 2 PRs to make reviewing easier. I'll land them together.

src/ui/components/SecondaryToolbox/TimelineContextAdapter.tsx
hbenl: This is more of a conceptual thing. The FrameCache allows `undefined` values but the callers of `getFramesIfCached()` currently don't distinguish between a cache hit with an `undefined` value and a cache miss.

bvaughn: Right, I understand that. A comparable API might be something like `Map` which has both a `has` and `get` method to differentiate these cases.

src/ui/suspense/frameCache.ts
bvaughn: Cool!

PR #8080, opened 2022-11-04
New Source viewer highlights current line for corresponding sources

bvaughn:
I wish this wasn't so pervasive/complicated and easy to get wrong.

### DIFF
--- a/packages/bvaughn-architecture-demo/components/sources/CurrentLineHighlight.tsx
+++ b/packages/bvaughn-architecture-demo/components/sources/CurrentLineHighlight.tsx
@@ -1,5 +1,5 @@
 import { SourceId } from "@replayio/protocol";
-import { Suspense, useContext } from "react";
+import { Suspense, memo, useContext } from "react";
 
 import { getPauseDataSuspense } from "bvaughn-architecture-demo/src/suspense/PauseCache";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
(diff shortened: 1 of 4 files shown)


bvaughn:
Going to land this now bc it _does_ fix a bug but I will be happy to make changes in a follow up 🙇🏼‍♂️

packages/bvaughn-architecture-demo/components/sources/CurrentLineHighlight.tsx
```diff
@@ -31,12 +32,22 @@ function CurrentLineHighlightSuspends({ lineNumber, sourceId }: Props) {
...
-            location => location.line === lineNumber && location.sourceId === sourceId
-          )
+          selectedFrame.location.find(location => {
+            if (correspondingSourceIds.includes(location.sourceId)) {
+              const correspondingLocations = client.getCorrespondingLocations(location);
+              return (
+                correspondingLocations.findIndex(
+                  correspondingLocation =>
+                    correspondingLocation.line === lineNumber &&
+                    correspondingLocation.sourceId === sourceId
```
jcmorrow, line 46: My head hurts

bvaughn: Mine too.

jcmorrow (commented):
Heh, I think all I can do in this case is agree that it is complicated 😅 . I was just bit by a corresponding source issue in combination with a cache thing this morning too. I do think it would be nice if we could wrap corresponding sources earlier in the `linker -> backend -> frontend` chain. I still don't feel like I know exactly what it would look like, but I'm sympathetic to not wanting code like this this far away from what feels like the source of the complexity.

packages/bvaughn-architecture-demo/components/sources/CurrentLineHighlight.tsx
bvaughn: The logic here is basically saying:

This is the file that's open.

If the currently selected frame matches this line– for _any_ source that "corresponds to" this one, show the highlight.

PR #8083, opened 2022-11-05
Fix mismatch Pause id in PreviewPopup component

bvaughn:
This was causing us to fetch hover preview information using one Pause id (the currently selected frame) but then retrieve object previews for hovered values using _a different_ Pause id 😞 I wish this wasn't so complicated.

(See Replay [8cdacfc1-06b1-4252-a684-cb795086681f](https://app.replay.io/recording/8cdacfc1-06b1-4252-a684-cb795086681f)) for a bug repro.)


bvaughn:
This is a pretty painful bug to have in production (even behind an opt-in flag) so I'm going to land this fix now and I'll be happy to make any post-PR changes.

PR #8085, opened 2022-11-06
Improve new log point panel type-ahead

bvaughn:
- [x] Use Lezer output for auto-complete expression parsing
- [x] Support template string variables and mid-token cursor positions
- [x] Support auto-completing variables in the middle of a string
  - [x] Support restoring the cursor position after auto-complete
- [x] Add unit tests

This PR is a big step toward JS terminal. The primary thing that's missing after it lands is syntax highlighting while in edit mode.

Overview / demo:
https://www.loom.com/share/c9063f20cd984183a8f3cef0054c35ee

### DIFF
--- a/packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
+++ b/packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
@@ -1,6 +1,7 @@
 import { ChangeEvent, KeyboardEvent, Suspense, useRef, useState } from "react";
 
 import getExpressionFromString from "../utils/getExpressionFromString";
+import updateStringWithExpression from "../utils/updateStringWithExpression";
 import AutoCompleteList from "./AutoCompleteList";
 import styles from "./AutoComplete.module.css";
 
@@ -33,14 +34,9 @@ export default function AutoComplete({
       onChangeProp(newValue);
     }
 
-    const cursorIndex = input.selectionStart;
-    const shouldAutoComplete =
-      cursorIndex === newValue.length &&
-      newValue.length > 0 &&
-      newValue.charAt(newValue.length - 1) !== " ";
-    const expression = shouldAutoComplete
-      ? getExpressionFromString(newValue, cursorIndex - 1)
-      : null;
+    const cursorIndex = input.selectionStart || value.length;
+    const shouldAutoComplete = newValue.length > 0 && newValue.charAt(newValue.length - 1) !== " ";
+    const expression = shouldAutoComplete ? getExpressionFromString(newValue, cursorIndex) : null;
 
     setExpression(expression);
   };
@@ -70,16 +66,8 @@ export default function AutoComplete({
     const input = inputRef.current;
     if (input) {
       const value = input.value;
-      let cursorIndex = input.selectionStart || value.length;
-      while (cursorIndex >= 0) {
-        const character = value.charAt(cursorIndex - 1);
-        if (character === "." || character === " ") {
-          break;
-        }
-        cursorIndex--;
-      }
-
-      const newValue = value.substr(0, cursorIndex) + match;
+      const cursorIndex = input.selectionStart || value.length;
+      const newValue = updateStringWithExpression(value, cursorIndex, match);
 
       onChangeProp(newValue);
 
--- a/packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
+++ b/packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
@@ -26,19 +26,38 @@ export default function AutoComplete({
 }) {
   const inputRef = useRef<HTMLInputElement>(null);
   const [expression, setExpression] = useState<string | null>(null);
+  const [cursorIndex, setCursorIndex] = useState<number | null>(null);
+
+  useLayoutEffect(() => {
+    if (cursorIndex !== null) {
+      const input = inputRef.current;
+      if (input) {
+        input.setSelectionRange(cursorIndex, cursorIndex);
+        setCursorIndex(null);
+      }
+    }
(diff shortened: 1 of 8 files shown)


packages/bvaughn-architecture-demo/src/suspense/createGenericCache.ts
```diff
@@ -18,7 +18,7 @@ interface HookState<TValue> {
 }
 
 export function createGenericCache<TParams extends Array<any>, TValue>(
-  fetchValue: (...args: TParams) => Promise<TValue>,
+  fetchValue: (...args: TParams) => Promise<TValue> | TValue,
```
bvaughn, line 21: @hbenl I don't think there's any reason that a Suspense cache's underlying function _has_ to be async, right?

bvaughn: Maybe we could do something special in this case and just avoid throwing on cache misses entirely?

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
```diff
@@ -33,14 +34,9 @@ export default function AutoComplete({
...
-    const shouldAutoComplete =
-      cursorIndex === newValue.length &&
-      newValue.length > 0 &&
-      newValue.charAt(newValue.length - 1) !== " ";
-    const expression = shouldAutoComplete
-      ? getExpressionFromString(newValue, cursorIndex - 1)
-      : null;
+    const cursorIndex = input.selectionStart || value.length;
+    const shouldAutoComplete = newValue.length > 0 && newValue.charAt(newValue.length - 1) !== " ";
+    const expression = shouldAutoComplete ? getExpressionFromString(newValue, cursorIndex) : null;
```
markerikson, line 39: :question: this means we oughta be able to replace use of Babel for parsing in the JSTerm as well, right?

packages/bvaughn-architecture-demo/components/sources/utils/updateStringWithExpression.test.ts
```diff
@@ -0,0 +1,61 @@
+import updateStringWithExpression from "./updateStringWithExpression";
+
+// Converts a string like: "foo |bar"
+// Into the string "foo bar" and the cursor index 4
+function getExpressionHelper(text: string, expression: string): string | null {
+  let cursorIndex = text.length;
+
+  for (let index = 0; index < text.length; index++) {
+    const character = text.charAt(index);
+    if (character === "|") {
```
markerikson, line 10: :question: where's the `'|'` coming from in the first place? Is that being auto-inserted at the cursor position by Lezer? Are these utils inserting it?  What happens if there's a literal pipe character being written?

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
bvaughn: Yes! 😄 Once that project is officially under way.

packages/bvaughn-architecture-demo/components/sources/utils/updateStringWithExpression.test.ts
bvaughn: There is no "|" in the real code. This is just a test helper I wrote so that I can write human-readable tests (with a visible "cursor").

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
bvaughn: TBH this feels so close now...all that's missing is syntax highlighting in edit mode.

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
```diff
@@ -25,24 +26,38 @@ export default function AutoComplete({
...
-      newValue.charAt(newValue.length - 1) !== " ";
-    const expression = shouldAutoComplete
-      ? getExpressionFromString(newValue, cursorIndex - 1)
-      : null;
+  const updateExpression = () => {
+    const input = inputRef.current;
+    if (input) {
+      const value = input.value;
+      const cursorIndex = input.selectionStart || value.length;
+      // const shouldAutoComplete = value.length > 0 && value.charAt(value.length - 1) !== " ";
```
markerikson, line 56: :hand: dead code

packages/bvaughn-architecture-demo/components/sources/utils/getExpressionFromString.ts
```diff
@@ -1,37 +1,70 @@
...
-    if (character === " ") {
-      break;
+  const parsed = highlighter(expression, ".js");
+  if (parsed == null || parsed.length === 0) {
+    return null;
+  } else {
+    const line = parsed[parsed.length - 1];
+
+    const element = document.createElement("div");
+    element.innerHTML = line;
```
markerikson, line 14: :speech_balloon: huh. so we're letting Lezer parse things into HTML, and then using an unattached DOM node to figure out what we care about?

That's... hacky, but hey, if it works! :)

markerikson (approved):
One leftover dead code line but seems reasonable otherwise

packages/bvaughn-architecture-demo/components/sources/AutoComplete/AutoComplete.tsx
bvaughn: Dead _comment 😆

packages/bvaughn-architecture-demo/components/sources/utils/getExpressionFromString.ts
bvaughn: I agree that it is a bit funky 😁 but we're doing it in a few places and it seems to work well

PR #8091, opened 2022-11-07
Handle "linker version" errors for hit counts

markerikson:
This PR:

- Fixes UI crashes when the recording is so old it doesn't support requesting hit counts:
  - Wraps `<SourceOutline>`in an `<ErrorBoundary>` just in case
  - Updates `fetchSourceHitCounts` to treat "linker version" errors as a "nope, no hit counts available" result

As a specific example, currently loading up https://app.replay.io/recording/503-clicks--7b36ed70-b486-459d-97af-367ae6e49c03 and opening any source file will crash the entire UI.  (I actually find this surprising given that we have the entire app wrapped in an `<ErrorBoundary>`)

### DIFF
--- a/packages/shared/utils/error.ts
+++ b/packages/shared/utils/error.ts
@@ -7,6 +7,7 @@ export enum ProtocolError {
   RecordingUnloaded = 38,
   TooManyLocationsToPerformAnalysis = 67,
   TooManyPoints = 55,
+  LinkerDoesNotSupportAction = 48,
 }
 
 export const commandError = (message: string, code: number): CommandError => {
@@ -26,6 +27,11 @@ export const isCommandError = (error: unknown, code: number): boolean => {
       // TODO [BAC-2330] The Analysis endpoint returns an error string instead of an error object.
       return error === "There are too many points to complete this operation";
     }
+  } else if (code === ProtocolError.LinkerDoesNotSupportAction) {
+    if (typeof error === "string") {
+      console.error("Unexpected error type encountered (string):", error);
+    }
+    return error === "The linker version used to make this recording does not support this action";
   }
 
   return false;
--- a/src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
+++ b/src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
@@ -12,6 +12,7 @@ import React, {
 import AutoSizer from "react-virtualized-auto-sizer";
 import { FixedSizeList as List } from "react-window";
 
+import ErrorBoundary from "bvaughn-architecture-demo/components/ErrorBoundary";
 import { FocusContext } from "bvaughn-architecture-demo/src/contexts/FocusContext";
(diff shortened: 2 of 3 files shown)


src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
```diff
@@ -187,12 +188,14 @@ export default function SourceOutlineWrapper() {
...
 
   return (
-    <Suspense fallback={null}>
-      <SourceOutline
-        cursorPosition={cursorPosition || null}
-        selectedSource={selectedSource || null}
-        symbols={symbols}
-      />
-    </Suspense>
+    <ErrorBoundary>
```
bvaughn, line 191: Just curious, but is this new boundary a redundant fix (given the change to check error code 48 in the cache)?

Probably still a good idea FWIW 👍🏼  Just curious

packages/shared/utils/error.ts
```diff
@@ -26,6 +27,11 @@ export const isCommandError = (error: unknown, code: number): boolean => {
       // TODO [BAC-2330] The Analysis endpoint returns an error string instead of an error object.
       return error === "There are too many points to complete this operation";
     }
+  } else if (code === ProtocolError.LinkerDoesNotSupportAction) {
+    if (typeof error === "string") {
+      console.error("Unexpected error type encountered (string):", error);
+    }
+    return error === "The linker version used to make this recording does not support this action";
```
bvaughn, line 34: Does this happen for this error code too? I thought it was specific to the way the Analysis stuff bundled along errors.

bvaughn (approved):
Thanks for digging into this 👍🏼 

Not sure if we want the type-of-string check, but stamping to unblock landing!

packages/shared/utils/error.ts
markerikson: It was partly copy-pasta, but yeah, by the time we get to here the error really is a raw string with that exact text

src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
markerikson: It's probably redundant, and tbh I'm still confused why the boundary we have around the root of the app isn't catching this 🤷‍♂️ .

But this is also probably one of those cases where it doesn't hurt to have the boundary there anyway.

packages/shared/utils/error.ts
bvaughn: Really! Even for this error code? 😮

bvaughn: I think this had the unintended side effect of causing the log point panel to show "0" instead of the too-many-points error.

bvaughn: Added logic to account for this in 0c9ac73ca98a3b9cb008ad5bf15d83ef054e6195

Seems like we should fix it higher up though

PR #8092, opened 2022-11-07
Add clipboard-based breakpoint import/export

markerikson:
This PR:

- Implements simple import/export of print statements / breakpoints via the clipboard, with a couple new entries for the Command Palette

image

### DIFF
--- a/src/ui/actions/app.ts
+++ b/src/ui/actions/app.ts
@@ -1,6 +1,7 @@
 import { KeyboardEvent, NodeBounds, unprocessedRegions } from "@replayio/protocol";
 import groupBy from "lodash/groupBy";
 
+import { isValidPoint } from "bvaughn-architecture-demo/src/contexts/PointsContext";
 import { openQuickOpen } from "devtools/client/debugger/src/actions/quick-open";
 import { shallowEqual } from "devtools/client/debugger/src/utils/compare";
 import { prefs } from "devtools/client/debugger/src/utils/prefs";
@@ -202,6 +203,56 @@ export function loadMouseTargets(): UIThunkAction {
   };
 }
 
+export function setBreakpointsFromClipboard(): UIThunkAction {
+  return async (dispatch, getState, { ThreadFront }) => {
+    const currentFocusedElement = document.activeElement;
+    let text = "";
+    try {
+      // The `readText()` API requires that the current DOM be focused somehow
+      document.body.focus();
+      text = await navigator.clipboard.readText();
+
+      // Was the clipboard data even valid JSON?
+      const parsedValue = JSON.parse(text);
+      // Assuming it was JSON, does it appear to be an array of Point objects?
+      if (Array.isArray(parsedValue) && parsedValue.every(isValidPoint)) {
+        // TODO [FE-874] This will need to change if we modify how we do per-recording persistence
+
+        // This value appears to be actual breakpoints. Our current implementation watches
+        //  for changes in `localStorage` and will auto-update itself when that event fires.
+        const storageKey = `${ThreadFront.recordingId!}::points`;
+        // Just reuse the text from the clipboard
+        localStorage.setItem(storageKey, text);
+        // Apparently the "storage" event only fires across tabs, and setting a value
+        // in the _same_ tab won't trigger it. Do that manually:
+        window.dispatchEvent(new Event("storage"));
+      }
+    } catch (err) {
+      if (err instanceof SyntaxError) {
+        console.error("Could not parse clipboard text as JSON:", `"${text}"`);
+      } else if (err instanceof DOMException) {
+        console.error("Could not copy clipboard text - document not focused", err);
+      } else {
+        console.error(err);
+      }
+    } finally {
+      if (currentFocusedElement && "focus" in currentFocusedElement) {
+        (currentFocusedElement as HTMLElement).focus();
+      }
+    }
+  };
+}
+
+export function copyBreakpointsToClipboard(): UIThunkAction {
+  return async (dispatch, getState, { ThreadFront }) => {
+    const storageKey = `${ThreadFront.recordingId!}::points`;
+    const currentPointsValue = localStorage.getItem(storageKey);
+    if (currentPointsValue) {
+      await navigator.clipboard.writeText(currentPointsValue);
+    }
+  };
+}
+
 export function executeCommand(key: CommandKey): UIThunkAction {
   return (dispatch, getState) => {
     const recordingId = getRecordingId();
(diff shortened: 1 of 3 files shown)


src/ui/actions/app.ts
```diff
@@ -202,6 +203,56 @@ export function loadMouseTargets(): UIThunkAction {
...
+      const parsedValue = JSON.parse(text);
+      // Assuming it was JSON, does it appear to be an array of Point objects?
+      if (Array.isArray(parsedValue) && parsedValue.every(isValidPoint)) {
+        // TODO [FE-874] This will need to change if we modify how we do per-recording persistence
+
+        // This value appears to be actual breakpoints. Our current implementation watches
+        //  for changes in `localStorage` and will auto-update itself when that event fires.
+        const storageKey = `${ThreadFront.recordingId!}::points`;
+        // Just reuse the text from the clipboard
+        localStorage.setItem(storageKey, text);
```
bvaughn, line 225: Would it be much harder to _merge_ these? Seems like we could do it by comparing locations?

Might be annoying if someone sent you their log points and it clobbered all of your existing ones.

bvaughn (approved):
Looks nice in general. Left an idea but take it or leave it.

src/ui/actions/app.ts
```diff
@@ -202,6 +203,56 @@ export function loadMouseTargets(): UIThunkAction {
...
+
+      // Was the clipboard data even valid JSON?
+      const parsedValue = JSON.parse(text);
+      // Assuming it was JSON, does it appear to be an array of Point objects?
+      if (Array.isArray(parsedValue) && parsedValue.every(isValidPoint)) {
+        // TODO [FE-874] This will need to change if we modify how we do per-recording persistence
+
+        // This value appears to be actual breakpoints. Our current implementation watches
+        //  for changes in `localStorage` and will auto-update itself when that event fires.
+        const storageKey = `${ThreadFront.recordingId!}::points`;
```
bvaughn, line 223: nit: Would be nice to have a single function for this:
```js
function getLocaleStorageKeyForPoints(recordingId: RecordingId): string {
  return `${recordingId}::points::transition`;
}
```

That way if we ever change one, we'll _know_ about the other one.

bvaughn (commented):
Actually Mark, why are these methods Redux actions? I don't think they're reading/writing Redux state.

Might be nice if they lived closer to (in) the PointsContext. Then the connection between them (format and `localStorage` key etc) would be clearer.

markerikson:
There's a few reasons why:

- They're being triggered from the "Command Palette" modal, which is main app
- All the other handlers for the Command Palette are also thunks
- They do need access to the recording ID.  In the main app, that's in `ThreadFront`, and that's generally accessed via dependency injection in thunks rather than direct imports.

I know that the code inside the actual `PointsContextRoot` component has access to `recordingId` via `SessionContext`, but that would be _inside_ the component, and I can't immediately think of a way to trigger logic inside there from the Command Palette code.

src/ui/actions/app.ts
markerikson: It's possible. My mental use case atm is the scenario where you and I were looking at the `<BreakpointNavigation>` bug last week - I'd spent a day building up a set of print statements, you were about to look for the first time, and I could have potentially sent over my list as a starting point for you to paste in and then tweak further.

If we really want me to tweak this further I can, but my goal here was just get the MVP in with minimal effort and time spent.

bvaughn:
> They're being triggered from the "Command Palette" modal, which is main app

I don't understand why this would make them more appropriate as Redux actions vs something else (regular functions exported from a module).

It feels like making them _actions_ when they neither read nor write from the Redux store is a little confusing.

> They do need access to the recording ID. In the main app, that's in ThreadFront, and that's generally accessed via dependency injection in thunks rather than direct imports.

That's fair I guess. Quite a few things import `ThreadFront` directly but I guess that's not great and we probably don't want to encourage more of that.

jonbell-lot23:
I put this comment into the wrong issue. Here it is:

Two more details:

1. We need messaging for when the item is copied or pasted.
2. It'd be great to make that its own component

If you look into [this stalled PR](https://github.com/replayio/devtools/pull/7943/files), you can see the code I used to do this in our cURL functionality:

Here's the element itself:

```
<div id="showCopied" className="mb-1.5 flex shrink rounded-lg bg-black p-1.5 text-center text-white opacity-100 shadow-2xl transition-all duration-700 ease-in-out">Copied to clipboard</div></div>
```

And here's Brian's suggestion for how I animate it:

```
  
  timeout = setTimeout(() => {
    setShowCopied(false);
  }, 700);
}, 2000);
return () => {
  clearTimeout(timeout);
}
```

Brian asked in the other ticket if we can design this out rather than one-offing it, and yup, that's what I'm requesting here. We need this in a component that we can use in a few places like here, the "copy URL" functionality in share, the "copy as cURL" that's currently in a PR, and a few others.

bvaughn:
> Brian asked in the other ticket if we can design this out rather than one-offing it, and yup, that's what I'm requesting here. We need this in a component that we can use in a few places like here, the "copy URL" functionality in share, the "copy as cURL" that's currently in a PR, and a few others.

I think we should write a ticket and design the toast messaging component(s) separately from this PR.

jonbell-lot23:
I'll trust you two to figure out the best way to chunk it out, but a feature like this does need UX feedback.

As for making it a standard component, I'd vote for taking my simple code, making a component, and using that component. Then the separate PR can be for starting to use that in multiple places in the system, which is of course lower priority than many other items we're working through.

bvaughn:
> I'll trust you two to figure out the best way to chunk it out, but a feature like this does need UX feedback.

Agreed. Would be happy to create a DES ticket for you to triage if you'd like?

> As for making it a standard component, I'd vote for taking my simple code, making a component, and using that component. Then the separate PR can be for starting to use that in multiple places in the system, which is of course lower priority than many other items we're working through.

Just to be clear, the "design" I was referring to above was mostly things like:
* How is the state structured?
* How does the Toast behave if there are multiple messages shown at once?
* What happens if the user clicks the toast or clicks outside of the toast?

So we don't just copy-paste a component and introduce a new set of quirky edge-cases that might break.

markerikson:
I'm getting directly conflicting feedback on this PR atm :(  That's blocking me from trying to finish this up.

jasonLaster:
In the spirit of a 30 minute exploration, I'm pretty sure we dont need to add a toast :) They'll be plenty of other use cases down the line for toasting 🥂

bvaughn:
Yes. Let's skip the toast.

markerikson:
After further discussion, we're going to skip the toast for now on the grounds that there's questions about how to properly implement it, and this is a technically-public-but-really-only-meant-for-internal-usage MVP feature.

PR #8097, opened 2022-11-08
Add NUX nags to new Source viewer

bvaughn:
https://www.loom.com/share/35fd9c792bbc4d9286e6d983ae41df1e


jasonLaster:
The nags up to the border look great. The border by itself looks a bit too much like an error state. I wonder if there is a natural way to pulse the color so it is more of a call to action. I get that this is an easy way to burn a lot of time, so hoping @jonbell-lot23 or others have a good idea here

bvaughn:
> The border by itself looks a bit too much like an error state.

Oooh, that's interesting. For comparison:

### Nag
Screen Shot 2022-11-08 at 11 27 30 AM

### Error
Screen Shot 2022-11-08 at 11 27 45 AM

The nag might even look _more_ like an error than the error does.

jonbell-lot23:
Hey gang! Here are a few thoughts:

1. Let's keep the banner at the top and use it to say "Now edit your print statement" or similar. That helps with messaging.
2. We could remove the border, auto-select the text, and make that selection fuchsia coloured. (I like auto-selecting the text anyway, adding a fuchsia color is just a nice touch for NUX consistency/messaging)

Let me know if that's confusing, I can make a visual after breakfast.

bvaughn:
@jonbell-lot23 Yes. Those are both doable.

https://www.loom.com/share/ee878af6232c4b259edc7397bdd96851

jonbell-lot23:
> @jonbell-lot23 Yes. Those are both doable.
> 
> https://www.loom.com/share/ee878af6232c4b259edc7397bdd96851

Love it!

The auto-selected fuchsia is quite intense ... but that's good! We lose people during this step, so the more clear we can be, even if it's garish, the better it is for people's onboarding success.

bvaughn:
Agreed. It's intense. Didn't expect to like it, but once I saw it– I think it's good.

jonbell-lot23 (approved):

PR #8100, opened 2022-11-08
Handle cypress anotations from chromium

jaril:


markerikson:
@jaril fyi you'll want to re-run `yarn` to make sure you've got the new Prettier-based import sorting rules installed correctly, then either reformat files via VS Code or running Prettier from the CLI (such as `yarn prettier --write "src/**/*.{js,jsx,ts,tsx}"` )

src/devtools/client/debugger/src/components/TestInfo/TestCase.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export function TestCase({ test, location }: { test: TestItem; location?: SourceLocation }) {
+  const [expandSteps, setExpandSteps] = useState(false);
+  const dispatch = useAppDispatch();
+  const cx = useAppSelector(getThreadContext);
+  const expandable = test.steps || test.error;
+
+  const onClick = () => {
+    if (location) {
+      dispatch(selectLocation(cx, location as any));
```
markerikson, line 21: :question: what's the TS mismatch here?

src/devtools/client/debugger/src/components/TestInfo/TestInfo.tsx
```diff
@@ -1,128 +1,27 @@
...
-      {expandSteps ? <TestSteps test={test} startTime={test.relativeStartTime} /> : null}
-    </div>
-  );
-}
-
-function TestSteps({ test, startTime }: { test: TestItem; startTime: number }) {
-  const { steps } = test;
+  // The test start times in metadata may be incorrect. If we have the reporter annotations,
+  // we can use those instead
+  const correctedTestCases = testCases.map((t, i) => ({
```
markerikson, line 14: :hand: seems like a good candidate for a `useMemo`.

src/ui/reducers/reporter.ts
```diff
@@ -0,0 +1,26 @@
+import { compareNumericStrings } from "protocol/utils";
+import { ReporterAction } from "ui/actions/reporter";
+import { UIState } from "ui/state";
+import { ReporterState } from "ui/state/reporter";
+
+export default function update(
```
markerikson, line 6: :stop: Please use `createSlice` from Redux Toolkit instead :)

src/ui/setup/dynamic/devtools.ts
```diff
@@ -225,7 +228,8 @@ export default async function setupDevtools(store: AppStore, replayClient: Repla
   setupGraphics();
   initOutputSyntaxHighlighting();
   setupNetwork(store, ThreadFront);
-  setupReactDevTools(store, ThreadFront);
+  // setupReactDevTools(store, ThreadFront);
```
markerikson, line 231: :hand: Need to restore the React behavior

src/ui/state/reporter.ts
```diff
@@ -0,0 +1,11 @@
+import { ExecutionPoint } from "@replayio/protocol";
+
+export interface Annotation {
```
markerikson, line 3: :hand: The early codebase defined TS types separately, but it's better to define the slice state types _inside_ the reducer/slice file.  Move this over to be alongside the corresponding reducer.

(Also, ideally our `UIState` type would _completely_ infer the entire type of the state from our root reducer, but we're still manually adding the slice state types to `UIState` atm.)

src/devtools/client/debugger/src/components/TestInfo/TestCase.tsx
jaril: Wrong type!

src/ui/actions/reporter.ts
```diff
@@ -0,0 +1,46 @@
...
+  await ThreadFront.ensureAllSources();
+
+  // Annotations come in piecemeal over time. Cut down the number of dispatches by
+  // storing incoming Annotations and debouncing the dispatch considerably.
+  let receivedAnnotations: Annotation[] = [];
+
+  // Debounce loading of annotations
+  const onAnnotationsReceived = debounce(() => {
+    const filtered = receivedAnnotations.filter(a => a.kind === kind);
+    store.dispatch(
```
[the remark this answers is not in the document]
jaril: I might be missing something. The way I wrote it, I'm using a reporter-specific action similar to `addAnnotations`. That shouldn't trip reactDevtools's `addAnnotations`, right?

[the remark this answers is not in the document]
markerikson: Ah, good point, missed that you were defining it in this same file.

Um... FYI **_please_ don't write Redux code this way any more :)**

With Redux Toolkit and `createSlice`, you don't write action creators or action type strings yourself - `createSlice` generates the action creators for you automatically.

Also, defining the TS union types of possible actions doesn't provide a benefit.

See files like `ui/actions/app.ts` for an example of how to use `createSlice`.

FWIW I'll note that if you remove those bits of code from this file, the only thing that's left is `setupReporter()`.  That _could_ potentially either move to another file, or even inline itself into some other existing `setupSomeStuff()` function, although that's less critical overall.

markerikson (approved):

PR #8101, opened 2022-11-08
Use standard "something failed" display in <ErrorBoundary>

markerikson:
This PR:

- Updates the existing main app `<ErrorBoundary>` component to use our standard `<Error>` component to say "something went wrong, please refresh", instead of leaving the entire page as a blank gray screen

Turns out it was already trying to do half of that by dispatching `setUnexpectedError`, but it wasn't actually _dispatching_ that action.

Updated to render the `<Error>` component if appropriate, and actually dispatch the action:

image

### DIFF
--- a/src/ui/components/ErrorBoundary.tsx
+++ b/src/ui/components/ErrorBoundary.tsx
@@ -7,6 +7,7 @@ import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
 import { UnexpectedError } from "ui/state/app";
 import { isDevelopment } from "ui/utils/environment";
 
+import Error from "./shared/Error";
 import { BlankViewportWrapper } from "./shared/Viewport";
 
 export const ReplayUpdatedError: UnexpectedError = {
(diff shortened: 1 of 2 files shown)


jonbell-lot23 (approved):
My name is Jon Bell and I approve this PR

src/ui/components/ErrorBoundary.tsx
```diff
@@ -24,20 +25,22 @@ export default function ErrorBoundary({ children }: { children: ReactNode }) {
       return dispatch(setUnexpectedError(ReplayUpdatedError, true));
     }
 
-    if (error.name === "CommandError" || isDevelopment()) {
+    if (error.name === "CommandError") {
       return;
     }
 
-    setUnexpectedError({
```
bvaughn, line 31: Oh lol

src/ui/components/ErrorBoundary.tsx
```diff
@@ -24,20 +25,22 @@ export default function ErrorBoundary({ children }: { children: ReactNode }) {
...
+        content: "Something went wrong, but you can refresh the page to try again.",
+        action: "refresh",
+      })
+    );
   };
 
   return (
     <Sentry.ErrorBoundary onError={onError}>
-      {unexpectedError ? <BlankViewportWrapper /> : children}
+      {unexpectedError ? <Error /> : children}
```
bvaughn, line 43: Would this look weird if it rendered inside of a nested component?

bvaughn (approved):

src/ui/components/ErrorBoundary.tsx
markerikson: Hmm. So it looks like this component is only used in two places:

- `_app.tsx`
- `Editor/Footer.tsx`

Everything else uses the `bvaughn/components/ErrorBoundary` component

Let me swap the old editor footer to use that component too, so this only gets used around the main app

bvaughn: Sounds good!

PR #8102, opened 2022-11-08
Points panel UI tweaks

bvaughn:
1. Change capsule from fixed with to min-width
2. Increased timeline background for dark theme
3. Add more space between badge picker and log point text
4. Speed up badge expand/collapse animation by 50%


jonbell-lot23:
Looks good, thanks!

PR #8103, opened 2022-11-08
Log point timeline interaction tweak

bvaughn:
Don't disable next/prev buttons while log point panel is in the middle of an optimistic update.

https://www.loom.com/share/326a2938e2de4c798cd2970566ad9b26

FYI @jonbell-lot23


jonbell-lot23:
I'm in a call so I can't watch the Loom with audio, but I watched without audio and it's feeling good.

If you're ok with it, I'm ok with it. If there are outstanding questions/concerns, I'll address in a few hours.

PR #8109, opened 2022-11-09
Start adding test styles

jasonLaster:
Starts a couple of things

1. Adds icons for the common types of steps
2. Adds step indices

Screen Shot 2022-11-08 at 9 14 37 PM


jonbell-lot23:
Chiming in to say that we have several PRs floating around, this one is a late addition, and it's late in the US and NZ.

Can't commit to doing anything with this on short notice, but I agree with the line numbers and we'll massage the icons in the coming days.

jonbell-lot23:
Looks like Jaril's build also has the line numbers, so that's something to consider when we merge

jaril:
i'll take this over and make sure it's merged after the core bits are in!

PR #8116, opened 2022-11-09
Test Suite steps polish pass

jaril:
Fixes SCS-255


jaril:
@jonbell-lot23 this deviates from the original figma:
- For 3, hovering on the step buttons (view source, focus on test case) changes the background color instead of the icon color. The existing icon color makes it difficult to change it on hover.
- For 4, it's quite involved to get a border for the one (or multiple) paused steps. Can I sell you on doing a bg color instead? In the meantime, I bolded the text for the step that's paused so that we're not losing that information. Though it's very subtle.
- Re: "Use the error color on error areas and green success color on success areas", no idea what that means. Sync later to talk about it?

Merging this now but we can do follow ups!

PR #8123, opened 2022-11-10
Use server side props instead of static props for recording metadata

ryanjduffy:
## Issue

If the first render of a replay page (`/recording/[id]`) includes a Bearer `Authorization` header, we were inadvertently caching that identity in the prerendered response due to the convergence of our auth design for testing and SSR support added for social cards. This causes subsequent requests for the same recording to authenticate the user as the cached identity rather than prompting them to log in.

## Resolution

Switch to using server-side props instead of static props.

This has the side effect of requiring a graphql query for every recording page request whereas previously it was only required for the first request. I don't expect this overhead to be prohibitive because in many cases, we were performing this query anyway because the first user for the replay was the only user.

If we notice this impacting FCP or TTI materially, we can pursue other options like edge caching the recording metadata or alternative auth approaches.


ryanjduffy:
Tagged the whole FE team because I wasn't sure who might want to put eyes on this

bvaughn (commented):
Just wanted to acknowledge the PR. I don't think I have enough context to know any ways this might be bite us, so 👍🏼

PR #8124, opened 2022-11-10
New JS Term component

bvaughn:
- [x] Show syntax highlighted text during edit (built with `contenteditable`)
- [x] Auto-complete suggestions should horizontally align to with the start of the current expression
- [x] Fix overflow behavior for long log statements
  - [x] Scroll to overflow
  - [x] Typing when overflowed should not unintentionally trigger save/cancel buttons
- [x] Auto-complete suggestions use the selected pause and frame ID rather than the top frame (so they work better than the old suggestions did in certain cases).
- [x] Add Playwright tests?

Overview walk through:
https://www.loom.com/share/187faf078dbb444aa4605ff25b02a2b6

Kapture 2022-11-10 at 08 46 29


packages/bvaughn-architecture-demo/components/sources/utils/getExpressionRange.ts
```diff
@@ -0,0 +1,24 @@
+export default function getExpressionRange(
+  string: string,
+  cursorIndex: number
+): [startIndex: number, endIndex: number] {
+  let startIndex = cursorIndex;
+  while (startIndex >= 0) {
+    const prevCharacter = string.charAt(startIndex - 1);
+    if (prevCharacter === "." || prevCharacter === " " || prevCharacter === "{") {
```
narutosstudent, line 8: I know I have zero context here :sweat_smile:

If it is a space, dot, or backward curly bracket, we want to break out of the loop? I wonder if it makes sense to extract i.e. `"."` into a constant? Or have an object with the characters we don't want to or can't get the expression range from?

I know I've zero context here, feel free to ignore me, just got curious from discord. :smile:

narutosstudent (commented):
I got too curious :joy:

markerikson (approved):

bvaughn:
Playwright tests all pass locally. I think the CI failure was a false negative. 😢

bvaughn:
Ooof. The failure was caused by outdated fixture JSON data 😅 Okay my bad.

PR #8127, opened 2022-11-10
[DRAFT] Rework recording title input to have better focus / edit behavior

markerikson:
This PR:

- _Mostly_ gets the recording title input to show its border on hover and focus, and be more explicit about when it's editing

Generally, changed it from a contenteditable span to a controlled `<input>`, and tried to make the logic handle saving correctly.  I can see a glitch if you delete the text entirely and hit "Save" - it's reverting back to a title from the underlying recording instead of "Untitled", but this is along the right lines.

Styling / behavior is slightly different than what Jon suggested in FE-911 - it shows an outline on hover, and has a single-click to start editing, instead of a double-click.

### DIFF
--- a/src/ui/components/Header/Header.tsx
+++ b/src/ui/components/Header/Header.tsx
@@ -87,36 +87,65 @@ function HeaderTitle({
   recordingId: RecordingId;
 }) {
   const [editing, setEditing] = useState(EditState.Inactive);
-  const inputNode = useRef<HTMLSpanElement>(null);
+  const inputNode = useRef<HTMLInputElement>(null);
+  const [title, setTitle] = useState(recording?.title ?? "Untitled");
   const updateRecordingTitle = hooks.useUpdateRecordingTitle();
   const canEditTitle = recording.userRole !== "none";
+  const isEditing = editing === EditState.Active;
 
-  const className = "ml-2 text-lg p-0.5 whitespace-pre overflow-hidden overflow-ellipsis";
+  const className =
+    "ml-2 text-lg p-0.5 whitespace-pre overflow-hidden overflow-ellipsis rounded-lg border-transparent hover:border-current focus:border-current";
 
   const onKeyPress: React.KeyboardEventHandler = (e: any) => {
     if (e.code == "Enter" || e.code == "Escape") {
       e.preventDefault();
-      inputNode.current!.blur();
+      saveOrReset(e.code === "Enter" ? EditState.Saving : EditState.Inactive);
     }
   };
   const onFocus = () => {
     trackEvent("header.edit_title");
     return editing === EditState.Inactive && setEditing(EditState.Active);
   };
+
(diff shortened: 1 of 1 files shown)


jasonLaster:
Here's some product feedback 
https://www.loom.com/share/63ea66c724604626aa1883e08abd5730

cc @jonbell-lot23

jonbell-lot23:
Hey team! Agreeing with Jason and adding some thoughts:

1. Definitely agree we don't want it full width
2. The border on hover is too bold -- but I do think something on hover is important

Here's what I'd propose:

1. No full width
2. On hover, don't do a border, do a background color (stay tuned next week for the exact variable name)
3. On click, border
4. On second click, enter edit state

It's so uncommon to rename that I think a double-click is a more appropriate level than a single click. It should be a bit harder to get to, because I'd rather it take an extra click than to accidentally capture mis-clicks.

... and that's why I'm referring to highlighting it on hover. It should signal that it is actionable (hover effect) but not be too easy (double-click). If this was more common/desired behaviour, I'd go with a pencil on hover and a single click. But because this feature should be more tucked away, that's why I think this is the best UX.

jasonLaster:
Sounds good. It'd be easy to dismiss this discussions as overkill, but the top left of the app is incredibly visible real estate and these interactions can help give people a positive/negative initial impression.

jonbell-lot23:
Oh I don't think this is overkill at all. This sort of thing is my specialty and I'm happy to see us working on it.

bvaughn:
> 3. On click, border
> 4. On second click, enter edit state

What does a border on click indicate?

For most editable things, I think the focus outline indicates "you can now edit this thing" but in this case, what does it mean?

What if you single click again after you're already in the single clicked state? What if you press Enter?

markerikson:
Just for the record, requiring a multi-single-click sequence to start editing _will_ add to internal complexity of implementation, and wouldn't be my first choice.

jonbell-lot23:
> > 5\. On click, border
> > 6\. On second click, enter edit state
> 
> What does a border on click indicate?
> 
> For most editable things, I think the focus outline indicates "you can now edit this thing" but in this case, what does it mean?
> 
> What if you single click again after you're already in the single clicked state? What if you press Enter?
> 
> > 1. No full width
> 
> Definitely agree with this! 😁

A lot of web and consumer UI is more on the Fischer-Price side of the scale, where hovers and single-clicks do everything you need them to. But you can overdo it, as we did here where we took a very rare action (renaming a Replay) and made it a single click with a huge hit target.

There is precedent for non-80% features being tucked away, and that's what I was going for here. The hover explained actionability, and a second click would get into editing for the people who want to edit.

That said, I need to prioritise my time, so it's fine to go with a single-click here because I don't have the cycles to work through the discussion on it.

bvaughn:
I wasn't trying to say that the "click vs double-click" was a bad decision. I was just asking what the spec was for a few cases that I think are pretty likely to occur.

PR #8130, opened 2022-11-11
New terminal (only added to test harness for the moment)

bvaughn:
I think we have a viable `JSTerm` replacement 😄  This is mostly built with the new type-ahead component I wrote for the log point panel. Eager evaluation is based on the same logic used for the terminal expression Console renderer. All in all, very little code to add this new functionality in which I'm happy about!

https://user-images.githubusercontent.com/29597/201264780-4f6cad44-e5a6-407f-aa9d-80b447d441e7.mp4


bvaughn:
This change only impacts the test harness app, so I'm going to go ahead and land it 😄

PR #8131, opened 2022-11-11
Allow multiple tabs for sources with the same URL (FE-806)

hbenl:
The current source tab logic explicitly avoids showing multiple tabs for sources with the same URL. This means that if there are multiple different sources with the same URL, only one of them will get a source tab and selecting any other of these sources won't show a selected tab (see FE-806).
This PR allows multiple source tabs to be shown in this case and cleans up the tab logic a little.

### DIFF
--- a/src/devtools/client/debugger/src/utils/tabs.ts
+++ b/src/devtools/client/debugger/src/utils/tabs.ts
@@ -56,16 +56,6 @@ export function getTabMenuItems() {
   };
 }
 
-export function isSimilarTab(tab: Tab, url: string, isOriginal?: boolean) {
-  return tab.url === url && tab.isOriginal === isOriginal;
-}
-
-export function persistTabs(tabs: Tab[]) {
-  return [...tabs]
-    .filter(tab => tab.url)
-    .map(tab => {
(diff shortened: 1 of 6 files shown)


src/devtools/client/debugger/src/utils/tabs.ts
```diff
@@ -56,16 +56,6 @@ export function getTabMenuItems() {
...
-export function persistTabs(tabs: Tab[]) {
-  return [...tabs]
-    .filter(tab => tab.url)
-    .map(tab => {
-      const newTab = { ...tab };
-      newTab.sourceId = null;
-      return newTab;
-    });
+export function persistTabs(tabs: Tab[]): Tab[] {
+  return [...tabs].filter(tab => tab.url).map(tab => ({ url: tab.url, sourceId: "" }));
```
bvaughn, line 60: nit: `[...tabs]` seems unnecessary?

`tabs.reduce` might be a more efficient way to manage the filter+map operation too but it's probably a small enough collection of things that it really doesn't matter.

bvaughn (approved):

PR #8132, opened 2022-11-11
Fix the location of column breakpoints (GTM-467)

hbenl:
The conversion from 1-based to 0-based line numbers was lost in the points refactor.

### DIFF
--- a/src/devtools/client/debugger/src/components/Editor/ColumnBreakpoint.tsx
+++ b/src/devtools/client/debugger/src/components/Editor/ColumnBreakpoint.tsx
@@ -71,7 +71,7 @@ export default function ColumnBreakpoint({
     });
 
     const bookmark = doc.setBookmark(
-      { line: point.location.line, ch: point.location.column },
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/components/Editor/ColumnBreakpoint.tsx
```diff
@@ -71,7 +71,7 @@ export default function ColumnBreakpoint({
     });
 
     const bookmark = doc.setBookmark(
-      { line: point.location.line, ch: point.location.column },
+      { line: point.location.line - 1, ch: point.location.column },
```
bvaughn, line 74: I thought Code Mirror used 1-based lines and 0-based columns?

hbenl: I don't know, all I can tell is that we had this before the refactor and it makes column breakpoints show up on the correct line again.

bvaughn (approved):

PR #8133, opened 2022-11-11
Replace JSTerm with new Console terminal input

bvaughn:
- [x] Replace `JSTerm` with the new input (this removes the last of the Code Mirror 5 dependencies)
- [x] Update e2e tests to work with the new input (which has no global methods)
- [x] Delete the old `JSTerm` and `EagerEvalFooter` components

This PR also updates the auto-complete UI to:
- [x] "Tab" and "Enter" keys both accepts the selected suggestion
- [x] Find matches using fuzzy search (rather than `string.startsWith()`
- [x] Sort result list using substring location, string length, and alpha-sort (more closely mimics what Chrome does, but the current implementation is kind of naive)
- [x] Highlight substring character matches within auto-complete list

Console type-ahead demo

* [Loom overview](https://www.loom.com/share/81ab3ba3ea974f32ba3975d685370f34)
* [Preview deploy](https://devtools-git-new-js-term-recordreplay.vercel.app/recording/why-2--283227b5-4d52-4ed5-919c-cb4d76b842e7?point=10384593717703026520674819326872385&time=5971&hasFrames=true)


bvaughn:
The two failing tests (`breakpoints-07` and `console_warp-02`) seem to be pause/scope related, which I would _not_ have expected to change based on this PR.

packages/bvaughn-architecture-demo/components/sources/AutoComplete/utils/findMatches.ts
```diff
@@ -1,66 +1,70 @@
...
   }
 }
 
+function calculateCharacterProximityWeight(delta: number): number {
+  return Math.max(0, 10 - delta);
+}
+
+function calculateCharacterIndexWeight(characterIndex: number): number {
+  return Math.max(0, 10 - characterIndex);
+}
```
bvaughn, line 47: This is really basic.

bvaughn:
Going to split this into smaller PRs: #8134, #8135, and #8136.

PR #8134, opened 2022-11-13
Fix TimelineContextAdapter Pause caching bug

bvaughn:
Split off of #8133

See [3a14ea93-d832-467f-918e-2a285f7e3ee8](https://app.replay.io/recording/warp02-branch-broken--3a14ea93-d832-467f-918e-2a285f7e3ee8) for annotations.

### DIFF
--- a/packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
+++ b/packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
@@ -17,6 +17,8 @@ import { cacheFrames } from "./FrameCache";
 import { preCacheObjects } from "./ObjectPreviews";
 import { cacheScope } from "./ScopeCache";
 
+const callStacksByPauseId: Map<PauseId, CallStack | null> = new Map();
+
 export const {
   getValueSuspense: getPauseIdForExecutionPointSuspense,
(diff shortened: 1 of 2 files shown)


packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
```diff
@@ -54,12 +56,18 @@ export const {
...
+  return callStacksByPauseId.get(pauseId) || null;
+}
+
 export function cachePauseData(
   client: ReplayClientInterface,
   pauseId: PauseId,
   pauseData: PauseData,
   stack?: CallStack
 ) {
+  callStacksByPauseId.set(pauseId, stack || null);
```
hbenl, line 69: The `stack` should only be cached if it comes from `Session.createPause` or `Pause.getAllFrames`, there are other messages that can contain `PauseData` and don't provide a `stack` (even if it isn't empty for that pauseId).

bvaughn: Can we just not pass the `stack` param in those cases then? (It's optional.)

hbenl (approved):

PR #8137, opened 2022-11-13
Polish print statement panel

jasonLaster:
[Loom overview of changes](https://www.loom.com/share/3f0bafa78829478089e4e008f7e869a9)


jonbell-lot23:
Hey gang! I'm doing some triangulation between the original Figma, this branch, production, and some other ideas that have been floating around. Stay tuned, but at a high level this is feeling really good and I'll work to address the remaining issues.

jonbell-lot23:
Here's a Loom where I spend two minutes calling out about 5-10 issues I'm seeing: https://www.loom.com/share/70334d271bcc4e29b94013f57d45666b

Next I'll formalize these in Figma and here so we're all seeing the same thing.

packages/bvaughn-architecture-demo/components/sources/CommentButton.module.css
```diff
@@ -17,9 +17,11 @@
...
   align-items: center;
   justify-content: center;
-  height: 2em;
-  min-width: 2em;
-  border-radius: 2em;
+  height: 2.5em;
+  min-width: 2.5em;
+  border-radius: 2.5em;
+  font-weight: 600;
+  font-family: "Inter";
```
bvaughn, line 24: I've been trying to avoid a smattering of assorted font families in the new components– using the CSS font family variables instead. I think we should stick to that. I think consistency in fonts is good UI.

packages/bvaughn-architecture-demo/components/sources/PointPanel.module.css
```diff
@@ -56,8 +56,8 @@
   overflow: auto;
   height: 100%;
   flex: 1;
-  background-color: var(--point-panel-input-background-color);
-  border: 2px solid transparent;
+  background-color: var(--body-bgcolor);
```
bvaughn, line 59: This variable: `-point-panel-input-background-color`

Exists specifically for this case. If you think it should be a different color, update the variable? 😄

packages/bvaughn-architecture-demo/components/sources/PointPanelTimeline.module.css
```diff
@@ -70,18 +74,18 @@
 
 .HitPointTimelineBackground {
   width: 100%;
-  height: 0.5em;
-  background-color: var(--point-panel-timeline-background-color);
-  border-radius: 0.5em;
+  height: 0.35em;
```
bvaughn, line 77: I've been trying to stick with increments of `0.25em` so we (a) avoid fractional sizes and (b) have a consistent grid for things to line up to. Do we really need 0.35 and 0.8 in this file?

packages/bvaughn-architecture-demo/components/sources/PointPanelTimeline.tsx
```diff
@@ -86,7 +86,7 @@ export default function PointPanelTimeline({
 
   const label =
     currentHitPointIndex !== null
-      ? `${currentHitPointIndex + 1} / ${hitPoints.length}`
+      ? `${currentHitPointIndex + 1}/${hitPoints.length}`
```
bvaughn, line 89: Huh. Was this a Prettier change? 🤔

packages/bvaughn-architecture-demo/components/sources/SourceListRow.module.css
```diff
@@ -145,7 +145,7 @@
   height: var(--line-height);
 }
 .LineHitCountBar0 {
-  background-color: var(--color-hit-counts-bar-0);
+  background-color: var(--point-panel-background-color);
```
bvaughn, line 148: Seems a bit sketchy to use `--point-panel-background-color` variable outside of the point panel. Maybe it's fine?

bvaughn (commented):
A little confused by these changes. It looks like we're going _back_ to more like how things looked before– but I just changed them in #8096 based on Jon's feedback. Are we going in circles here?

jonbell-lot23:
Hey @jasonLaster and @bvaughn , here's a quick loom: 
https://www.loom.com/share/11fea23cc3034ed79a7304cb2d0ee42d

Summary: yes, because Jason was using old visuals, we are revisiting some design approaches (for example bringing back the border on the textifled) and I'm trying to reconcile them in one Figma doc.

jonbell-lot23:
_A little confused by these changes. It looks like we're going back to more like how things looked before– but I just changed them in https://github.com/replayio/devtools/pull/8096 based on Jon's feedback. Are we going in circles here?_

That ticket had changed some colours to be more consistent, and it's true that Jason's branch adjusted them again as a proposal. But there are other items around spacing, line height, font sizes, etc which were going to need another round of polish that we're working through here.

packages/bvaughn-architecture-demo/components/sources/CommentButton.module.css
jonbell-lot23: There's a broader conversation to have here -- it's really good to start with a mathematically rigid approach, but the concept of "optical alignment" means that designs can often not feel right without some special-casing and fine-tuning.

Everyone's in agreement that we don't want to special-case everything because it looks amateur. But one of the most common tensions between design/dev is this topic -- trying to keep everything into a mathematical framework can lead to worse design, counter-intuitively.

So in this case, I'm surprised we need to set inter, but we absolutely need to be able to add more font weight sometimes.

jasonLaster: I did this because I didn't know where the variable was. By the way, this is for the add comment label, not the panel.

packages/bvaughn-architecture-demo/components/sources/PointPanel.module.css
jasonLaster: I think it's nice to remove variables that are unnecessary. With that said, there are several variables that i could probably kill now, but haven't :)

packages/bvaughn-architecture-demo/components/sources/PointPanelTimeline.module.css
jasonLaster: We don't. I can probably take this down to 0.25, but if 0.35 looks better we can add height so that it is 0.5 as a unit.

packages/bvaughn-architecture-demo/components/sources/PointPanelTimeline.tsx
jasonLaster: nope - the figma design had it and it looked better. In reality i think some padding is ideal, this is too little, before was too much.

packages/bvaughn-architecture-demo/components/sources/SourceListRow.module.css
jasonLaster: Ideally we'd have one variable that is more general, but unfortunately the hit counts and print statement panel are introducing more blue themed colors

bvaughn:
I'll review this in the morning :)

jonbell-lot23:
> I'll review this in the morning :)

Sorry these notifications came through on your Sunday!

jonbell-lot23:
Hey gang, check out this Figma link: https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=14587%3A152410&t=jnd6TMkeifAdPdrC-4

We're going to be discussing this more tomorrow, and this is my attempt at showing the (latest, revised) design POV in one place. As we find alignment, we can move this from Figma into actual code. But for now it's meant for discussion.

image

bvaughn:
I see!

I think there's some inefficiencies to the way we're doing this. (For example, the input background color changed from dark background to light background a few days ago, now back to dark background.) Is this because of how we're using Figma+Linear? Or is it just an unavoidable part of the way we want to iterate on our UI?

image

RE: This image– not sure if you're looking for feedback, but:
1. The next/previous hit buttons don't look very button-lick or clickable. (The pill, which is at least for now not clickable, looks the same as the buttons which are clickable.)
2. The dark theme input doesn't look editable to me. Maybe a border would help?

packages/bvaughn-architecture-demo/components/sources/PointPanelTimeline.tsx
bvaughn: Oh, sorry. I misread this at first pass.

packages/bvaughn-architecture-demo/components/sources/SourceListRow.module.css
bvaughn: I don't think we should use variables this way. Changing a _scoped_ variable should not be expected to have impact outside of that scope. (That makes refactoring UI harder.)

packages/bvaughn-architecture-demo/components/sources/PointPanel.module.css
bvaughn: I think we might disagree here a bit.

Our UI mocks often have divergences between light/dark theme that can be hard to implement if we're using building block variables (like `--body-bgcolor`). Variables like `--point-panel-input-background-color` are a way to mange that complexity/scoping. (We can't change `--body-bgcolor` for example, without breaking a lot of things.)

jonbell-lot23:
Loom explorations after our chat are here: https://www.loom.com/share/eabb4e1c44ad4908b99bdeba999219b6

jonbell-lot23:
Hello @bvaughn ! I have some decisions for you on the light theme side of things, so I've written up the code changes in this Notion document: https://www.notion.so/replayio/Print-statement-polish-8dbf67a84aea4fd9bc9f29b0ffbeef7e

I'm trying to give enough documentation to make your work easy without going overboard on documentation that might end up being rewritten soon! So please let me know if you need anything, then we'll see how it feels in production, and once we get things locked in I can turn around and do the design documentation to make sure it's up-to-date.

bvaughn:
@jonbell-lot23 There there dark mode colors?

I'd prefer not to fork the light and dark mode themes so drastically that we have drop shadows on the light theme but not on the dark theme etc. unless you feel strongly that we should do that. If it's just "I haven't had time to get to it yet" then I'd prefer we wait and do both at the same time.

jonbell-lot23:
Ahoy! [Here's a link to Figma](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=14640%3A165942&t=LBRWUdJsW5XGJz1o-4) where I documented each of the colours for light/dark theme, as well as adding the line number stuff. A few notes:

1. We're going to keep the drop shadow on both light and dark theme, but of course in dark theme you won't see it because it's too dark. But in the future we may use a coloured drop shadow to denote things like NUX callouts, etc.
2. Light theme -> dark theme is never 1:1, so there are little differences. For example, I'm calling for a border on the right side of the line number label for light theme, but none for dark theme.
3. On that same element, it looks awkward to have a grey line next to a light blue rectangle. So the design actually calls for no border for the coloured sections. See image ->
image
4. It is likely that in all of my copy/pasting of colors I goofed somewhere when documenting. If so, you can always rely on the visual to get the right color. My apologies if that happens.

When we get this live, any additional colors that need tweaking can probably come off Brian's plate and I can just handle them via simple CSS tickets. I suspect we'll want more fine-tuning but that's more of a me task than anyone else. We need to free up the devs to dev!

bvaughn:
Going to make the exact changes requested _but_ I want to mention out one thing or two:

The difference between `#F3F4F5` (border) and `#F3F4F6` (no hits background color) is too little to be visible, even on a fancy MacBook display:
Screen Shot 2022-11-16 at 2 20 38 PM

Maybe we should consider `#f9fafb` for the background color?
Screen Shot 2022-11-16 at 2 20 57 PM

Alternately we could go with the colors Figma _actually renders_ which is `#f3f4f6` (border) and `#fbfbfb` (no hits background color).

jasonLaster:
Yep - i agree with your suggestion, but we can also tweak some of these things in a second pass as well

jonbell-lot23:
> **Edit** I'm just going to go with the actual color values that Figma is rendering. I think that's probably more in line with your intent. 

Yup, there was an opacity issue there that I just fixed. Apologies, and a shout out to me fearing this would happen -> "4. It is likely that in all of my copy/pasting of colors I goofed somewhere when documenting. If so, you can always rely on the visual to get the right color. My apologies if that happens."

bvaughn:
RE: This comment in Figma:
> `--theme-base-100` (white) is used everywhere you see white. Button background, text background, timeline background. Both light and dark themes.

This kind of contradicts FE-955 which called for using `#e6e6e6` as the background color for inputs.

This was part of why I pushed back against the issue calling for the use of a specific CSS variable. Seems like we want this input to be different than the others (which is fine, we'll manage that with a different variable).

jonbell-lot23:
We chatted in Discord, but just to reiterate -- no apologies necessary! I didn't find that snippy and I appreciate your attention to detail when you see things that don't make sense.

packages/bvaughn-architecture-demo/components/sources/CommentButton.module.css
bvaughn: FWIW we shouldn't use "Inter" (at least not by itself like this) – it's not a standard font and broke the button rendering for me.

bvaughn:
https://www.loom.com/share/3f0bafa78829478089e4e008f7e869a9

bvaughn:
Question. Should we preserve the badge background color for the current step label?

Before:
image

After:
image

I didn't consciously change that but it regressed. I can change it back.

jasonLaster:
@bvaughn lets leave badges out for now. That's an easy follow up

bvaughn:
I think this is good to land then, assuming no complaints.

jasonLaster (commented):
Go for it!

jonbell-lot23:
Thank you! Excited :)

PR #8139, opened 2022-11-14
Replace JSTerm with new autocomplete (with a different fix for selecting the top frame)

hbenl:


bvaughn (commented):
Looking at commit 20b1fc96cf9b3be72e75b3c55f21807d97c82433 only– this change looks good! Maybe re-target the PR from `main` to my branch before merging though?

The only thing I'm left wondering (not necessarily in the scope of this PR) is "can we name `hasFrames` better?" because it's kind of confusing to me...but maybe it goes away anyway once we rip out the rest of the old `Pause` cache?

bvaughn (approved):

hbenl:
> Maybe re-target the PR from main to my branch before merging though?

OK

> The only thing I'm left wondering (not necessarily in the scope of this PR) is "can we name hasFrames better?" because it's kind of confusing to me...but maybe it goes away anyway once we rip out the rest of the old Pause cache?

Yeah, `hasFrames` will be removed together with the `Pause`. The `paused` redux action interprets `hasFrames` as telling it whether we want to open the top frame's source (e.g. if you're seeking to a console message) or not (if you're clicking on a point in the timeline). This flag should not come from the `Pause` but from the caller of the `seek` action (which ultimately triggers the `paused` action) and should have a better name.

bvaughn:
Agreed!

PR #8141, opened 2022-11-14
Remove Pause.getFrameSteps() and move the FrameSteps cache

hbenl:
Part of FE-949


packages/bvaughn-architecture-demo/src/suspense/FrameStepsCache.ts
```diff
@@ -0,0 +1,33 @@
...
+  getValueIfCached: getFrameStepsIfCached,
+} = createGenericCache2<
+  ReplayClientInterface,
+  [pauseId: PauseId, frameId: FrameId],
+  PointDescription[] | undefined
+>(
+  async (client, pauseId, frameId) => {
+    try {
+      const frameSteps = await client.getFrameSteps(pauseId, frameId);
+      const updatedFrameSteps = frameSteps.map(frameStep => cloneDeep(frameStep));
```
bvaughn, line 21: Just kind of curious if this would work?
```suggestion
const updatedFrameSteps = cloneDeep(frameSteps);
```
No need to change it regardless.

src/ui/suspense/frameStepsCache.ts
```diff
@@ -1,41 +1,19 @@
...
-    }
-  },
-  (pauseId, frameId) => `${pauseId}:${frameId}`
-);
+  getFrameStepsAsync,
+  getFrameStepsIfCached,
+} from "bvaughn-architecture-demo/src/suspense/FrameStepsCache";
+import { ReplayClientInterface } from "shared/client/types";
 
 export const useGetFrameSteps = createUseGetValue<
```
bvaughn, line 10: I think I'm getting a little less clear on the purpose of this cache (in `src/ui/suspense`) as being separate from the caches in bvaughn

packages/shared/client/ReplayClient.ts
```diff
@@ -340,6 +340,12 @@ export class ReplayClient implements ReplayClientInterface {
     return countsObject;
   }
 
+  async getFrameSteps(pauseId: PauseId, frameId: FrameId): Promise<PointDescription[]> {
```
bvaughn, line 343: I wonder if we could name this method something a little more descriptive (for people like me who get confused by it) – maybe even just something like `getFrameStepPoints` or something?

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -283,7 +283,11 @@ async function getResumePoint(replayClient: ReplayClientInterface, state: UIStat
   if (!frames || !frame || frame === frames[0]) {
     return;
   }
-  const frameSteps = await getFrameStepsAsync(selectedFrameId.pauseId, selectedFrameId.frameId);
+  const frameSteps = await getFrameStepsAsync(
```
bvaughn, line 286: I'm a little confused about why this function now calls both `getFramesAsync` and `getFrameStepsAsync`. Seems like we can remove the `getFramesAsync` call now? We only really need the steps?

bvaughn (approved):

packages/shared/client/ReplayClient.ts
hbenl: Hmm, there's dozens of places in the code where we talk about "FrameSteps", so I'd like to keep the name here for consistency.

bvaughn: I guess that's fair but also...isn't it an array of execution points? Do we ever call "points" "steps" except for here?

src/devtools/client/debugger/src/reducers/pause.ts
hbenl: We still check if the currently selected frame is the top frame and bail out in that case.

bvaughn: D'oh. I misread that. Ok.

PR #8143, opened 2022-11-14
Remove Pause.repaintGraphics()

hbenl:
Part of FE-949


bvaughn (approved):
I'm not sure I understand the sequencing here. The new `repainted-graphics-cache` seems pretty similar to a Suspense cache (except that it supports a `force` flag) but it's located in the _protocol_ package so... 😅 

I trust your direction here just not sure I understand it

hbenl:
I chose not to make this a suspense cache because of the `force` flag (which means it doesn't exactly fit the logic of the generic cache) and it's not used by a React component but by imperative code.

PR #8144, opened 2022-11-14
Remove Pause.evaluate()

hbenl:


bvaughn (approved):

bvaughn:
Need to update the new Playwright test fixture JSON data I think, since this change affects the params we're sending to the backend.

https://github.com/replayio/devtools/blob/main/packages/bvaughn-architecture-demo/playwright/README.md#updating-protocol-fixture-data

hbenl:
I've changed the PR so that `ThreadFront.evaluateNew()` calls the protocol client directly instead of using the evaluation cache: I felt uneasy about caching evaluations, especially those used by the react devtools frontend to communicate with the backend - if there's anything stateful in that communication, caching could break it.

bvaughn:
Seems fine

PR #8145, opened 2022-11-14
Source viewer should not re-request hit counts for lines

bvaughn:
Related to FE-934, although I think the real "bug" there is a backend one (BAC-2441).

This PR updates the Source Suspense cache so that it doesn't request hit counts for the same set of lines more than once. Tests are included, since this caching behavior is more complex than our typical Suspense cache.


bvaughn:
Not sure what went on with the e2e tests in CI on that first run. Ran a few locally and they all passed so I'm going to assume it was a transient issue with the backend.

markerikson (approved):

PR #8151, opened 2022-11-15
Add the ability to cache scan data

jcmorrow:
And change what it means for regions to be "loaded". We have several different things we track on the the region level:

- Loading: are we ever going to try and load this region given the current focus window?
- Indexed: have we processing the manifests which produce scan data for this region, which allows things like hitCounts to work
- Loaded: Have we done a run-to-point and fork at the beginning of this region yet or not?

Ironically, for the colloquial "loading" case (can I interact with the recording) we don't really care about the `loaded` state as defined above. If we have not RTP and forked in a region, then doing things like pause operations/analyses will be slow there, but we can totally get started queueing up those requests, they'll just take a little longer to respond. By not including that in the loading percentage we can get people interacting with editor, hitcounts, paints, etc. faster.

As always, it's been a while since I was fluent in devtools, so happy to get feedback from @bvaughn or @markerikson if there are better ways to accomplish these same goals!

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -198,9 +198,10 @@ export function createSocket(
       const experimentalSettings: ExperimentalSettings = {
         // Uncomment this to get a new session every time you refresh
         // controllerKey: String(Math.floor(Math.random() * 10e8)),
-        listenForMetrics: !!prefs.listenForMetrics,
+        cacheScanData: !!features.cacheScanData,
         disableCache: !!prefs.disableCache,
         disableQueryCache: !features.enableQueryCache,
+        listenForMetrics: !!prefs.listenForMetrics,
         profileWorkerThreads: !!features.profileWorkerThreads,
       };
 
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -100,7 +100,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
                   className="overflow-hidden overflow-ellipsis whitespace-pre"
                   title={recording.metadata.source?.branch}
                 >
-                  {time} ago
+                  {time === "Now" ? time : time + " ago"}
                 </div>
               </Row>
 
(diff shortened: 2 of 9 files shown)


src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -100,7 +100,7 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
                   className="overflow-hidden overflow-ellipsis whitespace-pre"
                   title={recording.metadata.source?.branch}
                 >
-                  {time} ago
+                  {time === "Now" ? time : time + " ago"}
```
jcmorrow, line 103: This was unrelated, I just happened to notice that sometimes we say `Now ago`.

jasonLaster (approved):

bvaughn (commented):
I feel a little silly but...is the new `cacheScanData` setting even used for anything?

Looks like you changed a few places from index+loaded to indexed, which is okay (I think) based on your PR description.

But I don't understand what the new experimental setting is meant to be used for? What are we "scanning" or "caching" here?

src/ui/actions/session.ts
```diff
@@ -198,9 +198,10 @@ export function createSocket(
       const experimentalSettings: ExperimentalSettings = {
         // Uncomment this to get a new session every time you refresh
         // controllerKey: String(Math.floor(Math.random() * 10e8)),
-        listenForMetrics: !!prefs.listenForMetrics,
+        cacheScanData: true,
```
jcmorrow, line 201: Thanks for that comment @bvaughn! The setting was *supposed* to be here! But I forgot to thread it through 💀

bvaughn: Gotcha. So whatever value you have set at the start of your session gets passed to the API. If you change it, you'd have to reload for it to have an impact?

bvaughn (approved):

src/ui/actions/session.ts
jcmorrow: Yep, if you change it you'll have to refresh, and because your session settings have changed you will automatically get a new session.

PR #8154, opened 2022-11-15
Clicking on a comment should not open the DevTools tab

bvaughn:
See [ff01dbee-81e1-4737-9108-27cc62778104](https://app.replay.io/recording/clicking-on-a-comment-opens-devtools-tab--ff01dbee-81e1-4737-9108-27cc62778104) for a bit of background.

Essentially, the high level sequence before this change was:
1. Clicking on a comment called `seekToComment` (`ui/actions/comments`)
1. `seekToComment` calls `seek` (`ui/actions/timeline`)
1. `seek` calls `ThreadFront.timeWarp`
1. `timeWarp` emits a "paused" event
1. The `paused` action creator is listening for "paused", and it calls `setLocation`
1. `selectLocation` calls `setViewMode` (which opens DevTools)

This is a lot of indirection. I think the following changes seem reasonable to me though:
1. `seekToComment` should only open the "DevTools" tab if you click on the location preview. (Otherwise it should just seek.)
1. The `paused` action creator should _never_ open a tab directly. It's too far removed from the user interaction to know what the right behavior should be.
1. The `selectLocation` action should only change the view mode (to "DevTools") if it's also been told to open the Sources tab.

[Loom demo](https://www.loom.com/share/db6b2938456142528db18ee33a23b5f6)

### DIFF
--- a/src/devtools/client/debugger/src/actions/pause/paused.ts
+++ b/src/devtools/client/debugger/src/actions/pause/paused.ts
@@ -85,7 +85,12 @@ export function paused({
         currentLocation.column !== selectedFrame.location.column
       ) {
         const { selectLocation } = await import("../sources");
-        dispatch(selectLocation(cx, selectedFrame.location, true));
+        // This action creator is triggered whenever we seek to a new exception point.
+        // It's pretty far removed from the cause of the seek,
+        // so it shouldn't make the decision of whether to change UI tabs.
+        const openSourcesTab = false;
+
+        dispatch(selectLocation(cx, selectedFrame.location, openSourcesTab));
       }
 
       if (pause !== ThreadFront.currentPause) {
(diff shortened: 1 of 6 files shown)


bvaughn:
> It looks like the `openSourcesTab` flag is only used to decide if we should switch view mode but the source tab is still opened even if `openSourcesTab` is `false`

Not sure I understand. That param is referenced twice:

```js
if (openSourcesTab) {
  if (getViewMode(getState()) == "non-dev") {
    dispatch(setViewMode("dev"));
  }
}

// ...

if (layout !== "ide" && openSourcesTab) {
  dispatch(setSelectedPanel("debugger"));
}
```

Are you referring to the call to `setShownSource`? That just changes which sources tab is in the "front".

hbenl:
OK, I had misunderstood the flag - I thought it should disable opening (a tab for) the source but it disables switching to the debugger panel (and I had overlooked that second reference).

hbenl (approved):

PR #8155, opened 2022-11-15
Recording title size/focus/edit tweaks

bvaughn:
Pickup up #8127 since Mark's on vacation.

~~This PR basically implements this comment: https://github.com/replayio/devtools/pull/8127#issuecomment-1312328180 - [Loom demo](https://www.loom.com/share/6221692bbe0945e2a5b1deaec83cc175)~~

[Loom demo](https://www.loom.com/share/0f7ede192f5147ebb85e72a17306835e)

### DIFF
--- a/packages/bvaughn-architecture-demo/components/sources/AutoComplete/utils/contentEditable.ts
+++ b/packages/bvaughn-architecture-demo/components/sources/AutoComplete/utils/contentEditable.ts
@@ -61,14 +61,18 @@ export function getCursorIndexAfterPaste(
 export function selectAllText(contentEditable: HTMLElement): void {
   contentEditable.focus();
 
-  const selection = window.getSelection();
-  if (selection) {
-    const range = document.createRange();
-    range.selectNodeContents(contentEditable);
+  // HACK
+  // Waiting until the end of the microtask queue works around a selection bug in Safari.
+  setTimeout(() => {
+    const selection = window.getSelection();
+    if (selection) {
+      const range = document.createRange();
+      range.selectNodeContents(contentEditable);
 
-    selection.removeAllRanges();
-    selection.addRange(range);
-  }
+      selection.removeAllRanges();
+      selection.addRange(range);
+    }
+  }, 0);
 }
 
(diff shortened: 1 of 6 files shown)


jonbell-lot23 (commented):
Hi! I need to focus on other higher-pri items but I can give some quick feedback to get this out the door.

[I started writing notes but it's taking too long, so I'm going to dial down my level of design support]

* Let's move to one-click, that command-A "select all" scenario is likely
* On hover, the background color should be really faint. The bg color we use in console for "filter input" is fine, although that variable is out of sync with the variable we use for this, --theme-text-field-bgcolor. 
* Moving to one click means we can auto-select the text on that first click, meaning it'll stand out more

I apologise for the lack of details, I'm one designer stretched thin.

bvaughn:
Should be ready to go then, with the revised/more simplified ux.

jonbell-lot23 (commented):
It's feeling really nice! Can you:

1. Add a more noticeable hover color in dark theme? The grey in var(--theme-toggle-bgcolor) is probably a good one to match. (Notably this means the different themes using different variables -- this is super common (and annoying) when aligning themes. Dark theme never works as a direct 1:1 with light theme because of optical/psychological stuff)
2. Right now when I click in Safari I'm seeing a flash of everything being selected, then it de-selects. It's not happening in Chrome and is therefore not a priority. But FYI.

Thank you! <3 Excited to get this long-standing frustration fixed.

bvaughn:
That dark-theme hover color seems a lot higher contrast than the light-theme one, but I'll plug it in sure thing! 😄

bvaughn:
Thanks. Made that hover color change and fixed the Safari issue.

jonbell-lot23:
Love it, thanks!

I did a quick check with the new contrast -- I agree it's higher contrast but yeah, super common to have to break out of a 1:1 relationship between dark and light theme because they have such different feels.

Was pleased to note that the colour I recommended is indeed somewhat standardised, although of course not perfect. 
image

Will file a bug for this one outlier.

PR #8157, opened 2022-11-15
Improve AutoComplete input paste/cut operations

bvaughn:


bvaughn:
Going to land this bc it's a bug fix and I don't think anyone's around to review it at the moment. Happy to change it in a follow up PR.

PR #8159, opened 2022-11-15
Add Controller Key experimental feature

jcmorrow:
This is a nice flag to have in prod, because it allows you to easily get a new session whenever you want. It's not super useful for users, so I've stuck it under the experimental tab. This is especially useful for testing scan data caching, since you need to get a new session to see the caches come in.


src/ui/components/shared/UserSettingsModal/ExperimentalSettings.tsx
```diff
@@ -64,6 +64,11 @@ const EXPERIMENTAL_SETTINGS: ExperimentalSetting[] = [
     description: "Cache the results of indexing the recording",
     key: "cacheScanData",
   },
+  {
+    label: "New Controller on Refresh",
```
jcmorrow, line 68: I don't like leaking the concept of a controller into the frontend, but since this is buried in the experimental settings, I feel ok about it.

bvaughn (approved):

PR #8163, opened 2022-11-16
Remove the Pause (FE-949)

hbenl:
... and replace the `hasFrames` flag with an `openSourcesTab` parameter to seek()


packages/protocol/thread/bounds.ts
```diff
@@ -1,18 +1,15 @@
...
 // that might not be loaded yet, for the node picker.
 export class NodeBoundsFront {
-  private _pause: Pause;
   private _rects: Rect[];
   nodeId: ObjectId;
   then = undefined;
 
-  constructor(pause: Pause, nodeId: ObjectId, rects: Rect[]) {
-    this._pause = pause;
+  constructor(nodeId: ObjectId, rects: Rect[]) {
```
bvaughn, line 12: Wouldn't this be more accurately/consistently typed as `PauseId`?

Both happen to be aliases for `string` but it still seems worth changing.

packages/protocol/thread/thread.ts
```diff
@@ -64,7 +72,7 @@ export interface Source {
 export interface PauseEventArgs {
   point: ExecutionPoint;
   time: number;
-  hasFrames: boolean;
+  openSourcesTab: boolean;
```
bvaughn, line 75: I dig this rename 👍🏼

packages/protocol/thread/thread.ts
```diff
@@ -205,13 +210,22 @@ class _ThreadFront {
...
-    return this._currentPause;
+  get currentPause(): Pause {
+    return {
+      point: this.currentPoint,
+      time: this.currentTime,
+      pauseId:
+        this.currentPauseId ??
+        getPauseIdForExecutionPointIfCached(this.currentPoint)?.value ??
+        null,
+    };
```
bvaughn, line 221: This seems like an unexpected change to me. Was it done so that we could remove the `ensurePause` waiter? I'm not sure I know what the potential effects of this change are.

packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
```diff
@@ -36,6 +36,34 @@ export const {
...
+
+export async function getPauseIdAsync(
+  replayClient: ReplayClientInterface,
+  point: ExecutionPoint,
+  time: number
+) {
+  const pauseId = await getPauseIdForExecutionPointAsync(replayClient, point);
+  pointAndTimeByPauseId.set(pauseId, { point, time });
+  return pauseId;
+}
```
bvaughn, line 65: Seems like maybe the changes in this file are unnecessary? I think we could just do this, couldn't we?
```js
const pointAndTimeByPauseId = new Map<
  PauseId,
  { point: ExecutionPoint; time: number }
>();

export const {
  getValueSuspense: getPauseIdForExecutionPointSuspense,
  getValueAsync: getPauseIdForExecutionPointAsync,
  getValueIfCached: getPauseIdForExecutionPointIfCached,
} = createGenericCache2<
  ReplayClientInterface,
  [executionPoint: ExecutionPoint],
  PauseId
>(
  async (client, executionPoint) => {
    const createPauseResult = await client.createPause(executionPoint);
    await client.waitForLoadedSources();
    cachePauseData(
      client,
      createPauseResult.pauseId,
      createPauseResult.data,
      createPauseResult.stack
    );

    // This is the only new line
    pointAndTimeByPauseId.set(pauseId, { point, time });

    return createPauseResult.pauseId;
  },
  (executionPoint) => executionPoint
);

export function getPointAndTimeForPauseId(pauseId: PauseId) {
  return pointAndTimeByPauseId.get(pauseId);
}
```

bvaughn (approved):
Mostly skimmed but looks good. I assume you've tested to make sure we don't regress on things like FE-953.

packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
hbenl: > `pointAndTimeByPauseId.set(pauseId, { point, time });`

But where do we get the point/time from?

packages/protocol/thread/thread.ts
hbenl: To simplify the refactor, I'm returning an object containing `point`, `time` and `pauseId` as a replacement for the `Pause`. `Pause` contained a `pauseId` that may initially be set to `null` and then to the `pauseId` once the pause was created. Now we have that `pauseId` in the suspense cache, so pulling it from there should have the same effect.

packages/protocol/thread/bounds.ts
hbenl: This isn't used anymore anyway so I'll just delete the file.

packages/bvaughn-architecture-demo/src/suspense/PauseCache.ts
bvaughn: Yeah, sorry. Realized that a minute or two after posting.

packages/protocol/thread/thread.ts
bvaughn: I guess if you've tested it thoroughly enough that you're confident 😁

hbenl:
The `hasFrames` -> `openSourcesTab` part broke this PR, so I'll split it off.

PR #8165, opened 2022-11-16
Improve focus region error handling

bvaughn:
[Loom overview of change](https://www.loom.com/share/4c9039a37f3945b0a7c16023dfea0b0c)

This PR changes multiple things. Commits are atomic to simplify review.
- [x] Wrap Console errors in boundary so they don't take down the larger app.
- [x] Timeline reducer shouldn't accept/apply an invalid focus region.
- [x] `TestCase` UI shouldn't apply an invalid focus region
- [x] Console and `LoggablesContext` should more gracefully handle message loading failures.

Note that d83d6b0c9d0a7aa5fdd3759d30aa9a5db330a19e should not be necessary, but it seems like TypeScript types aren't correct for the Test Suites integration. We probably need to fix this type (or the upstream bug that caused the data to be broken?)
https://github.com/replayio/devtools/blob/93459cfb6a3d147d4329d6647381834b4f4f5dd3/src/ui/types/index.ts#L169-L179

### DIFF
--- a/packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
+++ b/packages/bvaughn-architecture-demo/components/ErrorBoundary.module.css
@@ -1,6 +1,10 @@
 .Error {
-  padding: 0.5em;
-  text-align: center;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
   background-color: var(--background-color-error);
   color: var(--color-default);
 }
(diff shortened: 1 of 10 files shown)


jaril (approved):
LGTM! Filed a separate issue for me to clean up the test types a bit later today https://linear.app/replay/issue/SCS-290/focusing-on-a-test-sometimes-breaks-the-timeline

bvaughn:
Looks like `TestCase.tsx` has changed in a way that required me to change my workaround too. Should be okay to remove once the types get fixed?

PR #8172, opened 2022-11-17
Don't request hit counts outside of loaded regions

bvaughn:
The gist of the bug was this:
* When editing the focus range, we update the (pending) range in memory and attempt to also show updated values for hit counts, filtered console messages and networks requests, etc.
* If the focus range contracts, we can filter in memory, but if it expands– we need to fetch new hit count values remotely.
* Unfortunately, we don't actually update the focus range (on the backend) until the user hits "Save"
* This means we were potentially requesting hit counts _outside of the focus region_, and so the values returned by the backend were not valid. To make matters worse, our Suspense cache _cached_ these invalid responses so that even after we applied the new focus range, we continued to show incorrect values.

This PR makes the smallest commit I could think of to fix this:
* The `ReplayClient` automatically waits for the focus range to be fully loaded before fetching hit counts or running analysis.
* Source hit counts dim slightly while fetching newer data (so it's not confusing to see them jump-change on a delay).

I'm seeing an error in the Console while fetching messages that may be related to this change (not sure). Need to dig in.

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -53,7 +53,7 @@ import { RecordingCapabilities } from "protocol/thread/thread";
 import { binarySearch, compareNumericStrings, defer } from "protocol/utils";
 import { TOO_MANY_POINTS_TO_FIND } from "shared/constants";
 import { ProtocolError, isCommandError } from "shared/utils/error";
-import { isPointInRegions } from "shared/utils/time";
+import { isRangeInRegions, toPointRange } from "shared/utils/time";
 
 import {
   HitPointStatus,
@@ -612,6 +612,10 @@ export class ReplayClient implements ReplayClientInterface {
 
     const hitCounts: LineNumberToHitCountMap = new Map();
 
+    // Don't try to fetch hit counts in unloaded regions.
+    // The result might be invalid (and may get cached by a Suspense caller).
+    await this.waitForLoadedRegions(focusRange);
+
     await Promise.all(
       correspondingSourceIds.map(async sourceId => {
         const { hits: protocolHitCounts } = await client.Debugger.getHitCounts(
@@ -801,6 +805,10 @@ export class ReplayClient implements ReplayClientInterface {
         locations,
       };
 
+      // Don't try to run analysis in unloaded regions.
+      // The result might be invalid (and may get cached by a Suspense caller).
+      await this.waitForLoadedRegions(params.range || null);
+
       try {
         await analysisManager.runAnalysis(analysisParams, {
           onAnalysisError: (error: unknown) => {
@@ -879,18 +887,19 @@ export class ReplayClient implements ReplayClientInterface {
     }
   }
 
-  async waitForLoadedRegions(focusRange: PointRange | null): Promise<void> {
+  async waitForLoadedRegions(focusRange: TimeStampedPointRange | PointRange | null): Promise<void> {
     return new Promise(resolve => {
       const checkLoaded = () => {
         const loadedRegions = this.loadedRegions;
         let isLoaded = false;
-
         if (loadedRegions !== null) {
-          isLoaded =
-            focusRange !== null
-              ? isPointInRegions(focusRange.begin, loadedRegions.loaded) &&
-                isPointInRegions(focusRange.end, loadedRegions.loaded)
-              : loadedRegions.loading.length === 0;
+          if (focusRange !== null) {
+            const pointRange = toPointRange(focusRange);
+
+            isLoaded = isRangeInRegions(pointRange.begin, pointRange.end, loadedRegions.loaded);
+          } else {
+            isLoaded = loadedRegions.loaded.length > 0 && loadedRegions.loading.length === 0;
+          }
         }
 
         if (isLoaded) {
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -53,6 +53,7 @@ import { RecordingCapabilities } from "protocol/thread/thread";
 import { binarySearch, compareNumericStrings, defer } from "protocol/utils";
 import { TOO_MANY_POINTS_TO_FIND } from "shared/constants";
 import { ProtocolError, isCommandError } from "shared/utils/error";
+import { isPointInRegions } from "shared/utils/time";
 
 import {
   HitPointStatus,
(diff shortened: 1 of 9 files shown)


bvaughn:
Test failures likely unrelated to this change. We are currently having backend/capacity problems.

jcmorrow:
I think based on what I was seeing in https://linear.app/replay/issue/FE-966/analysis-is-not-rerun-upon-changing-focus-window, I'm surprised that this change (which seems entirely about hit counts, not analyses) would fix that?

bvaughn:
> I think based on what I was seeing in https://linear.app/replay/issue/FE-966/analysis-is-not-rerun-upon-changing-focus-window, I'm surprised that this change (which seems entirely about hit counts, not analyses) would fix that?

I'm still testing this. Trying to prepare the fix in between a bunch of meetings this morning.

packages/shared/client/ReplayClient.ts
```diff
@@ -878,6 +887,36 @@ export class ReplayClient implements ReplayClientInterface {
...
+      const checkLoaded = () => {
+        const loadedRegions = this.loadedRegions;
+        let isLoaded = false;
+        if (loadedRegions !== null) {
+          if (focusRange !== null) {
+            const pointRange = toPointRange(focusRange);
+
+            isLoaded = isRangeInRegions(pointRange.begin, pointRange.end, loadedRegions.loaded);
+          } else {
+            isLoaded = loadedRegions.loading.length === 0;
```
jcmorrow, line 901: I'm not sure that it matters, because I suspect it's impossible (at least, I hope so), but the case where there are no regions being loaded seems like it would be an error.

packages/shared/client/ReplayClient.ts
```diff
@@ -611,6 +612,10 @@ export class ReplayClient implements ReplayClientInterface {
 
     const hitCounts: LineNumberToHitCountMap = new Map();
 
+    // Don't try to fetch hit counts in unloaded regions.
+    // The result might be invalid (and may get cached by a Suspense caller).
+    await this.waitForLoadedRegions(focusRange);
```
jcmorrow, line 617: What happens if the user drags the window wide, triggering this request, but then cancels and never hits save, or makes the region smaller again and hits save? Does this just `await` forever?

jcmorrow (commented):
I think all of the logic for checking whether the point/range is covered all looks solid, and it makes sense that this approach would work (although I do think the potential "wait forever" case might be a bit weird), but I suspect that it would be cleaner to not issue any requests until the focus window has been changed and the calls to `unloadRegion` and `loadRegion` have completed. What do you think? I know it gets a bit weird, because it means we get in-memory filtering during the shrinking window case, which makes it feel like the UI is responding to the window changing, but then we get nothing in the window expanding case, which feels broken. I'm not sure how to reconcile that, but it seems like `await`ing those regions loading every time the user drags the slider, even if they don't hit save, feels a bit awkward. I don't know if I voiced that clearly, lmk if you wanna chat this one through!

packages/shared/client/ReplayClient.ts
jcmorrow: Or will it like potentially batch a bunch of requests up and then issue them when the focus window goes wide again?

bvaughn: Potentially, but it doesn't matter. Either...
1. The new focus window will contain this range, in which case it would silently complete later (and go into the Suspense cache) or
2. The new focus window won't contain this range, in which case it will continue to wait until/unless the user updates the region again, in which case it would resume from where it left off (b'c the promise is cached)

packages/shared/client/ReplayClient.ts
bvaughn: Good thought. I'll add a guard for it.

packages/shared/client/ReplayClient.ts
jcmorrow: > 1. The new focus window will contain this range, in which case it would silently complete later (and go into the Suspense cache)

Yeah, I guess I'm curious if we might queue up like 50 requests while the user is dragging the window around and then burst 50 requests when the load command finishes. That seems likely to be really inefficient, but maybe if they are all really fast then it does not matter 🤷

bvaughn:
> I suspect that it would be cleaner to not issue any requests until the focus window has been changed and the calls to unloadRegion and loadRegion have completed.

This is what we're doing. No requests are made (nothing gets sent over the `WebSocket`) until it's within a loaded region. We don't want until regions have _finished_ loading (but I suspect you're not suggesting we should do that).

I think I hear what you're suggesting, but I don't think it's straight forward to build it. The UX spec (as I recall it) called for things (console logs, network requests, hit counts) to update in realtime as the user drags the focus window, to help them narrow down. If they're dragging within already fetched territory, we do that (filter in-memory) but if they drag outside– we can't do that, so we wait.

I _think_ the only thing you're suggesting we do differently is:
1. Don't even initiate a request if it's outside of the loaded region (bail out early) but I think that would be difficult to do because if how Suspense caching works. We could do it on a cache-by-cache basis, but I was hoping to fix "all" issues with this change. (We could optimize a specific cache later.)
2. Don't show the _is pending_ dimmed style when we get to this case. (Not sure if you're actually asking for this or not, but I think it would be a bad change because suddenly the count would stop updated– _and be wrong_- and the UI would give no hints that this had happened.)

packages/shared/client/ReplayClient.ts
bvaughn: Yes, we would queue up a few. I don't think this makes a big difference _in practice_ because they're generally pretty fast. It's not ideal though. TBH I'm feeling a little overwhelmed at the moment though and I'm just looking to fix things (even if not in the most efficient way)

jcmorrow: Yeah, I totally trust your judgement 💯

jcmorrow (approved):

packages/shared/client/ReplayClient.ts
bvaughn: Thank you for the discussion 🙇🏼

jcmorrow:
> This is what we're doing. No requests are made (nothing gets sent over the WebSocket) until it's within a loaded region. We don't want until regions have finished loading (but I suspect you're not suggesting we should do that).

Yeah, we don't need them to finish loading, the backend will make sure that the regions are loaded before it responds, as long as the request to load them came in first.

> I think I hear what you're suggesting, but I don't think it's straight forward to build it. The UX spec (as I recall it) called for things (console logs, network requests, hit counts) to update in realtime as the user drags the focus window, to help them narrow down. If they're dragging within already fetched territory, we do that (filter in-memory) but if they drag outside– we can't do that, so we wait.

Console messages are indeed weird, because you can request them in a specific range for overflow reasons, but I think with the exception of console messages this should not affect any of the basic data components like network requests. It's just the dynamic stuff (basically analyses, hit counts, and pauses) where the request to load the region has to come in before the request to analyze/create pause/count hits. So if it makes it any easier, I think we could restrict the changes to those resources.

>I think the only thing you're suggesting we do differently is:
>
> 1. Don't even initiate a request if it's outside of the loaded region (bail out early) but I think that would be difficult to do because if how Suspense caching works. We could do it on a cache-by-cache basis, but I was hoping to fix "all" issues with this change. (We could optimize a specific cache later.)
> 2. Don't show the is pending dimmed style when we get to this case. (Not sure if you're actually asking for this or not, but I think it would be a bad change because suddenly the count would stop updated– and be wrong- and the UI would give no hints that this had happened.)

Yeah, I was suggesting 1, had not even thought about 2! But I def don't understand all of the reactions with suspense, so this all seems reasonable to me!

bvaughn:
I think some of the e2e tests are legit. Need to dig in.

packages/shared/client/ReplayClient.ts
```diff
@@ -878,6 +887,36 @@ export class ReplayClient implements ReplayClientInterface {
...
+      const checkLoaded = () => {
+        const loadedRegions = this.loadedRegions;
+        let isLoaded = false;
+        if (loadedRegions !== null) {
+          if (focusRange !== null) {
+            const pointRange = toPointRange(focusRange);
+
+            isLoaded = isRangeInRegions(pointRange.begin, pointRange.end, loadedRegions.loaded);
+          } else {
+            isLoaded = loadedRegions.loaded.length > 0 && loadedRegions.loading.length === 0;
```
bvaughn, line 901: Oh. This isn't right. `loading` will never get cleared. `loaded` will just equal `loading` 🤪

PR #8177, opened 2022-11-17
Support repainting

jasonLaster:


jasonLaster:
Landing because it'll help us learn

PR #8181, opened 2022-11-18
Source search panel

bvaughn:
This is a fairly large PR. If the size of the PR is a concern, it could be broken down into 2-3 separate PRs.

At a high-level it adds a new Sources search panel (to replace the one that was accidentally deleted during the old Editor removal). At a slightly lower level, it does a few things:
- [x] Rewrite Source Search panel (using Suspense caches and other shared components).
  - [x] Integrate the new panel into the legacy app and re-add the side menu option.
  - [x] Add new e2e tests for the new panel.
- [x] Update the file search to clear the selected index between Sources. This more closely mirrors VSCode's behavior and prevents the file-search state from interfering with the sources-search state.
  - [x] Added/updated tests around this as well.
- [ ] Optionally integrate with new `limit` backend param (BAC-2495). (Can't do this yet because the protocol API package has not been released. We can always do this as a follow up though.)

~~There are a couple of open questions that I'm holding off on Monday to sync with @hbenl about. The old code used [this helper](https://github.com/replayio/devtools/blob/e5ea6098528ae3c60aef058bc01125c16601f48c/src/ui/reducers/sources.ts#L316-L318) to sort sources so that original sources come first, but the new code doesn't have access to that information. It also used [this helper](https://github.com/replayio/devtools/blob/e5ea6098528ae3c60aef058bc01125c16601f48c/src/ui/reducers/sources.ts#L316-L318) to filter out prettified sources. Presumably both helpers could be injected into `ReplayClient` like we've done in other places.~~ For now, I'll just do this and we can revisit.

[**Here is a slightly longer Loom overview**](https://www.loom.com/share/d052c19cb5004b7ba0871fb25b1de1a0) and [**here is an integration demo**](https://www.loom.com/share/2dd2e6cc89c441d6b1467ce003d70c65).

Kapture 2022-11-20 at 15 32 52


packages/bvaughn-architecture-demo/components/sources/SyntaxHighlightedLine.tsx
```diff
@@ -5,20 +5,27 @@ import { parse } from "bvaughn-architecture-demo/src/suspense/SyntaxParsingCache
...
 
-function SyntaxHighlightedLineSuspends({ code, fileExtension = ".js" }: Props) {
+function SyntaxHighlightedLineSuspends({ code, className = "", fileExtension = ".js" }: Props) {
   const htmlLines = parse(code, fileExtension) || [];
 
   return (
-    <span className={styles.Code} dangerouslySetInnerHTML={{ __html: htmlLines.join("\n") }} />
+    <span
+      className={`${className} ${styles.Code}`}
+      dangerouslySetInnerHTML={{ __html: htmlLines.join("\n") }}
```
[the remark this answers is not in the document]
bvaughn: Shut up. I know.

packages/bvaughn-architecture-demo/components/search-files/HighlightMatch.tsx
```diff
@@ -0,0 +1,24 @@
...
+    if (index < pieces.length - 1) {
+      children.push(
+        <span className={styles.Match} key={`${index}-needle`}>
+          {needle}
+        </span>
+      );
+    }
+  }
+
+  return children as any;
```
hbenl, line 23: Nit: avoid `any` by using `return <>{children}</>;`

packages/bvaughn-architecture-demo/src/suspense/SearchCache.ts
```diff
@@ -0,0 +1,180 @@
...
+    } else {
+      return 1;
+    }
+  };
+
+  sources.forEach(source => {
+    const sourceId = source.sourceId;
+
+    if (client.isPrettyPrintedSource(sourceId)) {
+      return;
```
hbenl, line 162: This means we're using minified sources instead of their pretty-printed versions but I think we should do it the other way around.

hbenl (commented):
When I click on a search result, the source is opened in the currently selected source tab (instead of creating a new source tab or switching to an already open tab for that source) and if there is no source open, clicking on a search result does nothing.

packages/bvaughn-architecture-demo/components/search-files/HighlightMatch.tsx
bvaughn: This is a goofy aspect of TypeScript IMO. Casting to `any` isn't great, but returning `children as any` is _slightly more performant_ since it avoids creating an unnecessary Fiber.

packages/bvaughn-architecture-demo/src/suspense/SearchCache.ts
bvaughn: Hmm, this is what the old implementation did:
```js
async function initializeSourceIds(client: ReplayClientInterface) {
  sourceIdsWithNodeModules = [];
  sourceIdsWithoutNodeModules = [];

  const sources = await getSourcesAsync(client);

  // Insert sources in order so that original sources are first.
  const compareSources = (a: SourceId, b: SourceId) => {
    const aIsOriginal = client.isOriginalSource(a);
    const bIsOriginal = client.isOriginalSource(b);
    if (aIsOriginal === bIsOriginal) {
      return 0;
    } else if (aIsOriginal) {
      return -1;
    } else {
      return 1;
    }
  };

  sources.forEach(source => {
    const sourceId = source.sourceId;

    if (client.isPrettyPrintedSource(sourceId)) {
      return;
    }
```

bvaughn: I agree with you that it seems like using the pretty-printed one would be better if there are both. Will we _always have_ a pretty-printed version?

bvaughn:
> When I click on a search result, the source is opened in the currently selected source tab (instead of creating a new source tab or switching to an already open tab for that source) and if there is no source open, clicking on a search result does nothing.

Huh. You're right. Let me look into what broke there. Sorry.

bvaughn:
Ah, that's because I was testing with at least one source open. That's because we were only rendering the `NewSourceAdapter` (which does the React context <-> Redux mapping) if there was at least one source. Lemme fix.

bvaughn:
Fixed in 2a7ddb0. Thanks for catching that over sight.

packages/bvaughn-architecture-demo/components/search-files/HighlightMatch.tsx
hbenl: I think that's rather a goofy aspect of the types shipped by React, I wish they would just "admit" that a render function is allowed to return `ReactNode[]`.

bvaughn: Agreed that should be the return type. I don't know the legacy reasons for why it is what it is.

FWIW though the React team doesn't ship TypeScript types. They're community maintained ([over at DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)).

packages/bvaughn-architecture-demo/src/suspense/SearchCache.ts
hbenl: No, we don't always have a pretty-printed version, only if the file is minified (i.e. if it contains a line with more than 500 characters). So we'd have to check if there is a pretty-printed version for a non-pretty-printed file before skipping it.

bvaughn: How would we do that?

hbenl (approved):
LGTM except for the search using minified instead of pretty-printed sources.

packages/bvaughn-architecture-demo/src/suspense/SearchCache.ts
hbenl: Check if there is a pretty-printed source that has the current source as its "generated" source ("generated" according to the backend, in the case of pretty-printing the backend declares the minified source as the generated one, even though in reality it's the other way around).

bvaughn: I filed FE-983 for this follow up.

PR #8184, opened 2022-11-21
ReplayClient should not make requests in unloaded regions

bvaughn:
This is not fool proof, since the backend could unload a region while a request is in-flight, but in general this commit hardens the ReplayClient to check that the execution point or focus range that it is about to make a request for has been loaded before proceeding.

### DIFF
--- a/packages/shared/utils/time.test.ts
+++ b/packages/shared/utils/time.test.ts
@@ -4,44 +4,81 @@ import { isRangeInRegions } from "./time";
 
 describe("time util", () => {
   describe("isRangeInRegions", () => {
-    function toTimeStampedPointRange(...tuples: ExecutionPoint[]): TimeStampedPointRange[] {
-      const range: TimeStampedPointRange[] = [];
-      for (let i = 0; i < tuples.length; i += 2) {
-        const beginPoint = tuples[i];
-        const endPoint = tuples[i + 1];
-        range.push({
-          begin: {
-            point: beginPoint,
-            time: parseInt(beginPoint, 10),
-          },
-          end: {
-            point: endPoint,
-            time: parseInt(endPoint, 10),
-          },
-        });
+    function createRange(
+      beginPoint: ExecutionPoint,
+      beginTime: number,
+      endPoint: ExecutionPoint,
+      endTime: number
+    ): TimeStampedPointRange {
+      return {
+        begin: {
+          point: beginPoint,
+          time: beginTime,
+        },
+        end: {
+          point: endPoint,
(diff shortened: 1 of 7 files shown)


markerikson (approved):

jcmorrow:
I'm still trying to grok all of this. But reading the gymnastics that we are doing here makes me inclined to just change every protocol request to have a `begin` and `end` on it (in points!!!) and just make it the backend's problem whether or not things need to be loaded to process that request. I don't actually know how hard this would be, but I sort of doubt that we can ever have full confidence in the result of protocol requests with the current API.

jcmorrow:
Also, I am fully on board about the backend checking whether the requested ranges actually match the currently loaded regions. That was a good idea @bvaughn.

packages/shared/utils/time.test.ts
```diff
@@ -4,44 +4,81 @@ import { isRangeInRegions } from "./time";
...
 
     it("should reject non-contiguous overlaps", () => {
-      expect(isRangeInRegions("2", "5", toTimeStampedPointRange("0", "3", "4", "8"))).toBe(false);
+      expect(isRangeInRegions(tupleToRange("2", "5"), tuplesToRanges("0", "3", "4", "8"))).toBe(
+        false
+      );
+    });
+
+    // Imitate scenario where in-memory time-to-points mapping is too coarse
+    // to correctly differentiate between different time ranges.
```
jcmorrow, line 74: I think I'm having a hard time envisioning this case 🤔

bvaughn: I left a comment this morning (on one of your Replays IIR) where this seemed to be happening.

bvaughn:
> Also, I am fully on board about the backend checking whether the requested ranges actually match the currently loaded regions. That was a good idea @bvaughn.
> ether or not things need to be loaded to process that request. I don't actually know how hard this would be, but I sort of doubt that we can ever have full confidence in the result of protocol requests with the current API.

I think there was pushback from someone on the backend team (don't remember) when I mentioned this a few days ago? The tricky case of "things unloaded _after_ we asked about them_" seems pretty darn hard to avoid if the logic is on the frontend alone, though.

bvaughn:
I guess I'm going to push another small change to this branch so that– if the Console input's _pause_ create request fails, we'll show this:
Screen Shot 2022-11-21 at 2 45 47 PM

Rather than showing this:
Screen Shot 2022-11-21 at 2 46 46 PM

Neither is great but one seems worse.

PR #8185, opened 2022-11-21
Re-add "editor hovered" analytics event for onboarding

markerikson:
- Re-adds the `"editor.mouse_over"` Mixpanel event used as part of our onboarding flow, which was lost during the removal of the old source editor


bvaughn (approved):
I wish we didn't have to scatter this sort of analytics code throughout the app but oh well 😁

PR #8187, opened 2022-11-21
Select the whole input content on cmd+F

Andarist:
This behavior matches VS Code and Chrome. Whenever you press CMD+F in those the search input gets shown+focused+selected.

cc @bvaughn


jasonLaster:
@Andarist makes sense to me. Will let @bvaughn or @markerikson do the review though

bvaughn (approved):
Sure

PR #8188, opened 2022-11-21
General cleanup and TS migration of tree components and shared logic

markerikson:
- Removed a bunch of useless `() => ({})` args to `connect`
- Migrated the `Tree` and `ManagedTree` components to TS
- Rewrote all remaining components that still used the legacy `react-dom-factories` package:
  - Migrated them to TS (or in some cases deleted them because they're dead)
  - Replaced all uses of `react-dom-factories` with normal JSX syntax
- Deleted some more dead files in `client/shared`


markerikson:
Eyeballed the changes on the preview:

- CSS rules / values display okay
- Accordion renders and toggles
- Sources tree renders and contents look fine

PR #8193, opened 2022-11-22
Fixed broken default import

bvaughn:
Fixed minor regression from #8188 that was breaking prod for me.

https://www.loom.com/share/e6098439cc334f659b683422c4e4c5e8

### DIFF
--- a/src/devtools/client/inspector/layout/components/LayoutApp.tsx
+++ b/src/devtools/client/inspector/layout/components/LayoutApp.tsx
@@ -5,13 +5,12 @@
 import React from "react";
 import { ConnectedProps, connect } from "react-redux";
 
+import Accordion from "devtools/client/shared/components/Accordion";
+import { prefs as servicePrefs } from "devtools/shared/services";
 import type { UIState } from "ui/state";
 
 import { BoxModel } from "../../boxmodel/components/BoxModel";
 
-const Services = require("devtools/shared/services");
-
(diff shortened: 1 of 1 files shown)


src/devtools/client/inspector/layout/components/LayoutApp.tsx
```diff
@@ -5,13 +5,12 @@
...
 
+import Accordion from "devtools/client/shared/components/Accordion";
+import { prefs as servicePrefs } from "devtools/shared/services";
 import type { UIState } from "ui/state";
 
 import { BoxModel } from "../../boxmodel/components/BoxModel";
 
-const Services = require("devtools/shared/services");
-
-const Accordion = require("devtools/client/shared/components/Accordion");
```
bvaughn, line 14: This was the bug. `Accordion` is a _default export_ but we don't have `.default` on the end, so we're trying to render _the module itself_.

I assume it's okay to just use `import` statements here instead of `require`? Dunno why we were doing the other to begin with.

hbenl: IIRC this was necessary to import javascript (instead of typescript) modules, perhaps because we didn't have `allowJs` set in our tsconfig. It doesn't seem to be necessary anymore.

hbenl (approved):

bvaughn:
That's good to know!

I checked and it didn't seem to break anything, but I was curious.

PR #8195, opened 2022-11-22
Sync focus and loaded regions between server and client

bvaughn:
Refine the in-memory focus region if the backend decides to load a _smaller_ region due to region boundaries.

---

### Background

I think this ties back to the "soft focus" efforts from #7114. My understanding of _why_ we're sending the "soft focus" time rather than the `TimeStampedPoint` time is that we think the backend can find an execution point that more closely matches the time than the client can. (The client may only have paint points, so the data might be pretty sparse.)

Unfortunately this can also break things in the way that's noted in [Replay fe7a5a5d-d483-4634-9d93-ae4ce03bdc80](https://app.replay.io/recording/query-never-happens--fe7a5a5d-d483-4634-9d93-ae4ce03bdc80). Essentially the scenario shown below:
Screen Shot 2022-11-22 at 3 28 40 PM

After trying to load the focus range shown above, the frontend might get stuck in a situation where it's waiting for an execution point that's _before_ the nearest region boundary (and will never finish loading).

### DIFF
--- a/src/ui/components/FocusContextReduxAdapter.tsx
+++ b/src/ui/components/FocusContextReduxAdapter.tsx
@@ -1,7 +1,17 @@
-import { PropsWithChildren, useCallback, useEffect, useMemo, useState, useTransition } from "react";
+import isEqual from "lodash/isEqual";
+import {
+  PropsWithChildren,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  useTransition,
+} from "react";
 
 import { FocusContext } from "bvaughn-architecture-demo/src/contexts/FocusContext";
 import { Range } from "bvaughn-architecture-demo/src/types";
+import { isPointInRegions, isRangeInRegions } from "shared/utils/time";
 import { setFocusRegion } from "ui/actions/timeline";
 import { getLoadedRegions } from "ui/reducers/app";
 import { getFocusRegion } from "ui/reducers/timeline";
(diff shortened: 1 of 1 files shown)


src/ui/components/FocusContextReduxAdapter.tsx
```diff
@@ -12,17 +23,61 @@ import { rangeForFocusRegion } from "ui/utils/timeline";
...
   useEffect(() => {
-    startTransition(() => {
-      setDeferredFocusRegion(focusRegion);
-    });
-  }, [focusRegion, loadedRegions]);
+    let newFocusRegion = focusRegion;
+
+    let loading = loadedRegions?.loading || null;
+    let loadingBeginTime = loading?.[0]?.begin.time ?? null;
+    let loadingEndTime = loading?.[0]?.end.time ?? null;
```
bvaughn, line 43: There will only ever be 1 `loading` range, right? (Wouldn't make sense to break this up, right?)

jcmorrow: Yep, that is my understanding as well!

jcmorrow (commented):
I'm about to hop into a meeting, but there's still one thing I don't totally understand:

- If we have requested a focus window load via times
- AND we are waiting for a point *whose time is within the times we requested*
- The backend should *always* load that point, despite the coarseness of times, because of monotonicity and the fact that we always load a little *more* than the request window, never less.
- Does that reflect the cases that this is handling? Or is there something I'm missing? I think if the top 3 points are not holding, that's a backend bug.

src/ui/components/FocusContextReduxAdapter.tsx
bvaughn: Good to confirm.

bvaughn:
@jcmorrow Check out the comments on [fe7a5a5d-d483-4634-9d93-ae4ce03bdc80](https://app.replay.io/recording/query-never-happens--fe7a5a5d-d483-4634-9d93-ae4ce03bdc80) for the case I'm trying to guard against:
1. The frontend has a focus range (by points and time) locally
2. It asks the backend to load the range (by time)
3. It waits for the range to finish loading (by points) before it fetches e.g. hit counts
4. The range that eventually loads matches (by time) but not by points
5. The frontend is stuck waiting

jcmorrow:
> It waits for the range to finish loading (by points) before it fetches e.g. hit counts

I don't even understand how this is possible 😆 If we *know* the range we want via points, I would have expected that we had issued the `loadRegion` request for the times attached to those points

bvaughn:
Just noticed something I missed earlier, because I'm not as familiar with the `FocusRegion` structure.

jcmorrow:
I just tested https://github.com/replayio/backend/pull/6634 in preprod and it *does* fix this case (at least for [Replay fe7a5a5d-d483-4634-9d93-ae4ce03bdc80](https://app.replay.io/recording/query-never-happens--fe7a5a5d-d483-4634-9d93-ae4ce03bdc80)), and I am pretty darn sure it solves it in general too, but we'll have to wait and confirm.

bvaughn:
@jcmorrow You're saying we don't need this change at all, after that one's been made?

jcmorrow:
> @jcmorrow You're saying we don't need this change at all, after that one's been made?

I am pretty sure we don't. The important part is: it's now impossible for the protocol to return a point which is in another region, because:

A) We only consider points from inside of the region that you are querying (this was actually always true, but sometimes we would error)
B) We will not error for any point in the recording

jcmorrow:
I've also started to pull out some other ugliness in https://github.com/replayio/devtools/pull/8204, and there might be other things that can also be simplified because of this change?

bvaughn:
Correct me if I'm wrong but I _think_ we would need a _different_ change instead, to avoid caching a possibly fuzzy/incorrect guestimate point?

bvaughn:
I don't think this change is necessary any more after https://github.com/replayio/backend/pull/6634 and https://github.com/replayio/devtools/pull/8204. I was concerned about the front-end _caching_ imprecise values, but we are not currently caching these lookups. The Redux action always asks the backend for the time on-seek:
https://github.com/replayio/devtools/blob/25f0b7733d1ab8d29eb23d27f574f04983d134c4/src/ui/actions/timeline.ts#L250-L273

We probably _do_ need to make a change to the `ReplayClient` so that it no longer waits for loaded regions before calling `Session.getPointNearTime` but I'll do that in a follow up PR:
https://github.com/replayio/devtools/blob/25f0b7733d1ab8d29eb23d27f574f04983d134c4/packages/shared/client/ReplayClient.ts#L536-L543

PR #8199, opened 2022-11-23
Rewrite sources tree to display separate nodes for multiple versions of a file

markerikson:
- Fixed a broken lookup for sources that caused file tree nodes to not have icons (the record is keyed by URL, not source ID)
- Cleaned up some `$FixTypeLater` types and used the proper `TreeNode` types
- Rewrote source tree construction logic to better handle multiple versions of the same file:
  - added a new `"multiSource"` tree node type that acts as an expandable parent to the actual source entries
  - Rewrote checks for node types to handle this case
  - Improved React key selection for tree node components to use URLs+IDs for source entries instead of path
  - Added checks for existing files by content hash to prevent showing full duplicates
  - Multiple-version tree nodes are prefixed with a number to distinguish them, like `(1) App.tsx`
  - Ensured "find file node" logic works for these duplicates, allowing "Reveal in Tree" and highlighting based on selected source tab to work right
- Tweaked tree node CSS to shrink whitespace between icon and text from 10px to 3px per Jon's direction

Note that the source tabs do not show the same `(N) ` prefix, because that's internal to the source tree construction logic for now.

To see an example of this behavior, open up https://devtools-git-feature-fe-162-source-tree-duplicates-recordreplay.vercel.app/recording/vite-hmr-multi-file-test-with-duplicates--3ed2de41-1d87-4b53-8c7f-e1016d48547f on the preview branch.  I edited both `src/App.tsx` and `src/main.tsx` several times, including flipping back and forth between identical contents.  There should be 8 total versions of each file, but only 4 will show up in the tree because of the duplication.  (Currently, the "Quick Open" dialog will still show all 8 versions of each file.)

Visual:

image

### DIFF
--- a/src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.tsx
+++ b/src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.tsx
@@ -19,22 +19,21 @@ import { getContext } from "../../selectors";
 import { copyToTheClipboard } from "../../utils/clipboard";
 import { getSourceQueryString } from "../../utils/source";
 import { getPathWithoutThread, isDirectory } from "../../utils/sources-tree";
+import { TreeNode } from "../../utils/sources-tree/types";
 import AccessibleImage from "../shared/AccessibleImage";
 import SourceIcon from "../shared/SourceIcon";
 
-type $FixTypeLater = any;
-
 interface STIProps {
-  item: $FixTypeLater;
+  item: TreeNode;
   depth: number;
   focused: boolean;
   autoExpand: boolean;
   expanded: boolean;
-  focusItem: (item: $FixTypeLater) => void;
-  selectItem: (item: $FixTypeLater) => void;
+  focusItem: (item: TreeNode) => void;
+  selectItem: (item: TreeNode) => void;
   source: SourceDetails;
   debuggeeUrl: string;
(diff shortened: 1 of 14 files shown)


jonbell-lot23 (approved):

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTreeItem.tsx
```diff
@@ -174,7 +174,7 @@ class SourceTreeItem extends Component<FinalSTIProps> {
   renderItemTooltip() {
     const { item, depth } = this.props;
 
-    return item.type === "source" ? unescape(item.contents.url) : getPathWithoutThread(item.path);
+    return item.type === "source" ? unescape(item.contents.url!) : getPathWithoutThread(item.path);
```
jasonLaster, line 177: I am pretty sure that console evals can be a source and not have a url

markerikson: I can tweak it, but this is the same logic that was there already. Only thing that changed is that `SourceTreeItem` now actually knows the right type for `TreeNode` and `SourceDetails`, instead of being `type $FixTypeLater = any`.

jasonLaster: come to think of it, we probably dont add sources without urls to the tree so it'll be caught higher up

PR #8202, opened 2022-11-24
Link to the correct location for test steps

jaril:
This:
- links to a more accurate location by working up from a step's enqueue event to find the corresponding caller. That means we have a working link to the step's source, whether that function call was in a spec file or not.
- fixes the play from here and jump to step actions so it lands/starts in the right place

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
@@ -77,7 +77,7 @@ export function TestSteps({
     dispatch(startPlayback({ beginTime: testStart, endTime: testEnd - 1 }));
   };
   const onPlayFromHere = (beginTime: number) => {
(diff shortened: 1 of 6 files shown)


src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -184,6 +181,29 @@ function TestStepItem({
   const onJumpToAfter = () => {
     dispatch(seekToTime(startTime + adjustedDuration - 1));
   };
+  const onGoToLocation = async () => {
+    const { point } = annotations[index];
+
+    const frame = await (async point => {
```
ryanjduffy, line 187: Assuming this is Good Enough for now but noting that we don't need to wrap this with async and this `app.sendMessage` should be replaced with either the underlying `sendMessage` command from the protocol or use the new ReplayClient class.

ryanjduffy (approved):

PR #8206, opened 2022-11-28
Use inspector component for console props

ryanjduffy:
I'd like to get rid of the prototype entry but this'll be a lot better than building a custom inspector for test suites.

image

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
@@ -1,15 +1,22 @@
+import { Object as ProtocolObject } from "@replayio/protocol";
+import cloneDeep from "lodash/cloneDeep";
 import React, { useContext, useState } from "react";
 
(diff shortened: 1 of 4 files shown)


ryanjduffy:
@jaril - this is targeting your open branch for console props

src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
```diff
@@ -95,8 +95,21 @@ export function TestStepItem({
...
+            consolePropsProperty.object,
+            pauseResult.pauseId
+          );
+          const consoleProps = consolePropsPauseData.objects?.find(
+            o => o.objectId === consolePropsProperty.object
+          );
+
+          if (consoleProps?.preview) {
+            // suppress the prototype entry in the properties output
+            consoleProps.preview.prototypeId = undefined;
```
markerikson, line 108: :hand: I think this is technically mutating the cache entry. Probably better to make a defensive nested copy here.

ryanjduffy: It's probably safe for this object in this pause point but agree we should be defensive here

PR #8208, opened 2022-11-29
Fix the logic for matching annotations to steps

jaril:
We were using indices before, which was unreliable.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useMemo, useState } from "react";
 
 import { startPlayback } from "ui/actions/timeline";
 import Icon from "ui/components/shared/Icon";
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -29,26 +29,26 @@ function useGetTestSections(
           acc[0].push({
             ...step,
             annotations: {
```
markerikson, line 31: :hand: Three thoughts here:

- This is a good candidate for a `useMemo()`
- Personally I'd like to see it as an object with named fields like `beforeEach` instead of `acc[0]` :)
- We probably don't have _tons_ of annotations, so `O(n)` runtime likely isn't a big deal, but we could consider normalizing each of those annotations arrays into a lookup table keyed by ID to avoid the repeated `.find() calls

ryanjduffy (approved):

ryanjduffy:
I pushed a commit that added `useMemo` and switched the reduce to use objects since we're returning that structure anyway. 🚀

PR #8209, opened 2022-11-29
Move console props to the footer

jaril:
Fix SCS-355.


src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -22,33 +23,35 @@ function useGetTestSections(
   const annotationsEnqueue = useAppSelector(getReporterAnnotationsForTitle(testTitle));
   const annotationsEnd = useAppSelector(getReporterAnnotationsForTitleEnd(testTitle));
 
+  console.log({ annotationsEnqueue, annotationsEnd, steps });
```
ryanjduffy, line 26: 🪵

src/ui/components/SidePanel.module.css
```diff
@@ -0,0 +1,15 @@
...
+  font-size: 15px;
+  line-height: 19px;
+  font-weight: 400;
+  color: var(--body-color);
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  padding: 8px;
+  border-bottom: 1px;
+  border-color: var(--theme-border);
```
ryanjduffy, line 14: Let's add `align-items: center` here so the spec title is aligned

ryanjduffy (approved):
This is better so I'm 👍  on it.

I think the change to make the info panel focused on the test is a good direction. The UI is feeling a bit cluttered around here (duplicate spec from the replay title and test header, two back buttons with different destinations, extra indenting levels when focused on a test) so that's a possible next step for us.

image

PR #8210, opened 2022-11-29
Show network requests in line with the test sidebar

jaril:
Fix SCS-331, Fix SCS-332, Fix SCS-317, Fix SCS-324, Fix SCS-351

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/NetworkEvent.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/NetworkEvent.tsx
@@ -1,0 +1,59 @@
+import React from "react";
+
+import { setSelectedPanel, setViewMode } from "ui/actions/layout";
+import { selectAndFetchRequest } from "ui/actions/network";
+import { RequestSummary } from "ui/components/NetworkMonitor/utils";
+import { useAppDispatch } from "ui/setup/hooks";
+import { AnnotatedTestStep } from "ui/types";
+
+import { XHR_TYPE } from "./TestSteps";
+
+export function NetworkEvent({
+  method,
+  status,
+  url,
+  id,
+}: {
+  method: string;
+  status?: number;
+  url: string;
+  id: string;
+}) {
+  const dispatch = useAppDispatch();
+  const onClick = () => {
+    dispatch(setViewMode("dev"));
+    dispatch(setSelectedPanel("network"));
+    dispatch(selectAndFetchRequest(id));
+  };
+
+  return (
+    <button className="font-italic flex border-b border-themeBase-90 p-1 px-2" onClick={onClick}>
+      {method} {status} {new URL(url).pathname}
+    </button>
+  );
+}
+export const getDisplayedEvents = (
+  step: AnnotatedTestStep,
+  steps: AnnotatedTestStep[],
+  data: RequestSummary[],
+  startTime: number
+) => {
+  return data.filter(r => {
+    if (!r.end) {
+      return false;
+    }
+
+    const isDuringStep = (time: number, start: number, end: number) => time >= start && time < end;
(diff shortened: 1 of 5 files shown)


src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
-import { AnnotatedTestStep, Annotation, TestItem, TestStep } from "ui/types";
+import { AnnotatedTestStep, TestItem, TestStep } from "ui/types";
 
+import { NetworkEvent, getDisplayedEvents } from "./NetworkEvent";
 import { TestInfoContext } from "./TestInfo";
 import { TestStepItem } from "./TestStepItem";
 
+// const XHR_TYPE = "xhr";
+export const XHR_TYPE = "text/jsx";
```
ryanjduffy, line 19: Is this intentional? seems like an odd value

src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -155,26 +169,33 @@ function TestSection({
...
+            isLastStep={steps.length - 1 === i}
+            onReplay={onReplay}
+            onPlayFromHere={() => onPlayFromHere(startTime + s.relativeStartTime)}
+            selectedIndex={selectedIndex}
+            setSelectedIndex={setSelectedIndex}
+            id={s.id}
+          />
+          <div className="flex flex-col">
+            {getDisplayedEvents(s, steps, data, startTime).map((r, i) => (
+              <NetworkEvent key={i} method={r.method} status={r.status} url={r.url} id={r.id} />
```
ryanjduffy, line 195: `key`-ing on `r.id` is probably better but likely doesn't change much here

src/devtools/client/debugger/src/components/TestInfo/NetworkEvent.tsx
```diff
@@ -0,0 +1,59 @@
...
+  data: RequestSummary[],
+  startTime: number
+) => {
+  return data.filter(r => {
+    if (!r.end) {
+      return false;
+    }
+
+    const isDuringStep = (time: number, start: number, end: number) => time >= start && time < end;
+    const applicableSteps = steps.filter(s =>
```
ryanjduffy, line 47: We should pre-calc the start/end times for each step so we don't have to pass the whole steps array around to do things like this

src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -155,26 +169,33 @@ function TestSection({
     <>
       {header ? <div className="py-2">{header}</div> : null}
       {steps.map((s, i) => (
```
ryanjduffy, line 171: Another way to approach this would be to merge and sort steps and events in one array and then render each in one array pass

ryanjduffy (approved):

src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
ryanjduffy: @jaril - This is my only real question. The other comments are mostly editorial and can be deferred.

jaril: Bandaid for now since we seem to be getting incorrect types for network events. Kannan's on it, and when it's fixed we can just switch to the commented line.

jaril: https://linear.app/replay/issue/RUN-867/chromium-netmonitor-fix-xhr-event-attributes

PR #8233, opened 2022-12-01
Move test step actions into a context menu

jaril:
Fix SCS-302, Fix SCS-356.

https://www.loom.com/share/ca62c04777e845a5a542d43fb141e176

This moves the actions into a context menu. And makes it so that assertions are considered as 0 duration events.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/ContextMenu.module.css
+++ b/src/devtools/client/debugger/src/components/TestInfo/ContextMenu.module.css
@@ -5,10 +5,9 @@
   border-radius: 0.5em;
   overflow: hidden;
   white-space: nowrap;
-  transform: translateX(-100%);
 
   /* Ensure the context menu is always above console rows and the current time indicator */
-  z-index: 4;
+  z-index: 11;
 }
 
 .ContextMenuItem {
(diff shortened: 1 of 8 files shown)


src/devtools/client/debugger/src/components/TestInfo/ContextMenu.module.css
```diff
@@ -0,0 +1,41 @@
+.ContextMenu {
+  position: absolute;
+  flex-direction: column;
+  display: flex;
+  border-radius: 0.5em;
```
ryanjduffy, line 5: Verifying we're using `em` intentionally throughout this since it'll scale in potentially unexpected ways with the font size change.

src/devtools/client/debugger/src/components/TestInfo/ContextMenu.tsx
```diff
@@ -0,0 +1,119 @@
...
+    hide();
+    dispatch(seekToTime(testStep.endTime - 1));
+  };
+  const onGoToLocation = async () => {
+    hide();
+    if (!testStep.enqueuePoint) {
+      return;
+    }
+
+    const frame = await (async point => {
```
ryanjduffy, line 55: This doesn't need to be wrapped in another async iife because we're already in an async fn

ryanjduffy (approved):

src/devtools/client/debugger/src/components/TestInfo/ContextMenu.module.css
jaril: It wasn't intentional, since I copied it from [here](https://github.com/replayio/devtools/blob/main/packages/bvaughn-architecture-demo/components/console/ContextMenu.module.css).

My instinct is to just change it to `rem` — thoughts?

PR #8238, opened 2022-12-01
Find test source locations from non-spec files

ryanjduffy:


src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
```diff
@@ -13,8 +13,11 @@ import { TestInfoContext } from "./TestInfo";
 import { TestInfoContextMenuContext } from "./TestInfoContextMenuContext";
 import { TestStepContext } from "./TestStepRoot";
 
-function returnFirst<T, R>(list: T[] | undefined, fn: (v: T) => R | null) {
-  return list ? list.reduce<R | null>((acc, v) => acc ?? fn(v), null) : null;
+export function returnFirst<T, R>(
```
markerikson, line 16: :question: why is this needed?  Could this just be `frames.find()` instead?

ryanjduffy: It probably isn't any more. I added it to facilitate walking deeper into object trees with multiple levels of arrays. It'll probably be removed when we refactor the test suite code soonish.

PR #8241, opened 2022-12-01
Disable step context menu buttons if they're not applicable

jaril:
Fix SCS-384.


src/devtools/client/debugger/src/components/TestInfo/ContextMenu.tsx
```diff
@@ -85,11 +87,21 @@ function ContextMenu({
         <MaterialIcon>play_circle</MaterialIcon>
         Play from here
       </div>
-      <div className={styles.ContextMenuItem} onClick={onJumpToBefore}>
+      <div
+        className={`${styles.ContextMenuItem} ${
+          testStep.startTime === currentTime ? styles.disabled : ""
+        }`}
+        onClick={onJumpToBefore}
```
ryanjduffy, line 94: We should handle disabled here as well probably by conditionally applying the handler. I know we have `pointer-events: none` on the node but I think it'd still navigate on `Enter` if the element gained focus. Bit edge casey but easy to fix.

ryanjduffy (approved):

PR #8246, opened 2022-12-02
Load console messages earlier

jcmorrow:
and use `indexed` rather than `loading` for checking whether we should consider a part of the recording ready to interact with.

Loom explainer: https://www.loom.com/share/096286a1cda645918ae2fd8c79e34d20

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -241,11 +241,25 @@ export class ReplayClient implements ReplayClientInterface {
   }> {
     const sessionId = this.getSessionIdThrows();
 
-    // Don't try to fetch messages in unloaded regions.
-    // The result might be invalid (and may get cached by a Suspense caller).
-    await this._waitForRangeToBeLoaded(focusRange);
-
     if (focusRange !== null) {
+      // We *only* care about loaded regions when calling `findMessagesInRange`.
+      // Calling `findMessages` is always safe and always returns the console
+      // messages for all parts of the recording, regardless of what is
+      // currently loading or loaded. The reason we sometimes use
+      // `findMessagesInRange` is because `findMessages` can overflow (if the
+      // replay contains more than 1,000 console messages). In that case, we
+      // might be able to fetch all of the console messages for a particular
+      // section by using `findMessagesInRange`, but it requires sending
+      // manifests to a running process, so it will only work in loaded regions.
+
+      // It would be better if `findMessagesInRange` either errored when the
+      // requested range could not be returned, or returned the boundaries of
+      // what it *did* successfully load (see BAC-2536), but right now it will
+      // just silently return a subset of messages. Given that we are extra
+      // careful here not to to fetch messages in unloaded regions because the
+      // result might be invalid (and may get cached by a Suspense caller).
+      await this._waitForRangeToBeLoaded(focusRange);
+
       const response = await client.Console.findMessagesInRange(
         { range: { begin: focusRange.begin.point, end: focusRange.end.point } },
         sessionId
@@ -980,13 +994,11 @@ export class ReplayClient implements ReplayClientInterface {
       const checkLoaded = () => {
         const loadedRegions = this.loadedRegions;
         let isLoaded = false;
-        if (loadedRegions !== null) {
+        if (loadedRegions !== null && loadedRegions.loading.length > 0) {
           if (focusRange !== null) {
-            isLoaded = isRangeInRegions(focusRange, loadedRegions.loaded);
+            isLoaded = isRangeInRegions(focusRange, loadedRegions.indexed);
           } else {
-            isLoaded =
-              loadedRegions.loaded.length > 0 &&
-              areRangesEqual(loadedRegions.loaded, loadedRegions.loading);
+            isLoaded = areRangesEqual(loadedRegions.indexed, loadedRegions.loading);
           }
         }
 


packages/shared/client/ReplayClient.ts
```diff
@@ -981,12 +981,10 @@ export class ReplayClient implements ReplayClientInterface {
         const loadedRegions = this.loadedRegions;
         let isLoaded = false;
         if (loadedRegions !== null) {
-          if (focusRange !== null) {
-            isLoaded = isRangeInRegions(focusRange, loadedRegions.loaded);
+          if (focusRange === null) {
```
jcmorrow, line 984: I switched the order of these branches, it felt like if we are going to have an `else` clause anyways, then removing negatives slightly reduces cognitive load while reading, but I don't feel strongly about it.

packages/shared/client/ReplayClient.ts
```diff
@@ -981,12 +981,10 @@ export class ReplayClient implements ReplayClientInterface {
...
+            isLoaded = areRangesEqual(loadedRegions.indexed, loadedRegions.loading);
           } else {
-            isLoaded =
-              loadedRegions.loaded.length > 0 &&
```
jcmorrow, line 988: I removed the `length > 0` check, because AFAIK `loading` can never be an empty array (so if `loaded` is empty they can never be equal), but again, I don't feel super strongly about it. The important part of the change is `loaded` -> `indexed`.

hbenl: I'd prefer to keep that check because we shouldn't rely on undocumented details of the protocol:
- since they're undocumented anyone will feel free to change them in the backend and then the frontend may break
- it should be possible to fully understand the frontend code without insider knowledge that even most replay employees (including myself) don't have

hbenl (approved):
I *guess* this is fine!?
But the explanations in the Loom sound very handwavy. The loaded regions stuff is pretty hard to deal with for the frontend and having only a vague idea of what to expect from the backend in some situations doesn't help. We need stronger guarantees from the backend and we need them to be documented. Just my 2 cents.

packages/shared/client/ReplayClient.ts
```diff
@@ -241,11 +241,11 @@ export class ReplayClient implements ReplayClientInterface {
   }> {
     const sessionId = this.getSessionIdThrows();
 
-    // Don't try to fetch messages in unloaded regions.
-    // The result might be invalid (and may get cached by a Suspense caller).
-    await this._waitForRangeToBeLoaded(focusRange);
```
bvaughn, line 246: Why was this okay to do?

If there is no focus range, and we try to fetch messages for the whole recording, we might end up asking about unloaded regions still, right?

`_waitForRangeToBeLoaded` is smart enough to handle this case.

https://github.com/replayio/devtools/blob/fcae274fba899c2ec5db8fbf12f0fe393400ab53/packages/shared/client/ReplayClient.ts#L976-L989

packages/shared/client/ReplayClient.ts
bvaughn: I don't really feel strongly about one vs another– but this feels pretty subjective to me. I'd prefer we don't make changes like this often, because it seems like it would be a lot of churn for a subjective prefernce.

packages/shared/client/ReplayClient.ts
bvaughn: I agree with Holger.

bvaughn (commented):
Not sure if this change is safe.

packages/shared/client/ReplayClient.ts
jcmorrow: Ah, sorry, I mentioned this in the linked loom explainer, but I did not write it in the commit message. We have two different methods for fetching console messages: `findMessages` and `findMessagesInRange`. Now, I've gone on this rant probably a dozen times at this point, but the reason that this *one* basic data resource has multiple methods for fetching exactly the same kind of information is because it has an `overflow` flag, so discovering console messages in a particular region might allow you to bypass that flag. Unfortunately, they have different requirements because of how they work.

`findMessages` is much closer to all of the other basic data resources: it does not interact with loaded or indexed regions at all. It's always available, and always returns all data for the whole recording, regardless of what is currently loaded. For instance, we do not wait for the whole recording to be loaded before calling `findSources`, nor do we expect that `findSources` will only return sources that were used or first downloaded in the currently loaded regions.

Because `findMessagesInRange` is designed to get around the overflow flag, it has to go directly to the running linker and ask it to run through particular part of the recording and re-collect the console messages. This requires a running linker in those regions, and so it requires that those regions be loaded. I think we should either make this endpoint return the actual window for which it ran with the results, or have it error if it can't run through the requested window, rather than what it does now (silently return a different window) but that's a slightly tangential topic.

Anyways, because of the way `findMessages` works, we can issue that request as soon as the Replay is loaded, and in *most* cases we will never need to request again, because there will be fewer than 1,000 console messages.

bvaughn: Thank you for elaborating! That was very helpful. Could we add some abbreviated form of that as a comment inline for future me (or others) who might see this and think it's a mistake?

packages/shared/client/ReplayClient.ts
jcmorrow: I agree about the diff noise created by things that are just based on preference. I really do think that the `!=` operator causes more cognitive load than the `==` operator, but I don't have proof that this is the case.

FWIW, I can find some people on the internet who agree:

- https://csharpdeveloper.wordpress.com/2016/01/20/roberts-rules-of-coders-8-avoid-negative-conditionals-and-method-names/
- https://schneide.blog/2014/08/03/dont-ever-not-avoid-negative-logic/

But if that were a convincing argument we would have larger problems so...

packages/shared/client/ReplayClient.ts
jcmorrow: Alright, fair enough.

jcmorrow: I've slightly adjusted this logic. It sounds like the correct interpretation is "if we are not currently trying to load anything, then we cannot possibly be loaded yet". If I'm interpreting that correctly, I feel like the easiest way to signal that is to put it in the top level conditional, rather than `&&`ing it together in one of the branches but not the other!

packages/shared/client/ReplayClient.ts
jcmorrow: Yeah sounds great, will do.

jcmorrow: I guess I'll add one other thing while I'm in here: it's really not the focus window that should be deciding which of these we use, but rather we should probably always try `findMessages` first and only in the case of overflow fall back to `findMessagesInRange` as the focus window changes. Otherwise, users who load a replay with a specific window encoded in the URL will still end up waiting for the regions to load before fetching the console messages (even though in most cases they could issue a single `findMessages` and be done). But that's a larger change I think.

packages/shared/client/ReplayClient.ts
bvaughn: > FWIW, I can find some people on the internet who agree

lol 😆

packages/shared/client/ReplayClient.ts
bvaughn: I think it's more complicated than the above paragraph makes it sound– because in a bad case, we'd end up fetching all messages, seeing overflow, and then re-fetching a slice of messages. Adding the extra logic for that doesn't seem worth it to me.

bvaughn (approved):
This looks good, I think

packages/shared/client/ReplayClient.ts
jcmorrow: Oh, funny, I 100% agree when the truthiness check is used, and I disagree when the explicit `!== null` or `!= null` is used. For me it's two things that make it difficult: `undefined !== null` evaluates to `true`, so even though TS is helping, I still hold that possibility in my head just in case we bungled up our types somewhere, whereas `=== null` or `== null` alleviates that (but so does `if (thing)`. The rubyist in me really wants `if (thing)`, but I know that's not anyone's cup of tea. Anyways, I'll happily cede to you hear, thanks for explaining your thought process!

bvaughn: Unfortunately `!= null` and `== null` are generally better options, because `if (thing)` is often _not going to give the result you want_ because of "falsey" values like 0 and `""` (empty string)

jcmorrow:
Tried to minimize the the total diff to not make too unnecessary/controversial changes and squashed (and added my lengthy explanation to the commit message). Just gonna wait for CI to double check that things look good and then I'll merge.

packages/shared/client/ReplayClient.ts
markerikson: Heh. And on my side, I've internalized "never use the loose single-`=` comparison operators, ever" :)  so seeing Brian frequently use `!=` has been kinda jarring.

bvaughn: Curious why you've internalized that?

My reasoning for it is mostly just that... it's easy for `undefined` to slip into the chain unintentionally (e.g. `map.get(key)`) and _in most cases_ `null` and `undefined` have similar semantic meaning (unlike other "falsey" values).

jcmorrow:
> I guess this is fine!?
> But the explanations in the Loom sound very handwavy.

I am pretty sure I have a solid handle on what makes this change an improvement, but I might have done a bad job of explaining it in the loom. If after reading the written responses you still feel like it's unclear or handwavy, maybe I can clarify some bits? 

>  The loaded regions stuff is pretty hard to deal with for the frontend and having only a vague idea of what to expect from the backend in some situations doesn't help. We need stronger guarantees from the backend and we need them to be documented. Just my 2 cents.

Agreed. I'm hesitant to put this level of detail in the protocol documentation itself, especially the bits that say super explicit internal things like "this requires sending a manifest to a linker", because that just feels like too much internal detail and is too tied to a specific implementation. I made https://linear.app/replay/issue/BAC-2536/consolefindmessagesinrange-silently-returns-incomplete-results to document what I think would be improvements of this API particularly, feel free to add on if you have related things you'd like to see!

PR #8247, opened 2022-12-02
Fetch repaint for test start annotation

ryanjduffy:
Fetches repaints for test start annotations so we have the correct video frame when pausing.

* Adds a new method, `repaintAtPause`, which will repaint at a specific point and save that paint to the playback paint array. This involved some refactoring of the existing `repaint` method but the behavior should be the same for that method.
* Changes the jump to before/after context menu options to use the start/end annotations when the exist
* Limits asserts to only "jump to after"
* Compare step "past" and "paused" conditions to use points when possible instead of times


src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
```diff
@@ -44,96 +49,112 @@ export function TestStepItem({ argString, index, id }: TestStepItemProps) {
...
-  }, [client, messageEnd, pointEnd, setConsoleProps, setPauseId]);
+
+    // return async () => {
+    //   if (endPauseResult) {
+    //     await client.releasePause(endPauseResult.pauseId);
+    //   }
+    //   if (startPauseResult) {
+    //     await client.releasePause(startPauseResult.pauseId);
+    //   }
+    // }
```
ryanjduffy, line 140: There's a `Session.releasePause` command but it isn't supported on the client yet. Should it be? Should we be releasing pauses?

src/ui/actions/timeline.ts
```diff
@@ -191,6 +192,20 @@ export function setTimelineToTime(time: number | null, updateGraphics = true): U
...
+export function setTimelineToPauseTime(
+  pauseId: string,
+  time: number | null,
+  updateGraphics = true
+): UIThunkAction {
+  return async dispatch => {
+    dispatch(setTimelineToTime(time, updateGraphics));
+
+    if (time) {
+      dispatch(() => repaintAtPause(time, pauseId));
```
ryanjduffy, line 204: @markerikson - this might be a dumb way to add a one-off dispatch. I'll claim ignorance because I haven't used redux. I _think_ I want it to be scheduled by the dispatcher so I know it's called after the `setTimelineToTime` but maybe this doesn't guarantee that. Advice welcome!

packages/protocol/graphics.ts
```diff
@@ -274,7 +283,8 @@ export async function repaint(force = false) {
     }
   }
 
-  if (!rv || pauseId !== ThreadFront.currentPause.pauseId) {
+  // if (!rv || pauseId !== ThreadFront.currentPause.pauseId) {
+  if (!rv) {
```
ryanjduffy, line 287: @hbenl - Looks like you originally added this condition in #4396. In this refactor, I want to explicitly allow fetching and repainting the screen outside of the current pause. However, I can also rework this method (`repaintAtPause`) to not be used by the original `repaint` method so that we can retain this condition. I wasn't sure why it was there so I wanted to ask.

ryanjduffy: @hbenl - Lemme know if you recall why we had this condition originally. Thanks!

hbenl: The condition is there to avoid painting a screenshot for a pause that the user has already switched away from.
I think the `repaintAtPause()` function should not paint the screenshot that it fetches - we should always show the screenshot for the current pause. Instead it should only fetch it (and be renamed to `fetchRepaintForPause()` or something) and the `paintGraphics()` call should be moved to the `repaint()` method together with the `pauseId !== ThreadFront.currentPause.pauseId` condition.

ryanjduffy: We need the ability to paint the screenshot for a different pause -- that's the feature we're implementing here to preview a test step without jumping to that point -- but your explanation for that condition makes sense so I'll refactor with that in mind.

src/ui/actions/timeline.ts
ryanjduffy: I think this was bad but needed to take a different approach anyway.

src/devtools/client/debugger/src/components/TestInfo/ContextMenu.tsx
```diff
@@ -113,9 +125,9 @@ function ContextMenu({
       </div>
       <div
         className={classnames("ContextMenuItem", {
-          disabled: testStep.absoluteStartTime === currentTime,
+          disabled: canJumpToBefore,
         })}
-        onClick={testStep.absoluteStartTime === currentTime ? undefined : onJumpToBefore}
+        onClick={canJumpToBefore ? undefined : onJumpToBefore}
```
jaril, line 130: While we're at it, we should probably remove jumping before/after in an assert. Since the way we've been framing it, assert is a single point in time.

src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
```diff
@@ -29,107 +32,135 @@ export interface TestStepItemProps {
...
+          });
         }
       } catch {
         setLocalPauseData(undefined);
       }
 
       return null;
     })();
-  }, [client, messageEnd, pointEnd, setConsoleProps, setPauseId]);
+
```
jaril, line 146: Remove the comment?

jaril (approved):
Left some notes, lint's failing but otherwise looks good!

PR #8248, opened 2022-12-02
Fix ReplayClient's event emitter implementation (FE-1005)

hbenl:


bvaughn (approved):
Was this causing a bug? It would be nice to know the context for this change.

Seems like a good one though.

hbenl:
> Was this causing a bug? 

Yes, it caused FE-1005 and that may have caused the console getting stuck loading.

jcmorrow:
Ohhhhhh right. Good catch @hbenl! @jasonLaster and I were super stuck on this last night! Definitely having watchpoints would be super awesome here 💖

PR #8249, opened 2022-12-02
Console input UX fixes

bvaughn:
- [x] Type-ahead popup no longer flickers before dismissing an empty list (or a single-item list with an exact match)
- [x] Up/down arrow restores previous commands in Console
- [x] Fixed small UX regression with vertical alignment for truncated strings in Console's eager-eval row
- [x] Updated screenshots


bvaughn:
Happy to respond to feedback for this issue retroactively. Since no one on FE is online at the moment though, and these changes are small, I'm going to optimistically land them.

PR #8252, opened 2022-12-02
Rewrite per-recording persistence logic to be in IndexedDB

markerikson:
- Implemented a `useIndexedDB` hook that can store values in `IndexedDB`
- Updated all `bvaughn`-folder functionality that stored values per-recording in `localStorage` to use `useIndexedDB` instead
- Changed "console filter open" flag to be per-user instead of per-recording
- Implemented migration logic to convert existing per-recording persisted values in `localStorage` to be in IDB instead
- Updated persisted location value to be per-recording in the `replaySessions` field
- Disabled broken "hit counts mode" option
- Disabled broken "copy points to clipboard" command feature
- Removed incorrect types for `enableLargeText`


packages/bvaughn-architecture-demo/components/console/hooks/useTerminalHistory.ts
```diff
@@ -1,16 +1,19 @@
...
   maxHistoryLength: number = 20
 ): [expressionHistory: string[], addExpression: (value: string) => void] {
-  const [expressionHistory, setExpressionHistory] = useLocalStorage<string[]>(
-    `${recordingId}::expressionHistory`,
-    []
-  );
+  const { value: expressionHistory, setValue: setExpressionHistory } = useIndexedDB<string[]>({
+    database: CONSOLE_SETTINGS_DATABASE,
+    initialValue: [],
+    recordName: recordingId,
```
bvaughn, line 14: Why is this named "recording name"? Seems like we use "recording id" everywhere (even GraphQL). Seems simpler / less confusing to call it what it is.

packages/bvaughn-architecture-demo/src/contexts/ConsoleFiltersContext.tsx
```diff
@@ -59,18 +58,30 @@ export type ConsoleFiltersContextType = Toggles & {
   ) => void;
 };
 
+// NOTE: If any change is made like adding a store name, bump the version number
+// to ensure that the database is recreated properly.
+export const CONSOLE_SETTINGS_DATABASE: IDBOptions = {
```
bvaughn, line 63: nit Since this isn't filter-specific but console-wide, maybe this constant should live somewhere else? Don't feel strongly about it.

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
```diff
@@ -2,22 +2,31 @@ import { Location, TimeStampedPoint } from "@replayio/protocol";
...
 
 import useBreakpointIdsFromServer from "../hooks/useBreakpointIdsFromServer";
-import useLocalStorage from "../hooks/useLocalStorage";
+import useIndexedDB, { IDBOptions } from "../hooks/useIndexedDB";
 import { SessionContext } from "./SessionContext";
 
+const EMPTY_ARRAY: Point[] = [];
+
+// NOTE: If any change is made like adding a store name, bump the version number
+// to ensure that the database is recreated properly.
```
bvaughn, line 23: Will re-creating reset values to their defaults?

Is there guidance for when a new key should be added to an existing database vs a new database? (Seems like the migrate vs recreate question would impact this)

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
```diff
@@ -52,26 +61,51 @@ export function PointsContextRoot({ children }: PropsWithChildren<{}>) {
...
+      setPoints(newPoints);
+      setPointsForAnalysis(newPoints);
     },
     [setPoints, setPointsForAnalysis]
   );
 
+  useLayoutEffect(() => {
+    if (pointsStatus === "initialization-complete") {
+      prevPointsRef.current = points!;
+    }
```
bvaughn, line 103: This seems like it could burn us. If we try to add a point before previous ones load (even though it's probably unlikely, timing wise) we'd read an empty array and override the previous points value.

I think we should just prevent the UI from being enabled until the status was complete, and then...restore the `SetStateAction` logic (no ref based thing)

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
```diff
@@ -151,7 +152,7 @@ function UiPreferences() {
           />
         </div>
       </div>
-      {hitCountsEnabled && (
+      {/* {hitCountsEnabled && (
```
bvaughn, line 155: Couldn't we just wire this setting up to the new one index db store? Not sure I understand why we're _removing_ this vs just connecting the two

src/ui/actions/app.ts
```diff
@@ -209,6 +209,8 @@ export function loadMouseTargets(): UIThunkAction {
 
 export function setBreakpointsFromClipboard(): UIThunkAction {
   return async (dispatch, getState, { ThreadFront }) => {
+    // TODO [FE-998] Figure out a new implementation that works with IndexedDB instead
```
bvaughn, line 212: Seems like maybe we could accomplish this (and simplify the other code?) in two parts:
1. Add index db initialization to the "basic info" loading step that blocks the rest of the app from loading.
2. Copy values into React state (so we can read/write them sync, in-memory) and just async update the DB.

Seems like that would enable us to undo a few of the awkward ref-copying, `[] | undefined` types logic in other places? am I missing something?

bvaughn (commented):
Some initial thoughts. I think my brain is too tired to finish the review atm.

packages/bvaughn-architecture-demo/components/console/hooks/useTerminalHistory.ts
markerikson: It's `recordName`, as in 'the record in the database".  You wrote that variable name in your original WIP commit :)

packages/bvaughn-architecture-demo/src/contexts/ConsoleFiltersContext.tsx
markerikson: Yeah, I wasn't sure whether to put this in its own file or not - it's related to the rest of the console logic, but wasn't strictly a "component", so I figured this was the closest place to put it. I can move it around.

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
markerikson: I briefly tried modifying the database definition to add a third "store", and it looked like it got added without modifying the first couple.  Based on what I've seen, what matters is the logic you run in the `upgrade()` handler, and right now the hook is implemented to just "make sure there's a store in existence for each string name", no deleting or anything.

Haven't thought further on grouping settings, other than the per-recording values we had prior to this PR seemed to fall into two buckets: "points"-related and "console"-related.

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
markerikson: This was also left over from your initial WIP commit :)  

I can review it and see if it can be improved.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
markerikson: This is a totally separate persistence problem.

The issue here is that we used to store a "hit counts mode" enum (on/color-bar-only/show-numbers), in `localStorage`, using the legacy FF prefs system.

When you rewrote the source viewer, it got switched to a single boolean, in `localStorage`, via the `useLocalStorage` hook.

That broke this checkbox.  It still toggles the FF-based prefs value, but the source viewer code never looks at that value.

So, this checkbox is broken, and I'm hiding it from the UI until we get around to making it work again.

src/ui/actions/app.ts
markerikson: This line here is also separate from the "React+IDB" question.

My first attempt at the "copy points to clipboard / paste points from clipboard" functionality relied on the fact that points were synced via `localStorage`, which has the `"storage"` event that fires whenever a value is set.

IndexedDB doesn't have any such event.  There's no way to know when it's been changed from somewhere else.

The "paste points from clipboard" feature actually didn't entirely work 100% right, so it's been sorta broken ever since I added it.

The change from storing points in `localStorage` to IDB further means that this implementation is not a viable approach.

So, I'm also turning this off until it can be rethought.

packages/bvaughn-architecture-demo/components/console/hooks/useTerminalHistory.ts
bvaughn: Oh wow I completely misread "record" as "recording" 😆  Sorry!

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
bvaughn: Just to be clear, the initial WIP commit was very half baked. I didn't finish writing it. I definitely didn't test it. Please feel free to question _anything_ from it.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
bvaughn: Sure sure.  I was just asking if we couldn't keep this check box around, but wire it up to the new `localStorage` boolean instead. If that's hard for some reason bc of how the prefs settings dialog is wired up, that's cool. I was just wondering.

src/ui/actions/app.ts
bvaughn: Sounds fair.

bvaughn (approved):
I slept poorly and don't have the mental capacity this morning to give this a proper review.

I trust you, and I know that what was here before was a scattered mess, so I'm sure this is a nice incremental improvement even if we end up doing something else. If you feel good about this, land it.

If you'd rather wait on a deeper review, I can do it later today (or we could talk through it, or Holger could do it– whichever is more convenient)

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
markerikson: Tried logging some timings here.  In local dev, it's about 100ms from the time the `useIndexedDB` hooks first render to when we have read the data and set state...

and that looks like it's well faster than the time it takes to fetch the initial source file even if it was open on recording load.

Timings are always fickle, but I'm okay with not worrying about code for preventing interactions before the points have been read from IDB.

src/ui/components/shared/UserSettingsModal/PreferencesSettings.tsx
markerikson: Mmmm... couple thoughts. One is that I mis-spoke - it's not actually a "checkbox", it's a dropdown for the enum.

We _could_ theoretically change this to target `useLocalStorage`, but I also don't know if the source line rendering logic is set up to allow the 3 separate options (no hit display, color bar only, show numbers).  So, we'd either have to switch this to be a checkbox, or update the rendering logic to match prior 3-mode behavior.  Either way I'd rather do that in a follow-up PR, and there's an open issue for it.

packages/bvaughn-architecture-demo/src/contexts/PointsContext.tsx
bvaughn: I am pretty uncomfortable with that assumption.

bvaughn: Maybe I'm misunderstanding the implications of what you're saying though. Let's chat about it after the sync?

bvaughn:
This PR broke the architecture demo app (which I use for local development and testing/iteration). Log points (at least) no longer work– can't add them and they don't sync over from the main app.

bvaughn:
I think this introduced a regression in FE-1033

PR #8253, opened 2022-12-02
Comment @mentions include Workspace members and collaborators

bvaughn:
- [x] `CommentEditor` to pass collaborator and workspace member names to mentions plug-in
- [x] Add new query for fetching collaborator and workspace member names
- [x] Fix edge-case issue with parsing legacy Tip Tap comment JSON format

---

Once this PR lands, I'll follow up with a separate one to fix a (pre-existing) duplicate query name and update the other GraphQL types.

### DIFF
--- a/src/ui/hooks/useGetCollaboratorNames.ts
+++ b/src/ui/hooks/useGetCollaboratorNames.ts
@@ -1,0 +1,114 @@
+import { gql, useQuery } from "@apollo/client";
+import { useContext, useMemo } from "react";
+
+import { SessionContext } from "bvaughn-architecture-demo/src/contexts/SessionContext";
+import { GetCollaboratorNames, GetCollaboratorNamesVariables } from "graphql/GetCollaboratorNames";
+import { useGetRecordingId } from "ui/hooks/recordings";
+
+export type PartialUser = {
+  id: string;
+  name: string;
+};
+export type PartialUsers = PartialUser[];
+
+const EMPTY_ARRAY: PartialUsers = [];
+
+// Returns a unified list of collaborators and workspace members with access to the current recording.
+export default function useRecordingUsers(excludeCurrentUser: boolean): PartialUser[] {
+  const { currentUserInfo } = useContext(SessionContext);
+
+  const recordingId = useGetRecordingId();
+  const queryData = useQuery<GetCollaboratorNames, GetCollaboratorNamesVariables>(QUERY, {
+    variables: { recordingId },
+  });
+
+  const users = useMemo<PartialUsers>(() => {
+    const owner = queryData.data?.recording?.owner;
+    const collaboratorEdges = queryData.data?.recording?.collaborators?.edges;
+    const workspaceEdges = queryData.data?.recording?.workspace?.members?.edges;
+
+    const partialUsers: Map<string, PartialUser> = new Map();
+    const currentUserId = currentUserInfo!.id;
+
+    const addPartialUser = (partialUser: PartialUser) => {
+      if (!excludeCurrentUser || partialUser.id !== currentUserId) {
+        partialUsers.set(partialUser.id, partialUser);
+      }
+    };
+
+    if (owner) {
+      const { id, name } = owner;
+      if (name !== null) {
+        addPartialUser({ id, name });
+      }
+    }
+
+    if (collaboratorEdges) {
+      collaboratorEdges.forEach(collaboratorEdge => {
+        if (collaboratorEdge.node.__typename === "RecordingUserCollaborator") {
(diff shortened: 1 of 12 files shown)


src/ui/hooks/workspaces_user.ts
```diff
@@ -23,50 +23,52 @@ import {
...
-                    __typename
-                    id
-                    roles
-                    user {
-                      id
-                      name
-                      picture
-                    }
-                  }
+export const GET_WORKSPACE_MEMBERS = gql`
```
bvaughn, line 26: Had to extract this GQL so I could use it with the imperative Apollo API too, because hooks can't be called conditionally and both `recording` and `recording.workspace` are nullable 😞

src/ui/hooks/useGetCollaboratorNames.ts
```diff
@@ -0,0 +1,114 @@
...
+          if (node.user.name) {
+            newUsers.push({
+              id: node.user.id,
+              name: node.user.name,
+            });
+          }
+        }
+      });
+
+      setWorkspaceUsers(newUsers.length > 0 ? newUsers : EMPTY_ARRAY);
```
bvaughn, line 68: These types are so fucking frustrating

src/ui/hooks/workspaces_user.ts
jasonLaster: wow - this gql looks super verbose. Also where are the types? I thought they were built?

bvaughn: The types are here:
https://github.com/replayio/devtools/blob/main/src/graphql/GetWorkspaceMembers.ts

I found them really confusing and difficult to work with.

jasonLaster: oh yeah - these are auto generated

src/ui/hooks/useGetCollaboratorNames.ts
jasonLaster: agreed

src/ui/hooks/workspaces_user.ts
bvaughn: Apparently we could improve the generated output by changing something on the backend/server but I'm not super familiar with GraphQL.

src/ui/hooks/useGetCollaboratorNames.ts
```diff
@@ -0,0 +1,114 @@
...
+export type PartialUsers = PartialUser[];
+
+const EMPTY_ARRAY: PartialUsers = [];
+
+// Returns a unified list of collaborators and workspace members with access to the current recording.
+export default function useRecordingUsers(excludeCurrentUser: boolean): PartialUser[] {
+  const { currentUserInfo } = useContext(SessionContext);
+
+  // This info is loaded by a nested component because we can't conditionally load it otherwise.
+  const [workspaceUsers, setWorkspaceUsers] = useState<PartialUsers>(EMPTY_ARRAY);
```
hbenl, line 26: This would not be necessary if you use a query that starts with the `recordingId` instead of the `workspaceId`. For example you could fetch all the collaborator names with a single query like this:
```graphql
query GetCollaboratorNames($recordingId: UUID!) {
  recording(uuid: $recordingId) {
    owner {
      id
      name
    }
    collaborators {
      edges {
        node {
          ... on RecordingUserCollaborator {
            user {
              id
              name
            }
          }
        }
      }
    }
    workspace {
      members {
        edges {
          node {
            ... on WorkspaceUserMember {
              user {
                id
                name
              }
            }
          }
        }
      }
    }
  }
}
```
(I left out the `RecordingPendingUserCollaborator`s here even though they are included by your PR - do we want to show them or leave them out like the `WorkspacePendingUserMember`s?)

hbenl (approved):
LGTM (although it could be simplified)

src/ui/hooks/useGetCollaboratorNames.ts
bvaughn: > (I left out the `RecordingPendingUserCollaborator`s here even though they are included by your PR - do we want to show them or leave them out like the `WorkspacePendingUserMember`s?)

The main reason I left them out was that they didn't have _names_ (so nothing to filter on).

hbenl: `WorkspacePendingUserMember` does have a `name` (although it's optional, which I think it shouldn't be), only `WorkspacePendingEmailMember` doesn't have a name.
But I guess we should leave out all pending collaborators and members since they haven't agreed yet to be a collaborator/member.

hbenl (approved):

PR #8255, opened 2022-12-02
Highlight spec file line after opening

ryanjduffy:
Hijacks `setCurrentSearchResultLocation` to highlight the current step in the source viewer


bvaughn:
I'm looking into FE-1009. Let's hold off on this PR if that's okay! :)

ryanjduffy:
> I'm looking into [FE-1009](https://linear.app/replay/issue/FE-1009). Let's hold off on this PR if that's okay! :)

That's fine! This was never supposed to be the solution :)

bvaughn:
#8265

PR #8265, opened 2022-12-05
Refactor selected/focused location

bvaughn:
[Overview demo](https://www.loom.com/share/7cdb998e84a746ed8ef4f57476b30b02)

- [x] Combine `focusedLineIndex` and `focusedSourceId` into `focusedSource` and add `focusedType` (`"search-result" | "view-source"`). I don't think it made sense for us to track these separately.
- [x] Remove `currentSearchResultLocation` and setter in favor of `focusedSource`
- [x] Add start/end line range support to `focusedSource` (in prep for future Test Suites functionality)
- [x] Improve syncing of focused/selected state between React context and Redux to avoid potential cycles.

cc @ryanjduffy, @jaril

### DIFF
--- a/src/devtools/client/debugger/src/reducers/tabs.ts
+++ b/src/devtools/client/debugger/src/reducers/tabs.ts
@@ -55,7 +55,12 @@ function update(state: TabsState = initialTabState(), action: AnyAction) {
       return updateSourceIds(state, action.payload);
 
     case locationSelected.type: {
-      return addSelectedSource(state, action.payload.source);
+      const source = action.payload.source;
+      if (source) {
+        return addSelectedSource(state, source);
+      } else {
+        return state;
+      }
     }
 
     default:
--- a/src/devtools/client/debugger/src/reducers/tabs.ts
+++ b/src/devtools/client/debugger/src/reducers/tabs.ts
@@ -55,12 +55,7 @@ function update(state: TabsState = initialTabState(), action: AnyAction) {
       return updateSourceIds(state, action.payload);
 
     case locationSelected.type: {
-      const source = action.payload.source;
-      if (source) {
-        return addSelectedSource(state, source);
-      } else {
-        return state;
-      }
+      return addSelectedSource(state, action.payload.source);
     }
 
     default:
--- a/packages/bvaughn-architecture-demo/components/search-files/ResultsListRow.tsx
+++ b/packages/bvaughn-architecture-demo/components/search-files/ResultsListRow.tsx
@@ -130,7 +130,7 @@ function MatchRow({
 }) {
   const { match } = result;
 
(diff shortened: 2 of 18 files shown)


src/devtools/client/debugger/src/reducers/tabs.ts
```diff
@@ -55,7 +55,12 @@ function update(state: TabsState = initialTabState(), action: AnyAction) {
       return updateSourceIds(state, action.payload);
 
     case locationSelected.type: {
-      return addSelectedSource(state, action.payload.source);
+      const source = action.payload.source;
+      if (source) {
+        return addSelectedSource(state, source);
+      } else {
+        return state;
+      }
```
bvaughn, line 63: Pointing this change out to @markerikson in case it seems bad.

Alternately, maybe we could change the `SourcesContextAdapter` to read/find the corresponding `SourceDetails` thing (that the bvaughn architecture code doesn't know anything about)

markerikson: There's nothing wrong with the check in and of itself, but at least as the code existed prior to this PR, the TS types say that we shouldn't ever be in a case where `source` doesn't exist.  The only place that dispatched `locationSelected` always ensured `source` is included.

Looking at this PR, I see that we're now also dispatching `locationSelected({location})`, and _not_ including `source`.

Is that specifically what we want?  Based on what I know of the codebase atm, the intent is to always go through the `selectLocation` thunk that also ensures we have a source, fetches symbols, etc, rather than the `locationSelected` plain action by itself.

I _think_ it makes sense to have `SourcesContextAdapter` do `dispatch(selectLocation(location))` instead.  Does that work the way you want?

jasonLaster:
@bvaughn looks great!

src/devtools/client/debugger/src/reducers/tabs.ts
bvaughn: > I think it makes sense to have SourcesContextAdapter do dispatch(selectLocation(location)) instead. Does that work the way you want?

I think it would have to dispatch context too, whatever that is.

bvaughn: I think I made the change you were asking for

markerikson: Yep! Change looks good to me code-wise, just a question of if the final UI behavior is what you're trying to make it do.

(the `context` or `cx` variable is, loosely, the current pause info, and there's some checks in the logic to bail out of certain steps if it looks like we've changed to a new pause while in the middle of a multi-step operation)

bvaughn:
Playwright test failure was expected because of a regression in another PR that recently landed. That has been fixed separately.

PR #8270, opened 2022-12-05
Clear the selected step whenever we seek somewhere in the timeline

jaril:
Fix SCS-436.


jaril:
This isn't quite working for playback, the HUD still shows until you hit pause. Will revisit later this afternoon.

ryanjduffy (approved):

PR #8283, opened 2022-12-06
Fix selectLocation() to not open a source unless requested (FE-1010)

hbenl:
Dispatching the `locationSelected` redux action is enough to open a source and is not necessary if we don't want to open it.


bvaughn:
TBH it's pretty confusing that we have "select location" and that it sometimes calls "location selected" but sometimes not.

PR #8284, opened 2022-12-06
Skip framesteps that point to the beginning of a function body (SCS-172)

hbenl:
This is a minimal experiment for SCS-172: the FrameTimeline ignores steps that point to the beginning of a function body (but the stepping buttons don't).
This fixes the example in SCS-172 but we still need to check if we're ignoring too many steps in some places.

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
@@ -12,12 +12,18 @@ import ReactTooltip from "react-tooltip";
 import { getFrameStepsSuspense } from "bvaughn-architecture-demo/src/suspense/FrameStepsCache";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import { actions } from "ui/actions";
-import { SourcesState, getPreferredLocation, getSelectedLocation } from "ui/reducers/sources";
+import {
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
```diff
@@ -92,7 +99,23 @@ class FrameTimelineRenderer extends Component<FrameTimelineProps, FrameTimelineS
...
+    let adjustedDisplayIndex = Math.min(displayIndex, numberOfPositions - 1);
+
+    while (adjustedDisplayIndex < numberOfPositions - 2) {
+      const location = frameSteps[adjustedDisplayIndex].frame?.find(
+        location => location.sourceId === selectedLocation?.sourceId
+      );
+      if (
+        symbols?.symbols?.functionBodyLocations.some(
+          bodyLocation =>
+            bodyLocation.line === location?.line && bodyLocation.column === location.column
```
jasonLaster, line 111: nit: we should have an `isSameLocation` helper somewhere

jasonLaster (commented):
It looks like this could be a bit more general:

We want to 
1. show only the valid steps in the timeline
2. let you scrub through valid steps
3. only step forward/back to valid points 

We probably should doing the validation at a higher level as a result.

PR #8285, opened 2022-12-06
Convert new architecture styles from EMs to REMs

bvaughn:
`em` (or `ch`) units are relative to the font size of the element being styled, and since we use a couple of different font sizes in our app– using either means that e.g. `padding` won't necessarily be the same between two elements (which may have different font sizes). `rem` units avoid this problem since there's only one root font size.

This PR makes a rather large change of replacing all `em` units with `rem` and updating existing styles to keep things from changing _too much_. This should provide a more solid starting point for us _and hopefully avoid scenarios where we need to use one-off sizes_ like `0.35em` or `0.9em`.

- [x] Replace `em` styles with `rem`
- [x] Update screenshots
- [x] Add [Style Lint rule](https://stylelint.io/user-guide/rules/unit-disallowed-list/) to prevent `em` units from being reintroduced

Trunk (and Style Lint) is happy if there are no `em` units in the bvaughn folder:
Screen Shot 2022-12-06 at 1 36 51 PM

Trunk complains if there are:
Screen Shot 2022-12-06 at 1 38 27 PM

cc @jasonLaster, @jonbell-lot23 

Specifically the issue raised in FE-1030 is also fixed by this:
Screen Shot 2022-12-06 at 4 09 21 PM


bvaughn:
The `tests/repaint.test` e2e test failed on CI. Don't see how that would be related to these changes.

PR #8286, opened 2022-12-06
Re-implement "hit counts" preference

markerikson:
- Removed dead "experimental setting" option
- Updated `useLocalStorage` hook to force updates in the same tab
- Replaced tri-state "hide/color only/show numbers" hit counts pref with a boolean that is either "color only/show numbers".

Confirmed that this neatly toggles the behavior in the source viewer as soon as you click the checkbox.

### DIFF
--- a/packages/bvaughn-architecture-demo/src/hooks/useLocalStorage.ts
+++ b/packages/bvaughn-architecture-demo/src/hooks/useLocalStorage.ts
@@ -1,6 +1,20 @@
-import { SetStateAction, useCallback, useLayoutEffect, useState, useTransition } from "react";
+import {
+  SetStateAction,
+  useCallback,
+  useLayoutEffect,
+  useReducer,
+  useRef,
+  useState,
+  useTransition,
+} from "react";
 
 import { localStorageGetItem, localStorageSetItem } from "../utils/storage";
+
+function manuallyTriggerStorageEvent() {
+  // Apparently the "storage" event only fires across tabs, and setting a value
+  // in the _same_ tab won't trigger it. Do that manually:
+  window.dispatchEvent(new Event("storage"));
+}
 
 // Stores value in localStorage and synchronizes it between sessions and tabs.
 // The API mirrors useState.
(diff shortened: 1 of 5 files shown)


packages/bvaughn-architecture-demo/src/hooks/useLocalStorage.ts
```diff
@@ -63,5 +80,14 @@ export default function useLocalStorage<T>(
...
   }, [key, value]);
 
+  // Order of effects matters here. Don't trigger this until _after_ we've updated `localStorage`.
+  useLayoutEffect(() => {
+    // Only notify if this hook has had its setter run
+    if (updateCounter > 0) {
+      // Notify any other components subscribed to this value that it has changed
+      manuallyTriggerStorageEvent();
+    }
+  }, [updateCounter]);
```
bvaughn, line 90: Can we just move this into the same hook as updates local storage?

https://github.com/replayio/devtools/blob/5a30eb174e7f8c4a39a6ee5031f817d4a44ebda4/packages/bvaughn-architecture-demo/src/hooks/useLocalStorage.ts#L76-L81

Basically
```js
localStorageSetItem(key, string);

manuallyTriggerStorageEvent();
```

bvaughn (commented):
Looks fine but I think you could simplify the effects stuff by combining them.

packages/bvaughn-architecture-demo/src/hooks/useLocalStorage.ts
markerikson: Maybe. I was trying to avoid any potential stringification render loop cases, as well as ensure that it only runs specifically once per attempt to set a value and _only_ in the component that had its setter called.

bvaughn: I'm pretty sure what I suggested still meets those criteria, doesn't it?

bvaughn: I dunno. Whatever. Either way is fine I guess.

PR #8289, opened 2022-12-06
Preload `useIndexedDB` values

markerikson:
- Added Suspense caches to `useIndexedDB` to cache the DB instance itself for a given database, and the _initial_ value read for a given store + record.  
- Updated `bootstrapApp()` to know about all of our IDB database definitions, and preload the _initial_ value for each store during the setup process so that we can read it synchronously later
- Updated `useIndexedDB` to synchronously read that initial value if available so that we avoid flashes during rendering (such as applying console filters)

### DIFF
--- a/packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
+++ b/packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
@@ -20,6 +20,27 @@ type Status =
   | typeof STATUS_INITIALIZATION_ERRORED
   | typeof STATUS_UPDATE_PENDING;
 
+export async function preloadIDBInitialValues(
+  idbPrefsDatabase: IDBOptions[],
+  recordingId?: string
+) {
+  const idbPrefsPromises: Promise<any>[] = [];
+  if (!recordingId) {
+    // We don't have a recording ID if it's the library
+    return;
+  }
+
+  if (recordingId) {
+    for (let dbOptions of idbPrefsDatabase) {
+      for (let storeName of dbOptions.storeNames) {
+        idbPrefsPromises.push(getInitialIDBValueAsync(dbOptions, storeName, recordingId));
+      }
+    }
+  }
+
+  return Promise.all(idbPrefsPromises);
+}
+
 export const {
   getValueSuspense: getIDBInstanceSuspense,
   getValueAsync: getIDBInstanceAsync,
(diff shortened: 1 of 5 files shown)


packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
```diff
@@ -30,20 +70,28 @@ export default function useIndexedDB<T>(options: {
...
   const { databaseName, databaseVersion, storeNames } = options.database;
   const { initialValue, recordName, scheduleUpdatesAsTransitions = false, storeName } = options;
 
   const [isPending, startTransition] = useTransition();
-  const [value, setValue] = useState<T>(initialValue);
+  // If requested, suspend entirely until we've read the initial value
+  //  and can use that to seed the `useState`
+  const [value, setValue] = useState<T>(
+    options.suspend === true ? waitForValueLoadedSuspense() : initialValue
+  );
```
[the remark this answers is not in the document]
bvaughn: Shut the fuck up seriously

packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
```diff
@@ -39,11 +87,18 @@ export default function useIndexedDB<T>(options: {
   const { initialValue, recordName, scheduleUpdatesAsTransitions = false, storeName } = options;
 
   const [isPending, startTransition] = useTransition();
-  const [value, setValue] = useState<T>(initialValue);
+  // If requested, suspend entirely until we've read the initial value
+  //  and can use that to seed the `useState`
```
bvaughn, line 91: This comment doesn't seem accurate/useful anymore?

packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
```diff
@@ -39,11 +87,18 @@ export default function useIndexedDB<T>(options: {
...
+  const [value, setValue] = useState<T>(() => {
+    // In order for persistence to work properly, `getInitialIDBValueAsync` _must_ have been
+    // called in the main app, in `src/ui/setup/index.ts::bootstrapApp()`.
+    return (
+      getInitialIDBValueIfCached(options.database, storeName, recordName)?.value ?? initialValue
+    );
```
bvaughn, line 97: Should we just lean into this and _throw_ if the value hasn't been preloaded/cached? I kind of think maybe we should. It would simplify the rest of this hook a lot. (We could get rid of the `useEffect` async initialization.)

markerikson: On second thought I'm not sure we _can_ throw here, because we still have to handle the case of no value existing on first access of a given recording, and I'm not sure we can differentiate "wasn't initialized" from "first open".

bvaughn: The primary thing I'm concerned about is if we forget to pre-fetch one of these stores, it might behave differently if the IDB value loads faster than other data or not, and that might vary based on environment or recording (which could potentially cause a value to flicker in some cases and not in others).

markerikson: All right, I can try to apply that change.

bvaughn: Okay

PR #8291, opened 2022-12-06
Fix showing console props

ryanjduffy:
Variable shadowing strikes again!

Also moved mutating the console props object for prototype into the sanitizer so we don't mutate the possibly cached object ref.


jasonLaster (approved):
wow

PR #8296, opened 2022-12-07
Workaround for the source viewer flickering in the Replay browser (FE-1025)

hbenl:
This is a pretty weird workaround for a pretty weird rendering issue in the Replay browser (see FE-1025).
Don't ask me how I came up with it :laughing:

### DIFF
--- a/packages/bvaughn-architecture-demo/components/sources/SourceListRow.module.css
+++ b/packages/bvaughn-architecture-demo/components/sources/SourceListRow.module.css
@@ -53,6 +53,13 @@
   width: 1.5rem;
   height: 1.5rem;
 }
+.HoverButtonCompanion {
+  position: absolute;
+  width: 1px;
+  height: 1rem;
+  background-color: white;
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):
I cannot test this change in Replay because you cannot login to preview branches from the Replay browser, but i assume it is fine to land as is because it should not harm anything

bvaughn:
Any explanation here by chance? This looks pretty baffling.

hbenl:
> Any explanation here by chance?

I have no idea *why* this works, all I can say is that after hours of experimentation I got the feeling that the flickering is triggered by an absolutely positioned element (in this case the `HoverButton`) being added to a `SourceListRow` or becoming visible there. So I tried placing the `HoverButton` in an absolutely positioned and always (technically) visible `<div>` and it fixed the flickering. Then I realized that putting the `<div>` next to the `HoverButton` also worked.
¯\\\_(ツ)\_/¯

PR #8297, opened 2022-12-07
Fix broken replaySessions definition and improve TS types

markerikson:
- Fixed an incorrect change to the `replay-sessions` key definition in `asyncStoreHelper`
- Improved the TS types for `asyncStoreHelper` to correctly capture the type of `"Json"` fields

There was a mismatch between how `PrefsHelper` and `asyncStoreHelper` define their string lookup keys and default values.  I had made a change to the one `asyncStoreHelper` usage thinking they were the _same_, but that was incorrect, and resulted in the field being incorrectly saved in IDB.


bvaughn:
I think this broke _unit_ tests, specifically: `src/devtools/shared/async-store-helper.test.js`

PR #8303, opened 2022-12-07
Add stable/unstable query cache flags

jcmorrow:
I kind of don't like how complicated this is getting, but I *do* want the ability to toggle these things on/off separately.

Pairs with https://github.com/replayio/backend/pull/6698/

### DIFF
--- a/src/ui/types/index.ts
+++ b/src/ui/types/index.ts
@@ -19,8 +19,9 @@ export type LocalExperimentalUserSettings = {
   basicProcessingLoadingBar: boolean;
   disableScanDataCache: boolean;
   consoleFilterDrawerDefaultsToOpen: boolean;
-  enableQueryCache: boolean;
+  disableStableQueryCache: boolean;
   enableColumnBreakpoints: boolean;
+  enableUnstableQueryCache: boolean;
   profileWorkerThreads: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


jasonLaster (commented):
Absolutely. Anything that helps us test faster. My only piece of feedback is that maybe these should be advanced features since in reality we probably dont want to expose those to others.

src/ui/types/index.ts
```diff
@@ -19,8 +19,9 @@ export type LocalExperimentalUserSettings = {
   basicProcessingLoadingBar: boolean;
   disableScanDataCache: boolean;
   consoleFilterDrawerDefaultsToOpen: boolean;
-  enableQueryCache: boolean;
+  disableStableQueryCache: boolean;
   enableColumnBreakpoints: boolean;
+  enableUnstableQueryCache: boolean;
```
loganfsmyth, line 24: Stuff like this is why I hate forced property sorting hahah

jcmorrow:
> Absolutely. Anything that helps us test faster. My only piece of feedback is that maybe these should be advanced features since in reality we probably dont want to expose those to others.

The reason I did not do this is because it can sometimes be helpful when troubleshooting with a user to have them turn a feature off to see if it fixes a bug/provides a work-around.

PR #8312, opened 2022-12-08
Check server time during focus update

jcmorrow:
This is maybe not actually wise, because it can burst a *lot* of protocol messages (getPointsBoundingTime). That command is fast, but it would probably be best to debounce it somehow, if possible. I've tried it out on the recording in https://linear.app/replay/issue/FE-1041/console-not-fetching-messages and it *does* fix the bug 🤷

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -500,7 +500,7 @@ export function clearHoveredItem(): UIThunkAction {
 export function setFocusRegion(
   focusRegion: { beginTime: number; endTime: number } | null
 ): UIThunkAction {
-  return (dispatch, getState) => {
+  return async (dispatch, getState, { replayClient }) => {
     const state = getState();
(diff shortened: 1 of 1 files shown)


hbenl:
This PR triggers another bug in the focus region UI: we have two inputs for focus regions: the draggable range in the timeline and text inputs to the right of it. In this case the value in the text input is changed (by the `clearFocusRange()` test helper), then the value from the draggable range is saved to redux and sent to the backend and then the value from the text input is saved to redux (and not sent to the backend). As a result we try to run an analysis for the range from the text inputs, which is larger than the range from the draggable, and the analysis gets stuck in `ReplayClient._waitForRangeToBeLoaded()` because the range that we told the backend to load is smaller.
I have added comments to the replay of the e2e test failure.

jcmorrow:
This is superceded by @hbenl 's excellent change in 2de522bb51101c4e836a261052f69c54ade5edfd

PR #8317, opened 2022-12-09
Handle mismatch between throttled callback and server promise result

markerikson:
- Fixed issue where the throttled `onMatches` callback would end up trying to push into an array after it had already been dispatched in an action, added to the Redux state, and frozen by Immer.  That resulted in an error saying the array couldn't be mutated.

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -49,7 +49,7 @@ import { client, initSocket } from "protocol/socket";
 import { ThreadFront } from "protocol/thread";
 import { MAX_POINTS_FOR_FULL_ANALYSIS } from "protocol/thread/analysis";
 import { RecordingCapabilities } from "protocol/thread/thread";
-import { binarySearch, compareNumericStrings, defer } from "protocol/utils";
+import { binarySearch, compareNumericStrings, defer, waitForTime } from "protocol/utils";
 import { TOO_MANY_POINTS_TO_FIND } from "shared/constants";
 import { ProtocolError, isCommandError } from "shared/utils/error";
 import { isPointInRegions, isRangeInRegions, isTimeInRegions } from "shared/utils/time";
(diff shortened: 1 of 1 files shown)


packages/shared/client/ReplayClient.ts
```diff
@@ -800,6 +800,14 @@ export class ReplayClient implements ReplayClientInterface {
     } finally {
       client.Debugger.removeFunctionsMatchesListener(matchesListener);
     }
+
+    // Because the matches callback is throttled, we may still have a bit of
+    // leftover data that hasn't been handled yet, even though the server API
+    // promise has resolved. Delay-loop until that's done, so that the logic
+    // that called this method knows when it's safe to continue.
+    while (pendingMatches.length > 0) {
+      await waitForTime(10);
```
bvaughn, line 809: Ah, fuck. Nice catch. Sorry for missing this.

I don't think waiting for 10ms is as solid a work around as we could do though. Too timing sensitive. Can we do something like this?

https://github.com/replayio/devtools/blob/dab9f8bef1af9fd678eeb0b06bd351d1ef1d9ca8/packages/shared/client/ReplayClient.ts#L831-L875

markerikson: It's a `while` loop, so the intent is that we'd end up waiting repeatedly until after the throttled function expires.

We should only end up waiting at all _if_ there's data left to pass on, and the throttled function is at 100ms, so I would assume at most it would step through that loop 8-9 times before the throttled function fires one more time.

I think your snippet could work, but I think the PR also solves it unless you can see a reason why the loop would be an issue.

bvaughn: Oh geeze. I totally overlooked the `while` and just saw the `waitForTime`. My bad!

No. I don't see an issue why the `while` loop wouldn't work. Seems like there's something nice to be had for consistency between these methods but otherwise one seems as good as the other :)

bvaughn (approved):

packages/shared/client/ReplayClient.ts
markerikson: Hah, I wondered if you were missing the `while` bit :)

markerikson:
The E2E test failures all appear to be due to a backend change that Josh just reverted, so I'm going to merge.

PR #8323, opened 2022-12-10
NewSourceAdapter more reliably syncs selected location

bvaughn:
This requires taking Redux's `getSelectedLocationHasScrolled` value into account as well, so that re-selecting the same location as previously selected will still result in the Source list scrolling to the selected line.


src/devtools/client/debugger/src/components/Editor/NewSourceAdapter.tsx
```diff
@@ -65,11 +66,12 @@ function NewSourceAdapter() {
     // Sync focused state from Redux to React context,
     if (
       focusedSource?.sourceId !== location.sourceId ||
-      focusedSource?.startLineIndex !== lineIndex
+      focusedSource?.startLineIndex !== lineIndex ||
+      !locationHasScrolled
```
bvaughn, line 70: Even if the location hasn't changed, this flag indicates that we need to re-open and re-scroll the source back into view because the user has re-clicked on it.

markerikson (approved):

PR #8324, opened 2022-12-10
Fix log point panel white-space collapsing behavior in Firefox

bvaughn:
Referencing the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) for `white-space`, it seems like `"pre"` is the only value that doesn't collapse spaces _or_ wrap implicitly. It _does_ wrap explicitly though, so we need to disable that programmatically.


packages/bvaughn-architecture-demo/components/lexical/plugins/code/CodePlugin.tsx
```diff
@@ -92,6 +92,7 @@ function codeNodeTransform(node: CodeNode, editor: LexicalEditor) {
       }
 
       const textContent = currentNode.getTextContent();
+      console.log(`CodePlugin "${textContent}"`);
```
markerikson, line 95: :hand: Dead code

jasonLaster (approved):

PR #8326, opened 2022-12-11
More intuitive auto-complete for log point panel

bvaughn:
[Loom demo of behavioral change](https://www.loom.com/share/2d69c9c43a8e4a858739ec92bfa48770)

- [x] For a log point/location, find the hit nearest to the current execution point and use it for auto-complete

To make this easier, I also put some work into the array utilities methods:
- [x] Add support to `findIndex` to return the index for the closest match (if not exact)
- [x] Added unit test coverage fro `findIndex` and `findIndexString`

---

Test updates
- [x] Add test with log points inside blocks with `let`/`const` vars to verify different in-scope variables
- [x] Update API proxy snapshots (since scope data is being fetched for new pauses)

Generated new recording fixture [4074e685-5374-4f02-b4f1-35b713d07c06](https://app.replay.io/recording/4074e685-5374-4f02-b4f1-35b713d07c06) by running:
```sh
./scripts/save-examples.ts --example=log_points_and_block_scope
```


packages/bvaughn-architecture-demo/components/sources/PointPanel.tsx
```diff
@@ -49,14 +54,7 @@ export default function SourcePanelWrapper({
...
-  const graphQLClient = useContext(GraphQLClientContext);
-  const { showCommentsPanel } = useContext(InspectorContext);
-  const { editPoint } = useContext(PointsContext);
   const client = useContext(ReplayClientContext);
-  const { accessToken, recordingId, trackEvent } = useContext(SessionContext);
-  const { executionPoint: currentExecutionPoint, time: curentTime } = useContext(TimelineContext);
-
-  const [showEditBreakpointNag, dismissEditBreakpointNag] = useNag(Nag.FIRST_BREAKPOINT_EDIT);
 
   const [hitPoints, hitPointStatus] = getHitPointsForLocationSuspense(
```
bvaughn, line 59: Moved this call into its own component because I noticed that we were calling a method that potentially _suspended_ before calling hooks.

packages/bvaughn-architecture-demo/src/utils/array.ts
```diff
@@ -1,45 +1,97 @@
+import BigInteger from "big-integer";
```
markerikson, line 1: :question: why do we need a library for these?

packages/bvaughn-architecture-demo/components/sources/PointPanel.tsx
```diff
@@ -65,6 +63,36 @@ function PointPanel({ className, point }: { className: string; point: Point }) {
     focusRange
   );
 
+  return (
+    <PointPanelWithHistPoints
```
markerikson, line 67: :hand: typo in this component name

PR #8335, opened 2022-12-13
Refactor focus region UI (FE-1040, FE-1041)

hbenl:
- replaced `focusRegion.beginTime`/`focusRegion.endTime` (the times for the `Focuser` component) by `displayedFocusRegion` to avoid the kind of mistakes we were running into
- changes to `displayedFocusRegion` will be synced to `focusRegion` using `getPointsBoundingTime()` (instead of trying to find nearby points from those collected from backend messages)
- these updates will be throttled if they come from dragging the `Focuser`
- removed the thunk that tried to "improve" the focus region begin/end points when new points were received
- added `PointsBoundingTimeCache`

### DIFF
--- a/src/ui/reducers/timeline.ts
+++ b/src/ui/reducers/timeline.ts
@@ -1,16 +1,12 @@
 import { PayloadAction, createSlice } from "@reduxjs/toolkit";
-import { TimeStampedPoint } from "@replayio/protocol";
+import { TimeRange, TimeStampedPoint } from "@replayio/protocol";
 import sortBy from "lodash/sortBy";
 
 import { UIThunkAction } from "ui/actions";
 import { MAX_FOCUS_REGION_DURATION } from "ui/actions/timeline";
 import { UIState } from "ui/state";
 import { FocusRegion, HoveredItem, TimelineState } from "ui/state/timeline";
-import {
-  displayedBeginForFocusRegion,
-  displayedEndForFocusRegion,
-  mergeSortedPointLists,
-} from "ui/utils/timeline";
+import { mergeSortedPointLists } from "ui/utils/timeline";
 
 function initialTimelineState(): TimelineState {
   return {
@@ -18,6 +14,7 @@ function initialTimelineState(): TimelineState {
     currentTime: 0,
     focusRegion: null,
     focusRegionBackup: null,
+    displayedFocusRegion: null,
     hoverTime: null,
     hoveredItem: null,
     playback: null,
@@ -59,6 +56,9 @@ const timelineSlice = createSlice({
     setFocusRegion(state, action: PayloadAction<FocusRegion | null>) {
       state.focusRegion = action.payload;
     },
+    setDisplayedFocusRegion(state, action: PayloadAction<TimeRange | null>) {
+      state.displayedFocusRegion = action.payload;
+    },
     pointsReceived(state, action: PayloadAction<TimeStampedPoint[]>) {
       const mutablePoints = [...state.points];
       state.points = mergeSortedPointLists(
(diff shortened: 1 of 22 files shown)


hbenl:
Hmm, the focus_mode-01 test passes for me locally but its flaky, I'll try to make it more robust.

src/ui/reducers/timeline.ts
```diff
@@ -92,46 +92,12 @@ export const {
   setPlaybackPrecachedTime,
   setPlaybackStalled,
   setFocusRegion,
+  setDisplayedFocusRegion,
   setTimelineState,
   pointsReceived,
   paintsReceived,
 } = timelineSlice.actions;
 
-export const pointsReceivedThunk = (points: TimeStampedPoint[]): UIThunkAction => {
```
jcmorrow, line 100: Hmmm... if we don't do this at all, then we are going to have bugs around analysis windows when points come in at times outside the displayed time window but inside the focus window points. Am I missing something there?

packages/bvaughn-architecture-demo/src/suspense/PointsBoundingTimeCache.ts
```diff
@@ -0,0 +1,14 @@
...
+import { createGenericCache2 } from "./createGenericCache";
+
+export const {
+  getValueSuspense: getPointsBoundingTimeSuspense,
+  getValueAsync: getPointsBoundingTimeAsync,
+  getValueIfCached: getPointsBoundingTimeIfCached,
+} = createGenericCache2<ReplayClientInterface, [time: number], getPointsBoundingTimeResult>(
+  async (client, time) => client.getPointsBoundingTime(time),
+  time => `${time}`
+);
```
jcmorrow, line 14: This might be out of scope for these changes, but it would be great if we could make this cache a little more sophisticated. For instance:

If you ask for the points bounding `0:02` and the points you get are `point: 123, time: 0:01` and `point: 500, time: 0:05`, then those responses could be cached such that *any* time requested between those two points is going to return that same answer. I suspect this could really cut down on the number of requests we make.

Also, it's worth pointing out that [this PR](https://github.com/replayio/backend/pull/6733/files) is ready to go out, which will remove the `precise` flag, because it turns out it is not needed.

jcmorrow (commented):
I think generally this is great, I'm not so sure about the removing of the window refining though. I don't love that code, and it would be awesome if we have found a way to elegantly sidestep that problem altogether, but I don't see how we've done that here (but maybe we did and I just missed it!).

jcmorrow:
> Hmmm... if we don't do this at all, then we are going to have bugs around analysis windows when points come in at times outside the displayed time window but inside the focus window points. Am I missing something there?

We talked offline. I was missing something 😄 . In this new iteration, we still have a time window that is not using timestamped points, but we only use that *when the user is selecting via the focuser*. This means that we will still end up loading a slightly different window than requested, but that discrepancy will only be visible when comparing the focuser component to the timeline that gets loaded. This means that the user will still have a smooth dragging experience, and we will still be able to operate on points only for all backend calls and display purposes outside of the focuser component. 

We discussed that this still *might* be a rather large discrepancy (visually, in terms of percentage of the timeline) for smaller replays, where being off by half a second (in, for example, a 10 second replay) means that the windows will be visibly different by 5% of the width of the timeline. However, I still think this is a good approach, and we can potentially fix some of those discrepancies with backend changes. For instance, we have discussed actually using stepping code to run `getPointsBoundingTime` which would produce *much* smaller tolerances than what we currently do (which is using static points like widget events). Let's ship this and evaluate how it does in prod across a range of recordings, and we can revisit in a few weeks.

Also, if after a few weeks we are happy with this, we should go forward with changing the `loadRegion` API to take timestamped points. But I don't want to make that change until we are confident that we can use this approach for the long-term.

jcmorrow (approved):
Very happy to say goodbye to that thunk! 🚢

packages/bvaughn-architecture-demo/src/suspense/PointsBoundingTimeCache.ts
```diff
@@ -0,0 +1,14 @@
+import { getPointsBoundingTimeResult } from "@replayio/protocol";
+
+import { ReplayClientInterface } from "shared/client/types";
+
+import { createGenericCache2 } from "./createGenericCache";
+
+export const {
```
bvaughn, line 7: I think this should probably go into the `PointsCache` module. That's where the call to `getPointsBoundingTime` is and this seems basically like a variant of that.

PR #8338, opened 2022-12-13
remove unloadRegion calls

miriambudayr:
Related to https://github.com/replayio/backend/pull/6739
Ticket: https://linear.app/replay/issue/BAC-2510/migrate-from-load-and-unload-region-commands-to-a-setfocuswindow

@bvaughn When Josh and I took a look at this code yesterday, we were a bit confused about `zoomTime`: https://github.com/replayio/devtools/pull/8338/files#diff-0b3b6041213ad0d8eec8fbfb5111600076ab4ab1178b839e802d76a4e8e44689L685

From what I can tell, it is pretty much always the last timestamp in the recording. Is this the correct understanding? What I have done instead is to get that timestamp from the backend.


bvaughn:
> @bvaughn When Josh and I took a look at this code yesterday, we were a bit confused about `zoomTime`

I think this is both a bit of a _holdover_ (from code/concepts that predate me, cc @jaril) and a _placeholder_ for the "zoom" feature the frontend eventually plans to (re)add. At the moment, the zoom level is always **1** (measured as 0ms through the recording duration).

src/ui/actions/timeline.ts
```diff
@@ -668,22 +668,18 @@ export function setFocusRegionBeginTime(beginTime: number, sync: boolean): UIThu
 }
 
 export function syncFocusedRegion(): UIThunkAction {
-  return async (dispatch, getState, { replayClient }) => {
+  return async (_dispatch, getState, { replayClient }) => {
```
bvaughn, line 671: nit/curiosity: why?

bvaughn (approved):
Seems fine

src/ui/actions/timeline.ts
miriambudayr: This is a what we do on the backend repo. It is a way to tag unused variables. This way, the editor (the way mine is set up) won't display the parameter with a fade:

image
vs
image

packages/bvaughn-architecture-demo/src/contexts/FocusContext.tsx
```diff
@@ -68,7 +68,7 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
     }
 
     const timeoutId = setTimeout(() => {
-      client.loadRegion({ begin: range.begin.time, end: range.end.time }, duration);
+      client.requestFocusRange({ begin: range.begin.time, end: range.end.time });
```
bvaughn, line 71: Frontend has screenshot tests that run against a pre-recorded set of API calls to minimize test flakiness from backend issues. (We used to have a lot of those.) Adding new API calls invalidates the pre-recorded test data and causes tests to fail.

The new code channels all API calls through the Replay "client". That's how we record (and replay) them. This change (`client.loadRegion` -> `client.requestFocusRange`) looks like it might introduce a _new_ call to some of the tests, which might explain the failure.

You can read more about those tests here if you're interested:
https://github.com/replayio/devtools/tree/main/packages/bvaughn-architecture-demo/playwright

bvaughn: I pulled this branch down and ran it locally to confirm the failing tests. I noticed a few tests had outdated screenshots, and one test failed because of the reason mentioned above: "_should support setting focus range via the context menu_"

1. To fix the hard failure above, I followed [these steps](https://github.com/replayio/devtools/tree/main/packages/bvaughn-architecture-demo/playwright#updating-protocol-fixture-data) to update the protocol fixture data. (Basically I focused the test mentioned above and then ran `yarn test:update-fixture-data`).
2. To update the changed screenshots, I followed [these steps](https://github.com/replayio/devtools/tree/main/packages/bvaughn-architecture-demo/playwright#locally-to-update-snapshots). (Basically I ran `earthly +playwright-update-snapshots --HOST=host.docker.internal --push +deploy` in the "playwright" directory.) The resulting changes looked _wrong_ though, and I realized that this branch was based off of a broken revision in master (fixed by 2ceb3a70ada47ef81b614f3eab386cddb09a449d) so I rebased.

I force pushed to this branch with the updated fixtures JSON. I'd expect CI to pass now.

bvaughn (approved):
The Workflow failure is not related to your change. Feel free to merge whenever you'd like and someone on frontend will sort it out.

PR #8339, opened 2022-12-13
E2e tests

bvaughn:
- [x] Don't render "Add comment" button in Source context menu unless user is authenticated
- [x] Pass `apiKey` prop through to `SessionContext` (this will unblock future work on e2e tests that require authorization via an API key)


markerikson (approved):

bvaughn:
I reverted this PR because I think it might have broken something. I'll look closer in the morning.

bvaughn:
Actually I think it was something messed up with my session from trying out an api key because it happened even on main.

bvaughn:
I'm going to try to re-land the `apiKey` portion of the PR from yesterday now.

PR #8349, opened 2022-12-14
Add preview images to visual comments

bvaughn:
### [Loom overview](https://www.loom.com/share/33b6a1b48cc645b4be3cc087354e22d3)

This change makes use of the generic `primaryLabel` and `secondaryLabel` attributes to encode base-64 image data for visual comments. I've filed FE-1058 for some cleanup work around comments in general.

### DIFF
--- a/src/ui/components/KeyboardShortcuts.tsx
+++ b/src/ui/components/KeyboardShortcuts.tsx
@@ -2,6 +2,7 @@ import { useEffect, useMemo } from "react";
 import { ConnectedProps, connect } from "react-redux";
 
 import { SHOW_GLOBAL_SEARCH_EVENT_TYPE } from "bvaughn-architecture-demo/components/search-files/SearchFiles";
+import { getBase64Png } from "bvaughn-architecture-demo/src/utils/canvas";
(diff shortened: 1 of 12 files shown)


src/ui/components/KeyboardShortcuts.tsx
```diff
@@ -111,15 +112,25 @@ function KeyboardShortcuts({
...
         e.preventDefault();
-        createFrameComment(null, recordingId);
+
+        let base64PNG: string | null = null;
+        const canvas = document.querySelector("canvas#graphics");
+        if (canvas) {
+          base64PNG = await getBase64Png(canvas as HTMLCanvasElement, {
+            maxWidth: 200,
+            maxHeight: 200,
+          });
```
bvaughn, line 130: The exact max dimension is debatable. We can't send the full image (too large, GraphQL server will barf)

markerikson (approved):

src/ui/components/KeyboardShortcuts.tsx
bvaughn: Alternately we could take an image of the area around the marker.

PR #8351, opened 2022-12-14
Source viewer context menu changes

bvaughn:
Includes several incremental changes being made in support of issues FE-932 and FE-1059:
- [x] Expose context menu `target` element via new `ContextMenuContext`. This could be used by the source viewer to associated a context menu operation with a source code column
  - [x] Disable pointer events for search result highlight overlay (since they might interfere with `event.target`)
- [x] Add `data-column-index` attribute to markup generated for syntax highlighted code. This can be used to quickly determine the position of the token in the source code line
- [x] Add `data-parsed-token` attribute as a more reliable way of detecting an HTML element containing a source token than the previous `className` based approach


hbenl (approved):

bvaughn:
The three failing e2e tests also fail for me locally on both this branch and `main` so I'm not sure what that indicates.

PR #8356, opened 2022-12-15
Source search result highlight takes column breakpoints into account

bvaughn:
Fixes FE-1059:
- [x] Search highlight should account for column breakpoints
- [x] Add e2e tests for this case

### Before
Given the following search highlight:
Screen Shot 2022-12-15 at 10 31 14 AM
Adding a log point with column breakpoints before the "0" would cause the highlight to become misaligned:
Screen Shot 2022-12-15 at 10 31 23 AM

### After
After this change, the search highlight will _split around_ the column breakpoint:
Screen Shot 2022-12-15 at 10 29 06 AM


packages/bvaughn-architecture-demo/components/sources/SearchResultHighlight.tsx
```diff
@@ -1,24 +1,91 @@
+import { ReactNode } from "react";
+
 import styles from "./SearchResultHighlight.module.css";
 
+// Must stay in sync with --column-breakpoint-width in SourceListRow.module.css
+const COLUMN_BREAKPOINT_WIDTH_IN_PIXELS = 20;
```
bvaughn, line 6: This feels like an unfortunate hack. Maybe I can think of a way to decouple these.

PR #8361, opened 2022-12-15
Code complete type-ahead fetches prototype properties also

bvaughn:
This implementation does not seem ideal. Preferably [`Pause.getObjectPreview`](https://static.replay.io/protocol/tot/Pause/#method-getObjectPreview) would support a parameter to include own properties or all properties. As it is, having to walk the prototype chain and request each level as a waterfall seems pretty inefficient. Still, this fixes the bug being reported.

- [x] Add/update e2e tests


packages/bvaughn-architecture-demo/playwright/tests/console.ts
```diff
@@ -393,14 +394,35 @@ test("should evaluate terminal expressions at an execution point", async ({ page
...
 });
 
+test("should suggest type-ahead options from prototype objects as well", async ({ page }) => {
+  await setup(page);
+  await toggleProtocolMessage(page, "logs", true);
+
+  const listItem = await locateMessage(page, "console-log", "This is a log");
+  await seekToMessage(page, listItem);
+
+  await page.fill("[data-test-id=ConsoleTerminalInput]", "window.location.propertyIs");
```
bvaughn, line 413: This is an example of a property that _only_ exists on the prototype.

packages/bvaughn-architecture-demo/components/lexical/plugins/code-completion/findMatches.ts
```diff
@@ -61,6 +61,20 @@ function fetchQueryData(
           !PREVIEW_CAN_OVERFLOW
         );
         properties = preview?.properties || null;
+
+        // Auto-complete should not just include own properties.
+        // If there's a prototype chain, fetch those properties also.
+        let currentPrototypeId = preview?.prototypeId;
+        while (currentPrototypeId) {
```
markerikson, line 68: :speech_balloon: only real question is if we'd want to limit the depth here

markerikson (approved):

packages/bvaughn-architecture-demo/components/lexical/plugins/code-completion/findMatches.ts
bvaughn: Hm. Yeah, that's an interesting question. I wonder what that limit should be? I guess I could pick something arbitrary (like 10) to avoid an infinite loop but high enough to still probably handle 99% of the cases

hbenl (approved):

PR #8364, opened 2022-12-16
Don't wrap parens around expressions sent to Pause.evaluateInGlobal

bvaughn:
We were wrapping expressions in parens (e.g. `a + 1` was modified to `(a + 1)` before being sent to `Pause.evaluateInGlobal`). This operation could cause a _valid expression_ to become _invalid_.

I'm not sure why we were doing this, but it doesn't seem to be necessary based on some limited testing, so this commit removes the parens.

### Before

https://user-images.githubusercontent.com/29597/208111550-bf445d0d-b15b-4cdf-b4ed-056aae49fc9c.mp4

### After

https://user-images.githubusercontent.com/29597/208111656-8c037d8b-ff66-4ac1-8d8a-993d3fd206b9.mp4

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -176,7 +176,7 @@ export class ReplayClient implements ReplayClientInterface {
     if (frameId === null) {
       const response = await client.Pause.evaluateInGlobal(
         {
-          expression: `(${expression})`,
+          expression,
           pure: false,
         },
         sessionId,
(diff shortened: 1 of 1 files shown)


hbenl (approved):

loganfsmyth:
I'm 99% sure this would have been to allow people to type an object literal like `{ foo: "a" }` and actually print an object to the console. As far as I'm aware, Firefox, Chromium, and Node all try to special-case things like that, because otherwise it's not parsed as an object.

I agree that the solution you removed isn't a very good one, but we may want to consider adding support for that in with a bit more validation or something.

bvaughn:
Interesting! I'll follow up.

PR #8367, opened 2022-12-16
Show all breakable column positions for a line

bvaughn:
FE-1070 brought to my attention the fact that the Source viewer has never displayed markers for more than the _first_ breakable column on a given line:
Screen Shot 2022-12-16 at 11 28 46 AM

Compare this to Chrome:
Screen Shot 2022-12-16 at 9 25 58 AM

At first I thought this was a regression with the new Source viewer, but looking at the old Source viewer confirmed that it also showed only the first breakable location.

This commit fixes the Source viewer to display _all_ breakable locations (assuming the experimental "column breakpoints" feature has been enabled). It also tidies up the rendering logic for this a bit to make it less redundant and easier to read.

### Plain text
Screen Shot 2022-12-16 at 10 42 45 AM

### Syntax highlighted
Screen Shot 2022-12-16 at 11 23 23 AM

- [x] Update `SourceListRow` to account for multiple breakable point per line
  - [x] Log points should remain relative to the _first_ breakable location
  - [x] Mirror Chrome's behavior when clicking to toggle a particular column breakpoint
- [x] Refactor `SyntaxParsingCache` to return an data structure rather than an HTML string
- [x] Update search result highlighting logic to account for multiple column breakpoint markers
- [x] Add some new e2e/screenshot tests for this feature

### Demo

https://user-images.githubusercontent.com/29597/208173947-7cfb2bf2-b015-4923-8297-4d30a5d1dcb9.mp4


jcmorrow:
Nice, thanks @bvaughn !

PR #8373, opened 2022-12-18
Wakeables detect (sync) circular thennable chains and throw

bvaughn:
Given the underlying _cause_ of FE-1073 seems to be code that unconditionally (re-)attaches listeners to a _wakeable_, I think the best initial step is to detect and throw an error in that case.

The heuristics used are a bit arbitrary, but I wanted to avoid triggering false positives for cases where e.g. many React components in a tree a _suspended_ on the same source of data.

### DIFF
--- a/packages/bvaughn-architecture-demo/src/utils/suspense.ts
+++ b/packages/bvaughn-architecture-demo/src/utils/suspense.ts
@@ -1,5 +1,7 @@
 import { Wakeable } from "../suspense/types";
 
+let SYNCHRONOUS_CIRCULAR_THENABLE_CHECK_MAX_COUNT = 1_000;
+
 // A "thennable" is a subset of the Promise API.
 // We could use a Promise as thennable, but Promises have a downside: they use the microtask queue.
 // An advantage to creating a custom thennable is synchronous resolution (or rejection).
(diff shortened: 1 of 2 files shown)


packages/bvaughn-architecture-demo/src/utils/suspense.ts
```diff
@@ -12,6 +14,27 @@ export function createWakeable<T>(): Wakeable<T> {
...
+  const checkCircularThenableChain = () => {
+    if (++callbacksRegisteredDuringCurrentFrame > SYNCHRONOUS_CIRCULAR_THENABLE_CHECK_MAX_COUNT) {
+      throw Error("Circular thenable chain detected (infinite loop?)");
+    }
+
+    if (resetCounterTimer === null) {
+      resetCounterTimer = setTimeout(() => {
+        resetCounterTimer = null;
+
+        callbacksRegisteredDuringCurrentFrame = 0;
```
Domiii, line 33: Why reset the count? Shouldn't we just have one count per `wakeable`?
This reset might overlook infinite loops that are triggered asynchronously via non-microtasks?

Domiii (approved):
Not sure, I can make a good enough assessment of this, but I'll be curious to learn how the faulty call graph looks like as a whole - please let me know! 😂

packages/bvaughn-architecture-demo/src/utils/suspense.ts
bvaughn: Perhaps it’s an unnecessary detail, but I was mostly trying to avoid potential false positives.

PR #8375, opened 2022-12-18
Test e2e tests (ef4acca)

bvaughn:
Every PR/commit has failed at least some of the e2e tests in the past few days.

Stepping back through recent devtools repo commits, ef4acca (~4 days ago) was the most recent one to pass all e2e tests. (The commit before it, 3c67e01, also passed e2e tests.) It seems like _something_ regressed after this point– either on the frontend or backend (or both 💩) so this PR re-runs the frontend code from ef4acca to see if it still passes.

If it does, we'll assume a frontend regression is most likely. If not, a backend regression.


bvaughn:
Interesting outcome. Only one test failed, and it's one that was added more recently than the commit in question.

bvaughn:
Getting conflicting results from #8377 so going to run again.

bvaughn:
This branch has now twice passed all but the newer Cypress e2e test.

PR #8376, opened 2022-12-18
Test e2e tests (e2f1d0c)

bvaughn:
Given the commit history on `main` and the outcome of #8375, I would expect multiple e2e tests to fail on this PR (although I don't yet understand why). If it does, that would seem to point at a frontend cause for the e2e flakiness.


bvaughn:
This branch failed all but the newer, Cypress e2e test:
* cypress-01: Test basic cypress reporter functionality

PR #8377, opened 2022-12-18
Test e2e tests (3c67e01)

bvaughn:
Given the commit history on `main` and the outcome of #8375, I would expect e2e tests to pass on this PR (although I don't yet understand why).


bvaughn:
First run of tests on this branch (which was green 4 days ago) shows the same number of failures as #8374 (opened today) which seems to suggest that recent frontend changes are _not_ related to the flakiness.

Could be recently backend changes (given the streak of passing tests followed by the streak of failures) _or_ test fragility and chance.

bvaughn:
This branch has now twice failed e2e tests.

PR #8380, opened 2022-12-19
Add workaround for bad sourcemap entries (SCS-172)

hbenl:
Before: https://app.replay.io/recording/stepping-broken--dc27d57a-9c41-4ccf-9dcb-c843cdc09c71
After: https://app.replay.io/recording/stepping-fixed--fa2379ab-76d3-47ee-92a6-e84086d6bb23


ryanjduffy (approved):
I don't have a lot of context on these changes but they seem reasonable given the feature is behind a flag.

PR #8391, opened 2022-12-20
Add debug label to wakeable to help track circular chains

bvaughn:
Sentry error [3825236237](https://sentry.io/organizations/replay/issues/3825236237) doesn't have enough information to help uniquely identify **where** it happens in the React tree or **what** it hangs while loading. This commit adds some additional debug metadata to help with that.

It's not great and I'm open for alternate suggestions.

### DIFF
--- a/packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
+++ b/packages/bvaughn-architecture-demo/src/hooks/useIndexedDB.ts
@@ -46,6 +46,7 @@ export const {
   getValueAsync: getIDBInstanceAsync,
   getValueIfCached: getIDBInstanceIfCached,
 } = createGenericCache<[dbOptions: IDBOptions], IDBPDatabase>(
+  "useIndexedDB: getIDBInstance",
   async dbOptions => {
     const { databaseName, databaseVersion, storeNames } = dbOptions;
     // Create a single shared IDB instance for this DB definition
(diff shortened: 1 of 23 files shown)


markerikson (approved):

Domiii:
It is maybe worth noting that `react` uses `setTimeout` for scheduling render and other calls, but, in some environments, for some cases, it uses `setImmediate` (which afaik has a higher priority than the standard macrotask queue for `setTimeout`) but only in Node + jsdom:

```js
  var localSetImmediate = typeof setImmediate !== 'undefined' ? setImmediate : null; // IE and Node.js + jsdom
```

(link: [react-dom source](https://unpkg.com/react@18.2.0/umd/react.development.js))

So your current code using `setTimeout(reset)` should be able to reset between multiple render calls in `chrome` :thinking:. In that case, 1000 does sound like an awful lot, but who knows what else is happening here :shrug:

Domiii:
Also: Chrome does not provide complete asynchronous stacktraces (except for `async/await` stacks afaik), unless the debugger is active. Meaning that `Promise.then`, `setTimeout` and other asynchronous schedulers are not going to be part of `Error.stack`.

If you want complete asynchronous stacktraces, instrumentation is sadly necessary. This is so bad that `sequelize` recently decided to add extra stack patching code in a few special places, just so you can find out which `query` caused an exception (which since they migrated away from `bluebird` (which had async promise stacks)  and until recently has not been possible).

[Here is a discussion on incomplete async stacktraces](https://github.com/nodejs/node/issues/36126) pertaining to Node.js, but this is of course a general problem in all of V8, and afaik, also applies to all other contemporary JS engines.

Similarly, (as you are probably aware :smile:) `replay` also does not yet have proper async stacktrace support. This is tracked in [BAC](https://linear.app/replay/issue/BAC-1326/improve-async-call-stacks) and [RUN-1022](https://linear.app/replay/issue/RUN-1022/runtime-support-for-asynchronous-call-stacks-acs).

PR #8392, opened 2022-12-20
tests: Avoid using 'DEBUG' env var, and improve 'addBreakpoint' more

loganfsmyth:


packages/e2e-tests/helpers/source-panel.ts
```diff
@@ -28,6 +28,7 @@ export async function addBreakpoint(
   if (url) {
     await openSourceExplorerPanel(page);
     await openSource(page, url);
+    await page.waitForSelector('[data-test-name="Source"]:visible');
```
bvaughn, line 31: Maybe this line would be better placed inside of the `openSource` helper function?

packages/e2e-tests/package.json
```diff
@@ -10,7 +10,7 @@
     "test": "playwright test",
     "test:chromium": "playwright test --project chromium --reporter=@replayio/playwright/reporter,line",
     "test:replay-chromium": "playwright test --project replay-chromium --reporter=@replayio/playwright/reporter,line",
-    "test:debug": "DEBUG=1 playwright test",
+    "test:debug": "playwright test --workers 1 --headed",
```
bvaughn, line 13: No objection to this change but curious what motivates it. Is there any practical difference (other than ... just using the built-in flags?)

bvaughn (approved):

packages/e2e-tests/package.json
loganfsmyth: I could also have renamed `DEBUG` to `TEST_DEBUG` or something. DEBUG in the node ecosystem is already heavily used by https://www.npmjs.com/package/debug which I ran into because I set it to debug a playwright issue and it unexpectedly made the tests run headed which took me a bit to track down.

bvaughn: Oh I see. That's interesting.

PR #8393, opened 2022-12-20
Run console expression eval at ExampleFinished.

loganfsmyth:


packages/e2e-tests/tests/console_eval.test.ts
```diff
@@ -10,17 +10,13 @@ import { seekToPreviousScreenshot } from "../helpers/timeline";
...
-test("console_eval: support global console evaluations", async ({ page }) => {
+test("console_eval: support console evaluations", async ({ page }) => {
   await startTest(page, url);
   await openDevToolsTab(page);
   await openConsolePanel(page);
 
-  // in e2e tests replay always starts at the very end of the recording but at that point
-  // console evaluations always fail, so we seek to the point of the last screenshot instead
-  await seekToPreviousScreenshot(page);
-  await executeAndVerifyTerminalExpression(page, "333", 333);
```
bvaughn, line 21: Interesting that this (apparently?) used to cover this case. I wonder when it stopped working? Presumably it worked at one point?

cc @hbenl in case you have ideas

bvaughn (approved):

packages/e2e-tests/tests/console_eval.test.ts
loganfsmyth: I just don't think there's any guarantee that the previous screenshot will be at a location that can eval. It's probably more likely, so given that it's a hardcoded list of example recordings, it would only really be an issue if the example fixture gets re-recorded. I wasn't seeing this fail locally, but I was seeing it fail pretty randomly when running the backend e2e tests since they always re-record the fixtures

bvaughn: As someone unfamiliar with this code, I assumed a _screenshot_ was like a _paint_ and that we wouldn't have one without a DOM but...maybe I just don't understand what the `seekToPreviousScreenshot` helper is supposed to do. Looks like it just focuses on the timeline and presses the left arrow.

bvaughn: We should probably delete that `seekToPreviousScreenshot` helper entirely. Seems sketchy.

bvaughn: We should probably delete that `seekToPreviousScreenshot` helper entirely. Seems sketchy.

PR #8395, opened 2022-12-20
Show error message when Elements panel fails to load document

bvaughn:
https://user-images.githubusercontent.com/29597/208764902-9ca0fc7d-7ad2-4366-a065-bb9c9a2be3af.mp4


src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -133,15 +134,21 @@ export function setupMarkup(store: UIStore, startAppListening: AppStartListening
...
+      if (!isInspectorSelected(getState())) {
         await condition((action, currState) => {
           return isInspectorSelected(currState);
         });
+      }
+
+      try {
         await loadNewDocument();
+      } catch (error) {
+        console.error(error);
```
bvaughn, line 146: Should we try to narrow the error down to avoid catching other errors? I'm not sure how many types of error could be thrown here.

markerikson (approved):

PR #8396, opened 2022-12-20
inspector-02 test: reduce flakiness

bvaughn:
In FE-1075, Logan pointed out that the `inspector-02` test sometimes gets stuck waiting for the Elements panel to load because the test loads at the _end_ of the recording, and document may not still be available at that time.

This commit changes the test to seek to an earlier point in time before waiting for the Elements panel to load.

### DIFF
--- a/packages/e2e-tests/tests/inspector-02.test.ts
+++ b/packages/e2e-tests/tests/inspector-02.test.ts
@@ -1,7 +1,7 @@
 import test, { expect } from "@playwright/test";
 
 import { openDevToolsTab, startTest } from "../helpers";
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/tests/inspector-02.test.ts
```diff
@@ -1,6 +1,7 @@
 import test, { expect } from "@playwright/test";
 
 import { openDevToolsTab, startTest } from "../helpers";
+import { openConsolePanel, warpToMessage } from "../helpers/console-panel";
```
loganfsmyth, line 4: Unused `openConsolePanel` import FYI

loganfsmyth (approved):

PR #8397, opened 2022-12-20
Update React DevTools client to work with Chromium

markerikson:
This PR:

- Updates our Webpack config to support importing certain files as raw strings
- Ports the backend logic I wrote for injecting source files and individual functions into a specific paused browser instance over to the client side
- Ports the logic that actually mutates the global React DevTools stubs from our Chromium fork and injects the real RDT extension bundles
- Updates the RDT bundles to be built from our React repo fork from the latest `main`, with changes applied to the original RDT source to work with our persistent object IDs for React fiber objects and actually initialize the RDT extension logic on injection (as well as stripping out unnecessary chunks of code to partially shrink the extension bundles)

For the changes to the RDT source, see:

- https://github.com/replayio/react/compare/main...replayio:react:feature/replay-react-devtools

As far as I can tell, this logic generally works.  Assuming that the correct React operations annotations can be fetched, the RDT UI component will show the correct component tree, mousing over component tree nodes will highlight them in our preview, and clicking on a component node will fetch its props from the paused backend browser instance:

image

The `ReactDevTools.tsx` file is still a bit hacked up, especially with the logic for loading the correct React DevTools version.  That will need to be fixed.

The main issue at this point will be figuring out how to deploy the RDT data extraction "routine" to our backend so that it can collect the proper operations annotations and persist them.  The client already knows how to ask for those annotations, so that part won't actually need to change.

For my local dev work so far, I have been copy-pasting a generated annotations array directly into `reactDevTools.ts` and pretending that's what got fetched, so I am confident the rest of the client-side handling for those works okay.


markerikson:
Okay, we've got evidence that the backend piece works, and the client setup in this PR works. Merging!

PR #8399, opened 2022-12-23
Fix stepping when the selected frame is not the top frame (FE-1082)

hbenl:
Before: https://app.replay.io/recording/broken-stepout--0ef0a398-6f72-4d8b-8725-9ef6bf4a54ba
After: https://app.replay.io/recording/fixed-stepout--ddf04b03-bece-4ea3-86af-f3f8002ed546

This happened because the current execution point was after the last frame step in the selected frame, so we didn't step over or out from any of the frame steps but from the current execution point instead. Furthermore we never chose a frame step when stepping out because the operation type is called `stepOut` but we were looking for `stepUp`.


jasonLaster (commented):
Could you add some e2e tests for this behavior?

PR #8403, opened 2022-12-27
Try to manually refresh auth state for unauthenticated users

ryanjduffy:
I thought the Auth0 SDK would automatically use refresh tokens to refresh access tokens for expired sessions but apparently not.

When `this.update()` is called and the user is not authenticated, try to refresh via `getAccessTokenSilently()`. If it fails, silently redirect to the login page because either the user was logged out or there was some error but the user didn't have an active session so a fresh login will likely fix things.

### DIFF
--- a/src/ui/utils/tokenManager.tsx
+++ b/src/ui/utils/tokenManager.tsx
@@ -172,37 +172,49 @@ class TokenManager {
       return;
     }
 
-    if (this.auth0Client.isAuthenticated) {
-      if (this.isTokenRequested) {
-        if (refresh) {
-          this.reset();
-        } else {
-          return;
-        }
-      }
+    const wasAuthenticated = this.auth0Client.isAuthenticated;
 
-      this.isTokenRequested = true;
-      const deferredState = this.deferredState;
-
-      const item = window.localStorage.getItem("__cypress");
-      if (item) {
-        const token = JSON.parse(item).body.access_token;
-        this.setState({ token }, deferredState);
+    if (this.isTokenRequested) {
+      if (refresh) {
+        this.reset();
+      } else {
         return;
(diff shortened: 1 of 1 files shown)


src/ui/utils/tokenManager.tsx
```diff
@@ -212,7 +224,7 @@ class TokenManager {
     try {
       return await this.auth0Client.getAccessTokenSilently({ audience, ignoreCache: refresh });
     } catch (e: any) {
-      if (e.error !== "login_required" && e.error !== "consent_required") {
```
ryanjduffy, line 215: `login_required` is the expected response if the user was not authenticated when we call this under the new circumstances of this PR so we'll let that be thrown and caught above.

src/ui/utils/tokenManager.tsx
```diff
@@ -172,37 +172,49 @@ class TokenManager {
...
-        const token = await this.fetchToken(refresh);
+    this.isTokenRequested = true;
+    const deferredState = this.deferredState;
 
-        this.setState({ token }, deferredState);
-        if (deferredState === this.deferredState) {
-          this.setupTokenRefresh(token);
-        }
-      } catch (e) {
+    const item = window.localStorage.getItem("__cypress");
```
jasonLaster, line 188: what's cypress doing in there?

jasonLaster (approved):

src/ui/utils/tokenManager.tsx
ryanjduffy: I dunno. I assume it's legacy automated test workaround code. I'm pretty sure it can be removed. I almost did here actually but didn't want to mix it in with this change.

PR #8429, opened 2022-12-31
Tweak context menu background colors

bvaughn:


jonbell-lot23 (commented):
I think we might have had a misunderstanding here. In this screen, I was saying both blue and red were wrong:

image

They should be #3d4552 for dark theme and #F8F8F9 in light theme to align with our other menus. And when we resolve our design tokens, we won't be hard-coding like this, we'll be pointing these to --menu-bgcolor so we have a single place to update all menu items.

Sorry that this is the issue that won't go away!

PR #8448, opened 2023-01-01
Design token naming convention alignment

jonbell-lot23:
We have always used the conventions "bgcolor" and "color" in our design system. While updating the Figma, I noticed that we fell out of this convention in a few key places, so I updated them.

**Q&A**

Q: I think "bgcolor" is dumb, and prefer "background-color"
A: The design system standardized on bgcolor a long time ago, so we can update everything to background-color in a separate PR. But in the meantime, we need to standardize on one thing, and this PR does that. (for some common classes, not all)

Q: I feel strongly about this. 
A: I think we all feel strongly about consistency, and this PR makes things more consistent. As stated above, we can always choose to take the time to move the entire system to a new word, but that's less important to me than consistency. Our code got less consistent, and this PR fixes that. We can discuss naming conventions as a separate topic.


jonbell-lot23:
I see 26 green tests (1 skipped), nothing failing, and visually my check looks good. Squashing and merging.

bvaughn:
> But in the meantime, we need to standardize on one thing, and this PR does that.

This commit changed **six** references in `packages/bvaughn-architecture-demo/pages/variables.css` to "bgcolor" and left ~80 other references as "background-color".

I don't really understand the motivation here.

jonbell-lot23:
While analysing our design system, those were the six examples that were misaligned with the conventions we had established. See screenshot from [this Figma page](https://www.figma.com/file/ASas6u2DMihEEzw8jPT1XC/Replay-Component-Library?node-id=936%3A4358):
image

By fixing those, I got our "Foundations" page back to using a common naming convention, which is step one for getting back to a single, documented design system instead of two out-of-sync ones.

It's true that the bvaughn-architecture-demo is still extremely different to the core design system we use, and I'm eager to get that resolved as part of https://linear.app/replay/issue/FE-1090/align-our-design-token-system-to-our-new-variablescss-document .

jonbell-lot23:
In the meantime, I'm happy to see this PR reverted in the short term.

I apologise for the friction and confusion, and know we'll soon figure out a way to reconcile the original system with your new code. Once we do, this will become moot.

PR #8450, opened 2023-01-02
Adjusting width of privacy window

jonbell-lot23:
[Linear ticket](https://linear.app/replay/issue/DES-54/our-privacy-dropdown-menu-is-too-wide)

We were forcing the dropdown to be w-80 rather than letting it flex based on the content inside. I had trouble allowing the flex, but removing the w-80 and the "members of" string should address the 99.9% case.

Old:
image

New:
image


jonbell-lot23:
48 passing tests and no failures. Squashing and merging.

PR #8451, opened 2023-01-02
Initialize devtools as soon as we have a sessionId (FE-1076)

hbenl:
We used to wait for the first `LoadedRegions` event before initializing devtools, but that was unreliable and apparently not necessary (see FE-1076). Now we just wait for the session to be created.

### DIFF
--- a/packages/e2e-tests/tests/object_preview-02.test.ts
+++ b/packages/e2e-tests/tests/object_preview-02.test.ts
@@ -1,7 +1,7 @@
 import test, { expect } from "@playwright/test";
 
 import { openDevToolsTab, startTest } from "../helpers";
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/tests/object_preview-02.test.ts
```diff
@@ -18,6 +18,7 @@ const url = "doc_rr_objects.html";
 test(`object_preview-02: should allow objects in scope to be inspected`, async ({ page }) => {
   await startTest(page, url);
   await openDevToolsTab(page);
+  await waitForTerminal(page);
```
jasonLaster, line 21: Nice. Happy to see the improvements to the e2e tests

PR #8453, opened 2023-01-02
Open HTML elements from the object inspector (FE-448)

hbenl:


packages/bvaughn-architecture-demo/components/inspector/values/HTMLElementRenderer.tsx
```diff
@@ -77,14 +79,8 @@ export default function HTMLElementRenderer({
...
     if (inspectHTMLElement !== null) {
-      if (timestampedPointContext !== null) {
-        inspectHTMLElement(
-          protocolValue,
-          pauseId,
-          timestampedPointContext.executionPoint,
-          timestampedPointContext.time
-        );
-      }
+      const { executionPoint, time } = timestampedPointContext || timelineContext;
```
bvaughn, line 82: I think this makes sense but it could probably use an inline comment explaining _why_ we're doing this. (I don't think it's very common to destructure values from multiple contexts like this.)

src/devtools/client/debugger/src/components/shared/InspectorContextReduxAdapter.tsx
```diff
@@ -36,58 +43,20 @@ export default function InspectorContextReduxAdapter({ children }: { children: R
...
-            dispatch(openNodeInInspector(objectId));
-          }
-        }
-      })();
+    (protocolValue: ProtocolValue, pauseId: PauseId, point: ExecutionPoint, time: number) => {
+      if (pauseId !== ThreadFront.currentPause.pauseId) {
+        ThreadFront.timeWarpToPause({ point, time, pauseId }, false);
+      }
+      dispatch(selectNode(protocolValue.object!));
+      dispatch(setSelectedPanel("inspector"));
```
bvaughn, line 52: Nifty!

bvaughn (approved):

PR #8463, opened 2023-01-03
Highlight inspectable tokens on hover

bvaughn:
[Loom overview](https://www.loom.com/share/0758b04370a24c08a42f9c255a2f1c24)

https://user-images.githubusercontent.com/29597/210426768-c9533dfa-a810-4198-aab2-3e5f7678fbda.mp4

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/sources/utils/classNameToTokenTypes.ts
+++ b/packages/replay-next/components/sources/utils/classNameToTokenTypes.ts
@@ -1,0 +1,7 @@
+import { TokenType } from "../../lexical/plugins/code/types";
+
+export default function classNameToTokenTypes(className: string): TokenType[] {
+  // e.g. "tok-comment" -> ["comment"]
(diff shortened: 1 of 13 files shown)


packages/replay-next/components/sources/utils/classNameToTokenTypes.ts
```diff
@@ -0,0 +1,7 @@
+import { TokenType } from "../../lexical/plugins/code/types";
+
+export default function classNameToTokenTypes(className: string): TokenType[] {
+  // e.g. "tok-comment" -> ["comment"]
+  // e.g. "tok-variable tok-definition" -> ["variable", "definition"]
```
bvaughn, line 5: This was a case we weren't properly handling before:
```js
let foo = "abc";
```

In the above example, `foo` is both a variable _and_ a definition.

markerikson (approved):

PR #8190, opened 2022-11-22
A series of small visual tweaks

jonbell-lot23:
Old:
image

New:
image

Old:
image

New:
image

Also removed the "Resolve recording" experiment. I did that a long time ago and it's a long-expired test.

PR #8466, opened 2023-01-03
Fix Info panel horizontal alignment of icons and text

bvaughn:
### After

I think the _schedule_ icon was always slightly too small (causing the text after it to be misaligned). The "Anyone with link..." dropdown misalignment seems to have regressed in #8190


jonbell-lot23 (approved):
Egg on my face since I'm the one that caused this bug. Apologies.

markerikson (approved):

bvaughn:
> Egg on my face since I'm the one that caused this bug. Apologies.

It happens 😁

PR #8468, opened 2023-01-03
Fix Replay Delta screenshot workflow

bvaughn:
Trying two things here:
- [ ] ~~Clone new Workflow but leave Yarn cache enabled~~ ([failed](https://github.com/replayio/devtools/actions/runs/3832655916/jobs/6523221561))
- [x] Disable Yarn cache for old Workflow ([worked](https://github.com/replayio/devtools/actions/runs/3832655930/jobs/6523221918))


bvaughn:
cc @jasonLaster Disabling Yarn cache for now as it breaks things. (I don't know why.)

PR #8469, opened 2023-01-03
BAC-2511: allow create session to take focus window

miriambudayr:


miriambudayr:
Note to self: the protocol package will need to be bumped to 37 once it is deployed.

package.json
```diff
@@ -42,7 +42,7 @@
     "@redux-devtools/app": "^2.1.4",
     "@redux-devtools/extension": "^3.2.2",
     "@reduxjs/toolkit": "^1.8.4",
-    "@replayio/protocol": "0.36.0",
+    "@replayio/protocol": "0.37.0",
```
bvaughn, line 45: I think this change should have an accompanying change in `yarn.lock`. Did you run `yarn install` ?

src/ui/actions/session.ts
```diff
@@ -229,67 +229,87 @@ export function createSocket(
...
-              // and at most we might want to see if the initial lines match or something.
-              clonedResponse.result = {
-                ...clonedResponse.result,
-                contents: clonedResponse.result.contents.slice(0, 1000),
-              };
+      const initialPausePoint = await getInitialPausePoint(ThreadFront.recordingId!);
+      const focusRegion =
+        initialPausePoint && "focusRegion" in initialPausePoint
+          ? initialPausePoint.focusRegion
+          : undefined;
```
bvaughn, line 236:
```suggestion
      const focusRegion = initialPausePoint?.focusRegion;
```

src/ui/utils/environment.ts
```diff
@@ -98,6 +98,12 @@ export function hasLoadingParam() {
   return getURL().searchParams.get("loading") != null;
 }
 
+export function getFocusRegion() {
+  const url = getURL();
+  const focusRegionParam = url.searchParams.get("focusRegion");
+  return focusRegionParam ? (decodeBase64FromURL(focusRegionParam) as TimeStampedPointRange) : null;
```
bvaughn, line 104: Should we `try`/`catch` this in case? (I know it's not _new_ code, just code that's been moved.)

src/ui/actions/session.ts
miriambudayr: I think the way I wrote it is necessary because `initialPausePoint()` can return one of two types:

`{point: string, time: number, focusRegion: TimestampedPointRange | null}`

or

`{point: string, time: number}`

src/ui/utils/environment.ts
miriambudayr: Sure. Are you worried about the call to `decodeBase64FromURL`?

bvaughn: Hm, I was, but then I looked and _that_ function catches and returns `undefined` so...my bad :smile:

src/ui/actions/session.ts
bvaughn: The code before was:
```js
const focusRegion =
  initialPausePoint && "focusRegion" in initialPausePoint
    ? initialPausePoint.focusRegion
    : undefined;
```

I think that's functionally equivalent to:
```js
const focusRegion = initialPausePoint?.focusRegion;
```

Isn't it?

miriambudayr: It isn't functionally equivalent because this type doesn't allow `focusRegion` to be optional:

`{point: string, time: number, focusRegion: TimestampedPointRange | null}`

The easy thing here is to just make `focusRegion` optional.

src/ui/utils/environment.ts
miriambudayr: :)

src/ui/actions/session.ts
bvaughn: Sorry @miriambudayr. Can you point out to me _what_ the functional difference is between those two statements? Because they should evaluate exactly the same. (Or are you saying TSC complains?)

src/ui/utils/environment.ts
```diff
@@ -98,6 +98,18 @@ export function hasLoadingParam() {
...
 
+export function getFocusRegion() {
+  const url = getURL();
+  const focusRegionParam = url.searchParams.get("focusRegion");
+  try {
+    return focusRegionParam
+      ? (decodeBase64FromURL(focusRegionParam) as TimeStampedPointRange)
+      : undefined;
+  } catch (e) {
+    return undefined;
```
bvaughn, line 109: This _used_ to be `null` and now it's `undefined`. I don't know if that matters or not but it looks like a change in behavior?

bvaughn (approved):

PR #8470, opened 2023-01-03
Try re-enabling Yarn cache for Delta Workflow

bvaughn:
Follow up to #8468

### DIFF
--- a/packages/replay-next/package.json
+++ b/packages/replay-next/package.json
@@ -37,7 +37,7 @@
     "typescript": "^4.6.3"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


packages/replay-next/package.json
```diff
@@ -37,7 +37,7 @@
     "typescript": "^4.6.3"
   },
   "scripts": {
-    "dev": "next dev",
+    "dev": "npx next dev",
```
bvaughn, line 40: So if the Yarn cache is slightly broken, use NPX as a fallback. I don't _love_ this but...

PR #8473, opened 2023-01-04
Send caller stacktrace to Sentry on CommandError

hbenl:
When our frontend receives an error from the backend, the corresponding Sentry issues are not very useful because their stacktraces don't show where in our code the request originated.
With this PR we send the caller stacktrace to Sentry whenever we receive an error from the backend (note that this will create a second Sentry issue for every `CommandError`).

### DIFF
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -12,6 +12,7 @@ import {
   analysisPoints,
   analysisResult,
 } from "@replayio/protocol";
+import { captureException } from "@sentry/react";
 
 import { commandError } from "shared/utils/error";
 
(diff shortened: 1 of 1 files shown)


packages/protocol/socket.ts
```diff
@@ -12,6 +12,7 @@ import {
   analysisPoints,
   analysisResult,
 } from "@replayio/protocol";
+import * as Sentry from "@sentry/react";
```
markerikson, line 15: :hand: We probably _shouldn't_ import the _entire_ Sentry library - just `captureException` by itself.

bvaughn (approved):
image

PR #8474, opened 2023-01-04
Replay Next Reconciliation: Tabs

jonbell-lot23:
[See Notion doc here](https://www.notion.so/replayio/Replay-next-reconciliation-e5761570ddec4888984afc8c031ca663)

This is a very small change to see if anything breaks. It didn't!


bvaughn (approved):
Quick spot-check looks good 👍🏼

PR #8475, opened 2023-01-04
Replay Next Reconciliation: Core design tokens 

jonbell-lot23:
We started with tabs and nothing broke. So this is a meatier PR following the same approach, and things are still not breaking:

Timeline
Menus
Code Mirror
Breakpoints
Testsuites
Toolbar
Toolbar buttons
Buttons
Accordion headers
Selection
Border color that splits the toolbars/panels/headers.
Icon colors

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -219,7 +219,7 @@
   /*
   Migration notes:
     * removed from src/devtools/client/themes/variables.css
-    * things below thing point are not organized yet
+    * things below this point are not organized yet
   */
 
   /* Timeline (Dark) */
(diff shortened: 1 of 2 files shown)


bvaughn (changes requested):
Looks like this change broke something.

On `main` (and the previous branch) this is what the Settings dialog looks like in light mode:
Screen Shot 2023-01-05 at 10 35 11 AM

On this branch it looks like this:
Screen Shot 2023-01-05 at 10 35 01 AM

jonbell-lot23:
Found the typo, fixed it, ready for re-review

bvaughn:
Code looks good but the Lint task is failing because of a formatting issue. You can fix this by running: this command in the root directory
```sh
npm run lint
```

jonbell-lot23:
Whoops, lint issue addressed.

(Side note: I see that merging is blocked, but I'm having trouble finding the way to say "resolve issue.")

bvaughn (approved):

PR #8476, opened 2023-01-04
Add a "Copy" button to the Object Inspector

bvaughn:
Resolves FE-989

[Loom overview](https://www.loom.com/share/974ec9cfbf6a43d08c536129b9c8abf9)

- [x] Fix interop between inspector context menu and preview popover
- [x] Add context menu with "copy" operation to various inspector locations (mimic Chrome behavior)
- [x] Copy serialized value to clipboard
  - [x] Simple values (string, number, boolean, null, undefined)
  - [x] Special formatted values (regex, date, error)
  - [x] Nested values (object, array, map, set, html element)
- [x] e2e tests
  - [x] Update old tests to not trigger false positives on the new context menus
  - [x] Add new context menu tests for copying various complex and simple values
  - [x] Review and compare updated screenshots ([here](https://delta.replay.io/project/dcb5df26?branch=FE-989))


packages/replay-next/components/inspector/useInspectorContextMenu.tsx
```diff
@@ -0,0 +1,90 @@
...
+          case "set":
+            console.log("TODO [FE-989]", object);
+            break;
+          case "object":
+            console.log("TODO [FE-989]", object);
+            break;
+        }
+      }
+
+      if (valueToCopy != null) {
```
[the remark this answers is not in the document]
bvaughn: What the fuck is this

packages/replay-next/components/inspector/protocolValueToText.ts
```diff
@@ -1,89 +1,196 @@
...
 import { dateProtocolObjectToString } from "replay-next/components/inspector/values/DateRenderer";
 import { errorProtocolObjectToString } from "replay-next/components/inspector/values/ErrorRenderer";
+import { functionProtocolObjectToString } from "replay-next/components/inspector/values/FunctionRenderer";
 import { regExpProtocolObjectToString } from "replay-next/components/inspector/values/RegExpRenderer";
 import { getObjectWithPreviewHelper } from "replay-next/src/suspense/ObjectPreviews";
 import { protocolValueToClientValue } from "replay-next/src/utils/protocol";
-import { functionProtocolObjectToString } from "replay-next/components/inspector/values/FunctionRenderer";
+import { ReplayClientInterface } from "shared/client/types";
+
+const MAX_DEPTH_TO_COPY = 5;
```
markerikson, line 19: :speech_balloon: probably good enough for now. I guess if they want more depth they can right-click on a more deeply-nested field and grab that?

(Could also consider adding a secondary context option to copy _everything_.)

markerikson (approved):

bvaughn:
The only screenshot failure that looks potentially legit is the `html-text` one from the `source-preview.ts` test. I'm going to dig in a little more.

### main
Screen Shot 2023-01-05 at 6 43 04 PM

### this branch
Screen Shot 2023-01-05 at 6 43 11 PM

packages/replay-next/components/inspector/protocolValueToText.ts
bvaughn: > (Could also consider adding a secondary context option to copy everything.)

I think I'd rather avoid that. The UX feels awkward.

>  I guess if they want more depth they can right-click on a more deeply-nested field and grab that?

Yes. You can expand an object and copy any of its properties.

bvaughn:
Added some explicit tests for the deeply nested/truncated values.

PR #8482, opened 2023-01-05
Remove unsafe uses of ThreadFront.currentPause.pauseId

hbenl:
In some places (mostly the Elements panel) we still used `ThreadFront.currentPause.pauseId`. The problem is that this property is `null` when the user switched to a new execution point and the new `pauseId` has not yet been received from the backend. That's why `await ThreadFront.getCurrentPauseId()` should be preferred whenever possible.
I think the only uses of `ThreadFront.currentPause.pauseId` that can be considered safe are:
- to check if the user has navigated (or is currently navigating) to a different pause
- if we need to get the `pauseId` synchronously and bail out if it's `null`

This fixes some of the Sentry issues in FE-1056.

### DIFF
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -294,7 +294,7 @@ export async function repaint(force = false) {
     ThreadFront.currentTime,
     await ThreadFront.getCurrentPauseId(replayClient),
     (_time, pauseId) => {
-      return pauseId !== ThreadFront.currentPause.pauseId;
+      return pauseId !== ThreadFront.currentPauseIdUnsafe;
     },
     force
   );
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -219,15 +219,14 @@ class _ThreadFront {
     });
   }
 
-  get currentPause(): Pause {
-    return {
-      point: this.currentPoint,
-      time: this.currentTime,
-      pauseId:
-        this.currentPauseId ??
-        getPauseIdForExecutionPointIfCached(this.currentPoint)?.value ??
-        null,
-    };
+  /**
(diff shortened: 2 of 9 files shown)


bvaughn:
Lots of e2e tests failures.

bvaughn:
>  That's why await `ThreadFront.getCurrentPauseId()` should be preferred whenever possible.

I'm concerned that having the following two properties/methods disagree is *very unintuitive*:
* `ThreadFront.currentPause.pauseId`
* `ThreadFront.getCurrentPauseId()`

A method with the name "get current thing" should reasonably be expected to return the same value as a property "current thing". Seems like this will be a very easy thing for us to reintroduce.

Can we rename `get currentPause` to `get currentPauseUnsafe` or something that makes it more explicit that it's not the thing we should be using in most cases?

hbenl:
> Can we rename get currentPause to get currentPauseUnsafe or something that makes it more explicit that it's not the thing we should be using in most cases?

Theoretically the types should be enough as a warning but in practice it's too common to just append `!` if typescript "annoys" with type errors.
So I've replaced `ThreadFront.currentPause.pauseId` with `ThreadFront.currentPauseIdUnsafe` (I noticed that the other properties of `ThreadFront.currentPause` weren't used anymore so I removed them - I had only added them originally for backward compatibility during the removal of the `Pause` class).
While doing that I stumbled on some dead code which I also removed.

bvaughn:
> So I've replaced `ThreadFront.currentPause.pauseId` with `ThreadFront.currentPauseIdUnsafe`

Nice! Love it.

bvaughn (approved):
Just skimmed it, but this looks like a really nice cleanup 👍🏼

PR #8483, opened 2023-01-05
Handle ChunkLoadErrors in the app's ErrorBoundary (FE-1079)

hbenl:
STR:
- build locally with `yarn build`
- remove the file `.next/static/chunks/1824.*.js`
- `yarn start`
- open `http://localhost:8080`
- click on "View settings"

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -106,14 +106,14 @@ function Routing({ Component, pageProps }: AppProps) {
         <link rel="icon" type="image/svg+xml" href="/images/favicon.svg" />
         <title>Replay</title>
       </Head>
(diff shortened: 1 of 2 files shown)


bvaughn (approved):

markerikson:
Out of curiosity, when are these `ChunkLoadError`s occurring in prod?

hbenl:
> Out of curiosity, when are these ChunkLoadErrors occurring in prod?

When you use app.replay.io so that it tries to load a chunk (e.g. if you're in the Library and open the settings panel) but a new version has been deployed since you initially opened app.replay.io.

PR #8496, opened 2023-01-07
Add a video endpoint

jasonLaster:
This adds a video endpoint that uses the protocol to synthesize a movie. Eventually, this should live in the backend, but given that it seemed fairly easy to implement with a vercel endpoint with sufficient caching, i was curious to see how far this experiment could go...

Motivation:
I want to be able to include replay videos in linear issues

Steps
- [x] Add a video.ts endpoint that can create a video given a recordingId
- [x] Add a video.js script that can download a video and write it to disk
- [x] Expose the endpoint in the UI so that users can download the video
- [x] Add authentication so that logged in users can download their own videos
- [x] Add a UI for downloading the video

I think this is ready to go
https://www.loom.com/share/54e7f05b39ba4175819a00ab4bf9c5d5

### DIFF
--- a/packages/node-protocol/README.md
+++ b/packages/node-protocol/README.md
@@ -1,0 +1,21 @@
+## Replay Node Protocol
+
+Provides a protocol client that can be used from Node.js scripts to communicate with the Replay.io backend via the [protocol](https://replay.io/protocol).
+
+### Example usage
+
+This script, creates a Replay client that can create a debugging session.
+
+```js
+import { createClient } from "node-protocol";
+
+const recordingId = "889363e9-08b9-4755-904f-447549558575";
+
+const { client } = await createClient();
+const { sessionId } = await client.Recording.createSession({ recordingId });
+console.log(`sessionId: ${sessionId}`);
+```
+
+### Case Study: Downloading a video
+
+See [video.ts](https://github.com/replayio/devtools/blob/main/pages/api/video.ts) for an example of how the protocol can be used to create a video of a recording on demand.
--- a/pages/api/video.ts
+++ b/pages/api/video.ts
@@ -1,0 +1,119 @@
+import fs from "fs";
+import os from "os";
+import path from "path";
+import { PaintPoint, ProtocolClient } from "@replayio/protocol";
+import { NextApiRequest, NextApiResponse } from "next";
+
+import { defer } from "node-protocol/utils";
+import { createClient } from "node-protocol";
+
+const ffmpegPath = require("@ffmpeg-installer/ffmpeg").path;
+
+const ffmpeg = require("fluent-ffmpeg");
+ffmpeg.setFfmpegPath(ffmpegPath);
+
+const address = process.env.NEXT_PUBLIC_DISPATCH_URL;
+
+async function fetchPaintPoints({
+  client,
+  sessionId,
+}: {
+  client: ProtocolClient;
+  sessionId: string;
+}) {
+  const findPaints = client.Graphics.findPaints({}, sessionId);
+  let gPaints: PaintPoint[] = [];
+  client.Graphics.addPaintPointsListener(({ paints }) => {
+    gPaints.push(...paints);
+  });
+
+  await findPaints;
+  console.log("found points", gPaints.length);
+
+  return gPaints;
+}
+
+async function fetchGraphics(
+  tmpDir: string,
+  paintPoints: PaintPoint[],
+  { client, sessionId }: { client: ProtocolClient; sessionId: string }
+) {
+  let prevTime = 0;
+  let input = "";
+  const numPaints = paintPoints.length;
+  await Promise.all(
+    paintPoints.map(async (point, index) => {
+      const duration = (point.time - prevTime) / 1000;
+      const filename = path.join(tmpDir, `/${index}-${point.point}-${point.time}.jpeg`);
+      input += `file '${filename}'\nduration ${duration}\n`;
+
+      prevTime = point.time;
+
+      const { screen } = await client.Graphics.getPaintContents(
+        { point: point.point, mimeType: point.screenShots[0].mimeType },
+        sessionId
+      );
+      console.log(`${index}/${numPaints} - ${filename}`);
+      fs.writeFileSync(filename, Buffer.from(screen.data, "base64"));
+    })
+  );
+
+  const inputFilename = path.join(tmpDir, "/input.txt");
+  console.log(inputFilename);
+  fs.writeFileSync(inputFilename, input);
+  return inputFilename;
+}
+
+async function createVideo(tmpDir: string, input: string, recordingId: string): Promise<string> {
+  const { promise, resolve, reject } = defer<string>();
+  const filename = path.join(tmpDir, `/${recordingId}.mp4`);
+
+  console.log(`starting ffmpeg`);
+  const proc = ffmpeg();
+  proc.input(input).inputOptions(["-safe 0"]).inputFormat("concat");
+
+  proc
+    .on("end", () => {
+      console.log("Finished processing");
+      resolve(filename);
+    })
+    .on("error", (err: any) => {
+      console.log("Error:", err);
+      reject(err);
+    })
+    .outputOptions(["-c:v libx264", "-r 30", "-pix_fmt yuv420p"])
+    .save(filename);
+
+  return promise;
+}
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { recordingId, token } = req.body;
+  const tmpDir = os.tmpdir();
+  console.log(`creating video for ${recordingId}...`);
+
+  try {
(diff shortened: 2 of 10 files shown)


jasonLaster:
Added a simple download button that downloads the video. We should
1. add a loom style flow for downloading
2. add affordances to show loading, finished, failed state

We also need to figure out how to pass along headers/cookies etc so that the function can download private replays

jasonLaster:
Turns out that authentication is fairly straight forward with `Authentication.setAccessToken`
https://static.replay.io/protocol/tot/Authentication/#method-setAccessToken

pages/api/video.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const fileStream = fs.createReadStream(video);
+    const filename = path.basename(video);
+
+    res.setHeader("Content-Type", "video/mp4");
+    res.setHeader("Content-Disposition", `inline; filename="${filename}"`);
+
+    fileStream.pipe(res);
+  } catch (error) {
+    console.log(error);
```
jazzdan, line 115: Maybe this should be `console.error`?

pages/api/video.ts
```diff
@@ -0,0 +1,118 @@
...
+    res.status(500).send(error);
```
jazzdan, line 116: We might not want to send the entire error to the client because it could leak internal information. Just logging the error should be OK.

src/api/socket.ts
```diff
@@ -0,0 +1,110 @@
...
+  EventParams,
+  PauseId,
+  ProtocolClient,
+  SessionId,
+} from "@replayio/protocol";
+import WebSocket from "ws";
+
+import { defer } from "./utils";
+
+const address = "wss://dispatch.replay.io";
```
jazzdan, line 15: This should be parameterized so that it will work in cloud-dev and ephemeral environments. We have existing environment variables for this

src/api/socket.ts
```diff
@@ -0,0 +1,110 @@
...
+export function createClient(): Promise<clientRes> {
+  const { promise, resolve } = defer<clientRes>();
+  function addEventListener<M extends EventMethods>(
+    event: M,
+    handler: (params: EventParams<M>) => void
+  ) {
+    gEventListeners.set(event, handler);
+  }
+
+  function removeEventListener() {}
```
jazzdan, line 42: This fucntion is empty, is that intended?

pages/api/hello.ts
```diff
@@ -0,0 +1,5 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  return res.status(200).json({ name: "John Doe" });
+}
```
bvaughn, line 5: Did you accidentally commit his file?

pages/api/video.ts
```diff
@@ -0,0 +1,118 @@
...
+  sessionId: string;
+}) {
+  const findPaints = client.Graphics.findPaints({}, sessionId);
+  let gPaints: PaintPoint[] = [];
+  client.Graphics.addPaintPointsListener(({ paints }) => {
+    gPaints.push(...paints);
+  });
+
+  await findPaints;
+  console.log("found points", gPaints.length);
```
bvaughn, line 29: I assume this (and the other console logs) were intentionally left in place?

pages/api/video.ts
```diff
@@ -0,0 +1,118 @@
...
+    await client.Authentication.setAccessToken({ accessToken: token });
+
+    const { sessionId } = await client.Recording.createSession({ recordingId });
+    console.log(`sessionId: ${sessionId}`);
+
+    const tmpDir = os.tmpdir();
+    const paintPoints = await fetchPaintPoints({ client, sessionId });
+    const inputFilename = await fetchGraphics(tmpDir, paintPoints, { client, sessionId });
+
+    const video = await createVideo(tmpDir, inputFilename, recordingId);
```
bvaughn, line 103: nit: maybe naming this `videoFilename` would be a little clearer (and match `inputFilename` above)

src/api/socket.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  const client = new ProtocolClient({
+    sendCommand: sendMessage,
+    addEventListener,
+    removeEventListener,
+  });
+
+  return promise;
+}
```
bvaughn, line 110: It's not clear to me why we need a new `socket.ts` vs the one we already have.

src/ui/components/shared/SharingModal/SharingModal.tsx
```diff
@@ -123,9 +127,178 @@ function EnvironmentVariablesRow() {
...
+          <div className="flex flex-col space-y-1 overflow-hidden">
+            <div className="font-bold">Privacy Settings</div>
+            <PrivacyDropdown {...{ recording }} />
+            {showEnvironmentVariables ? <EnvironmentVariablesRow /> : null}
+          </div>
+        </div>
+        <CopyButton recording={recording} />
+      </section>
+      <section className="bg-menuHoverBgcolor px-8 pb-6">
+        {!recording.private && recording.operations && (
```
bvaughn, line 160: Why'd you add the `!recording.private &&` check?

src/ui/components/shared/SharingModal/SharingModal.tsx
```diff
@@ -123,9 +127,178 @@ function EnvironmentVariablesRow() {
...
+      console.log(`Video downloaded ${recording.id}`);
+      setDownloadState("success");
+
+      a.download = "replay.mp4";
+      document.body.appendChild(a);
+      a.click();
+      document.body.removeChild(a);
+    };
+    xhr.onerror = function () {
+      console.log(`Failed to download video ${recording.id}`);
```
bvaughn, line 269: nit: `console.error`

src/ui/components/shared/SharingModal/SharingModal.tsx
```diff
@@ -123,9 +127,178 @@ function EnvironmentVariablesRow() {
...
+      setDownloadState("error");
+    };
+    xhr.open("POST", "/api/video");
+    xhr.setRequestHeader("Content-Type", "application/json");
+    xhr.send(JSON.stringify({ recordingId: recording.id, token: token.token }));
+    console.log(`Downloading recording ${recording.id}`);
```
bvaughn, line 275: Generally I think we should probably remove `console.log`s from code that runs in the browser. I think those are probably only good to leave in when we're development/debugging stuff.

pages/api/hello.ts
jasonLaster: haha - yep

pages/api/video.ts
jasonLaster: Yes - on the backend that's the only way to see what happened

src/api/socket.ts
jasonLaster: I spent some time trying to use the original socket.ts, but I don't want to make changes to it for the Node use case

src/ui/components/shared/SharingModal/SharingModal.tsx
jasonLaster: This is probably a different PR. good catch

packages/node-protocol/README.md
```diff
@@ -0,0 +1,20 @@
+## Replay Node Protocol
```
jasonLaster, line 1: I think moving this out to its own package makes it clearer why it's needed and also easier to re-use if we decide to publish it down the line

bvaughn: Yeah, that seems like a good call.

src/ui/components/shared/SharingModal/SharingModal.tsx
```diff
@@ -123,9 +127,176 @@ function EnvironmentVariablesRow() {
...
+          <div className="flex flex-col space-y-1 overflow-hidden">
+            <div className="font-bold">Privacy Settings</div>
+            <PrivacyDropdown {...{ recording }} />
+            {showEnvironmentVariables ? <EnvironmentVariablesRow /> : null}
+          </div>
+        </div>
+        <CopyButton recording={recording} />
+      </section>
+      <section className="bg-menuHoverBgcolor px-8 pb-6">
+        {recording.operations && (
```
bvaughn, line 160: Sorry, I missed this the first time. Why'd this get moved _inside_ of the section? Seems like we might end up with an empty `<section></section>` now. Would that look weird?

bvaughn (approved):
I just skimmed it this second time but 👍🏼  If it seems to be working well, let's try it out?

src/ui/components/shared/SharingModal/SharingModal.tsx
jasonLaster: Ugh - I should really decouple this change from the video change. The rationale was that before the UI was missing bottom padding when there weren't any operations. Will quickly revert this change.

PR #8504, opened 2023-01-09
Skip auth token refresh in the replay browser

ryanjduffy:


src/ui/utils/tokenManager.tsx
```diff
@@ -172,6 +172,11 @@ class TokenManager {
       return;
     }
 
+    if (window.__IS_RECORD_REPLAY_RUNTIME__) {
+      this.setState({}, this.deferredState);
```
ryanjduffy, line 176: Settings the state here clears the default `loading: true` state used by `useToken` which was what was blocking rendering the app in the first attempt.

jaril (approved):

PR #8506, opened 2023-01-10
Playwright screenshot helper clips visible bounds

bvaughn:
So far as I can tell from the [sparse documentation](https://playwright.dev/docs/screenshots#element-screenshot) there is no way to tell Playwright to trim the hidden/overflowing parts of screenshots. This causes unnecessary churn for parts of our UI that overflow and get hidden. For example, lines in the Source viewer may overflow horizontally:
image

Because of this, screenshots incorrectly include partial Console messages:
image

This has the downstream effect of causing screenshots to be invalidated when unrelated things change (like Console messages).

This PR changes our screenshot helper to auto trim. The above screenshot should now look like this:
image

I expect a lot of Delta conflicts from this PR.

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -621,7 +621,7 @@
   --theme-selection-color: #ffffff;
 
   /* Border color that splits the toolbars/panels/headers. */
-  --theme-splitter-color: var(--grey-25);
+  --theme-splitter-color: #e0e0e2; /* TODO @jonbell var(--grey-25) is undefined */
   --theme-emphasized-splitter-color: var(--grey-30);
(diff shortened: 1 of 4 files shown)


bvaughn:
Looks like this had an unexpected second benefit of making the images crisper for some reason.

bvaughn:
Scrolling through all of the screenshot differences in Delta and it's looking good.

e.g. before
image

and after
image

bvaughn:
I see literally only **one** image that looks worse after this change: "[uncaught exception stack collapsed](https://delta.replay.io/project/dcb5df26?branch=FE-1114&snapshot=140593)"

before:
image

after:
image

Going to take a closer look at that one.

jasonLaster (approved):
Nice

PR #8508, opened 2023-01-10
Jon fixing merge conflicts 2/2

jonbell-lot23:


jonbell-lot23:
1. Went through preview branch manually and things look ok
2. Looked at Delta -- the 3 different screens appear to be different data, not visual bugs
3. Lint isn't failing anymore, woo

This merge got messy but the best I can tell we're in good shape.

bvaughn:
Those screenshot diffs look _wrong_ so I'm going to re-run the Playwright test.

bvaughn (commented):
I think this looks goo overall, but diffing the two theme sections in `packages/replay-next/pages/variables.css` after this changes shows there are still a few differences that are important to get sorted out asap.

Light has the following variables that Dark doesn't have:
* `--theme-toggle-inner`
* `--theme-toggle-selected`
* `--theme-toolbar-panel-icon-color`
* `--theme-errant-popup-color`

And Dark has this variable that Light doesn't have:
* `--theme-toolbar-hover`

Check out this comment for an example of how that can cause trouble:
https://github.com/replayio/devtools/pull/8506#issuecomment-1376534226

If you can patch up those differences, I think this is otherwise good to land 👍🏼

jonbell-lot23:
I am keen to get this live! Will address this when I start my day

bvaughn (commented):
https://www.loom.com/share/052d7cb1aebc406eb71b03430610b3e5

These variables only exist in the light theme:
* `--tab-standard-color`
* `--tab-standard-hover-bgcolor`

These variables only exist in the dark theme:
* `--cm-property`
* `--cm-atom`

And this value is defined multiple times in each theme:
* `--tab-bg-color`

jonbell-lot23:
> Looking at the `variables.css` file in this branch, after your changes.
> 
> These variables only exist in the light theme:
> 
> * `--tab-standard-color`
> * `--tab-standard-hover-bgcolor`
> 
> These variables only exist in the dark theme:
> 
> * `--cm-property`
> * `--cm-atom`
> 
> And this value is defined multiple times in each theme:
> 
> * `--tab-bg-color`
> 
> Here's a Loom walk through of how I put together that list: https://www.loom.com/share/052d7cb1aebc406eb71b03430610b3e5

Thanks, Brian! I went through the files and the issues you called out are already in production, rather than being introduced in this PR. I'm in Merge Hell and every day I'm getting less confident in the state of these files. Can we land this despite the issues you found, considering they're not regressions from prod, then I'll loop back to fix them in another ticket?

bvaughn (approved):

bvaughn:
Yeah that's fair. Let's fix them with a small follow up!

bvaughn:
The Playwright task failed on its most recent run so let me kick it off one more time.

jonbell-lot23:
I moved these final variable misalignments into this Linear ticket: https://linear.app/replay/issue/DES-222/reconcile-misaligned-styles

PR #8511, opened 2023-01-10
Better caching for terminal expressions and eager eval

bvaughn:
### [Loom demo and overview of this change](https://www.loom.com/share/7354e449e266476b879e138886d4285e)

- [x] Suspense cache supports (optional) instance id per expression being evaluated
- [x] Console eager eval and terminal expression renderers pass this instance id to cache-bust
- [x] Add new e2e test and ~~update test fixture data~~

cc @Andarist


bvaughn:
Merging this for now. Happy to make changes via follow up PR.

PR #8512, opened 2023-01-10
Share Modal + Privacy Panel fix

jasonLaster:
When we designed the privacy panel, the idea was that we would only show the privacy information for public replays. @jaril / @jonbell-lot23 was there a moment where we intentionally decided to show it all the time?

Screenshot 2023-01-10 at 8 07 58 AM

Screenshot 2023-01-10 at 8 07 53 AM


jonbell-lot23:
> @jaril / @jonbell-lot23 was there a moment where we intentionally decided to show it all the time?

Nope, this isn't by design

PR #8513, opened 2023-01-10
Left sidebar can be fully collapsed

bvaughn:
### [Loom overview/demo](https://www.loom.com/share/5b15dd06e205442ca48627ee89f427ac)

- [x] Upgrade `react-resizable-panels` package from 0.0.30 to 0.0.31
- [x] Left toolbar panel is now fully collapsible via mouse and keyboard interactions.
  - [x] Toggle button added to left toolbar to expand/collapse as well.
- [x] New e2e test added for collapsible behavior.

cc @jonbell-lot23

### DIFF
--- a/src/ui/components/Toolbar.tsx
+++ b/src/ui/components/Toolbar.tsx
@@ -1,6 +1,6 @@
 import classnames from "classnames";
 import classNames from "classnames";
-import React, { useContext, useEffect, useState } from "react";
+import React, { ReactNode, useContext, useEffect, useState } from "react";
 
 import AccessibleImage from "devtools/client/debugger/src/components/shared/AccessibleImage";
 import { getPauseId } from "devtools/client/debugger/src/selectors";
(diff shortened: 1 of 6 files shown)


jonbell-lot23:
Copy/pasting from Discord:

> Another UX feature we've wanted for a long time, and I continue to be very excited about it.
> 
> Some could argue that the little arrow in the corner is too much. Strongly disagree. Requiring a gesture to undo an action is a big no-no. You always need a clickable target for something like this, and messaging of state. 

#shipit!

src/ui/components/Toolbar.tsx
```diff
@@ -147,6 +147,13 @@ export default function Toolbar() {
           </>
         ) : null}
         {logProtocol ? <ToolbarButton icon="code" label="Protocol" name="protocol" /> : null}
+
+        {sidePanelToggle !== null && (
+          <>
+            <div className="grow"></div>
+            <div className="relative px-2">{sidePanelToggle}</div>
+          </>
+        )}
```
jasonLaster, line 156: Not sure if we need the extra button or if the vs code ux where you click the active button to toggle it is fine?

My personal preference is that we should bring back the vs code ux and remove the sidebar toggle button. But, I'd be open to re-adding the button if we see users struggle to collapse the sidebar.

bvaughn: Jon and I agreed this button was a nice addition.

PR #8514, opened 2023-01-10
Ensure frame order

jcmorrow:
And speculatively try to fix aspect ratio.

### DIFF
--- a/pages/api/video.ts
+++ b/pages/api/video.ts
@@ -39,24 +39,25 @@ async function fetchGraphics(
   { client, sessionId }: { client: ProtocolClient; sessionId: string }
 ) {
   let prevTime = 0;
-  let input = "";
   const numPaints = paintPoints.length;
-  await Promise.all(
-    paintPoints.map(async (point, index) => {
-      const duration = (point.time - prevTime) / 1000;
-      const filename = path.join(tmpDir, `/${index}-${point.point}-${point.time}.jpeg`);
-      input += `file '${filename}'\nduration ${duration}\n`;
+  const input = (
+    await Promise.all(
+      paintPoints.map(async (point, index) => {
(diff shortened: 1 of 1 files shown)


pages/api/video.ts
```diff
@@ -39,24 +39,25 @@ async function fetchGraphics(
   { client, sessionId }: { client: ProtocolClient; sessionId: string }
 ) {
   let prevTime = 0;
-  let input = "";
   const numPaints = paintPoints.length;
-  await Promise.all(
-    paintPoints.map(async (point, index) => {
-      const duration = (point.time - prevTime) / 1000;
-      const filename = path.join(tmpDir, `/${index}-${point.point}-${point.time}.jpeg`);
-      input += `file '${filename}'\nduration ${duration}\n`;
```
jcmorrow, line 48: this mutative operation can theoretically happen in any order, I think?

pages/api/video.ts
```diff
@@ -81,7 +82,7 @@ async function createVideo(tmpDir: string, input: string, recordingId: string):
       console.log("Error:", err);
       reject(err);
     })
-    .outputOptions(["-c:v libx264", "-r 30", "-pix_fmt yuv420p"])
+    .outputOptions(["-c:v libx264", "-r 30", "-pix_fmt yuv420p", `-vf "scale=100:-2,setdar=16/9"`])
```
jcmorrow, line 85: > If one and only one of the values is -n with n >= 1, the scale filter will use a value that maintains the aspect ratio of the input image, calculated from the other specified dimension. After that it will, however, make sure that the calculated dimension is divisible by n and adjust the value if necessary.

http://ffmpeg.org/ffmpeg-filters.html#scale-1

PR #8516, opened 2023-01-11
Don't disable the console when fetching frames fails (FE-1110)

hbenl:


bvaughn (changes requested):
This fix breaks how Suspense works. We _expect_ `getFramesSuspense` to throw.

I'm also not sure how this fix lines up with the comments I left on [the recording](https://app.replay.io/recording/cannot-evaluate--8da26b2a-6448-4c08-bd92-6dab5979d9e6?point=26935039955686650221307487328403555&time=11242). Can you help me understand the connection?

hbenl:
> This fix breaks how Suspense works. We expect getFramesSuspense to throw.

D'oh!

> I'm also not sure how this fix lines up with the comments I left on the recording. Can you help me understand the connection?

There is no connection - I just spotted another issue which I think is the real culprit.
I don't think `useContext` really throws here, there are many ways in which the hit counts can be misleading, I wouldn't rely on them to deduce the control flow.

bvaughn:
> There is no connection - I just spotted another issue which I think is the real culprit.

I don't think that's backed up by the Replay though. Did you check out where my comments are?

bvaughn (commented):
Maybe this change is okay but it doesn't solve the original issue filed in FE-1110.

I am a little confused about this change though. Do we expect frame fetching to fail for some reason? More than any of the other Suspense fetching requests we make? Why?

hbenl:
> You can step in from that point and see that it throws.

Huh, I didn't see that; admittedly I didn't really investigate your theory since `getFrames()` definitely throws in this case and that would trigger the `ErrorBoundary`, so it *seemed* like the right explanation.
Can you add a comment to the recording where `useContext` throws?

> I am a little confused about this change though. Do we expect frame fetching to fail for some reason? More than any of the other Suspense fetching requests we make?

Well, I've seen it fail a couple of times and I thought falling back to global evaluations in that case was a reasonable fix.

bvaughn:
> Can you add a comment to the recording where `useContext` throws?

I did yesterday. Added another one this morning at the specific spot where React throws. Are you not seeing them?

bvaughn (approved):
I don't know. Maybe this change is okay? I'll leave it up to your judgement.

It seems a little fishy though. Might cause subtle, unexpected changes in behavior?

hbenl:
- I think it's reasonable to fall back to global evaluations if we can't fetch frames (and I can't think of a situation where this would cause an unwanted change in behavior - currently we don't allow evaluations at all in this case)
- I could reproduce the issue by opening https://app.replay.io/recording/replay-of-ts-node--9926d09f-a6c2-4033-8c47-3dfe3be3e319?point=0&time=0 (note the `point`/`time` parameters) and this fix works in that case
- I've seen recordings where `useContext` also throws without causing any visible issues, so I think that was really not the culprit
- the pause for which we tried to fetch frames in the original recording was created for execution point 0 but the timeline claims that we're paused at the end of the recording - I'll fix this in a separate PR (we should initially pause at the endpoint if we find no other usable initial pause point)
- that would probably also fix this issue (because we would fetch frames for the endpoint instead of 0) but since I've occasionally seen `getFrames()` failing I still think this PR is a good idea

bvaughn:
> I've seen recordings where `useContext` also throws without causing any visible issues, so I think that was really not the culprit

I'm confused by your statement. The recording _shows_ that as the issue for FE-1110.

PR #8519, opened 2023-01-11
Updated Playwright tests to fix flakiness/failures

bvaughn:
Updates an e2e tests that has been cause false negatives from our Playwright/Delta actions lately:
- [x] `tests/console`: _should show a button to clear terminal expressions_

This PR also updates tests to `await expect` when testing async values<sup>1</sup>. This causes Playwright to [_wait until the expected condition is met_](https://playwright.dev/docs/test-assertions) (with a timeout) and so it should also help cut down on small timing related flakes.

Would be nice if we could enforce this with a Lint rule but I don't know how that would work.

<sup>1</sup> Note that I only did this for `await expect(await thing)`. I didn't replace `await expect(asyncValue)` because I didn't have an easy way to grep for them.

### DIFF
--- a/packages/e2e-tests/helpers/console-panel.ts
+++ b/packages/e2e-tests/helpers/console-panel.ts
@@ -183,8 +183,9 @@ export async function seekToConsoleMessage(
   await consoleMessage.hover();
   await consoleMessage.locator('[data-test-id="ConsoleMessageHoverButton"]').click();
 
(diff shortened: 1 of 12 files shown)


bvaughn:
I think we're good here. Three runs, no Playwright failures. (We had a Delta upload failure– aka DEL-16– but that's unrelated).

PR #8520, opened 2023-01-11
Always jump to an initial pause point

hbenl:
If the initial pause point was not in a loading region, `jumpToInitialPausePoint()` would not call `ThreadFront.timeWarp()` at all - it should jump to the endpoint instead (that's what it already does if it doesn't find a pause point from URL parameters, comments or paints but the case that it *does* find such a point but it isn't in a loading region wasn't handled correctly).


bvaughn (approved):
Not sure I understand totally but trust your judgement 😁

PR #8530, opened 2023-01-12
Remove next-transpile-modules

hbenl:
In recordings of app.replay.io, the "original" sources under `packages/replay-next/` are actually transpiled sources. When I remove `next-transpile-modules` from our next config, these sources are the real originals. I've used this config locally for a while now and didn't notice any problems.


bvaughn (approved):
Huh...could have sworn this was require for something. @markerikson do you remember?

Seems okay when I gave it a quick test though 🤷🏼

PR #8535, opened 2023-01-13
Removing unused design tokens

jonbell-lot23:
This was a fun one.

1. I wrote a script to find the number of times every variable is used
2. I wrote a second script to use that list to remove everything used zero times
3. Which resulted in this PR!


jonbell-lot23:
Check out this cool spreadsheet where I tracked this stuff:

image

jonbell-lot23:
I got things to the point where Delta is happy. (Other than one screenshot that I think we are frequently seeing as a Delta bug) So this should be ready for review.

ryanjduffy (approved):

PR #8537, opened 2023-01-13
Do not switch from cypress panel when changing view modes

ryanjduffy:
* Adds `cypress` as a viewer panel
* Suppresses switching to the source explorer when switching to Devtools

## Additional Considerations

I think this is worth trying but I worry the devtools features will be missed by not switching the primary panel which draws attention to that side. The new icons appearing on the left are easy to miss.

cc: @jonbell-lot23

### DIFF
--- a/src/ui/actions/layout.ts
+++ b/src/ui/actions/layout.ts
@@ -62,7 +62,10 @@ export function setViewMode(viewMode: ViewMode): UIThunkAction {
     // Otherwise, it's possible for the nag to not be properly dismissed.
     if (viewMode === "dev" && !localNags.includes(LocalNag.YANK_TO_SOURCE)) {
       dispatch(dismissLocalNag(LocalNag.YANK_TO_SOURCE));
-      dispatch(setSelectedPrimaryPanel("explorer"));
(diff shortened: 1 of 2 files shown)


jasonLaster:
I think we should probably take a second pass on switching panels in general. It was always meant to be a NUX feature and it's now become something pretty disruptive as you switch modes...

The biggest thing here is the context switch when you jump into devtools and lose your bearings. The intuition is coming from being on a lot of calls and folks feeling comfortable with cypress and then when they lose that panel getting nervous :)

jaril (approved):

jonbell-lot23:
Good point, @jasonLaster -- I think there are two things at play here:

1. When looking at Cypress, what's the most important panel when switching to DevTools?
2. When looking at other replays, what's the most important panel when switching to DevTools?

I agree with this PR because it optimises for Cypress familiarity in item 1 above, so people can be effective.

I think 2 is an interesting conversation, because it's harder to argue that our custom Replay Info / Events panel feels familiar to people coming from VSCode, Chrome, etc. And as Ryan said, it's easy to miss those icons when they appear on the side. Good context if/when we have that conversation.

PR #8541, opened 2023-01-15
Improve our share modal (Share 2.0)

jonbell-lot23:
[Loom video showing differences between old and new](https://www.loom.com/share/4b042783fda44df8971f44f9bd6b31f3)

[Figma link](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=7572%3A42043)

Linear task: https://linear.app/replay/issue/DES-226/share-20

Before:
image

After:
image


jonbell-lot23:
@bvaughn I have some remaining details I'd like to address around spacing, colours, and making sure dark theme is feeling right, so I'm not fully done.

But would you mind reviewing what I've done so far? Curious to know if I'm heading off in the right direction or if I might be stepping on toes.

packages/replay-next/pages/variables.css
```diff
@@ -201,7 +201,7 @@
   --background-color-current-execution-point: #25364e;
   --background-color-current-search-result: #25364e;
   --background-color-warning: #42381f;
-  --background-color-inputs: #0e1119;
+  --background-color-inputs: var(--theme-text-field-bgcolor);
```
bvaughn, line 204: nit: Since both themes use this same value, may be a good opportunity to collapse two variables?

src/global-css.ts
```diff
@@ -79,7 +79,7 @@ import "ui/components/shared/SettingsModal/SettingsModal.css";
 import "ui/components/shared/SettingsModal/SettingsNavigation.css";
 import "ui/components/shared/SharingModal/CollaboratorsList.css";
 import "ui/components/shared/SharingModal/EmailForm.css";
-import "ui/components/shared/SharingModal/ReplayLink.css";
+import "ui/components/shared/SharingModal/SharingModal.module.css";
```
bvaughn, line 82: CSS modules should only be imported by components, e.g.

```tsx
// MyComponent.tsx
import styles from "MyComponent.module.css";
```

Importing one from a global style sheet like this has no impact (because the style names are all mangled and unused.)

```suggestion
```

src/ui/components/shared/SharingModal/CollaboratorsList.css
```diff
@@ -6,6 +6,19 @@
   max-height: 240px;
 }
 
+.sharing-modal .permissions-list {
```
bvaughn, line 9: I don't think we need this nesting/specificity, do we?

Nice to avoid it as much as possible.
```suggestion
.permissions-list {
```

src/ui/components/shared/SharingModal/SharingModal.module.css
```diff
@@ -0,0 +1,37 @@
+.copy-link input[type="text"] {
```
bvaughn, line 1: This looks like a mistake. Styles in CSS module files (`*.module.css`) will have their names mangled during build to avoid naming conflicts. That's one of the nice things about CSS styles. The names get matched up because you generally refer to them like so:

```tsx
// MyComponent.tsx
import styles from "./MyComponent.module.css";

<div className={styles.Example}>...</div>
```

Generally they shouldn't have hyphens in them because then you have to refer to them like this:
```tsx
<div className={styles["copy-link"}>...</div>
```

In this case, it looks like the style is being referred to like this:
```tsx
<div className="copy-link relative flex flex-shrink-0 flex-col items-center">
```

This won't actually do anything, because the "copy-link" style name in this module will be converted to something random.

Also, looking at the `ReplayLink` file where this is being used– there doesn't seem to be any scenario in which there's an `<input>` inside of a "copy-link" styled component anyway?

bvaughn (commented):
I just kind of skimmed it. Couple of minor visual things I noticed on the preview branch:

Spacing seems a bit off for the focused outline here (bottom button should probably have top margin rather than all top padding):
Screen Shot 2023-01-18 at 9 23 11 AM


Expanded state layout feels a bit wonky (left side feels cramped compared to the right side):
Screen Shot 2023-01-18 at 9 22 47 AM

jonbell-lot23:
Thanks! Yeah, the right-hand drawer has looked strange for a while. Might be worth its own PR.

packages/replay-next/pages/variables.css
jonbell-lot23: I hear you, but prefer the semantic division here.

Because "inputs" are the superset and "textfields" are a subset, I like the idea of the code referring to inputs as inputs and textfields as textfields. It's true that we currently are using the same color between the two, but it's not uncommon to want to apply a border to all textfields but not also get things like checkboxes or forms.

If we do end up changing this out, I'd propose we add it to a separate cleanup ticket.

src/ui/components/shared/SharingModal/SharingModal.module.css
jonbell-lot23: Good eye. This was from @jaril about a year ago, and I'm seeing the same things you are. But I do remember this functionality existing somewhere in the code so I'm reluctant to remove it in case I blow something up.

Because this already exists on production, I'd propose a separate task to investigate this and possibly remove those lines of code if they're truly unused.

jonbell-lot23:
**Status:**

Made a series of small tweaks based on Brian's feedback above, also reached out to Ryan for a once-over. Make sure to check out the Loom video that shows a [90 second walkthrough](https://www.loom.com/share/4b042783fda44df8971f44f9bd6b31f3).

src/ui/components/shared/SharingModal/CollaboratorsList.css
```diff
@@ -6,6 +6,19 @@
   max-height: 240px;
 }
 
+.permissions-list {
+  animation: fadeIn 0.4s ease-in-out;
+}
```
bvaughn, line 11: I think my earlier comment is still relevant here:
https://github.com/replayio/devtools/pull/8541/files/5273ffda3a10103876478a8a58eee869a61f9fd5#diff-380e2a93729a0b270e3bc9f8aeb3af1a097d967993100c0326b42f70eb794334

This style was _added_ in this PR and it's definitely _not_ being referenced, so we should remove it (and preferably the one above it)

src/ui/components/shared/SharingModal/SharingModal.tsx
```diff
@@ -87,19 +88,44 @@ function CollaboratorRequests({ recording }: { recording: Recording }) {
...
+    <section className="space-y-4 bg-modalBgcolor p-4">
       <div className="flex w-full flex-col justify-between space-y-3">
         <div className="w-full space-y-4">
-          <div className="space-y-1.5">
-            <div className="font-bold">Add People</div>
+          <div>
+            <div className="mb-2 font-bold">Team</div>
+
+            <div className="rounded-md border border border-transparent bg-themeTextFieldBgcolor p-2 hover:bg-themeTextFieldBgcolorHover">
+              <PrivacyDropdown {...{ recording }} />
```
bvaughn, line 114: Unnecessary spread?
```suggestion
              <PrivacyDropdown recording={recording} />
```

bvaughn (approved):
Approved overall but I think that new style should be cleaned up.

src/ui/components/shared/SharingModal/SharingModal.tsx
jonbell-lot23: Good call, not actually sure because this is older code. But I trust you! Committed.

src/ui/components/shared/SharingModal/CollaboratorsList.css
jonbell-lot23: Feel free to squash and merge if I'm asleep and we get on the same page with the animation.

src/ui/components/shared/SharingModal/SharingModal.tsx
bvaughn: The expression `...{ recording }` is doing two things:
1. Creating a new object with the property "recording" that points to the value of whatever is in that variable.
2. _Spreading it_ as props passed to `PrivacyDropdown` (one prop per attribute– so a `recording` prop).

So this code:
```jsx
<PrivacyDropdown {...{ recording }} />
```

And this code:
```jsx
<PrivacyDropdown recording={recording} />
```

Will _both_ get compiled to something like this code:
```js
createElement(PrivacyDropdown, { recording });
```

The second one is just easier to read.

src/ui/components/shared/SharingModal/CollaboratorsList.css
bvaughn: Okay. I should have checked out your branch and searched it locally before commenting.

You're right. There is one component still referencing the _global_ class name:
```jsx
export default function CollaboratorsList({
  owner,
  collaborators,
}: {
  owner: User;
  collaborators: CollaboratorDbData[] | null;
}) {
  return (
    <div className="permissions-list overflow-auto">
      <Author user={owner} />
      {collaborators ? <Collaborators {...{ collaborators }} /> : null}
    </div>
  );
}
```

My [initial comment](https://github.com/replayio/devtools/pull/8541/files/5273ffda3a10103876478a8a58eee869a61f9fd5#diff-380e2a93729a0b270e3bc9f8aeb3af1a097d967993100c0326b42f70eb794334) was on a different file and I think I 
mixed the two up when I saw the `permissions-list` style. My apologies.

jonbell-lot23:
> Screen Shot 2023-01-19 at 8 24 56 AM
> 
> 
> 
> This failure means that formatting is wrong. You can fix it by running locally:
> 
> ```
> 
> npm run lint
> 
> ```

Ugh, I lint all the time but I guess this one slipped by me. Apologies! Will lint and merge soon.

jonbell-lot23:
Seeing odd behaviour here. When I type npm run lint from the command line, it says:
✔ No issues

But when I look into the issue, from the link above, it says:
image

So I go to my IDE and I mash prettier a few times just to be sure, but nothing changes.

--

Any suggestions, @bvaughn ? Remove the cartridge and blow in it?

jonbell-lot23:
Update: after fussing with prettier/lint for a while, I guessed that I was just hitting an odd error. Pushed things live, then confirmed things were working.

PR #8553, opened 2023-01-19
Print Statements v3.0

bvaughn:
### [Feature overview (Loom)](https://www.loom.com/share/c7cc7675bc3d4556bbd55bf377698115)

1. Re-arrange components into their own package
2. Add capsule drop-down menu with options to focus and add/remove conditional
3. New UI/UX styles (demonstrated in the video below)

https://user-images.githubusercontent.com/29597/213473117-edab43ee-c0eb-410b-8193-002fe3931678.mp4

Resolves DES-220, FE-1125, FE-1132


packages/replay-next/pages/variables.css
```diff
@@ -308,8 +310,8 @@
   --point-panel-timeline-label-background-color: #000000;
   --point-panel-timeline-label-color: var(--color-brand);
 
-  --badge-picker-background-color: #28282a;
-  --badge-default-color: #747476;
+  --badge-picker-background-color: rgba(0, 0, 0, 0.04);
+  --badge-default-color: rgba(0, 0, 0, 0.08);
```
bvaughn, line 314: cc @jonbell-lot23 These changes roughly approximate the colors we had before, but also handle error states without requiring _additional_ variables to track. We can iterate on this if needed though.

bvaughn: Hm, disregard. I think I'm going to have to add new variables for these invalid states– because of this 💩
Screen Shot 2023-01-19 at 9 48 50 AM

bvaughn:
I think this is ready for review 😄 

Going to look at `breakpoints-07` to see why it's failing (and if it's related to this PR). I may also write a couple of additional screenshot tests. If I change anything though, I'll make it a new commit so as to not interfere with the review.

jonbell-lot23:
Exciting! Will take a look later when I'm up.

jasonLaster:
@bvaughn this looks amazing!

jasonLaster:
Some capsule feedback
https://www.loom.com/share/fef7e709de8741dfb2cd418c3a0b7403?from_recorder=1&focus_title=1

bvaughn:
> Some capsule feedback

@jasonLaster Re the "9" vs "? of 9" – Jon and I chatted about this [on the Linear issue](https://linear.app/replay/issue/FE-1125/implement-capsule-dropdown-in-print-statement-panel#comment-1d6afa37) and "9" was what we agreed on. We can chat more about this off of the PR if you'd like, but I think reasoning holds.

jasonLaster:
Some feedback on focus/blur states. It's a bit rambly, but wanted to share nonetheless
https://www.loom.com/share/d09a2b0414794fdbabc84f1fd9e4c309

bvaughn:
> Some feedback on focus/blur states.
> 0:49: "_It feels a little funny to me that it's only while we're focused_"

@jasonLaster This is intentional. The blue border indicates _focus_ not _editing_ state. I think this is kind of standard focus behavior. (There are also two inputs (potentially) and we need to show which one you're focused in.)

jasonLaster:
re capsule: thanks for linking to the issue. 
* Agreed that sticking with prod and pre-layout is a good call. 
* Will file a follow up issue to discuss the other states since I think it's valuable to know where you are.
* focus start/end could work off of time and closest point, no?

re focus states: yeah - i think you're right. thanks

bvaughn:
> Some capsule feedback

@jasonLaster Re `font-size`, the issue I'm working off of (FE-1125) shows this mockup:

Screen Shot 2023-01-19 at 1 11 23 PM

It does not specify a font size, but comparing our two standard sizes (shown below) the smaller one _seemed_ to match the proportions a little more closely.
Screen Shot 2023-01-19 at 1 11 36 PM
Screen Shot 2023-01-19 at 1 11 48 PM

I don't care about which size we use personally. I think we should just go with one now and iterate later.

bvaughn:
> focus start/end could work off of time and closest point, no?

I think this might be kind of ambiguous and confusing, in practice.

bvaughn:
Okay I applied the `font-size` and edit-hover-button feedback.

bvaughn:
I fixed a few misc things (mostly e2e testing related, plus some of Jason's feedback) in a separate commit– 2e9c25c– in case anyone had begun reviewing the first commit.

jonbell-lot23:
Brian and I chatted and we're feeling good! I have some follow-on feedback I can handle in quick follow-on PRs.

markerikson:
About to look at the code. My main question after having watched the Looms is:

does the print statement timeline always show _all_ hits regardless of the current focus region?  or, if you select "Set Focus Region" for one of those points, will it recalculate and limit which points are actually shown in its timeline?  example: if there are 20 hits for a print statement, and I set point 6 as the focus region start, does it still show all 20, or does it limit to 15 total (the former 6...20, with the former 6 now being hit 1 because of the region)?

bvaughn:
The timeline has always shown all hits.

markerikson:
Hmm. Looks like it does narrow down the list of hits that are in the print statement timeline, which matches my existing mental model because we _do_ ask "how many times did this line get hit _in the focus region_".  So, per my example above, selecting point 6 and clicking "Set Focus Start" _will_ drop it from 20 shown hits to 15, with point 6 becoming the new point 1.

Given how the system works today, I think that's what i would expect.

(it would be _nice_ if we didn't have that annoying 200 hits limitation and we could always show _all_ the hits even when we're focused in, but oh well.)

markerikson (approved):

bvaughn:
Thanks Mark! Let's chat with @jonbell-lot23 and see if it makes sense to make some changes to the focus stuff like you mentioned (but as a follow up)

PR #8554, opened 2023-01-19
Use object and eval caches for step details

ryanjduffy:
The protocol expects clients to cache objects already returned in previous calls but this code wasn't using the caches already in place for protocol objects. Swapping both the object lookups and the expression eval greatly improves the UX and reduces UI crashes for that component caused by objects missing from the cache.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
@@ -1,8 +1,9 @@
-import { Object as ProtocolObject, createPauseResult } from "@replayio/protocol";
+import { Object as ProtocolObject } from "@replayio/protocol";
 import React, { useCallback, useContext, useEffect, useRef, useState } from "react";
 
 import { highlightNodes, unhighlightNode } from "devtools/client/inspector/markup/actions/markup";
 import { getObjectWithPreviewHelper } from "replay-next/src/suspense/ObjectPreviews";
+import { evaluateAsync } from "replay-next/src/suspense/PauseCache";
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
```diff
@@ -125,50 +119,66 @@ export function TestStepItem({ step, argString, index, id }: TestStepItemProps)
...
+              const cmdObject = await getObjectWithPreviewHelper(
+                client,
+                endPauseResult.pauseId,
+                cmdObjectId,
+                true
+              );
 
-            const nodeIds = subjects.filter(s => s?.preview?.node).map(s => s?.objectId!);
-            setSubjectNodePauseData({ pauseId: endPauseResult.pauseId, nodeIds });
+              const length: number | undefined = cmdObject?.preview?.properties?.find(
```
markerikson, line 139: :hand: Have to say this is pretty hard to read :)

If it at all helps, there's a `getJson` util I wrote that will recursively fetch contents of an object or array and return the final result as normal data.  It might be easier to use that to fetch `cmdObject` and then pull out values, if you want to try it

jaril (approved):

src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
ryanjduffy: Thanks for the pointer! I tried it but it was returning empty objects instead of the DOM node objects. I don't actually need the objects themselves here and and ids are sufficient but I take your point that the code was messy so I refactored it to be more straightforward.

PR #8565, opened 2023-01-21
Support jump to source for chai assertions

ryanjduffy:


src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -72,10 +72,8 @@ function useGetTestSections(
           // of the failed command but successful asserts with their start time
           if (s.error) {
             absoluteStartTime = absoluteEndTime - 1;
-            annotations.start = annotations.end;
```
ryanjduffy, line 75: We need the start annotation to be the actual start annotation so I've reverted this. It was added by #8263 but I'm not sure it's still required so I need to investigate that further before we land this.

ryanjduffy: I pushed 5ac9776 with some fixes for this so I think it's ready for review now

jaril (approved):

PR #8567, opened 2023-01-22
Add disabled print statement treatment

jonbell-lot23:
Old, where a disabled state just looked like floating in space:
image

New, where we use a pattern treatment to call out disabled state:
image

FAQ

Q: Do we use a pattern treatment like this anywhere else in the product?
A: Nope, this is our first time.

Q: Have we considered other approaches?
A: Yup, we explored colors. They don't work very well because we have white on grey textfields, surrounded by white. (or the reverse in dark theme). Using another shade of gray wouldn't have explained much, and colors such as red and yellow mean "warn" and "error," whereas things like blue and green often mean success or action. 

Q: Shouldn't we dim/disable the textfields themselves?
A: Maybe? I don't know if it's strictly necessary, but I could see it being useful. On the other hand, I could also see it getting in people's way when they're trying to add a conditional or otherwise manipulate the field. I'd propose trying this treatment without disabling the functionality.

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
+++ b/packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
@@ -4,12 +4,21 @@
   flex-direction: column;
   padding: 0.5rem;
   gap: 0.25rem;
-
   --badge-picker-button-size: 1.25rem;
   --badge-picker-icon-size: 1rem;
 }
 .PanelEnabled {
   background-color: var(--point-panel-background-color);
+}
+
+.PanelDisabled {
+  background: repeating-linear-gradient(
+    45deg,
+    var(--theme-base-100),
+    var(--theme-base-100) 5px,
+    var(--point-panel-background-color) 5px,
+    var(--point-panel-background-color) 10px
+  );
 }
 
 .Loader {
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.css
@@ -54,6 +54,16 @@
   border-top: 1px solid var(--theme-splitter-color);
 }
 
+.disabled {
+  background: repeating-linear-gradient(
+    45deg,
+    var(--theme-base-95),
+    var(--theme-base-95) 5px,
+    var(--theme-base-100) 5px,
+    var(--theme-base-100) 10px
+  );
+}
+
 .breakpoints-list .breakpoint:hover {
   background: var(--theme-toolbar-background-alt);
 }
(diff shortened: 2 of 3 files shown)


packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
```diff
@@ -4,14 +4,23 @@
...
   background-color: var(--point-panel-background-color);
 }
 
+.PanelDisabled {
+  background: repeating-linear-gradient(
+    45deg,
+    var(--theme-base-100),
+    var(--theme-base-100) 5px,
+    var(--point-panel-background-color) 5px,
+    var(--point-panel-background-color) 10px
```
bvaughn, line 20: This looks like a bad mix of CSS variables. Since `--point-panel-background-color` is detached from the `--theme-base-*` colors, it could change in a way that makes this not work. We should probably add a new `--point-panel-background-color*` for this.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.css
```diff
@@ -54,6 +54,16 @@
   border-top: 1px solid var(--theme-splitter-color);
 }
 
+.disabled {
+  background: repeating-linear-gradient(
+    45deg,
+    var(--theme-base-95),
+    var(--theme-base-95) 5px,
+    var(--theme-base-100) 5px,
+    var(--theme-base-100) 10px
```
bvaughn, line 63: This should probably use the same CSS variables as `--point-panel-background-color` and the new one, so they stay in sync?

bvaughn (approved):
Approved after the CSS variable change is made!

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.css
jonbell-lot23: You'd think so, but this is on the sidebar, with a smaller area. It didn't look right using an identical treatment, so the color was lightened slightly for readability.

packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
bvaughn: Is it important that we use 5px increments here rather than our normal 0.25rem increments? I'd prefer the latter if possible.

jonbell-lot23:
Hey gang! Closing this PR while we revise the design.

packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
jonbell-lot23: Moot point for now, since we're moving away from stripes!

PR #8571, opened 2023-01-23
Wait for the first LoadedRegions event before initializing devtools

hbenl:
This is a partial revert of #8451.
That PR was necessary because we couldn't rely on getting at least one `LoadedRegions` event from the backend (see [FE-1076](https://linear.app/replay/issue/FE-1076/session-is-created-but-sessionwaiter-never-resolves)) and then devtools would be stuck. But the PR also introduced a different problem: if [this check](https://github.com/replayio/devtools/blob/ba7e3ab36e7253cb92e91178a7470f5c35ea089d/src/ui/actions/timeline.ts#L125) was executed before we received a `LoadedRegions` event, devtools would use the end of the recording as the initial pause point.
The backend/protocol issue was fixed in replayio/backend#6871, so now we can wait for the first `LoadedRegions` event again and ensure we use the right initial pause point.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -160,6 +160,9 @@ class _ThreadFront {
   // Waiter which resolves when the debugger has loaded and we've warped to the endpoint.
   initializedWaiter = defer<void>();
 
+  // Waiter which resolves when there is at least one loading region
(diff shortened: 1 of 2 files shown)


bvaughn (approved):
Thanks for the extra context in the PR description.

PR #8578, opened 2023-01-24
FE-1150: making `chromium` events work

Domiii:
* Set `supportsEventTypes: true` for `chromium`
* Use new `Debugger.getAllEventHandlerCounts` protocol command ([BAC-2724](https://linear.app/replay/issue/BAC-2724/add-debuggergetalleventhandlercounts-to-protocol))
* Make sure to render `EventsList` correctly for `chromium`
* Depends on [this `BAC` PR](https://github.com/replayio/backend/pull/6951)
* Note: **Not all sessions have events enabled yet**. But we could probably flip the switch 🤔 
  * Here is a recording that has them: http://localhost:8080/recording/f6a88545-1f09-4d96-b54d-3f1f4e5f0788

For more details and explanations as to why this messy code is necessary, please consider the Linear issue:
https://linear.app/replay/issue/FE-1150/add-chromium-events-to-devtools

Short video: https://www.loom.com/share/656bf2aa796745479b1d1774990335a8


packages/shared/client/ReplayClient.ts
```diff
@@ -371,6 +371,14 @@ export class ReplayClient implements ReplayClientInterface {
     return countsObject;
   }
 
+  async getAllEventHandlerCounts(): Promise<Record<string, number>> {
+    const sessionId = this.getSessionIdThrows();
+    const { counts }: { counts: { type: string; count: number }[] } =
+      await client.Debugger.getAllEventHandlerCounts({}, sessionId);
```
bvaughn, line 377: `Debugger.getAllEventHandlerCounts` is not a function in the protocol package version installed in devtools, and it looks like we're depending on the latest published version of that package.

I think you need to publish an updated version of `@replayio/protocol` to NPM, then update `package.json` and `yarn.lock` in this repo to include that version as part of this PR.

Domiii: Yep. Working on that.
It is still in a `BAC` PR (https://github.com/replayio/backend/pull/6951), which is also failing checks /o\

bvaughn: Going to hold off on reviewing this until I can build and run it locally. Let me know when that's ready!

Domiii:
Finally ready for a review 👀 
@bvaughn @markerikson

packages/replay-next/components/console/filters/EventsList.tsx
```diff
@@ -84,7 +83,9 @@ function EventsListCategories({
           filterByText={filterByText}
         />
       ))}
-      <div className={styles.Header}>Other Events</div>
+      {commonEventCategories.length && otherEventCategories.length && (
+        <hr className="border-splitter" />
+      )}
```
bvaughn, line 88: Removing these category headers looks like a UI change that's maybe out of scope of the "Chromium events" change?

cc @jonbell-lot23

packages/replay-next/components/console/filters/EventType.tsx
```diff
@@ -18,8 +18,14 @@ export default function EventType({
 }) {
   const { eventTypesForDisplay: eventTypes, update } = useContext(ConsoleFiltersContext);
 
-  const checked = eventTypes[event.type] === true;
-  const toggle = () => update({ eventTypes: { [event.type]: !checked } });
+  const checked = eventTypes[event.rawEventTypes[0]] === true;
```
bvaughn, line 21: I wonder if there's an easier way to model this that doesn't involve toggling _all_ event types for a category?

packages/replay-next/src/suspense/EventsCache.ts
```diff
@@ -160,3 +160,239 @@ export function getEventAsync(eventType: EventHandlerType, point: ExecutionPoint
...
+}
+
+async function countEvents(eventCountsRaw: Record<string, number>) {
+  const recordingTarget = await getRecordingTarget();
+  const targetCountEvents = CountEvents[recordingTarget];
+  if (!targetCountEvents) {
+    if (!isEmpty(eventCountsRaw)) {
+      console.error(
+        `Internal Error: recordingTarget "${recordingTarget}" has events but is missing CountEvents function`
+      );
```
bvaughn, line 200: This doesn't look right. Node recordings are configured with `supportsEventTypes: true` but this new implementation would always log an error and count no events.

bvaughn (commented):
This is a pretty big chunk of new functionality to have no new tests 😅 

Can you add an e2e test or two for this? Maybe one that explicitly tests Gecko events and one that explicitly tests Chromium events?

packages/replay-next/components/console/filters/EventsList.tsx
jonbell-lot23: Thanks for letting me know! I need to track changes like this, so it's very helpful to be pinged.

In this case, we talked about making that change many weeks ago and I thought it was in prod already. I agree it's not related to this ticket, but when we're ready for it to go live, it has a thumbs up from design.

Domiii: I reached out to Jon back then, and he remembers after all this time ♥

packages/replay-next/src/suspense/EventsCache.ts
Domiii: Uhm... Why do Node recordings have `supportsEventTypes`?
Also what do you recommend, remove the `if`?

packages/replay-next/src/constants.ts
```diff
@@ -1,227 +1,960 @@
...
+          },
+          {
+            type: "progress",
+            label: "progress",
+          },
+          {
+            type: "timeout",
+            label: "timeout",
+          },
+        ],
```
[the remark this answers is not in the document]
Domiii: It's not ideal. I agree.

packages/shared/client/ReplayClient.ts
```diff
@@ -380,6 +380,14 @@ export class ReplayClient implements ReplayClientInterface {
     return Object.fromEntries(counts.map(({ type, count }) => [type, count]));
   }
 
+  async getAllEventHandlerCounts(range: PointRange | null): Promise<Record<string, number>> {
+    const sessionId = this.getSessionIdThrows();
+    const { counts }: { counts: { type: string; count: number }[] } =
```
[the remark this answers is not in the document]
Domiii: It was a copy-and-paste from `getEventCountForTypes`. I removed it in both.

packages/replay-next/components/console/filters/EventType.tsx
Domiii: `rawEventTypes` is what the `runtime` and `backend` knows, and thus, that is what needs to be queried for. Usually, you won't have too many of these in one "normalized event type" though. It will be negligible for performance.

As we have previously discussed, the code for this ideally should at least be isolated and modularized. On the other hand, chromium should also not be crashing all the time, and is that latter part that is a bit more anxiety inducing for me for now 😅

What is your concern here in particular?

Domiii:
@bvaughn 
> This is a pretty big chunk of new functionality to have no new tests 😅
> 
> Can you add an e2e test or two for this? Maybe one that explicitly tests Gecko events and one that explicitly tests Chromium events?

Sounds like a good idea. I see that `packages/e2e-tests/tests/logpoints-03.test.ts` and `packages/e2e-tests/tests/stepping-05.test.ts` already do some testing in this regard for gecko (I guess?).
How do I make an e2e test target a specific build?

packages/replay-next/components/console/filters/EventType.tsx
bvaughn: I just meant there's maybe a simpler way to _model_ which categories are checked then adding an attribute to every event in the array.

packages/replay-next/src/suspense/EventsCache.ts
bvaughn: https://github.com/replayio/devtools/blob/b1a5e77285d64c269e17fe3f5e81810a0baae83a/packages/protocol/thread/thread.ts#L249-L257

bvaughn:
> How do I add an e2e test to target chromium?

Make a recording with Chromium and add a test here: https://github.com/replayio/devtools/tree/main/packages/e2e-tests/tests

I think the README has some info that might be helpful
https://github.com/replayio/devtools/tree/main/packages/e2e-tests

packages/replay-next/src/suspense/EventsCache.ts
Domiii: Should I remove the `if`?

bvaughn: If you did, it would break with your current implementation– wouldn't it? (There is no `CountEvents["node"]`.)

That seems like an oversight or a disagreement at least in my understanding of what Node was supposed to support. Maybe my understanding was just wrong though and the "capabilities" code had the wrong default on the client. Could totally be that! In which case, the thing to do would be to edit the `thread/thread.ts` code to match.

Domiii: I think, you added those flags [here](https://github.com/replayio/devtools/pull/7713/files#diff-6e85b32eceb414e3df67e264b8594beb834c30fd38ce5b8ffa774dcd08dabc9eR259) 6 months ago?

From a runtime perspective, I don't think we would expect Node events to work. The `OnEvent` function is also not being used [here](https://github.com/replayio/node/blob/1f7d292d2cc117103eb1ca4246d870b0e7a0cf75/deps/v8/src/api/api.cc#L10893).
So, I'll just go ahead and set it to `false` for Node.

Domiii:
@jasonLaster, currently e2e tests only allow for one "golden recording" per test. Can we change it s.t. `examples.json` can have more than one recording (e.g. one for gecko, one for chromium)?

bvaughn:
> Currently, e2e tests only allow for one "golden recording" per test. Can we change it s.t. `examples.json` can have more than one recording (e.g. one for gecko, one for chromium)?

We _could_ change this, certainly. The logic for matching a test to its recording is here:

https://github.com/replayio/devtools/blob/39b5f9ac86028c6c9050cbe94ae30475e428bcf8/packages/e2e-tests/helpers/index.ts#L40-L57

And each test calls it on start, like so:
https://github.com/replayio/devtools/blob/39b5f9ac86028c6c9050cbe94ae30475e428bcf8/packages/e2e-tests/jail/cypress-01.test.ts#L14-L17

That log _could_ be changed to add support for Chromium vs Gecko...how would you want that to work?

markerikson:
Latest status update:

- Merged down from `main` to pick up all the recent changes
- Made a copy of `doc_events.html` -> `doc_events_chromium.html`. Successfully created a new "golden recording" for that test file, and confirmed that loading up the recording in this branch shows events. 
- _However_, there are differences in the object preview formatting for events between Gecko and Chromium, due to a combo of many fields being getters (and getters mostly being disabled in Chromium atm), and the event object prototype chain.

I've pushed the test code changes up. Domi is going to work on the event preview behavior.  Once he's got a new Chromium build with those changes ready, I can recreate the "golden recording" and try running the test again.

I tried making a copy and recording for `doc_minified.html` for use with the `stepping-05` test, but that recording exploded with command errors as soon as I tried to view it, and so did my next two recording attempts.

markerikson:
Ended up copying the changes over to a new PR, #9048 . Thanks for all the work putting this together!

PR #8581, opened 2023-01-24
Move test suites e2e tests into their own workflow job

ryanjduffy:
Re-enables the cypress e2e test in a separate GH workflow job (that always passes via `|| exit 0`) so we can start capturing replays of it and catch issues with the cypress panel without blocking other devtools PRs with a ❌ .


ryanjduffy:
@bvaughn + @markerikson - Thoughts on this approach for adding a "jail" for the cypress e2e test so we can start getting replays in for it to identify why it's flaky without blocking other work?

packages/e2e-tests/testsuites/cypress-01.test.ts
```diff
@@ -15,7 +15,7 @@ import {
 
 const url = "cypress/basic";
 
-test.skip(`cypress-01: Test basic cypress reporter functionality`, async ({ page }) => {
+test(`cypress-01: Test basic cypress reporter functionality`, async ({ page }) => {
```
bvaughn, line 18: Could we move this file into a folder with a name that's more meaningful than "testsuites"? This is pretty ambiguous:
```
e2e-tests
├── tests
└── testsuites
```

Maybe even something like "jail"

bvaughn (commented):
Don't have a strong opinion other than folder name

bvaughn (approved):
<dog science>

ryanjduffy:
@markerikson , @bvaughn , @hbenl - I'm merging this now but wanted to specifically point out that if anything in "Jail" fails, it'll still show in the bot comment even though the workflow won't fail. Seemed like that could be a surprise for folks so wanted to call it out beforehand.

PR #8583, opened 2023-01-24
Jump to the function location for info sidebar "Click" events

markerikson:
This PR:

- Implements complex introspection logic to determine the first plain JS or React event listener that ran in association with a "Click" event listed in our info panel sidebar, and opens the source location of that event listener function

Loom demo: https://www.loom.com/share/eee1c44b28414155ad034a9a5815088f

### DIFF
--- a/src/ui/components/Events/Event.tsx
+++ b/src/ui/components/Events/Event.tsx
@@ -1,12 +1,139 @@
+import {
+  ExecutionPoint,
+  Location,
+  Node,
+  Object as ProtocolObject,
+  MouseEvent as ReplayMouseEvent,
+  Value,
+} from "@replayio/protocol";
 import classNames from "classnames";
 import React, { KeyboardEvent, MouseEvent, ReactNode } from "react";
 
+import { selectLocation } from "devtools/client/debugger/src/actions/sources/select";
+import { getThreadContext } from "devtools/client/debugger/src/reducers/pause";
+import type { ThreadFront as TF } from "protocol/thread";
+import { createGenericCache } from "replay-next/src/suspense/createGenericCache";
+import { MAPPER as EVENTS_MAPPER, EventLog } from "replay-next/src/suspense/EventsCache";
+import { getFramesAsync } from "replay-next/src/suspense/FrameCache";
+import { getObjectWithPreviewHelper } from "replay-next/src/suspense/ObjectPreviews";
+import { getPauseIdAsync } from "replay-next/src/suspense/PauseCache";
+import { compareExecutionPoints } from "replay-next/src/utils/time";
+import { ReplayClientInterface } from "shared/client/types";
+import type { UIThunkAction } from "ui/actions";
+import {
+  FunctionWithPreview,
+  REACT_16_EVENT_LISTENER_PROP_KEY,
+  REACT_17_18_EVENT_LISTENER_PROP_KEY,
+  formatEventListener,
+} from "ui/actions/event-listeners";
 import useEventContextMenu from "ui/components/Events/useEventContextMenu";
+import { getLoadedRegions } from "ui/reducers/app";
+import { getPreferredLocation, getSourceDetailsEntities } from "ui/reducers/sources";
+import { useAppDispatch } from "ui/setup/hooks";
+import { UIState } from "ui/state";
 import { ReplayEvent } from "ui/state/app";
 import { getFormattedTime } from "ui/utils/timeline";
 
 import MaterialIcon from "../shared/MaterialIcon";
 import { getReplayEvent } from "./eventKinds";
+
+const { getValueAsync: getNextClickEventAsync } = createGenericCache<
+  [replayClient: ReplayClientInterface, point: ExecutionPoint, endTime: number],
+  EventLog | undefined
+>(
+  "clickEventCache",
+  async (replayClient, point, endTime) => {
+    const pointNearEndTime = await replayClient.getPointNearTime(endTime);
+
+    const entryPoints = await replayClient.runAnalysis<EventLog>({
+      effectful: false,
+      eventHandlerEntryPoints: [{ eventType: "event.mouse.click" }],
+      mapper: EVENTS_MAPPER,
+      range: {
+        begin: point,
+        end: pointNearEndTime.point,
+      },
+    });
+
+    entryPoints.sort((a, b) => compareExecutionPoints(a.point, b.point));
+    return entryPoints[0];
+  },
+  (replayClient, point) => point
+);
+
+const { getValueAsync: getClickEventListenerLocationAsync } = createGenericCache<
+  [
+    ThreadFront: typeof TF,
+    replayClient: ReplayClientInterface,
+    getState: () => UIState,
+    pauseId: string
+  ],
+  Location | undefined
+>(
+  "clickEventListenerLocationCache",
+  async (ThreadFront, replayClient, getState, pauseId) => {
+    const stackFrames = await getFramesAsync(replayClient, pauseId);
+    if (!stackFrames) {
+      return;
+    }
+    const topFrame = stackFrames[0];
+    const { frameId } = topFrame;
+
+    await ThreadFront.ensureAllSources();
+
+    const state = getState();
+
+    // Introspect the click event's target DOM node, and find the nearest
+    // React `onClick` event handler if any exists.
+    const res = await ThreadFront.evaluateNew({
+      replayClient,
+      pauseId,
+      text: FIND_MOUSE_EVENT_TARGET_AND_REACT_CLICK_HANDLER,
+      frameId,
+      pure: false,
+    });
+
+    let sourceLocation: Location | undefined;
+
+    if (res.returned?.object) {
+      const preview = await getObjectWithPreviewHelper(replayClient, pauseId, res.returned.object);
+
+      // The evaluation may have found an `onClick` function somewhere.
+      const onClickProp = preview?.preview?.properties?.find(p => p.name === "onClick");
+
+      const sourcesById = getSourceDetailsEntities(state);
+
+      if (onClickProp) {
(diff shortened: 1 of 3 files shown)


src/ui/components/Events/Event.tsx
```diff
@@ -1,13 +1,144 @@
...
+  [replayClient: ReplayClientInterface, point: ExecutionPoint, time: number],
+  EventLog | undefined
+>(
+  "clickEventCache",
+  async (replayClient, point, time) => {
+    // Actual browser click events get recorded a fraction later then the
+    // "mouse events" used by the sidebar.
+    // Look for the next click event within a short timeframe after the "mouse event".
+    // Yes, this is hacky, but it does seem pretty consistent.
+    const arbitraryEndTime = time + 200;
```
bvaughn, line 49: This makes me nervous. I wonder how many times we may get false positives from this? How does this work for double clicks?

cc @jasonLaster @bhackett1024  in case they have more context about edge cases here.

src/ui/components/Events/Event.tsx
```diff
@@ -31,15 +162,142 @@ export const getEventLabel = (event: ReplayEvent) => {
...
+  // Outer body runs in scope of the "current" click handler.
+  // Grab the click handler's arguments.
+  const args = [...arguments];
+  
+  (function findEventTargetAndClickHandler() {
+    // One of the args should be a mouse event
+    const foundMouseEvent  = args.find(a => typeof a === "object" && a instanceof MouseEvent);
+
+    if (foundMouseEvent) {
+      // We _could_ probably just return the onClick function or undefined here
```
bvaughn, line 175: Not sure I understand this comment. Are you saying we could get rid of the wrapper `{ target: ... }` object?

src/ui/components/Events/Event.tsx
```diff
@@ -31,15 +162,142 @@ export const getEventLabel = (event: ReplayEvent) => {
...
+            key.startsWith("${REACT_16_EVENT_LISTENER_PROP_KEY}") || 
+            key.startsWith("${REACT_17_18_EVENT_LISTENER_PROP_KEY}")
+          )
+        });
+
+        if (reactPropsKey) {
+          const props = currentNode[reactPropsKey];
+          if (props.onClick) {
+            res.onClick = props.onClick;
+            res.handlerNode = currentNode;
```
bvaughn, line 197: This matches props that are literally named `onClick` ?

I think this means it will always find the one attached to the DOM element? What location would it return for the following scenario?
```jsx
function SomeGenericComponent({ handleClick }) {
  return (
    <div onClick={handleClick}>
      {/* ... */}
    </div>
  );
}
```

What about a scenario like this?
```jsx
function noop() {}

<div onClick={outerHandler}>
  <div onClick={enabled ? innerHandler : noop}>
    <button>Click target</button>
  </div>
</div>
```

I think we'd find `noop` and not `outerHandler`? (The above example is stupid but I think this pattern sometimes exists in actual apps– just further apart in the React tree.)

src/ui/components/Events/Event.tsx
```diff
@@ -31,15 +162,142 @@ export const getEventLabel = (event: ReplayEvent) => {
...
+        const reactPropsKey = keys.find(key => {
+          return (
+            key.startsWith("${REACT_16_EVENT_LISTENER_PROP_KEY}") || 
+            key.startsWith("${REACT_17_18_EVENT_LISTENER_PROP_KEY}")
+          )
+        });
+
+        if (reactPropsKey) {
+          const props = currentNode[reactPropsKey];
+          if (props.onClick) {
```
bvaughn, line 195: Do we care about `onClickCapture` ?

src/ui/components/Events/Event.tsx
```diff
@@ -31,15 +162,142 @@ export const getEventLabel = (event: ReplayEvent) => {
...
+
+      // If we did have a click event, timewarp to that click's point
+      onSeek(nextClickEvent.point, nextClickEvent.time);
+
+      const sourceLocation = await getClickEventListenerLocationAsync(
+        ThreadFront,
+        replayClient,
+        getState,
+        pauseId
+      );
```
bvaughn, line 268: Should we get this first, and only _seek_ if we can find a location too? Otherwise the timeline would jump to...nothing?

bvaughn (commented):
This is a lot of new code. Can you add an e2e test for it? Ideally maybe one for React and one for non-React? 😁

src/ui/components/Events/Event.tsx
markerikson: Yeah. I originally was returning an object with 3-5 fields here for debugging purposes (target node, possible `onClick` prop, possible ancestor node, `string[]` serialized ancestor tree, etc).  In the end the only field the outer logic actually ends up using is the potential `onClick` prop, since we're getting the vanilla JS function location from the execution point stack frame.

src/ui/components/Events/Event.tsx
markerikson: Yes, literally `onClick`, because that's what the `props` object ends up as: `{onClick: someFunctionHere}`.

src/ui/components/Events/Event.tsx
markerikson: I don't know how the runtime specifically determines which events it records for the `Session.findMouseEvents` protocol request.  I think @Domiii recently implemented that for Chromium and might have more context on how that works internally.

That said, I checked this across several recordings and am consistently seeing _some_ `"click"` event listener entry show up very shortly after the timestamp for the `findMouseEvents` entry.  

Like I said, it's not ideal, but it's a sufficiently good heuristic for now.

src/ui/components/Events/Event.tsx
markerikson: Nope. Not trying to deal with that for now - it's extremely rarely used in my experience.

src/ui/components/Events/Event.tsx
markerikson: Not sure I understand this question.

Previously, clicking any of the info sidebar list items unconditionally did a `seek()` to that execution point, period.  As I've found, that time is always offset from the actual browser event listener handling.  So, it gets you the general visuals of what's in the page and where the click event occurred, but not the exact line of code execution point.

Now, it does the seek to that initial execution point, but then _tries_ to refine it to the time of the click event listener if possible.

bvaughn: That's fair!

bvaughn (approved):
You have more context than I do at this point :grin:

I think we should add some e2e tests for this, but otherwise it seems good to move forward with

jasonLaster:
Excited to see this land! Would love to discuss next steps after this lands

src/ui/components/Events/Event.tsx
Domiii: @markerikson I am surprised that this works, since `getPointNearTime` throws an error if there is no active region at exactly that point in time :smile:

If it does not error out, the `runAnalysis` part should probably work just fine on gecko.
However, on chromium, event names are a total mess, so we will have to talk, or I have to just write you a function that you can use to get the right events to query for.
Please keep that in mind when my `events` PR finally hopefully lands before/by next week :eyes:

markerikson: Hmm. Yeah, I've only actually tested on this relatively small recordings, and both made with Firefox.  Those are good points!  Let me think about this a bit more.

Domiii: @markerikson It would theoretically make sense for the mouse click event to be associated with its "raw event" somehow. The "raw event", as you might have already figured out, records both, begin and end points, so as to allow searching for the first user frame between the two.

markerikson: on the flip side: 

- why do we even _have_ `Session.findMouseEvents`?  What purpose does that API server?
- Why are these events labeled as "Click" in the UI, but with a type of `"mousedown"`?
- Why does there seem to be a different number of results for those `findMouseEvents` entries than there are for "real" `"click"` or `"mousedown"` events?

src/ui/components/Events/Event.tsx
bvaughn: Note to self:
What happens if there's an imperative handler that uses capture higher up in the DOM? and what if it stops prop? Will we incorrectly tag a nested (declarative) handler in that case?

src/ui/components/Events/Event.tsx
Domiii: @markerikson 
1. The three `Session.find*Events` calls are probably from an earlier time, and were thought of and added over time (I am guessing). The data points collected for them have more information attached to them. However, they might be somewhat legacy, since we are likely able to use the (very different) data we use for the `Debugger.find*Event*Count*` calls to get mostly the same data that we desire from `Session.find*Events`. Cleaning that up and reasoning about the different user scenarios, is probably going to be a low priority for now.
2. I am not sure, but the entire mixing-and-matching algorithm generally gets a bit hacky. The way it works in `gecko` is also different from `chromium`, and from what I saw, they both have different types of inaccuracies.
3. The `findMouseEvents` data points are collected upon event dispatch inside the render engine, so they should always register any "click". The `Debugger` events, on the other hand, should only report on events that have a user-space JS listener associated with them (but possibly with false positives). Furthermore, the `Debugger` events are collected using mechanisms but ultimately come down to `gecko`'s and `chromium`'s "Event Listener Breakpoint" feature. In `gecko`, if you enable logging for all "Event Listener Breakpoints", you should see the same events as in Replay. If that is not the case, feel free to file a bug. (I don't think anyone has bothered to look into this yet.) I also detected some inconsistencies, e.g. [here](https://linear.app/replay/issue/RUN-1061#comment-a6859804). In `chromium` it might be even worse because their "Event Listener Breakpoints" are not as well written, not well maintained, and they also don't have the "log-only" feature that `gecko` has.

PR #8590, opened 2023-01-25
Use the logpoint's condition in the logpoint panel (FE-1165)

hbenl:


packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -84,7 +84,7 @@ function PointPanel(props: ExternalProps) {
   const [hitPoints, hitPointStatus] = getHitPointsForLocationSuspense(
     client,
     point.location,
-    null,
+    point.condition,
```
bvaughn, line 87: Aw, hell

bvaughn (approved):
Do we have another instance of this that we should fix here?
https://github.com/replayio/devtools/blob/71733921a6c32bad2c10753911b6eb8907e999da/src/ui/components/Timeline/PreviewMarkers.tsx#L41-L53

hbenl:
> Do we have another instance of this that we should fix here?

Good question - in general we don't have a logpoint there to get the condition from (since this is generally used to preview hitpoints for any source line as soon as it's hovered). We could add logic to detect if there is a logpoint on that line and use that logpoint's condition in that case. However, I'd prefer to keep it as is, so the global timeline will always show all hits for the hovered line.

bvaughn:
Makes sense 👍🏼

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
jasonLaster: 😂

bvaughn:
1. Edit: `tests/source-and-console.ts`
1. Focus "_should support conditional log points_" test
1. Run `yarn test:update-fixture-data`

bvaughn:
I think this PR broke one of our Playwright tests:
* tests/source-and-console.ts:153:1 › should support conditional log points

Failing on main and other branches now.

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
bvaughn: Didn't think about this at the time, but this change means that editing a point's condition text (in the point panel overlay) will cause the point panel to re-suspend.

(Looking into this as part of FE-1138)

PR #8597, opened 2023-01-26
Use original variables in autocomplete (FE-1147)

hbenl:


packages/e2e-tests/helpers/lexical.ts
```diff
@@ -24,11 +24,11 @@ export async function focus(page: Page, selector: string) {
 }
 
 export async function hideTypeAheadSuggestions(page: Page, selector: string) {
-  const list = page.locator('[data-test-name="TypeAheadList"]');
+  const list = page.locator('[data-test-name="CodeTypeAhead"]');
```
bvaughn, line 27: Hm 😆 Guess I hadn't tested this helper yet.

packages/replay-next/playwright/tests/utils/general.ts
```diff
@@ -160,6 +160,7 @@ export async function takeScreenshot(page: Page, locator: Locator, name: string)
 
   // Make sure any suspended components finish loading data before taking the screenshot.
   await awaitNoLoaders(page, locator);
+  await delay(1);
```
bvaughn, line 163: What's this about?

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -166,6 +168,16 @@ function PointPanelWithHitPoints({
     };
   }
 
+  let source = getSource(client, point.location.sourceId);
+  if (source?.kind === "prettyPrinted") {
+    assert(
+      source.generatedSourceIds,
```
bvaughn, line 174: This just asserts that there's an array, right? Not that it's non-empty?

packages/replay-next/components/lexical/plugins/code-completion/findMatches.ts
```diff
@@ -44,16 +51,20 @@ function fetchQueryData(
...
+  pauseId: PauseId | null,
+  useOriginalVariables: boolean
 ): {
   properties: WeightedProperty[] | null;
   scopes: Scope[] | null;
 } {
   let properties: WeightedProperty[] | null = null;
 
-  let scopes = null;
+  let generatedScopes: Scope[] | null = null;
```
bvaughn, line 62: I'm curious– why do we use _generated_ scopes for global/Window properties?

bvaughn (approved):
Thanks for adding a new log point test!

I wonder if it would be worth also adding one for the console's input?

packages/replay-next/playwright/tests/utils/general.ts
hbenl: I noticed that sometimes dark screenshots were flaky when their light counterparts weren't. Since the dark screenshots are taken first I thought that maybe a little delay (to let the UI settle(?)) would help and it seems it does.

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
hbenl: Yeah, that was more about appeasing typescript.

packages/replay-next/components/lexical/plugins/code-completion/findMatches.ts
hbenl: I noticed that the original scopes that we get from the backend don't contain a global scope(!).

hbenl:
> I wonder if it would be worth also adding one for the console's input?

Good idea, I'll do that.

packages/replay-next/playwright/tests/utils/general.ts
bvaughn: Whaaaat the fuck 😆 

Mind adding an inline comment about this? 😁

packages/replay-next/components/lexical/plugins/code-completion/findMatches.ts
bvaughn: Interesting! I guess that's why we were using generated scopes for this before. I just hadn't thought about it.

PR #8598, opened 2023-01-26
Change how inline Array previews render non-numeric properties

bvaughn:
Given the following code, here's how browsers each render the array as an inline value:
```js
const arrayWithExtraProperties = [1, 2, 3];
arrayWithExtraProperties.test = "FE-1164";
```

| Browser | Rendering
| :--- | :---
| Chrome | Screen Shot 2023-01-26 at 10 25 48 AM
| Firefox | Screen Shot 2023-01-26 at 10 25 58 AM
| Safari | Screen Shot 2023-01-26 at 10 26 04 AM

Compare the above to how Replay renders this value (currently, on production):

Screen Shot 2023-01-26 at 10 29 35 AM

This branch changes the rendering to more closely mirror what Chrome does– displaying the property _name_ for values with non-numeric property names:
Screen Shot 2023-01-26 at 10 33 24 AM

cc @Andarist


markerikson (approved):

bvaughn:
Going to update our screenshot tests to cover this case as well before landing.

PR #8605, opened 2023-01-26
Add a checkbox flag for tracking recording assets in the database.

cweld510:


cweld510:
I opened this yesterday, but then got pulled into other work; will return to it at some point today to figure out what's going wrong with CI

bvaughn (approved):

PR #8607, opened 2023-01-26
Treat jQuery wrappers as arrays

ryanjduffy:
Cypress uses jQuery to query for elements and returns that wrapped object as the `subject` for assertions which we display in the Step Details pane. It's currently presented as an object which, while true, is a bit unexpected for the user who doesn't care about jQuery.

The proposed change may be (is?) controversial as we likely don't want to blindly assume `className`'s for third party libraries but opening up for discussion. This isn't a big UX win so fine to ignore it if this doesn't work and an alternative isn't obvious.

Before:
image
image

After:
image
image

### DIFF
--- a/packages/replay-next/src/utils/protocol.ts
+++ b/packages/replay-next/src/utils/protocol.ts
@@ -179,6 +179,7 @@ export function protocolValueToClientValue(
         case "Uint8ClampedArray":
         case "Uint16Array":
         case "Uint32Array":
+        case "jQuery.fn.init":
           type = "array";
           break;
         case "Date":


jasonLaster (approved):
Nice idea!

bvaughn:
> The proposed change may be (is?) controversial as we likely don't want to blindly assume className's for third party libraries but opening up for discussion

Could we hold off landing this until Monday. I'm on PTO but I'd like to think through the implications.

ryanjduffy:
@bvaughn - Let me know what you think about this!

bvaughn:
It's in my queue. I'm still trying to catch up from being gone on Friday.

bvaughn:
Are the before/after screenshots messed up?

bvaughn:
I don't feel great about this change. I'm not familiar with what `"jQuery.fn.init"` is. (It's been ages since I've used anything jQuery.) But a quick Google search of it turns up results like [this](https://stackoverflow.com/questions/4754560/help-understanding-jquerys-jquery-fn-init-why-is-init-in-fn) and [this](https://blog.quickui.org/2012/06/07/jquery-fn-init/) that suggest it may also contain other value types– like functions.

I personally prefer we don't make this change.

ryanjduffy:
Is there a change to treat things that are "Array-like" like arrays that would be better?

bvaughn:
Haven't thought deeply about it, but initially– not necessarily, considering we don't necessarily have any preview information loaded in the object at the point where we would need to decide what it was– only the class name.

> This isn't a big UX win

Given this, I'd suggest we just pass on it.

PR #8608, opened 2023-01-26
Improve Replay Info

jonbell-lot23:
* Tightened padding
* Removed redundant timestamp
* Tightened security copy
* Removed blue icon highlights on hover
* Added underline to URL hover
* We no longer show Events if it's empty 🎉 

image

Stay tuned for another iteration on Replay Info that will look like this:

image


src/ui/components/Events/Events.tsx
```diff
@@ -23,6 +23,7 @@ function CurrentTimeLine({ isActive }: { isActive: boolean }) {
 
 function Events({ currentTime, events, executionPoint, seek }: PropsFromRedux) {
   const { userSettings } = hooks.useGetUserSettings();
+  const [items, setItems] = useState([]);
```
bvaughn, line 26: I don't understand why these were aded. Doesn't look like anything is referencing `items` or `setItems`?

jonbell-lot23: This PR is old, sorry about that.

(And yes, those lines were dumb)

PR #8610, opened 2023-01-27
Improve visuals in Replay Info pane

jonbell-lot23:
**Changes**

* Tightened padding
* Removed redundant timestamp
* Tightened security copy
* Removed blue icon highlights on hover
* Added underline to URL hover
* We no longer show Events if it's empty 🎉
* New icons

**Notes**

1. I meant to make two PRs, one for UX and the other for visual, but this combines them. Here's [the other PR](https://github.com/replayio/devtools/pull/8608), which I've closed.
2. We use a *lot* of legacy Tailwind in these files. We will be using Tailwind less (and removing old Tailwind) but for this ticket I'm hoping to focus on the visual updates, then I can come back around in Tailwind removal tickets and clean things more.
3. There are likely to be a ton of improvements requested. Thank you for helping me learn! 

--

Old:
image

New:
image


jonbell-lot23:
I am marking this as ready for review, but I'm expecting a fair amount of feedback before it goes live

src/ui/components/SidePanel.tsx
```diff
@@ -100,24 +102,25 @@ export default function SidePanel() {
...
+  items.push({
+    header: "Info",
+    buttons: resolveRecording ? <StatusDropdown /> : null,
+    className: "replay-info",
+    component: <ReplayInfo />,
+    opened: !replayInfoCollapsed,
+    onToggle: () => setReplayInfoCollapsed(!replayInfoCollapsed),
+  });
+
+  if (events.length > 0) {
```
bvaughn, line 114: There is no `events` variable in scope here. The TypeScript error is because of this.

src/ui/components/Events/Events.tsx
```diff
@@ -33,26 +32,30 @@ function Events({ currentTime, events, executionPoint, seek }: PropsFromRedux) {
...
-    </div>
-  );
+          );
+        })}
+        <CurrentTimeLine isActive={currentEventIndex === events.length && !!events.length} />
+      </div>
+    );
+  } else {
+    return null;
+  }
```
bvaughn, line 58: Super tiny nit that you should probably ignore– but I think we generally return null first, e.g.

```js
if (someThing) {
  return null;
}

// return actual view
```

Either way is totally fine though. This way just reads a little clearer to me because returning null at the end can be easy to miss.

src/ui/components/SidePanel.tsx
```diff
@@ -96,28 +97,30 @@ export default function SidePanel() {
...
+  items.push({
+    header: "Info",
+    buttons: resolveRecording ? <StatusDropdown /> : null,
+    className: "replay-info",
+    component: <ReplayInfo />,
+    opened: !replayInfoCollapsed,
+    onToggle: () => setReplayInfoCollapsed(!replayInfoCollapsed),
+  });
+
+  if (events && Array.isArray(events) && events.length > 0) {
```
bvaughn, line 114: I _think_ `getFlatEvents` will _always_ return `ReplayEvent[]` – is there any reason we're checking for undefined and is-array here?

bvaughn (approved):
Looks okay to me

src/ui/components/Events/Events.tsx
jonbell-lot23: Thanks! No nit is too small for me, I appreciate you letting me know.

src/ui/components/SidePanel.tsx
jonbell-lot23: Short answer: I was flailing and I will revert this.

Long answer: when you called out the Typescript error I was trying to figure out how it snuck in. I think it was because Ryan also committed code at the same time and I didn't see that the merge changed stuff. So this was me troubleshooting, and I thought it'd be fine to keep in. But happy to pull it back out.

PR #8618, opened 2023-01-27
Improve cypress network event formatting

ryanjduffy:
* Adds some spacing before network and navigation events to align them with the step text so its easier to scan
* Adds a click handler to navigation events so the user can seek to that time
* Adds a status circle for network events for cypress parity

Before:
image

After:
image


jasonLaster:
@ryanjduffy looks much nicer!

jaril (approved):

PR #8621, opened 2023-01-27
Add missing steps warning to test panel

ryanjduffy:
* Adds warning with docs when a cypress replay is missing test steps
* Replaces the `semver` dependency with an inline util version (cc: @markerikson)
* Adds `useTestInfo` hook to start to consolidate test suite-related data / logic
* Refactors `TestSuitesPanel` out of `SidePanel` into its own file

image


jonbell-lot23:
Quick note: any time we use a background or foreground colour, please make sure not to hard-code anything or else it looks broken in dark theme. Also please tap me on the shoulder so I can make sure my mockups are reflecting the code.

PR #8632, opened 2023-01-30
Improve handling of Session.createPause failures (FE-1021)

hbenl:
- `TimelineContextAdapter`: leave fetching the `pauseId` to the `seek()` thunk, which will also handle any failures
- handle `Session.createPause` failures in `LogPointPanel`
- show an error message in the Frames panel when `Session.createPause` fails
- ensure that the `ErrorBoundary` of `ConsoleInput` is reset when seeking to a different point, otherwise the console would remain disabled for the entire session after a `Session.createPause` failure
- fixed a bug in `TerminalContextAdapter` that showed up in e2e tests: for some reason `setMessages()` inside `startTransition()` in `addMessage()` was called repeatedly for one of the messages, hence the id for that message changed repeatedly, hence the evaluation was started repeatedly and its result was never displayed. See https://app.replay.io/recording/replay-of-localhost8080--6e9e7698-de48-4204-b621-b2fa8a75d2cf

### DIFF
--- a/packages/replay-next/components/ErrorBoundary.tsx
+++ b/packages/replay-next/components/ErrorBoundary.tsx
@@ -1,20 +1,46 @@
 import { captureException } from "@sentry/browser";
-import React, { Component, PropsWithChildren, ReactNode, createElement } from "react";
+import React, { Component, PropsWithChildren, ReactNode } from "react";
 
 import styles from "./ErrorBoundary.module.css";
 
-type ErrorBoundaryState = { error: Error | null };
+type ErrorBoundaryState = {
+  error: Error | null;
+  erroredOnResetKey: string | number | undefined | null;
+};
 
 type ErrorBoundaryProps = PropsWithChildren<{
   fallback?: ReactNode;
   fallbackClassName?: string;
+  /**
+   * If `resetKey` changes after the ErrorBoundary caught an error, it will reset its state.
+   * Use `resetKey` instead of `key` if you don't want the child components to be recreated
+   * (and hence lose their state) every time the key changes.
+   */
+  resetKey?: string | number;
 }>;
 
 export default class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
-  state: ErrorBoundaryState = { error: null };
+  state: ErrorBoundaryState = { error: null, erroredOnResetKey: null };
 
   static getDerivedStateFromError(error: Error): ErrorBoundaryState {
-    return { error };
+    return { error, erroredOnResetKey: null };
+  }
+
+  static getDerivedStateFromProps(
+    props: ErrorBoundaryProps,
+    state: ErrorBoundaryState
+  ): Partial<ErrorBoundaryState> | null {
+    if (state.error) {
+      if (state.erroredOnResetKey === null) {
+        // we just caught the error and need to save the current resetKey
+        return { erroredOnResetKey: props.resetKey };
+      }
+      if (props.resetKey !== state.erroredOnResetKey) {
+        // the resetKey changed after an error was caught so we need to reset the state
+        return { error: null, erroredOnResetKey: null };
+      }
+    }
+    return null;
   }
 
   componentDidCatch(error: Error) {
--- a/packages/replay-next/components/console/ConsoleInput.tsx
+++ b/packages/replay-next/components/console/ConsoleInput.tsx
@@ -22,8 +22,9 @@ import useTerminalHistory from "./hooks/useTerminalHistory";
 import styles from "./ConsoleInput.module.css";
 
 export default function ConsoleInput({ inputRef }: { inputRef?: RefObject<ImperativeHandle> }) {
+  const { executionPoint } = useContext(TimelineContext);
   return (
-    <ErrorBoundary fallback={<ErrorFallback />}>
+    <ErrorBoundary resetKey={executionPoint} fallback={<ErrorFallback />}>
       <Suspense fallback={<Loader />}>
         <ConsoleInputSuspends inputRef={inputRef} />
       </Suspense>
@@ -186,7 +187,7 @@ function ErrorFallback() {
   return (
     <div className={styles.FallbackState}>
       <Icon className={styles.Icon} type="terminal-prompt" />
-      Input disabled for session because of an error
+      Input disabled because of an error
     </div>
   );
 }
--- a/packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
+++ b/packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
@@ -38,6 +38,7 @@ import {
   POINT_BEHAVIOR_ENABLED,
   Point,
 } from "shared/client/types";
+import { isThennable } from "shared/proxy/utils";
 
 import Loader from "../../Loader";
 import { SetLinePointState } from "../SourceListRow";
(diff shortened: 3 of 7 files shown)


markerikson (approved):

packages/replay-next/components/console/ConsoleInput.tsx
```diff
@@ -22,8 +22,9 @@ import useTerminalHistory from "./hooks/useTerminalHistory";
 import styles from "./ConsoleInput.module.css";
 
 export default function ConsoleInput() {
+  const { executionPoint } = useContext(TimelineContext);
   return (
-    <ErrorBoundary fallback={<ErrorFallback />}>
+    <ErrorBoundary key={executionPoint} fallback={<ErrorFallback />}>
```
bvaughn, line 27: I see why you're doing this, and it makes sense. I think we should do it slightly differently though. Else any input that's in the console when you change execution point (even if just stepping through a function) will be erased, which seems bad.

bvaughn: I think maybe a better UX (though more complicated code) would be for us to increment the key only if:
1. an error has already occurred
2. the execution point changes

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -157,15 +158,21 @@ function PointPanelWithHitPoints({
...
+    if (frameId !== null) {
+      pauseAndFrameId = {
+        frameId,
+        pauseId,
+      };
+    }
+  } catch (errorOrPromise) {
+    if (isThennable(errorOrPromise)) {
+      throw errorOrPromise;
+    }
```
bvaughn, line 175: Do we _expect_ this error to happen? If not, we should probably at least log it to the console.

If we do though, can we add an inline comment with an explanation of why?

src/ui/components/SecondaryToolbox/TerminalContextAdapter.tsx
```diff
@@ -20,16 +20,15 @@ export default function TerminalContextController({ children }: PropsWithChildre
   const idCounterRef = useRef(0);
 
   const addMessage = useCallback((partialTerminalExpression: NewTerminalExpression) => {
+    const terminalExpression: TerminalExpression = {
```
bvaughn, line 23: nit: Can you ad dan inline comment of why we're doing this? It looks like the kind of thing that might get refactored back to how it was before if someone doesn't know.

I think the only/main thing you're trying to do is prevent the `idCounterRef.current` value from being incremented more than once?

packages/replay-next/components/console/ConsoleInput.tsx
hbenl: I've added a `resetKey` prop to `ErrorBoundary` that does exactly that. Please have a look if my implementation of that behavior is correct.

src/ui/components/SecondaryToolbox/TerminalContextAdapter.tsx
hbenl: > I think the only/main thing you're trying to do is prevent the idCounterRef.current value from being incremented more than once?

That was my initial concern but now I've added some more logic to ensure that the same message isn't added multiple times if the callback to `setMessages()` is called multiple times. I'm not sure if that extra logic is needed but I added it just in case. I've also added a warning that the callback may be called multiple times.

src/ui/components/SecondaryToolbox/TerminalContextAdapter.tsx
```diff
@@ -20,16 +20,22 @@ export default function TerminalContextController({ children }: PropsWithChildre
...
-        },
-      ]);
+      setMessages(prev => {
+        // warning: this callback may be called multiple times
+        // for a single addMessages() call
+        if (prev.some(expression => expression.id === terminalExpression.id)) {
+          return prev;
+        }
+        return [...prev, terminalExpression];
+      });
```
bvaughn, line 38: Ah! It seems like I misunderstood what you were saying the other day. I thought you said the `startTransition` callback was called multiple times, which should not happen.

The `setState` callback definitely _might_ be called multiple times. Particularly if React has other updates and it wants to "rebase" this one on top of the latest state. That's expected behavior for React and one of the reasons `setState` callbacks should not have side effects (so I guess, technically my bad for incrementing the counter that way).

The "fix" for this should just be:

```js
const addMessage = useCallback(
  (partialTerminalExpression: NewTerminalExpression) => {
    // New expressions should be added in a transition because they suspend.
    startTransition(() => {
      const id = idCounterRef.current++;

      setMessages((prev) => [
        ...prev,
        {
          ...partialTerminalExpression,
          id,
          type: "TerminalExpression",
        },
      ]);
    });
  },
  []
);
```

packages/replay-next/components/ErrorBoundary.tsx
```diff
@@ -1,20 +1,43 @@
...
+  ): Partial<ErrorBoundaryState> | null {
+    if (state.error) {
+      if (state.erroredOnKey === null) {
+        // we just caught the error and need to save the current resetKey
+        return { erroredOnKey: props.resetKey };
+      }
+      if (props.resetKey !== state.erroredOnKey) {
+        // the resetKey changed after an error was caught so we need to reset the state
+        return { error: null, erroredOnKey: null };
+      }
```
bvaughn, line 38: I think there's a mismatch between the _optional_ `resetKey` prop (which will be `undefined` by default) and the `erroredOnKey` state (that we initialize to `null`) that will cause unnecessary updates here.

I would make the key optional with a type of `string | null` and then default the value to `null` when the prop wasn't present.

packages/replay-next/components/ErrorBoundary.tsx
```diff
@@ -1,20 +1,43 @@
...
+  resetKey?: string | number;
 }>;
 
 export default class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
-  state: ErrorBoundaryState = { error: null };
+  state: ErrorBoundaryState = { error: null, erroredOnKey: null };
 
   static getDerivedStateFromError(error: Error): ErrorBoundaryState {
-    return { error };
+    return { error, erroredOnKey: null };
```
bvaughn, line 23: Kind of a nit but "key" has [special meaning](https://reactjs.org/docs/lists-and-keys.html#keys) in React and isn't actually a "prop" that can be accessed. So the thing we're storing here is actually the `resetKey` at the time of the error. Maybe the state field could more accurately reflect that? I dunno, `resetKeyAtTimeOfError` or something?

packages/replay-next/components/ErrorBoundary.tsx
hbenl: > that will cause unnecessary updates here

How?

> I would make the key optional with a type of string | null and then default the value to null when the prop wasn't present.

I deliberately didn't allow the value `null` for `resetKey` because I want `null` to mean that `erroredOnKey` still needs to be set to the current `resetKey` value.

bvaughn: > > that will cause unnecessary updates here
>
> How?

```js
// state.erroredOnKey will be "null"
if (state.erroredOnKey === null) {
  // props.resetKey will be "undefined"
  // so this will schedule an unnecessary update with React
  return { erroredOnKey: props.resetKey };
}
```

I think this should be:
```js
if (state.erroredOnKey === null) {
  if (props.resetKey != null) {
    return { erroredOnKey: props.resetKey };
  }
} else if (props.resetKey !== state.erroredOnKey) {
  // the resetKey changed after an error was caught so we need to reset the state
  return { error: null, erroredOnKey: null };
}
```

hbenl: But that would mean the `ErrorBoundary` is reset if `resetKey` changes *except* if it changes to `undefined`. That exception seems unexpected.

hbenl: Also, isn't `getDerivedStateFromProps()` called as part of an update, before rendering, so it wouldn't trigger another rerender?

bvaughn: > But that would mean the `ErrorBoundary` is reset if `resetKey` changes *except* if it changes to `undefined`. That exception seems unexpected.

Sorry. I don't understand what you're saying.

I'm suggesting that this whole `getDerivedStateFromProps` should do _nothing_ unless a `resetKey` prop is provided.

> Also, isn't `getDerivedStateFromProps()` called as part of an update, before rendering, so it wouldn't trigger another rerender?

It's still going to require React to restart the render (since it's a _different state value_).

bvaughn: I think you're thinking of "render" as literally what's being done in a component's `render` method and I'm referring to "render" as "all of the work React does to render a component– which includes rebasing and diffing state.

> If I have `<ErrorBoundary resetKey={myKey} />`, it will reset if `myKey` changes e.g. from `"foo"` to `"bar"` but not if it changes from `"foo"` to `undefined`.

That's fair. I wasn't considering  the case where we'd pass a `resetKey` and then stop passing one. Hmm... what's the _expected_ behavior there? I could see an argument for why state _shouldn't_ reset in that case.

This is probably well within the "nit" category though so feel free to ignore my input on this.

hbenl: > I'm referring to "render" as "all of the work React does to render a component– which includes rebasing and diffing state.

OK, got it now. Thanks for the entire discussion, I feel like I understand React a little better now.

> what's the expected behavior there? I could see an argument for why state shouldn't reset in that case.

Yeah, it could be argued either way. I'll go with my initial gut feeling that it should also reset in that case.

PR #8636, opened 2023-01-30
Clicking on Console (messages) will auto focus on input

bvaughn:
Small UX improvement.

Note that clicking anywhere in the left or top filters area will not set focus.

Also fixed a small UX thing that was nagging me before: when the Console search is dismissed ("Escape" key) the Console input will be re-focused.


markerikson (approved):

bvaughn:
Looks like the e2e tests caught a legit regression. Going to try pushing a small fix here momentarily.

bvaughn:
Note to self that I think I may need to re-think this slightly tomorrow. Specifically, this auto-focus approach makes the following use case weird:
1. Select text in a Console message
2. Focus will jump to the terminal input (and text will be deselected)

bvaughn:
I've made the auto-focus logic a lot more constrained– basically only setting focus when you've clicked on "empty space" below the last console log message. If the console is full of messages, then the only way to focus will be to click on the input directly (or tab to it) as before.

PR #8643, opened 2023-01-31
Fixed bad column breakpoint conditional

bvaughn:
Verified the fix by:
1. Recreated part of the Glide source code.
2. Verified it reproduced the original bug.
3. Verified the bug was fixed after the change.

### DIFF
--- a/packages/replay-next/components/sources/SourceListRow.tsx
+++ b/packages/replay-next/components/sources/SourceListRow.tsx
@@ -168,7 +168,7 @@ const SourceListRow = memo(
       if (tokens) {
         for (let index = 0; index < tokens.length; index++) {
           const token = tokens[index];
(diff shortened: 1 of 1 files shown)


packages/replay-next/components/sources/SourceListRow.tsx
```diff
@@ -168,7 +168,7 @@ const SourceListRow = memo(
       if (tokens) {
         for (let index = 0; index < tokens.length; index++) {
           const token = tokens[index];
-          if (token.columnIndex + token.value.length > columnIndexEnd) {
+          if (token.columnIndex >= columnIndexEnd) {
```
bvaughn, line 171: I'm not sure what I was thinking originally, but this _seems_ like a more appropriate termination check. Note that we'd never expect to break _within_ a token because of how column breakpoints work.

markerikson (approved):

PR #8647, opened 2023-01-31
Add createdByUserId and recordingId to Point object

bvaughn:
This is an incremental step toward shared print statements:
- [x] Add `createdByUserId` to `Point` objects.
- [x] Add `recordingId` to `Point` objects.

~~I think the user information stored should be sufficient to render the user's picture (or some fallback) without also needing to fetch the user info from Replay.~~ Decided instead to just store `createdByUserId` and fetch whatever information we need for the user using the `useRecordingUsers` hook.

Next steps:
- [ ] Move log points from Indexed DB to GraphQL.
- [ ] Fetch and display all (shared) log points for the current recording.

### DIFF
--- a/packages/replay-next/src/contexts/PointsContext.tsx
+++ b/packages/replay-next/src/contexts/PointsContext.tsx
@@ -105,6 +105,7 @@ export function PointsContextRoot({ children }: PropsWithChildren<{}>) {
             }
           : null,
         createdAtTime: Date.now(),
+        recordingId,
         shouldBreak: POINT_BEHAVIOR_DISABLED,
         shouldLog: POINT_BEHAVIOR_DISABLED,
         ...partialPoint,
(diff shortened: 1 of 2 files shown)


packages/replay-next/src/contexts/PointsContext.tsx
```diff
@@ -96,7 +96,9 @@ export function PointsContextRoot({ children }: PropsWithChildren<{}>) {
         badge: null,
         content: "",
         condition: null,
+        createdByUserId: currentUserInfo?.id ?? null,
```
jasonLaster, line 99: `user_id` would be more consistent with our DB schema which we could transform on the JS side if we'd like

bvaughn: I think the only thing that interacts with this would be the client-side JS, so I have a slight preference to follow the client side naming conventions (and transfer before sending the server, if need be)

PR #8649, opened 2023-01-31
Remove source contents loading from Redux and fix breakpoint text

markerikson:
After the source viewer rewrite, a handful of components were still looking to find loaded source file contents in the Redux store:

- `<QuickOpenModal>` was doing an update bailout if not loaded
- `<BreakpointOptions>` uses the text to show a line snippet

Additionally, the source text was needed for Babel symbol parsing.

Along the way, I found out that our `<BreakpointOptions>` component has been broken since the "points refactor" in October - it was always trying to use the _selected_ source to read out line snippets, when it actually needs to be the source where the _breakpoint_ is located. That led to a display bug like this, where the line snippets are from the wrong file:

image

I've created Suspense caches for source contents, split source lines, and Babel-parsed symbols.

I also added a new `getStatus` method to the caches to let us check the loading status of a record.

`<QuickOpenModal>` now just imports the contents cache and and checks to see if the given source is loaded.

`<BreakpointOptions>` now suspends waiting for the source lines, and does the line text trimming itself.

This let me remove a couple hundred lines of Redux-based loading state tracking, and a lot of dead code in `utils/source.ts`.

You can see that the breakpoint snippets now show the correct text even if another file or _no_ file is selected, and they show "Loading" if suspended:

image

### DIFF
--- a/packages/replay-next/src/suspense/createGenericCache.ts
+++ b/packages/replay-next/src/suspense/createGenericCache.ts
@@ -13,11 +13,16 @@ import {
   Wakeable,
 } from "./types";
 
+export { STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED } from "./types";
+
+type CacheRecordStatus = typeof STATUS_PENDING | typeof STATUS_RESOLVED | typeof STATUS_REJECTED;
+
 export interface GenericCache<TParams extends Array<any>, TValue> {
   getValueSuspense(...args: TParams): TValue;
   getValueAsync(...args: TParams): Thennable<TValue> | TValue;
   getValueIfCached(...args: TParams): { value: TValue } | undefined;
   addValue(value: TValue, ...args: TParams): void;
+  getStatus(...args: TParams): CacheRecordStatus | undefined;
 }
 
 export interface GenericCache2<TExtra, TParams extends Array<any>, TValue> {
@@ -25,6 +30,7 @@ export interface GenericCache2<TExtra, TParams extends Array<any>, TValue> {
   getValueAsync(extra: TExtra, ...args: TParams): Thennable<TValue> | TValue;
   getValueIfCached(...args: TParams): { value: TValue } | undefined;
   addValue(value: TValue, ...args: TParams): void;
+  getStatus(...args: TParams): CacheRecordStatus | undefined;
 }
 
 interface HookState<TValue> {
@@ -48,6 +54,7 @@ export function createGenericCache<TParams extends Array<any>, TValue>(
     getValueAsync: (...args) => cache.getValueAsync(undefined, ...args),
     getValueIfCached: (...args) => cache.getValueIfCached(...args),
     addValue: (value, ...args) => cache.addValue(value, ...args),
+    getStatus: (...args) => cache.getStatus(...args),
   };
 }
 
@@ -138,6 +145,11 @@ export function createGenericCache2<TExtra, TParams extends Array<any>, TValue>(
     addValue(value: TValue, ...args: TParams) {
       const cacheKey = getCacheKey(...args);
       recordMap.set(cacheKey, { status: STATUS_RESOLVED, value });
+    },
+
+    getStatus(...args: TParams) {
+      const cacheKey = getCacheKey(...args);
+      return recordMap.get(cacheKey)?.status;
     },
   };
 }
--- a/src/devtools/client/debugger/src/actions/ui.ts
+++ b/src/devtools/client/debugger/src/actions/ui.ts
@@ -6,8 +6,9 @@
 
 import type { Context } from "devtools/client/debugger/src/reducers/pause";
 import { copyToClipboard as copyTextToClipboard } from "replay-next/components/sources/utils/clipboard";
+import { getSourceContentsIfCached } from "replay-next/src/suspense/SourcesCache";
 import type { UIThunkAction } from "ui/actions";
-import { SourceDetails, getSourceContent, getSourceDetails } from "ui/reducers/sources";
+import { SourceDetails, getSourceDetails } from "ui/reducers/sources";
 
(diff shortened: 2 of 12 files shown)


packages/replay-next/src/suspense/createGenericCache.ts
```diff
@@ -13,18 +13,24 @@ import {
...
+export { STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED } from "./types";
+
+type CacheRecordStatus = typeof STATUS_PENDING | typeof STATUS_RESOLVED | typeof STATUS_REJECTED;
+
 export interface GenericCache<TParams extends Array<any>, TValue> {
   getValueSuspense(...args: TParams): TValue;
   getValueAsync(...args: TParams): Thennable<TValue> | TValue;
   getValueIfCached(...args: TParams): { value: TValue } | undefined;
   addValue(value: TValue, ...args: TParams): void;
+  getStatus(...args: TParams): CacheRecordStatus | undefined;
```
bvaughn, line 25: I wonder if `STATUS_UNINITIALIZED` would be a bit clearer?

My personal preference is to avoid `undefined` since it can mask unhandled edge cases.

src/devtools/client/debugger/src/actions/ui.ts
```diff
@@ -99,10 +100,11 @@ export function flashLineRange(location: HighlightedRange): UIThunkAction {
...
 export function copyToClipboard(source: SourceDetails): UIThunkAction {
-  return (dispatch, getState) => {
-    const content = getSourceContent(getState(), source.id);
-    if (content?.value?.type === "text") {
-      copyTextToClipboard(content.value.value);
+  return async (dispatch, getState, { replayClient }) => {
+    const sourceContents = await getSourceContentsAsync(replayClient, source.id);
+
+    if (typeof sourceContents?.contents === "string") {
+      copyTextToClipboard(sourceContents.contents);
```
bvaughn, line 107: This change seems iffy. Could take _a while_ for source contents to load (if they haven't already loaded) in which case the UX would be kind of weird:
1. Click the copy menu item
2. Go paste the text, seeing either _nothing_ or a stale value
3. Confused, maybe do something else
4. Your clipboard is async updated unexpectedly (possibly wiping out a different thing you'd copied)

I think we should avoid async copy interactions, especially because of 4.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointOptions.tsx
```diff
@@ -1,23 +1,42 @@
...
+    }
+
+    return snippet;
+  }, [sourceLines, breakpoint]);
+
+  if (!snippet) {
+    return null;
+  }
+
+  return <SyntaxHighlightedLine code={snippet} />;
```
bvaughn, line 36: I wonder if we should read from the syntax parsing cache rather than the source cache? Sometimes parsing a small snippet in isolation will yield bad results.

src/ui/suspense/sourceCaches.ts
```diff
@@ -0,0 +1,72 @@
...
+  async (replayClient, sourceId) => {
+    const res = await getStreamingSourceContentsHelper(replayClient, sourceId);
+    if (res) {
+      // Ensure that follow-on caches have the entire text available
+      const sourceContents = await res.resolver;
+      return sourceContents;
+    }
+  },
+  (replayClient, sourceId) => sourceId
+);
```
bvaughn, line 27: If we need a wrapper around the streaming API, I think it would be nicer to co-locate it with the rest of the source cache (rather than string it out in multiple files): `packages/replay-next/src/suspense/SourcesCache`

src/devtools/client/debugger/src/actions/ui.ts
markerikson: Hmm. This only ends up getting called from `useTabContextMenu`.  I think it's fairly reasonable to assume that by the time you've clicked on the tab, the source has been loaded.

If we're _really_ concerned about it, I guess we could do a sync cache check here instead.

bvaughn: The Source viewer is optimized for streaming (because of use cases like Andarist's). Source tabs show as soon as we know file names– after which point, you could use the context menu.

Source content streams in (and is parsed) in chunks. This approach to the copy action would wait on the streaming/chunks to finish, which could be a bit of time.

packages/replay-next/src/suspense/createGenericCache.ts
markerikson: Hmm. We don't have a `STATUS_UNINITIALIZED` value atm.

bvaughn: I know. I was suggesting we _add one_ rather than `undefined` because `undefined` seems like a bad pattern since it can mask unhandled edge cases.

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointOptions.tsx
bvaughn: @markerikson Looks like this already got merged, but it seems like this would be worth doing.

markerikson: Yeah, filed FE-1208 for that.

bvaughn: Oh! Thanks I hadn't even noticed that.

markerikson: I only filed it after you made that comment :)

PR #8666, opened 2023-02-01
Render the stacktrace of Error objects in the console (FE-1158)

hbenl:
This is especially useful when a caught error is logged:
```javascript
try {
  ...
} catch (error) {
  console.log(error);
}
```
The console message lets you jump to the location where the error was *caught and logged* but the stacktrace in the Error object will give you the location where it was *thrown* (or rather where the Error object was created but in most cases that'll be the same).


packages/replay-next/components/console/renderers/MessageRenderer.tsx
```diff
@@ -97,12 +99,32 @@ function MessageRenderer({
   const primaryContent =
     argumentValues.length > 0 ? (
       <Suspense fallback={<Loader />}>
-        {argumentValues.map((argumentValue: ProtocolValue, index: number) => (
-          <Fragment key={index}>
+        {argumentValues.map((argumentValue: ProtocolValue, index: number) => {
+          const argumentInspector = (
             <Inspector context="console" pauseId={message.pauseId} protocolValue={argumentValue} />
```
bvaughn, line 104: Doesn't `Inspector` render object types with `<Expandable>` around them anyway?

Is the issue here that it _also_ fetches the preview and we don't need that in this case? Does this error have some kind of a different structure than our standard objects?

packages/replay-next/components/console/ErrorStackRenderer.tsx
```diff
@@ -0,0 +1,145 @@
...
+  pauseId,
+  errorObjectId,
+}: {
+  pauseId: PauseId;
+  errorObjectId: ObjectId;
+}) {
+  const client = useContext(ReplayClientContext);
+  const stack = getObjectPropertySuspense(client, pauseId, errorObjectId, "stack").value;
+  assert(typeof stack === "string", "no stack string found in error object");
+  const frames = parseStackString(stack);
```
bvaughn, line 41: Consider using [the `error-stack-parser` module](https://www.npmjs.com/package/error-stack-parser) instead?

bvaughn (approved):

packages/replay-next/components/console/renderers/MessageRenderer.tsx
hbenl: > Doesn't Inspector render object types with <Expandable> around them anyway?

No, for Error objects it doesn't (since they're not expandable).

PR #8667, opened 2023-02-01
Testsuites: Hide alias badge when the window gets tiny

jonbell-lot23:
This PR hides the alias badges when things get too narrow:
https://www.loom.com/share/2755b1377d374dcd807d1f4adb155081

We can discuss other visual treatments like little minimised icons or something, but this is a good start.

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -486,6 +486,10 @@
   --breakpoint-tip: #fff;
 
   /* Testsuites (Dark) */
+  --testsuites-capsule-alias-bgcolor: var(--theme-base-85);
+  --testsuites-capsule-alias-color: var(--body-color);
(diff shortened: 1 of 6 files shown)


ryanjduffy (changes requested):
Couple implementation notes below. Happy to pair on them if you'd like. Thanks for the fix, @jonbell-lot23 !

jonbell-lot23:
Update:

This code could get the job done, but there are some things we'd like to do to make the code less tacky:

- [ ] Jon is going to do a PR that combines our CSS modules. This is important because we're trying to affect some css classes based on parent classes, and it'd be much easier if we weren't splitting things across CSS files.
- [ ] Separately, Ryan is going to wire this up with a resize eventlistener at a higher level rather than making it specific to this one component

PR #8670, opened 2023-02-01
Convert large analysis/evaluation strings of JS into real TS functions

markerikson:
We've complained for a long time that trying to write code run on the backend via evaluation or analysis is really hard, because we have to write that evaluated code as giant strings of JS text.  This means no syntax highlighting and no type checking, and it's really easy to make mistakes.

After a bunch of recent discussion ( ref: FE-1199 ), we've concluded that it's feasible to actually write our "code to be evaluated" as _real_ TS functions.

There's a few tricks involved:

- In JS, calling `someFunction.toString()` gives you back the _entire_ function definition as an action string, like `function someFunction() { console.log('whatever'); }`
- We can write the function as normal in TypeScript. The TS syntax wouldn't run on the backend because it's not legal JS... but _the TS syntax has already been compiled away at build time_.  So, the stringified function at runtime will _only_ be JS code.
- We frequently need to refer to values that already exist in scope.  For example, Analysis mappers have access to `sendCommand()`, and there's an `input` variable in scope that has details on the current execution point.  In other cases, we may be evaluating code in a particular stack frame, and have expectations that certain variables exist.  Normally, referencing a variable name that hasn't been declared would cause a compile error.  If we did a `declare global { let someVar }`, this introduces a major footgun because now _any_ file in our codebase could reference `someVar` without an error. But, we can use TS's `declare` keyword in the same file that we define the evaluated function, _now_ we can reference it without error.  In the case above, that means `declare let sendCommand: SendCommand` and `declare let input: AnalysisInput`.
- Because this is evaluated code, we can't close over any actual variables in scope in our source files. There's no magic bullet for this - you just have to be _very_ careful when writing the logic.  It does also mean that when we have dynamic behavior, such as determining which React event type to look for, we have to do ~~string search-and-replace on the resulting JS string~~ okay no _not_ search-and-replace. Instead, write it as an IIFE and inject the values as the IIFE arg.
- The Analysis code won't let us call `sendCommand()` inside a nested function for some reason, so we have to extract _just_ the body of the function and send that over. Per https://stackoverflow.com/q/3179861/62937 , this can be done with some `string.indexOf('{')` checks.

Given all that, this PR:

- Adds util functions for extracting the stringified body of a function and doing some multiple search-and-replaces
- Adds predefined TS types for `sendCommand` and `input`, which are likely to be used in analysis
- Replaces all of the current major analysis "mapper" strings (Events, Exceptions, Print Statement contents and conditions) with the same code converted to TS functions (and using string search-and-replace as necessary)
- Replaces the React Event Listeners evaluation JS logic the same way
- Deletes some dead utils
- Renames `ThreadFront.evaluateNew()` back to just `ThreadFront.evaluate()`

**update** As I found out the hard way, the "stringify a TS function" thing initially broke in prod, because the build step minifies the source. That meant that my initial attempts to do string search-and-replace would fail.

I've finally settled on this form:

```ts
// Define values we'll be injecting
interface InjectedValues {
  $REACT_16_EVENT_LISTENER_PROP_KEY: string;
  $REACT_17_18_EVENT_LISTENER_PROP_KEY: string;
  EVENT_CLASS_NAME: string;
  possibleReactPropNames: string[];
  args: any[];
}

function createReactEventMapper(eventType: SEARCHABLE_EVENT_TYPES) {
  // The actual function that will be evaluated
  function findEventTargetAndHandler(injectedValues: InjectedValues) {
    // EVALUATED JS CODE HERE
  }

  const reactEventPropNames = REACT_EVENT_PROPS[eventType];
  const eventClassName = EVENT_CLASS_FOR_EVENT_TYPE[eventType];

  // Use that function as the body of an IIFE,
  // and pass in the injected values as a single object arg
  let evaluatedEventMapperBody = `
    (${findEventTargetAndHandler})({
      $REACT_16_EVENT_LISTENER_PROP_KEY: "${REACT_16_EVENT_LISTENER_PROP_KEY}",
      $REACT_17_18_EVENT_LISTENER_PROP_KEY: "${REACT_17_18_EVENT_LISTENER_PROP_KEY}",
      EVENT_CLASS_NAME: "${eventClassName}",
      possibleReactPropNames: ${JSON.stringify(reactEventPropNames)},
      
      // Outer body runs in scope of the "current" event handler.
      // Grab the event handler's arguments.
      args: [...arguments]
    })
  `;

  return evaluatedEventMapperBody ;
}
```


markerikson:
Very likely that I broke something :( Investigating...

markerikson:
Looks like the issue is that the `someFunction.toString()` approach works fine _in development_... but in prod, the function has been mangled.  This breaks things in two ways:

- The string search-and-replace technique I was using may not match correctly in some cases
- I'm hitting that weird error/limitation that `"sendCommand can only be referenced by a function callee"`, which appears to be coming from our own backend in `analysis-worker/transform.ts` as part of a Babel transformation applied to all analysis mapper code.

markerikson:
At this point:

- The rewritten evaluation logic appears to work in prod builds
- Delta screenshots seem very messed up
- Not sure what's up with the E2E tests.  All jobs passed, the comment bot shows two failures. Glanced at the tests and not seeing anything obviously wrong.

bvaughn:
Looks like a bunch of Playwright tests are _failing_ before writing snapshot data to disk (so the files show up as "deleted" since they're missing in the branch results).

https://github.com/replayio/devtools/actions/runs/4070173941/jobs/7010736717

PR #8672, opened 2023-02-01
Consolidating CSS in Testsuites

jonbell-lot23:
See https://github.com/replayio/devtools/pull/8667 for context.

That PR is attempting to hide certain classes when the panel gets too small. @ryanjduffy and I decided to split things across this CSS-oriented PR and a PR where he'll wire it up.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestInfo.module.css
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestInfo.module.css
@@ -35,16 +35,3 @@
 .TestsuitesErrorColor {
   color: var(--testsuites-error-color);
 }
-
-#TestSteps {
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/components/TestInfo/TestInfo.module.css
```diff
@@ -35,3 +35,16 @@
 .TestsuitesErrorColor {
   color: var(--testsuites-error-color);
 }
+
+#TestSteps {
```
ryanjduffy, line 39: This doesn't seem to be applying to anything but let's use a class name here anyway

ryanjduffy (approved):

PR #8674, opened 2023-02-02
Consolidate GenericCache and GenericCache2 (FE-1204)

hbenl:


bvaughn (approved):
Thanks for consolidating these!

My only nit isn't a new detail, but it seems odd to me that the "extra" param comes at the _beginning_ for some of the functions 🤷🏼 but it's not a big deal since I mostly copy-paste an existing cache when creating a new one anyway 😁

PR #8676, opened 2023-02-02
Store (shared) Points in GraphQL

bvaughn:
### [Loom demo of feature](https://www.loom.com/share/291cbd6b23264a40a7cbe5015e917453)

This PR enables points (log points and breakpoints) to be shared with other users on a per-recording basis. Although there are some similarities to Comments, Points are a little more complex because:
1. They can be created by _unauthenticated_ users, and
2. They may cause React to _suspend_ during an update

Because of this, the frontend needs to store multiple copies of points, in:
* Memory (React state)
* IndexedDB
* GraphQL (for authenticated users)

---

Follow up on:
- [ ] Should breakpoints be shared?
- [ ] Should shared print statements be enabled or disabled by default?
- [ ] Shared breakpoint blocks print statement for non-owners
        I think this shakes out of our design constraints: only one point per line, toggle logging on/off

---

Relates to FE-1138,, FE-1201, and FE-1223 and https://github.com/replayio/backend/pull/7025


bvaughn:
I've given this some thought and I think we should separate the should-log and should-break attributes from the `Point` type and store them separately. We'll need to do something similar in order to allow local users to override logging/breaking behavior for shared points anyway, and I think we should use the same approach. Here is what I propose:

<table>
  <tr>
    <td colspan="2"><h3><code>Point</code> <small>(GraphQL, IndexedDB)</small></h3></td>
  </tr>
  <tr>
    <td colspan="2"><h4>Static <code>Point</code> attributes</h4></td>
  </tr>
  <tr>
    <td>
      <br>
      <ul>
        <li>recording</li>
        <li>user</li>
        <li>location</li>
      </ul>
    </td>
    <td>
      These are assigned when the point is created and cannot be changed.
      Note there is no explicit "id" attribute, rather the combination of recording id, user id, and source location are used to uniquely identify a point.
      (Note that <a href="https://github.com/hasura/graphql-engine/pull/2148">Hasura supports composite keys and unique constraints</a>.)
    </td>
  </tr>
  <tr>
    <td colspan="2"><h4>Dynamic <code>Point</code> attributes</h4></td>
  </tr>
  <tr>
    <td>
      <br>
      <ul>
        <li>badge</li>
        <li>conditional</li>
        <li>content</li>
      </ul>
    </td>
    <td>
      These attributes may be edited after a point is created.
      Current thinking is that they are only editable by the user who created the point although there is no technical reason why this has to remain the case.
    </td>
  </tr>
  <tr>
    <td colspan="2"><h4>Point <em>behaviors</em></h4></td>
  </tr>
  <tr>
    <td>
      <br>
      <ul>
        <li>breaking</li>
        <li>logging</li>
      </ul>
    </td>
    <td>
      These control how a point should behave in a local session (e.g. should it log?)
      These are persisted to IndexedDB because viewers should be able to control the behavior of points independently.
      They are stored in-memory in a Map, keyed by point "id" (recording id, user id, and source location).
    </td>
  </tr>
</table>

There are a few constraints that may be worth explicitly noting about the above proposal:
* Recording owners can't control the default enabled/disabled states of points with this model (unless we were to add additional metadata to the recording). I think this is okay though– Replay should pick a good default behavior (e.g. shared points always enabled by default?) and allow users to override this default somehow.
* Enabled/disabled state wouldn't be synced between multiple browsers for the same user. (This is probably not a common enough use case to even worry about.)

bvaughn:
I need to do some end-to-end testing (which I can't at the moment due to bad Amtrak wifi) but I think the general structure of this PR should be ready for review.

packages/replay-next/src/contexts/PointsContext.tsx
```diff
@@ -34,132 +52,299 @@ export type PointInstance = {
...
+) => void;
 
 export type PointsContextType = {
   addPoint: AddPoint;
   deletePoints: DeletePoints;
   editPoint: EditPoint;
+  editPointBehavior: EditPointBehavior;
   isPending: boolean;
+  pointBehaviors: PointBehaviorsMap;
+  pointBehaviorsForSuspense: Map<PointId, PointBehavior>;
```
hbenl, line 77: why not use `PointBehaviorsMap` here as well?

hbenl (approved):

packages/replay-next/src/contexts/PointsContext.tsx
bvaughn: Ha, just an oversight.

bvaughn:
Backend PR has been merged so the frontend is unblocked now.

bvaughn:
Maybe we can further simplify things here for _pending edits_ (which are currently tracked in a `LinePointState` map shared via context) by adding a 3rd, in-memory-only bucket of points:
Screen Shot 2023-02-09 at 12 16 48 PM

bvaughn:
I think maybe this would more or less work:
Screen Shot 2023-02-09 at 12 26 47 PM

markerikson:
@bvaughn my main concern with that is that `IndexedDB` _doesn't_ have any kind of subscription.  What are you intended to subscribe to, and how?

bvaughn:
> @bvaughn my main concern with that is that `IndexedDB` _doesn't_ have any kind of subscription mechanism (unlike `localStorage`, which has the `"storage"` event).  What are you intended to subscribe to, and how?

My wording was poor. I was thinking of the `useIndexedDB` hook (which stores a copy in state, so when you update a point– state is also updated). By comparison, the Points GraphQL cache is static after the initial read.

bvaughn:
I pushed the points context refactor. It touched a lot of imports (unsurprisingly). I need to spend a while smoke testing now though. It probably regressed some things.

PR #8677, opened 2023-02-02
Overboard "ready when you are" screens

jonbell-lot23:
**Background**

We're looking for ways to incorporate more Replay branding elements into our app. We also have related topics like the ethics of AI art, whether we should hire an illustrator, etc. This is a down payment on the idea, using our standard overboard.dev hoverboard in a range of colors. We'll be iterating further from here.

image


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-248/explore-low-hanging-fruit-for-illustrations

jonbell-lot23:
Adding Jason as reviewer – my hope is to ease us into the idea of having an illustration here, which will inspire further iterations where we can get into more details about ai art, illustrators, overall branding, etc.

jasonLaster (approved):
Ship it

PR #8678, opened 2023-02-02
Disable repainting graphics in Chromium

jcmorrow:
This fails constantly, and it can take a while and block other (possibly successful) manifests.


jasonLaster (approved):
Great idea

PR #8683, opened 2023-02-03
Remember state of Expandables in the Scopes panel (FE-939)

hbenl:
This uses the `functionLocation` that we get for call frames from the backend in the `persistenceKey` for which the state is saved. So the Scopes panel remembers the expandable states when stepping within a function and when coming back to it after stepping into a different function.

### DIFF
--- a/packages/e2e-tests/helpers/pause-information-panel.ts
+++ b/packages/e2e-tests/helpers/pause-information-panel.ts
@@ -352,7 +352,12 @@ export async function waitForPaused(page: Page, line?: number): Promise<void> {
   }
 }
 
-export async function waitForScopeValue(page: Page, name: string, expectedValue: Expected) {
+export async function waitForScopeValue(
+  page: Page,
+  name: string,
+  expectedValue: Expected,
+  expandScopes = true
+) {
   await debugPrint(
     page,
     `Waiting for scope with variable "${chalk.bold(name)}" to have value "${chalk.bold(
@@ -364,7 +369,9 @@ export async function waitForScopeValue(page: Page, name: string, expectedValue:
   const escapedValue =
     typeof expectedValue === "string" ? expectedValue.replace(/"/g, '\\"') : expectedValue;
 
-  await expandAllScopesBlocks(page);
+  if (expandScopes) {
+    await expandAllScopesBlocks(page);
+  }
 
   const scopesPanel = getScopesPanel(page);
   const scopeValue = scopesPanel
--- a/packages/replay-next/components/inspector/HTMLChildrenRenderer.tsx
+++ b/packages/replay-next/components/inspector/HTMLChildrenRenderer.tsx
@@ -9,6 +9,7 @@ import { getObjectWithPreviewSuspense } from "replay-next/src/suspense/ObjectPre
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 
 import KeyValueRenderer from "./KeyValueRenderer";
+import { addPathSegment } from "./PropertiesRenderer";
 
 // HTML entries are a special case.
 // Unlike the other property lists, HTML entries only display child nodes:
@@ -20,9 +21,11 @@ import KeyValueRenderer from "./KeyValueRenderer";
 //         </li>
 export default function HTMLChildrenRenderer({
   object,
+  path,
   pauseId,
 }: {
   object: ProtocolObject;
+  path?: string;
   pauseId: ProtocolPauseId;
 }) {
   const childNodes = object.preview?.node?.childNodes ?? [];
@@ -33,7 +36,12 @@ export default function HTMLChildrenRenderer({
   return (
     <>
       {childNodes.map((objectId: ProtocolObjectId, index: number) => (
-        <HTMLChildRenderer key={index} objectId={objectId} pauseId={pauseId} />
+        <HTMLChildRenderer
+          key={index}
+          objectId={objectId}
+          path={addPathSegment(path, `${index}`)}
+          pauseId={pauseId}
+        />
       ))}
     </>
   );
@@ -41,9 +49,11 @@ export default function HTMLChildrenRenderer({
 
 function HTMLChildRenderer({
   objectId,
+  path,
   pauseId,
 }: {
   objectId: ProtocolObjectId;
+  path?: string;
   pauseId: ProtocolPauseId;
 }) {
   const client = useContext(ReplayClientContext);
@@ -61,6 +71,7 @@ function HTMLChildRenderer({
     <KeyValueRenderer
       context="nested"
       layout="vertical"
+      path={path}
       pauseId={pauseId}
       protocolValue={protocolValue}
     />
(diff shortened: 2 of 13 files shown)


hbenl:
For some reason object_preview-01 keeps failing in CI although it works locally and the recordings of the failures all fail to load.

markerikson (approved):

packages/e2e-tests/helpers/pause-information-panel.ts
```diff
@@ -352,7 +352,12 @@ export async function waitForPaused(page: Page, line?: number): Promise<void> {
   }
 }
 
-export async function waitForScopeValue(page: Page, name: string, expectedValue: Expected) {
+export async function waitForScopeValue(
+  page: Page,
+  name: string,
+  expectedValue: Expected,
+  expandScopes = true
```
bvaughn, line 359: What would it mean to call this function _without_ this value being true? I think it would mean we'd be stuck waiting forever for a value that wasn't in the HTML. Seems bad?

bvaughn, line 359: What would it mean to call this function _without_ this value being true? I think it would mean we'd be stuck waiting forever for a value that wasn't in the HTML. Seems bad?

bvaughn, line 359: What would it mean to call this function _without_ this value being true? I think it would mean we'd be stuck waiting forever for a value that wasn't in the HTML. Seems bad?

packages/replay-next/components/inspector/HTMLChildrenRenderer.tsx
```diff
@@ -33,17 +36,24 @@ export default function HTMLChildrenRenderer({
   return (
     <>
       {childNodes.map((objectId: ProtocolObjectId, index: number) => (
-        <HTMLChildRenderer key={index} objectId={objectId} pauseId={pauseId} />
+        <HTMLChildRenderer
+          key={index}
+          objectId={objectId}
+          path={addPathSegment(path, `${index}`)}
```
bvaughn, line 42: Calling a function during render is a [side effect](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects). This will end up being called many times which is probably _not_ what we want.

Should either use memo in `PropertiesRenderer` to pre-compute (and memoize) all paths, _or_ we could move this callback into an effect. (Not sure which makes the most sense as I'm still reading through this PR– but I assume `useMemo`).

packages/replay-next/components/inspector/HTMLExpandable.tsx
```diff
@@ -22,20 +24,29 @@ export default function HTMLExpandable({
...
   defaultOpen = false,
   object,
+  path,
   pauseId,
   protocolValue,
 }: {
   before?: ReactNode;
   defaultOpen?: boolean;
   object: ProtocolObject;
+  path?: string;
```
bvaughn, line 34: Kind of a nit: but why is this called "persistence key" some places and "path" others? Might be nice to name it consistently.

packages/replay-next/components/inspector/ScopesInspector.tsx
```diff
@@ -25,11 +28,13 @@ export default function ScopesInspector({
             <Inspector
               context="default"
               key={index}
+              path={addPathSegment(path, protocolValue.name)}
```
bvaughn, line 31: Same as above RE: side effects

packages/replay-next/src/contexts/ExpandablesContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+  setIsExpanded: () => {},
+});
+
+export function ExpandablesContextRoot({ children }: PropsWithChildren) {
+  const expandedKeys = useMemo(() => new Map<string, boolean>(), []);
+  const isExpanded = useCallback((key: string) => expandedKeys.get(key), [expandedKeys]);
+  const setIsExpanded = useCallback(
+    (key: string, expanded: boolean) => expandedKeys.set(key, expanded),
+    [expandedKeys]
+  );
```
bvaughn, line 19: I think this works but I wonder if you've considered the fact that `setIsExpanded` _sounds_ like it should trigger a state update (like other React "setters") when it won't actually in this case.

Maybe it would be nicer to name thing something more passive sounding– like "onExpandedChange" or something.

Alternately you could refactor `expandedKeys` to be stored in state but that's probably overkill.

packages/replay-next/components/inspector/PropertiesRenderer.tsx
```diff
@@ -100,81 +102,99 @@ export default function PropertiesRenderer({
     }
   }
 
+  let renderedPrototype: ReactNode = null;
+  if (prototype != null) {
+    const prototypePath = addPathSegment(path, "[[Prototype]]");
```
bvaughn, line 107: Same comment RE: side effects

bvaughn: D'oh– I just _saw_ `addPathSegment` and it's just concatenating strings. Feeling kind of silly now. Ignore the "side effects" comments!

packages/e2e-tests/helpers/pause-information-panel.ts
hbenl: It means that the function doesn't try to expand anything, in this case it will only succeed if the scopes are already expanded - which is exactly what I wanted to test (see the changes in object_preview-03).

packages/replay-next/src/contexts/ExpandablesContext.tsx
hbenl: To me `setIsExpanded` sounds like it sets a value and I don't feel that ties it to React state updates. I don't like `onExpandedChange` because it doesn't say anything about what it does, only when it is expected to be called.

> Alternately you could refactor expandedKeys to be stored in state but that's probably overkill.

Yeah, I considered that but also saw it as overkill.

packages/e2e-tests/helpers/pause-information-panel.ts
bvaughn: I guess that makes sense, but I think maybe we should use a different method for that– an _expect_ method, rather than a _wait for_ method. I guess this is kind of subjective but I think it would fit a little more with our other existing methods.

packages/replay-next/src/contexts/ExpandablesContext.tsx
bvaughn: > To me `setIsExpanded` sounds like it sets a value and I don't feel that ties it to React state updates. I don't like `onExpandedChange` because it doesn't say anything about what it does, only when it is expected to be called.

I'm going to push back pretty strongly on this. "value" and "setValue" have _very specific implications_ in React components. They suggest values stored in React state, that updating will trigger re-renders. I don't think we should use that naming convention for this.

hbenl: OK, I'll change it tomorrow (perhaps to `persistIsExpanded`?)

bvaughn: I like that name!

PR #8685, opened 2023-02-03
Refactor sidebar button logic and fix collapsed persistence

markerikson:
The state of the sidebar panel was originally completely kept in Redux, in a `startPanelCollapsed` field. At some point, with the migration to the new "resizeable panels" library, we stopped using that field to drive the UI.  However, it was still getting persisted and read in one place: `<ToolbarButton>`

Somehow we ended up with two different sets of checks inside the the click handler in `ToolbarButton`. The mismatched logic led to the click handler thinking that the panel was already open when it was actually closed, so sometimes a click on a button wouldn't do anything.

I've refactored the logic to pull the toggle handling up into `<Toolbar>`, removed the Redux `start/endPanelCollapsed` fields and all the persistence related to those, and added persistence of the component state `sidePanelCollapsed` field via `useLocalStorage`.

Also found out that we had a keyboard shortcut ( `CTRL-B` ) that used to toggle the sidebar, but that broke with the panel migration.  Per discussion, killing that.

### DIFF
--- a/src/ui/components/Toolbar.tsx
+++ b/src/ui/components/Toolbar.tsx
@@ -58,39 +58,17 @@ function ToolbarButton({
   icon,
   label,
   name,
-  sidePanelRef,
+  onClick,
   showBadge,
 }: {
   icon: string;
   label: string;
   name: PrimaryPanelName;
-  sidePanelRef: RefObject<ImperativePanelHandle>;
+  onClick: (name: PrimaryPanelName) => void;
   showBadge?: boolean;
 }) {
   const selectedPrimaryPanel = useAppSelector(selectors.getSelectedPrimaryPanel);
-  const panelCollapsed = useAppSelector(selectors.getPaneCollapse);
-  const dispatch = useAppDispatch();
 
-  const handleClick = (panelName: PrimaryPanelName) => {
-    if (panelCollapsed || (selectedPrimaryPanel == panelName && !panelCollapsed)) {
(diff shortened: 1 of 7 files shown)


src/ui/components/Toolbar.tsx
```diff
@@ -159,6 +138,32 @@ export default function Toolbar({
...
+
+    if (!samePanelSelected) {
+      trackEvent(`toolbox.primary.${panelName}_select`);
+      dispatch(actions.setSelectedPrimaryPanel(panelName));
+    }
+
+    if (shouldTogglePanel) {
+      trackEvent(`toolbox.toggle_sidebar`);
+      togglePanel();
+    }
```
bvaughn, line 164: Sorry if this is a silly question, but why are we checking this conditional before toggling? Shouldn't we always toggle expanded/collapsed when a panel header is clicked?

markerikson: This isn't panel headers, this is the sidebar buttons.

If you go from, say, "Info" to "Pause", and the sidebar was already open, then there's nothing to toggle (which is part of the trickiness here because we've got imperative "toggle" behavior rather than "ensure this is open/closed" logic).

bvaughn: Oh right. Thanks for the reminder. 😄 Reviewing with bad connectivity (so I'm not trying the code out like I normally would) and I got those mixed up. My bad.

PR #8697, opened 2023-02-07
Add "jump to line of code" click target to events panel

jonbell-lot23:
Loom of it in action:
https://www.loom.com/share/c96ffced5e2143e2b59471356905c555

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ storybook-static/
 test/.yarn/cache/
 *.mp4
 videos/
+.vercel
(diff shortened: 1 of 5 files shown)


jasonLaster:
Pushed some small fixes which will help us get this out the door. For example, let's do the `showToast` thing in a separate PR.

jasonLaster:
Will land this because it's a big step forward and I'd like to have it for 2 demos today. As a follow up

1. some buttons are not formatted
2. Do we want to use more consistent labels for the console / event timeline 
3. I'd prefer when we unfurl to keep the icon in the same place, curious what others think

markerikson:
Another bug: we're showing it on _all_ list items, but that includes "navigation" events. We should only be showing it on click and keypress event list items.

jonbell-lot23:
Filed a new issue here for follow-up: https://linear.app/replay/issue/DES-318/events-follow-up-polish

PR #8699, opened 2023-02-07
Rewrite frames context menus to add focus-related menu items 

markerikson:
This PR:

- Migrates the `Group` and `Frame` components from the "Call Stack" panel to function components
- Replaces the old legacy context menu with our new `useContextMenu`
- Adds logic to find the point/time of a stack frame's location
- Adds context menu items for "Set focus start/end at frame" and "Jump to frame"

image

`<Group>` component context menus only have the last two menu items.


packages/replay-next/components/Icon.tsx
```diff
@@ -173,6 +174,9 @@ export default function Icon({
       path =
         "M3 17.25V21h3.75L17.81 9.94l-3.75-3.75L3 17.25zM20.71 7.04c.39-.39.39-1.02 0-1.41l-2.34-2.34c-.39-.39-1.02-.39-1.41 0l-1.83 1.83 3.75 3.75 1.83-1.83z";
       break;
+    case "empty-placeholder":
```
bvaughn, line 177: Hm. This feels like an odd choice to me. We can just use a spacer `<div />` with `width: 1rem; height: 1rem;` which I think is more semantically clear.

bvaughn (commented):
I think we should adjust this icon as I mentioned here: https://discord.com/channels/779097926135054346/1072641433443508445/1072673222396498031

jonbell-lot23 (approved):
Seconding Brian's points and what we talked about on Discord: I think once we figure out the right way to do icons and make those changes, this is looking good!

packages/replay-next/components/Icon.tsx
markerikson: Looks like it needs to be `1.5rem` to line up, but yeah.

PR #8702, opened 2023-02-08
When seeking to a console message, only open the source if this doesn't hide the console (FE-941)

hbenl:


packages/replay-next/src/contexts/LayoutContext.tsx
```diff
@@ -0,0 +1,10 @@
+import { createContext } from "react";
+
+export type LayoutContextType = {
+  // whether console and sources can be shown at the same time
+  canShowConsoleAndSources: boolean;
```
bvaughn, line 5: tiny nit: I wonder if `sourceViewerVisible` might be a clearer name? The only thing that uses this _is the Console_ view code, so it's kind of obvious that it's visible.

bvaughn (approved):

bvaughn:
Nice fix!

PR #8706, opened 2023-02-08
Modernize sources tree context menu

markerikson:
This PR:

- Removes the dead `FrameMenu` file
- Cleans up the imports in `SourcesTree` to use specific imports instead of the `actions` and `selectors` barrel imports
- Migrates `<SourcesTreeItem>` to be a function component
- Replaces the source tree context menu with our current `useContextMenu` implementation
- Deletes the dead legacy context menu implementation entirely
- Fixes the "Copy" icon usage for the stack frames context menu that I forgot to switch in the last PR


bvaughn (approved):
Deleted code 🥳

PR #8708, opened 2023-02-08
Implement "view React component source"

markerikson:
The React DevTools UI has support for "view component source".  Provide an appropriate callback prop, and the RDT logic will give you the ID of the selected component.  From there, you can talk to the RDT "backend" logic, send it a `"viewElementSource"` message, and it will save a reference to the component as `window.$type`.  We can then evaluate code in the paused browser to retrieve that function reference, and do our usual "map the location and jump to the source" steps.

For function components, it jumps straight to the function.  For class components, it jumps to `render()`.  This behavior came straight from the extension:

https://github.com/facebook/react/blob/v18.0.0/packages/react-devtools-extensions/src/main.js#L194-L220

Results:

image

image


jasonLaster:
💃

bvaughn (approved):
Awesome! This is great!

Looks like an RDT integration test is failing though. Didn't look at it but we should make sure it's not a regression before landing.

markerikson:
Yeah, I'll review that one in the morning.

markerikson:
Test definitely fails for me locally, so I broke something. Investigating...

markerikson:
Apparently wrapping `createReactDevTools()` in a `useMemo()` is bad somehow 🤷‍♂️

PR #8713, opened 2023-02-09
Handle errors in the TypeAheadPopup (FE-1216)

hbenl:


packages/replay-next/components/lexical/plugins/typeahead/TypeAheadPopup.tsx
```diff
@@ -45,7 +47,15 @@ export default function TypeAheadPopUpSuspends<Item>({
...
+    items = findMatches(queryData.query, queryData.queryAdditionalData);
+  } catch (errorOrPromise) {
+    if (isThennable(errorOrPromise)) {
+      throw errorOrPromise;
+    }
+    console.error("Failed to find matches for the TypeAheadPopup", errorOrPromise);
+  }
 
   return (
     <Suspense>
```
bvaughn, line 61: Just noticing this now, but given this component may _already_ suspend due to code completion `findMatches` function, this inner `<Suspense>` wrapper is kind of useless.

bvaughn (commented):
I guess this change is okay but it's also masking the real underlying error, which seems bad. (I added a few comments to [7f9fb5f3-2b30-4f89-a64d-caebd3c0ff33](https://app.replay.io/recording/cypressintegrationallgroupingspecjs--7f9fb5f3-2b30-4f89-a64d-caebd3c0ff33).)

Can we log this sort of error to Sentry as well so it's not invisible?

bvaughn (approved):

bvaughn:
I think this change is good since it's better to show _nothing_ than to crash.

My main concern is that we not lose visibility into this type of issue, so can we track it somewhere (Sentry) rather than just the console?

I also filed a backend issue about this because I think the `CommandError` seems unexpected given what the recording shows.

packages/replay-next/components/lexical/plugins/typeahead/TypeAheadPopup.tsx
hbenl: I noticed that too but thought I must be missing something. I'll remove it.

hbenl:
> My main concern is that we not lose visibility into this type of issue, so can we track it somewhere (Sentry) rather than just the console?

I'll send it to Sentry.

PR #8714, opened 2023-02-09
Clear timers on component unmount to prevent memory leaks

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of replayio's devtools, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some uncleared timers, causing the memory to leak (screenshots below).

[before]
before Screen Shot 2023-02-10 at 5 12 11 AM

Hence we added the fix by clearing the timers on component unload, and you can see the heap size and # of leaks reducing noticeably:
 <br />
Screen Shot 2023-02-10 at 5 00 13 AM

Note. Static properties are used in classes to avoid losing the context of the correct ‘this’ in case any nesting changes are made to the enclosing functions in future.

We executed the unit test suite after the fixes and it successfully passed 43 of 43 (in non-watch mode) and 14 of 14 (in watch mode) respectively. 

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[devtools-scenario-memlab.txt](https://github.com/replayio/devtools/files/10701446/devtools-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from React's internal objects, and hence were ignored.

### DIFF
--- a/packages/design/AddCommentButton/index.tsx
+++ b/packages/design/AddCommentButton/index.tsx
@@ -1,6 +1,6 @@
 import classNames from "classnames";
 import type { MouseEventHandler } from "react";
-import { forwardRef, useRef, useState } from "react";
+import { forwardRef, useRef, useState, useEffect } from "react";
 import mergeRefs from "react-merge-refs";
 import { Transition } from "react-transition-group";
 
@@ -47,6 +47,12 @@ export const AddCommentButton = forwardRef<HTMLButtonElement, AddCommentButtonPr
       },
       className
     );
+
+    useEffect(() => {
+      return () => {
+        clearTimeout(hoverRef.current);
+      };
+    }, []);
 
     return (
       <div className={styles.Root}>
--- a/packages/replay-next/components/lexical/CommentEditor.tsx
+++ b/packages/replay-next/components/lexical/CommentEditor.tsx
@@ -155,18 +155,23 @@ export default function CommentEditor({
 
   // Toggle editor's enabled state when the prop changes.
   useEffect(() => {
+    let timeOutID: any = null;
     const editor = editorRef.current;
     if (editor != null) {
       if (editor.isEditable() !== editable) {
         editor.setEditable(editable);
 
         if (editable) {
-          setTimeout(() => {
+          timeOutID = setTimeout(() => {
             editor.focus();
           });
         }
       }
     }
+
+    return () => {
+      clearTimeout(timeOutID);
+    };
   }, [editable]);
 
   // Save a backup of the editor state in case we need to cancel/discard pending changes.
--- a/src/devtools/client/inspector/markup/components/Node.tsx
+++ b/src/devtools/client/inspector/markup/components/Node.tsx
@@ -32,6 +32,7 @@ interface NodeProps {
 type FinalNodeProps = NodeProps & PropsFromRedux;
 
 class _Node extends PureComponent<FinalNodeProps> {
+  private static timeOutID: any;
   onExpanderToggle = (event: MouseEvent) => {
     event.stopPropagation();
     const { node } = this.props;
@@ -72,6 +73,7 @@ class _Node extends PureComponent<FinalNodeProps> {
       // Chrome sometimes ignores element.scrollIntoView() here,
       // calling it with a little delay fixes it
       setTimeout(() => el.scrollIntoView({ behavior: "smooth", block: "center" }));
+      _Node.timeOutID = setTimeout(() => el.scrollIntoView({ behavior: "smooth", block: "center" }));
     }
   };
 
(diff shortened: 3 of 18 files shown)


packages/replay-next/components/lexical/CommentEditor.tsx
```diff
@@ -155,18 +155,23 @@ export default function CommentEditor({
...
     if (editor != null) {
       if (editor.isEditable() !== editable) {
         editor.setEditable(editable);
 
         if (editable) {
-          setTimeout(() => {
+          timeOutID = setTimeout(() => {
             editor.focus();
           });
         }
```
bvaughn, line 168: I think the `timeoutId` variable and effect cleanup function could be moved inside of this conditional. The effect only needs to return that cleanup if it actually creates the timeout.

markerikson:
Neat, thanks for filing this!  Looks like there's some TS type issues.  If you've got time to fix those that'd be great, otherwise we can do that.

packages/design/AddCommentButton/index.tsx
```diff
@@ -48,6 +48,12 @@ export const AddCommentButton = forwardRef<HTMLButtonElement, AddCommentButtonPr
       className
     );
 
+    useEffect(() => {
+      return () => {
+        clearTimeout(hoverRef.current);
+      };
+    }, []);
```
bvaughn, line 55: Note for follow up: This code pre-dates `useDebouncedCallback` but we could probably simplify this a bit by using that hook here.

src/devtools/client/inspector/markup/components/Node.tsx
```diff
@@ -72,6 +73,7 @@ class _Node extends PureComponent<FinalNodeProps> {
       // Chrome sometimes ignores element.scrollIntoView() here,
       // calling it with a little delay fixes it
       setTimeout(() => el.scrollIntoView({ behavior: "smooth", block: "center" }));
+      _Node.timeOutID = setTimeout(() => el.scrollIntoView({ behavior: "smooth", block: "center" }));
```
bvaughn, line 76: This change seems mistaken:
* We now have _two_ `setTimeout` calls where there was one before.
* Also the use of a static variable seems like a mistake, since setting and clearing the time out are both associated with instances of the component.

src/ui/components/UploadScreen/UploadScreen.tsx
```diff
@@ -30,9 +30,13 @@ function DeletedScreen({ url }: { url: string }) {
...
   };
   useEffect(() => {
-    setTimeout(() => {
+    const timeOutID = setTimeout(() => {
       navigateToUrl();
     }, 1200);
+
+    return () => {
+      clearTimeout(timeOutID);
+    };
```
bvaughn, line 39: This component looks like it was kind of broken before since this effect had no dependencies. The added change looks okay but I think we also should fix the effect so that it doesn't re-run the effect every time it happens to be rendered. (This is for _us_ to follow up on.)

src/ui/utils/tokenManager.tsx
```diff
@@ -253,6 +254,11 @@ class TokenManager {
     );
     this.refreshTimeout = window.setTimeout(() => this.update(true), refreshDelay);
   }
+
+  componentWillUnmount() {
+    clearTimeout(this.refreshTimeout);
+    clearTimeout(TokenManager.timeOutID);
```
bvaughn, line 260: This looks like another troublesome mix of static property (`TokenManager.timeOutID`) and _component instance_ timeouts.

bvaughn (changes requested):
Thank you for sharing this PR with us!

I think the majority of the component updates (effects that clear timeouts of functions that just update component state) aren't _strictly_ necessary, since calling `setState` on an unmounted React component is a no-op– but the changes don't hurt, and avoid potential false positive React DEV warnings, so they're fine 😄 

I think in a lot of cases we might be able to simplify the code a bit by swapping in the newer `useDebouncedCallback` hook too.

src/ui/utils/tokenManager.tsx
Arooba-git: Thank you for your comments and notes @bvaughn. Actually these changes are auto-generated by our leak-fixing tool that we are working on. Our tool is not yet smart enough to detect whether the timers are associated with the class or instance (hence the static props). If you'd like, I can manually fix up these issues as per your comments. :)

bvaughn: Ah, that's interesting!

No, you don't have fix them. We could certainly do it. I just wanted to note them so we didn't merge them as-is.

bvaughn:
Looks like this PR has gotten stale so I'm going to close it.

PR #8715, opened 2023-02-09
Revamp "Protocol Viewer" to always enable for devs and show messages from recordings of Replay

markerikson:
Per FE-1214, I recently had a brainstorm. We have a "Protocol Viewer" panel that is only usable by Replay devs, and shows a neatly-formatted log of all the protocol requests and responses in the current open session.

We spend a lot of time debugging _recordings_ of Replay itself.  Wouldn't it be great if we could reuse the same `<ProtocolViewer>` component to visualize the protocol messages _from the copy of Replay we recorded_?  That would be really useful for debugging our own app.

So... I did that :)

Along the way I got a bit sidetracked and noted that we have 3.5 different implementations of tabs in this codebase in terms of CSS classes, and _no_ prebuilt components for any of those.

I ended up componentizing some of the HTML used by the tabs in the "Elements" and "Network" sub-panels.  I also noted that we have `@headlessui/react` in our codebase already, which has prebuilt headless tab management logic: https://headlessui.com/react/tabs .  So, I opted to try using that to switch between the "Live" and "Recorded" versions of the panel.

This first pass requires that you click to fetch the recorded data, doesn't cache any of it, and doesn't persist those results if you switch to a different tab or panel.  Also need to figure out how to handle cases where there's >200 analysis hits. Wanted to get what I have working up as a PR.

Overall, this PR:

- Removes the existing internal-user-only feature checkbox for enabling the `<ProtocolViewer>` ( accessible from the console as `window.app.features.logProtocol` ), and instead just checks if the current user is a Replay dev. (Brian and I agreed there's no reason to make this optional for us devs.)
- Extracts a reusable `<Tab>` component out of the HTML from those couple panels
- Updates the Protocol Viewer file to split it into:
  - `<ProtocolViewerPanel>`
    - `<LiveAppProtocolViewer>`, which reads from Redux
      - `<ProtocolViewer>`, which is now a "presentational" component
    - `<RecordedAppProtocolViewer>`, which fetches messages via analysis
- Updates `AnalysisCache.ts` to rename the `AnalysisResults` type, which is actually a callback function that returns a result and _not_ an array, to `GetAnalysisResult` for clarity
- Uses more analysis wizardry to find the protocol messages from the recording:
  - Checks the list of sources for our known `actions/session.ts` file
  - Finds the `onRequest`, `onResponse`, and `onResponseError` callback functions in that file
  - Finds the first breakable position in each function
  - Queries for all hit points of those positions
  - Uses analysis to fetch the `event` object in scope in each callback
  - Recursively downloads the `event` object to turn it into a POJO
  - Reformats the fetched event objects to match the normalized structure needed by `<ProtocolViewer>`

Here's how the "Protocol Viewer" panel looks with the "Recorded" tab selected:

image

### DIFF
--- a/packages/replay-next/src/suspense/AnalysisCache.ts
+++ b/packages/replay-next/src/suspense/AnalysisCache.ts
@@ -19,13 +19,13 @@ import {
 } from "protocol/evaluation-utils";
 import { ReplayClientInterface } from "shared/client/types";
 
-import { createWakeable } from "../utils/suspense";
+import { createFetchAsyncFromFetchSuspense, createWakeable } from "../utils/suspense";
 import { cachePauseData } from "./PauseCache";
 import { Record, STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Wakeable } from "./types";
 
 type Value = any;
 
-type AnalysisResult = {
+export type AnalysisResult = {
   executionPoint: ExecutionPoint;
   isRemote: boolean;
   pauseId: PauseId | null;
@@ -33,7 +33,7 @@ type AnalysisResult = {
   values: Value[];
 };
 
-export type AnalysisResults = (timeStampedPoint: TimeStampedPoint) => AnalysisResult | null;
+export type GetAnalysisResult = (timeStampedPoint: TimeStampedPoint) => AnalysisResult | null;
 
 export type RemoteAnalysisResult = {
   data: { frames: Frame[]; objects: ProtocolObject[]; scopes: Scope[] };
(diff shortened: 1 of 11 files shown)


jasonLaster:
Nice job!

Out of curiosity, why do we need the fetch button?

markerikson:
@jasonLaster because I was trying to repeatedly re-fetch in local dev while changing the code every time.

Now that the logic has stabilized, I can switch it over to being in one of our caches, and fetching that as soon as you switch over to this tab.

As for the feature flag: currently in prod, the feature flag _only_ has meaning for Replay devs.  The "Advanced" settings panel is only shown in the settings dialog for Replay internal users.  Even if a non-Replay user were to set `window.app.features.logProtocol = true`, there's another check of `userInfo.internal` that will bail out of rendering the toolbar button.

Given that, all I've changed here is that Replay devs _always_ get that toolbar button and there's no specific setting.

If we _want_ to expose the protocol viewer to non-Replay users we can do that too. I'm just making it easier for _us_ to use it atm.

jasonLaster:
Thanks for the explanation. 
- sounds like we're in agreement that default on works
- I'd love for other folks to be able to see it too. Perhaps we remove the unnecessary `userInfo.interal` check and then it's just a matter of letting folks also enable settings from the console... or get to the advanced panel

PR #8718, opened 2023-02-10
Sort by point instead of time (FE-1229)

hbenl:
`mergeSortedPointLists()` asserted that the lists were sorted by `point` and then merged them sorted by `time`, so it's no surprise that this fails sometimes (e.g. https://app.replay.io/recording/unsorted-points-assertion--8732a7ee-c197-4ac0-a24c-938dd187e824).
`time` should not be used for *anything* other than user display, it's simply not precise and reliable enough for anything else!

### DIFF
--- a/src/ui/utils/timeline.ts
+++ b/src/ui/utils/timeline.ts
@@ -336,10 +336,10 @@ export function mergeSortedPointLists(
     if (destinationIndexPoint == null) {
       destination.push(sourcePoint);
       sourceIndex++;
-    } else if (sourcePoint.time === destinationIndexPoint.time) {
+    } else if (sourcePoint.point === destinationIndexPoint.point) {
       // Don't add duplicates.
       sourceIndex++;
-    } else if (sourcePoint.time < destinationIndexPoint.time) {
+    } else if (BigInt(sourcePoint.point) < BigInt(destinationIndexPoint.point)) {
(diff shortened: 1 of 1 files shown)


jcmorrow (approved):
Good catch!

markerikson (approved):

src/ui/utils/timeline.ts
```diff
@@ -336,10 +336,10 @@ export function mergeSortedPointLists(
     if (destinationIndexPoint == null) {
       destination.push(sourcePoint);
       sourceIndex++;
-    } else if (sourcePoint.time === destinationIndexPoint.time) {
+    } else if (sourcePoint.point === destinationIndexPoint.point) {
       // Don't add duplicates.
       sourceIndex++;
-    } else if (sourcePoint.time < destinationIndexPoint.time) {
+    } else if (BigInt(sourcePoint.point) < BigInt(destinationIndexPoint.point)) {
```
markerikson, line 342: :question: We've got utils for comparing points already - should we use those here?

hbenl: Yeah, we have several of those and use them inconsistently. Originally they were introduced when we tried to avoid `BigInt` for Safari compatibility but that doesn't seem to be a concern anymore. :shrug:

PR #8719, opened 2023-02-10
Add generic range cache (FE-1231)

hbenl:
and use it to reimplement the hitpoints cache.

The generic range cache can be used to cache anything that can be fetched for a point range. If parts of a requested range have already been cached, it will only fetch the missing parts of the requested range. It is the first step towards the new analysis cache.


bvaughn (commented):
Do you mind holding off on landing this for a day or two? It touches a few of the points files– and while the conflicts would be minimal, I'd love to avoid making that PR any more complicated to land.

hbenl:
@bvaughn No problem, I'm not in a hurry landing this.

hbenl:
I still have to figure out how to teach the new `ReplayClient.streamAnalysis()` to the recorder/player for our replay-next tests.

packages/protocol/analysisManager.ts
```diff
@@ -166,7 +167,10 @@ class AnalysisManager {
       this.handlers.set(analysisId, handler);
       await Promise.all([
         !handler.onAnalysisResult || client.Analysis.runAnalysis({ analysisId }, sessionId),
-        !handler.onAnalysisPoints || client.Analysis.findAnalysisPoints({ analysisId }, sessionId),
+        !(handler.onAnalysisPoints || handler.onPointsFinished) ||
+          client.Analysis.findAnalysisPoints({ analysisId }, sessionId).then(
+            handler.onPointsFinished
+          ),
```
bvaughn, line 173: This looks wrong, I think.

If you pass both `onAnalysisPoints` but not `onPointsFinished` – this will add a handler for `onPointsFinished`?

packages/shared/client/types.ts
```diff
@@ -215,6 +211,14 @@ export interface ReplayClientInterface {
...
     onMatches: (matches: SearchSourceContentsMatch[], didOverflow: boolean) => void
   ): Promise<void>;
+  streamAnalysis(
+    params: AnalysisParams,
+    handlers: {
+      onPoints?: (points: PointDescription[]) => void;
+      onResults?: (results: AnalysisEntry[]) => void;
+      onError?: (error: any) => void;
+    }
+  ): { pointsReceived: Promise<void>; resultsReceived: Promise<void> };
```
bvaughn, line 221: nit: Would "finished" be clearer than "received"? Seems like "received" could potentially mean either "first result received" or "all results received"

src/ui/components/FocusContextReduxAdapter.tsx
```diff
@@ -24,6 +34,14 @@ export default function FocusContextReduxAdapter({ children }: PropsWithChildren
     });
   }, [focusRegion, loadedRegions]);
 
+  const rangeForAnalysis = useMemo(
+    () =>
+      deferredFocusRegion
+        ? { begin: deferredFocusRegion.begin.point, end: deferredFocusRegion.end.point }
+        : { begin: "0", end: endpoint },
+    [deferredFocusRegion, endpoint]
+  );
```
bvaughn, line 43: This just a convenience value? I guess that makes sense, since we're using it so many places.

bvaughn (commented):
Haven't dug into the actual range code yet– is this ready for revieW?

packages/protocol/analysisManager.ts
hbenl: The `onAnalysisPoints()` handler will be called by the `AnalysisManager.onAnalysisPoints()` method which receives the `Analysis.analysisPoints` events. But the backend will only send these events to us if we call the `Analysis.findAnalysisPoints` command, that's why the `onAnalysisPoints()` handler appears in this condition.

src/ui/components/FocusContextReduxAdapter.tsx
hbenl: > This just a convenience value? I guess that makes sense, since we're using it so many places.

Exactly. I had put it in a hook at first but then I thought that it makes sense to add it to the `FocusContext` :shrug:

hbenl:
> is this ready for revieW?

Yes, I just need to fix the tests before merging.

packages/protocol/analysisManager.ts
bvaughn: I don't think I understand. Those things seem kind of far apart from each other to be an uncommented dependency in this function. (Also maybe it's just me but I really have trouble parsing the way this code is written.)

Could we write it without as much `!` and `||` to maybe make it more clearer?
Is this the same code?

```js
const promises: Promise<any>[] = [];
if (handler.onAnalysisResult) {
  promises.push(client.Analysis.runAnalysis({ analysisId }, sessionId));
}
if (handler.onAnalysisPoints && handler.onPointsFinished) {
  promises.push(
    client.Analysis.findAnalysisPoints({ analysisId }, sessionId).then(
      handler.onPointsFinished
    )
  );
}
await Promise.all(promises);
```

hbenl: Yes, this is the same code except we still need to say `if (handler.onAnalysisPoints || handler.onPointsFinished)` instead of `if (handler.onPointsFinished)`.

bvaughn: Can you replace the above with something like this then, for readability?
```js
const promises: Promise<any>[] = [];
if (handler.onAnalysisResult) {
  promises.push(client.Analysis.runAnalysis({ analysisId }, sessionId));
}
if (handler.onAnalysisPoints || handler.onPointsFinished) {
  // We check onAnalysisPoints too even though it isn't directly referenced here
  // because unless we call Analysis.findAnalysisPoints,
  // that callback won't be called.
  promises.push(
    client.Analysis.findAnalysisPoints({ analysisId }, sessionId).then(
      handler.onPointsFinished
    )
  );
}
await Promise.all(promises);
```

packages/replay-next/src/suspense/HitPointsCache.ts
```diff
@@ -0,0 +1,138 @@
...
+
+function createHitPointsCache(location: Location, condition: string | null) {
+  return createGenericRangeCache<TimeStampedPoint>(
+    async (client, range, cacheValues, cacheError) => {
+      const locations = client.getCorrespondingLocations(location).map(location => ({
+        location,
+      }));
+
+      if (condition) {
+        const mapper = `
```
bvaughn, line 33: Shouldn't this be typed code now, after @markerikson's recent change?

packages/replay-next/src/utils/range-list.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function isInRange(b: bigint, range: Range) {
+  return b >= range.begin && b < range.end;
+}
+
+export function rangesOverlap(range1: Range, range2: Range) {
+  return range1.end > range2.begin && range1.begin < range2.end;
+}
+
+export function rangeContains(range1: Range, range2: Range) {
```
bvaughn, line 22: nit: Name one of these something like `containerRange` or something so it's clear which directly the args are expected in? (I often finds "contains" type functions confusing when I'm using them).

packages/replay-next/src/utils/range-list.ts
```diff
@@ -0,0 +1,96 @@
...
+  return range1.begin <= range2.begin && range1.end >= range2.end;
+}
+
+export function getMissingRanges(requested: Range, cached: Range[]): Range[] {
```
bvaughn, line 26: Functions like this should have associated unit tests– particularly to cover edge cases.

packages/replay-next/src/utils/range-list.ts
```diff
@@ -0,0 +1,96 @@
...
+      begin: point,
+      end: min(cached[index].begin, requested.end),
+    });
+    point = cached[index].end;
+    index++;
+  }
+  return missing;
+}
+
+export function mergeRanges(ranges1: Range[], ranges2: Range[]): Range[] {
```
bvaughn, line 54: Same RE: unit tests

packages/replay-next/src/suspense/createGenericRangeCache.ts
```diff
@@ -0,0 +1,227 @@
...
+        requestedRange,
+        mergeRanges(store.cachedRanges, store.pendingRanges)
+      );
+
+      const failedRange = failedRanges.find(failedRange =>
+        missingRanges.some(missingRange => rangeContains(missingRange, failedRange.range))
+      );
+      if (failedRange) {
+        throw failedRange.error;
+      }
```
bvaughn, line 83: Won't this potentially give us false positives?

If we try to fetch all events for a whole recording, and it fails, then wouldn't any subsequent focus ranges be flagged as failed?

packages/replay-next/src/suspense/createGenericRangeCache.ts
```diff
@@ -0,0 +1,227 @@
...
+
+        pendingRange.promise = new Promise<void>(async (resolve, reject) => {
+          pendingRanges.push(pendingRange);
+          store.pendingRanges = mergeRanges(store.pendingRanges, [missingRange]);
+
+          let failed = false;
+          function onError(error: any) {
+            if (!failed) {
+              failed = true;
+              failedRanges.push({ range: missingRange, error });
```
bvaughn, line 99: Maybe I'm reading this wrong, but I _think_ we should only gate the call to `reject` in `if (!failed)` but we should always push failed ranges?

These generic caches should have unit tests too I think, that explicitly cover edge cases like this.

packages/shared/client/ReplayClient.ts
```diff
@@ -928,6 +776,32 @@ export class ReplayClient implements ReplayClientInterface {
...
+  ): { pointsReceived: Promise<void>; resultsReceived: Promise<void> } {
+    let resolvePointsReceived!: () => void;
+    const pointsReceived = new Promise<void>(resolve => (resolvePointsReceived = resolve));
+    let resolveResultsReceived!: () => void;
+    const resultsReceived = new Promise<void>(resolve => (resolveResultsReceived = resolve));
+
+    this._waitForRangeToBeLoaded(params.range || null).then(() =>
+      analysisManager.runAnalysis(params, {
+        onAnalysisPoints: handlers.onPoints,
+        onAnalysisResult: handlers.onResults,
```
bvaughn, line 795: Naming nit: Why is a param called "on results" (plural) re-mapped to something called "on result" (singular)?

packages/replay-next/src/suspense/HitPointsCache.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        await client.streamAnalysis(
+          {
+            effectful: false,
+            locations,
+            mapper,
+            range,
+          },
+          {
+            onResults: results => {
```
bvaughn, line 77: I know the analysis manager pre-dates this PR, but I'm confused about the difference between `onResult` (used here) and `onPoints` (used below). Looks like we could almost avoid this fork entirely (and just pass a mapper or an empty string) _except_ for that.

I guess one calls `Analysis.findAnalysisPoints` and the other calls `Analysis.runAnalysis`. Is one more efficient than the other?

packages/replay-next/src/suspense/createGenericRangeCache.ts
hbenl: No, it's the other way around: only ranges that completely contain a failed range will be flagged as failed. (Yeah, I find contains type functions confusing as well)

packages/replay-next/src/suspense/createGenericRangeCache.ts
hbenl: If `failed` is `true` then we have already pushed the `missingRange` to `failedRanges`, no need to do that again.

bvaughn: That's only true for a single range though, right? Can't there be more than one?

packages/replay-next/src/suspense/HitPointsCache.ts
hbenl: Yes, finding analysis points is a lot faster than getting the results - the first is just a lookup in the backend while the second needs to run a browser to each point to execute the mapper. So `pointsFinished` will resolve a lot earlier than `resultsFinished`.
When we don't have a `condition` we only need to ask for the points but when we have one we need to execute it for every point to find out if that point should be included.

packages/replay-next/src/suspense/HitPointsCache.ts
hbenl: This mapper hadn't been converted yet.

packages/replay-next/src/suspense/createGenericRangeCache.ts
hbenl: Yes, there can be more than one, but since `failed` is declared within the loop for the ranges, each range gets its own `failed` flag.

bvaughn: I see.

packages/replay-next/src/suspense/HitPointsCache.ts
markerikson: Yep, this is the underlying problem with hit points and print statements to begin with, and this is also why the "breakpoints" implementation that Josh and I had come up with mid-last-year intentionally separated out the "find points" and "run actual analysis" steps in that `protocol/thread/analysis.ts` file.

bvaughn:
I forget where we landed on unit tests– but this PR adds a lot of new functionality and no new tests. (It actually _removes_ a pre-existing test.) Can you do a follow up PR with unit tests for this stuff, @hbenl?

hbenl:
Yep, I created FE-1266 for that.

PR #8725, opened 2023-02-12
Improve our comment colors

jonbell-lot23:
Comment color touch-up


jonbell-lot23:
This addresses https://linear.app/replay/issue/DES-272/fix-comment-background-color-regression-in-dark-theme

PR #8729, opened 2023-02-13
Add simpler e2e cypress test example

ryanjduffy:
* Adds simpler test which should be quicker to load
* Adds generator for cypress e2e test


jaril (approved):

ryanjduffy:
Tagging y'all on FE for some extra 👀 on the ancillary changes I had to make. @bvaughn , @markerikson , @hbenl 

* There were a couple places we're using `uuid` in both devtools proper and replay-next but we didn't have a direct dependency on it. When I brought in `@replayio/cypress` and bumped the other `@replayio` dependencies, we had some compilation issues because they're using v8 of uuid and we were assuming v7 with our config. So I've added that dependencies in both packages. It means we have two copies locally but only v7 should be included in the bundle (which we already had from `react-tooltip` and from where I assume we were pulling for our imports)
* I added `cypress` as a dev dependency for `e2e-tests` so we can regenerate replays used for the cypress e2e tests. It adds some more weight but again its only for dev deps.
* There are some config gymnastics for typescript so chai `expect()` declarations from the newly added `cypress` don't conflict with the existing jest globals. Open for improvements here but I think everything is at least working now.

tsconfig.json
```diff
@@ -87,7 +87,8 @@
   },
   "exclude": [
     "src/devtools/client/debugger/src/workers/parser/tests/fixtures/scopes/*",
-    "node_modules"
+    "node_modules",
+    "packages/e2e-tests",
```
bvaughn, line 91: This seems like a bad change. We'll lose a lot of type coverage if we exclude this whole package. Why?

package.json
```diff
@@ -154,7 +154,7 @@
     "@types/react-virtualized-auto-sizer": "^1.0.1",
     "@types/react-window": "^1.8.5",
     "@types/styled-components": "^5.1.25",
-    "@types/uuid": "^8.3.1",
+    "@types/uuid": "^7.0.3",
```
bvaughn, line 157: What's this change about?

packages/e2e-tests/cypress/support/e2e.ts
```diff
@@ -0,0 +1 @@
+import "@replayio/cypress/support";
```
bvaughn, line 1: What's this file for?

package.json
ryanjduffy: There were a couple places we're using uuid in both devtools proper and replay-next but we didn't have a direct dependency on it. When I brought in @replayio/cypress and bumped the other @replayio dependencies, we had some compilation issues because they're using v8 of uuid and we were assuming v7 with our config. So I've added that dependencies in both packages. It means we have two copies locally but only v7 should be included in the bundle (which we already had from react-tooltip and from where I assume we were pulling for our imports)

tsconfig.json
bvaughn: I see you added a nested TS config but it's not clear to me why this is needed.

packages/e2e-tests/cypress/support/e2e.ts
ryanjduffy: Assists in capturing the test steps for the cypress test.

bvaughn: I don't understand what you mean.

Can you add some inline comments to this file maybe? I would imagine others seeing this and also being confused by it too.

package.json
ryanjduffy: this specifically is aligning the types import to the version the app was actually using (v7.0.3)

bvaughn: Yeah. That makes sense. I hadn't noticed the `@types` prefix. 👍🏼

tsconfig.json
ryanjduffy: @markerikson - any ideas on how to manage this ts config better? the problem is that both jest and cypress add globals (e.g. for `expect()`) and they're conflicting. I was excluding the e2e tests to suppress the `cypress` import from the root devtools typechecking but the side effect seems to be that no typechecking is happening in that package at all now. I've tried a few other configs but nothing is achieving the desired result.

packages/e2e-tests/cypress/support/e2e.ts
ryanjduffy: I can but I don't expect it to be too foreign for anyone accustomed to how cypress configuration happens. Which, to be fair, I wasn't until all of this fun started but here we are!

bvaughn: A module that does nothing but _import_ another module (no exports, no code) _looks_ like a mistake. If I saw this module in isolation, I would assume it was a mistake and raise a PR to delete it.

Given that I'm probably not the only one who would be confused by this in our code base, I think it's probably best to leave some comments as to what it's doing?

bvaughn: > I don't expect it to be too foreign for anyone accustomed to how cypress configuration happens

Put another way– who on the frontend team would actually accustomed to how Cypress works enough to understand this? (Not me.)

tsconfig.json
ryanjduffy: @markerikson - hoping I can borrow some time from you tomorrow to get to the bottom of this

ryanjduffy: I've minimized this change to exclude the cypress-related files so the conflict is avoided. The side effect at the moment is that the cypress config and record-cypress files are not typechecked by `tsc` though they are in VS Code.

I'm going to move ahead with this because we retain the existing typechecking of everything else with this config and the impact of those files being missed is minimal.

PR #8732, opened 2023-02-13
WIP Use Babel synchronously to debug symbol parsing

markerikson:


markerikson:
Okay, conclusion here is that it's a minifier bug.  Fails to parse with a normal deploy, but works okay if I turn off minification in Webpack.

bvaughn:
Cleaning up our PR queue this morning and closing anything that seems stale or to have fallen through the cracks.

Seems like this is no longer relevant given that symbol parsing has been moved to the backend by @hbenl so I'm going to close this too. Please re-open if I'm wrong!

PR #8734, opened 2023-02-13
AvatarImage uses useTooltip hook to show user's names as Replay tooltips

bvaughn:
This finishes out the UI spec in FE-1201 and I think is probably a good general change for anywhere we have a user avatar:

Screen Shot 2023-02-13 at 2 24 24 PM

Screen Shot 2023-02-13 at 2 24 28 PM

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/AvatarImage.tsx
+++ b/packages/replay-next/components/AvatarImage.tsx
@@ -15,6 +15,7 @@ export default function AvatarImage({ className, name, src, title, ...rest }: Pr
   const tooltipOrTitle = title ?? name ?? "";
 
   const { onMouseEnter, onMouseLeave, tooltip } = useTooltip({
+    position: "above",
     tooltip: tooltipOrTitle,
   });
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
Surprised we dont show the tooltip on top. Also - would we show the avatar in the print statements panel when it's yours? Would be surprised if we do

jonbell-lot23:
> Surprised we dont show the tooltip on top. Also - would we show the avatar in the print statements panel when it's yours? Would be surprised if we do

Agree with both points. Can we align to top and make sure your own avatar doesn't show?

hbenl (approved):

bvaughn:
> Surprised we dont show the tooltip on top.

Why would we? I don't think that was part of a spec anywhere I saw, so we just showed it like a normal tooltip.

> Also - would we show the avatar in the print statements panel when it's yours? Would be surprised if we do

No. The video is from an _anonymous_ account looking at print statements created by my user.

bvaughn:
https://user-images.githubusercontent.com/29597/218746492-88d0b972-7f2a-4def-b150-58404f1173e5.mp4

PR #8736, opened 2023-02-13
Print statement v3

bvaughn:
### Features
- [x] Add "in-between" colors/states
- [x] Add "editable hit count"

### Tests
- [x] Fix existing tests
- [x] Add new screenshot tests for:
  - [x]  ~~In-between state~~ Covered by existing screenshot tests
  - [ ] Editing state
- [x] Add new e2e tests for editing current index

### Demo
https://user-images.githubusercontent.com/29597/218588817-b1828f1f-618f-446c-8864-688543fd0552.mp4

Resolves FE-1169 and FE-1170

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
+++ b/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
@@ -8,8 +8,8 @@ import useLogPointPanelContextMenu from "./useLogPointPanelContextMenu";
 import styles from "./Capsule.module.css";
 
 export default function Capsule({
+  closestHitPointIndex,
   currentHitPoint,
-  currentHitPointIndex,
   editable,
   hasConditional,
   hitPoints,
@@ -19,8 +19,8 @@ export default function Capsule({
   toggleConditional,
   toggleShouldLog,
 }: {
+  closestHitPointIndex: number;
   currentHitPoint: TimeStampedPoint | null;
-  currentHitPointIndex: number | null;
   editable: boolean;
   hasConditional: boolean;
   hitPoints: TimeStampedPoint[];
@@ -67,11 +67,11 @@ export default function Capsule({
 
   let label: ReactNode = "10k+";
   if (!tooManyPointsToFind) {
-    if (currentHitPointIndex != null) {
-      label = `${currentHitPointIndex + 1}/${hitPoints.length}`;
-    } else {
-      label = `${hitPoints.length}`;
-    }
+    label = (
+      <span className={currentHitPoint !== null ? styles.ExactMatch : styles.NearMatch}>
+        {closestHitPointIndex + 1}
+      </span>
+    );
   }
 
   return (
@@ -89,7 +89,7 @@ export default function Capsule({
             minWidth: `${minLabelWidthCh}ch`,
           }}
         >
-          {label}
+          {label}/{hitPoints.length}
         </div>
 
         {/* Use a none-standard SVG component/size to simplify layout/margin/gap styles */}
--- a/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
+++ b/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
@@ -112,6 +112,8 @@ export default function Capsule({
     selectAll(currentTarget);
   };
 
+  const inputDefaultValue = tooManyPointsToFind ? "10k" : `${closestHitPointIndex + 1}`;
+
   return (
     <>
       <div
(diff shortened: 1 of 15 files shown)


jasonLaster:
@bvaughn what do you think about using the same color we use for the disabled next/prev button? It's basically a grey with the same hue

jonbell-lot23:
> @bvaughn what do you think about using the same color we use for the disabled next/prev button? It's basically a grey with the same hue

image

Top one: design proposal
Middle one: aligning to the pencil gray
Bottom one: aligning to disabled color

It's important that this gray be visible, so I am concerned that aligning to the disabled color is too dark. The pencil shade is pretty close, but reducing the number of grays isn't as important as making sure each element looks good in its own context. I think the top one is best.

jonbell-lot23:
@bvaughn I think it's important that the progress indicator be the same color as the current hit. Right now we're using purple for the text color (my design) and fuchsia for the progress dot (the current Replay behaviour across the app) because we didn't quite land our POV on this.

image

My proposal is to go with the fuchsia color for the text (in both light and dark theme) to visually align the concepts (which is super important) and see how it feels. Maybe in the future we'll standardise on purple, but that can be a later discussion I think.

bvaughn:
I tweaked the `--point-panel-timeline-label-*` CSS variables in my latest commit to address your feedback @jonbell-lot23 

Screen Shot 2023-02-14 at 7 55 13 AM

Screen Shot 2023-02-14 at 7 55 46 AM

jasonLaster:
Colors look good.
1. next/prev buttons don't work
2. editing the value doesnt work when trying to go from one to two digits

bvaughn:
Please note that this is still marked as _draft_ and isn't ready for review 😄  I'm still testing things out locally and writing e2e tests. I only pushed the work in progress for my own self (so I can update the change log as I'm going).

bvaughn:
I think we're ready for some testing now.

packages/replay-next/components/sources/log-point-panel/Capsule.tsx
```diff
@@ -65,44 +70,108 @@ export default function Capsule({
...
+  const onChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const input = inputRef.current!;
+    const numChars = input.value.length;
+    const maxChars = `${hitPoints.length}`.length;
+    input.style.width = `${Math.min(numChars, maxChars)}ch`;
+  };
+
+  const onKeyDown = (event: KeyboardEvent) => {
+    switch (event.code) {
+      case "Enter":
```
markerikson, line 90: :hand: Apparently this does _not_ work with `Enter` on the numpad :(

If I go to https://w3c.github.io/uievents/tools/key-event-viewer.html and first press `Enter` in the main keyboard and then `Enter` in the numpad, it looks like the `event.code` field differs (`"Enter"` vs `"NumpadEnter"`):

image


Can we tweak this to work with both `Enter` keys?  (also any other places we're checking for the `Enter` key like this?)

bvaughn: Want to file a Linear issue? I don't know how much we care to support "NumpadEnter" but if we _do_ then we have a bunch of places to check for it.

Screen Shot 2023-02-14 at 2 26 14 PM

markerikson: Okay. It was certainly confusing for me when I tried to type in numeric values in the hit count edit box and then hit the numpad `Enter`, and nothing happened.

jonbell-lot23:
This has been a longstanding problem in our UX that prevents people from understanding/using print statements, thank you! Hurray! 🎉 I have feedback not for this PR, but for follow-ons.

--

1. I think it's important that the numbers count up when you hit play. If you're between hits 4 and 5, it should show a gray 4 until you pass hit 5, at which point it should turn into a gray 5. (Right now it doesn't update until playback is paused) We discussed doing this as a follow-on, Brian, just tracking here to have design feedback in one place.

2. It'd be good to have some more optimistic loading here. Right now it can feel sluggish/jittery in different scenarios. For example, when I hit play I can see the forward/backward arrows flickering because the data is still loading which is leading to a stop-start-stop-start effect on the button state/visuals. Or when I click on the timeline, there can be a noticeable wait between when I click and when the step updates. That can make it feel feel sluggish.

To be clear, I know both of these topics are likely to be technically difficult and they are nit-level issues. But from a polish/magic wand perspective, those are things that could help the feature feel that much more premium and powerful.

packages/replay-next/components/sources/log-point-panel/Capsule.tsx
bvaughn: That's fair. I haven't had a numpad in so long I can't remember.

markerikson (approved):

PR #8739, opened 2023-02-14
[Testsuites] Improve wrapping

jonbell-lot23:
We have a lot of lines that have no spaces, so we can't get by with word-wrap. I made some changes to make this read better across long lines:

Wrapping before and after-1
Wrapping before and after-2


jasonLaster:
Looks much better

PR #8743, opened 2023-02-14
Clamp times so they are within the recording (FE-1241)

hbenl:


src/ui/actions/timeline.ts
```diff
@@ -298,6 +302,13 @@ export function seekToTime(targetTime: number, autoPlay?: boolean): UIThunkActio
   };
 }
 
+const endTimeWaiter = defer<number>();
+
+async function clampTime(time: number) {
+  const recordingEndTime = await endTimeWaiter.promise;
```
bvaughn, line 308: This change seems a bit unsafe. The module exports a bunch of functions like `setFocusRegionFromTimeRange` and `seekToTime ` that now have a hard dependency on `jumpToInitialPausePoint` being called first.

bvaughn: I think maybe we should pass `replayClient` as a param to `clampTime` and it should JIT initialize the end time. I think this should be doable since `replayClient` is passed as an extra arg to our thunk things?

bvaughn (approved):

PR #8748, opened 2023-02-14
Ensure we fetch breakpoint positions before running an analysis (FE-1252)

hbenl:
I also removed some related code that is no longer used or needed.


packages/protocol/analysisManager.ts
```diff
@@ -179,35 +179,6 @@ class AnalysisManager {
     return handler.onFinished?.();
   }
 
-  async runAnalysisBatches<T>(
```
markerikson, line 182: :hand: I was actually considering possibly using this in the near future.  Can we leave this in?

PR #8750, opened 2023-02-14
Remove unused screenshot logic that led to modal crash in library

markerikson:
- `<DownloadSection>` in `<SharingModal>` was using a selector that read `state.timeline.currentTime`, but that doesn't exist in the library because we lazy-load most of our reducers when DevTools boots up, so the modal crashes when used in the library.  Looks like this "show a screenshot" functionality wasn't even being used, so I removed it.
- Replaced uses of `import * as selectors` with specific named imports for clarity.


markerikson:
Got the issue ID wrong in the branch name, recreating

PR #8752, opened 2023-02-14
Fix console freezing when there are too many event filter hits

markerikson:
Per FE-1206, the console freezes and is stuck in a loading state if you try to show events like `mousemove` that have too many hits.

The root cause was our analysis API throwing an error when >200 hits, but the real error was that `EventsCache.ts` didn't handle that case and never resolved the wakeable.  

I've added a `try/catch` there to at least resolve it.

In addition, we don't have any messaging that there were too many hits.  I've cobbled together a bandaid by creating a fake `EventLog` entry with an appropriate message and a timestamp of `0`, so it will end up at the top of the messages list:

image

We should come up with a better UX later, but this at least fixes the freezing and indicates what happened.

### DIFF
--- a/packages/replay-next/src/suspense/EventsCache.ts
+++ b/packages/replay-next/src/suspense/EventsCache.ts
@@ -5,6 +5,7 @@ import {
   Location,
   PauseData,
   PauseId,
+  Value as ProtocolValue,
   Object as RecordReplayObject,
 } from "@replayio/protocol";
 
@@ -14,7 +15,7 @@ import { ReplayClientInterface } from "shared/client/types";
 import { STANDARD_EVENT_CATEGORIES } from "../constants";
 import { createWakeable } from "../utils/suspense";
 import { cachePauseData } from "./PauseCache";
-import { Record, STATUS_PENDING, STATUS_RESOLVED, Wakeable } from "./types";
+import { Record, STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Wakeable } from "./types";
 
 export type Event = {
   count: number;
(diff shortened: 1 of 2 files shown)


jasonLaster:
I think we have another ⚠️ icon we're using already.

By the way, my comment about print statements and 200+ hits is one thing. I think we could do something here where we show the first 200, show the total count, and then have a load more button. That's much more tractable from product direction.

markerikson:
The immediate issue for the icon is that the `<EventLogRenderer>` only knows how to work with protocol `Value` types.  In this case, I'm handing it back a `Value` that's a string, so there's showing that as-is.  Icons wouldn't immediately fit in there, unless I start making changes to `<EventLogRenderer>` itself.

"Load More" is very much the type of behavior Brian and I were debating in that Discord thread. It ought to be semi-feasible to handle cases where there's >200 hits if we split up "find points" and "run the analysis".  On the other hand, loading _more_ results later would require analysis logic we haven't even tried to construct yet.

jonbell-lot23 (approved):
As Jason said, the icon's not ideal but if that's a pain let's roll this out as-is

packages/replay-next/src/suspense/EventsCache.ts
```diff
@@ -110,27 +111,52 @@ async function fetchEventTypeEntryPoints(
...
+          objects: [],
+        },
+        location: [],
+        pauseId: "",
+        point: "",
+        time: 0,
+        type: "EventLog",
+        values: [
+          {
+            value: `❗Too many messages of type "${eventType}" found!`,
```
bvaughn, line 152: The embedded "❗" character feels kind of yucky. Especially the lack of a space between "❗" and the subsequent "T" character. I don't think we should mix iconography with error message.

bvaughn (commented):
I see the appeal of showing something in the console, but I'm not a huge fan initially with the idea of catching an error and inserting a _fake_ message type.

packages/replay-next/src/suspense/EventsCache.ts
bvaughn: Can we remove this bit?

PR #8753, opened 2023-02-14
Move devtools nag into Replay Tour location

jonbell-lot23:
**The problem we're trying to solve**

We are improving our "high funnel" features that help developers get into our product and find value from it. This feature is moving our "click on DevTools!" banner from the top right of the screen to the top left, with more space.

**Old version:**

image

**New version:**

image

Zoomed out:
image

**Loom walkthrough for NUX:**

Loom: https://www.loom.com/share/cd9a2d57a4a344d29594346f8b205021

**Additional details**

This is step one for our Replay Tour, and addresses Linear ticket [DES-290](https://linear.app/replay/issue/DES-290/1-move-click-the-devtools-blue-badge-to-the-top-left-corner-of-the-app). [Read more about Replay Tour 2.0 here](https://www.notion.so/replayio/Replay-Tour-2-0-85e71db5569d45439690408e3d9d63ae?pvs=4).

Here's a grid where we're tracking the various things we can light up with Replay Tour: 
image

[See more in the Figma document here](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=16125%3A100501&t=O6gLm2iZQhQ3bwIz-4)

### DIFF
--- a/src/ui/components/SidePanel.tsx
+++ b/src/ui/components/SidePanel.tsx
@@ -4,31 +4,27 @@ import PrimaryPanes from "devtools/client/debugger/src/components/PrimaryPanes";
 import SecondaryPanes from "devtools/client/debugger/src/components/SecondaryPanes";
 import Accordion from "devtools/client/debugger/src/components/shared/Accordion";
 import { setSelectedPrimaryPanel } from "ui/actions/layout";
+import { setViewMode } from "ui/actions/layout";
 import Events from "ui/components/Events";
 import { shouldShowDevToolsNag } from "ui/components/Header/ViewToggle";
 import SearchFilesReduxAdapter from "ui/components/SearchFilesReduxAdapter";
+import MaterialIcon from "ui/components/shared/MaterialIcon";
 import hooks from "ui/hooks";
 import { useFeature } from "ui/hooks/settings";
 import { Nag } from "ui/hooks/users";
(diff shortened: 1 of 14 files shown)


src/ui/components/SidePanel.tsx
```diff
@@ -65,17 +87,27 @@ export default function SidePanel() {
...
+          <h2>Welcome to Replay!</h2>
+          <p>To get started, click into DevTools so we can show off some time travel features!</p>
+          <button type="button" onClick={() => handleToggle("dev")} style={{ padding: "5px 12px" }}>
+            <div>Open DevTools</div>
+            <MaterialIcon style={{ fontSize: "16px" }}>arrow_forward</MaterialIcon>
+          </button>
+        </div>
+      )}
+
+      <div className="h-full w-full overflow-hidden rounded-lg bg-bodyBgcolor text-xs">
```
ryanjduffy, line 102: Nit: I prefer to use `h-0 flex-grow` instead of `h-full`. `h-full` works because `flex-shrink` defaults to `1` but I prefer the explicit "no natural height but grow to fill the space" vs "try to be 100% but you can shrink"

ryanjduffy (approved):

jasonLaster:
This is probably not blocking, but I think we can spruce up the copy

PR #8754, opened 2023-02-15
Add replay to snapshot tests

ryanjduffy:
Blocked by #8805 

* Uses replay browser to record and upload failing snapshot tests
* Sets the suite name for both snapshot and e2e tests so they both are rendered with headers in the replay bot comment


jasonLaster (commented):
Exciting! When will we see the github comment? Does this mean we get two comments for the price of one?

ryanjduffy:
Looks like it overwrites the comment. :(

ryanjduffy:
Converting this to draft for now. I don't think we can land this as is until we handle running multiple tests for the same PR (SCS-827).

ryanjduffy:
@bvaughn - One alternative would be to remove the `--reporter` configuration which would mean that we wouldn't see the test panel in the replay (because we wouldn't capture metadata to know the steps, result, etc) but then it also wouldn't be detected as a test suite replay by the comment bot so there wouldn't be a conflict between the e2e tests with replay and the screenshot tests with replay. To find the replays of the failed screenshot tests, you'd have to open the action details and scroll to the end to see the links.

We'll eventually get this fixed but it isn't going to happen in the next few weeks I expect.

Thoughts? If that doesn't seem helpful, we can keep this as draft or close it and reopen it when we have a better story around multiple test runs for the same PR.

ryanjduffy:
The comment is a bit extra long on this PR because the comment bot can't distinguish between a renamed suite for this PR and an entirely new test for this PR.

jasonLaster:
@ryanjduffy ship it! 😉

bvaughn (approved):
@ryanjduffy I don't have a strong opinion one way or another 😄  If you think this is good, let's try it and see!

bvaughn:
PS It does look like it's _failing_ on GitHub though?

bvaughn:
This seems to have broken the ability for us to run these tests locally.
> Error: chromium is not supported on this platform

Can we fix it by changing it to?
```diff
diff --git a/packages/replay-next/playwright/playwright.config.ts b/packages/replay-next/playwright/playwright.config.ts
index 9a6ca0f4d..b2a2bf415 100644
--- a/packages/replay-next/playwright/playwright.config.ts
+++ b/packages/replay-next/playwright/playwright.config.ts
@@ -22,7 +22,6 @@ const config: FullConfig = {
   use: {
     browserName: "chromium",
     launchOptions: {
-      ...devices["Replay Chromium"].launchOptions,
       slowMo,
     },
 
@@ -36,6 +35,14 @@ const config: FullConfig = {
   timeout: 30_000,
 };
 
+if (CI) {
+  // @ts-ignore
+  config.use.launchOptions = {
+    ...devices["Replay Chromium"].launchOptions,
+    slowMo,
+  };
+}
+
 if (VISUAL_DEBUG || RECORD_PROTOCOL_DATA) {
   config.workers = 1;
 }
```

PR #8755, opened 2023-02-15
Log point panel: Don't show 0/0 while hit points are loading

bvaughn:
Show loading text "–" instead (for lack of anything better).

### Before

Kapture 2023-02-14 at 23 09 33

### After

Kapture 2023-02-14 at 23 03 36

---

Ignore the timing differences above as they are somewhat arbitrary. The key difference is showing "0/0" vs "–" when hit points data is still loading.

Unfortunately we can't prevent some amount of size jumping (without knowing how many hits to allow space for). I could perhaps estimate more closely than I'm currently doing (e.g. assume single digits) but it didn't seem worth it for a temporary fix.

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/Capsule.module.css
+++ b/packages/replay-next/components/sources/log-point-panel/Capsule.module.css
@@ -12,6 +12,14 @@
   font-variant-numeric: tabular-nums;
 }
 
+.Capsule[data-loading] {
+  padding: 0 2ch;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--color-dimmer);
+}
+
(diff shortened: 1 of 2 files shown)


jasonLaster (approved):
Looks good. Although I wonder if we can use the line hit count for the badge width

bvaughn:
> Although I wonder if we can use the line hit count for the badge width

I don't understand what you mean?

If we had this hit count, we wouldn't need to show the "–" loading indicator.

jasonLaster:
🤦 carry on

bvaughn:
😆  It's late. Happens to us all.

PR #8756, opened 2023-02-15
Defer loading Pause and Frame info for print statement and console inputs

bvaughn:
Speeds up various parts of the UI:
* Print statement panels now defer loading Pause and Frame information until _required_ by the type-ahead input (aka when a user actually starts typing)
* Console now defers loading pause and frame info until the user starts typing _or_ enters a terminal expression (at which point the information is needed to evaluate and render the expression)

Here are some before/after demos:
* [Print statements panel](https://www.loom.com/share/24e68cf418ae44e6a5186dd2c23d651a)
* [Console (and input)](https://www.loom.com/share/68b2e42071374ae1bff351bed7e8b343)

cc @jcmorrow

### DIFF
--- a/packages/replay-next/src/utils/suspense.ts
+++ b/packages/replay-next/src/utils/suspense.ts
@@ -1,6 +1,10 @@
+import { isThennable } from "shared/proxy/utils";
+
 import { Wakeable } from "../suspense/types";
 
-let CIRCULAR_THENABLE_CHECK_MAX_COUNT = 1_000;
+type AnyFunction<ReturnType> = (...args: any[]) => ReturnType;
+
+let MAX_LOOP_COUNT = 1_000;
 
 // A "thennable" is a subset of the Promise API.
 // We could use a Promise as thennable, but Promises have a downside: they use the microtask queue.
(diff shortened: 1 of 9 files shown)


packages/replay-next/src/utils/suspense.ts
```diff
@@ -98,13 +102,36 @@ export function createWakeable<T>(debugLabel: string): Wakeable<T> {
   return wakeable;
 }
 
-type AnyFunction = (...args: any[]) => any;
+export async function convertSuspenseToAsyncAwait<ReturnType>(
```
markerikson, line 105: :hand: this looks like it's doing a superset of what I'd added in `createFetchAsyncFromFetchSuspense` at the bottom of this file :)  Should we replace all references to `createFetchAsyncFromFetchSuspense` with this function instead?

bvaughn: Ah, good point. I thought there was a helper like this– and I searched for it! Guess I used the wrong search keywords.

Sure!

jasonLaster (approved):

PR #8757, opened 2023-02-15
Fix display cases where `error.message` is a getter

markerikson:
Apparently in some cases `error.message` can be a getter. We were trying to stringify `messageProperty.value`, and if it's a getter there _isn't_ a `.value` field from the protocol.

Addressed this by fetching the getter contents and rendering that.

This _doesn't_ fix the "copy to string" use case, but that seems rare enough we can skip it for now.

As an example, see https://app.replay.io/recording/fe-1233-babel-parsing-errors--13a939a5-703c-400c-a491-27323947d6ad?point=7463926734574730657574443599003984&time=3613 , where there's some nested `SyntaxError`s from Babel.

Result:

image


packages/replay-next/components/inspector/values/ErrorRenderer.tsx
```diff
@@ -1,14 +1,53 @@
...
+        "message"
+      );
+      errorContent = (
+        <ValueRenderer
+          context="nested"
+          layout="vertical"
+          pauseId={pauseId}
+          protocolValue={getterValue}
+        />
+      );
```
bvaughn, line 40: Should we use the `GetterRenderer` for this case instead?

packages/replay-next/components/inspector/values/ErrorRenderer.tsx
```diff
@@ -18,5 +57,6 @@ export function errorProtocolObjectToString(protocolObject: ProtocolObject) {
     property => property.name === "message"
   );
 
+  // TODO We'd need to suspend / await to find the getter value. Skip that case for now.
```
bvaughn, line 60: Could you add more detail to this comment? Like a link the Linear issue at least? It's kind of vague.

bvaughn (approved):

packages/replay-next/components/inspector/values/ErrorRenderer.tsx
markerikson: I tried that, but it includes the field name, so you end up with something like:

`SyntaxError:  message:  something went wrong`

The `message` field name seems unnecessary - we just want to show the message value itself, and we don't want differing behavior between when there's `messageProperty.value` and a getter.

packages/replay-next/components/inspector/values/ErrorRenderer.tsx
markerikson: There's no separate issue.  This is just me noting that the logic you wrote for doing the "copy to string" functionality _appears_ to be synchronous at first glance, so I didn't want to follow that rabbit hole and try to do something async here.

packages/replay-next/components/inspector/values/ErrorRenderer.tsx
bvaughn: Ah, understood.

packages/replay-next/components/inspector/values/ErrorRenderer.tsx
bvaughn: Okay fair. I was just saying that this comment could use more detail. It's pretty vague.

markerikson: Double-checked how the "copy to string" logic works, and I see that the parent function _is_ `async` already.  So, I went ahead and converted this function to be `async` and do the same "fetch the getter" logic as the component does.

bvaughn: Nice 👍🏼

PR #8758, opened 2023-02-15
Don't update current time marker while editing Focus mode

bvaughn:
### [Loom demo](https://www.loom.com/share/959bd0e024364ff3a9bb24646319245c)

cc @jonbell-lot23

### DIFF
--- a/src/ui/components/Timeline/Focuser.tsx
+++ b/src/ui/components/Timeline/Focuser.tsx
@@ -1,7 +1,7 @@
 import classNames from "classnames";
 import React, { useEffect, useRef } from "react";
 
-import { setTimelineToTime, updateDisplayedFocusRegion } from "ui/actions/timeline";
+import { updateDisplayedFocusRegion } from "ui/actions/timeline";
 import { selectors } from "ui/reducers";
 import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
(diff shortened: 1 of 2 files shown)


jasonLaster:
@bvaughn looks great

jonbell-lot23:
Hey @bvaughn check out this Loom question: https://www.loom.com/share/b86f1eeeed2646c899cfdfbe04cb8cdf

I think it's important that the red dot always be equal to what we're showing in the seek/video region. Thoughts?

bvaughn:
> I think it's important that the red dot always be equal to what we're showing in the seek/video region. Thoughts?

No, at least I don't agree.

If you're _dragging_ a focus handle– you'll see a preview of where the handle is. (The handle you're dragging is also red.) This is important when you're trying to focus in _around_ a point you're paused at– which is something I do often

The _dot_ is where you're paused at. If that moves when you click to drag, then you _can't_ focus in around a point (because the point would jump as soon as you started dragging). That was our previous UX and it could be pretty annoying in my opinion.

jonbell-lot23:
I agree the dragging dot can be frustrating. But I wonder if we can do anything about the red dot feeling misleading now.

Maybe the red dot can turn blue while dragging? That way we're not claiming that red == current video location during the drag. That would address my concern while not going back to the frustrating behaviour. Would that work?

PR #8762, opened 2023-02-15
Replay Tour: Add logged out experience

jonbell-lot23:
Logged out

Addresses [DES-291](https://linear.app/replay/issue/DES-291/2-our-logged-out-state-needs-a-welcome-to-replay-box) and is related to https://github.com/replayio/devtools/pull/8753 .

Right now the links are going to our standard login and https://docs.replay.io/recording-a-web-app/viewing-and-collaborating. I've filed https://linear.app/replay/issue/DES-328/create-a-better-quickstart-guide-link to come up with better quickstart content. (And we'd want different Quickstart content based on whether or not you're using testsuites)


jasonLaster:
One of our goals has been to provide a good logged out experience where you can add print statements and prompt folks to log in w/ comments. Why are we moving this up when we could use this top of funnel action to help people get started w/o the friction of google sso?

jonbell-lot23:
> One of our goals has been to provide a good logged out experience where you can add print statements and prompt folks to log in w/ comments. Why are we moving this up when we could use this top of funnel action to help people get started w/o the friction of google sso?

Sorry, I don't follow. I think you're saying you don't want a login button there, but want a better way to get people started?

I think better quickstart guides (one for testsuites, one for standard replays) is a good way to do that, and this PR (in concert with [better documentation](https://linear.app/replay/issue/DES-328/create-a-better-quickstart-guide-link) ) is setting up that improvement.

jasonLaster:
Probably worth discussing in a video call. 

* Not sure what the larger plan is for the Tour
* Not sure when we'd show the go to devtools vs login + quick start.

jonbell-lot23:
Follow-up: 

Jason pointed out that if we want people to use the product without the friction of logging in, we shouldn't lead with a log in button. I totally agree. Also: this message telling people to log in is missing something important: a link to log in! I'll be adding that.

image

jasonLaster:
I wonder if the empty comment panel could be a part of the Replay Tour work. 

i.e. prompt folks to add a comment in the video.

jasonLaster:
Nice! I bet we can land this incrementally too

src/ui/components/SidePanel.tsx
```diff
@@ -104,12 +106,34 @@ export default function SidePanel() {
...
+            style={{ padding: "4px 8px" }}
+          >
+            <div className="mr-1">Documentation</div>
+            <MaterialIcon style={{ fontSize: "16px" }}>arrow_forward</MaterialIcon>
+          </button>
+        </div>
+      )}
+
+      {!isAuthenticated && info.isTestSuiteReplay && (
+        <div className={styles.TourBox}>
```
ryanjduffy, line 126: This would be a good candidate to refactor into a shared component

src/ui/components/SidePanel.tsx
```diff
@@ -104,12 +106,34 @@ export default function SidePanel() {
...
+          <h2>Welcome! 👋</h2>
+          <p>We've written some docs to get the most out of Replay testsuites. Check them out!</p>
```
ryanjduffy, line 128: testsuites => test suites ?

ryanjduffy (approved):

PR #8763, opened 2023-02-16
Source search UX tweak

bvaughn:
### [Loom demo](https://www.loom.com/share/14b4988c94d54730a3cd3e821c6c98e8)

Hide search highlight when source search panel is closed. Remember previous search when panel is re-opened. (Matches VS Code.)


packages/replay-next/components/sources/SourceSearch.tsx
```diff
@@ -142,7 +142,14 @@ export default function SourceSearch({
...
             className={styles.ResultsIconButton}
             data-test-id="SourceSearchClearButton"
-            onClick={searchActions.disable}
+            onClick={() => {
+              searchActions.disable();
+
+              const container = containerRef.current;
+              if (container) {
+                container.focus();
+              }
```
bvaughn, line 151: This change restores focus on the Source (so if you type CMD+F again, the source search will re-open). It's just a little tweak I noticed while poking around.

PR #8767, opened 2023-02-16
Show failed event analysis as inline error

bvaughn:
Resolves FE-1257 and FE-1395.

### [Loom overview](https://www.loom.com/share/e7699912d1aa41a7b36b6795de92b16d)

cc @jonbell-lot23


bvaughn:
I want to tweak the UX for this one more now that the filter-counts-by-range change has landed 👍🏼

packages/replay-next/src/suspense/createGenericCache.ts
```diff
@@ -24,6 +30,10 @@ export interface GenericCache<TExtraParams extends Array<any>, TParams extends A
   addValue(value: TValue, ...args: TParams): void;
   getStatus(...args: TParams): CacheRecordStatus | undefined;
   getCacheKey(...args: TParams): string;
+  subscribeToStatus: (
```
markerikson, line 33: :speech_balloon: I have to say that it _really_ feels like we're reinventing Redux at this point :)

bvaughn: Redux manages data fetching and subscriptions too, so yeah for sure there are similarities.

This stuff is all being built on top of React core primitives though– and without any constraints of broader public use cases (or legacy usage patterns) so I think this is still an okay direction for us.

jasonLaster:
@bvaughn UX looks great

bvaughn:
Rebasing this so keep it up to date but I think it's still blocking on #9009

bvaughn:
I think this should be unblocked now, at long last.

hbenl (approved):

packages/replay-next/src/suspense/EventsCache.ts
```diff
@@ -166,10 +166,12 @@ const CountEvents: Partial<Record<RecordingTarget, CountEventsFunction>> = {
...
-      [key: string]: {
-        count: number;
-        rawEventTypes: string[];
-      };
+      [key: string]:
+        | {
+            count: number;
+            rawEventTypes: string[];
+          }
+        | undefined;
```
bvaughn, line 174: cc @markerikson @hbenl I would have expected TypeScript to catch this as potentially `undefined`, but [it doesn't](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgQwA4C4YG8BQMYG0DWApgJ5rQBOAlmAOYC6aWOOoArmFGmK3AEaHkA3NhgBfYaJgBeDOMyZQkWADcEAG1aFp8ZLgBEAMxAg9dYQHpzMAAJQIAWkIAPJIWBRH5ciHIxVGrUoIGCQQKEIOSnU1Yhh2ABNCA2pCeMx-TQA6Ng4AalzBIA).

The only way I know to fix this is to [mark the value as possibly undefined](https://www.typescriptlang.org/play?ssl=5&ssc=3&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgLJwA4ElzXk5AbwChlkBtAawgE8AuZAZzClAHMBdBksshAewCu4BiEEBbAEbQA3KWQBfZAB9kwgCYQYoCOrkK5xASGbJxmBumy5YiFAF4iB4kf4mwyAG5wANoIdmmOQARDD8-MEccgD00cgAAmCMALQQAB4YEAhgqVBQ-FBevv7IwIzIGPyQ4MC+PjRqIJraILrE3n4QAHQCwmAA1P0yQA).

PR #8771, opened 2023-02-16
Defer subject node pauses until needed

ryanjduffy:
## Issue

We're currently proactively creating pauses and running evaluations for each test step to retrieve the subject DOM node IDs so we can highlight the nodes on hover and show the step details as soon as possible. This is causing some consternation on the backend for tests with many steps.

## Resolution

As a first step, this defers creating that pause until it's required by either clicking or hovering a step.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestStepItem.tsx
@@ -1,5 +1,5 @@
 import classNamesBind from "classnames/bind";
-import React, { useCallback, useContext, useEffect, useRef, useState } from "react";
+import React, { useCallback, useContext, useEffect, useMemo, useRef, useState } from "react";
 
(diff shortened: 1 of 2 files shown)


jasonLaster (commented):
Can't follow this code, but I wonder if either

1. we should debounce hover
2. remove the hover behavior because it's inconsistent with the rest of devtools and noisy

ryanjduffy:
I don't think the hover is the culprit here since it's not the trigger for creating pauses on mount but does require the same data to do it's job. I'm not particularly concerned with the highlighting on hover but i'll note that we still need the pause on hover in order to paint the right screenshot so removing the highlighting behavior won't address the performance concerns unless we're also willing to accept inaccurate screenshots (which I don't think we are).

jaril:
Clicking a step might not be a good preload signal, since we don't highlight the node while you're paused on that step. But there's not much we can do about that.

Either way, hover should lighten the load here 🤷‍♂️

jaril (approved):

PR #8773, opened 2023-02-17
Refactored createGenericCache to remove required length param

bvaughn:
This reverses the order of params passed to the fetch function so that the "extra" params come at the end. This seems like a reasonable trade off.

Possible alternate approach to #8774.


bvaughn:
Looks like we've exceeded GitHub's API rate limits somehow so Vercel deploys aren't working 😢 

CI ran unit and Playwright screenshots tests, and I ran the other e2e tests locally on this branch.

bvaughn:
Superceeded by #8774

markerikson (approved):

PR #8774, opened 2023-02-17
Removed required length param from createGenericCache

bvaughn:
Makes use of [`Function.prototype.length`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) to programmatically determine the number of "extra args" passed to the _fetch_ function vs the _key_ function. The previous `extraParamsLength` param can be calculated and cached like:
```js
const extraParamsLength = fetchValue.length - getCacheKey.length;
```

This is a much smaller/nicer approach than #8773.

### DIFF
--- a/packages/replay-next/components/sources/SourceSearchContext.tsx
+++ b/packages/replay-next/components/sources/SourceSearchContext.tsx
@@ -1,7 +1,7 @@
 import { ReactNode, createContext, useContext, useEffect, useMemo } from "react";
 
 import { SourcesContext } from "replay-next/src/contexts/SourcesContext";
-import { getStreamingSourceContentsHelper } from "replay-next/src/suspense/SourcesCache";
+import { getStreamingSourceContentsAsync } from "replay-next/src/suspense/SourcesCache";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 
 import useSourceSearch, { Actions, SetScope, State } from "./hooks/useSourceSearch";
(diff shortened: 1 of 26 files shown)


bvaughn:
Looks like this change may have caused a regression in a couple of our source search screenshot tests. Going to dig into why now.

hbenl:
This sounds like a dangerous approach if it can be broken by transpiler settings.
I *think* we could get rid of that length param if we put the extra arguments at the end (i.e. `getValueSuspense(...args: [...TParams, ...TExtraParams])` instead of `getValueSuspense(...args: [...TExtraParams, ...TParams])`):
we only need the length param because `getOrCreateRecord(...args: [...TExtraParams, ...TParams])` calls `getCacheKey: (...args: TParams)`, so it needs to slice the extra arguments from its `args`. But if we put the extra arguments at the end, we don't need to slice them off (that means `getCacheKey` will be called with more arguments than its signature expects but I think that's OK).
(FWIW the only reason why I put the extra arguments at the beginning was because I couldn't figure out at the time how to do the Typescript types otherwise)

markerikson:
Yeah, some thoughts here:

- The `function.length` thing doesn't help at the TS level, which was part of the goal of Holger's change
- I think this would also require that `getCacheKey` _always_ have all of its parameters declared.  What if the cache gets `(a, b, c, d)` as its args, but the `getCacheKey` function only needs `(a, b) => a + b`, and thus `c, d` get omitted from the `getCacheKey` declaration?  That seems like a footgun
- The default values thing also seems problematic

bvaughn:
> This sounds like a dangerous approach if it can be broken by transpiler settings.

It can't be broken by compiler settings. I think maybe you misread the final edit on my comment above.

> But if we put the extra arguments at the end, we don't need to slice them off (that means getCacheKey will be called with more arguments than its signature expects but I think that's OK).

PR #8773 did this. If you two prefer that approach, we could go with that PR instead.

hbenl:
Personally I'd prefer the approach in #8773.

bvaughn:
Okay. I'm fine with that.

PR #8776, opened 2023-02-17
Replace several uses of `Pause.getAllFrames` with `getTopFrame`

markerikson:
The backend/runtime teams have noted that returning all frames can be expensive due to serializing all the frame-related object data.  They have plans to fix things on their end.  But, per discussion with Domi and Logan, there may still be at least _some_ benefit to having the front-end skip calling `getAllFrames` if possible.

We have several places in our code where we _only_ care about the top frame.  Or, in the case of `<CurrentLineHighlight>`, the 95% use case is that the user is in the top frame and hasn't specifically selected another frame earlier in the stack.  So, this PR refactors those cases to focus on getting the top frame specifically.


packages/replay-next/src/suspense/FrameCache.ts
```diff
@@ -37,3 +37,40 @@ export function getFrameSuspense(
...
+    // that comes back _should_ be the same either way.
+    const framesResult = await client.getTopFrame(pauseId);
+    const { frame: frameId } = framesResult;
+
+    if (frameId === undefined) {
+      return;
+    }
+    await client.waitForLoadedSources();
+    cachePauseData(client, pauseId, framesResult.data);
+    const topFrame = framesResult.data?.frames?.find(frame => frame.frameId === frameId);
```
hbenl, line 71: The frame still needs to have its location(s) updated (to account for corresponding locations), like [here](https://github.com/replayio/devtools/blob/f929de2bb80b9facf629571f0f6bcafa8fe4215b/packages/replay-next/src/suspense/PauseCache.ts#L120-L126)

hbenl (changes requested):

hbenl (approved):

PR #8777, opened 2023-02-17
Add popup and hover tooltips to warn if we can't jump to code

markerikson:
We're agreed that the right solution long-term is to precalculate "does this event have a valid jumpable location?" in a routine, but until then it's nice to have a tooltip warning you that you can't jump for this event.

I was able to cobble together both a 5-second popup and an on-hover version of the tooltip:

**[Loom demo here](https://www.loom.com/share/ffa8f8b315f74d648c4fccaf74d3ad82)**

To test this out:

- Open https://devtools-git-feature-fe-1221-jump-code-toast-recordreplay.vercel.app/)/recording/redux-todos-app-with-events--8835992a-2708-4c59-928e-85ea1f0216a5?point=32451855387679743845332202100359236&time=15602.54918032787
- Scroll down to the first two "Click" events, and click "Jump to Code"

image

### DIFF
--- a/packages/replay-next/src/hooks/usePositionedTooltip.tsx
+++ b/packages/replay-next/src/hooks/usePositionedTooltip.tsx
@@ -10,48 +10,35 @@ import {
 } from "react";
 
 import Tooltip from "replay-next/components/Tooltip";
-import useDebouncedCallback from "replay-next/src/hooks/useDebouncedCallback";
 
-export type TooltipPosition = "above" | "below" | "left-of" | "right-of";
+import { MARGIN_LARGE, MARGIN_SMALL, TooltipPosition } from "./useTooltip";
 
-export const MARGIN_LARGE = 8;
-export const MARGIN_SMALL = 4;
-
-export default function useTooltip({
+// A copy-pasted version of the other `useTooltip` hook,
+// but exposing `setShowTooltip` for control from the parent component
+// instead of controlling display itself on hover.
+// TODO Find a better way to consolidate these or extract common logic
+export function usePositionedTooltip({
   className,
   containerRef,
-  delay = 0,
   position,
   tooltip,
+  targetRef,
 }: {
   className?: string;
   containerRef?: RefObject<HTMLElement>;
-  delay?: number;
-  position?: TooltipPosition;
+  position: TooltipPosition;
+  targetRef: RefObject<HTMLElement>;
   tooltip: ReactNode;
 }): {
-  onMouseEnter: (event: MouseEvent) => void;
-  onMouseLeave: (event: MouseEvent) => void;
-  onMouseMove: (event: MouseEvent) => void;
+  setShowTooltip: Dispatch<SetStateAction<boolean>>;
   tooltip: ReactNode;
 } {
   const tooltipRef = useRef<HTMLDivElement>(null);
 
-  const [mouseCoordinates, setMouseCoordinates] = useState<Pick<
-    MouseEvent,
-    "clientX" | "clientY"
-  > | null>(null);
-
-  const [mouseTarget, setMouseTarget] = useState<HTMLElement | null>(null);
-
   const [showTooltip, setShowTooltip] = useState(false);
 
-  const setShowTooltipDebounced = useDebouncedCallback((value: boolean) => {
-    setShowTooltip(value);
-  }, delay);
-
   useLayoutEffect(() => {
-    if (mouseCoordinates === null || mouseTarget === null || showTooltip === false) {
+    if (!targetRef.current || showTooltip === false) {
       return;
     }
 
@@ -59,6 +46,8 @@ export default function useTooltip({
     if (tooltip === null) {
       return;
     }
+
+    const mouseTarget = targetRef.current;
 
     const container = containerRef?.current ?? document.body;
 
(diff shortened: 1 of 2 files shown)


packages/replay-next/src/hooks/useTooltip.tsx
```diff
@@ -8,6 +17,129 @@ type TooltipPosition = "above" | "below" | "left-of" | "right-of";
 export const MARGIN_LARGE = 8;
 export const MARGIN_SMALL = 4;
 
+// A copy-pasted version of the other `useTooltip` hook,
+// but exposing `setShowTooltip` for control from the parent component
+// instead of controlling display itself on hover.
+export function usePositionedTooltip({
```
bvaughn, line 23: This seems like an awkward mix of default and named exports. Why is this in the same file?

packages/replay-next/src/hooks/useTooltip.tsx
```diff
@@ -8,6 +17,129 @@ type TooltipPosition = "above" | "below" | "left-of" | "right-of";
...
+  tooltip,
+  targetRef,
+}: {
+  className?: string;
+  containerRef?: RefObject<HTMLElement>;
+  position: TooltipPosition;
+  targetRef: RefObject<HTMLElement>;
+  tooltip: ReactNode;
+}): {
+  setShowTooltip: Dispatch<SetStateAction<boolean>>;
```
bvaughn, line 36: This feels awkward to me. I think maybe a more straight forward approach would be to update `useTooltip` to accept an optional flag `forceVisible` or something that would override the internal state when present.

```ts
export default function useTooltip({
  className,
  containerRef,
  delay = 0,
  forceVisible = false,
  position,
  tooltip,
}: {
  className?: string;
  containerRef?: RefObject<HTMLElement>;
  delay?: number;
  forceVisible?: boolean;
  position?: TooltipPosition;
  tooltip: ReactNode;
}): {
  onMouseEnter: (event: MouseEvent) => void;
  onMouseLeave: (event: MouseEvent) => void;
  onMouseMove: (event: MouseEvent) => void;
  tooltip: ReactNode;
} {
```

Then I don't think we'd need to fork at all?

bvaughn (changes requested):

packages/replay-next/src/hooks/useTooltip.tsx
markerikson: This and the other comment are the same reason: I was going for the simplest thing that would work, you said it made sense to fork it, and given that we might re-combine them later it didn't seem worth putting this in another file.

packages/replay-next/src/hooks/useTooltip.tsx
markerikson: Okay, I'll give that a shot.

packages/replay-next/src/hooks/useTooltip.tsx
bvaughn: Not knowing what you had in mind, forking seemed reasonable. Seeing this proposed implementation though, it seems like overkill. I dunno 😄

packages/replay-next/src/hooks/useTooltip.tsx
markerikson: @bvaughn : ah, I remember why I really ended up copy-pasting this.

`useTooltip` _expects_ that you're using its `onMouseEnter/onMouseLeave` callbacks in order to get a reference to the target element.  It doesn't have a way to pass that in.

It was easier to fork the entire hook and swap that out to be an argument instead.

bvaughn: k. Fork it then, but I think it should be in its own file. Mixing named and default exports like this seems confusing to me.

Seems like we could share the _position_ logic between the two though.

bvaughn: I guess the _other_ options would be to change `useTooltip` to accept a `targetRef` prop but then that makes the regular "tooltip" use cases a little more verbose so maybe it's not worth it.

I dunno. End of a long week and my brain is feeling friend tbh.

markerikson: Okay. Personally I've shifted to not being a fan of default exports in the first place :)  But that's a separate discussion.

And yeah, extracting the positioning logic was what I started with, but that's also its own rabbit hole and I didn't want to deal with that for now.

bvaughn: > Personally I've shifted to not being a fan of default exports in the first place

Yeah. I think I've been coming around to that position myself, although default exports do have a nice benefit of speeding up file navigation for me in VS Code (e.g. Cmd+P -> type name of function).

jasonLaster:
The UX seems fine
1. I would not change the cursor - we dont do that elsewhere and it's a bit awkard
2. It'd be nice to improve the placement of the tooltip. cc @jonbell-lot23 
3. If we're going with a tooltip, we probably dont need the 5 seconds, it could just show up on hover for a failed jump

markerikson:
@jasonLaster : define "improve the placement of the tooltip"?

bvaughn:
> If we're going with a tooltip, we probably dont need the 5 seconds, it could just show up on hover for a failed jump

If _that's_ the case then we could just use the original `useTooltip` hook. (Is that the case? I dunno what @jonbell-lot23 and @markerikson decided here.)

packages/replay-next/src/hooks/usePositionedTooltip.tsx
```diff
@@ -0,0 +1,138 @@
...
+  position: TooltipPosition;
+  targetRef: RefObject<HTMLElement>;
+  tooltip: ReactNode;
+}): {
+  setShowTooltip: Dispatch<SetStateAction<boolean>>;
+  tooltip: ReactNode;
+} {
+  const tooltipRef = useRef<HTMLDivElement>(null);
+
+  const [showTooltip, setShowTooltip] = useState(false);
```
bvaughn, line 38: I think sharing state inside and outside of the hook this way is still a bit odd, but I can see why you do it with the hide-on-scroll behavior. 🤷🏼

bvaughn (approved):
Would be good to get clarification from @jonbell-lot23

jasonLaster:
Would defer to Jon/Brian on alignment.

jonbell-lot23:
Hey team! It's Saturday and I'm writing from a beach trip.

If it's cool with everyone, I'd like to hold off on this until next week. I can write more details later.

Quick summary:

1. Yeah, let's not change the cursor
2. The text of the button itself can contain the error, no tooltip needed

But there's no harm in holding off, so that'd be my preference to reduce back and forth.

jonbell-lot23:
@markerikson Here's what I recommend ([Figma link](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?node-id=16420%3A208823&t=wRqsi8GkLSYOUsRL-4)):
image

1. No need for a tooltip, let's use the capsule itself to message
2. No custom cursor needed, we can use a standard cursor to show it's inert (see bottom left image)

I considered changing the forward/backward icon into something like an X for this state, but it doesn't work as well as keeping the forward/backward buttons. We're already juggling a lot of concepts here, so it's important to keep things consistent where we can.

jonbell-lot23 (approved):
Love this! Sorry for the runaround where we tried tooltips, but I think it's clear that this is great way to make it as clear yet simple as possible. Ship it!

PR #8779, opened 2023-02-20
Initialize focusRegion with the initially loaded region (FE-1270)

hbenl:


packages/replay-next/components/sources/HoverButton.tsx
```diff
@@ -93,7 +93,7 @@ export default function HoverButton({
               sourceId: source.sourceId,
             },
             null,
-            rangeForAnalysis
+            { begin: focusRange.begin.point, end: focusRange.end.point }
```
jasonLaster, line 96: I wonder if we want a helper for mapping between these two datastructures?

PR #8784, opened 2023-02-21
Add bigger click target for code comments

jonbell-lot23:
**Summary**

When you're in DevTools and you click on a comment, we should load the line of code in Source Viewer. Before this change, we only took you to code when you clicked the small code callout, but now it's the whole region.

image

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -764,7 +764,7 @@
 
   --comment-card-preview-background-color-hover: #e6e6e6;
   --comment-card-preview-background-color: #f2f3f2;
-  --comment-card-source-preview-background-color-hover: #E3ECF9;
+  --comment-card-source-preview-background-color-hover: #e3ecf9;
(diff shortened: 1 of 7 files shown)


jonbell-lot23:
Loom demo: https://www.loom.com/share/c3c12c91df724607aec44a842bd0c168

Before/after showing off the visual polish (notice that the code callout is using the same color as the current line of code)
Click on comment

jonbell-lot23:
Stay tuned @ryanjduffy , @jasonLaster wonders if we can remove the filename and line number from the blurb and put it behind a hover instead. I'll be working on that to see how it feels.

jonbell-lot23:
Ok, now we have tooltips and this should be ready for review. here's a quick Loom of the tooltip behaviour: https://www.loom.com/share/8dd2aeec8a41485980e5c0626ef229b4

jasonLaster (approved):
Looks great to me

PR #8785, opened 2023-02-22
Add new AnalysisCache and LogPointAnalysisCache (FE-1280)

hbenl:
Note that `AnalysisCache.ts` used to handle only logpoint analyses but now it's more generic and logpoint analyses were moved to `LogPointAnalysisCache.ts`.


packages/replay-next/components/console/renderers/LogPointRenderer.tsx
```diff
@@ -148,12 +138,15 @@ function AnalyzedContent({ logPointInstance }: { logPointInstance: PointInstance
     [timeStampedHitPoint]
   );
 
-  const entry = analysisResults(timeStampedHitPoint);
-  if (entry === null) {
-    console.error(`No analysis results found for execution point "${timeStampedHitPoint.point}"`);
-
-    return <span className={styles.AnalysisError}>Analysis error</span>;
-  }
+  assert(focusRange, "No focusRange in LogPointRenderer");
```
bvaughn, line 141: This _looks_ like an unsafe change in behavior. (The `range` property in `FocusContext` _can be null_.)

 We used to check and handle the case where there was no focus region.

What changed (outside of this code) that makes that check no longer necessary?

packages/replay-next/src/suspense/LogPointAnalysisCache.ts
```diff
@@ -0,0 +1,274 @@
...
+
+export function getLogPointAnalysisResultSuspense(
+  client: ReplayClientInterface,
+  range: PointRange,
+  point: TimeStampedPoint,
+  location: Location,
+  code: string,
+  condition: string | null
+): AnalysisResult {
+  if (!condition && canRunLocalAnalysis(code)) {
```
bvaughn, line 44: (I know this code existed before but as I'm reading it again after not looking for a while–) it might be nice to update `canRunLocalAnalysis` to accept both `code` and `condition` 😄  since it is kind of silly, semantically, to have a conditional for a local analysis.

hbenl: But a `condition` with a constant value doesn't make much sense, so I don't think we need to optimize for that.

packages/replay-next/components/console/renderers/LogPointRenderer.tsx
bvaughn: Could you add an inline comment to this effect? That relationship is kind of distance and might not be apparent.

packages/replay-next/src/suspense/LogPointAnalysisCache.ts
bvaughn: I agree. I'm not saying the `canRunLocalAnalysis` function should _check_ if the condition is local, just that it might be more semantically accurate if we moved the check there, e.g.:

```ts
export function canRunLocalAnalysis(code: string, condition: string | null): boolean {
  if (condition) {
    return false;
  }

  // ...
```

PR #8800, opened 2023-02-22
Remove unprocessedRegions usage

jcmorrow:
The whole `ensureProcessed` command is probably going to either get deleted or change considerably in the near future. It seems like basically all of this is not really powering any important functionality at the moment. The SourceMap visualizer has a comment about showing a loading bar, but actually removing this has no affect on the source map viewer loading screen.

Blocks https://linear.app/replay/issue/BAC-2871/make-sure-that-the-dispatcher-waiting-on-regions-is-not-a-problem

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -435,23 +435,23 @@ class _ThreadFront {
     const abilities = await this.recordingCapabilitiesWaiter.promise;
     const { result } = frameId
       ? await client.Pause.evaluateInFrame(
-        {
-          frameId,
-          expression: text,
-          useOriginalScopes: true,
-          pure: abilities.supportsPureEvaluation && pure,
-        },
-        this.sessionId!,
-        pauseId
-      )
+          {
+            frameId,
+            expression: text,
+            useOriginalScopes: true,
+            pure: abilities.supportsPureEvaluation && pure,
+          },
+          this.sessionId!,
+          pauseId
+        )
       : await client.Pause.evaluateInGlobal(
-        {
-          expression: text,
-          pure: abilities.supportsPureEvaluation && pure,
-        },
-        this.sessionId!,
-        pauseId
-      );
+          {
+            expression: text,
+            pure: abilities.supportsPureEvaluation && pure,
+          },
+          this.sessionId!,
+          pauseId
+        );
     cachePauseData(replayClient, pauseId, result.data);
 
     if (repaintAfterEvaluationsExperimentalFlag) {
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -31,10 +31,8 @@ import {
   TimeStampedPointRange,
   Value,
   findAnnotationsResult,
-  missingRegions,
   requestBodyData,
   responseBodyData,
-  unprocessedRegions,
 } from "@replayio/protocol";
 import groupBy from "lodash/groupBy";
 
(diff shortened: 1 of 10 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -455,23 +435,23 @@ class _ThreadFront {
...
-          this.sessionId!,
-          pauseId
-        );
+        {
+          expression: text,
+          pure: abilities.supportsPureEvaluation && pure,
+        },
+        this.sessionId!,
+        pauseId
+      );
```
jcmorrow, line 454: My `prettier` made these indent changes. Is it wrong or should I leave them in?

pages/recording/[id]/sourcemap/[sourceId].tsx
```diff
@@ -54,14 +54,6 @@ async function loadSourceMap(
     }
     const { sessionId } = await client.Recording.createSession({ recordingId });
 
-    // this will show a progress bar in the LoadingScreen, note that the
-    // sourcemap visualizer is shown as soon as the requested source and its sourcemap
-    // have been loaded, which may happen before the progress bar reaches 100%
-    client.Session.ensureProcessed({ level: "basic" }, sessionId);
-    client.Session.addUnprocessedRegionsListener(regions =>
-      store.dispatch(onUnprocessedRegions(regions))
-    );
```
Comment (jcmorrow) [inline pages/recording/[id]/sourcemap/[sourceId].tsx:63]: This was not actually showing a progress bar.

src/ui/reducers/app.ts
```diff
@@ -64,13 +64,6 @@ const appSlice = createSlice({
     setMouseTargetsLoading(state, action: PayloadAction<boolean>) {
       state.mouseTargetsLoading = action.payload;
     },
-    durationSeen(state, action: PayloadAction<number>) {
-      // If this is the longest duration we have seen yet, we should update the
-      // store
-      if (action.payload > state.recordingDuration) {
-        state.recordingDuration = action.payload;
-      }
-    },
```
jcmorrow, line 73: We should be using `getEndpoint` for this. I have not seen anything not work with this deleted, but lmk if there's something else I should check

packages/protocol/thread/thread.ts
markerikson: ~~Eh, if it passes lint let's leave it.~~

well lint failed :)  double-check settings I guess?

src/ui/reducers/app.ts
markerikson: I do see several references to `recordingDuration`, particularly around playback.  Looks like it also gets passed via context into some of the `replay-next` functionality.  Can we swap this out with the better API right now?

jcmorrow: Yeah I'll see what I can do in a bit!

jcmorrow: It seems like [this line](https://github.com/replayio/devtools/blob/main/src/ui/actions/timeline.ts#L115) is doing exactly what I was suggesting?

packages/protocol/thread/thread.ts
jcmorrow: Ah, shoot. Well, I'm trying out lazy.vim and it has some different lsp configs than I usually use 🤔

src/ui/reducers/app.ts
markerikson: Hmm. Seems like, it, yeah?  (Also, putting on my Redux hat: I _hate_ that we have an action named `setTimelineState`. That is literally the exact opposite mentality of how we want people to _think_ about using Redux :(  )

jcmorrow: Fair!

jcmorrow:
I'm not sure why the 🦄 would have changed with this PR? @markerikson is that a thing I should be concerned about?

src/ui/actions/session.ts
```diff
@@ -320,7 +320,6 @@ export function createSocket(
       // We don't want to show the non-dev version of the app for node replays.
       if (recordingTarget === "node") {
         dispatch(setViewMode("dev"));
-        dispatch(onLoadingFinished());
```
markerikson, line 323: How did we end up dispatching this _twice_ in the case where it's a Node recording?  Looks like.... Jason added that in June 2022.

markerikson (approved):

src/ui/actions/session.ts
jcmorrow: Haha, yeah I dunno, many hands make light work... but also sometimes the work is light and also a little confused!

Issue #8769, opened 2023-02-16
Migrate remaining uses of React-Redux `connect` to function components


markerikson:
Filed internally as FE-1222 , but copying this publicly since there's some external interest in contributing.

We currently have roughly 50 uses of `connect(`) and 10 uses of `React.Component` (with or without React. ).  I'd love to modernize all of those so that we are exclusively using the React-Redux hooks API and React function components.

Some of the remaining components will likely be tricky due to having more complex logic inside the classes, such as `<SourcesTree>`.  We can leave those for last.

A good first step would be to compile a list of all remaining files that use `connect` or class components so we know which ones need to be updated.

ChaiWithJai:
Checking through the `devtools` repository I only see XX imports and invocations of `connect` from `react-redux`. 

Here's the breakdown - 10 usages within `client/debugger`.

```
App
QuickOpenModal
Tabs
PrimaryPanes/index
QuickOpenButton
SourcesTrees
Breakpoint
BreakpointHeading
Popover
```

2 usages on 2 dynamic pages:

```
[...view].tsx
[...id].tsx
```

34 usages across 33 files in `ui/components`

```
src/ui/components/App.tsx
src/ui/components/DevTools.tsx
src/ui/components/KeyboardShortcuts.tsx
src/ui/components/SkeletonLoader.tsx
src/ui/components/WaitForReduxSlice.tsx
src/ui/components/CommandPalette/CommandPalette.tsx
src/ui/components/Comments/CommentMarker.tsx
src/ui/components/Comments/index.tsx
src/ui/components/Comments/VideoComments/CommentsOverlay.tsx
src/ui/components/Events/Events.tsx
src/ui/components/Events/ReplayInfo.tsx
src/ui/components/Header/ShareButton.tsx
src/ui/components/Header/UserOptions.tsx
src/ui/components/NetworkMonitor/index.tsx
src/ui/components/shared/Error.tsx
src/ui/components/shared/LaunchBrowserModal.tsx
src/ui/components/shared/LaunchButton.tsx
src/ui/components/shared/LoadingScreen.tsx
src/ui/components/shared/LoomModal.tsx
src/ui/components/shared/Modal.tsx
src/ui/components/shared/NewAttachment.tsx
src/ui/components/shared/PrivacyModal.tsx
src/ui/components/shared/TeamOnboarding.tsx
src/ui/components/shared/FirstReplayModal/FirstReplayModal.tsx
[...]

PR #8802, opened 2023-02-22
refact:  to function component

ChaiWithJai:
This PR is part of a larger block of work #8769 to migrate away from `connect` to use React hooks.

This PR is refactoring the `A11yIntention` component from a class-based component to function component.

### DIFF
--- a/src/devtools/client/debugger/src/components/A11yIntention.js
+++ b/src/devtools/client/debugger/src/components/A11yIntention.js
@@ -3,28 +3,22 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 //
-import React from "react";
+import React, { useState } from "react";
 
-export default class A11yIntention extends React.Component {
-  state = { keyboard: false };
+export default function A11yIntention({ children }) {
+  const [isKeyboard, setIsKeyboard] = useState(false);
 
-  handleKeyDown = () => {
-    this.setState({ keyboard: true });
-  };
+  handleKeyDown = () => setIsKeyboard(true);
 
-  handleMouseDown = () => {
-    this.setState({ keyboard: false });
-  };
+  handleMouseDown = () => setIsKeyboard(false);
 
-  render() {
-    return (
-      <div
-        className={this.state.keyboard ? "A11y-keyboard" : "A11y-mouse"}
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/A11yIntention.js
```diff
@@ -3,28 +3,22 @@
...
 
-export default class A11yIntention extends React.Component {
-  state = { keyboard: false };
+export default function A11yIntention({ children }) {
+  const [isKeyboard, setIsKeyboard] = useState(false);
 
-  handleKeyDown = () => {
-    this.setState({ keyboard: true });
-  };
+  handleKeyDown = () => setIsKeyboard(true);
```
markerikson, line 11: :hand: Both of these callbacks probably need a `const ` in front.

src/devtools/client/debugger/src/components/A11yIntention.js
```diff
@@ -3,28 +3,22 @@
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
 //
-import React from "react";
+import React, { useState } from "react";
```
markerikson, line 6: :hand: While we're at it, bonus request: can you convert this file to TypeScript?

Looking at it, I'm pretty sure that _all_ you need to do is:

- Rename it to `.tsx`
- Add a type declaration for the props that includes `children` (probably `{children: React.ReactNode}`), and use that

markerikson (changes requested):
Good start! Can you fix the callback declarations, and also try converting it to TypeScript while you're at it?

src/devtools/client/debugger/src/components/A11yIntention.js
ChaiWithJai: Roger that!

src/devtools/client/debugger/src/components/A11yIntention.js
ChaiWithJai: Good catch!

markerikson (approved):

markerikson:
Hmm. Seems like the UI snapshot tool had an upload failure, shouldn't be related.

markerikson:
Thanks!

PR #8803, opened 2023-02-22
Improve "Jump to Code" precision by finding first hit in functions

markerikson:
The original "Jump to Code" implementation limited itself to finding the function _declaration_ that handled the event, and opening that location.  This was good, but it's even nicer if we can get you to the _time_ that the function actually ran.

This means that we need to find the first breakable location _inside_ the function after its declaration, and then do hit point analysis to find the next time that breakable position was hit.

And I've got that working!

Tested this out on several recordings and it works pretty consistently.

I've also added some additional logic to the evaluated introspection logic to handle cases where a click event might be on an `<input>` that has an `onChange` (such as an `<input type="checkbox">`, even though that wouldn't normally be a prop name you'd associate with a click event otherwise.

Example:

Open up https://devtools-git-feature-fe-1228-jump-location-recordreplay.vercel.app/recording/redux-todos-app-with-events--8835992a-2708-4c59-928e-85ea1f0216a5 , click the "Info" panel icon on the left, hover over one of the events list items like "Click" and click the "Jump to Code" flyout button on the right side of the list item:

image

### DIFF
--- a/packages/replay-next/src/suspense/createGenericCache.ts
+++ b/packages/replay-next/src/suspense/createGenericCache.ts
@@ -30,6 +30,7 @@ export interface GenericCache<TExtraParams extends Array<any>, TParams extends A
   getValueAsync(...args: [...TParams, ...TExtraParams]): Thennable<TValue> | TValue;
   getValueIfCached(...args: TParams): { value: TValue } | undefined;
   getValueSuspense(...args: [...TParams, ...TExtraParams]): TValue;
+  remove(...args: TParams): void;
   subscribeToStatus: (
     callback: SubscribeCallback,
     ...args: TParams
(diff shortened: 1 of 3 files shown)


jonbell-lot23 (commented):
Hello! This is very exciting.

I am having a clunky experience when I click things from the Viewer, because it can take a long time to load the line of code. I think we need to do two things:

1. We should switch tabs to DevTools immediately
2. We should update the button to say "Loading..." so there's an indication that something is happening behind the scenes
3. Then if it fails, we have the gray button treatment we already shipped. And if it succeeds, great. The "loading" text can go back to normal ("Jump to code") and the file will open in the Source Viewer.

jasonLaster (commented):
Looks great! Would definitely add some playwright tests when you have a chance, but this is much closer to what we want to move into the routine, so excited to see this come together

jonbell-lot23 (approved):
Thumbs up!

There are follow-up tasks for the future  like what do we do if you're already on a line, and clicking this makes you go directly to that line again? But those can be considered later, this works well and I'm excited to see it live.

PR #8809, opened 2023-02-23
Fix RDT prop fetching (FE-1274)

hbenl:
Our ReactDevTools panel uses the `getJSON()` helper when forwarding a message from the RDT backend to the RDT frontend. That helper allowed the object previews to overflow, so the message properties could be incomplete.

### DIFF
--- a/src/ui/utils/objectFetching.ts
+++ b/src/ui/utils/objectFetching.ts
@@ -28,7 +28,7 @@ export async function loadObjectProperties(
   pauseId: string,
   objectId: string
 ) {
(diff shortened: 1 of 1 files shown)


jasonLaster:
@hbenl great find. Could you share a before/after replay? Also - could we update an e2e test to catch this?

hbenl:
Before: https://app.replay.io/recording/rdt-fail--23bc406f-a4f0-4fc5-becc-aca7ba66219b
After: https://app.replay.io/recording/rdt-success--cb420ee4-938b-4580-b36e-b281e5e6f98d

markerikson:
Nice catch!

PR #8814, opened 2023-02-23
refact:  to ts and functional comp

ChaiWithJai:
This PR is part of #8769 and it converts the component from a `js` file to `tsx` along with moving a function component for `Modal`.

### DIFF
--- a/src/devtools/client/debugger/src/components/shared/Modal.tsx
+++ b/src/devtools/client/debugger/src/components/shared/Modal.tsx
@@ -14,8 +14,8 @@ interface IModalProps {
   additionalClass: string;
   children: React.ReactNode;
   handleClose: () => void;
-  status: string;
   width: string | number;
+  status?: string;
 }
 
 interface ISlideProps extends IModalProps {
--- a/src/devtools/client/debugger/src/components/shared/Modal.tsx
+++ b/src/devtools/client/debugger/src/components/shared/Modal.tsx
@@ -10,27 +10,34 @@ import { Transition } from "react-transition-group";
 
 export const transitionTimeout = 50;
 
-export class Modal extends React.Component {
-  onClick = e => {
-    e.stopPropagation();
-  };
-
-  render() {
-    const { additionalClass, children, handleClose, status, width } = this.props;
-
-    return (
-      <div className="modal-wrapper" onClick={handleClose}>
-        <div
-          style={{ width }}
-          className={classnames("modal", additionalClass, status)}
-          onClick={this.onClick}
-        >
-          {children}
-        </div>
-      </div>
-    );
-  }
+interface IModalProps {
+  additionalClass: string;
+  children: React.ReactNode;
+  handleClose: () => void;
+  status: string;
+  width: string | number;
 }
+
+interface ISlideProps extends IModalProps {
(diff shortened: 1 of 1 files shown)


src/devtools/client/debugger/src/components/shared/Modal.tsx
```diff
@@ -10,35 +10,42 @@ import { Transition } from "react-transition-group";
...
+  handleClose: () => void;
+  status: string;
+  width: string | number;
+}
 
-  render() {
-    const { additionalClass, children, handleClose, status, width } = this.props;
+interface ISlideProps extends IModalProps {
+  in: boolean | undefined;
+}
```
ChaiWithJai, line 23: @markerikson I tried to read the component API and the associated invocations but not entirely sure if I captured the types properly here.

src/devtools/client/debugger/src/components/shared/Modal.tsx
```diff
@@ -10,35 +10,42 @@ import { Transition } from "react-transition-group";
...
   width = "50%",
   in: inProp,
   children,
   additionalClass,
   handleClose,
-}) {
+}: ISlideProps) {
   return (
     <Transition in={inProp} timeout={transitionTimeout} appear>
       {status => (
```
ChaiWithJai, line 51: `Slide` doesn't include `status` in its properties. Where does the reference for status come from here?

markerikson: That looks like a typical "function as a child" setup, where `<Transition>` calls `props.children(status)`

src/devtools/client/debugger/src/components/shared/Modal.tsx
markerikson: Looks like `IModalProps` doesn't actually have a `status` field or it needs to be optional, based on the TS error from the build step

src/devtools/client/debugger/src/components/shared/Modal.tsx
ChaiWithJai: So this is a `render callback` pattern? We're relying on `Transition` to run logic when the component mounts and unmounts. However, it's a heavier lift because there's the scheduling of callbacks because of CSS transitions and timeouts. All of this is extracted into `Transition` and the "function as child" pattern allows us to pass in the relevant stateful property and markup we want rendered in an easy-to-use API... and this is a pattern that would normally be a code smell in a codebase but intended for use by library authors. Am I close to understanding this concept?

ChaiWithJai: Reading the source code for `react-transition-group` makes me wonder.... this `Transition` component kinda breaks the hooks pattern. I would imagine these actions should be handled in `useLayoutEffect` and there should be some hooks for safely dispatching (and cancelling callbacks). But... this package is maintained by the core React team and is giving people access to lower level APIs like `componentWillEnter`. So I'm guessing there's some more advanced scheduling required for CSS animations/transitions that requires lower-level APIs from React Fiber for scheduling purposes and that's why this library isn't using the "hooks way" of solving this problem. Its been a few years since I've touched React and I'm slightly baffled by this.

PR #8815, opened 2023-02-23
Enable Elements panel even for runtimes that don't support repainting graphics

bvaughn:


ryanjduffy (approved):

markerikson (approved):

jasonLaster:
I think we need a new capability so that we don't show the elements panel for node

bvaughn:
Yeah.

PR #8832, opened 2023-02-26
Standardize CSS variables across design systems

jonbell-lot23:
This ticket is part CSS maintenance and part visual upgrade. As I went through our code to standardise our colors, it had the side-effect of making our visuals nicer. Here's what changed:

* Nicer preview popup
* More standardised menus (with better contrast)
* More contrast work in dark theme
* Improved selection state when searching the console

Popup
Better dark menus
Contrast
Selection state


jasonLaster:
Great changes! Especially happy to see the preview popup get some love. 

The one debatable decision for me was the blue hover bg for the console. We used that blue for the snippet last week. Is there some thinking that we'll be using it more?

jonbell-lot23:
The one debatable decision for me was the blue hover bg for the console. We used that blue for the snippet last week. Is there some thinking that we'll be using it more?

Great question -- this is actually a hover state, it's for search results. Check out this image:

image

In this case, the user is searching for something so it's brighter and more obvious. But our hover states continue to be very subtle, especially since we have the little forward/backward button following your cursor.

PR #8843, opened 2023-02-28
Fixed an error caused by accessing an empty ref

Andarist:
I'm not sure when this can actually be `null` since React's `onBlur` shouldn't be fired when unmounting (see the long-standing issue~ [here](https://github.com/facebook/react/issues/12363)).

However, I noticed this error in the console and it was pointing to this exact line:
Call stack of an error in the browser devtools. Stating that: 'Uncaught TypeError: Cannot read properties of null (reading textContent)'


hbenl:
Thanks!

PR #8844, opened 2023-02-28
Improved Editable remark UI

Roaster05:
Improved the UI of the comments and the reply by modifying the Editable Remark components and its styling

PR #8845, opened 2023-02-28
Improved the UI of buttons by adjusting their placements

Roaster05:
Some buttons were not placed in their right position and were cut off while clicking them or hovering, this issue was resolved by providing them enough spacing and margin .


Domiii:
Duplicate of #8844

PR #8847, opened 2023-02-28
Reimplement ExceptionsCache using AnalysisCache (FE-1281)

hbenl:
The new `ExceptionsCache` not only benefits from `AnalysisCache`'s improved caching but it also does analysis result streaming, i.e. the exceptions show up in the console as soon as the analysis points are received and then every such console message suspends individually until its result is received.
I didn't port `ExceptionsCache.test.ts` because it basically tested the same functionality as `createGenericRangeCache.test.ts` (except some tests for the status subscription, but that doesn't exist anymore).


hbenl:
I need to investigate some of the differences in the screenshot tests.

packages/replay-next/components/console/LoggablesContext.tsx
```diff
@@ -126,8 +131,14 @@ export function LoggablesContextRoot({
...
-  if (showExceptions) {
-    exceptions = getExceptionsSuspense(client, focusRange);
+  if (focusRange && showExceptions) {
+    try {
+      exceptions = getExceptionPointsSuspense(client, toPointRange(focusRange));
+    } catch (errorOrPromise) {
+      if (isThennable(errorOrPromise)) {
+        throw errorOrPromise;
+      }
+    }
```
bvaughn, line 141: nit: Consider using `createInfallibleSuspenseCache` for this?

Outside of the component (module level)
```js
const infallibleExceptionPointsSuspense = createInfallibleSuspenseCache(getExceptionPointsSuspense);
```

Inside of the component
```ts
let exceptions: UncaughtException[] = EMPTY_ARRAY;
if (focusRange && showExceptions) {
  exceptions =
    infallibleExceptionPointsSuspense(client, toPointRange(focusRange)) ??
    EMPTY_ARRAY;
}
```

bvaughn (approved):

PR #8852, opened 2023-03-01
Implement Replay Tour MVP

jonbell-lot23:
**Summary**

Replay Tour MVP is a replacement for our "fuchsia tour." It's meant to walk people through their first print statement. There are tangential issues like adding a tour for Cypress, showing content to non-developers, etc, but those will be follow-on PRs. [Here's a loom that discusses the latest](https://www.loom.com/share/0b8d7cd55ef746a2a6a7ba8055ec903c).

**User journey**
image

**Next steps**

In this review, I'm looking for two things:

1. General code improvement – this started as a prototype, so I suspect I'll need to tighten things
2. Copy – please propose copy changes directly into the review process. It's inefficient to chase down a bunch of screens in Figma, especially as they go through iterations. Let's track it all here in a standard GitHub review process so we can track efficiently.


yarn.lock
```diff
@@ -20867,7 +20869,7 @@ __metadata:
     next-global-css: ^1.3.1
     node-fetch: 2.x
     node-protocol: "workspace:*"
-    parse-script-tags: ^0.1.7
+    parse-script-tags: ^0.1.7
```
ryanjduffy, line 20872: Extra space here

src/ui/components/Toolbar.tsx
```diff
@@ -156,6 +165,15 @@ export default function Toolbar() {
   return (
     <div className="toolbox-toolbar-container flex flex-col items-center justify-between py-1">
       <div id="toolbox-toolbar">
+        {recording?.metadata?.test?.runner?.name !== "cypress" && showTour ? (
```
ryanjduffy, line 168: `useTestInfo` will give you the name as `runner` as well if you want to simplify this lookup.

src/ui/components/SidePanel.tsx
```diff
@@ -93,19 +97,6 @@ export default function SidePanel() {
 
   return (
     <div className="flex w-full flex-col gap-2">
-      {shouldShowDevToolsNag(nags, viewMode) && (
```
ryanjduffy, line 96: Do we need to remove the import for `shouldShowDevToolsNag` from this file after dropping this code?

src/devtools/client/debugger/src/components/Editor/NewSourceNag.tsx
```diff
@@ -7,15 +7,20 @@ import styles from "./NewSourceNag.module.css";
...
   if (!showAddBreakpointNag && !showEditBreakpointNag) {
     return null;
   }
 
   return (
-    <div className={styles.Nag}>
-      <MaterialIcon iconSize="xl">auto_awesome</MaterialIcon>
-      {showAddBreakpointNag ? "Now hover on a line number" : "Now edit your print statement"}
-    </div>
+    <>
```
ryanjduffy, line 17: Fragment isn't needed here

ryanjduffy (approved):
Some clean up notes but the code seems sound. Curious if we should offload the GIFs to static.replay.io so they're easier to update than pushing a PR. Thoughts?

jonbell-lot23:
Thanks @ryanjduffy ! Will be addressing these soon

src/devtools/client/debugger/src/components/Editor/NewSourceNag.tsx
jonbell-lot23: Funny thing, the page fails when I don't have it there. Can't figure out why! Marking as resolved for now.

src/ui/components/Toolbar.tsx
jonbell-lot23: Will address in follow-up, thanks!

src/devtools/client/debugger/src/components/Editor/NewSourceNag.tsx
ryanjduffy: Suspect you still have the `{` before `!showConsoleNavigate` which is only needed in JSX-land.

This should be valid:

```
return !showConsoleNavigate ? (<...>) : null;
```

ryanjduffy: Or, even better imo:

```
if (showConsoleNavigate) {
  return null;
}

return <...>
```

jonbell-lot23: Thanks!

jonbell-lot23:
Ok, everything is green and got reviewed so I'm merging! There will be follow-ups:

1. I'm going to move our images to static.replay
2. I have some animation tweaks I'd like to add
3. The demo could use a touch-up
4. Any regressions/issues that may come up

PR #8871, opened 2023-03-01
Add a small gradient and dropshadow to our toggle

jonbell-lot23:
You can barely spot the difference, but it adds some depth and helpful contrast.

Toggle visual improvement


jasonLaster:
Love this!

PR #8872, opened 2023-03-02
Add international date formats

jonbell-lot23:
We've had non-US customers surprised when our library lists July 9th as 7/9/2023 rather than 9/7/2023. This fixes that!

International

Love from dd/mm/yyyy New Zealand,
Jon


markerikson:
OR WE COULD JUST USE ISO8601 AND FORMAT IT AS `YYYY-MM-DD` LIKE GOD INTENDED!

jonbell-lot23:
> OR WE COULD JUST USE ISO8601 AND FORMAT IT AS `YYYY-MM-DD` LIKE GOD INTENDED!

Lol, I feel the same way about files that need to be sorted in a filesystem. 

But speaking as a New Zealander, (and I assume other commonwealth countries) they really enjoy their dd/mm/yyyy.

PR #8873, opened 2023-03-02
Adjust default textfield color

jonbell-lot23:
It was too light, now it's a little darker.


jasonLaster:
@jonbell-lot23 screenshots plz

PR #8874, opened 2023-03-03
Design system alignment across our form inputs

jonbell-lot23:
This one is flippin' enormous, but here's the summary:

* Our textfields pretty pretty consistent
* A lot of other stuff like our replay-next checkboxes weren't
* We're better aligned across the board now
* Also some freebie fixes to some of our dialogs

input alignment

checkboxes
filters
rename

### DIFF
--- a/packages/replay-next/components/console/filters/FilterToggles.module.css
+++ b/packages/replay-next/components/console/filters/FilterToggles.module.css
@@ -1,6 +1,6 @@
 .Filters {
   height: 100%;
-  padding: 0.5rem 0;
+  padding: 0.5rem 0 0 0;
   flex: 0 0 auto;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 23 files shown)


packages/replay-next/components/console/filters/FilterToggles.tsx
```diff
@@ -82,21 +82,28 @@ export default function FilterToggles() {
...
-        label="Node modules"
-        onChange={showNodeModules => update({ showNodeModules })}
-      />
-      <Toggle
-        checked={showTimestamps}
-        label="Timestamps"
-        onChange={showTimestamps => update({ showTimestamps })}
-      />
+
+      <div className="py-1">
```
bvaughn, line 86: This change caused a small regression in the console.
Screen Shot 2023-03-14 at 5 15 05 PM

Can we avoid using Tailwind styles inside of the "replay-next" package? At least for consistently within this package– we're using CSS modules.

packages/replay-next/components/console/filters/FilterToggles.module.css
```diff
@@ -1,6 +1,6 @@
 .Filters {
   height: 100%;
-  padding: 0.5rem 0;
+  padding: 0.5rem 0 0 0;
```
bvaughn, line 3: This caused a bit of a bottom padding regression in the new components.

PR #8876, opened 2023-03-03
Initialize logProtocol value from searchParam

jaril:
For the playwright test suite for user SLOs, I'm trying to capture the protocol commands. But by default, `logProtocol = false`, so the events aren't stored when playwright runs the tests.

This adds some initialization logic that checks the searchParams, so we can enable that when we start up a replay from that test suite.

@bvaughn @markerikson this does the trick, but it's not as clean as I'd like. Are there any alternatives I'm missing?

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -27,6 +27,7 @@ import tokenManager from "ui/utils/tokenManager";
 
 import "../src/base.css";
 import useAuthTelemetry from "ui/hooks/useAuthTelemetry";
+import { useFeature } from "ui/hooks/settings";
 
 if (isMock()) {
   // If this is an end to end test, bootstrap the mock environment.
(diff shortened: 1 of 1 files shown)


markerikson:
Seems okay, just gotta fix the hooks rule lint issue.

bvaughn:
This doesn't seem like something we should be doing at the `_app` level. It's a pretty far down preference.

pages/_app.tsx
```diff
@@ -83,11 +84,19 @@ function AppUtilities({ children }: { children: ReactNode }) {
 }
 function Routing({ Component, pageProps }: AppProps) {
   const [store, setStore] = useState<Store | null>(null);
+  const { update: updateLogProtocol } = useFeature("logProtocol");
   const { getFeatureFlag } = useLaunchDarkly();
 
   useEffect(() => {
+    // Check what the initial value of logProtocol should be
+    if (window) {
```
bvaughn, line 92: I think this would need to be `typeof window !== "undefined"` to avoid causing a runtime error. (I assume that's what's breaking the tests?)

bvaughn (commented):
Can we move this somewhere other than `pages/_app`? That feels like the wrong place to have arbitrary preference checks.

Is this something that the generic `useFeature` might do itself?

bvaughn:
I think this PR seems to have stalled out so I'm going to close it. Happy to review it again if you decide to pick it back up though.

PR #8877, opened 2023-03-03
Allow seeking to unloaded points (FE-1217)

hbenl:
Currently when the user tries to seek to an unloaded point, we update the current time but not the current point, resulting in an inconsistent and confusing UI (see FE-1217).
This PR:
- always updates current time *and* point even if they're in an unloaded region
- disables the Elements, ReactDevTools and ConsoleInput panels and the PreviewPopup if the current point is unloaded (Frames and Scopes were already disabled in that case)
- fixes the Frames panel so that it doesn't try to create a pause in an unloaded region and then shows "Error loading frames"
- reenables caller stacktraces for CommandErrors that are due to unloaded regions - this should help finding any other places that may try to create or use a pause in an unloaded region


markerikson (approved):

bvaughn:
@hbenl Can you follow up with a test for this?

PR #8880, opened 2023-03-04
[PoC] AI Explanations

TejasQ:
This PR is a proof-of-concept and just an experiment. I do not expect it to be merged (although if y'all want to live on the edge like that I'd be flattered). It introduces AI to Replay, having lines of code explained in comments.

https://user-images.githubusercontent.com/9947422/222916883-3ea394af-70eb-4830-bb47-b39e387c7d5b.mp4

## Motivation

Just exploring the possibilities of AI and Replay DevTools and intending to inspire some ideas in the short-medium term. I see tremendous potential here and was motivated by the great, supportive team vibe and by @jasonLaster's document about the subject.

## Try it Yourself

You'll need an OpenAI API key that you paste into `.env` as a value to key `OPENAI_API_KEY`.

## Rough Edges

1. Currently, all the context our AI has is the single file that's open. It would be far more powerful with the entire project—and we can easily send it each file—but I haven't done it here because of the following limitations:
    1. I don't know where the data lives in the code. I found [`rawTextByLine`](https://github.com/replayio/devtools/blob/494951d9cf968704f176940e326eb6ef70faa667/packages/replay-next/components/sources/SourceListRow.tsx#L114), but I couldn't find _all_ the files. If I could, I'd feed it as context to the AI which would definitely deliver better results (though the existing results are already pretty nice).
    1. OpenAI charges by _token_, which is TL;DR the length of your input prompts and output responses. Including more context will get more expensive and I'm cheap/paying for this out of my personal account so 😛 .
1. The explanation comment is posted as the currently logged in user. This is because I'm not sure I have access to the backend/database resources needed to create a machine user and hook it up here. Since it's a PoC anyway, I am using the user's `accessToken` to post the comment. This should be trivial to swap out later.
1. Requests to OpenAI are not memoized. This leads to unnecessarily duplicated requests to them when clicking "explain" on the same read-only line multiple times. Ideally, we cache this to save money and be faster.
1. There is noticeable latency when talking to OpenAI. This is not our issue, but something to be aware of.
1. This does not handle any network failure cases. 😢 
1. When the comment is posted, we don't switch to the comments pane automatically, mainly because I couldn't find the code that would do that.
1. The explain button does a weird effect on click, but I kind of like it. It's like a poor man's [genie effect](https://raw.github.com/Ciechan/BCGenieEffect/master/Screens/genieOut.gif). 😂 

## How Do We Move Forward With This?

As mentioned, I don't expect this PR to be merged but closed. IF this is something y'all feel would benefit the devtools project with some polish, I'm happy to work with y'all to get it in a mergeable state however that looks. If not, let's smile, be inspired, close this PR, and go about our lives. 😄

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -12,3 +12,5 @@ NEXT_PUBLIC_IMAGE_URL=https://opengraph.replay.io/
 FRONTEND_API_SECRET=
 # API Key to source icons from Figma
 FIGMA_TOKEN=
+# API Key to talk to OpenAI
+OPENAI_API_KEY=
--- a/packages/replay-next/components/Icon.tsx
+++ b/packages/replay-next/components/Icon.tsx
@@ -2,74 +2,81 @@ import { CSSProperties, ReactNode } from "react";
 
 import styles from "./Icon.module.css";
 
+export type IconType =
+  | "add"
+  | "ai"
+  | "arrow"
+  | "arrow-left"
+  | "arrow-right"
+  | "breakpoint"
+  | "cancel"
+  | "chevron-down"
+  | "chevron-right"
+  | "check"
+  | "close"
+  | "comment"
+  | "comments"
+  | "conditional"
+  | "confirm"
+  | "continue-to-next"
+  | "continue-to-previous"
+  | "copy"
+  | "delete"
+  | "document"
+  | "down"
+  | "eager-evaluation"
+  | "edit"
+  | "error"
+  | "fast-forward"
+  | "file"
+  | "folder"
+  | "invisible"
+  | "invoke-getter"
+  | "menu-closed"
+  | "menu-open"
+  | "pause"
+  | "preview"
+  | "print"
+  | "prompt"
+  | "protocol"
+  | "protocol-viewer"
+  | "remove"
+  | "remove-alternate"
+  | "rewind"
+  | "save"
+  | "search"
+  | "set-focus-end"
+  | "set-focus-start"
+  | "share"
+  | "source-explorer"
+  | "spinner"
+  | "terminal-prompt"
+  | "terminal-result"
+  | "toggle-off"
+  | "toggle-on"
+  | "up"
+  | "view-function-source"
+  | "view-html-element"
+  | "visible"
+  | "warning";
+
 export default function Icon({
   className = styles.DefaultIcon,
   style,
   type,
 }: {
   className?: string;
-  type:
-    | "add"
-    | "arrow"
-    | "arrow-left"
-    | "arrow-right"
-    | "breakpoint"
-    | "cancel"
-    | "chevron-down"
-    | "chevron-right"
-    | "check"
-    | "close"
-    | "comment"
-    | "comments"
-    | "conditional"
-    | "confirm"
-    | "continue-to-next"
-    | "continue-to-previous"
-    | "copy"
-    | "delete"
-    | "document"
-    | "down"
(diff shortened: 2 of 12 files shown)


jonbell-lot23:
Hey @TejasQ ! This is AMAZING and I'm going to reiterate what I said [in Discord](https://discord.com/channels/779097926135054346/1081333739109503036/1081670103202021406):

* Once we feel good enough about code quality, I am in favour of shipping: 
   - Behind a context menu (as seen in our Figma)
   - With results showing up in comments (as you wired up)
* And then we should do a bunch of tire-kicking
* And then we should iterate both UX and code

Woo!

TejasQ:
58?!

markerikson:
@TejasQ : just skimming the description, I noticed you said:

> Requests to OpenAI are not memoized. This leads to unnecessarily duplicated requests to them when clicking "explain" on the same read-only line multiple times. Ideally, we cache this to save money and be faster.

That sounds like a good case for our `createGenericCache` util, which is used for a couple dozen other caches.

TejasQ:
Fantastic! Thanks, will play!

Do we have any feature flags mechanisms in this repo?

markerikson:
@TejasQ : yeah, the primary one we've got is a `useFeature` hook that is ultimately persisted to `localStorage`, via the legacy Firefox "prefs" system  There's also `useLocalStorage` and `useIndexedDB` hooks as well.

TejasQ:
So good! Thanks, Mark!!

TejasQ:
I'm rethinking the LLM-based approach of inferring interesting variables since this is pretty trivial to do with a parser since we have the code and tokens.

What might make more sense for auto-generated print statements is:

- AI only generates the prefix (3 words about what's happening)
- We capture 2-3 variables from lexical scope
- We return `${aiReponse} ${variables.join(',)`.

At this point, one might question why even have the AI prefix. When I debug, I never `console.log("this happens", ...vars)`, but instead do `console.log({ vars })` to see a nice key/value pair. My brain already knows what's happening since it's _my_ code. 

All that to say, maybe we don't need AI-generated print statements. 😬 

The AI explainer feature makes sense because there's no way we can do that with the data we have (lexical scope), but I'm not so sure about print statements. Thoughts?

bvaughn:
Hi 👋🏼 

I am going through all open GitHub issues and PR and closing ones that seem stale or un-mergable.

Given the description and the fact that this PR [doesn't compile/build](https://vercel.com/recordreplay/devtools/5tqoeSeFW6p2CprvFd9Zwz5BrbE2), I think we should probably close it also (for now).

I think that pulling in a dependency on something like OpenAI isn't a decision we should make lightly or without discussion. For example, there are pricing/licensing concerns.

I think that one of our Tuesday afternoon architectural syncs would be a good place to start this discussion. (Side note: Would you like to be invited to this reoccurring meeting, @TejasQ?)

jasonLaster:
@bvaughn agreed that it would be nice to discuss AI enabled features in an upcoming design/product sync. 

With that said, this also feels like a good area for some exploration. The OpenAI functionality comes through an API endpoint which adds a fetch so the scope is pretty narrowly contained. And at that point, it's just new UI.

TejasQ:
> Would you like to be invited to this reoccurring meeting, @TejasQ?

Wow, I totally missed this notification. I'd love that!

bvaughn:
I don't own that meeting but I'll email Jon and ask him to add you.

Issue #8820, opened 2023-02-24
No scrollbar is present in inspector-sidebar-container


anikdhabal:
**Description:**

No scrollbar is present in inspector-sidebar-container. It is difficult to scroll horizontally as well as vertically.
**Steps to Reproduce:**
1.open Replay browser
2.Open Elements in Devtools
3.try to scroll Rules, Layout or computed list. 

**Expected Result:**
Scrollbar should be  there in order to see the complete list.
**Additional Context:**

* Screenshots:
* 
![Screenshot 2023-02-24 180446 (3)](https://user-images.githubusercontent.com/81948346/221687695-6d22f049-cab4-4f46-a7ef-b915715f70b2.jpg)

* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):

anikdhabal:
@jasonLaster can you check this.

jasonLaster:
I think this is expected if there is no mouse

anikdhabal:
If there is a mouse also the same issue occurred.
![scroollbar_issue](https://user-images.githubusercontent.com/81948346/221767330-22d23ed3-5ba7-4a77-a43e-aa78970a939a.jpeg)

anikdhabal:
<img width="716" alt="Screenshot 2023-02-28 115220" src="https://user-images.githubusercontent.com/81948346/221770202-7dfed3ec-5492-4b39-9ab7-5067a2a893e9.png">

I think this would be the proposed solution @jasonLaster

Domiii:
hi @anikdhabal,

This looks like a design change. Without being a designer, I have a few concerns:

1. The scrollbars are way too big. Can you find other scrollbars in the app and see if they are also this big? Not sure if this is intended or not.
2. I am not sure why you are hardcoding `height` to `220px` but its generally bad practice. I can imagine you found it in other places in the code? If so, please refer/link to where you found it. Also, see if you can find a better place to adapt from, where we are not hardcoding pixel values?
3. You did not go through the guidelines checklist before committing. We added that recently. For example, it asks you to run lint/prettier. Please always check all items before submitting a PR.

Roaster05:
> hi @anikdhabal,
> 
> This looks like a design change. Without being a designer, I have a few concerns:
> 
> 1. The scrollbars are way too big. Can you find other scrollbars in the app and see if they are also this big? Not sure if this is intended or not, but I'd say, it should not be like that.
[...]

PR #8881, opened 2023-03-04
Added Scrollbars in Element panel of devtools

Roaster05:
This PR is linked to the issue #8820 and the Replay for the UI change is 
https://app.replay.io/recording/facebook-log-in-or-sign-up--2d99b605-bea6-4174-8952-15a69fafc964
The actual issue was the overlapping css styling of the classes `tabs` and `panel` which were prioritized over other classes and now they have been merged with their parent classes

### DIFF
--- a/src/devtools/client/inspector/computed/components/ComputedApp.tsx
+++ b/src/devtools/client/inspector/computed/components/ComputedApp.tsx
@@ -7,9 +7,7 @@ export default function ComputedApp() {
   return (
     <div id="sidebar-panel-computedview" className="theme-sidebar inspector-tabpanel">
       <ComputedToolbar />
-      <div style={{overflow:"scroll",height:"220px"}}>
       <ComputedProperties />
-      </div>
     </div>
   );
 }
--- a/src/devtools/client/inspector/computed/components/ComputedApp.tsx
+++ b/src/devtools/client/inspector/computed/components/ComputedApp.tsx
@@ -7,7 +7,9 @@ export default function ComputedApp() {
   return (
     <div id="sidebar-panel-computedview" className="theme-sidebar inspector-tabpanel">
       <ComputedToolbar />
+      <div style={{overflow:"scroll",height:"220px"}}>
       <ComputedProperties />
+      </div>
     </div>
   );
 }
--- a/src/devtools/client/inspector/rules/components/RulesApp.tsx
+++ b/src/devtools/client/inspector/rules/components/RulesApp.tsx
@@ -124,9 +124,11 @@ export const RulesApp: FC = ({}) => {
 
     return (
       <>
+      <div style={{overflow:"scroll",height:"220px"}}>
         {pseudoElementRules.length !== 0 && renderPseudoElementRules(pseudoElementRules)}
         {styleRules.length !== 0 && renderStyleRules(styleRules)}
         {inheritedRules.length !== 0 && renderInheritedRules(inheritedRules)}
+        </div>
       </>
     );
   }, [rules, rulesQuery, renderInheritedRules, renderStyleRules, renderPseudoElementRules]);
--- a/src/devtools/client/inspector/rules/components/RulesApp.tsx
+++ b/src/devtools/client/inspector/rules/components/RulesApp.tsx
@@ -124,11 +124,9 @@ export const RulesApp: FC = ({}) => {
 
     return (
       <>
-      <div style={{overflow:"scroll",height:"220px"}}>
         {pseudoElementRules.length !== 0 && renderPseudoElementRules(pseudoElementRules)}
         {styleRules.length !== 0 && renderStyleRules(styleRules)}
         {inheritedRules.length !== 0 && renderInheritedRules(inheritedRules)}
-        </div>
       </>
     );
   }, [rules, rulesQuery, renderInheritedRules, renderStyleRules, renderPseudoElementRules]);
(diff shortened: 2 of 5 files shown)


src/devtools/client/inspector/computed/components/ComputedApp.tsx
```diff
@@ -7,7 +7,9 @@ export default function ComputedApp() {
   return (
     <div id="sidebar-panel-computedview" className="theme-sidebar inspector-tabpanel">
       <ComputedToolbar />
+      <div style={{overflow:"scroll",height:"220px"}}>
```
manavagr1108, line 10: Hey @Roaster05, I feel it's not a good practice to give inline css, rather we can change the css in `computed` css file.

src/devtools/client/inspector/rules/components/RulesApp.tsx
```diff
@@ -124,9 +124,11 @@ export const RulesApp: FC = ({}) => {
 
     return (
       <>
+      <div style={{overflow:"scroll",height:"220px"}}>
```
manavagr1108, line 127: Same goes for this styling as well. May be change the css in `rules.css`.

manavagr1108 (commented):
I am just a contributor and I had suggested the changes according to my experience, do confirm it with the maintainers and then proceed.

src/devtools/client/inspector/computed/components/ComputedApp.tsx
Roaster05: Hey Manav thanks for your feedback but the previous styling was also inline therefore I preferred not to change that if the maintainers want the css to be mentioned explicitly in the styling file i will do the same

Roaster05:
> I am just a contributor and I had suggested the changes according to my experience, do confirm it with the maintainers and then proceed.

Hey Manav thanks for your feedback but the previous styling was also inline therefore I preferred not to change that if the maintainers want the css to be mentioned explicitly in the styling file i will do the same

manavagr1108:
> > I am just a contributor and I had suggested the changes according to my experience, do confirm it with the maintainers and then proceed.
> 
> Hey Manav thanks for your feedback but the previous styling was also inline therefore I preferred not to change that if the maintainers want the css to be mentioned explicitly in the styling file i will do the same

It was inline only for `src/devtools/client/inspector/rules/components/RulesApp.tsx` not for other two changes that I have commented.

Roaster05:
> > > I am just a contributor and I had suggested the changes according to my experience, do confirm it with the maintainers and then proceed.
> > 
> > 
> > Hey Manav thanks for your feedback but the previous styling was also inline therefore I preferred not to change that if the maintainers want the css to be mentioned explicitly in the styling file i will do the same
> 
> It was inline only for `src/devtools/client/inspector/rules/components/RulesApp.tsx` not for other two changes that I have commented.

If the maintainers want i will commit the new changes

Domiii:
As I said before (and accidentally wrote it in the wrong place, that is in the linked issue, rather than here):
1. I'd recommend not hardcoding pixel values.
2. `lint`/`prettier` issues have not been fixed.

Domiii:
@Roaster05 Thanks for the hard work! I see that you are trying. But I think PR #8922 is simpler and seems to get the job done? Doesn't it?

Roaster05:
> @Roaster05 Thanks for the hard work! I see that you are trying. But I think PR #8922 is simpler and seems to get the job done? Doesn't it?

@Domiii  Actually no because I haven't modified the css for any of the classes and the pr u mentioned has modified the class which has been used in other places too rather i just paired up two classes and rather everything is same

Roaster05:
> > @Roaster05 Thanks for the hard work! I see that you are trying. But I think PR #8922 is simpler and seems to get the job done? Doesn't it?
> 
> @Domiii Actually no because I haven't modified the css for any of the classes and the pr u mentioned has modified the class which has been used in other places too rather i just paired up two classes and everything else  is same

Aimajoke:
This seems to work fine aswell,  the problem seems to be just heigh inheritance, so merging tabs div with its child also avoid it as you did in your pr, in my [pr](https://github.com/replayio/devtools/pull/8922) instead of heigh:100% it could has been heigh: inherit and it get the job done, as the class tabs was only created to these devtools sidebars. I just dont get why did you also merge the div panel with its child? I dont see any change on that part.

Roaster05:
> This seems to work fine aswell, the problem seems to be just heigh inheritance, so merging the class tabs div with its child also avoid it as you did in your pr, in my [pr](https://github.com/replayio/devtools/pull/8922) instead of heigh:100% it could has been heigh: inherit and it get the job done, as the class tabs was only created to these devtools sidebars. I just dont get why did you also merge the class panel div with its child? I dont see any visual change with that part.

Well because the overflow properties of panel class was set to hidden and setting the inherit property just for the tab class won't get this work done

Roaster05:
> @Roaster05 mmmm at first glance you got me because it sounds reasonable but take a closer look. Because of the tab-panel-box div having the same height as the panels div (its child) the property `overflow: hidden` is not doing anything, the scrollbar appears on element ul with class accordion which has the property `overflow: auto` and to me it seems to work fine after just changing the heigh of class tabs, although only tested on linux version browsers, can you send me a ss if it is not working where you tested pls.

Well actually while debugging this issue I tried to observe the effect of each of these classes and finally after seeing the results that it's the effect of both the classes as per my knowledge because in order to fix this issue I had to adjust both of them also the css for the class panels is also similar to the tabs panel which made it clear for me that what actually is the issue

bvaughn:
Thank you for this contribution 🙇🏼

I've decided to merge PR #8922 instead though because it included before and after screenshots which made it easier to review (and this PR has a merge conflict). Both solutions seem fine FWIW though and this was just a subjective choice

Issue #8789, opened 2023-02-22
Network Monitor -- Copy as CURL


jasonLaster:
We started it here, https://github.com/replayio/devtools/pull/7943

faga295:
Can I work on this?

jasonLaster:
@faga295 go for it

faga295:
@vikneshwar I want to give you some advice: If you are interested in a task and see that someone is already preparing to do it, even if the draft PR has been submitted, you should at least let them know before you prepare to do it. Instead of silently submitting the PR and causing someone’s efforts to be wasted in the end.

vikneshwar:
Hi @faga295 sorry that your effort on this PR is wasted, but I was totally unaware if this was being worked now I was aware that there is a draft PR already. I started to analyze on this way before, in fact you can see my message in the #development channel [here](https://discord.com/channels/779097926135054346/795692423513767956/1080546891453968505)

bvaughn:
Thanks @faga295 and @vikneshwar for your contributions on this 🙇🏼‍♂️  and sorry for the miscommunications on our end.

This feature should be rolling out sometime today.

PR #8885, opened 2023-03-06
Add Copy as cURL context menu to network tab

vikneshwar:
This PR fixes the #8789 
- Adds the copy as cURL context menu in the network tab
  - However, there is one issue for which we need to decide how we go about handling it
     - When you right-click and choose `Copy as cURL`  for a request without actually selecting and opening the request first, the request body won't be copied in the cURL, this is because we download the body for the particular request only when you click and select the request.So no request body is available to be copied. 
     - But once it is downloaded(by clicking on the request and trying it after a few seconds) and we try to copy, the body is copied exactly as expected
     - This also causes one more issue which we might have left unnoticed. For the first time, as soon as you select a request and navigate to Request or Response body tab, you won't be able to see any information because the socket call is yet to be completed
     - I can think of 2 solutions here (but looking for better suggestions from others):
        - Download the request and response body for all the requests when we render the network panel. But that will result in significant calls, which is useless (most users won't check all the requests). This also fixes both the above-mentioned issues
        - When the user clicks `copy as cURL`, we make the socket call for the particular request to fetch the request and response body, show the loading state until the call completes, and copy to the clipboard. Similarly, when the user clicks on a request and navigates to the request tab quickly, we can show the loading state 
- Removes the root object in the request body json view. 
  - Currently, when you view the request payload by clicking on the request tab, the body is enclosed inside the `root` object, which is not expected. This PR fixes that too. 

Replay: https://app.replay.io/recording/overboard--999b4bf9-463e-4ca5-9336-7b585e28ad5b


vikneshwar:
A couple of unit test fails when i run `npm run test`, but those are unrelated to the changes, I tried to run test in main branch and it still fails.

vikneshwar:
@jasonLaster can you review this ?

bvaughn:
> When the user clicks copy as cURL, we make the socket call for the particular request to fetch the request and response body, show the loading state until the call completes, and copy to the clipboard. Similarly, when the user clicks on a request and navigates to the request tab quickly, we can show the loading state

This (fetch on demand, if needed) seems like the right way to go.

bvaughn:
Delta is suggesting this PR also regresses another recently added context menu. For example:

Before:
image

After:
image

I think this is just indicating that you need to rebase this branch on `main` though

vikneshwar:
@bvaughn Thanks for your suggestion. I have merged the latest changes and also implemented the suggested solution. Let me know how it looks now. 

Changes done as part of the previous commit:
 - Create a hook to handle copy to clipboard state.
- Show Copying status until we get the request body from the server.
- Copy the body and show Copied status as soon as we get a response.
- For all other cases (i.e with no request body or body is already available) copy immediately and show copied status.

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -36,12 +37,14 @@ const RequestTable = ({
 
   const loadedRegions = useAppSelector(getLoadedRegions);
 
+  const [showCopied, setShowCopied] = useState(false);
```
bvaughn, line 40: Why is this state here? Doesn't look like it's being read or written.

src/ui/reducers/network.ts
```diff
@@ -22,6 +22,7 @@ export type NetworkState = {
   requestBodies: Record<string, RequestBodyData[]>;
   requests: RequestInfo[];
   selectedRequestId: string | null;
+  isCopyTriggered: boolean;
```
bvaughn, line 25: Doesn't look like anything is reading or updating this piece of reducer state?

src/ui/components/NetworkMonitor/HttpBody.tsx
```diff
@@ -89,6 +89,7 @@ const HttpBody = ({
       <ReactJson
         style={{ backgroundColor: "none" }}
         theme={theme == "light" ? "rjv-default" : "tube"}
+        name={false}
```
bvaughn, line 92: Why is this change necessary/reated?

src/ui/components/NetworkMonitor/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,86 @@
...
+          `curl '${url}' ${formatRequestHeadersForCURL(requestHeaders)}${requestBodyForCURL}`
+        );
+
+        isCopied = true;
+        shouldShowLoading = false;
+        setTimeout(() => setIsCopyTriggered(false), 3000);
+
+      } else {
+        //dispatch to fetch request body
+        dispatch(fetchRequestBody(id));
```
bvaughn, line 71: Should you be `await`ing this?

src/ui/components/NetworkMonitor/RequestTable.tsx
vikneshwar: Ah ! I missed removing the old changes. Thanks for spotting

src/ui/reducers/network.ts
vikneshwar: Same reason as above. Missed removing the old changes

src/ui/components/NetworkMonitor/HttpBody.tsx
vikneshwar: Not setting the props results in request body json wrapped inside a `root` key, I guess that is not expected. Request Body json should show up exactly what was in the request. Correct me if I am wrong. Thanks

src/ui/components/NetworkMonitor/useCopyToClipboard.tsx
vikneshwar: Moved it to `useEffect`

vikneshwar:
Hi @bvaughn addressed the review comments. Thanks

src/ui/components/NetworkMonitor/useNetworkContextMenu.tsx
```diff
@@ -1,15 +1,30 @@
...
-export default function useNetworkContextMenu(row: Row<RequestSummary>) {
+export default function useNetworkContextMenu({
+  row,
+  onClipboardCopy,
+}: {
+  row: Row<RequestSummary>;
+  onClipboardCopy: () => void;
+}) {
   const dispatch = useAppDispatch();
+  const requestBodies = useAppSelector(getRequestBodies);
```
bvaughn, line 27: Looks like this new added value is not being used by anything.

src/ui/components/NetworkMonitor/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  useEffect(() => {
+    const fetchBody = async () => {
+      const shouldGetBody = isCopyTriggered && hasRequestBody && !requestBody;
+
+      if (shouldGetBody) {
+        await dispatch(fetchRequestBody(id));
+      }
+    };
+    fetchBody();
+  }, [dispatch, hasRequestBody, id, isCopyTriggered, requestBody]);
```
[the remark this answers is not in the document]
vikneshwar: I just realized we don't need to `await` for the results here. This is a socket call similar to `selectAndFetchRequest`, the response from the socket is updated to redux store through `NEW_REQUEST_BODY_PARTS `

src/ui/components/NetworkMonitor/useCopyToClipboard.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+        isCopied = true;
+        shouldShowLoading = false;
+        setTimeout(() => setIsCopyTriggered(false), 3000);
+      } else {
+        isCopied = false;
+        shouldShowLoading = true;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
vikneshwar: Have moved it to the effects and also updated the hook to manage flags through state

src/ui/components/NetworkMonitor/HeaderGroups.tsx
```diff
@@ -32,7 +26,7 @@ export function HeaderGroups({
               const { key, ...columnProps } = column.getHeaderProps();
               const isLastColumn = index === headerGroup.headers.length - 1;
               return (
-                <div className={classNames("p-1", styles[column.id])} {...columnProps} key={key}>
+                <div key={key} className={classNames("p-1", styles[column.id])} {...columnProps}>
```
vikneshwar, line 29: FIxes `key should be unique` warning showing up in the console. Not sure how having the props in the beginning works though

src/ui/components/NetworkMonitor/RequestRow.tsx
```diff
@@ -56,7 +59,7 @@ export const RequestRow = ({
         tabIndex={0}
         data-testid="NetworkMonitor-RequestTable-RequestRow"
       >
-        <div {...row.getRowProps()}>
+        <div {...rowProps}>
```
vikneshwar, line 62: Fixes warning about passing the key to div

src/ui/components/NetworkMonitor/HeaderGroups.tsx
bvaughn: Doesn't look like this should have changed anything, since `columnProps` (the value being spread) had `key` pulled out of it explicitly:
```js
const { key, ...columnProps } = column.getHeaderProps();
```

src/ui/components/NetworkMonitor/RequestRow.tsx
bvaughn: Doesn't look like this should have changed anything either. Can you explain why you think it fixed a warning?

bvaughn (commented):
Thanks for making the requested changes!

I have a few subjective follow up changed I'd like to make, but I'm going to go ahead and merge this PR and then make those changes myself on top of it 😄

src/ui/components/NetworkMonitor/RequestRow.tsx
vikneshwar: Hi @bvaughn, `getRowProps` return value contains the `key` so spreading it here in div causes a warning

src/ui/components/NetworkMonitor/HeaderGroups.tsx
vikneshwar: Yeah it does have the `key` pulled out. The warning here I saw is about key not being unique(but they are actually unique, I confirmed that), not sure how putting it in the beginning fixes it. I don't have a solid answer here, but I tried to move it to the beginning and it did work.

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -51,6 +53,17 @@ const RequestTable = ({
         className={classNames("flex h-full w-full flex-col overflow-x-auto", styles.request)}
         {...getTableProps()}
       >
+        {isCopied || shouldShowLoading ? (
+          <div className="absolute z-50 grid h-56 grid-cols-1 content-end place-self-center">
+            <div
+              id="showCopied"
+              className={`mb-1.5 flex shrink rounded-lg bg-black p-1.5 text-center text-white opacity-100 shadow-2xl transition-all duration-700 ease-in-out`}
+            >
+              {`${shouldShowLoading ? "Copying" : "Copied"} to clipboard`}
```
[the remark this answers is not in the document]
vikneshwar: 1. This was not due to `shouldShowLoading` flag, but due to `setTimeout` I have used to defer setting the state by `300ms`. I did this so when the tooltip changes from `Copying to clipboard` to `Copied to clipboard the tooltip is not shown for few milliseconds and that gives a clear visual for the user instead of just text changing from `Copying` to `Copied` . Here `Copying` to `Copied` happens only for the case where there is a request body, for all other cases it just displays `Copied` in the first place and yeah that delay with `setTimeout` has caused issue with items where there is no body. 

2. I was not sure how it is generally shown across the application, so I decided to have it centred.

PR #8891, opened 2023-03-07
Adds feature run-time override

ryanjduffy:
Adds support for `features` query string parameter accepting a comma-separated list of devtools features which will be force-enabled for the session (not permanently for the user).


jaril (approved):

bvaughn (approved):
I dunno sure

PR #8895, opened 2023-03-08
New SourceHitCountsCache (FE-1277)

hbenl:


bvaughn (changes requested):
Can you hold off on this PR for now?

It looks like it's going to conflict with #8884 which I've been working on for a few days.

bvaughn:
The "suspense" package has been added now (#8909) and I've migrated a few example caches to it (#8910, #8936, and WIP #8940). Can we add this new cache using it? 😄

jasonLaster:
@hbenl i think it's fine to show the hit counts for the currently loaded window. It's cool that the backend `getSymbols` call could contain all hit counts, but to be consistent with the rest of devtools we really should only show the hit counts for the current window

bvaughn:
> That's the plan. However it seems that the interval cache currently doesn't expose what I need to create hooks for it (i.e. I don't see a way to get cached values and to subscribe to changes) but I need that.

I started adding that last night, FWIW. Let me finish (this morning) first and then pick this back up.

hbenl:
> i think it's fine to show the hit counts for the currently loaded window.

I think there's a misunderstanding: I'm not talking about the loaded range of the recording but about the part of the source visible in the source panel. The Outline panel will show 0 hits for functions that are not currently visible in the source panel.
I'm going to change the UI to show no hitcount badge (instead of a hitcount badge with content "0") for functions that are not visible in the source panel. That was the part that I found confusing.

bvaughn:
>> Can we add this new cache using it?
>
> That's the plan. However it seems that the interval cache currently doesn't expose what I need to create hooks for it (i.e. I don't see a way to get cached values and to subscribe to changes) but I need that.

Released 0.0.26 with new [`useIntervalCacheStatus` hook](https://suspense.vercel.app/useIntervalCacheStatus) which should unblock this. I'll post a PR shortly that upgrades us.

hbenl:
about test failures:
- the replay bot comment has not been updated after the last screenshot test run: the tests actually all passed
- the changed screenshots show one logpoint has moved slightly - I guess that the backend gave me a slightly different `time` value when I recreated some protocol fixtures

src/ui/components/Timeline/PreviewMarkers.tsx
```diff
@@ -27,14 +28,15 @@ function PreviewMarkers() {
...
+      visibleLines?.start.line ?? 0,
+      visibleLines?.end.line ?? 0,
       replayClient,
       focusedSourceId,
-      visibleLines,
-      focusRange
+      focusRange ? toPointRange(focusRange) : null
     );
-    const hitCountsForLine = hitCounts.get(hoveredLineIndex + 1)!;
+    const hitCountsForLine = hitCounts.find(([line]) => line === hoveredLineIndex + 1);
```
bvaughn, line 37: Can we use our binary search util here to make this faster?

packages/replay-next/src/suspense/SourceHitCountsCache.ts
```diff
@@ -0,0 +1,99 @@
...
+    begin,
+    end,
+    client,
+    sourceId,
+    pointRange
+  );
+  if (status === "resolved") {
+    return sourceHitCountsCache.read(begin, end, client, sourceId, pointRange);
+  }
+}
```
bvaughn, line 91: If I'm understanding it correctly– the only reason it exists is to let hit counts be loaded _without_ using Suspense? (else we'd be using `sourceHitCountsCache.read` instead of `sourceHitCountsCache.readAsync`)

I don't think we need it. Should be able to just use `sourceHitCountsCache` directly.

packages/replay-next/components/sources/SourceList.tsx
```diff
@@ -99,13 +100,18 @@ export default function SourceList({
   // but neither should actually _block_ us from showing source text.
   // Fetch those in the background via the caches,
   // and re-render once that data is available.
-  const { value: hitCounts = null } = useImperativeCacheValue(
-    sourceHitCountsCache,
+  const hitCounts = useGetSourceHitCounts(
```
bvaughn, line 103: Why do we need `useGetSourceHitCounts`? I think we can just use `sourceHitCountsCache.read` directly here.

The only place `SourceList` is used is inside of `<Source><Suspense><SourceLoader>`

packages/replay-next/src/suspense/SourceHitCountsCache.ts
hbenl: > If I'm understanding it correctly– the only reason it exists is to let hit counts be loaded without using Suspense? 

Correct.

> I don't think we need it. Should be able to just use sourceHitCountsCache directly.

Yes, that would be possible, but this is a performance optimization we consciously added in #8695.

bvaughn: I see. So the goal here is not so much to bypass Suspense– but to complete the initial render _without suspending_ if data hasn't let been loaded. The `useImperativeCacheValue` hook (in "suspense") does this, although that's not its main purpose– and it doesn't work with interval caches.

Maybe I should just add a `useIntervalCacheValue` hook to the "suspense" package too.

hbenl: > Maybe I should just add a useIntervalCacheValue hook to the "suspense" package too.

Yes, I think that would make sense.

bvaughn: Oh it.

bvaughn: PR #8968 updates `suspense@0.0.28` which adds a new hook, `useImperativeIntervalCacheValues`:
https://suspense.vercel.app/useImperativeIntervalCacheValues

Should be usable here 😄

hbenl: I'll rebase when that PR is fixed and merged.

packages/replay-next/playwright/tests/object-inspector.ts
```diff
@@ -179,22 +179,25 @@ test("should render getters and setters correctly", async ({ page }) => {
   await takeScreenshot(page, arrayRow, "inspect-getter-nested-array");
 });
 
-test("should properly bucket properties for an overflowing array", async ({ page }) => {
+test.only("should properly bucket properties for an overflowing array", async ({ page }) => {
```
markerikson, line 182: :hand: Need to remove this `.only`

hbenl: D'oh!

packages/replay-next/src/suspense/SourceHitCountsCache.ts
bvaughn: Should be ready now!

markerikson (approved):

PR #8897, opened 2023-03-08
use UTCMinutes to convert duration into mm : ss

### LINKED ISSUE #8889: The timeline starts from 30:00 rather than from 00:00
**Description:**
The timeline below should start from 00:00, right? But it starts from 30:00

**Steps to Reproduce:**
Just record any replay and play it.

* Screenshots:
image

* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):

adarsh-sgh:
Fix #8889 

### Replay [After Fix](https://app.replay.io/recording/clock-within-a-clock-within-a-clock--ca4937d8-22da-4db0-a202-0c4a2bc2a33f)
image
(timestamp starts from 0:00)

 ### Unit tests after fix
image

### Replay [Before Fix ](https://app.replay.io/recording/clockreplay--1a13a10a-356c-45ce-aad3-f39e42d58f99) (taken from the issue)

image

### Unit tests before fix
image

**Note: you can set your timezone to IST to get these tests failing before the fix**

run `TZ=Asia/Calcutta yarn test`  to run tests with IST timezone

### Tests 
Although unit test for the function is already present it would be better if we can run tests in (two ?) different timezones ? 🤔 
TZ environment variable can be modified to change timezone but not sure if it's worth it.

### TODO
there are one or two other places with same issue
e.g when we hover over the timeline the tooltip still shows different minutes elapsed a/c to locale, will be fixing it soon.

### Alternate Solution
credits : copilot
```js
 let minutes = Math.floor(time / 1000 / 60);
 let seconds = Math.floor(time / 1000) % 60;
 const milliseconds = time % 1000;
```

also our current implementation works upto 1 hour only, that is a safe upper limit for replays ?
`3600000` milliseconds convert into `0:00`

### DIFF
--- a/src/ui/utils/timeline.ts
+++ b/src/ui/utils/timeline.ts
@@ -123,9 +123,9 @@ export function getNewZoomRegion({
 // Format a time value to mm:ss
 export function getFormattedTime(time: number, showMilliseconds: boolean = false) {
   const date = new Date(time);
-  let minutes = date.getMinutes();
-  let seconds = date.getSeconds();
-  const milliseconds = date.getMilliseconds();
+  let minutes = date.getUTCMinutes();
+  let seconds = date.getUTCSeconds();
+  const milliseconds = date.getUTCMilliseconds();
 
   if (!showMilliseconds) {
     if (milliseconds >= 500) {


adarsh-sgh:
https://github.com/replayio/devtools/blob/03d4de6da157a5422388e62dc17bedc619a00e15/src/ui/components/Library/Team/View/TestRuns/utils.tsx#L8-L14

https://github.com/replayio/devtools/blob/2de522bb51101c4e836a261052f69c54ade5edfd/src/ui/components/Timeline/Tooltip.tsx#L12-L19

these two are very similar to the function below, only diff i can see is minutes are padded in first.
should we try to use the below function for all these cases ?
https://github.com/replayio/devtools/blob/dc44c076ed40440054056b98b1cf748520741109/src/ui/utils/timeline.ts#L124

adarsh-sgh:
@ryanjduffy can you please review this.

bvaughn (approved):
> should we try to use the below function for all these cases ?

That sounds like a good idea.

Maybe you could add this format function to: `packages/shared/utils/time.ts` ?

PR #8901, opened 2023-03-09
Make the loading screen less flickery

jonbell-lot23:
When we go to the library, the loading screen is called several times, and oftentimes each load is only for a split second. This leads to a very flickery and heavy experience.

Old on the left, new on the right:
Better library load-v1

Here are the four small changes to loading messaging to get this effect:

* We used to say "Reloading team details" but it's not necessary.
* Same with "Loading account details"
* We changed the default from "Starting your session..." to the more user-friendly + standard "Loading..."
* We changed "Loading timeline..." to "Loading..." because that distinction isn't important enough for a flicker

* And: a small change to make the tips text a bit smaller.

--

Of note, we still have our standard messaging on devtools. That's where different states (and stalls) are much more relevant to show the user, and aren't as flickery.

### DIFF
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -223,7 +223,7 @@ function _DevTools({
   }, [recording]);
 
   if (!loadingFinished) {
-    return <LoadingScreen fallbackMessage="Starting your session..." />;
+    return <LoadingScreen fallbackMessage="Loading replay..." />;
   }
 
   if (loadedRegions === null) {
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -223,11 +223,11 @@ function _DevTools({
   }, [recording]);
 
   if (!loadingFinished) {
-    return <LoadingScreen fallbackMessage="Loading replay..." />;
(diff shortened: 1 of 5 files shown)


jasonLaster (commented):
Love the spirit of the change

src/ui/components/DevTools.tsx
```diff
@@ -223,11 +223,11 @@ function _DevTools({
   }, [recording]);
 
   if (!loadingFinished) {
-    return <LoadingScreen fallbackMessage="Starting your session..." />;
+    return <LoadingScreen fallbackMessage="Loading..." />;
   }
 
   if (loadedRegions === null) {
```
jasonLaster, line 229: Perhaps merge these into one IF ` if (!loadingFinished || loadedRegions === null) {` Or better yet, can we move the  `loadedRegions === null` check into the `createSocket` call

jonbell-lot23:
Status update: still have lots of failures, but here's the general idea! Stay tuned

jonbell-lot23:
This was a bit stale so I opened another PR:
https://github.com/replayio/devtools/pull/8971

PR #8910, opened 2023-03-09
Bvaughn/suspense migrate object preview caches

bvaughn:
Second piece of PR https://github.com/replayio/devtools/pull/8884. This piece migrates an initial cache (for reference).

- [x] Migrate object preview cache
- [x] Update fixture data
- [x] Ensure unit tests pass
- [x] Make sure e2e and screenshot tests pass


bvaughn:
After rebasing on the now-passing main, there are two e2e test failures:
* tests/object_preview-01.test
* tests/react_devtools.test

The first one, at least, looks likely to be related to this change. I'll dig in after lunch.

packages/replay-next/components/console/ErrorStackRenderer.tsx
```diff
@@ -37,7 +37,7 @@ function ErrorStackRendererSuspends({
   pauseId: PauseId;
 }) {
   const client = useContext(ReplayClientContext);
-  const stack = getObjectPropertySuspense(client, pauseId, errorObjectId, "stack").value;
+  const stack = objectPropertyCache.read(client, pauseId, errorObjectId, "stack")?.value;
```
markerikson, line 40: :speech_balloon: First time seeing usage of the `suspense` package. Part of me wonders if `.read()` should actually be named `.readSuspense()` instead for clarity.  (I suppose the counter-argument is that "using this _with_ actual `<Suspense>` is the default intended use case" :)  )

Interestingly, this now actually _looks_ more like what I remember seeing in the early Suspense demos, syntax-wise.

packages/replay-next/components/inspector/protocolValueToText.ts
```diff
@@ -59,7 +58,7 @@ async function protocolValueToTextHelper(
     } else {
       objectIdSet.add(objectId);
 
-      const object = await getObjectWithPreviewHelper(client, pauseId, objectId, true);
+      const object = await objectCache.readAsync(client, pauseId, objectId, "full");
```
markerikson, line 61: :speech_balloon: Thanks for turning this into an enum! It's now much more clear what the point of that arg is :)

packages/replay-next/src/suspense/ObjectPreviews.ts
```diff
@@ -1,288 +1,87 @@
...
+    objectId: ObjectId,
+    previewLevel: ObjectPreviewLevel
+  ],
+  Object
+> = createCache({
+  // The protocol only sends objects once per session;
+  // disable weak ref caching to ensure we don't lose data
+  config: { useWeakRef: false },
+  debugLabel: "objectCache",
+  getKey: ([client, pauseId, objectId, previewLevel]) => `${pauseId}:${objectId}:${previewLevel}`,
```
markerikson, line 26: :speech_balloon: Looking at https://suspense.vercel.app/createCache , I don't see any actual _reference_ docs or TS types describing the possible args to `createCache`, or what the resulting `Cache` object contains. It would be helpful to add those when you have time.

(related: I'm confused why `getKey()`'s first arg is an array of the arguments, instead of the separate args themselves directly.)

packages/replay-next/src/suspense/ObjectPreviews.ts
```diff
@@ -1,288 +1,87 @@
...
+import { Cache, createCache } from "suspense";
 
 import { ReplayClientInterface } from "../../../shared/client/types";
-import { createFetchAsyncFromFetchSuspense } from "../utils/suspense";
 import { cachePauseData } from "./PauseCache";
 
-type ObjectMap = Map<ObjectId, Object>;
-type RecordMap = Map<ObjectId, Record<Object>>;
-type PropertyRecordMap = Map<string, Record<ProtocolValue>>;
+export const objectCache: Cache<
```
markerikson, line 13: :speech_balloon: Had to go off and look at the final version of this file separate from the diff, but yeah, that _does_ look nice and compact!

markerikson (approved):

packages/replay-next/components/console/ErrorStackRenderer.tsx
bvaughn: Yeah, I get where you're coming from with this observation. I waffled on the naming a bit, but settled on `read` as a suggestion from someone in the community (initial name was `fetchSuspense`).

I think it's subjective enough that I'll leave it as is for now.

packages/replay-next/components/inspector/protocolValueToText.ts
bvaughn: Agreed! 😁

packages/replay-next/src/suspense/ObjectPreviews.ts
bvaughn: Ah, it's not an "args array" - it's using named parameters. So `getKey` comes earlier because I alpha-sort keys by habit. 😄 

Good point on explicitly calling out the generic types and how they relate to the `getKey` and `load` params though.

packages/replay-next/src/suspense/ObjectPreviews.ts
bvaughn: Thanks! 🙇🏼 

I'm pretty happy with how this specific cache turned out. Given that it was the most complex of the "simple" caches, I think it's a good signal for replacing the other `createGenericCache` instances.

Will be a bit more work to replace the interval and streaming cache types though.

bvaughn:
Okay. I think all tests should pass now!

PR #8912, opened 2023-03-09
Do not block loading on endpoint

jcmorrow:
In this Replay we wait for the endpoint for four minutes:

https://app.replay.io/recording/super-duper-never-loads-wtf--bcbb8ecc-36d6-4199-a43e-cade7a4433a1?point=157391498534412048370684558846197892&time=54376.98606502986&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6IjUwNDYyNjM1MDk2ODYxMzE2NjYyOTcxNzEzMDg5NjAxNzY0NSIsInRpbWUiOjE3OTE4OX0sImVuZCI6eyJwb2ludCI6IjY4NDQwOTYyOTcwNTYzOTU0MTMyMjEwNzg4Mzc2NzIwMTc5MiIsInRpbWUiOjIzOTIwM319

We should make that faster. But also, in the meantime, let's just not await it.

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -332,8 +332,8 @@ export function createSocket(
       ThreadFront.on("paused", ({ point }) => dispatch(setCurrentPoint(point)));
 
       await ThreadFront.waitForSession();
-      await dispatch(jumpToInitialPausePoint());
 
+      dispatch(jumpToInitialPausePoint());
       dispatch(actions.setLoadingFinished(true));
 
       if (!focusRegion) {


jasonLaster (approved):
Are we still waiting in DevTools for regions?

jcmorrow:
> Are we still waiting in DevTools for regions?

I assumed that was removed in the last pass after I pointed it out, but let me double check.

jcmorrow:
Confirmed. That was already removed here: https://github.com/replayio/devtools/pull/8904/files#diff-10caa866926e2aa822cb517138eea17237a290d7279302e01c8ff3deaa60876eL229-L232

jcmorrow:
Not sure what to do about these test failures.

hbenl:
I wanted to avoid opening devtools paused at point 0 and then jump to a different point later. In particular, it would be bad UX if the user had already selected a point in the recording and then we take him to a different point. That can be avoided however with a little change that I've added to this PR.

PR #8917, opened 2023-03-10
Custom Playwright formatter for local test runs

bvaughn:
The way logs were interleaved previous was annoying and difficult to read. While things were kind of down this afternoon I decided to write a formatter.

Here's an example run of a few random tests (with some expected failures):

https://user-images.githubusercontent.com/29597/224442133-48f3567d-fbab-4894-95c0-c653de1daf49.mp4

And here are some screenshots of the runner in progress:

Screen Shot 2023-03-10 at 6 02 38 PM

Screen Shot 2023-03-10 at 6 02 35 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,6 +155,7 @@
     "@types/react-window": "^1.8.5",
     "@types/styled-components": "^5.1.25",
     "@types/uuid": "^7.0.3",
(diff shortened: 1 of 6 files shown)


bvaughn:
This is probably not worth merging.

Issue #8905, opened 2023-03-09
 Scrollbar Size is too wide inside Replay Browser and Firefox in windows


anikdhabal:
**Description:**
 scrollbar issue on High DPI screen. Scrollbar of Replay is too wide inside Replay Browser and Firefox in window. But it's fine in  another browser.

**Steps to Reproduce:**
1.Open Replay in Replay browser or Firefox in Windows.
2.See scrollbar

**Expected Result:**
Its size should be perfect.

**Additional Context:**

* Screenshots:
<img width="716" alt="221770202-7dfed3ec-5492-4b39-9ab7-5067a2a893e9" src="https://user-images.githubusercontent.com/81948346/224034182-89d8ec1d-c70e-4cff-8b11-3dd90898c582.png">

* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):
https://app.replay.io/recording/scrollbar-dpi-issue--d446af32-6cea-4d45-9ae5-f40d3439aefe

swapn652:
I think this issue exists only in replay browser.

### Here's a screenshot of Replay in Microsoft Edge
![image](https://user-images.githubusercontent.com/70851202/224051957-5d70a8ff-fc4a-44c4-b5c3-739cdd8edd36.png)

### Local Host
![image](https://user-images.githubusercontent.com/70851202/224052673-b8abbf68-6398-4c1d-9623-20c27135ab44.png)

### Replay Browser
![image](https://user-images.githubusercontent.com/70851202/224052761-76c55ea9-ce01-48cb-a4ef-8eaa71b414d8.png)

swapn652:
I wonder why that happens. The code responsible for scroll bar is in **src/base.css**. Apparently the code isn't working for replay browser.

Domiii:
The current "Replay Browser" is based on Firefox. So maybe its a Firefox thing.

anikdhabal:
> The current "Replay Browser" is based on Firefox. So maybe its a Firefox thing.

Yes, currently **::-webkit-scrollbar** is not supported by firefox. Currently, styling scrollbars for Firefox is available with **scrollbar-width** and **scrollbar-color** properties.

PR #8920, opened 2023-03-11
Scrollbar size is Fixed inside Replay browser and Firefox in windows

anikdhabal:
closes: #8905 
currently ::-webkit-scrollbar is not supported by firefox. Currently, styling scrollbars for Firefox is available with scrollbar-width and scrollbar-color properties.
Replay:-
https://app.replay.io/recording/scrollbar-size-is-fixed--cd001945-9e6d-47b8-87d9-8819df900c96

### DIFF
--- a/src/base.css
+++ b/src/base.css
@@ -12,6 +12,8 @@ body {
 
 * {
   box-sizing: border-box;
+  scrollbar-width: thin;
+  scrollbar-color: transparent;
 }
 
 .showRedactions[data-private="true"] {


jasonLaster:
Thanks @anikdhabal this looks like a promising solution. Could you share before/after screenshots as well?

NOTE: I don't think that this will fix this issue, but seems related
https://github.com/replayio/gecko-dev/issues/746

anikdhabal:
Before:-
Screenshot 2023-03-12 223454

After:-
Screenshot 2023-03-12 223416
@jasonLaster

anikdhabal:
> Thanks @anikdhabal this looks like a promising solution. Could you share before/after screenshots as well?
> 
> NOTE: I don't think that this will fix this issue, but seems related [replayio/gecko-dev#746](https://github.com/replayio/gecko-dev/issues/746)

Yes It will not fix this issue, but I've checked that this styling property can fixed it.
Screenshot 2023-03-12 225721

jasonLaster:
@anikdhabal could you provide another before/after that more similar? For instance same window, viewport, theme...

anikdhabal:
Before:-


> @anikdhabal could you provide another before/after that more similar? For instance same window, viewport, theme...

Before:-
Screenshot 2023-03-13 001157

After:-
Screenshot 2023-03-13 001040

src/base.css
```diff
@@ -12,6 +12,8 @@ body {
 
 * {
   box-sizing: border-box;
+  scrollbar-width: thin;
+  scrollbar-color: transparent;
```
bvaughn, line 16: I wish there was a better way to scope this than `*`* but I don't suppose there is.

bvaughn (approved):

PR #8922, opened 2023-03-11
scrollbar not showing in inspector-sidebar-container fixed

Aimajoke:
Description:
Vertical scrollbar is  not appearing in inspector-sidebar-container, as mentioned in [#8820](https://github.com/replayio/devtools/issues/8820) .
Steps to Reproduce:
1. Open Replay browser
2. Change from viewer to Devtools
3. Open element on Devtools
4.Try to scroll Rules, Layout or computed list.

Expected Result:
The scrollbar should appear and work to see the complete list.

Screenshots:

Before (on Replay, although not showing in any browser):
replay_before

After:
Replay | Chrome
replay_after

Firefox | Edge

firefox_after

It was just a oneline change of a class to adjust the height and avoid overflowing.

### DIFF
--- a/src/devtools/client/shared/components/Tabs.css
+++ b/src/devtools/client/shared/components/Tabs.css
@@ -5,6 +5,7 @@
 /* Tabs General Styles */
 
 .tabs {
+  height: 100%;
   background: var(--theme-sidebar-background);
   display: flex;
   flex-direction: column;


Domiii:
FYI: Here [this PR](https://github.com/replayio/devtools/pull/8881) is an alternative solution.

@Aimajoke, in that other PR, [Roaster05](https://github.com/Roaster05) also points out that your change is to a general component, and might affect other places in the code in unintendend ways. Can you double check all places where `tabs` are used and make sure that they are not negatively affected?

Roaster05:
> 

@Domiii actually the number of lines changed is more just because of running the prettier tests and not because I have modified the code too much i have just mentioned the classes `panels` and `tabs` in their parent div elements

Aimajoke:
> @Aimajoke, in that other PR, [Roaster05](https://github.com/Roaster05) also points out that your change is to a general component, and might affect other places in the code in unintendend ways. Can you double check all places where `tabs` are used and make sure that they are not negatively affected?

Yup, after further inspection (for God's sake everything is called tabs here), the **class** tab is only used one more time in the project on the component PanelTabs that is rendered on NetworkMonitor, RequestDetails (The panel that shows details of a request on the Network tab) and my changes had no effect on it. 
In theory my changes should had also fix a scrollbar issue here but the way RequestDetails renders the panel is different from how is rendered in the main devtools panel so that issue does not occur here, but either way the changes made have no negative side effects. 

The other component  where the class tabs is used (Screenshots with and without the change):

withoutchanges

tabsclass_usage

Maybe related issue found here, because scrollbar is not showing either in this Network subpanel although it is working fine, It would be better if the scrollbar is visible. I think is a different issue because the component that renders this subpanel is a different one and works different, probably would not be a oneline fix as done on this one.

bvaughn (approved):
This seems like a reasonable solution. Thank you for the before and after screenshots. Those are helpful.

Issue #8919, opened 2023-03-11
Redirecting to the site in dashboard


Roaster05:
**Description:**

**Functionality Improvement** 
To provide a URL link for the site whose replay has been recorded in the dashboard section so we can have a look once again to view the current state 

**Steps to Reproduce:**
Just visit the page `/team/me/recordings` that will get you to the dashboard and then you can view the link mentioned below each recording title

**Expected Result:**
while clicking the link we are navigated to the recording itself but we might route to the website to view the current state of the website or if recording a new replay
**Additional Context:**

* Screenshots:
 
![Screenshot (1) issue](https://user-images.githubusercontent.com/100672872/224490413-a81e8a7a-4a7b-4f47-a22c-3f83832ddbef.png)

* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):
 https://app.replay.io/recording/overboard--90cfddf4-ced2-4bca-b67d-112e60132c52

bvaughn:
cc @jonbell-lot23 for triage/input

On the one hand, we _do_ show the "main" URL in a Replay's info panel. On the other hand– there's no reason why a Replay would have only one URL– or in the case of Node recordings, it might not have a URL at all.

So I'm leaning toward thinking we should not do this.

bvaughn:
Going to close this for now. If Jon disagrees, we can reopen.

jonbell-lot23:
Good conversation, thanks for the exploration and tagging me.

Yup, I agree with closing it for the reasons mentioned:

1. There can be multiple urls
2. This is more appropriate to put in the replay info panel itself
3. It's pretty common for people to keep hitting the same url (if I work at contoso, most of my recordings will be of contoso.com)
4. It's not super common to have multiple urls but it happens. To do a feature like this fully, you'd imagine showing each of the URLs or showing there we multiple

PR #8924, opened 2023-03-12
added href link for recording URL in Library

Roaster05:
A draft Pull request for the issue #8919 and added a href link so that the url can be visited to see the current state or to record a new replay for comparison
the replay for  the new changes is:
in the replay it got redirected in a new tab but was not shown in replay 
https://app.replay.io/recording/overboard--3a2065a9-46f5-48e2-8910-ad7a0c739f1d


bvaughn:
Thanks for the suggestion! I left some comments on #8919 though about why I don't think we should make this change. Essentially:

* Replays may have no URL at all (ones recorded with Node).
* Replays may have multiple URLs.
* I personally/subjectively think it's confusing to show this in the list view (and this would probably need to be a decision made by our Design team).

PR #8925, opened 2023-03-13
Add incremental parsing (FE-853)

hbenl:
I've reimplemented the `SyntaxParsingCache` to support incremental parsing and fixed `SourceListRow`, which rerendered every time a source content chunk was processed (even if that chunk didn't contain the line rendered by `SourceListRow`).
Opening a huge source will now parse the entire source again (in chunks) and this results in a noticeable lag when using the UI while the source is being parsed. This is exacerbated by the fact that multiple chunks received from the backend may be grouped into one chunk to be parsed, which may then freeze the UI for a second or more.
The next step is to move the parsing to a worker.


bvaughn:
I've been planning to migrate this cache to be built on top of the streaming cache API: https://suspense.vercel.app/createStreamingCache

If we're going to update it, can we make that change part of it too? (Not trying to make more work for you 😄  but I am hoping to avoid us doing the same amount of work twice)

hbenl:
I don't think `createStreamingCache` would be a good fit here: if I understand it correctly, the `useStreamingValue` hook would update every time a source chunk is received and parsed but in `SourceListRow` I create hooks that only update when that particular line is received and parsed. Would that be possible with `createStreamingCache`?

bvaughn:
I wrote `createStreamingCache` for use cases like the ones in Replay, so if it's _not_ a good fit– I'd like to know how it needs to be changed to be a good fit. It's meant to be used along with the [`useStreamingValue` hook](https://suspense.vercel.app/guide/writing-a-streaming-cache) to render as values stream in.

I'm not sure I understand what you're saying– but I think you're raising a concern about over render lines in the source editor as new values stream in? I don't think that would be very expensive since we're limiting the rows we render, and `useStreamingValue` debounces updates.

But if we did see over-rendering, we _could_ add our own `useSyncExternalStore` hook that works exactly like the one yo have 😄

hbenl:
> I think you're raising a concern about over render lines in the source editor as new values stream in?

Exactly.

> The actual tokens for a given line won't change (new lines are just added)

Yes, but the hook doesn't know that we're only interested in one particular line. The streaming value (containing the tokens for *all* parsed lines) will update every time new lines are added.

> It shoulnd't be very expensive to re-render even if we do because we're using windowing to limit the rows we render

When you open a source that is some megabytes large, you receive hundreds of chunks. Multiply that by dozens of lines that are visible and you get thousands of unnecessary rerenders. This led to the UI being almost completely unresponsive for several seconds (until the source is completely parsed) until I made the change to `SourceListRow`.

> If we did see over-rendering, we could add our own useSyncExternalStore hook that works exactly like the one yo have

That would be favorite solution. I just saw that `StreamingValue` exposes everything I need for that, so I'll try that.

hbenl:
I just noticed another problem: we expose both raw and parsed lines, each with their own progress value (raw lines usually finish a lot earlier than parsed lines if the chunks are received faster than they can be parsed, which is the case for me). But `createStreamingCache` only supports one progress value. I could try to create two caches in one go (one for raw and one for parsed lines) but I think the code for that could become a bit cumbersome, possibly negating the benefits of using `createStreamingCache` here.

bvaughn:
> we expose both raw and parsed lines, each with their own progress value (raw lines usually finish a lot earlier than parsed lines if the chunks are received faster than they can be parsed, which is the case for me). But createStreamingCache only supports one progress value.

It was designed with this in mind also. The interface is:
```ts
createStreamingCache<
  Params extends Array<any>,
  Value,
  AdditionalData = undefined
>
```

Additional data can be anything– in our case, raw lines could go there.

packages/replay-next/components/sources/utils/comments.ts
```diff
@@ -89,26 +89,26 @@ export async function createTypeDataForSourceCodeComment(
...
       }
 
-      rawText = parsedSource.rawTextByLine[lineNumber - 1];
+      rawText = parsedSource.data!.text[lineNumber - 1];
       if (rawText.length <= maxTextLength) {
         // If the raw text is longer than the max length, we can't safely use the parsed tokens.
-        if (parsedSource.parsedTokensByLine.length >= lineNumber) {
-          parsedTokens = parsedSource.parsedTokensByLine[lineNumber - 1] ?? null;
+        if (parsedSource.value?.length ?? 0 >= lineNumber) {
+          parsedTokens = parsedSource.value?.[lineNumber - 1] ?? null;
```
bvaughn, line 111: I think we don't need to still be using `!` and `?` at this point. We should be able to add a TypeScript assertion after  line 105.

packages/replay-next/src/suspense/SyntaxParsingCache.ts
```diff
@@ -1,407 +1,157 @@
...
+  },
+});
 
-  return {
-    isComplete: () => complete,
-    parsedTokensByLine: parsedTokens,
-    parseChunk,
-  };
+export function parseStreaming(source: StreamingSourceContents, fileName: string | null) {
+  return streamingParserCache.stream(source, fileName);
```
bvaughn, line 111: I think we should start exporting these caches _directly_ rather than wrapper methods. The caches usually have multiple methods that are potentially useful. The "suspense" package also exports utils/hooks for working with the caches that can't be used if the actual caches can't be referenced.

packages/replay-next/src/suspense/SyntaxParsingCache.ts
```diff
@@ -1,407 +1,157 @@
...
-      parsedTokens.push(currentLineState.parsedTokens);
-
-      currentLineState.parsedTokens = [];
-      currentLineState.rawString = "";
-    }
-
-    index = nextIndex + 1;
-    nextIndex = section.indexOf("\n", index);
+function handleParseResponse({ data }: MessageEvent<ParseResponse>) {
+  const state = internalParserStates.get(data.sourceId);
```
bvaughn, line 145: So this `data` message is _guaranteed_ to always break evenly at the line level? Even for sources with _really long lines_?

hbenl: What do you mean by "break evenly at the line level"?

bvaughn: I mean that the parsing will have to always finish at the line boundary

hbenl: OK, yeah, `data` will only contain completed lines: the worker removes the last line of a chunk if it isn't the last chunk, because that line will usually be incomplete. When the next chunk comes in, it doesn't start highlighting at the beginning of that chunk but at the beginning of the first line that hasn't been highlighted yet.

bvaughn: Great! Thanks. Just wanted to confirm that behavior was still in place.

bvaughn:
Looks like this broke some of our screenshot tests? on `main` after pulling I see

e.g.
image

is now showing up as 
image

PR #8928, opened 2023-03-13
Delete failing Cypress e2e test

bvaughn:
This has been causing CI to fail for a while now, even on empty commits.

Let's re-add it only once it's _not_ running in CI and causing failures.


jasonLaster:
I think `manifest.js` has a skip flag

PR #8930, opened 2023-03-13
Focus source viewer on source switch so that CMD+F works okay

markerikson:
This PR:

- Adds a `useEffect` to the main app's `<NewSourceAdapter>` so that we auto-focus the current source viewer's container element whenever the selected source changes.  This enables `CMD+F` to work immediately

(Hopefully this is the right spot to put that check - please let me know if there's a better place!)

### DIFF
--- a/packages/replay-next/components/sources/Source.tsx
+++ b/packages/replay-next/components/sources/Source.tsx
@@ -29,6 +29,9 @@ export type HoveredState = {
   target: HTMLElement;
 };
 
+/**
+ * Rendered in the main app by `<NewSourceAdapter>`
+ */
 export default function Source({
   source,
   showColumnBreakpoints,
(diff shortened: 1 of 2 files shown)


packages/replay-next/components/sources/Source.tsx
```diff
@@ -1,6 +1,14 @@
 import { newSource as ProtocolSource } from "@replayio/protocol";
 import debounce from "lodash/debounce";
-import { MouseEvent, Suspense, useContext, useLayoutEffect, useRef, useState } from "react";
+import {
+  MouseEvent,
+  Suspense,
+  useContext,
```
bvaughn, line 6: Looks like an unused `useEffect` import got added here?

bvaughn (approved):

packages/replay-next/components/sources/Source.tsx
markerikson: Whoops, yeah, I had tried that file first and then changed. Will revert.

PR #8931, opened 2023-03-13
Fix event filtering and remove remaining uses of string refs

markerikson:
This PR:

- Fixes a bug in the filtering of sidebar events that was apparently due to how we were comparing execution points
  - Replaces the unmemoized `getFlatEvents` selector with a memoized `getFilteredEventsForFocusRegion` selector
  - Uses our actual execution point comparison utils for the filtering
  - Converts `<Events>` from using `connect` to using hooks
- Removes the remaining uses of string refs because we recently upgraded React and it's now spitting out "string refs are deprecated" warnings in dev
  - Migrates `<ResultList>` to TS
  - Refactors the list scrolling logic in `<QuickOpenModal>` and `<ResultList>` to push that all down into `<ResultList>`, and only use ref objects instead of string refs
  - Replaces the string ref usage in `<Tabs>`


markerikson:
The "fuzzy file search" test failure probably indicates I really did actually break something.  Will investigate shortly.

src/devtools/client/debugger/src/components/Editor/Tabs.tsx
```diff
@@ -35,12 +35,13 @@ class Tabs extends PureComponent<PropsFromRedux> {
   _draggedSource: SourceDetails | { url: null; id: null } | null = null;
   _draggedSourceIndex: number | null = null;
 
+  sourceTabs = React.createRef<HTMLDivElement>();
```
bvaughn, line 38: Subjective naming nit (please feel free to ignore) – I prefer to prepend refs with "ref" (e.g. `sourceTabRef`) so it's clear at a glance that they are a wrapper and not the thing itself.

src/devtools/client/debugger/src/components/shared/ResultList.tsx
```diff
@@ -6,24 +6,51 @@ import classnames from "classnames";
...
+  role?: string;
+  items: SearchResultWithHighlighting[];
+  expanded: boolean;
+  selected: number;
+  size?: string;
+  selectItem: (e: any, item: SearchResultWithHighlighting) => void;
+}
+
+export default class ResultList extends Component<ResultListProps> {
+  listItemRefs: Record<number, HTMLElement> = {};
```
bvaughn, line 24: Subjective nit: This is kind of confusingly named because it's not a map of _refs_ but a map of actual HTML elements.

I guess those _are_ "refs" but I think of a "ref" as a thing with a `.current` property.

src/ui/reducers/app.ts
```diff
@@ -21,12 +22,10 @@ import {
 } from "ui/state/app";
 import { getNonLoadingRegionTimeRanges } from "ui/utils/app";
 import { getSystemColorSchemePreference } from "ui/utils/environment";
-import { compareBigInt } from "ui/utils/helpers";
 import { prefs } from "ui/utils/prefs";
 import { isPointInRegions, isTimeInRegions, overlap } from "ui/utils/timeline";
 
 export const initialAppState: AppState = {
-  // analysisPoints: {},
```
bvaughn, line 29: Delete?

src/ui/reducers/app.ts
```diff
@@ -278,27 +277,38 @@ const NO_EVENTS: MouseEvent[] = [];
...
+      return isExecutionPointsWithinRange(e.point, focusRegion.begin.point, focusRegion.end.point);
+    });
+    return filteredEvents;
+  }
+);
 
-  // Only show the events in the current focused region
-  return focusRegion
-    ? filteredEvents.filter(
-        e => e.point >= focusRegion.begin.point && e.point < focusRegion.end.point
```
bvaughn, line 298: Ooof, so easy to do this.

src/ui/reducers/app.ts
```diff
@@ -278,27 +277,38 @@ const NO_EVENTS: MouseEvent[] = [];
...
+export const getFilteredEventsForFocusRegion = createSelector(
+  getFocusRegion,
+  getSortedEventsForDisplay,
+  (focusRegion, sortedEvents) => {
+    if (!focusRegion) {
+      return sortedEvents;
+    }
+
+    const filteredEvents = sortedEvents.filter(e => {
+      return isExecutionPointsWithinRange(e.point, focusRegion.begin.point, focusRegion.end.point);
```
bvaughn, line 306: 👍🏼  Much nicer

bvaughn (approved):
Once you get the failing test figured out, this looks okay otherwise!

src/devtools/client/debugger/src/components/Editor/Tabs.tsx
markerikson: Yeah, agreed. I think I was keeping to the existing naming because it was already in `this.refs`.

src/ui/reducers/app.ts
markerikson: That _is_ a removal :) it was already commented out when I was looking at this file, so I just killed it.

bvaughn: lol it's been a long day

markerikson:
Hmm. The test failure is:

```
  1) tests/source-and-console.ts:534:1 › should render search results properly for lines with multiple breakable column positions 

    Error: expect(received).toBeFocused()
    Call log:
      - expect.toBeFocused with timeout 5000ms
      - waiting for selector "[data-test-id="SourcesRoot"]"
      -   selector resolved to <div tabindex="0" data-test-id="SourcesRoot" class="S…>…</div>
      -   unexpected value "false"
```

Which feels like it could be related to the change I made in #8930 for source focusing, except that PR had a green checkmark :(

PR #8932, opened 2023-03-13
[React] polish panel

jasonLaster:
Safety: ReactPanel is behind a feature flag. The toolbartab change is minimal

Changes
- Updates React icon
- Adds  react/redux logos
- adds extra guard around analysis
- removes unused code

Screenshot 2023-03-13 at 3 13 17 PM

### DIFF
--- a/src/ui/components/ReactPanel.tsx
+++ b/src/ui/components/ReactPanel.tsx
@@ -3,6 +3,7 @@ import classnames from "classnames";
 import { ReactNode, useContext, useEffect, useState } from "react";
 
 import { selectFrame as selectFrameAction } from "devtools/client/debugger/src/actions/pause/selectFrame";
+import AccessibleImage from "devtools/client/debugger/src/components/shared/AccessibleImage";
 import {
   PauseFrame,
   getExecutionPoint,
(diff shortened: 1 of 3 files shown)


src/ui/components/ReactPanel.tsx
```diff
@@ -383,118 +390,42 @@ function ReactQueuedRenderListItem({
   );
 }
 
-type CommittedRenderProps = CommonProps & {
-  commitPoint: TimeStampedPoint;
-};
-
-function ReactCommittedRenderListItem({
```
markerikson, line 390: :speech_balloon: FWIW I still _really_ think we ought to be showing commits somehow :)

src/ui/components/ReactPanel.tsx
```diff
@@ -383,118 +390,42 @@ function ReactQueuedRenderListItem({
...
   const { rangeForDisplay: focusRange } = useContext(FocusContext);
   const [renderDetails, setRenderDetails] = useState<RenderAnalysisResults | null>(null);
 
-  const handleClick = async () => {
-    console.log("Focus range: ", focusRange);
-    const renderAnalysisResults = (await dispatch(findQueuedRendersForRange(focusRange!))) ?? null;
-    setRenderDetails(renderAnalysisResults);
-  };
+  useEffect(() => {
+    setRenderDetails(null);
```
markerikson, line 401: :hand: one issue here is that this ends up re-fetching the values if you click to another tab and back, because `<Offscreen>` re-runs effects.

These really _ought_ to be cached somehow.  We can skip it for this PR.  A stopgap follow-up would be to use our existing cache and just do it for the focus range, without worrying about solving the over-fetching problem.

src/ui/components/Toolbar.tsx
```diff
@@ -46,11 +46,18 @@ function CypressIcon() {
   );
 }
 
-function ReactIcon() {
+function ReactIcon({ active = false }) {
```
markerikson, line 49: :question: Doesn't look like `active` is being used?

markerikson (approved):
Seems generally okay, albeit with my caveat that I'd still really like to show commits in the list :)  Double-check the `active` prop for the icon?

src/ui/components/ReactPanel.tsx
jasonLaster: Agreed - would just like a different UI treatment for it. This is where having a figma brainstorming session helps

src/ui/components/ReactPanel.tsx
jasonLaster: yep - any caching will do

PR #8938, opened 2023-03-14
Add mode to test run view

ryanjduffy:
Adds the test run mode (when set) to the list item and summary

image


ryanjduffy:
@jonbell-lot23 - FYI on using a couple material icons here as placeholders.

image
image
image

jaril (approved):

PR #8943, opened 2023-03-15
Added text to emoji feature like discord

Roaster05:
Created an Discord like Emoji feature which gives the user a feature to add Emoji by using the specification of the emoji also the feature is implemented with the help of an external package `node-emoji` and provides this feature for all the existing emojis
The replay for its working is :
https://app.replay.io/recording/overboard--d604f66e-6796-4189-80ff-9b2f70159c63


bvaughn:
Thank you for the suggestion/idea 🙇🏼  but I don't think we should proceed with this for now. It seems like a pretty aggressive change and (so far as I know) it's not one that's been run by our Design team.

Issue #8945, opened 2023-03-15
Input disabled message in console is formatted strangely


ryanjduffy:
**Description:**
When the console is narrow and the user is paused outside the focus window, the input disabled message is formatted strangely

**Steps to Reproduce:**
https://app.replay.io/recording/input-disabled-message-formatting-issue--0ff423cc-6eef-41bc-9d99-5af7d884e266

**Expected Result:**

**Additional Context:**

* Screenshots:
* Video (Loom):
* Replay ([replay.io/record-bugs](https://replay.io/record-bugs) for instructions):

MohitGupta14:
Hi, I found a similar  Issue. At the bottom right corner, PurchaseForm.tsx looks weird when we scroll. 
I want to know that should  I create this as a new issue and make a PR. I'm new to open-Source, and need some guidance.
 
https://user-images.githubusercontent.com/81846020/225627578-f5c83f20-400f-48a0-9f6d-46ac7e472bc1.mp4

bvaughn:
Hey 👋🏼  We use Linear rather than GitHub to track frontend work so I missed this issue. I did notice, and fix, the underlying problem yesterday in PR #8949 though so I'm going to close this.

PR #8952, opened 2023-03-16
Resolved Input disabled message

Roaster05:
This issue is related to the issue https://github.com/replayio/devtools/issues/8945 and is resolved by placing the content in a container so that they are treated as a single entity and now the overflow for the whole text is treated as a whole rather than breaking it into two parts.
the replay for the changes is:
https://app.replay.io/recording/facebook-log-in-or-sign-up--cd4093f4-ea65-4146-b0d3-a01240f0bcd0


bvaughn:
Hey 👋🏼  My team uses Linear rather than GitHub to track frontend issues, so I missed #8945 until just now.

I did notice, and fix, the underlying problem yesterday in PR #8949 though. Sorry for the waste of time, @Roaster05.

PR #8956, opened 2023-03-16
Implement Replay Tour MVP (Second attempt)

jonbell-lot23:


jonbell-lot23:
This went live last week but there was an issue where existing users were shown the tour.

This fixes that issue. I tested on two dummy accounts to make sure.

PR #8960, opened 2023-03-17
Migrate "range" caches to use the "suspense" package (FE-1330)

hbenl:


bvaughn (commented):
Thanks for doing the integration work with these caches! :bowing_man: After reading more of the PR, I think I understand some of what I was initially confused by. There's one thing that feels a little awkward to me still though, and it's the wrapper functions. It seems like we should be able to get rid of these. I think if we changed the signature of `createAnalysisCache` to lift up some of the mapper functions then we could use those functions from within the `load` method we define inline. Something like:

```ts
export function createAnalysisCache<
  Type extends { point: ExecutionPoint },
  Params extends Array<any>
>(
  createAnalysisParams: (...params: Params) => AnalysisParams,
  transformPoint: (point: PointDescription) => Type,
  debugLabel: string
): IntervalCache<
  ExecutionPoint,
  [ReplayClientInterface, ...Params],
  Type
> {
  return createIntervalCache<
    ExecutionPoint,
    [ReplayClientInterface, ...Params],
    Type
  >({
    comparePoints: compareNumericStrings,
    debugLabel,
    getPointForValue: pointDescription => pointDescription.point,
    load: async (
      start: ExecutionPoint,
      end: ExecutionPoint,
      replayClient: ReplayClientInterface,
      ...params: [...Params, IntervalCacheLoadOptions]
    ) => {
      // AnalysisCache does not use the AbortSignal passed in the load options
      params.pop()!;

      const analysisParams = createAnalysisParams(...params as any as Params);

      // Load then map with transformPoint ...
      return [] as any;
    },
  });
}
```

For caches like `ExceptionsCache` this would be very straight forward:
```ts
() => ({
  exceptions: true,
  mapper: getFunctionBody(exceptionsMapper),
});
```

For `EventsCache` it would be something like:
```ts
(eventType: EventHandlerType) => ({
  eventTypes: [eventType],
  mapper: getFunctionBody(eventsMapper),
});
```

Then the components that use these caches could just use their APIs directly (and we could get rid of the nested `pointsCache` attribute), e.g.
```ts
const results = eventsCache.read(
  range.begin,
  range.end,
  client,
  eventType,
);
```

Or
```ts
const results = logPointAnalysisCache.read(
  range.begin,
  range.end,
  client,
  location,
  code,
  condition
);
```

packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -68,29 +79,33 @@ function getCacheKey(params: AnalysisParams) {
...
+    analysisResultsCaches.set(key, new Map<ExecutionPoint, Record<RemoteAnalysisResult>>());
   }
-  return cacheMap.get(key)!;
+  return analysisResultsCaches.get(key)!;
 }
 
-function createCache<T extends { point: ExecutionPoint }>(
-  params: AnalysisParams,
-  transformPoint: (point: PointDescription) => T
+export function createAnalysisCache<T extends { point: ExecutionPoint }>(
```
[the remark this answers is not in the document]
hbenl: > Why do we need these redundant methods (getResultSuspense, getResultAsync, getResultIfCached)

They're not redundant: we get two types of objects from an analysis: first we get the execution points and then the results for these points. `pointsCache.read()` (which is used by the `LoggablesContext`) returns the points and `getResultSuspense()` (which is used by the message renderers) the results.

packages/replay-next/src/suspense/LogPointAnalysisCache.ts
```diff
@@ -51,15 +64,15 @@ export function getLogPointAnalysisResultSuspense(
...
+    logPointAnalysisCache.pointsCache.readAsync(
+      range.begin,
+      range.end,
+      client,
+      analysisParams,
       point => point
     );
-    cache.getPointsAsync(client, range);
-    const remoteResult = cache.getResultSuspense(point.point);
+    const remoteResult = logPointAnalysisCache.getResultSuspense(analysisParams, point.point);
```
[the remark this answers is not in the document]
hbenl: It's not about the call being non-suspensey. It's about the fact that `logpointAnalysisCache.pointsCache.read...()` needs to be called to run the analysis, but the `LoggablesContext` doesn't do that, so we're doing it here to ensure the analysis is run. I've added a comment. 
Note that this is not an issue for analysis caches in general: the `getResult...()` methods take an execution point as a parameter and you need to call `pointsCache.read...()` to get the points to feed to `getResult...()`. But for the `LogPointAnalysisCache` we know that there is another cache (the `HitPointsCache`) that contains exactly the same points and is shared by different parts of the UI, so we take the points from there instead.

packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -68,31 +72,38 @@ function getCacheKey(params: AnalysisParams) {
   return cacheKey + params.mapper;
 }
 
-export function getAnalysisCache<T extends { point: ExecutionPoint }>(
-  params: AnalysisParams,
-  transformPoint: (point: PointDescription) => T
-): AnalysisCache<T> {
+const analysisResultsCaches = new Map<string, Map<ExecutionPoint, Record<RemoteAnalysisResult>>>();
+function getAnalysisResultsCache(params: AnalysisParams) {
```
bvaughn, line 76: So this function is basically responsible for creating two, synchronized caches: an interval cache (for fetching points/hits) and a regular cache (for evaluating data at a point).

I think we could remove a bit of redundancy here by using `createCache` for the second cache. What do you think?

```ts
export function createAnalysisCache<
  TPoint extends { point: ExecutionPoint },
  TParams extends any[]
>(
  debugLabel: string,
  createAnalysisParams: (...params: TParams) => AnalysisParams,
  transformPoint: (point: PointDescription, ...params: TParams) => TPoint
): AnalysisCache<TPoint, TParams> {
  const resultsCache: Cache<
    [executionPoint: ExecutionPoint, ...params: TParams],
    RemoteAnalysisResult
  > = createCache({
    debugLabel: `${debugLabel} Cache`,
    getKey: ([executionPoint, ...params]) => {
      const analysisParams = createAnalysisParams(...params);
      const baseKey = getCacheKey(analysisParams);
      return `${baseKey}:${executionPoint}`;
    },
    load: async ([executionPoint, ...params]) => {
      // This cache doesn't load its own values;
      // Values are externally cached by the interval cache.
      return new Promise((resolve, reject) => {
        setTimeout(() => {
          reject(
            "Analysis results cache timed out; this may indicate that the points interval cache was not loaded."
          );
        }, 5_000);
      });
    },
  });

  const pointsIntervalCache = createIntervalCache<
    ExecutionPoint,
    [client: ReplayClientInterface, ...params: TParams],
    TPoint
  >({
    debugLabel: `${debugLabel} IntervalCache`,
    getKey: (client, ...params) => getCacheKey(createAnalysisParams(...params)),
    getPointForValue: pointDescription => pointDescription.point,
    comparePoints: compareNumericStrings,
    load: async (begin, end, client, ...paramsWithCacheLoadOptions) => {
      const params = paramsWithCacheLoadOptions.slice(0, -1) as TParams;
      const analysisParams = createAnalysisParams(...params);
      const locations = analysisParams.location
        ? client.getCorrespondingLocations(analysisParams.location).map(location => ({
            location,
          }))
        : undefined;
      if (locations) {
        await Promise.all(
          locations.map(location =>
            breakpointPositionsCache.readAsync(client, location.location.sourceId)
          )
        );
      }

      let allPoints: TPoint[] = [];
      let error: any;
      await client.streamAnalysis(
        {
          locations,
          eventHandlerEntryPoints: analysisParams.eventTypes?.map(eventType => ({ eventType })),
          exceptionPoints: analysisParams.exceptions,
          mapper: analysisParams.mapper,
          effectful: false,
          range: { begin, end },
        },
        {
          onPoints: points => {
            allPoints = allPoints.concat(points.map(point => transformPoint(point, ...params)));
          },
          onResults: analysisEntries => {
            for (const analysisEntry of analysisEntries) {
              const result = analysisEntry.value as RemoteAnalysisResult;
              cachePauseData(client, result.pauseId, result.data);

              // This will resolve any pending/in-flight request from the Cache
              resultsCache.cache(result, result.point, ...params);
            }
          },
          onError: err => (error = err),
        }
      ).pointsFinished;

      if (error) {
        throw error;
      }
      if (allPoints.length > MAX_POINTS_FOR_FULL_ANALYSIS) {
        throw new Error("Too many points to run analysis");
      }
      return allPoints;
    },
  });

  return {
    pointsIntervalCache,
    resultsCache,
  };
}
```

bvaughn: I kind of like that the inner cache will time out with an explicit error if we "mess up" (especially since it's a non-standard way of using the caches)

bvaughn: Note #8973 upgrades `suspense@0.0.29` with a small change I just made to ensure that calling `cache` will clear out/cancel any pending requests (so the above suggestion will work).

packages/replay-next/src/suspense/LogPointAnalysisCache.ts
```diff
@@ -51,15 +67,24 @@ export function getLogPointAnalysisResultSuspense(
...
-      {
-        mapper: createMapperForAnalysis(code, condition),
-        location,
-      },
-      point => point
+    // the LoggablesContext doesn't call logPointAnalysisCache.pointsCache.read(Async)
+    // because it uses points from the HitPointsCache instead (which is more efficient
+    // as it shares the points with other parts of the UI), so we call it here to ensure
+    // that the analysis is run for the given range
+    logPointAnalysisCache.pointsCache.readAsync(
```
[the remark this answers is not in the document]
hbenl: Yeah, I think that would make the code clearer. The interval cache doesn't have a `prefetch` method yet though.

packages/replay-next/src/suspense/AnalysisCache.ts
hbenl: I don't have a strong opinion here although I lean towards keeping it the way it is because using `createCache` with a dummy `load` method feels like we're overusing it.

bvaughn: I prefer a more explicit coupling (with a timeout that will throw ratner than just permanently hang). I'd also like to avoid avoiding duplicating a chunk of the cache API (like the PR is currently doing).

I agree that the way `load` is used is a bit odd. Maybe a special, explicit "externally managed cache" type could alleviate that concern? It would basically be the same thing but its config wouldn't require a `load` method and could instead accept an optional `timeout` param?

```ts
export function createExternallyManagedCache<Params extends Array<any>, Value>(
  options: Omit<CreateCacheOptions<Params, Value>, "load"> & {
    timeout?: number;
    timeoutMessage?: string;
  }
): Cache<Params, Value> {
  const { timeout = 5_000, timeoutMessage = "Timed out", ...rest } = options;

  return createCache<Params, Value>({
    ...rest,
    load: async ([executionPoint, ...params]) => {
      // This cache doesn't load its own values;
      // Values are externally cached by the interval cache.
      return new Promise((resolve, reject) => {
        setTimeout(() => {
          reject(timeoutMessage);
        }, timeout);
      });
    },
  })
}
```

bvaughn: I think part of the reason I prefer this is that there are fewer code paths– so if we have test coverage of one path, or fix a bug in one path, all paths benefit.

It's also nice to have actual `Cache` objects because they can be used with the other "suspense" hooks where as a one-off faux-cache can't be.

PR #8965, opened 2023-03-18
Set `overscroll-behavior: none;` on the `SourceList`

Andarist:
I went back in the browser's history accidentally on multiple occasions cause of the default behavior of the overscroll. I find this default behavior often annoying for "long-lived" scrollable containers (like this `SourceList`) so I propose to turn it off.


jasonLaster:
Oh wow. I didn't know this was a thing

bvaughn (approved):

PR #8968, opened 2023-03-20
Upgrade suspense@0.0.28

bvaughn:
Adds the following capabilities:
* Removed `useWeakRef` config option for `createCache` and replaced with `getCache` option in order to support LRU type caches in addition to `WeakRef` based caches. Thanks to @cevr for contributing to this release!
* Add `useImperativeIntervalCacheValue` hook for imperatively loading and subscribing to `createIntervalCache` data.
* Add `getValue` and `getValueIfCached` methods to `createIntervalCache` type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "reselect": "^4.1.5",
     "shared": "workspace:*",
     "slugify": "^1.6.5",
-    "suspense": "^0.0.26"
+    "suspense": "^0.0.28"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


hbenl (approved):

bvaughn:
This is weird. I pushed two commits to this branch, and the both show up in the branch history on GitHub, but only one shows up in the PR.

hbenl:
"scroll position should be restored when switching between sources" should be fixed in #8895

PR #8976, opened 2023-03-22
Move outline parsing to the backend (FE-1184)

hbenl:


src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
```diff
@@ -196,12 +199,20 @@ export default function SourceOutlineWrapper() {
...
+    replayClient,
+    selectedSource?.id
   );
+  if (symbolsCacheValue.status === "resolved") {
+    symbols = symbolsCacheValue.value;
+  }
 
   return (
-    <ErrorBoundary>
+    <ErrorBoundary key={selectedSource?.id}>
```
bvaughn, line 215: 👍🏼

packages/replay-next/src/suspense/SourceOutlineCache.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const sourceOutlineCache = createCache<
+  [replayClient: ReplayClientInterface, sourceId: SourceId | undefined],
+  getSourceOutlineResult
+>({
+  debugLabel: "sourceOutlineCache",
+  getKey: ([replayClient, sourceId]) => sourceId ?? "",
+  load: ([replayClient, sourceId]) =>
+    sourceId ? replayClient.getSourceOutline(sourceId) : { functions: [], classes: [] },
+});
```
bvaughn, line 14: Unlike hooks, Suspense caches can be called conditionally during render. This cache should _require_ a source ID and it should be on the caller to decide what to do if one is not provided.
```suggestion
export const sourceOutlineCache = createCache<
  [replayClient: ReplayClientInterface, sourceId: SourceId],
  getSourceOutlineResult
>({
  debugLabel: "sourceOutlineCache",
  getKey: ([replayClient, sourceId]) => sourceId,
  load: ([replayClient, sourceId]) => replayClient.getSourceOutline(sourceId),
});
```

bvaughn (approved):
Looks great. Nice cleanup!

Just one thing I think we should change.

packages/replay-next/src/suspense/SourceOutlineCache.ts
hbenl: The problem was that this cache is used with the `useImperativeCacheValue` hook in the `QuickOpenModal` - and that shouldn't block on fetching the source outline.

PR #8978, opened 2023-03-22
Use SourceOutline to find original function names (FE-1342)

hbenl:


packages/replay-next/src/utils/source.ts
```diff
@@ -62,3 +67,48 @@ export function isNodeModule(source?: ProtocolSource): boolean {
 
   return source.url.includes("node_modules");
 }
+
+export function isSourceMappedSource(sourceId: SourceId, sources: ProtocolSource[]): boolean {
+  const source = sources.find(source => source.sourceId === sourceId);
+  if (!source) {
+    return false;
```
bvaughn, line 74: Should we be throwing here instead? This seems like a very unexpected case.

bvaughn (approved):

PR #8979, opened 2023-03-22
Reuse The function to convert milliseconds into timestamp string

adarsh-sgh:
This pull request is a follow up to https://github.com/replayio/devtools/pull/8897#pullrequestreview-1347005359 

### Unit Tests
all existing unit tests are passing after refactor
image
***
merging this PR will also fix the issue of offset of certain minutes in few timezones
e.g. below is a screenshot when we hover over a replay in IST timezone
image
the timestamp shown in tooltip (30:12) is 30 minutes ahead of actual timestamp
it's because of fix in https://github.com/replayio/devtools/pull/8897

@bvaughn could you please review this.

### DIFF
--- a/packages/shared/utils/time.test.ts
+++ b/packages/shared/utils/time.test.ts
@@ -1,6 +1,6 @@
 import { ExecutionPoint, TimeStampedPointRange } from "@replayio/protocol";
 
-import { isRangeInRegions } from "./time";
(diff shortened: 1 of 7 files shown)


bvaughn (commented):
Looks good to me. Let's make sure CI is happy.

PR #8981, opened 2023-03-22
Improve Source line scroll alignment

bvaughn:
The "jump to code" button in our events panel uses `react-window` "smart" align, which mimics VS Code:
* If the line is partially visible, scroll just enough so that it's fully visible and no more.
* Otherwise scroll so that it's center aligned (assuming height allows).

I wrote the `react-window` scroll API with the assumption that the list was already visible, but in Replay's case– jumping to a source location may open a different source tab, at which point– it's preferable for us to center-align a line _even if it would have been visible otherwise_.

This PR adds a ref to track the first mount state, and updates our alignment mode based on that ref. Feels like an improvement:

https://user-images.githubusercontent.com/29597/227026158-06ece5b4-3334-422d-9368-2a068648efee.mp4


jasonLaster:
❤️

bvaughn:
This is small enough that I'm going to just land it. Happy to make changes _later_ if someone does eventually review it.

PR #8982, opened 2023-03-22
Re-enable React Timeline panel and rework to cache + stream results

markerikson:
This PR:

- Re-enables showing the React Timeline panel, which had been inadvertently removed in the Replay Tour work
- Heavily rewrites the panel processing logic:
  - The primary searching for hits _inside_ React are now fetched in a streaming cache, so they will pop in over time.  (This is slow, but at least it shows _something_ happening instead of waiting for all up-to-200 hits to come back.)
      - I was seeing `createPause` times get _extremely_ slow when I tried to create large numbers of pauses simultaneously, which I've seen while doing routines work too.
  - Moves the _exact_ calculation of the final file location and potential timestamp to be done when you click "Jump to Code", just to cut down on the amount of processing down during the initial search
  - Switches from using analysis to using `getFrameSteps` to identify the right time.  This does work, but with caveats:
      - I ran into issues where the column of the original frame did not match the columns in the frame steps, such as an `onClick={() => dispatch(increment())}` having the frame point to the `dispatch` (column 33), but the frame _steps_ pointing to the `onClick=` (column 25).
      - The [recording I was testing against](https://app.replay.io/recording/breakpoints-03-test-stepping-forward-through-breakpoints-when-rewound-before-the-first-one--07c4ac6d-ac19-4cd8-a3b7-fadda36b3e39) seems to have a bunch of cases where execution clearly _should_ have hit a certain line, and the _frame_ says it hit that line, but I'm not getting back any breakable positions or hit counts for that line.
      - Given that, I half-gave-up, and am jumping to the location even if there's no exact frame step / execution point time to seek to

In general, this still needs a lot of further work to figure out why these load relatively slowly, why there's frame step / frame mismatches, etc.

But I think it's generally finding the right location and opening it consistently.

### DIFF
--- a/src/ui/components/Events/Events.tsx
+++ b/src/ui/components/Events/Events.tsx
@@ -11,7 +11,7 @@ import { trackEvent } from "ui/utils/telemetry";
 
 import Event from "./Event";
 
-function CurrentTimeLine({ isActive }: { isActive: boolean }) {
+export function CurrentTimeLine({ isActive }: { isActive: boolean }) {
   return (
     <div
       className={classNames("m-0", isActive ? "bg-secondaryAccent" : "bg-transparent")}
--- a/src/ui/components/ReactPanel.tsx
+++ b/src/ui/components/ReactPanel.tsx
@@ -1,8 +1,21 @@
-import { TimeStampedPoint, TimeStampedPointRange } from "@replayio/protocol";
+import { createSelector } from "@reduxjs/toolkit";
+import {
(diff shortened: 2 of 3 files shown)


jasonLaster (commented):
Nice! Will let others review the code, but excited to see it coming back

markerikson:
fyi, I want to get this into prod by tomorrow morning so it's available for the video with Jason Lengstorf

PR #8983, opened 2023-03-23
[DRAFT] Implement Replay Assist

jonbell-lot23:
**Status**

- [x] Hook up each step (aka nag) in the code
- [x] Get a preliminary review from the FE team
- [x] Address the obvious stuff in the code review
- [x] Address the hard stuff
- [x] Build animated gifs and captions for each step (aka nag)
- [x] Put everything behind an experimental flag
- [ ] Get a second code review
- [x] Push things behind the flag and then iterate on the polish from there

**Updated note to FE team**

Hello again! I think I've addressed all the issues called out in the previous code review, now I'm looking for a follow-up code review. I have things behind an experimental flag, so I'm hoping to get this code good enough to push to prod, then I have some visual updates to work on from there.  I'd like to get this live soon so I can avoid a bunch of merge conflicts that happen when code has been in a branch for a while.

~~I touched a lot of files and there was a fair amount of flailing involved. (for example, see [PrimaryPanes/SourcesTree.tsx](https://github.com/replayio/devtools/pull/8983#diff-1949eb2aa612ae7066aec74ec8d34442d94cc8e73ec17bc4d805e913b9fa50b2)) I'm hoping for a preliminary code review to see how far off base my approach may or may not be. This is still really rough -- there's going to be console.logs in there, areas I didn't clean up, etc. After this preliminary review, I'll be working on polish (both UX and code) and then the real review will come later. Thanks!~~

ReplayAssist2

[Notion link](https://www.notion.so/replayio/Replay-Assist-ec41adc7020449a7861ac86da8c6d048?pvs=4)


bvaughn:
By the way– when this one is ready, let's run it through an actual review process 😄

jonbell-lot23:
> By the way– when this one is ready, let's run it through our normal review process? 😄
> Seems like the feature is complex enough to benefit from going through review review.

I agree 100%. This is a giant feature, with complex (for me) code, so I'll be sure to run it through the official channels.

jasonLaster:
Love that gradient

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTree.tsx
jonbell-lot23, line 1: This worked, but it seems pretty hacky. Curious to know how this feels. Worst idea ever? Not so bad?

src/devtools/client/debugger/src/components/PrimaryPanes/SourcesTree.tsx
```diff
@@ -94,21 +89,33 @@ interface STState {
...
+    const updatedProps = {
+      ...props,
+      cx: { navigateCounter: 0 },
+      sourcesLoading: false,
+      shownSource: null,
+      selectedSource: null,
+      focused: null,
+      sources: {},
+      sourceCount: 0,
+    };
```
bvaughn, line 101: Can you explain what this is doing and why it's necessary? `props` should always come from _outside_ of a component. It's okay to set _default_ values, but this is actually overriding any external values with hard-coded values (that won't ever change) so this seems like a bug for sure.

src/ui/hooks/users.ts
```diff
@@ -73,6 +73,21 @@ export enum Nag {
...
+  ADD_UNICORN_BADGE = "add_unicorn_badge",
+  RECORD_REPLAY = "record_replay",
+  EXPLORE_SOURCES = "explore_sources",
+  SEARCH_SOURCE_TEXT = "search_source_text",
+  QUICK_OPEN_FILE = "quick_open_file",
+  LAUNCH_COMMAND_PALETTE = "launch_command_palette",
+  JUMP_TO_EVENT = "jump_to_event",
+  INSPECT_ELEMENT = "inspect_element",
+  INSPECT_COMPONENT = "inspect_component",
+  USE_FOCUS_MODE = "use_focus_mode",
```
bvaughn, line 90: nit: It's nice to alpha-sort key values like this. (Easier to visually scan for a value and less likely to have a merge conflict if two PRs modify the same object.)

src/ui/components/Timeline/FocusModePopout.tsx
```diff
@@ -15,6 +17,8 @@ import Icon from "../shared/Icon";
 import styles from "./FocusModePopout.module.css";
 
 export default function FocusModePopout() {
+  const [useFocusModeState, dismissUseFocusModeNag] = useNag(Nag.USE_FOCUS_MODE);
```
bvaughn, line 20: nit: For readability
```suggestion
  const [, dismissUseFocusModeNag] = useNag(Nag.USE_FOCUS_MODE);
```

src/ui/components/NodePicker.tsx
```diff
@@ -40,6 +42,7 @@ const nodePickerInstance = new NodePickerClass();
 
 export function NodePicker() {
   const dispatch = useAppDispatch();
+  const [inspectElementState, dismissInspectElementNag] = useNag(Nag.INSPECT_ELEMENT); // Replay Assist
```
bvaughn, line 45: nit for readability
```suggestion
  const [, dismissInspectElementNag] = useNag(Nag.INSPECT_ELEMENT); // Replay Assist
```

src/ui/components/NagDismiss.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+    useEffect(() => {
+      dismissNag();
+    }, [dismissNag]);
+
+    return <WrappedComponent {...props} />;
+  };
+};
+
+export default NagDismiss;
```
bvaughn, line 19: I don't think this component is being used anywhere, so we should delete it.

_If_ we wanted to go with an auto-dismiss-nag-on-mount approach, I think we'd probably want to use a hook instead since it's a lot "lighter" in terms of the React overhead.

```ts
function useNagDismissOnMount(nagType: Nag) {
  const [, dismissNag] = useNag(nagType);

  useEffect(() => {
    dismissNag();
  }, [dismissNag]);
}
```

src/ui/components/Header/UserOptions.tsx
```diff
@@ -13,9 +14,10 @@ import ExternalLink from "../shared/ExternalLink";
 
 interface UserOptionsProps extends PropsFromRedux {
   noBrowserItem?: boolean;
+  toggleReplayAssist: () => void;
 }
 
-function UserOptions({ setModal, noBrowserItem }: UserOptionsProps) {
+function UserOptions({ setModal, noBrowserItem, toggleReplayAssist }: UserOptionsProps) {
```
bvaughn, line 20: The new `toggleReplayAssist` prop isn't being used. Remove it?

src/ui/components/Header/Header.tsx
```diff
@@ -161,7 +169,9 @@ function HeaderTitle({
   );
 }
 
-export default function Header() {
+interface HeaderProps extends PropsFromRedux {}
+
+function Header({ toggleReplayAssist }: HeaderProps) {
```
bvaughn, line 174: This new prop isn't being used. Remove it?

src/ui/components/Header/Header.tsx
```diff
@@ -62,7 +70,7 @@ function Links({ recordingTarget }: { recordingTarget: RecordingTarget | null })
       {showShareButton ? <ShareButton /> : null}
       <Avatars recordingId={recordingId} />
       {showViewToggle && <ViewToggle />}
-      <UserOptions />
+      <UserOptions toggleReplayAssist={toggleReplayAssist} />
```
bvaughn, line 73: `UserOptions` isn't using this new prop. Remove it (I think we can revert all of the changes in this file?

src/ui/components/Events/Event.tsx
```diff
@@ -282,11 +284,14 @@ export default React.memo(function Event({
   const label = getEventLabel(event);
   const { icon } = getReplayEvent(kind);
   const [jumpToCodeStatus, setJumpToCodeStatus] = useState<JumpToCodeStatus>("not_checked");
+  const [jumpToCodeState, dismissJumpToCodeNag] = useNag(Nag.JUMP_TO_CODE);
+  const [jumpToEventState, dismissJumpToEventNag] = useNag(Nag.JUMP_TO_EVENT);
```
bvaughn, line 288: nit for readability
```suggestion
  const [, dismissJumpToCodeNag] = useNag(Nag.JUMP_TO_CODE);
  const [, dismissJumpToEventNag] = useNag(Nag.JUMP_TO_EVENT);
```

src/ui/components/Comments/EditableRemark.tsx
```diff
@@ -51,6 +53,13 @@ export default function EditableRemark({
     setIsEditing(false);
   };
 
+  const handleSaveChanges = (editorState: SerializedEditorState) => {
+    saveChanges(editorState);
+    dismissAddCommentNag();
+  };
+
+  const [addCommentState, dismissAddCommentNag] = useNag(Nag.ADD_COMMENT);
```
bvaughn, line 61: Readability nit
```suggestion
  const [, dismissAddCommentNag] = useNag(Nag.ADD_COMMENT);
```

src/ui/components/Comments/EditableRemark.tsx
```diff
@@ -133,7 +142,7 @@ export default function EditableRemark({
           initialValue={content}
           onCancel={discardPendingChanges}
           onDelete={deleteRemark}
-          onSave={saveChanges}
+          onSave={handleSaveChanges} // Replace 'saveChanges' with 'handleSaveChanges'
```
bvaughn, line 145: Remove this TODO comment?

src/ui/components/CommandPalette/CommandPalette.tsx
```diff
@@ -117,6 +119,14 @@ function CommandPalette({
   const [activeIndex, setActiveIndex] = useState(0);
   const shownCommands = getShownCommands(searchString);
 
+  const [launchCommandPaletteState, dismissLaunchCommandPaletteNag] = useNag(
```
bvaughn, line 122: Readability nit
```suggestion
  const [, dismissLaunchCommandPaletteNag] = useNag(
```

src/ui/components/ReplayAssistButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  useEffect(() => {
+    const storedValue = localStorage.getItem("replayAssistEnabled");
+    if (storedValue) {
+      setShouldShowReplayAssist(JSON.parse(storedValue));
+    }
+  }, []);
+
+  useEffect(() => {
+    localStorage.setItem("replayAssistEnabled", JSON.stringify(shouldShowReplayAssist));
+  }, [shouldShowReplayAssist]);
```
bvaughn, line 23: You can simplify the local storage integration here with the `useLocalStorage` hook:
```suggestion
const dispatch = useDispatch();

const [shouldShowReplayAssist, setShouldShowReplayAssist] =
  useLocalStorage<boolean>(`Replay:replayAssistEnabled`, false);
```

It looks like there's also a syncing bug here though. The Redux state defaults to `false` but the component state defaults to whatever the most recent value in local storage was. These two could get out of sync, which is problematic.

src/ui/components/ReplayAssistButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <div className="row row-flat rounded-none border-t border-b">
+      <input
+        type="checkbox"
+        className="rounded-sm border-gray-400"
+        id="replay-assist-checkbox"
+        checked={shouldShowReplayAssist}
+        onChange={() => {
+          const newValue = !shouldShowReplayAssist;
+          setShouldShowReplayAssist(newValue);
+          dispatch(toggleReplayAssist());
```
bvaughn, line 39: Mixing set-boolean and toggle actions runs the risk of mismatches. Replace `toggleReplayAssist` with a Redux setter to make sure they stay in sync:
```js
dispatch(setReplayAssist(newValue));
```

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,284 @@
...
+  };
+
+  /*
+  const renderClickInfo = () => {
+    if (selectedItem >= 0) {
+      return <div className={styles.clickInfo}>You clicked on {checklistItems[selectedItem]}</div>;
+    }
+    return null;
+  };
+  */
```
bvaughn, line 171: Delete dead code.

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,284 @@
...
+    const initialPrimaryPanel = "events";
+    dispatch(setSelectedPrimaryPanel(initialPrimaryPanel));
+    dismissNag(Nag.DISMISS_TOUR);
+  };
+
+  return { dismissTourNag };
+};
+
+const Assist: React.FC = () => {
+  const [selectedItem, setSelectedItem] = useState(1);
```
bvaughn, line 57: If this is a 0-based value, let's call it "index". If it's a 1-based value, let's call it "step" or "number" or something. ( Otherwise it's likely to cause off-by-one error to someone less familiar with the code later.)

Also– why does it default to 1 when there is a 0? Is this a bug?

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,284 @@
...
+        "https://website-git-master-recordreplay.vercel.app/assist/inspect_an_element.gif"
+      );
+    } else if (index === 15) {
+      setVideoUrl(
+        "https://website-git-master-recordreplay.vercel.app/assist/inspect_an_component.gif"
+      );
+    } else if (index === 16) {
+      setVideoUrl("https://website-git-master-recordreplay.vercel.app/assist/use_focus_mode.gif");
+    }
+  };
```
bvaughn, line 151: Kind of a nit, but (1) this would probably be a little easier to read as a `switch` statement and (2) the video URL doesn't need to be its own state value (at least given the fact that it's always derived from the index).

I would structure this like so:
```ts
const [stepIndex, setStepIndex] = useState(0);

let videoUrl;
switch (stepIndex) {
  case 0:
    videoUrl =
      "https://website-git-master-recordreplay.vercel.app/assist/open_devtools.gif";
    break;
  case 1:
    videoUrl =
      "https://website-git-master-recordreplay.vercel.app/assist/time_travel_in_console.gif";
    break;
  // ...
}
```

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,284 @@
...
+          <div className="p-0 pt-3">
+            <div className={styles.h1}>Replay Assist</div>
+            <div className={styles.AssistBoxInternal}>
+              <div className={styles.checklist}>
+                {checklistItems.map((item, index) => (
+                  <div
+                    key={index}
+                    className={`flex ${getItemStyle(index)} ${styles.checklistItem}`}
+                    onClick={() => handleClick(index)}
+                  >
```
bvaughn, line 266: The selected item above is initialized to 1:
```js
const [selectedItem, setSelectedItem] = useState(1);
```

This index-based approach means that the first item in the list is index 0 though. That seems like a bug?

jonbell-lot23:
Thank you Brian! This is very helpful, and I appreciate the time commitment. 

Thinking out loud, here are my steps to getting this code in tip-top shape:

- [ ] Obvious stuff: remove comments, name variables better, delete unused references
- [ ] Conflicts: I have two conflicts to merge, shouldn't be too hard
- [ ] The big stuff: there are a few items in here where I was way out of my depth on, so I'm going to need to figure out how to handle those. I have a plan.

I still have a ways to go, but things are getting closer!

packages/replay-next/components/sources/log-point-panel/BadgePicker.tsx
```diff
@@ -22,8 +24,11 @@ export default function BadgePicker({
   const { editPointBadge } = useContext(PointsContext);
 
   const [state, setState] = useState<State>("initial");
-
+  const [addUnicornBadgeState, dismissAddUnicornBadgeNag] = useNag(Nag.ADD_UNICORN_BADGE);
```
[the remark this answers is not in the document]
jonbell-lot23: TIL I learned you can do this!

src/ui/components/Tour/Tour.tsx
```diff
@@ -160,19 +160,31 @@ const Tour: React.FC = () => {
...
+          <img
+            src="https://website-git-master-recordreplay.vercel.app/tour/editlogs.gif"
+            className={styles.videoExample}
+          />
         )}
 
         {hasCompletedTour && (
-          <img src="/images/tour/consoleupdate.gif" className={styles.videoExample} />
+          <img
+            src="https://website-git-master-recordreplay.vercel.app/tour/consoleupdate.gif"
```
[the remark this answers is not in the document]
jonbell-lot23: Two topics here:

1. I'm pointing to vercel but I need to change this to static.replay
2. But Ryan did suggest pointing to static.replay because I'm about to put a ton of videos and images into our app, and it's better to host them in a static place rather than bloating our package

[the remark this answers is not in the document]
bvaughn: I'm not sure how static.replay works. (I guess that's where we deploy our protocol docs?) But I guess that's fine if that's what Ryan suggested. We shouldn't be pointing to website-git-master-recordreplay.vercel.app though.

bvaughn:
Gotcha 👍🏼  Ping me when you're ready for me to take another look 😄

src/ui/components/ReplayAssistButton.tsx
jonbell-lot23: Leaving a note for myself that I understand the point here, but not how to fix it.

src/ui/components/Assist/Assist.tsx
jonbell-lot23: Good eye, this was because I wanted to pick that item in the list. Not a bug, but a better solution is "pick the first item that the user hasn't yet completed."

src/ui/components/ReplayAssistButton.tsx
jonbell-lot23: Hey @bvaughn I'm not super familiar with Redux setters and I'm having trouble getting this working. Can you share the steps I'll be taking so I can make sure I'm going in the right direction?

jonbell-lot23: I figured it out!

jonbell-lot23:
Hello again! I think I'm ready for another code review. See note I added in the description. Thank you!

src/ui/reducers/app.ts
```diff
@@ -60,6 +61,9 @@ const appSlice = createSlice({
   name: "app",
   initialState: initialAppState,
   reducers: {
+    replayAssistToggled(state) {
+      state.showReplayAssist = !state.showReplayAssist;
+    },
```
bvaughn, line 66: For simple actions like this– ones that only update one piece of state– the name of an action generally _matches_ the name of the state.

So I think this should be either `setShowReplayAssist` or `toggleShowReplayAssist`

src/ui/components/Tour/Tour.tsx
```diff
@@ -157,19 +157,31 @@ const Tour: React.FC = () => {
         )}
 
         {!isNewUser && showConsoleNavigate && showBreakpointAdd && showBreakpointEdit && (
-          <img src="/images/tour/fast-forward.gif" className={styles.videoExample} />
+          <img
+            src="https://website-git-master-recordreplay.vercel.app/tour/fast-forward.gif"
```
bvaughn, line 161: These URLs still look wrong.

I don't know what "website-git-master-recordreplay.vercel.app" or how to deploy things there.

src/ui/reducers/app.ts
```diff
@@ -194,6 +199,8 @@ export default appSlice.reducer;
 const getSelectedPanel = (state: UIState) => state.layout.selectedPanel;
 const getViewMode = (state: UIState) => state.layout.viewMode;
 
+export const getReplayAssist = (state: UIState) => state.app.showReplayAssist;
```
bvaughn, line 202: Similar nit about "get blah" vs "show blah" – these names should probably better match.

src/ui/components/ReplayAssistButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { getReplayAssist, replayAssistToggled } from "ui/reducers/app";
+import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
+import useAuth0 from "ui/utils/useAuth0";
+
+function toggleAndShowAssistPanel(): UIThunkAction {
+  return (dispatch, getState) => {
+    dispatch(replayAssistToggled());
+    const newValue = getReplayAssist(getState());
+    if (!newValue) {
+      dispatch(setSelectedPrimaryPanel("events"));
```
bvaughn, line 15: Toggling one panel off should just hide that panel. It shouldn't show another panel. That's how the other panels work:

https://user-images.githubusercontent.com/29597/229527858-b0874d2f-424c-4ba9-bb3e-90d380005cdf.mp4

src/ui/components/NodePicker.tsx
```diff
@@ -1,12 +1,14 @@
 import { NodeBounds } from "@replayio/protocol";
 import classnames from "classnames";
-import React, { useCallback, useLayoutEffect, useRef, useState } from "react";
+import React, { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
```
bvaughn, line 3: nit: new unused import `useEffect`

src/ui/components/Comments/EditableRemark.tsx
bvaughn: This nit still seems relevant ^ FWIW

src/ui/components/Comments/EditableRemark.tsx
```diff
@@ -51,6 +53,13 @@ export default function EditableRemark({
     setIsEditing(false);
   };
 
+  const handleSaveChanges = (editorState: SerializedEditorState) => {
+    saveChanges(editorState);
+    dismissAddCommentNag();
```
bvaughn, line 58: I think this call to `dismissAddCommentNag` should be put inside of `saveChanges`

I don't think there's any reason to add a new wrapper function that just does one more unconditional call, especially when `saveChanges` is never called without it.

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const info = useTestInfo();
+
+  const dismissTourNag = () => {
+    const initialPrimaryPanel = "events";
+    dispatch(setSelectedPrimaryPanel(initialPrimaryPanel));
+    dismissNag(Nag.DISMISS_TOUR);
+  };
+
+  return { dismissTourNag };
+};
```
bvaughn, line 54: Nothing' using this.
```suggestion
```

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const showAddUnicornBadge = shouldShowAddUnicornBadge(nags);
+  const showRecordReplay = shouldShowRecordReplay(nags);
+  const showExploreSources = shouldShowExploreSources(nags);
+  const showSearchSourceText = shouldShowSearchSourceText(nags);
+  const showQuickOpenFile = shouldShowQuickOpenFile(nags);
+  const showLaunchCommandPalette = shouldShowLaunchCommandPalette(nags);
+  const showJumpToEvent = shouldShowJumpToEvent(nags);
+  const showInspectElement = shouldShowInspectElement(nags);
+  const showInspectComponent = shouldShowInspectComponent(nags);
+  const showUseFocusMode = shouldShowUseFocusMode(nags);
```
bvaughn, line 78: About half of these values aren't being used, so they should be deleted:
```suggestion
  const showConsoleNavigate = shouldShowConsoleNavigate(nags);
  const showBreakpointEdit = shouldShowBreakpointEdit(nags);
  const showAddComment = shouldShowAddComment(nags);
  const showJumpToCode = shouldShowJumpToCode(nags);
  const showAddUnicornBadge = shouldShowAddUnicornBadge(nags);
  const showRecordReplay = shouldShowRecordReplay(nags);
  const showExploreSources = shouldShowExploreSources(nags);
  const showSearchSourceText = shouldShowSearchSourceText(nags);
  const showLaunchCommandPalette = shouldShowLaunchCommandPalette(nags);
  const showJumpToEvent = shouldShowJumpToEvent(nags);
  const showInspectElement = shouldShowInspectElement(nags);
  const showUseFocusMode = shouldShowUseFocusMode(nags);
```

FWIW dead code like this usually shows up as _gray_ in VS Code to help you spot it.

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,206 @@
...
+  shouldShowInspectElement,
+  shouldShowJumpToCode,
+  shouldShowJumpToEvent,
+  shouldShowLaunchCommandPalette,
+  shouldShowQuickOpenFile,
+  shouldShowRecordReplay,
+  shouldShowSearchSourceText,
+  shouldShowTour,
+  shouldShowUseFocusMode,
+} from "ui/utils/onboarding";
```
bvaughn, line 38: These imports can be reduced too because the usage of them below can be removed:
```suggestion
import {
  shouldShowAddComment,
  shouldShowAddUnicornBadge,
  shouldShowBreakpointEdit,
  shouldShowConsoleNavigate,
  shouldShowExploreSources,
  shouldShowInspectElement,
  shouldShowJumpToCode,
  shouldShowJumpToEvent,
  shouldShowLaunchCommandPalette,
  shouldShowRecordReplay,
  shouldShowSearchSourceText,
  shouldShowTour,
  shouldShowUseFocusMode,
} from "ui/utils/onboarding";
```

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,206 @@
+import React, { useEffect, useState } from "react";
+
+import Icon from "replay-next/components/Icon";
+import useLocalStorage from "replay-next/src/hooks/useLocalStorage";
+import { setViewMode } from "ui/actions/layout";
+import { setSelectedPrimaryPanel } from "ui/actions/layout";
+import Events from "ui/components/Events";
+import { shouldShowDevToolsNag } from "ui/components/Header/ViewToggle";
```
bvaughn, line 8: This can be removed.
```suggestion
```

src/ui/components/Assist/Assist.tsx
```diff
@@ -0,0 +1,206 @@
...
+    },
+  ];
+
+  useEffect(() => {
+    const firstUnselectedIndex = updatedChecklistItems.findIndex(item => !item.completed);
+    if (firstUnselectedIndex !== -1) {
+      setSelectedItem(firstUnselectedIndex);
+      setStepIndex(firstUnselectedIndex);
+    }
+  }, []);
```
bvaughn, line 171: nit: There's still no reason to store `selectedItem` in state.  It can be calculated from the `stepIndex` (also stored in state) and storing them separately can lead to mismatches.

It's not a big deal in this case since the component is small and it's unlikely that one value would be updated without the other, but just an FYI.

This could be simplified to use only one piece of state like so:
```ts
const [stepIndex, setStepIndex] = useState(0);
const selectedItem = updatedChecklistItems[stepIndex];
```

src/ui/components/Assist/Assist.tsx
jonbell-lot23: Thanks! I didn't know about that VS Code feature, that's going to come in quite handy.

src/ui/components/Tour/Tour.tsx
jonbell-lot23: Good call, we're going to standardise on https://vercel.replay.io/.

We're doing it this way to avoid having a ton of big media files in devtools. These are pushed via the website repo.

jonbell-lot23:
Hey @bvaughn I really appreciate your recent review. I didn't know about the dimmed dead code feature in VS Code because I prefer another IDE that doesn't have that feature. But I'll be using that from now on!

Dead code comments are a pretty big waste of your time. In the future, I want to make sure I'm using your attention for truly important code feedback. Thanks for helping me improve, and thanks again for the time commitment.

src/ui/components/Assist/Assist.tsx
bvaughn: This comment keeps getting marked as resolved, but the change I'm suggesting seems to be getting overlooked. Any reason?

bvaughn:
Let me know when this is ready for me to take another look. I saw your comment and thought it was but I think maybe I jumped the gun so I'll wait.

src/ui/reducers/app.ts
jonbell-lot23: In this case, I was trying to align to @markerikson 's suggestion for these items to be human-readable while reading through a series of steps.

But he don't *force* me to change it, and I understand it's a matter of style/taste/preference. CC'ing Mark to discuss the best way forward here, I have no opinions but wanted to call out the question.

bvaughn: Ah, I see.

I find it kind of confusing to have Redux state, React component state, and Redux actions all with slightly different naming variations. It's more mental overhead of things you have to keep track of while working in the code.

jonbell-lot23: I'm going to close this out as a won't fix for now, but I want to stress that I'm totally fine to change this!

src/ui/components/ReplayAssistButton.tsx
jonbell-lot23: In this case it's coming from the toggle itself, not a menu element ->

MenuToggle

Without this text, you can toggle off the Replay Assist pane, but then nothing is selected on the side, leading to a broken experience.

src/ui/components/Comments/EditableRemark.tsx
jonbell-lot23: True

src/ui/components/Assist/Assist.tsx
jonbell-lot23: Sorry, I thought I had addressed it. Should be better now.

jonbell-lot23:
Hi @bvaughn ! 

1. I think I only have one more unresolved issue, and I'm stuck on it. See above comments.
2. Between some clumsy merge conflicts a few days ago that might have reverted stuff and general GitHub UI confusion, it's possible I missed some items. I hope not!

I'm ready for another code review. Here's how I'm hoping things go:

0. This doesn't take much more of your time!
1. You have a quick suggestion for the final issue I called out
2. We can push this live (behind an experimental flag)
3. Then I have some great design concepts coming but I'd like them to be in their own branch because they're kind of bold

src/ui/reducers/app.ts
bvaughn: After that variable rename– there is a naming conflict. (You have two things named `showReplayAssist`).

One way to fix that is with an import alias:
```ts
import { showReplayAssist as showReplayAssistSelector } from "ui/reducers/app";
```

And then down in the component bodyL
```ts
const showReplayAssist = useAppSelector(showReplayAssistSelector);
```

That being said, maybe it's easier to ignore this and I'll do a small (sort of subjective) naming sweep myself.

src/ui/components/Assist/Assist.tsx
bvaughn: Hm. It's still the same 😄

src/ui/components/Comments/EditableRemark.tsx
bvaughn: This hasn't changed.

src/ui/components/Comments/EditableRemark.tsx
bvaughn: This hasn't changed.

bvaughn (commented):
Btw I'm going to just jump in on this branch if that's okay. 😄

bvaughn:
I think something is not wired up correctly because I see the Replay assist tab even though the experiment is disabled.

Screen Shot 2023-04-05 at 9 52 00 AM

I'll look into this shortly.

src/ui/components/Assist/Assist.tsx
bvaughn: Ah, I think I was seeing a stale version in GitHub. It _has_ changed, although there's now an unused value:

Screen Shot 2023-04-05 at 10 12 40 AM

I'll update it !

bvaughn:
Ah, I think I see a higher level issue going on which maybe caused the above experimental leak– there's two pieces of "state" controlling the Replay Assist feature. There's the experimental feature (normal) and the app-level Redux flag this PR adds (unusual).

The former is toggled in the Experiments panel:
Screen Shot 2023-04-05 at 9 52 00 AM

And the latter is toggled in the main drop down menu:
Screen Shot 2023-04-05 at 10 19 46 AM

This latter toggle looks kind of bad to me. Why do we need a _second_ way to select that panel? It's already visible in the left Toolbar, one click away?

src/ui/components/Toolbar.tsx
```diff
@@ -210,6 +207,15 @@ export default function Toolbar() {
           />
         ) : null}
 
+        {replayAssistExperimentEnabled && showReplayAssist ? (
```
bvaughn, line 210: This place should be checking both the Redux state (if we decide to keep this) _and_ the experimental setting (so that it doesn't leak).

Personally I think the redundant Redux state is a bit odd.

PR #8985, opened 2023-03-23
Show marker on timeline when hovering over an event

bvaughn:
I re-purposed the `PreviewMarkers` component to show this since it seemed like the same _type_ of thing.

https://user-images.githubusercontent.com/29597/227233550-7608bca4-3ff0-47cb-b37b-0501754e1694.mp4

cc @jonbell-lot23


src/ui/components/Timeline/PreviewMarkers.tsx
```diff
@@ -63,40 +72,41 @@ function PreviewMarkers() {
...
       : [null, null];
 
-  if (
-    hitPointStatus === "too-many-points-to-run-analysis" ||
-    hitPointStatus === "too-many-points-to-find" ||
-    hitPoints == null
-  ) {
-    return null;
-  }
+  const showHitPointMakers =
```
markerikson, line 75: :hand: Typo:  "Makers" -> "Markers"

src/ui/reducers/timeline.ts
```diff
@@ -15,8 +16,9 @@ function initialTimelineState(): TimelineState {
     focusRegion: getPausePointParams().focusRegion,
     focusRegionBackup: null,
     displayedFocusRegion: null,
-    hoverTime: null,
+    hoveredEvent: null,
```
markerikson, line 19: :question: Can we just make this a `TimeStampedPoint`?  We can use this same mechanism with things like the React Timeline panel too, which doesn't use the `ReplayEvent` type, and all we actually care about here _is_ the `ExecutionPoint` + time anyway.

PR #8987, opened 2023-03-23
Use requestFocusRange and getFocusWindow APIs

bvaughn:
### Verification:
* Before: [db03084b-6ac5-47f4-9224-e295615efd73](https://app.replay.io/recording/troublesome-azure-test--db03084b-6ac5-47f4-9224-e295615efd73)
* After: [c20072fa-125f-477d-9a8a-b90553c46504](https://app.replay.io/recording/troublesome-azure-test-after--c20072fa-125f-477d-9a8a-b90553c46504)

### API integration
- [x] Update @replayio/protocol package to 0.41.0
- [x] Add `getFocusWindow` to `ReplayClientInterface`
- [x] Update `requestFocusRange` to return new applied window result

### Redux actions
- [x] Sync focus window with backend at start of session using `getFocusWindow`
- [x] Sync actual focus window with backend after requested changes

### Tests
- [x] Update protocol fixture data (if needed)
- [x] ~~Add new e2e test with long recording~~ I'm going to hold off on this until BAC-2964 lands and we can do the better version of the UX.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -31,6 +31,10 @@ import { ThreadFront } from "protocol/thread";
 import { PauseEventArgs } from "protocol/thread/thread";
 import { waitForTime } from "protocol/utils";
 import { pointsBoundingTimeCache } from "replay-next/src/suspense/ExecutionPointsCache";
+import {
+  isExecutionPointsGreaterThan,
+  isExecutionPointsLessThan,
+} from "replay-next/src/utils/time";
 import { ReplayClientInterface } from "shared/client/types";
 import { getFirstComment } from "ui/hooks/comments/comments";
 import { mayClearSelectedStep } from "ui/reducers/reporter";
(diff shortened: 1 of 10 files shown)


bvaughn:
As I make this change, I realize there's an odd UX edge case that arises:

1. The user clicks on a test
2. The frontend sets the current time to the start of the test and asks the backend to focus on the range of time that includes the test.
3. If the region is too large, the backend focuses on the last minute or so of it.
4. The frontend shrinks its selected focus range to that last minute.
5. The current execution point (the start of the test) is now outside of a loaded region.

I wonder if we need a way for the frontend to tell the backend "if you can't load all of this, load the first bit rather"?

src/ui/actions/timeline.ts
```diff
@@ -687,22 +692,38 @@ export function setFocusRegionBeginTime(
...
 
     if (sync) {
-      dispatch(syncFocusedRegion());
+      await dispatch(syncFocusedRegion());
       dispatch(updateFocusRegionParam());
     }
   };
 }
 
 export function syncFocusedRegion(): UIThunkAction {
```
bvaughn, line 701: I _think_ this method is the single place that we need to add this two-way sync logic.

markerikson (approved):

PR #8988, opened 2023-03-24
Menu icons no longer flicker because they're preloaded

jonbell-lot23:
image

### DIFF
--- a/src/ui/components/Header/UserOptions.tsx
+++ b/src/ui/components/Header/UserOptions.tsx
@@ -41,6 +41,19 @@ function UserOptions({ setModal, noBrowserItem }: UserOptionsProps) {
     setModal("settings");
   };
 
+  const preloadIcons = () => (
+    <div style={{ display: "none" }}>
+      <Icon filename="docs" className="bg-iconColor" />
+      <Icon filename="help" className="bg-iconColor" />
+      <Icon filename="settings" className="bg-iconColor" />
+      <Icon filename="replay-logo" className="bg-iconColor" />
+    </div>
+  );
+
+  {
+    preloadIcons();
+  }
+
   return (
     <div className="user-options text-blue-400">
       <Dropdown
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -157,8 +157,18 @@ export default function PrivacyDropdown({ recording }: { recording: Recording })
     );
   }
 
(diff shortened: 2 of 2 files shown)


src/ui/components/Header/UserOptions.tsx
```diff
@@ -41,6 +41,19 @@ function UserOptions({ setModal, noBrowserItem }: UserOptionsProps) {
...
+    <div style={{ display: "none" }}>
+      <Icon filename="docs" className="bg-iconColor" />
+      <Icon filename="help" className="bg-iconColor" />
+      <Icon filename="settings" className="bg-iconColor" />
+      <Icon filename="replay-logo" className="bg-iconColor" />
+    </div>
+  );
+
+  {
+    preloadIcons();
```
markerikson, line 54: :hand: For the record I don't think this line does anything :)

You're calling the function, which returns JSX...

but that JSX isn't being returned from the _component_.

So, React will never look at it or render those `<Icon>` components.

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -157,8 +157,18 @@ export default function PrivacyDropdown({ recording }: { recording: Recording })
...
+      <MaterialIcon>globe</MaterialIcon>
+      <MaterialIcon>group</MaterialIcon>
+      <MaterialIcon>lock</MaterialIcon>
+      <MaterialIcon>expand_more</MaterialIcon>
+    </div>
+  );
+
   return (
     <>
+      {preloadIcons()}
```
markerikson, line 171: 💬 This one, on the other hand, _will_ do something, because it's included in the JSX return structure.

:hand: That said, in both cases, `preloadIcons` doesn't need to be a _function_.  It could just be a direct assignment to a variable, ie, `const preloadIcons = <div>Stuff Here</div>`

jonbell-lot23: I did not know that, thank you!

jonbell-lot23:
Revising via this ticket: https://linear.app/replay/issue/DES-455/improve-the-preload-code

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
bvaughn: PRs like this should start going through normal PR review process. Mark, Holger, and I review each other's changes as well. It's just got to have a second set of eyes on stuff that impacts production code.

PR #8990, opened 2023-03-24
Simplify loading page

jonbell-lot23:
image

**Background**

I've been eyeing this page for a long time. The icons are mismatched with the rest of the icons in our external branding and our UI. The text wasn't the same size, and it doesn't quite work even if you do match them. The text is often too lengthy for our new super-speedy loading page. The two boxes compete for attention despite one being much larger. The gap between them was always awkward but I couldn't find a way to make it less awkward.

**Tradeoffs**

Removing the tips means we lose a place to talk about security, and security is important. That's a reasonable argument for why this PR shouldn't go live. On the other hand, our loading screen is quick enough that it's possible those messages are just adding to clutter and a feeling of complexity.

**Next steps**

I'm proposing this in the hopes that we agree this is a reasonable PR to push live. But I'll understand if it's too bold :)


src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -36,15 +36,14 @@ export function LoadingScreenTemplate({
...
-        <div className="flex flex-col items-center space-y-8">
+      <div className="relative flex w-96 flex-col items-center space-y-1 rounded-lg bg-loadingBoxes p-4 py-12 shadow-sm">
+        <div className="flex flex-col items-center space-y-2">
           <div className="h-32 w-32" onClick={changeHoverboardColor}>
             <Hoverboard color={hoverboardColor} />
           </div>
           {children}
         </div>
       </div>
-      {showTips ? <LoadingTips /> : null}
```
bvaughn, line 47: Can you also delete the no-longer-in-use `showTips` variable that appears 3 other places in this file?

jonbell-lot23:
Ran this by Jason and implemented the changes Brian called out, so it should be good to go!

PR #8998, opened 2023-03-27
Rewrote StreamingSourceCache to use createStreamingCache

bvaughn:
Resolves FE-1355

- [x] Rewrite source (contents) cache to use `createStreamingCache`
- [x] Update syntax parsing cache to use new source contents cache
- [x] Cleaned up a few outstanding "TODO" comments in our e2e test helpers now that this migration has been finished

### DIFF
--- a/packages/replay-next/src/suspense/SyntaxParsingCache.ts
+++ b/packages/replay-next/src/suspense/SyntaxParsingCache.ts
@@ -1,10 +1,14 @@
 import { ContentType, SourceId } from "@replayio/protocol";
-import { createStreamingCache } from "suspense";
+import { StreamingCacheLoadOptions, createStreamingCache } from "suspense";
 
 import { assert } from "protocol/utils";
+import {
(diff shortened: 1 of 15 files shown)


markerikson (approved):

packages/replay-next/src/suspense/SyntaxParsingCache.ts
```diff
@@ -134,20 +136,24 @@ function postWorkerMessage(message: PrepareRequest | ParseRequest) {
...
+
+    state.value = state.value.concat(data.parsedTokens);
+    state.progress = state.value.length / state.streaming.data.codeUnitCount;
+    state.loadOptions.update(state.value, state.progress, state.data);
+
+    if (state.streaming.complete) {
+      internalParserStates.delete(data.sourceId);
+
+      state.loadOptions.resolve();
+    }
```
bvaughn, line 157: Note to self: I want to re-think this and the timing of it before merging.

PR #9000, opened 2023-03-28
Search input sometimes erased unexpectedly

bvaughn:
Resolves FE-1322

- [x] Fix loose memoization in `useSourceSearch` that caused an effect in `SourceSearchContext` to run too many times.
- [x] Don't wait for source contents to stream in before updating search scope.
- [x] Fix issue where "default" focus id causes source search to reset after a source loads.

See comment trail on [8ff1203c-2503-4aab-bbb3-ff4cb8dbc8a2](https://app.replay.io/recording/fe-1322-search-clear-bug--8ff1203c-2503-4aab-bbb3-ff4cb8dbc8a2) for why these changes were made.

cc @Andarist

### DIFF
--- a/packages/replay-next/components/sources/hooks/useSearch.ts
+++ b/packages/replay-next/components/sources/hooks/useSearch.ts
@@ -271,10 +271,6 @@ export default function useSearch<Item, Result, QueryData = never>(
       dispatch({ type: "updateItems", items });
     }
 
-    if (queryChanged || queryDataChanged) {
-      dispatch({ type: "updateQuery", query: deferredQuery, queryData: deferredQueryData });
-    }
-
     const results = deferredQuery
       ? stableSearch(deferredQuery, items, deferredQueryData)
       : EMPTY_ARRAY;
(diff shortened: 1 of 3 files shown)


packages/replay-next/components/sources/hooks/useSearch.ts
```diff
@@ -271,10 +271,6 @@ export default function useSearch<Item, Result, QueryData = never>(
       dispatch({ type: "updateItems", items });
     }
 
-    if (queryChanged || queryDataChanged) {
-      dispatch({ type: "updateQuery", query: deferredQuery, queryData: deferredQueryData });
-    }
```
bvaughn, line 276: This was the core of the bug. We shouldn't erase "global" query state when we encounter an invalid scope. We should just re-compute results for that scope (which we're already doing in this block).

packages/replay-next/components/sources/hooks/useSourceSearch.ts
```diff
@@ -125,23 +125,40 @@ export default function useSourceSearch(
...
+    },
+    [dispatchSearch, modifiers, state.query]
+  );
+
+  const memoizedSetScope = useCallback(
+    (sourceId: SourceId | null, code: string) => {
+      setScope({ code, sourceId });
+    },
+    [setScope]
+  );
```
bvaughn, line 152: This is just an optimization. I noticed `SourceSearchContext` effect was firing too often– because the `setScope` function inside of `memoizedActions` was being re-created every time `state.query` changed, even though the only value it uses (`setScope`) is stable.

markerikson (approved):

Andarist:
A guy saying 'thank you so much'

PR #9002, opened 2023-03-28
Implement initial Cypress "Jump to Code" button for click/type steps

markerikson:
This PR:

- Fixes two leftover React key warnings I ran across
- Extracts a reusable `<JumpToCodeButton>` component
- Fixes an issue where function preview locations sometimes hadn't been updated with proper mapped locations
- Adds the actual "Jump" button to Cypress "click/type" test step rows

General appearance:

- No hover: 
  image

- Button hovered:  
  image

The non-hovered state does appear a bit squished horizontally compared to the "Events" panel atm, but I'm going to skip that for now.

The "find location" logic seems to work decently for clicks.  The one example I have of a Cypress+FF recording with a `"type"` command (keyboard text) doesn't seem to be working right because the recording only shows `keydown` events captured, not `keypress`, even though Cypress is trying to fire all standard keyboard events in sequence.  That feels like a runtime bug of some kind.  Unfortunately I don't yet have other Cypress+FF recordings to compare against right now.

So, my vote is to go ahead and merge this in, let Jon tweak the design of the button, and then focus on getting the Chromium Events PR in so that we can start testing those pieces together.

### DIFF
--- a/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
+++ b/src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
@@ -275,7 +275,7 @@ function TestSection({
       ) : null}
       {events.map(({ event: s, type, time }, i) =>
         type === "step" ? (
-          <>
+          <React.Fragment key={s.id}>
             <TestStepItem
               step={s}
               key={s.id}
@@ -289,7 +289,7 @@ function TestSection({
               autoSelect={s.id === autoSelectId}
             />
             {s.error ? <TestError error={s.error!} /> : null}
-          </>
+          </React.Fragment>
         ) : type === "network" ? (
           <NetworkEvent key={s.id} request={s} />
         ) : (
--- a/src/devtools/client/debugger/src/components/shared/ResultList.tsx
+++ b/src/devtools/client/debugger/src/components/shared/ResultList.tsx
@@ -42,7 +42,6 @@ export default class ResultList extends Component<ResultListProps> {
     const { selectItem, selected } = this.props;
     const props: React.ComponentPropsWithRef<"li"> = {
       onClick: (event: any) => selectItem(event, item),
-      key: `${item.id}${item.value}${index}`,
       ref: element => {
         // Keep a lookup table of items by index for scrolling
         if (element) {
@@ -61,7 +60,7 @@ export default class ResultList extends Component<ResultListProps> {
     };
 
     return (
-      <li {...props}>
+      <li key={`${item.id}${item.value}${index}`} {...props}>
         {item.icon && (
           <div className="icon">
             <AccessibleImage className={item.icon} />
--- a/src/ui/actions/event-listeners.ts
+++ b/src/ui/actions/event-listeners.ts
@@ -279,6 +279,8 @@ export const IGNORABLE_PARTIAL_SOURCE_URLS = [
   "webpack:///src/sandbox/",
   "webpack:///sandpack-core/",
   "webpack:////home/circleci/codesandbox-client",
+  // or Cypress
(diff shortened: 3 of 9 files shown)


bvaughn:
Hey uh, did you want this code reviewed?

markerikson:
@bvaughn : It was somewhere between "want a design review" and "want a code review" :)

If you've got feedback on this, yeah, comment and I can address in a follow-up.

src/ui/components/shared/JumpToCodeButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+  targetExecutionPoint: ExecutionPoint;
+}
+
+export function JumpToCodeButton({
+  status,
+  onClick,
+  currentExecutionPoint,
+  targetExecutionPoint,
+}: JumpToCodeButtonProps) {
+  const [isHovered, setIsHovered] = useState(false);
```
bvaughn, line 29: Since this variable only impacts styles, it could probably be done more cleanly using CSS `:hover` state (and avoid the additional re-render)

src/devtools/client/debugger/src/components/shared/ResultList.tsx
```diff
@@ -61,7 +60,7 @@ export default class ResultList extends Component<ResultListProps> {
     };
 
     return (
-      <li {...props}>
+      <li key={`${item.id}${item.value}${index}`} {...props}>
```
bvaughn, line 63: Curious why `item.id` isn't sufficiently unique for the key here?

src/ui/actions/event-listeners.ts
```diff
@@ -277,6 +279,8 @@ export const IGNORABLE_PARTIAL_SOURCE_URLS = [
   "webpack:///src/sandbox/",
   "webpack:///sandpack-core/",
   "webpack:////home/circleci/codesandbox-client",
+  // or Cypress
+  "__cypress/runner/",
```
bvaughn, line 283: Kind of new to this code– but the `"react-dom"` string above seems maybe a bit too loose without "/" around it. (Seems like it would also impact any packages/paths with "react-dom" in the name– not just the `react-dom` package itself?)

src/devtools/client/debugger/src/components/TestInfo/TestSteps.tsx
```diff
@@ -275,7 +275,7 @@ function TestSection({
       ) : null}
       {events.map(({ event: s, type, time }, i) =>
         type === "step" ? (
-          <>
+          <React.Fragment key={s.id}>
```
bvaughn, line 278: Thanks for fixing this.

src/ui/components/shared/JumpToCodeButton.tsx
markerikson: That hovering behavior has gone through a couple copy-paste cycles at this point. I _think_ it was there in the console message items where we got the general behavior from in the first place? Agreed it could probably be simplified.

src/devtools/client/debugger/src/components/shared/ResultList.tsx
markerikson: Not sure. That key construction was there previously, I think.

src/ui/actions/event-listeners.ts
markerikson: I'm okay with that. For now the goal is primarily to jump to user code.

src/devtools/client/debugger/src/components/shared/ResultList.tsx
bvaughn: Just looks unnecessarily complicated :)

PR #9007, opened 2023-03-29
Polishing testsuites

jonbell-lot23:
* Removed the circular progress loader
* The button only appears when a row is selected
* Absolute positioning so it doesn't push against other text

image


jonbell-lot23:
Because of timezone differences and the rough state of this marquee feature, I'm going to merge this while also asking for a review from @markerikson

jasonLaster (approved):

markerikson (approved):

PR #9009, opened 2023-03-29
Replace runAnalysis with findPoints/runEvaluation (FE-1349)

hbenl:


packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -81,59 +66,83 @@ export function createAnalysisCache<
...
-          eventHandlerEntryPoints: analysisParams.eventTypes?.map(eventType => ({ eventType })),
-          exceptionPoints: analysisParams.exceptions,
-          mapper: analysisParams.mapper,
-          effectful: false,
-          range: { begin, end },
-        },
-        {
-          onPoints: points => {
-            allPoints = allPoints.concat(points.map(point => transformPoint(point, ...params)));
+      for (const evaluationParams of allEvaluationParams) {
```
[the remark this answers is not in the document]
loganfsmyth: I can add support for multiple locations to the API so we can avoid this

packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -81,59 +66,83 @@ export function createAnalysisCache<
...
-          onError: err => (error = err),
-        }
-      ).pointsFinished;
+              const values: Value[] = [];
+              if (result.exception) {
+                values.push(result.exception);
+              } else if (result.returned?.object) {
+                const length =
+                  (
+                    await objectPropertyCache.readAsync(
```
loganfsmyth, line 110: Does cache code like this trigger any protocol commands or does this just query the cache for the value?

packages/replay-next/src/suspense/ExceptionsCache.ts
```diff
@@ -1,74 +1,43 @@
...
-  exceptionsCache.pointsIntervalCache.read
-);
+export const exceptionValueCache = createCache<
+  [client: ReplayClientInterface, pauseId: PauseId],
+  Value | undefined
+>({
+  debugLabel: "ExceptionValueCache",
+  getKey: ([client, pauseId]) => pauseId,
+  load: async ([client, pauseId]) => {
+    const result = await client.getExceptionValue(pauseId);
```
loganfsmyth, line 39: Could you check how this behaves on a recording with a large number of exceptions? I realized that there might be some slowness with splitting this out if we're not careful about it.

packages/replay-next/src/suspense/AnalysisCache.ts
bvaughn: It potentially does both. If the value is already cached, it will return synchronously. Else it will query `Pause.getObjectProperty`.

markerikson: Looking at the code here, these _may_ have come back as part of the `results` handler on line 99?  Would need to run this / add some logging and see.

Pretty sure this _is_ necessary, though. The whole point of this file is to take whatever evaluation code was provided (such as a print statement content), eval that, and show it.  This chunk appears to be taking those results.

In other words, if a print statement was:

`"myFunction", state.value, counter`

We convert that into an array like `["myFunction", state.value, counter]`, and what's coming back here _is_ the array we need to show the results in the console.

But, we want to show each value individually in the console message, not wrapped in that artificial array.  So, we have to extract the values here, which means doing the array lookup to get them index-by-index.

(This is yet another example of how working with the protocol's design is frequently a pain.)

packages/shared/client/ReplayClient.ts
```diff
@@ -761,76 +804,43 @@ export class ReplayClient implements ReplayClientInterface {
...
+        expression: opts.expression,
+        frameIndex: opts.frameIndex,
+        fullReturnedPropertyPreview: opts.fullPropertyPreview,
+        pointLimits: opts.limits,
+        pointSelector: opts.selector,
+        runEvaluationId,
+      },
+      sessionId
+    );
+    removeEventListener("Session.runEvaluationResults", onResultsWrapper);
```
loganfsmyth, line 843: this should really be in a `finally` so the listener doesn't leak if there's an error

packages/replay-next/src/suspense/AnalysisCache.ts
loganfsmyth: If this code is going to live here, it should only be used in cases where `fullPropertyPreview: true` is passed, since it's designed to address that case as much as possible. That said, either way the client should still be getting back a `canOverflow` preview vs a `full` preview. Is the frontend required to have a `full`-level preview in order to render the object?

>   So, we have to extract the values here, which means doing the array lookup to get them index-by-index.

My concern is that they should already be in the cache, just potentially with a non `full` preview level. If they're not there at all, then something about that object preview has probably gone wrong.

What I'm saying for "user interaction" is primarily that from a perf standpoint:
* It'd be _way_ better to show the `canOverflow` preview and then have the `full` preview loading in the background, rather than `await`-ing the preview here.
* It'd be even better to just show the `canOverflow` preview, and only query the `full` preview if the user clicks on the object in some way. The `fullPropertyPreview` will do its best to expand things from `canOverflow` to `full` as part of `runEvaluation`, so if the frontend can just show the `full` when it's available and `canOverflow` overwise (until the user interacts), that would be ideal

markerikson: I feel like we're maybe talking past each other a bit here?

In general, the console message UI for print statements is wanting to do something akin to `values.map(value => <ObjectInspector value={value} />`, and then the inspector takes care of asking for more details.

I'm not sure of the specifics of overflow vs non-overflow currently.

We have other cases for analysis as well, but that's the main one where this matters, I think.

bvaughn: Sorry, I'm a few messages behind. Been in meetings.

> Hmm okay, I really don't think we should be doing this then. The frontend should do its absolute best to only query pause objects based on user interaction. Pause commands can be really slow, especially in this case where it's being used for the first time.

The frontend generally does this, only querying what it needs to display the data. Our caching supports 3 tiers of "preview" data– none, can-overflow, and full.

API requests made to `Pause.getObjectPreview` and `Pause.getObjectProperty` both cache data for a given Pause ID so we avoid re-requesting data unless (a) we need it to render something and (b) it hasn't already been cached as part of a prior request. (Indeed if we _forget_ this, the backend won't re-send the data anyway– so this cache is really important).

I'm not familiar with the new Analysis API (or the changes in this PR really) but given the call to `cachePauseData ` a few lines above:
https://github.com/replayio/devtools/blob/561d3ec018367fdfb16e12cde58875bbc38182ba/packages/replay-next/src/suspense/AnalysisCache.ts#LL102-L102C67

...I suspect the call you're commenting on (to `await objectPropertyCache.readAsync`) would be a **cache hit** in most cases.

loganfsmyth: Yeah I think we probably are a bit. Re-reading this code again, I misunderstood one thing, which is that this is more focused on the `ProtocolValue` than it is on the `ObjectPreview`.

Given that, I think the only comment I have left is that if these values aren't already in the cache, I'd consider that very bad and an indication that something's gone wrong, so because this is using the async version of the function, I'm worried that it could mask that problem by instead issuing further pause commands.

bvaughn: If that's the case (I don't have much context on this cache instance) we could replace–
```js
await objectPropertyCache.readAsync(...)
```
with–
```js
 await objectPropertyCache.getValue(...)
```
The latter function will return a cached value _or thrown_ if it's a cache miss.

hbenl:
The PR now contains 3 workarounds in separate commits: #9018, #9019 and a workaround for the protocol fixture recorder's inability to record errors thrown by `ReplayClient` (which I want to fix properly).

packages/replay-next/src/suspense/ObjectPreviews.ts
```diff
@@ -39,6 +40,23 @@ export const objectPropertyCache: Cache<
   debugLabel: "objectPropertyCache",
   getKey: ([client, pauseId, objectId, propertyName]) => `${pauseId}:${objectId}:${propertyName}`,
   load: async ([client, pauseId, objectId, propertyName]) => {
+    const cachedObject = objectCache.getValueIfCached(
+      null as any,
+      pauseId,
+      objectId,
+      "canOverflow"
+    );
+    if (cachedObject?.preview) {
```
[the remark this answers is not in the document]
hbenl: We may have a cached preview for the object whose property we want. In that case we can just return the cached value for that property.
This becomes necessary when we use `Session.runEvaluation` with an expression that returns an array: we want to get the values of that array from the array's preview. Without this change we'd have to send a request to the backend for every value that we want to extract from an array.

packages/replay-next/src/suspense/AnalysisCache.ts
hbenl: > Looking at the code here, these may have come back as part of the results handler on line 99?

Exactly.

> If this code is going to live here, it should only be used in cases where fullPropertyPreview: true is passed

That's exactly what's happening here (only in this case it's the `fullReturnedPropertyPreview` param for `Session.runEvaluation`). In my tests, all necessary previews were already cached, so this didn't send a single backend request.

packages/replay-next/src/suspense/ExceptionsCache.ts
hbenl: Looking good: https://app.replay.io/recording/exceptions--411587ca-3524-4653-8af3-10ac1aab14a1

hbenl: See here: https://app.replay.io/recording/exceptions--acf98816-1229-409a-8865-a0f7f8df624a
It's a lot slower than with the old Analysis API but I'd say our frontend handles that slowness rather well.

packages/replay-next/src/suspense/AnalysisCache.ts
bvaughn: >> If this code is going to live here, it should only be used in cases where fullPropertyPreview: true is passed
>
> That's exactly what's happening here

We could ensure this remains the case going forward by replacing `readAsync` with `getValue`.

Is there a reason not to do this, @hbenl ?

hbenl: > Is there a reason not to do this

I think we don't have a guarantee that the backend will include the necessary object previews in the events triggered by `Session.runEvaluation` and if it decides to not send an object preview for whatever reason, I'd prefer the frontend to request that rather than failing to show the message or showing an incomplete message.

packages/replay-next/src/suspense/ObjectPreviews.ts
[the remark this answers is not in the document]
hbenl: > I don't think we need this check, do we?

I guess you mean the `cachedObject?.preview` check? That is indeed not necessary but the `preview` property is marked as optional and Typescript doesn't know that in this case the object must have a `preview`. So I could use this check or `preview!`, I prefer to use a check.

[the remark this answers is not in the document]
bvaughn: I think we should probably use `assert(cachedObject?.preview)` here TBH. If it's _missing_ then something has gone pretty wrong.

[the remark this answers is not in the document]
hbenl: > It makes it clearer to a casual reader what the expected behavior is (a cache hit)

Good point, I'll use the `assert`.

markerikson:
Per my ramblings in Discord earlier today, I think there's a couple meaningful changes that are needed to this PR:

- We shouldn't be throwing an error if the pagination indicates there's more points available
- The rewritten `eventsCache`, which does full `runEvaluation` calls internally _and_ evaluates `"[...arguments]"` for each event handler hit, shouldn't be getting used for the `nextInteractionCache` logic in `Event.tsx`.  Colin reported that the backend is being taxed pretty hard as I was playing around with this today, and that _could_ be one of the issues.  That `nextInteractionCache` logic really only cares about hit points (I think).  What we actually need is an interval cache that does `Session.findPoints` with the event type info, but I haven't wrapped my mind around how to do that yet.

hbenl:
Copying my replies from Discord:

> We shouldn't be throwing an error if the pagination indicates there's more points available

The old API would throw this `CommandError` if there were too many points, the new API returns a `PointPage` with `hasNext` being `true`. So checking for this and throwing a `CommandError` is there to preserve the old behavior. The point of this PR was to switch to the new API without changing the behavior.

> The rewritten eventsCache, which does full runEvaluation calls internally and evaluates "[...arguments]" for each event handler hit, shouldn't be getting used for the nextInteractionCache logic in Event.tsx.

The old code would trigger `Analysis.runAnalysis` (that's why you have to pass in a mapper function), the only difference is that the new code will use `Session.runEvaluation` instead and will cache the results. So the performance should be the same or even better (due to caching), if it's worse then that's probably a bug in `Session.runEvaluation` that we should look into.

> What we actually need is an interval cache that does Session.findPoints with the event type info, but I haven't wrapped my mind around how to do that yet.

I'll do that.

hbenl:
The only changed screenshots are a result of [BAC-3017](https://linear.app/replay/issue/BAC-3017/internal-error-from-sessionfindpoints#comment-06858d68). Once that has been fixed I'll update the relevant fixtures and screenshots and remove the workaround for BAC-3017, I've created [FE-1393](https://linear.app/replay/issue/FE-1393/remove-workaround-for-bac-3017) for that.

packages/shared/client/ReplayClient.ts
```diff
@@ -332,6 +339,58 @@ export class ReplayClient implements ReplayClientInterface {
...
+    }
+
+    addEventListener("Session.findPointsResults", onPoints);
+
+    let result;
+    try {
+      result = await client.Session.findPoints(
+        { pointSelector, pointLimits, findPointsId },
+        sessionId
+      );
```
loganfsmyth, line 373: I'd recommend changing this to be
```
try {
  const maxCount = pointLimits?.maxCount ?? MAX_POINTS_TO_FIND;
  let begin = pointLimits?.begin ?? undefined;
  const end = pointLimits?.end ?? undefined;
  while (true) {
    const { pointPage } = await client.Session.findPoints(
      {
        findPointsId,
        pointSelector,
        pointLimits: {
          begin,
          end,
          // This should always be non-negative so but just adding for clarity.
          maxCount: Math.max(0, maxCount - points.length),
        },
      },
      sessionId
    );

    if (
      // Got to the end of the focus winow so definitely done.
      !pointPage.hasNext ||
      // We reached the end endpoint on the query so we're done.
      (end && compareNumericStrings(pointPage.end, end) >= 0) ||
    ) {
      break;
    }

    // If other parts of the codebase changed the focus window out from under 
    // this code this could happen, and if this isn't handled this code would be
    // at risk of being an infinite loop so it should be handled somehow.
    if (pointPage.begin === pointPage.end) {
      throw commandError("Too many points", ProtocolError.FocusWindowChange);
    }

    // If you hit the max count but didn't bail out for any of the cases
    // above then throw too many points.
    if (points.length >= maxCount) {
      throw commandError("Too many points", ProtocolError.TooManyPoints);
    }

    // Pick up querying after this page of results.
    begin = pointPage.end;
  }
} finally {
```
with the same type of thing for `runEvaluation`.

This has the great benefit that it separates the max results per page that the backend allows from the max count that the frontend wants to query for.

packages/shared/client/ReplayClient.ts
```diff
@@ -332,6 +339,58 @@ export class ReplayClient implements ReplayClientInterface {
...
+    } catch (error) {
+      //TODO remove this workaround when BAC-3017 is fixed
+      if (isCommandError(error, ProtocolError.InternalError)) {
```
loganfsmyth, line 376: You should be good to erase this now.

PR #9013, opened 2023-03-30
Timeline contex menu

bvaughn:
### [Loom overview video](https://www.loom.com/share/7ca6dd8732034440be16d21ef9179f00)

- [x] `ContextMenu` component updated to re-position itself to ensure menu stays within the viewport. (This was necessary to support a context menu at the bottom of the page, such as the Timeline.)
- [x] Shared `useContextMenu` hook supports a new optional parameter– `onContextMenuEvent`. (This function can be used to capture additional metadata, e.g. the relative position of the event within the timeline– which can be used to calculate the time that was clicked.)
- [x] Fixed an initialization bug where the app wasn't calling `Session.getFocusWindow` if there was a focus region defined. (If the region happened to be too large, we could have been out of sync initially.)
- [x] Disabled `pointer-events` for the Timeline hovered time tooltip (so that it didn't interfere with right-clicks).
- [x] Shared `useModalDismissSignal` hook stops event propagation. (This lets multiple right clicks, aka "contextmenu" events, to adjust the context menu position with only a single right click rather than two.)
- [x] Added new Timeline context menu via new `useTimelineContextMenu` hook.

Resolves FE-1365 and FE-1366.


packages/replay-next/components/context-menu/ContextMenu.tsx
```diff
@@ -1,10 +1,13 @@
-import { MouseEvent, ReactNode, useRef } from "react";
+import { MouseEvent, ReactNode, useLayoutEffect, useRef, useState } from "react";
 import { createPortal } from "react-dom";
 
 import useModalDismissSignal from "replay-next/src/hooks/useModalDismissSignal";
 
 import styles from "./ContextMenu.module.css";
 
+// TODO Should we be using clientX/clientY instead?
+// I guess it doesn't matter since Replay doesn't scroll.
```
bvaughn, line 9: I think this is out of scope of this PR but I wanted to leave it as a thought for later.

src/ui/actions/session.ts
```diff
@@ -337,8 +337,12 @@ export function createSocket(
...
       dispatch(actions.setLoadingFinished(true));
 
-      if (!focusRegion) {
-        const focusWindow = await replayClient.getFocusWindow();
+      const focusWindow = await replayClient.getFocusWindow();
+      if (
+        !focusRegion ||
+        focusRegion.begin.point !== focusWindow.begin.point ||
+        focusRegion.end.point !== focusWindow.end.point
+      ) {
```
bvaughn, line 345: This fixes FE-1366. I spotted it while testing the other changes in this PR.

src/ui/components/Timeline/Tooltip.css
```diff
@@ -12,6 +12,7 @@
   border-radius: 4px;
   border: 1px solid var(--theme-toggle-bgcolor);
   transform: translateX(-50%);
+  pointer-events: none;
```
bvaughn, line 15: This prevents the little time hover tooltip from blocking "contextmenu" events

PR #9016, opened 2023-03-30
Add "sort" and "filter" options to Comments

bvaughn:
### [Loom overview](https://www.loom.com/share/7ac637bed3304cfcb94c9ef31e4fd949)

Screen Shot 2023-03-31 at 3 17 46 PM

cc @jonbell-lot23 

---

**Update** Jon and I chatted on Discord and this basic UI treatment has his blessing– although I did merge #9020 into this PR and add a "hide" vs "dim" option to the menu, which Jon may have thoughts about.


src/ui/components/Comments/useUserCommentPreferences.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type Filter = "current-user" | null;
+type SortBy = "created-at" | "recording-time";
+
+export default function useUserCommentPreferences() {
+  const [filter, setFilter] = useLocalStorage<Filter>("Replay:CommentPreferences:Filter", null);
+  const [sortBy, setSortBy] = useLocalStorage<SortBy>(
+    "Replay:CommentPreferences:sortBy",
+    "recording-time"
+  );
```
bvaughn, line 13: Here's a question: Is this a _per recording_ preference or a global preference? My initial inclination was global, but I'm not 100% sure.

markerikson: Hmm. My own instinct says it makes more sense per-recording, but I'd be okay doing it globally as a starting point.

src/ui/components/Comments/CommentCardsList.tsx
```diff
@@ -11,57 +14,80 @@ import useAuth0 from "ui/utils/useAuth0";
...
   const recording = hooks.useGetRecording(recordingId);
   const auth = useAuth0();
 
+  const { filter, sortBy } = useUserCommentPreferences();
+
   const { isAuthenticated } = useAuth0();
 
   const displayedComments = useMemo(() => {
-    const clonedComments: Comment[] = [...comments];
+    const filteredComments =
```
markerikson, line 29: :speech_balloon: my one style nit here is that I'm _really_ not a fan of nested ternaries :)

markerikson (approved):

src/ui/components/Comments/CommentCardsList.tsx
bvaughn: Yeah that's fair. I'll rewrite this bit.

PR #9019, opened 2023-03-31
Source search: only select first match after query changes (FE-1376)

hbenl:
@bvaughn This is the second fix/workaround I mentioned in yesterday's standup.
It does fix the test but you had some reservations whether this fix is actually correct in all situations.

### DIFF
--- a/packages/replay-next/components/sources/hooks/useSearch.ts
+++ b/packages/replay-next/components/sources/hooks/useSearch.ts
@@ -285,7 +285,7 @@ export default function useSearch<Item, Result, QueryData = never>(
           index = results.indexOf(prevItem);
         }
 
-        if (index < 0) {
+        if (index < 0 && queryChanged) {
           index = findInitialIndex(results);
         }
       }


packages/replay-next/components/sources/hooks/useSearch.ts
```diff
@@ -285,7 +285,7 @@ export default function useSearch<Item, Result, QueryData = never>(
           index = results.indexOf(prevItem);
         }
 
-        if (index < 0) {
+        if (index < 0 && queryChanged) {
```
bvaughn, line 288: I left a few comments on [3c0263aa-7f88-4855-b1fe-30758a912528](https://app.replay.io/recording/replay-of-localhost3000--3c0263aa-7f88-4855-b1fe-30758a912528) about where the bug is happening.

Ideally I think this PR should probably also add a new unit test to cover the gap here.

I think I see what you're going for with this change– but in its current form, it would break other scenarios. (For example, if you changed a query modifier from e.g. case sensitive to case insensitive– then this would block us from finding a potential match).

_Maybe_ this would work–
```
if (index < 0 && (queryChanged || queryDataChanged)) {
```
–though I'd feel better about the change if we had some new tests that accompanied it.

I'm the most familiar with this code so if you don't want to deal with that– I can pick it up. I think we need a few new tests to accompany this change though.

hbenl: > if you don't want to deal with that– I can pick it up.

Thanks, I'd appreciate that.

bvaughn:
Closing in favor of #9041

PR #9022, opened 2023-03-31
Add global hooks to test tree

ryanjduffy:
* Adds `beforeAll` / `afterAll` hook support to the cypress panel

### DIFF
--- a/packages/shared/graphql/types.ts
+++ b/packages/shared/graphql/types.ts
@@ -242,12 +242,19 @@ export type TestMetadata = {
   result: TestResult;
   title: string;
   version: number;
+  hooks?: HookItem[];
   tests?: TestItem[];
   run?: { id: string; title?: string };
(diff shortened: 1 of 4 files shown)


ryanjduffy:
The basic structure is wired up.
image

Next is supporting selecting a global hook instead of a test step. The logic should be similar but there are probably some assumptions that need to be revisited.

bvaughn:
Hey! :wave: Maybe SUC and FE can have a chat this week about this component about the TestCase/TestInfo components and the short term plans for them.

I was thinking about maybe doing a refactoring pass at some of the code in `debugger/src/components/TestInfo` to help with some focus and perf issues. Maybe we can coordinate that work so we don't have big merge conflicts?

bvaughn:
I'm happy to pick this issue up and rebase it after #9056 lands.

bvaughn:
Going to roll this into #9056

PR #9043, opened 2023-04-03
More precise focus windows

bvaughn:
- [x] `FocusContext` supports two methods for refining focus: `update` (preferred) receives a `TimeStampedPointRange` and `updateForTimelineImprecise` (only used by the Timeline focuser UI) receives a range of times (which get mapped to execution points using `Session.getPointNearTime` or `Session.getPointsBoundingTime`)
  * Also updated the Log Point panel and Console context menu to both use execution points rather than times.
- [ ] ~~Update Redux actions (`updateFocusRegion`, `syncFocusedRegion`) to require execution points.~~
  * [`Session.requestFocusRange`](https://static.replay.io/protocol/tot/Session/#method-requestFocusRange) only accepts a time range, so I think this change maybe isn't worth making for the time being.

### DIFF
--- a/packages/replay-next/src/contexts/FocusContext.tsx
+++ b/packages/replay-next/src/contexts/FocusContext.tsx
@@ -18,7 +18,7 @@ import {
   imperativelyGetClosestPointForTime,
   preCacheExecutionPointForTime,
 } from "../suspense/ExecutionPointsCache";
-import { Range } from "../types";
+import { TimeRange } from "../types";
 import { SessionContext } from "./SessionContext";
 
 const FOCUS_DEBOUNCE_DURATION = 250;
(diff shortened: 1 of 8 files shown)


packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -94,33 +100,43 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
...
-    [client, debouncedSetDeferredRange]
+    [debouncedSetDeferredRange]
+  );
+
+  const updateForTimelineImprecise = useCallback(
+    async (value: TimeRange | null, debounce: boolean) => {
+      if (value) {
+        const [timeBegin, timeEnd] = value;
+
+        const pointBegin = await imperativelyGetClosestPointForTime(client, timeBegin);
```
markerikson, line 125: :question: Tiny nit/suggestion: can we `Promise.all()` these?

markerikson (approved):

packages/replay-next/src/contexts/FocusContext.tsx
bvaughn: Probably not. This portion was just a copy-paste but I don't mind making that change.

PR #9046, opened 2023-04-04
Video player context menu 

bvaughn:
- [x] Add context menu to video player with options to inspect an element and add a comment
- [x] Re-use node picker highlighting to show which element the context menu "applies" to
- [x] Don't show "add comment" button to unauthenticated users (we were doing this in a few places before)

https://user-images.githubusercontent.com/29597/229807080-6522dfcd-4790-40d6-a1e0-df0619e19a4f.mp4

Resolves FE-1347.


bvaughn:
Closing in favor of #9047

PR #9048, opened 2023-04-04
Chromium Events, take 2

markerikson:
This PR is a re-implementation of #8578 .  I started it initially to see if redoing it from scratch would get the snapshot tests working right, and ended up carrying through with it because the Git history ends up being simpler.

The actual changes should be the same.


bvaughn (approved):
I wish the complexity of the differences between Gecko and Chromium events weren't on the client like this but 🤷🏼

PR #9052, opened 2023-04-05
Enable Chromium "Jump to Code" support and improve Cypress handling

markerikson:
This PR:

- Updates the "find event hit points" cache to accept multiple event types for a single `findPoints` call
- Updates the "Jump to Code" event processing logic to query for _all_ known Chromium "raw event" types in a single `findPoints` call. (This is necessary because a single `click` event could be any one of the "raw" types like `"click,DIV"`, `"click,BUTTON"`, etc). This enables the existing "Jump to Code" logic to work for Chromium recordings.
- Improves the analysis mapper that finds possible React props by matching against multiple DOM event class types (such as `["InputEvent", "KeyboardEvent"]` for a keyboard operation) in order to find the right event object argument
- Updates the "find the first probable userland handled event" logic to ignore any handlers in `cypress/runner/` files specifically. These were consistently turning up as the first hit, and should be ignored.

To try this out, see these Cypress recordings.  Look in the Cypress panel for `type` and `click` commands.  Click on those test step items, then click the blue "Jump to Code" button.

- Cypress RealWorld:
  - FF: N/A atm
  - Chromium:
	  - https://devtools-git-feature-fe-1363-cypress-jump-location-recordreplay.vercel.app/recording/cypresstestsuibankaccountsspects--5109edd2-36fb-464d-9b43-2d142b869028
	  - https://devtools-git-feature-fe-1363-cypress-jump-location-recordreplay.vercel.app/recording/cypresstestsuiauthspects--d6f2df81-b4fa-4c0c-8e13-ec924f62f31a
- React-Hook-Form:
  - FF:
    - https://devtools-git-feature-fe-1363-cypress-jump-location-recordreplay.vercel.app/recording/cypresse2eusewatchusefieldarraynestedcyts--6692d05a-2d26-48e4-87b9-548e538e3ea5
  - Chromium:
    - https://devtools-git-feature-fe-1363-cypress-jump-location-recordreplay.vercel.app/recording/cypresse2eusewatchusefieldarraynestedcyts--ef618b0c-4084-4203-a44c-bc99e9e09752


packages/replay-next/src/suspense/EventsCache.ts
```diff
@@ -51,33 +52,40 @@ export const eventCountsCache: Cache<
 
 export const eventPointsCache = createIntervalCache<
   ExecutionPoint,
-  [client: ReplayClientInterface, eventType: EventHandlerType],
+  [client: ReplayClientInterface, eventTypes: EventHandlerType[]],
   PointDescription
 >({
   debugLabel: "EventPoints",
-  getKey: (client, eventType) => eventType,
+  getKey: (client, eventTypes) => eventTypes.join(),
```
bvaughn, line 59: nit: Should probably specify a separator string.

I don't think it's likely (maybe not even possible in this specific case) but without that, you could have a key conflict (e.g. "a", "bc" and "a", "b", "c would result in the same key).

src/ui/actions/event-listeners.ts
```diff
@@ -510,14 +517,42 @@ function createReactEventMapper(eventType: SEARCHABLE_EVENT_TYPES) {
       }
 
       return res;
+    } else {
+      // TODO Semi-debug logging of the arguments. Hopefully can be removed soon.
```
bvaughn, line 521: When is "soon"?  Under what conditions are we okay to remove this?

If we intend to remove this, would be nice to file a Linear issue for it. (Otherwise code like this tends to get overlooked).

src/ui/components/Events/Event.tsx
```diff
@@ -41,30 +46,45 @@ export interface PointWithEventType extends TimeStampedPoint {
...
-  Record<RecordingTarget, Record<SEARCHABLE_EVENT_TYPES, string>>
+  Record<RecordingTarget, Record<SEARCHABLE_EVENT_TYPES, EventListenerEntry>>
 > = {
   gecko: {
-    mousedown: "event.mouse.click",
-    keypress: "event.keyboard.keypress",
+    mousedown: { categoryKey: "Mouse", eventType: "event.mouse.click" },
+    keypress: { categoryKey: "Keyboard", eventType: "event.keyboard.keypress" },
   },
   // TODO [FE-1178] Fill in Chromium event types here?
```
bvaughn, line 63: Is this comment still revelant?

bvaughn (approved):
I don't have much context here 🤷🏼

bvaughn:
Looks like this PR broke Playwright test:
> tests/console.ts:297:1 › should log events in the console

packages/replay-next/src/suspense/EventsCache.ts
markerikson: For Firefox, this will be a single string like `["event.mouse.click"]`.  For Chromium, it should be strings like `["click,DIV","click,BUTTON"]`.  Also, `Array.join()` defaults to `","` as a separator string anyway.

markerikson:
Cleaned up the dead comment and the debug code, and updated the fixture data for that one test.

packages/replay-next/src/suspense/EventsCache.ts
bvaughn: TIL. Didn't realize it defaulted to "," as a separator.

PR #9063, opened 2023-04-12
Find jump locations for `button type=submit` in a `form onSubmit=`

markerikson:
This PR:

- Updates the "Jump to Code" heuristics to try to find cases where a `<button type="submit">` is clicked inside of a `<form>` with a React `onSubmit=` prop
- Checks the contents of function previews to see if they've been formatted correctly. Currently, Chromium does not include a `functionLocation` field in its previews, even though it should be (see RUN-1709 ), so we can't figure out where the function lives and jump to it.  (This will hopefully be fixed shortly by https://github.com/replayio/chromium/pull/541 )
- Adds callstack frame grouping for Cypress:

image


src/ui/actions/event-listeners.ts
```diff
@@ -66,13 +66,14 @@ export type NodeWithPreview = DeepRequired<
...
-    !!obj && "functionName" in obj && "functionLocation" in obj && "functionParameterNames" in obj
-  );
+  return !!obj && "functionName" in obj && "functionLocation" in obj;
 };
 
 export const isFunctionWithPreview = (obj: ProtocolObject): obj is FunctionWithPreview => {
-  return obj.className === "Function" && isFunctionPreview(obj.preview);
+  return (
+    (obj.className === "Function" || obj.className === "AsyncFunction") &&
+    isFunctionPreview(obj.preview)
```
bvaughn, line 75: I'm curious why we're checking this way (`"key" in obj`) rather than (`obj.key`)? Seems like this way would give false positives for empty strings or undefined/null values. Do we want that?

src/ui/actions/event-listeners.ts
```diff
@@ -467,6 +478,16 @@ function createReactEventMapper(eventType: SEARCHABLE_EVENT_TYPES) {
...
         searchedNodes.push(searchedNode);
 
+        const currentNodeName = currentNode.nodeName.toLowerCase();
+
+        if (
+          injectedValues.eventType === "mousedown" &&
+          currentNodeName === "button" &&
+          (currentNode as HTMLButtonElement).type === "submit"
+        ) {
+          clickWasInsideSubmitButton = true;
```
bvaughn, line 488: nit: We could just set `res.clickWasInsideSubmitButton` directly here and avoid needing the extra line below:
```js
res.clickWasInsideSubmitButton = clickWasInsideSubmitButton;
```

src/ui/actions/event-listeners.ts
markerikson: The general intent is to check if the fields exist at all, and at that point we can assume that what's _in_ the fields matches the TS types from the backend.

PR #9066, opened 2023-04-13
Minimal repro for unnecessary renders caused by a suspense cache

hbenl:
A cache created with the `suspense` package causes the RDT panel to rerender unnecessarily (which resets RDT, causing the UI issue that made me notice this in the first place).
STR:
- open http://localhost:8080/recording/replay-of-localhost8080--19fe1893-b9ea-41a5-a90a-90a209064c0b?point=5192296858539304422732884716028029&time=2720 (and ignore the errors from the backend - see [BAC-3105](https://linear.app/replay/issue/BAC-3105/assertionfailure-assertion-failed-missing-mapped-location)) or https://devtools-git-hbenl-rdt-render-repro-recordreplay.vercel.app/recording/replay-of-localhost8080--19fe1893-b9ea-41a5-a90a-90a209064c0b?point=5192296858539304422732884716028029&time=2720
- open the RDT panel
- select the "h" component
- expand the state hook
- expand the object in the state hook

-> RDT is reset and forgets the selection

### DIFF
--- a/src/ui/components/SecondaryToolbox/index.tsx
+++ b/src/ui/components/SecondaryToolbox/index.tsx
@@ -11,6 +11,7 @@ import React, {
   useState,
 } from "react";
 import { ImperativePanelHandle } from "react-resizable-panels";
+import { createSingleEntryCache } from "suspense";
 
 import { EditorPane } from "devtools/client/debugger/src/components/Editor/EditorPane";
 import { RecordingCapabilities } from "protocol/thread/thread";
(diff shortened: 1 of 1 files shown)


bvaughn:
> which resets RDT, causing the UI issue that made me notice this in the first place

I haven't had a chance to look into whatever this and PR #9067 is showing yet– but the fact that a re-render blows away component state is the bigger problem here, I think.

PR #9067, opened 2023-04-13
Another repro for unnecessary renders caused by a suspense cache

hbenl:


bvaughn:
Going to dig into this.

bvaughn:
We know the cause here and a "fix" so I think it's okay to close this and the other PR now 😁  Thanks for the helpful repro cases and recordings!

PR #9072, opened 2023-04-15
Enable repainting and network requests

jasonLaster:
cc @jcmorrow

### DIFF
--- a/packages/replay-next/src/suspense/BuildIdCache.ts
+++ b/packages/replay-next/src/suspense/BuildIdCache.ts
@@ -90,8 +90,8 @@ function getRecordingCapabilities(buildId: string) {
         supportsEagerEvaluation: false,
         supportsElementsInspector: false,
         supportsEventTypes: false,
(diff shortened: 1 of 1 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -228,8 +228,8 @@ class _ThreadFront {
           supportsEagerEvaluation: false,
           supportsElementsInspector: true,
           supportsEventTypes: true,
-          supportsNetworkRequests: false,
-          supportsRepaintingGraphics: features.chromiumRepaints ?? false,
+          supportsNetworkRequests: true,
+          supportsRepaintingGraphics: features.chromiumRepaints ?? true,
```
bvaughn, line 232: Line 231 looks okay. I'm not sure about 232 though. Do we mean to turn repaints on for Chromium recordings by default?

jasonLaster: Yep. that's the goal

PR #9077, opened 2023-04-18
Improve stability of some e2e tests

hbenl:
- the `rewindToLine()` helper could be called without a `lineNumber`, in that case it would click the rewind button without waiting for the new pause, so the next test step could be executed at the old pause. I've made the `lineNumber` mandatory and changed `rewindToLine()` calls without a `lineNumber` to `rewind()` which does wait for the new pause.
- increased the timeout in the `waitForPaused()` helper

### DIFF
--- a/packages/e2e-tests/helpers/pause-information-panel.ts
+++ b/packages/e2e-tests/helpers/pause-information-panel.ts
@@ -212,9 +212,9 @@ export async function rewind(page: Page) {
 export async function rewindToLine(
   page: Page,
   options: {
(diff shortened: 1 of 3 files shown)


markerikson (approved):

hbenl:
> the only two places we weren't passing that number have been changed to use rewind instead (which doesn't take a number)

Yes, and that should fix the flakiness because `rewind()` will call `waitForPaused()` whereas `rewindToLine()`, when called without a `lineNumber`, would not.

PR #9078, opened 2023-04-19
Improve Replay Passport edgecases + bugs

jonbell-lot23:
In our research I discovered that some people are clicking over to Viewer and getting confused because the Tour/Passport is no longer applicable to what they're seeing. (For example, it says "click x" but x isn't there)

This changes the logic around to guide people back into the Tour if they go to the Viewer. This is also important if someone returns a day later and they're halfway through the Tour.

image

### DIFF
--- a/src/ui/components/Tour/Tour.tsx
+++ b/src/ui/components/Tour/Tour.tsx
@@ -78,7 +78,7 @@ const Tour: React.FC = () => {
                   {viewMode === "non-dev" ? (
                     <>
                       <div className={styles.h1}>Oh, hi again!</div>
-                      <p>To continue, please click the DevTools toggle at the top left.</p>
+                      <p>To continue, please click the DevTools toggle at the top right.</p>
                     </>
                   ) : (
                     <>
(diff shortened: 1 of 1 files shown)


bvaughn:
I just tested this on the deployment and it doesn't seem to work. The passport is suggesting I should "time travel in the console" but the Viewer tab is open.

Screen Shot 2023-04-19 at 9 19 02 AM

src/ui/components/Tour/Tour.tsx
```diff
@@ -60,125 +60,144 @@ const Tour: React.FC = () => {
...
-                        bottom: `300px`,
-                        transform: `rotate(14deg)`,
-                      }}
-                    />
-                  </>
-                )}
-              </div>
-            )}
+            <div className={styles.intro}>
+              {isNewUser ? (
```
bvaughn, line 64: IMO inline ternaries like this are kind of hard to read when they're multi-lined, especially when they're nested. I think it improves readability to break the subsections down into their own components, e.g.

```js
<div>
  {isNewUser ? <NewUserInstructions /> : <ReturningUserInstructions />}
</div>
```

This component isn't too large so this is more of a nit which you can ignore though. Just sharing a personal preference.

jonbell-lot23: Yeah, I was feeling that too. I'm going to merge this one and then move to this format in the next iteration, because things are gonna get more complex soon

Thank you!

bvaughn:
@jonbell-lot23 I didn't see a response to my comment above: https://github.com/replayio/devtools/pull/9078#issuecomment-1514723541

> I just tested this on the deployment and it doesn't seem to work. The passport is suggesting I should "time travel in the console" but the Viewer tab is open.

Seems like this was maybe merged prematurely?

jonbell-lot23:
> @jonbell-lot23 I didn't see a response to my comment above: [#9078 (comment)](https://github.com/replayio/devtools/pull/9078#issuecomment-1514723541)
> 
> > I just tested this on the deployment and it doesn't seem to work. The passport is suggesting I should "time travel in the console" but the Viewer tab is open.
> 
> Seems like this was maybe merged prematurely?

Sorry, I didn't see this note and did merge it prematurely. Reverting now

PR #9080, opened 2023-04-19
Remove old BACKEND_CI timeouts

jcmorrow:
This was for when we were running in standalone and creating new recordings for every test. We don't do that anymore, so we no longer need to be as aggressive about waiting around (and making the tests run *forever*, which they do).

Fixes https://linear.app/replay/issue/BAC-3148/turn-down-backend-ci-timeouts


jazzdan (approved):

jcmorrow:
It looks like there's a few failures, but that seems common here these days as well as on the backend ☹️ . I don't *think* I could have directly caused any of those.

PR #9083, opened 2023-04-19
Refactor Passport code

jonbell-lot23:
* Making things easier to read by moving things to components
* Continuing to hunt down edge cases in the passport flow

### DIFF
--- a/src/ui/components/Tour/Tour.tsx
+++ b/src/ui/components/Tour/Tour.tsx
@@ -69,7 +69,7 @@ const Tour: React.FC = () => {
 
   const { dismissTourNag } = useNagDismissal();
 
-  const Intro: React.FC<IntroProps> = ({ typeOfReplay }) => (
+  const intro = (
     <div className={styles.intro}>
       <p className={styles.h1}>Hello and welcome!</p>
       {typeOfReplay === "events" ? (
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):
Approving so you can move this ahead but I do suggest one change:

We should move all of these component definitions out of `Tour()` and make them peers of that function instead. When they're inside, it'll cause React to create new DOM nodes every render because they're a completely new component (redefined for each render of Tour).

Editorially, that's a lot of components and probably not adding a lot of value to make them all separate. There are a handful that take no props and aren't reused anywhere so I'd generally leave them as part of Tour. Not a blocking issue though so take it or leave it.

bvaughn:
Ryan's right about inline functions.

```js
const Intro: React.FC<IntroProps> = ({ typeOfReplay }) => (
  <div className={styles.intro}>
    <p className={styles.h1}>Hello and welcome!</p>
    {typeOfReplay === "events" ? (
      <>
        <p>
          Replay is the first time-travel enabled DevTools. Let's get started!
        </p>
        <p>
          <em>Click on DevTools in the top right.</em>
        </p>
      </>
    ) : null}
  </div>
);
```

Could be replaced with this:
```js
const intro = (
  <div className={styles.intro}>
    <p className={styles.h1}>Hello and welcome!</p>
    {typeOfReplay === "events" ? (
      <>
        <p>
          Replay is the first time-travel enabled DevTools. Let's get started!
        </p>
        <p>
          <em>Click on DevTools in the top right.</em>
        </p>
      </>
    ) : null}
  </div>
);
```

There's no real reason for using inline components here. You can just declare an inline block. Then your ternary switches from this:
```js
{isNewUser ? <Intro typeOfReplay={typeOfReplay} /> : ...
```

To this:
```js
{isNewUser ? intro : ...
```

jonbell-lot23:
Thanks, gang! Will make changes

jonbell-lot23:
Ok, I updated most of the components to look like

```
intro
```

rather than <Intro>

I kept the <CompletedTour> one because it has props. And I kept the animated gifs because it read strangely to me for it to say

```
{showConsoleNavigate && showBreakpointAdd && showBreakpointEdit && timeTravelGif }
```

Rather than 

```
{showConsoleNavigate && showBreakpointAdd && showBreakpointEdit && <TimeTravelGif />}
```

Preparing to merge this soon, then on to other Passport improvements

PR #9085, opened 2023-04-20
Abort requests that failed with FocusWindowChange error (FE-1415)

hbenl:


packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -71,64 +72,79 @@ export function createAnalysisCache<
...
-          frameIndex: evaluationParams.frameIndex,
-          fullPropertyPreview: true,
-          limits: { begin, end },
-        },
-        async results => {
-          for (const result of results) {
-            setPointAndTimeForPauseId(result.pauseId, result.point);
-            cachePauseData(client, result.pauseId, result.data);
+      client
+        .runEvaluation(
```
bvaughn, line 89: Shouldn't we be _awaiting_ `client.runEvaluation` ?

For one, that method is async (and accepts an async callback) so not _awaiting_ it would cause our interval cache `load` method to signal completion prematurely.

But also, if we aren't awaiting then re-throwing a potential `error` below won't do anything. (It will be a global/uncaught error.)

bvaughn (commented):
This technique of ignoring `ProtocolError.FocusWindowChange` errors and re-throwing everything else– since we're using it for all of our interval caches, I wonder if it would be worth making an extended cache type that does this automatically?

```ts
export function createReplayIntervalCache<Point extends SerializableToString, Params extends Array<any>, Value>(options: {
  comparePoints?: ComparisonFunction<Point>;
  debugLabel?: string;
  getKey?: (...params: Params) => string;
  getPointForValue: GetPointForValue<Point, Value>;
  load: (start: Point, end: Point, ...params: [...Params, IntervalCacheLoadOptions]) => PromiseLike<Value[]> | Value[];
}): IntervalCache<Point, Params, Value> {
  const cache = createIntervalCache({
    ...options,
    load: async(start, end, ...paramsWithCacheLoadOptions) => {
      try {
        return await options.load(start, end, ...paramsWithCacheLoadOptions);
      } catch (error) {
        if (isCommandError(error, ProtocolError.FocusWindowChange)) {
          const params = paramsWithCacheLoadOptions.slice(0, -1) as Params;
          cache.abort(...params);
        }
        throw error;
      }
    }
  });

  return cache;
}
```

packages/replay-next/src/suspense/AnalysisCache.ts
hbenl: > not awaiting it would cause our interval cache load method to signal completion prematurely.

The interval cache stores the points from `findPoints()`, so we `await` only that, i.e. we *want* this load method to return as soon as points have been received. `runEvaluation()` will fetch the results, which are not stored in the interval cache but in the externally managed results cache.

> if we aren't awaiting then re-throwing a potential error below won't do anything

Well, it should trigger the error overlay in dev and send the error to Sentry in prod.

bvaughn: >> if we aren't awaiting then re-throwing a potential error below won't do anything
>
> Well, it should trigger the error overlay in dev and send the error to Sentry in prod.

I think this kind of only prevents the worst case though (no logging at all). Code calling this cache would expect it to fail in a way that would trigger an error boundary (or `try/catch`).

Interval caches are expected to block. (This isn't a _streaming cache_.)

hbenl: Ah, but that also means that the request is already resolved in the `pointsIntervalCache` when `runEvaluation` throws, so `abort` won't work and I need to use `evict` instead.

hbenl: > Code calling this cache would expect it to fail in a way that would trigger an error boundary

I would not expect `pointsIntervalCache.read()` to throw if the points were fetched successfully. But `resultsCache.read()` should throw in this case, so we should do something like `points.forEach(point => resultsCache.cacheError(error, point, ...params);` in the catch block. The suspense package doesn't offer a method for caching errors yet though.

bvaughn: Ah, that makes sense. Give me a few minutes to think on that.

bvaughn: https://github.com/replayio/devtools/pull/9086

PR #9087, opened 2023-04-20
CommandBar: resume/rewind buttons

bvaughn:
Disable `CommandBar` resume/rewind buttons unless paused and loaded frames.

I don't know the historical reason for why these buttons are always enabled, but the underlying Redux actions don't seem to do anything meaningful unless frame steps are available:

https://github.com/replayio/devtools/blob/44ecbe1943c7dc55d44ed56eb608dbda29afb2ea/src/devtools/client/debugger/src/reducers/pause.ts#L356-L364

Also updated the pause-info test helper to await for `button.isEnabled()` before clicking.

cc @jcmorrow

### DIFF
--- a/packages/e2e-tests/helpers/pause-information-panel.ts
+++ b/packages/e2e-tests/helpers/pause-information-panel.ts
@@ -184,18 +184,26 @@ export async function resumeToLine(
   }
 }
 
+export async function clickCommandBarButton(page: Page, title: string): Promise<void> {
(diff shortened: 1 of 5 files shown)


bvaughn:
Looks like these e2e test failures might be legit for this branch. Running an arbitrary one (`stepping-02`) I see things are now ending up stuck while waiting for the "Rewind" button to become enabled. Need to reconsider this approach then.

markerikson (approved):

PR #9097, opened 2023-04-25
Use a stable ErrorBoundary key for loggables (FE-1377)

hbenl:
Using the loggable's index as ErrorBoundary key led to an infinite render loop when new loggables were added to the console (e.g. by setting a logpoint) and hence the index changed.


markerikson (approved):

bvaughn (commented):
Thanks for digging into this!

PR #9100, opened 2023-04-25
Simplify ExecutionPoint comparison function

bvaughn:
This looks overly complicated:
```js
const collator = new Intl.Collator("en", {
  numeric: true,
  sensitivity: "base",
});

function compareExecutionPoints(a: ExecutionPoint, b: ExecutionPoint): number {
  return collator.compare(a, b);
}
```

Given that `ExecutionPoint`s are `BigInt`s then it should suffice to just use `Number(BigInt(a) - BigInt(b))` for comparison purposes.

Also based on a micro benchmark, it seems like using `Intl.Collator` is _slower_ than using `Number` (which I don't think is surprising) – https://jsben.ch/1TSZh

### DIFF
--- a/packages/replay-next/src/utils/time.ts
+++ b/packages/replay-next/src/utils/time.ts
@@ -1,4 +1,4 @@
-import { ExecutionPoint, PointRange, TimeStampedPointRange } from "@replayio/protocol";
+import { ExecutionPoint, TimeStampedPointRange } from "@replayio/protocol";
 import differenceInCalendarDays from "date-fns/differenceInCalendarDays";
 import differenceInMinutes from "date-fns/differenceInMinutes";
 import differenceInMonths from "date-fns/differenceInMonths";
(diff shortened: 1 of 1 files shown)


markerikson:
FWIW that's actually pretty intentional.  I don't have benchmarks on hand atm, but I remember looking around a while back and concluding that a simple string compare was much faster than trying to parse to `BigInt` every time.

bvaughn:
> I remember looking around a while back and concluding that a simple string compare was much faster than trying to parse to BigInt every time.

If you can find/remember this– would love to see it. I'm using the `Number` comparison in other places too.

markerikson:
Huh. Well.  Tried sorting an array of 3 ExecutionPoints 1M times:

```ts
{
  bigIntTotal: 966.2008500024676,
  intlCollatorTotal: 2232.274211999029
}
```

Pretty surprised, actually.

(I know that `Intl.Collator` was a lot faster than `String.localeCompare`.  Maybe that's what I was thinking of?)

bvaughn:
> (I know that Intl.Collator was a lot faster than String.localeCompare. Maybe that's what I was thinking of?)

Yeah that makes more sense.

bvaughn:
Delta is bugging on this PR.

bvaughn:
I'm going to go ahead and land this, given that it both simplifies and (seems to) speeds up the code.

PR #9098, opened 2023-04-25
Stream Console messages when possible

bvaughn:
## [Loom overview video](https://www.loom.com/share/fdbec033ade0451ba7c7d88357dd62b3)

In this video, we compare production Console loading times to development loading times with the new streaming behavior introduced in this PR. In this case we see the following results:

| Feature | production | development |
|:---|:---:|:---:|
| Console visible | 0:28 | 0:02 |
| First message(s) | 0:28 | 0:06 |

Here's a shorter video that shows how long it takes to load console messages for a fresh controller:

https://user-images.githubusercontent.com/29597/234404161-58984e33-27f8-4134-af48-fac96a9a828e.mp4

And here's messages loading with a reused controller:

https://user-images.githubusercontent.com/29597/234404558-cc77e5a1-9dac-43eb-915c-f7308169829a.mp4

- [x] Add new Console messages Suspense cache that...
  - [x] Attempts to stream (`Console.findMessages`) messages when possible and
  - [x] Falls back to the slower, blocking API (`Console.findMessagesInRange`) if needed
- [x] Tests
  - [x] Update `MessagesCache.test` since that cache has changed so much (hopefully rebase on top of #9101)
  - [x] Update e2e tests
  - [x] ~~Update screenshot e2e test fixture data~~

PR #9101, opened 2023-04-25
Jest mock extendded

bvaughn:
When working on #9098, I noticed that the TypeScript for our mock `ReplayClientInterface` methods were all `any` – which isn't great, but also that mock was getting pretty unwieldy to maintain.

To address both of these issues, I added the [jest-mock-extended](https://www.npmjs.com/package/jest-mock-extended) package to generate this mock with proper types. Also found and fixed a few minor type errors along the way.

### DIFF
--- a/packages/replay-next/package.json
+++ b/packages/replay-next/package.json
@@ -16,6 +16,7 @@
     "design": "workspace:*",
     "escape-html": "^1.0.3",
     "idb": "^7.1.1",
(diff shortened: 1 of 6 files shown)


markerikson (approved):

bvaughn:
Whoops~ looks like I need to re-add a couple of default return values :)

PR #9105, opened 2023-04-26
Clamp times so they are within the recording (FE-1428)

hbenl:
I also changed `ReplayClient.getSessionEndpoint()` to use `this.getSessionIdThrows()` instead of expecting the `sessionId` as an argument.
When updating the protocol fixtures I ran into the error "The focus range was changed while the command was in progress" again. This time I think I found the proper fix: `FocusContext` (which is only used in the screenshot tests) used to call `requestFocusRange()` with the initial range, now it does so only when the range changes.


packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -81,7 +84,7 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
 
   // Refine the loaded ranges based on the focus window.
   useEffect(() => {
-    if (range === null) {
+    if (range === null || range === initialRange) {
```
bvaughn, line 87: I think this might be a bug waiting to happen if we:
1. Start with the initial range
2. Change it to something else
3. Change it back to the full range using the focuser UI

Wouldn't we end up _not_ calling `client.requestFocusRange` at step 3 even though we should?

I think what you're trying to do is avoid calling the API if the range hasn't changed from whatever it started as. I think we can do that using a ref though?

```js
const prevRangeRef = useRef(initialRange);

useEffect(() => {
  if (range !== prevRangeRef.current) {
    prevRangeRef.current = range;
```

The ref would be initialized the first time the component renders (so we wouldn't request initially) and then we'd request each time it changed afterward.

packages/replay-next/src/suspense/ExecutionPointsCache.ts
```diff
@@ -293,12 +294,24 @@ export function preCacheExecutionPointForTime(timeStampedPoint: TimeStampedPoint
...
   [replayClient: ReplayClientInterface, time: number],
   PointsBoundingTime
 > = createCache({
   config: { immutable: true },
   debugLabel: "PointsBoundingTime",
   getKey: ([client, time]) => `${time}`,
-  load: async ([client, time]) => client.getPointsBoundingTime(time),
+  load: async ([client, time]) => {
+    const endpoint = await sessionEndPointCache.readAsync(client);
+    return client.getPointsBoundingTime(clamp(time, 0, endpoint.time));
```
bvaughn, line 315: I guess I'm not necessarily opposed to the `sessionEndPointCache`– I wonder if the Replay client shouldn't be the thing enforcing this– since we could still call client methods with potentially invalid times in other ways than just this cache?

packages/replay-next/src/contexts/FocusContext.tsx
hbenl: > Wouldn't we end up not calling client.requestFocusRange at step 3 even though we should?

No, because the `range` will still not be the same *object* even if it contains the same values.

packages/replay-next/src/suspense/ExecutionPointsCache.ts
hbenl: Personally I prefer `ReplayClient` to be a thin wrapper around the protocol and not add any logic to it.

packages/replay-next/src/contexts/FocusContext.tsx
bvaughn: Ah, right, because you wrapped it with `useMemo` 🤔 I guess that wouldn't break but it's also not the clearest relationship.

packages/replay-next/src/suspense/ExecutionPointsCache.ts
bvaughn: It already has logic for e.g. waiting for a range to be loaded before making a request. This doesn't seem substantially different from that, does it?

bvaughn: Btw I don't feel strongly enough about this to block the PR or anything 😄  It's just a preference.

PR #9107, opened 2023-04-26
Remove Playwright screenshot test fixture JSON data

bvaughn:


bvaughn:
Looks like this passed CI in about the same overall duration as before. So let's try landing it.

PR #9108, opened 2023-04-26
Change regex in E2E test detection of CodeTypeAhead

jcmorrow:
In this replay of a failing E2E test: https://app.replay.io/recording/logpoints-02-conditional-log-points--0bafa800-dcb1-4a27-9c9f-b4251317d833?point=37644152233979824960128594120545102&time=13798&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiNjEzMzQwMDY2NTEzMTgzMTk4Njg5ODMyMTE3NTg0NTI3MzYiLCJ0aW1lIjoyODQ3NX19

It looks like maybe we did not close the autocomplete because `page.locator('[data-test-name="CodeTypeAhead"]')` did not return anything because the actual component was `PointPanel-20-CodeTypeAhead`?

Might fix: https://linear.app/replay/issue/FE-1434/failing-logpoints-02

### DIFF
--- a/packages/e2e-tests/helpers/lexical.ts
+++ b/packages/e2e-tests/helpers/lexical.ts
@@ -24,7 +24,7 @@ export async function focus(page: Page, selector: string) {
 }
 
 export async function hideTypeAheadSuggestions(page: Page, selector: string) {
-  const list = page.locator('[data-test-name="CodeTypeAhead"]');
+  const list = page.locator('[data-test-name$="CodeTypeAhead"]');
 
   if (await list.isVisible()) {
     const input = page.locator(selector);


bvaughn:
> It looks like maybe we did not close the autocomplete because `page.locator('[data-test-name="CodeTypeAhead"]')` did not return anything because the actual component was `PointPanel-20-CodeTypeAhead`?

Looking at the code, there are two components that use an attribute with that value:

https://github.com/replayio/devtools/blob/73f03a64dd530a12cf863438ea53e6d350315542/packages/replay-next/components/lexical/plugins/code-completion/CodeCompletionPlugin.tsx#L20

https://github.com/replayio/devtools/blob/73f03a64dd530a12cf863438ea53e6d350315542/packages/replay-next/components/lexical/CommentEditor.tsx#L260

Both of these _can_ be overridden, and the `hideTypeAheadSuggestions` helper seems like maybe it's not the best to deal with that case.

I think your change _should_ work but it could still be broken by anything that passes in a different `dataTestName` prop. (Though I guess we'd deal with that if we have to.)

I think for now `CodeEditor` is the only place this plug-in is used and it's following the pattern.

https://github.com/replayio/devtools/blob/73f03a64dd530a12cf863438ea53e6d350315542/packages/replay-next/components/lexical/CodeEditor.tsx#L241-L244

bvaughn (approved):

jcmorrow:
Nice, thanks for the extra explanation!

PR #9110, opened 2023-04-27
Send unparseable editor state to Sentry (FE-1437)

hbenl:


packages/replay-next/components/lexical/CommentEditor.tsx
```diff
@@ -16,6 +16,7 @@ import { HistoryPlugin } from "@lexical/react/LexicalHistoryPlugin";
 import { createEmptyHistoryState } from "@lexical/react/LexicalHistoryPlugin";
 import { MarkdownShortcutPlugin } from "@lexical/react/LexicalMarkdownShortcutPlugin";
 import { RichTextPlugin } from "@lexical/react/LexicalRichTextPlugin";
+import * as Sentry from "@sentry/react";
```
markerikson, line 19: :hand: Do we really want to import _all_ of Sentry here? odds are it's probably all in the bundle, but maybe we could save a bit on bundle size by just importing `captureException` or something.

PR #9116, opened 2023-04-27
Shard screenshot tests

bvaughn:
This PR is large mostly because it splits up large test files into one file per test. Aside from that, there are very few significant changes (and I'll comment on each to make them easier to spot).

## Why split up the tests?
Playwright will group all tests within any given file into a single shard (https://github.com/microsoft/playwright/issues/21497). That means if there are e.g. 5 test files, we can use (at most) 5 shards. (In practice, it seems like we can use even fewer for some reason– many end up with 0 tests or 0 workers).

Note that I choose the convention that groups of tests still share a `beforeEach` function (in `./beforeEach.ts`) as well as shared helper methods (in `./shared.ts`) because it helps reduce repetition and will hopefully also make it easier to update recordings in the future.

## How much does this improve overall performance?
Most of our shards now complete in < 4 minutes (vs 17.5 minutes before). This is an increase in the _total time_ (because of installing redundant Playwright deps) but the effective time decreases by ~80%.

## Changes
- [x] Break Replay screenshot tests into multiple shards so they complete faster
- [x] Break test cases into separate files so Playwright will more efficiently schedule things.
- [x] Update Delta to pull from multiple actions/shards: https://github.com/replayio/delta/pull/39
- [x] Update PR comment action to account for multiple actions/shards: https://github.com/replayio/delta/pull/39


packages/replay-next/playwright/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "playwright:install": "playwright install chromium",
     "test": "playwright test tests",
-    "test:debug": "PWDEBUG=console VISUAL_DEBUG=true yarn test",
-    "test:write-snapshot-image-files": "WRITE_SNAPSHOT_IMAGE_FILES=true playwright test tests"
+    "test:ci": "WRITE_SNAPSHOT_IMAGE_FILES=true playwright test tests --shard=$SHARD/$SHARDS",
```
bvaughn, line 13: This is a core part of the change.

.github/workflows/playwright-snapshots.yml
```diff
@@ -18,6 +18,12 @@ env:
 jobs:
   run_tests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Playwright shards using test files (a file will never be split into more than one shard)
+        # Using more shards than we have test files means some will end up with 0 tests (and will fail)
+        shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
bvaughn, line 26: This is a core part of the change.

hbenl (approved):

bvaughn:
The Delta side of this has been merged and released.

I re-ran the Playwright task on main to generate the new data format needed see [Job 8634574604](https://github.com/replayio/devtools/actions/runs/4845578110/jobs/8634574604)).

Now this branch is re-running to see if the GitHub actions successfully interpret the resulting screenshots.

PR #9117, opened 2023-04-27
Show Passport immediately when completing Tour

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-569/investigate-why-passport-doesnt-always-appear-for-new-users

Our user flow used to look like this:

Complete Tour -> Click button -> Show Passport

But this meant that anyone who doesn't click the button misses out on the Passport. So this fix simply removes step two, so now it looks like this:

When you complete the Tour, Passport immediately appears.


jonbell-lot23:
I'm pushing this live because it's causing trouble for new users, it's a small fix, and the code is looking ok.

PR #9119, opened 2023-04-28
Split the passport into 4 sections

jonbell-lot23:
Was:
image

Now:
image


jonbell-lot23:
I'm pushing this live because it's basically copy-level changes, rather than something code-oriented.

PR #9124, opened 2023-04-28
Reimplement Cypress "Jump to Code" button and improve hit detection logic

markerikson:
The Cypress "Jump to Code" button's functionality got overwritten with just jumping to the test source location as part of https://github.com/replayio/devtools/pull/9056 .  Additionally, I think we lost the behavior that clicking the test step list item did a seek to the step timestamp.

This PR:

- Updates the event listener detection logic to accept an optional end point to constrain the event listener API call.  If provided we use that, otherwise we still fall back to an arbitrary end time (increased from 200ms to 500ms)
- Tweaks the logic inside `useJumpToSource` to hopefully improve location handling:
  - I think there was a variable name clash with the use of `location` that resulted in results not always getting assigned to the outer variable, and thus not opening the location
  - Added a `seek()` call to jump to the timestamp
- Removes the use of `setTransition()` in `TestCaseSectionRow`, as this seemed to somehow interfere with test step list item clicks and opening locations
  - Prior to this PR, clicking on a list item often took _two_ clicks on the same list item somehow. It still seems a bit inconsistent, but it's more reliably jumping to the right line on a single click now.
- Reworks `AnnotatedTestStepRow` to reapply the "Jump to Code" button logic I had originally implemented in #9002 
  - Moved the `useJumpToSource` click handler to the list item outer div
  - Re-applied the call to the "Jump to Code" event listener logic and passed that handler to the jump button

### DIFF
--- a/src/ui/components/TestSuite/views/TestItem/TestSteps/AnnotatedTestStepRow.tsx
+++ b/src/ui/components/TestSuite/views/TestItem/TestSteps/AnnotatedTestStepRow.tsx
@@ -1,15 +1,18 @@
-import { Suspense, memo, useContext, useMemo } from "react";
+import { Suspense, memo, useContext, useMemo, useState } from "react";
 
 import { getExecutionPoint } from "devtools/client/debugger/src/selectors";
 import Loader from "replay-next/components/Loader";
 import useSuspendAfterMount from "replay-next/src/hooks/useSuspendAfterMount";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
-import { JumpToCodeButton } from "ui/components/shared/JumpToCodeButton";
+import { seek } from "ui/actions/timeline";
+import { jumpToClickEventFunctionLocation } from "ui/components/Events/Event";
+import { PointWithEventType } from "ui/components/Events/Event";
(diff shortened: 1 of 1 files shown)


src/ui/components/TestSuite/views/TestItem/TestCaseSectionRow.tsx
```diff
@@ -79,7 +79,9 @@ export function TestCaseSectionRow({
       data-status={status}
       data-type={testStep.type}
       data-test-name="TestCaseSectionRow"
-      onClick={() => startTransition(() => selectTestStep(testStep))}
+      onClick={() => {
+        selectTestStep(testStep);
+      }}
```
bvaughn, line 84: > Prior to this PR, clicking on a list item often took two clicks on the same list item somehow. It still seems a bit inconsistent, but it's more reliably jumping to the right line on a single click now.

I wish we had dug into _why_ this required two clicks before. I wouldn't expect a change like this to "fix" behavior like you're describing. Sounds like it maybe just changed the timing of it but left it around still.

I think it's best to try updating values like this in transitions in case any views suspend in response to the change. (This can cause React to unmount already-mounted components so it can show their fallback states.)

I don't know if we'll see that happening here or not but it's something to look out for after this change.

PR #9125, opened 2023-04-28
Add a longer timeout for error stacks

jcmorrow:
I am not sure why this is taking so long, that would be worth investigating, but I am pretty sure this will at least make the test pass.

Example failure: https://app.replay.io/recording/a3e22dbb-61de-4c6c-a5d6-11769ba4a49d


bvaughn:
Seems like this PR was abandoned so I'm going to close it as part of cleanup. Please re-open and assign a reviewer if I'm mistaken though!

PR #9126, opened 2023-04-29
Allow the Playwright test suite panel to be shown

jcmorrow:
We will open the test suite panel by default if the recording is a test recording, but if it's a *playwright* recording then it's impossible to ever return to the test suite panel if you've ever navigated away. We use playwright tests for our E2E test suites, so this gets me all the time. I understand we might not want to show the Cypress logo so I've just thrown in a random logo instead (and I'd be happy to change it to something else if someone has a different suggestion).

Also, it's common for error messages to have white space breaks in them to show things like diffs, so we should display those rather than letting everything get crammed onto one line.

### Before
CleanShot 2023-04-29 at 12 01 52

### After

CleanShot 2023-04-29 at 12 01 08


jonbell-lot23 (approved):
Excellent -- I'll come up with another icon later

jonbell-lot23:
Btw: feel free to push this live, then I'll do the icon work as a separate topic

bvaughn (approved):

markerikson:
@jcmorrow fix the linting?

PR #9131, opened 2023-05-01
Remove flicker issue in Passport

jonbell-lot23:
The Passport was re-rendering a handful of times during page load, so these changes make sure the render only happens a single time.

### DIFF
--- a/src/ui/components/Passport/Passport.tsx
+++ b/src/ui/components/Passport/Passport.tsx
@@ -190,6 +190,10 @@ const Passport = () => {
     return Math.floor(Math.random() * (max - min + 1) + min);
   };
 
+  const [randomBottom, setRandomBottom] = useState(rand(220, 320));
+  const [randomRight, setRandomRight] = useState(rand(-50, 0));
+  const [randomRotation, setRandomRotation] = useState(rand(-20, 20));
+
   const renderSection = (section: Section, sectionIndex: number) => {
     return (
       <div className={styles.section}>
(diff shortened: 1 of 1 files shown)


src/ui/components/Passport/Passport.tsx
```diff
@@ -190,6 +190,10 @@ const Passport = () => {
     return Math.floor(Math.random() * (max - min + 1) + min);
   };
 
+  const [randomBottom, setRandomBottom] = useState(rand(220, 320));
```
markerikson, line 193: :hand: Small suggested tweak:

Technically, this is going to call this three `rand()` functions _every_ time the component renders. But, it will only _use_ the values from the _first_ call.

`useState` lets you pass a function as the "initial value", in which case it will call that function during the initial render, and _not_ during future renders.

In this case, we could do: `useState(() => rand(220, 320))`.

This isn't a real perf problem for this example, because `rand()` is a trivial function with no meaningful cost.  But it's a good practice to follow in general, and might be useful sometime in the future.

markerikson (approved):

src/ui/components/Passport/Passport.tsx
jonbell-lot23: Thank you!

I just realised that both my code and yours is doing something I don't want: it's forcing the badge's position. I want to be able to click 5 different line items and have each badge show up in a slightly different place, but this is fixing everything into one place.

That's fine for now, but I'll want to bring that back later.

jonbell-lot23:
I accidentally merged this without Mark's suggestions. Will follow up.

PR #9148, opened 2023-05-01
Add Playwright icon

jonbell-lot23:
30 second Loom: https://www.loom.com/share/459c2d1414b34f019b21464d5a37d017

cc @jcmorrow -- thanks again for fixing this bug, now here's the playwright version of your fix!


jonbell-lot23:
cc @jasonLaster Thanks for the Playwright icon, here's the PR!

src/ui/components/Toolbar.tsx
```diff
@@ -221,7 +259,7 @@ export default function Toolbar() {
             />
           ) : (
             <ToolbarButton
-              icon="folder"
+              icon="playwright"
```
ryanjduffy, line 262: This will show the playwright icon for regular (non-test suite) replays. Maybe we should make a `TestSuitesToolbarButton` that picks between the (eventually) various icons based on the test runner name?

ryanjduffy (changes requested):

src/ui/components/Toolbar.tsx
ryanjduffy: I was wrong ... ignore this

ryanjduffy (approved):

PR #9149, opened 2023-05-01
Update RDT bundles to latest minified, and combine evaluation calls

markerikson:
This PR:

- Updates the two React DevTools JS bundles to the latest stripped-down and minified versions that we're using on the backend with the RDT routine
- Updates the RDT injection logic to combine the two bundles + the various surrounding setup steps into a single TS function, to cut down on the number of evaluation calls

Confirmed that the RDT UI still works and I can still inspect components.  Example response from selecting a `<Counter>` component:

image


bvaughn:
Truthfully, I'm not sure why I'm tagged as a reviewer on this.

markerikson:
Trying to keep you in the loop and get someone else's eyes on it, I guess :)

PR #9165, opened 2023-05-08
use-context-menu

bvaughn:
Swap internal context menu hook and components for [`use-context-menu` package](https://www.npmjs.com/package/use-context-menu) which offers some better accessibility features (for keyboard menus).

---

Note that Delta failures are expected right now, unfortunately.


bvaughn:
This is pretty rote, so I'm going to land it.

PR #9166, opened 2023-05-08
Stop playback if Video context menu is shown

bvaughn:


src/ui/components/useVideoContextMenu.tsx
```diff
@@ -80,6 +83,11 @@ export default function useVideoContextMenu({
   };
 
   const onShow = async (event: UIEvent) => {
+    console.log("onShow() isPlaying?", isPlaying);
```
markerikson, line 86: :hand: Logging

markerikson (approved):
Fine other than the stray log

src/ui/components/useVideoContextMenu.tsx
bvaughn: Oops. Thanks

PR #9168, opened 2023-05-09
Add OutlineHitCountsCache (FE-1301)

hbenl:
- move the logic from `ReplayClient.getSourceHitCounts()` to the `SourceHitCountsCache`
- add `OutlineHitCountsCache` (which makes use of the `breakpointLocation` added to `FunctionOutline` in replayio/backend#7656)
- use the new cache in the outline panel
- the @replayio/protocol package still needs to be updated once replayio/backend#7678 has landed

### DIFF
--- a/packages/replay-next/src/suspense/OutlineHitCountsCache.ts
+++ b/packages/replay-next/src/suspense/OutlineHitCountsCache.ts
@@ -1,0 +1,80 @@
+import {
+  ClassOutline,
+  FunctionOutline,
+  PointRange,
+  SameLineSourceLocations,
+  SourceId,
+  SourceLocation,
+} from "@replayio/protocol";
+import { Cache, createCache } from "suspense";
+
+import { ReplayClientInterface } from "shared/client/types";
+
+import { sourceOutlineCache } from "./SourceOutlineCache";
+
+export interface FunctionOutlineWithHitCounts extends FunctionOutline {
+  hits: number;
+}
+
+export interface SourceOutlineWithHitCounts {
+  classes: ClassOutline[];
+  functions: FunctionOutlineWithHitCounts[];
+}
+
+export const outlineHitCountsCache: Cache<
+  [
+    replayClient: ReplayClientInterface,
+    sourceId: SourceId | undefined,
+    focusRange: PointRange | null
+  ],
+  SourceOutlineWithHitCounts
+> = createCache({
+  config: { immutable: true },
+  debugLabel: "OutlineHitCountsCache",
+  getKey: ([replayClient, sourceId, focusRange]) =>
+    sourceId && focusRange ? `${sourceId}:${focusRange.begin}-${focusRange.end}` : sourceId ?? "",
+  load: async ([replayClient, sourceId, focusRange]) => {
+    if (!sourceId) {
+      return { classes: [], functions: [] };
+    }
+
+    const { classes, functions } = await sourceOutlineCache.readAsync(replayClient, sourceId);
+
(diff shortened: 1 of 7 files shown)


packages/replay-next/src/suspense/OutlineHitCountsCache.ts
```diff
@@ -0,0 +1,80 @@
...
+        hitCounts.forEach(({ hits, location }) => {
+          const locationKey = `${location.line}:${location.column}`;
+          const previous = hitCountsByLocationKey.get(locationKey) ?? 0;
+          hitCountsByLocationKey.set(locationKey, previous + hits);
+        });
+      })
+    );
+
+    const functionsWithHitCounts = functions.map(functionOutline => {
+      //TODO update @replayio/protocol
```
bvaughn, line 68: If these TODOs don't already have a Linear task open to track them, will you open one?

src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
```diff
@@ -200,12 +177,14 @@ export default function SourceOutlineWrapper() {
...
+  const { range: focusRange } = useContext(FocusContext);
 
   let symbols: getSourceOutlineResult | null = null;
   const symbolsCacheValue = useImperativeCacheValue(
-    sourceOutlineCache,
+    outlineHitCountsCache,
     replayClient,
-    selectedSource?.id
+    selectedSource?.id,
+    focusRange ? toPointRange(focusRange) : null
```
bvaughn, line 187: Because `OutlineHitCountsCache` is a regular cache (not an interval one) I think this is going to cause us to over-fetch in many cases?

e.g.

1. User opens Replay (no focus)
2. We fetch values for the whole recording
3. User sets focus
4. We re-fetch values for the new focus range (even though we've already fetched them)

It's not awful but I wonder if you considered using an interval cache for this? I haven't thought much about it so maybe it wouldn't make sense in this case.

bvaughn (approved):
Skimmed it. Looks good but please regenerate the JSOn fixtures.

packages/shared/client/types.ts
```diff
@@ -205,10 +206,9 @@ export interface ReplayClientInterface {
   getSessionId(): SessionId | null;
   getSourceHitCounts(
     sourceId: SourceId,
-    locationRange: SourceLocationRange,
-    sourceLocations: SameLineSourceLocations[],
+    locations: SameLineSourceLocations[],
     focusRange: PointRange | null
-  ): Promise<LineNumberToHitCountMap>;
+  ): Promise<HitCount[]>;
```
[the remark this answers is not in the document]
hbenl: Those fixtures were removed (together with the ability to regenerate them) in #9107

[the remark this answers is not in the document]
bvaughn: Haha oh gosh, that's embarrassing. I had forgotten about that. 😆 I should wait until I'm fully awake to review ode

src/devtools/client/debugger/src/components/SourceOutline/SourceOutline.tsx
bvaughn: Ah, okay that makes sense.

packages/replay-next/src/suspense/OutlineHitCountsCache.ts
hbenl: Create [FE-1462](https://linear.app/replay/issue/FE-1462/update-the-replayioprotocol-package)

PR #9172, opened 2023-05-09
Adding docs links to Passport

jonbell-lot23:
Passport is performing well even without links to docs, but this makes the experience that much more robust for people trying to learn about our features. 

image

Here's what the code does:

1. If there's a docs link, show it when a line item is selected
2. When clicking on the link, open the docs into a new tab
3. If there's no docs link, don't show the icon (we'll be writing these docs links later)

### DIFF
--- a/src/ui/components/Passport/Passport.module.css
+++ b/src/ui/components/Passport/Passport.module.css
@@ -45,6 +45,13 @@
   text-overflow: ellipsis;
 }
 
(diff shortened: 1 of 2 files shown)


markerikson (approved):

markerikson:
Needs a TS fix (`docsLink` could be undefined), but looks okay once that's done.

PR #9173, opened 2023-05-10
Fix wait-on step in screenshot workflow

hbenl:


hbenl:
I thought the problem was caused by jeffbski/wait-on#133 but apparently it wasn't.

PR #9190, opened 2023-05-12
Move resume operations out of ThreadFront (FE-1476)

hbenl:


src/ui/utils/resumeOperations.ts
```diff
@@ -0,0 +1,142 @@
...
+    sourceId,
+    locationsToSkip,
+  }: ResumeOperationParams & { point: ExecutionPoint }
+) {
+  if (
+    loadedRegions.loaded.every(
+      region =>
+        BigInt(point) < BigInt(region.begin.point) || BigInt(point) > BigInt(region.end.point)
+    )
+  ) {
```
bvaughn, line 43: nit: We have utility methods for this:

```suggestion
  if (!isRangeInRegions(region.begin, loadedRegions.loaded)) {
```

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -129,7 +130,7 @@ export const executeCommandOperation = createAsyncThunk<
   const symbols = sourceId ? await sourceOutlineCache.readAsync(replayClient, sourceId) : undefined;
   const nextPoint = await getResumePoint(replayClient, state, command);
 
-  const resp = await ThreadFront[command]({
+  const resp = await resumeOperations[command](replayClient, {
```
bvaughn, line 133: Pleasantly surprised to see that TypeScript covers this without falling back to an `any` type  :)

bvaughn (approved):

src/ui/utils/resumeOperations.ts
bvaughn: Seems worth making this change?

src/ui/utils/resumeOperations.ts
```diff
@@ -0,0 +1,142 @@
...
+    return null;
+  }
+
+  await ThreadFront.ensureAllSources();
```
markerikson, line 47: :hand: Can we kill this too? Surely we can use our sources cache or something instead.

hbenl: Yeah, that'll be part of my next PR where I clean up how we handle sources.

PR #9193, opened 2023-05-12
Hide internal-only settings based on user info

bvaughn:
This includes unauthenticated users.


markerikson (approved):

bvaughn:
Some of our e2e tests failed (not the screenshot ones). I wonder if any of them were relying on these "internal" settings for unauthenticated users? Dunno. Will try re-running.

PR #9197, opened 2023-05-15
Fix e2e log point selector regression

bvaughn:
Regressed in #9191


markerikson (approved):

jasonLaster (commented):
This change makes a ton of sense and i bet is super common. I wonder what we can do on the runtime to make detecting these race conditions easier?

PR #9201, opened 2023-05-15
Refactor existing "event listeners" / "Jump to Code" logic for consistency

markerikson:
This PR refactors all the existing event listeners logic as a preparation for switching over to using annotations from the in-review event listeners routine:

- Extracted evaluation mapper and constants to match the backend structure as of https://github.com/replayio/backend/pull/7514
- Moved all J2C logic out of `Event.tsx` and into a new folder, `actions/eventListeners`
- Separated out specific J2C logic and cache from utils
- Separated out the function that gets evaluated in the backend
- Updated event listener and React Panel logic to use the new source outline `functionOutline.breakablePosition` field instead of manually determing the first breakable position


src/devtools/client/inspector/event-listeners/EventListenersApp.tsx
```diff
@@ -131,10 +124,11 @@ export const EventListenersApp = () => {
                         }
                       >
                         <div className="pl-4">
-                          <div className="theme-fg-color3">
+                          {/* TODO We haven't collected/shown real capture values in a long time */}
+                          {/* <div className="theme-fg-color3">
                             useCapture:{" "}
                             <span className="theme-fg-color1">{capture ? "true" : "false"}</span>
-                          </div>
+                          </div> */}
```
bvaughn, line 131: Delete this dead code and comment

src/ui/actions/eventListeners/constants.ts
```diff
@@ -0,0 +1,15 @@
+export type InteractionEventKind = "mousedown" | "keypress";
+
+export const REACT_EVENT_PROPS: Record<InteractionEventKind, string[]> = {
+  mousedown: ["onClick"],
+  // Users may have added `onChange` to an <input>, or `onkeyPress` to other elements
+  keypress: ["onChange", "onKeyPress"],
```
bvaughn, line 6: What about `"onKeyDown"` or `"onKeyUp"`?

src/ui/actions/eventListeners/constants.ts
```diff
@@ -0,0 +1,15 @@
+export type InteractionEventKind = "mousedown" | "keypress";
+
+export const REACT_EVENT_PROPS: Record<InteractionEventKind, string[]> = {
+  mousedown: ["onClick"],
```
bvaughn, line 4: What about `"onMouseDown"` or `"onMouseUp"`?

src/ui/actions/eventListeners/evaluationMappers.ts
```diff
@@ -0,0 +1,187 @@
...
+  searchedNodes: SearchedNode[];
+  searchedNodesString: string;
+  clickWasInsideSubmitButton?: boolean;
+}
+
+export function createReactEventMapper(eventType: InteractionEventKind) {
+  const reactEventPropNames = REACT_EVENT_PROPS[eventType];
+  const eventClassNames = EVENT_CLASS_FOR_EVENT_TYPE[eventType];
+
+  // This will became evaluated JS code
```
bvaughn, line 43:
```suggestion
  // This will become evaluated JS code
```

bvaughn (commented):
Started skimming this but it looks like this PR is mostly just moving code around?

src/ui/actions/eventListeners/constants.ts
markerikson: Intentionally not handling "down/up" events for the event listeners / jump to code logic.  We don't show those in the info sidebar or the Cypress panel - it's just "clicks" and "keypresses".

bvaughn: Why?

markerikson:
Going to have to rebase this to pick up Holger's sources changes

src/ui/actions/eventListeners/constants.ts
markerikson: That won't actually work given the current implementation.  The sidebar list items are _specifically_ the timestamps + points for just `click` and `keypress` events, so if we start trying to add the up/down events, the timestamps won't match up.

jasonLaster: Yep - we'd have to improve the UI to also include other click-like events as well and rollup events that are close together.

markerikson: Yeah, and per discussion in Discord, I'm not sure that makes sense "conceptually" given that the list is currently representing "times that the user interacted with the app" (and _not_ "individual sub-events for one overall interaction")

bvaughn: The things you call "sub events" are user interactions though.

If we're trying to help people find where their code handled mouse or keyboard events, then not including these commonly used event types seems like an ommission.

markerikson: If we want to have a larger design discussion around changing all this, we can. But everything I've built so far for "Jump to Code" / "Event Listeners" has been based around the original design and behavior of the info sidebar.  Arguing about it in this PR (which is just a refactoring PR in the first place) doesn't really seem helpful or relevant :(

bvaughn: FE-1496

jonbell-lot23: Hey gang, just reading up on this.

I agree with and sympathise with the idea that commonly used event types that don't show up can feel confusing. Coming up with ways to message that, even if it's just mentioned in docs, is better than nothing.

I am glad we have filed FE-1496 to split that conversation into its own thing.

At all times we need to remind ourselves that the goal is "better than before." Jump to Code is 100% better than not having it. And I assume this refactor is better than life before the refactor.

We can discuss in FE-1496 what our new "better than before" bar includes, and if it's worth the time commitment, etc.

PR #9206, opened 2023-05-16
First authenticated e2e tests (comments)

bvaughn:
Adds our first authenticated e2e tests for comments:
- [x] Adding, editing, and deleting a source code comment
- [x] Comments and replies between multiple users

The bulk of this PR is adding new e2e helper methods for interacting with context menus and comment cards.

Note there are a couple of _TODO_ comments in this PR, blocked by SCS-1066:
* Tests must begin by cleaning up after previously failed or timed out tests (since we have to re-use recordings)
* Each authenticated test must have its own, unique recording id (else parallel tests might interfere with each other)

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -13,5 +13,5 @@ FRONTEND_API_SECRET=
 # API Key to source icons from Figma
 FIGMA_TOKEN=
 # Authenticated e2e tests
-E2E_FE_TEST_USER_1_API_KEY=
-E2E_FE_TEST_USER_2_API_KEY=
+E2E_USER_1_API_KEY=
+E2E_USER_2_API_KEY=
--- a/.github/workflows/delta.yml
+++ b/.github/workflows/delta.yml
@@ -69,6 +69,8 @@ jobs:
         run: yarn test:ci --reporter @replayio/playwright/reporter,line
         working-directory: packages/replay-next/playwright
         env:
+          API_KEY_1: ${{ secrets.E2E_USER_1_API_KEY }}
+          API_KEY_2: ${{ secrets.E2E_USER_2_API_KEY }}
           SHARD: ${{ matrix.shard }}
           SHARDS: 10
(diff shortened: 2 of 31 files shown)


bvaughn:
Looks like the GitHub secrets aren't being passed through correctly, because the authenticated test is failing with a URL that doesn't include the `apiKey` parameter.

bvaughn:
I think the only missing piece here is figuring out how to pipe through the API keys for these two test users.

I've created GitHub "secrets" for each:
> **Repository secrets**:
> * `E2E_FE_TEST_USER_1_API_KEY`
> * `E2E_FE_TEST_USER_2_API_KEY`

I've updated the Delta workflow to pass them through:
```yml
- name: Run playwright tests
  run: yarn test:ci --reporter @replayio/playwright/reporter,line
  working-directory: packages/replay-next/playwright
  env:
    E2E_FE_TEST_USER_1_API_KEY: ${{ secrets.E2E_FE_TEST_USER_1_API_KEY }}
    E2E_FE_TEST_USER_2_API_KEY: ${{ secrets.E2E_FE_TEST_USER_2_API_KEY }}
    SHARD: ${{ matrix.shard }}
    SHARDS: 10
```

I've updated the `test:i` script to pass them through:
```
"test:ci": "E2E_FE_TEST_USER_1_API_KEY=$E2E_FE_TEST_USER_1_API_KEY E2E_FE_TEST_USER_2_API_KEY=$E2E_FE_TEST_USER_2_API_KEY WRITE_SNAPSHOT_IMAGE_FILES=true playwright test tests --shard=$SHARD/$SHARDS",
```

And then the tests access them from `process.env`:
```js
await startTest(page, url, process.env.E2E_FE_TEST_USER_1_API_KEY);
```

But they're getting lost somewhere along the way.

bvaughn:
Landed as part of #9215.

PR #9203, opened 2023-05-16
Improved logged out experience (NGX)

jonbell-lot23:
**Old version**

1. We'd put a documentation link up in all situations
2. The documentation link (aka "fuchsia hat") wasn't dismissible

image

**New version**

1. If there are comments, we default to that tab with special copy
2. ... and when you click onto any other tab, the "fuchsia hat" goes away (localStorage is used for this)
3. If there are no comments, we show the standard documentation link
4. ... but when you click into any other tab, the "fuchsia hat" goes away

image

Here's a little demo:
https://www.loom.com/share/445b9aab2f7144ef9a481c7f470c8220

PR #9207, opened 2023-05-16
Improved logged out experience (NGX) (Attempt #2 with hotfix)

jonbell-lot23:
Got an "implicit any" error [here](https://github.com/replayio/devtools/pull/9203), this is attempting to fix that.


jonbell-lot23:
After a few bumps and scrapes I figured out the TSC issues with Mark. Finally merged and live!

PR #9208, opened 2023-05-17
Pause info UI tweaks

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-623/improve-print-statement-panel

Old:
image

New:
image

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css
@@ -3,7 +3,7 @@
   flex-direction: row;
   overflow: hidden;
   gap: 1ch;
-  padding: 0.25rem 0.5em;
+  padding: 0.25rem 0.5rem;
   cursor: pointer;
 }
 .Point[data-test-state="paused"] {
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css
@@ -4,6 +4,7 @@
   overflow: hidden;
   gap: 1ch;
   padding: 0.25rem 0.5em;
+  cursor: pointer;
 }
 .Point[data-test-state="paused"] {
   background-color: var(--background-color-current-execution-point);
@@ -11,9 +12,6 @@
 
 .Location {
   color: var(--theme-comment);
-}
-.Point:hover .Location {
-  display: none;
 }
 
 .CreatedByAvatar {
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointHeading.module.css
jonbell-lot23, line 1: This is a good example where .25 increments are too rigid. 0.36 is awkward, but's less awkward than icons not lining up.

bvaughn: I suspect this is actually a case of a style being messed up in another place, so that we have to hack around it by using non-standard increments

bvaughn: Yes. This was caused by the `.Point` style in `Breakpoint.module.css` being set to use `em` instead of `rem`

https://github.com/replayio/devtools/blob/6440e23d2a8102c86c24f334fdd84215fed10783/src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css#L6

This was _my_ mistake originally, but the fix should be there rather than here. There's no relation between `em` and `rem` so these could be misaligned later.

bvaughn (changes requested):

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointHeading.module.css
```diff
@@ -4,7 +4,7 @@
   width: 100%;
   font-size: var(font-size-regular);
   line-height: 1rem;
-  padding: 0.25rem 0.5rem;
+  padding: 0.25rem 0.36rem;
```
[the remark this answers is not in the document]
bvaughn: More explanation at https://www.loom.com/share/df597449e6d14207b7ec7f743c12a83a

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/BreakpointHeading.module.css
jonbell-lot23: Sounds good - so I'm reverting back to .25 (rather than .36) and then I'm changing Breakpoint.module.css to say 0.5rem rather than 0.5em

src/devtools/client/debugger/src/components/SecondaryPanes/Breakpoints/Breakpoint.module.css
```diff
@@ -3,7 +3,8 @@
   flex-direction: row;
   overflow: hidden;
   gap: 1ch;
-  padding: 0.25rem 0.5em;
+  padding: 0.25rem 0.25rem;
```
bvaughn, line 6: nit
```suggestion
  padding: 0.25rem;
```

bvaughn (approved):

PR #9209, opened 2023-05-17
Fix tooltip z-index

jonbell-lot23:
Old:
image

New:
image

### DIFF
--- a/src/ui/components/Passport/Passport.module.css
+++ b/src/ui/components/Passport/Passport.module.css
@@ -28,7 +28,7 @@
   padding-left: 0.4rem;
   border-radius: 3rem;
   margin-bottom: 1px;
-  z-index: 10;
+  z-index: var(--z-index-1);
   white-space: nowrap;
   overflow: hidden;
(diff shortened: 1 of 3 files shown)


src/ui/components/Passport/Passport.module.css
```diff
@@ -68,7 +68,7 @@
 .videoExample {
   border-bottom-left-radius: 6px;
   border-bottom-right-radius: 6px;
-  z-index: 100;
+  z-index: 5;
```
bvaughn, line 71: Don't we have z-index CSS variables? (I think you added them) Why don't we use them?

bvaughn (approved):

src/ui/components/Passport/Passport.module.css
jonbell-lot23: I don't recall doing that, but can look later

jonbell-lot23: Moved to css vars

PR #9210, opened 2023-05-17
Improve NGX with a clearer onboarding message + library context

jonbell-lot23:
**Old**

We used to optimise our NGX (New Guest Experience) around downloading a browser, and we did it on a pretty ugly screen. We also offered people a chance to "go to that team" or "download replay" but the component buttons didn't match, because they said "skip" and "download replay". We filed that [here](https://linear.app/replay/issue/DES-586/ngx-button-and-copy-dont-match).

image

image

image

image

**New**

This design (which we used to use) is much better at introducing you to your library because it's a welcome message floating above your window. It doesn't provide a button for downloading a replay, but it does refer to the "launch replay" button at the top right.

image

**Hypothesis**

We often see people confused by the old approach, both on the big gray screen and the library. Some percentage of this confusion/failure rate is because invited users frequently are trying to *look at replays* rather than record their own. I predict we'll see our Tour/Passport numbers go up after this change, and I'll be tracking the progress to make sure.


jasonLaster:
Left aligning the text here feels off when the rest is centered. Thoughts?

image

jonbell-lot23:
> Left aligning the text here feels off when the rest is centered. Thoughts?
> 
> image

Agree, that's an old screen that I didn't touch yet. Propose filing as a separate quality of life fix.

jonbell-lot23:
https://linear.app/replay/issue/DES-649/interview-1

PR #9211, opened 2023-05-17
Fix testsuites error regression

jonbell-lot23:
**Old**

image

**New**

image

The margin got messed up here, this PR fixes it.


bvaughn (commented):
The border looks unbalanced on the bottom with only padding top.
It will look similar on the right if there's only padding left.

PR #9214, opened 2023-05-17
Shared log point (authenticated) e2e test

bvaughn:
Builds on top of #9206 and adds a test for shared log points.

Also updates the multi-user comments test (from #9206) to run each user in a separate browser context– to avoid any false positives from shared IndexedDB or localStorage.


markerikson (approved):

bvaughn:
Landed as part of #9215.

PR #9215, opened 2023-05-17
Run authenticated e2e tests in Chromium

bvaughn:
We can revert this based on the outcome of RUN-1984.


bvaughn:
`logpoints-02` is flaking on this but it's also flaking on main, and we have a Linear issue to address it (FE-1497). It passes for me locally. so I'm going to merge.

PR #9216, opened 2023-05-17
Rework "Jump to Code" buttons to use data from routine annotations

markerikson:
> **NOTE**: This PR depends on https://github.com/replayio/backend/pull/7759 landing to fix issues with the event listeners routine, and then we need that routine turned on for real and running consistently before we can switch from "figure out the location when the user clicks" to "use the precalculated location data from the routine".

This PR:

- Converts the existing annotation handling for React DevTools, Redux DevTools, and annotation kinds into single-entry caches
  - deletes the existing `state.reactDevTools` slice
  - Updates the TS types and some variable names in `ReactDevTools.tsx` to better describe the parsed annotation contents
- Updates the info events panel and Cypress test steps panelto use precalculated annotations data for the "Jump" buttons

This doesn't yet delete the existing client-side "jump" logic - I can do that in a follow-up PR.


bvaughn:
This is still titled "DRAFT" so I'll hold off on reviewing it.

markerikson:
FWIW the logic looks good atm, so I think the code is ready for review. I just don't want to merge it until I've had a further chance to confirm that it all works right with real annotations data coming from the backend.  Logan and I tried to check that a couple hours ago and it looked okay at first glance, but the backend was doing a revert and we had to set the checking aside.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -212,10 +214,16 @@ export default async function setupDevtools(store: AppStore, replayClient: Repla
...
   setupNetwork(store, replayClient);
-  setupReactDevTools(store, ThreadFront);
   setupBoxModel(store, startAppListening);
   setupRules(store, startAppListening);
 
+  ThreadFront.waitForSession().then(() => {
+    // Precache annotations
+    annotationKindsCache.readAsync(replayClient);
+    reactDevToolsAnnotationsCache.readAsync();
+    eventListenersJumpLocationsCache.readAsync();
```
bvaughn, line 224: Trivial naming nit: `cache.prefetch` might be a better semantic fit

bvaughn (approved):
I don't have a ton of context to this code so I just skimmed it. Looks good though.

bvaughn (approved):
I don't have a ton of context to this code so I just skimmed it. Looks good though.

PR #9218, opened 2023-05-18
Port workaround for preferred locations and corresponding source IDs

markerikson:
This PR:

- Ports a workaround for `getPreferredLocation`'s corresponding source IDs assertion that _was_ in place prior to #9204 over to the current sources handling logic.

Original hack was here:

- https://github.com/replayio/devtools/blob/4c972f026f400249b8e9e518f1c1b64e097882e4/src/ui/utils/preferredLocation.ts

I don't know why we have these corresponding source ID mismatches, but apparently we've been working around it for a long time.  Filed FE-1508 as a follow-up to review why that is and hopefully remove this hack.


bvaughn (approved):
<dog-science>

PR #9220, opened 2023-05-19
Improve loading screen states

jonbell-lot23:
**Problem we're trying to solve**

* Right now our loading screen can stall out silently, so this PR makes it more clear
* There's no option to contact us, so this PR adds one
* This is a good place to add some tone, so this PR adds some

Loom demonstration: https://www.loom.com/share/d397fe24860b4c4a8e16dd5c49796843

Addresses https://linear.app/replay/issue/DES-635/our-loading-screen-needs-to-provide-help-when-things-stall-out

### DIFF
--- a/src/ui/components/shared/LoadingScreen.module.css
+++ b/src/ui/components/shared/LoadingScreen.module.css
@@ -24,7 +24,7 @@
 }
 
 .messageWrapper a:hover {
-  color: rgb(80, 202, 255);
+  color: var(--primary-accent);
 }
 
 .viewportWrapper {
--- a/src/ui/components/shared/LoadingScreen.module.css
+++ b/src/ui/components/shared/LoadingScreen.module.css
@@ -1,0 +1,38 @@
+.loadingScreenWrapper {
+  position: relative;
+  display: flex;
+  width: 24rem;
+  flex-direction: column;
+  align-items: center;
+  padding: 1rem 1rem 3rem;
+  margin: 0.25rem;
+}
+
+.hoverboardWrapper {
+  height: 8rem;
+  width: 8rem;
+  cursor: pointer;
+}
+
+.messageWrapper {
+  font-size: 0.875rem;
+  text-align: center;
+}
+
+.messageWrapper a {
+  text-decoration: underline;
+}
+
+.messageWrapper a:hover {
+  color: rgb(80, 202, 255);
+}
+
+.viewportWrapper {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 1rem;
+  margin: 0.5rem;
+  border-radius: 0.5rem;
+}
--- a/src/ui/components/shared/LoadingScreen.module.css
+++ b/src/ui/components/shared/LoadingScreen.module.css
@@ -3,9 +3,6 @@
   display: flex;
   width: 24rem;
   flex-direction: column;
-  align-items: center;
-  padding: 1rem 1rem 3rem;
-  margin: 0.25rem;
 }
 
 .hoverboardWrapper {
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -69,16 +69,9 @@ function LoadingScreen({
       const randomIndex = Math.floor(Math.random() * phrases.length);
       setMessage(phrases[randomIndex]);
     };
-
     const phraseTimeout = setTimeout(changeMessage, 5000);
-
     return () => clearTimeout(phraseTimeout);
   }, []);
-
-  useEffect(() => {
-    const phraseTimeout = setTimeout(changeMessage, 5000);
-    return () => clearTimeout(phraseTimeout);
-  }, [changeMessage]);
 
   useEffect(() => {
     const stalledTimeoutId = setTimeout(() => {
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -64,9 +64,15 @@ function LoadingScreen({
 
   const [message, setMessage] = useState(fallbackMessage);
 
-  const changeMessage = useCallback(() => {
-    const randomIndex = Math.floor(Math.random() * phrases.length);
-    setMessage(phrases[randomIndex]);
+  useEffect(() => {
+    const changeMessage = () => {
+      const randomIndex = Math.floor(Math.random() * phrases.length);
+      setMessage(phrases[randomIndex]);
+    };
+
+    const phraseTimeout = setTimeout(changeMessage, 5000);
+
+    return () => clearTimeout(phraseTimeout);
   }, []);
 
   useEffect(() => {
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -18,15 +18,15 @@ const Hoverboard = dynamic(() => import("./Hoverboard"), {
 export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
   const [hoverboardColor, setHoverboardColor] = useState(colorOptions[2]);
 
-  const changeHoverboardColor = () => {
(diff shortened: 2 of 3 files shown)


src/ui/components/shared/LoadingScreen.module.css
```diff
@@ -0,0 +1,38 @@
...
+  font-size: 0.875rem;
+  text-align: center;
+}
+
+.messageWrapper a {
+  text-decoration: underline;
+}
+
+.messageWrapper a:hover {
+  color: rgb(80, 202, 255);
```
bvaughn, line 27: ⚠️  Hard-coded colors may cause problems with themes

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -28,49 +29,67 @@ export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
...
+  "Ensuring temporal paradox safeguards...",
+  "Being careful not to mess up your timeline...",
+];
+
 function LoadingScreen({
   uploading,
   awaitingSourcemaps,
   fallbackMessage,
-  stalledTimeout = 9000,
+  stalledTimeout = 15000,
```
bvaughn, line 60: Is this a high enough value to avoid false positives? I _think_ so.

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -28,49 +29,67 @@ export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
   }, [hoverboardColor]);
```
bvaughn, line 29: This should be triggering a Lint failure. This effect is missing a dependency (`changeHoverboardColor`)

You could fix this in a variety of ways:
1. Memoize `changeHoverboardColor` with `useCallback`
2. Or just add the inline callback dependency (and maybe change `useEffect` to use `setTimeout` so it's clearer what it's meant to be doing– one delayed change per render, at most)

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -28,49 +29,67 @@ export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
...
+    const randomIndex = Math.floor(Math.random() * phrases.length);
+    setMessage(phrases[randomIndex]);
+  }, []);
 
-    return () => clearTimeout(timeout);
-  }, [fallbackMessage, stalledTimeout]);
+  useEffect(() => {
+    const phraseTimeout = setTimeout(changeMessage, 5000);
+    return () => clearTimeout(phraseTimeout);
+  }, [changeMessage]);
```
bvaughn, line 75: Since the effect is the only code that uses `changeMessage`, you could simplify this:
```js
useEffect(() => {
  const changeMessage = () => {
    // Nit that it's possible to get the same index twice this way;
    // A more robust way of doing this would be sorting the array once (during module initialization?)
    // and then just stepping through in-order (`index = (index + 1) % length`)
    const randomIndex = Math.floor(Math.random() * phrases.length);
    setMessage(phrases[randomIndex]);
  };

  const phraseTimeout = setTimeout(changeMessage, 5000);

  return () => clearTimeout(phraseTimeout);
}, []);
```

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -28,49 +29,67 @@ export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
...
   );
 }
 
+const phrases = [
+  "Tuning the flux capacitor...",
+  "Prepping DeLorean for time travel...",
+  "Revving up to 88 miles per hour...",
+  "Gathering 1.21 gigawatts of power...",
+  "Ensuring temporal paradox safeguards...",
+  "Being careful not to mess up your timeline...",
```
jasonLaster, line 53: Can we go further? https://sharegpt.com/c/G7i464K

At some point we should add a section to our language guide around how we discuss time travel, but it's probably too early for that. At this stage, I love seeing how experimental we can be

jonbell-lot23: Love this idea, will play with it in a separate PR away from the code changes Brian's suggesting.

jonbell-lot23:
Ok, comments from PR addressed. How's it look now?

bvaughn:
I rebased this so the new Delta will run on it. I also fixed the Lint issue and _one_ of the TypeScript issues. Realized this didn't look ready for review at that point, so I left the other TS issue :)

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -17,60 +18,77 @@ const Hoverboard = dynamic(() => import("./Hoverboard"), {
...
-  };
+  }, []);
 
   useEffect(() => {
-    const interval = setInterval(changeHoverboardColor, 5_000);
-    return () => clearInterval(interval);
-  }, [hoverboardColor]);
+    const timeoutId = setTimeout(changeHoverboardColor, 5000);
+    return () => clearTimeout(timeoutId);
+  }, [changeHoverboardColor]);
```
bvaughn, line 29: I think there was a misunderstanding from my comment https://github.com/replayio/devtools/pull/9220#discussion_r1199032258

I meant do one or the other, not both. The code now will only change the color once. (I think you intended the color to change every 5 seconds?) Maybe I'm misunderstanding.

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -17,60 +18,77 @@ const Hoverboard = dynamic(() => import("./Hoverboard"), {
...
 
-    return () => clearTimeout(timeout);
-  }, [fallbackMessage, stalledTimeout]);
+  useEffect(() => {
+    const changeMessage = () => {
+      const randomIndex = Math.floor(Math.random() * phrases.length);
+      setMessage(phrases[randomIndex]);
+    };
+    const phraseTimeout = setTimeout(changeMessage, 5000);
+    return () => clearTimeout(phraseTimeout);
```
bvaughn, line 73: Since the effect will only run once now (good) this will only change the message once (maybe not what you wanted?)

bvaughn (commented):
It looks okay now except I think you have two timeouts where you want intervals.

bvaughn (commented):
It looks okay now except I think you have two timeouts where you want intervals.

src/ui/components/shared/LoadingScreen.tsx
jonbell-lot23: Ah yes, good call. This is indeed what is designed. We only want a single cutesy message to appear before the stalled message, so it'll go like this:

1. 5 seconds of "loading"
2. 10 seconds of a single cutesy message
3. Stalled message that never goes away

jonbell-lot23:
GitHub hiccupped and showed 3 copies of Brian's message, and my responses/resolved may have gotten a bit messed up too. Here's the status:

1. We do indeed want setTimeout for the phrase generator, because we only one one cutesy error to appear before the stalled error. So that part is correct.
2. But I did make a mistake with the color. It's true that we want it swapping every five seconds, so I needed to put setInterval back.
3. I had a quick chat with Mark to double-check my code, and we added some lines of code to clarify that the cutesy message is designed to only load once.

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -17,60 +18,80 @@ const Hoverboard = dynamic(() => import("./Hoverboard"), {
...
+    }, stalledTimeout);
+    return () => clearTimeout(stalledTimeoutId);
+  }, [stalledTimeout]);
 
+  const waitingForMessage =
+    awaitingSourcemaps || uploading ? (
+      <span>Uploading {Math.round(uploading?.amount ? Number(uploading.amount) : 0)}Mb</span>
+    ) : (
+      <span dangerouslySetInnerHTML={{ __html: message }}></span>
+    );
```
bvaughn, line 91: async PR comment: This (the `waitingForMessage` message) is no longer being used (and should be deleted). That would also mean the `awaitingSourcemaps` and `uploading` props are not being used, so we could remove them (and the Redux `connect()` HOC) entirely.

PR #9221, opened 2023-05-19
Write Delta snapshot files to unique paths per test

bvaughn:
Initially screenshots were being saved to a directory that included the test file name and the description, e.g. `packages/replay-next/playwright/snapshots/tests/console/should-be-filterable-on-complex-content.ts-snapshots/dark/…`

As of PR #8407, screenshots for CI tests we being written to a flat directory, `packages/replay-next/playwright/visuals/dark/…`

This made a race condition for tests that happened to have screenshots with the same names. This PR restores something more like the original behavior. (Note that it will generate _a lot_ of Delta changes.)

Tagging reviewers mostly as an FYI, not for actual review.

### DIFF
--- a/packages/replay-next/playwright/.gitignore
+++ b/packages/replay-next/playwright/.gitignore
@@ -4,3 +4,4 @@ cache_buster
 playwright-report*
 snapshots/tests
 test-results
+visuals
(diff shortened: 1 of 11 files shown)


bvaughn:
Looking at the Workflow, this is looking better:
> Upload file "./playwright/visuals/should-properly-bucket-properties-for-an-overflowing-array/dark/overflowing-array-first-bucket-expanded.png"

PR #9225, opened 2023-05-19
Removed global TestInfo in favor of passing through argument

bvaughn:


bvaughn:
Searching through the `uploadSnapshots.ts` output, I don't see any references to `"./playwright/visuals/dark` or `"./playwright/visuals/light` so this seems to have helped.

PR #9231, opened 2023-05-22
Reduce Console screenshot test flakiness

bvaughn:
Looking at Delta's [reported flaky screenshots](https://delta.replay.io/flaky/replay?branchId=76&runId=796&snapshotId=57899&) I noticed a common pattern for Console screenshots.

For example, compare this:
image

To this:
image

In the second image, the console "exceptions" toggle is still suspending and the "warnings" and "node modules" toggles aren't enabled yet (maybe blocked by the Suspense). In either case, it seemed we weren't waiting on the toggles to be fully applied before continuing with the test.

This PR adds two additional checks to:
1. Ensure the toggle has been applied (at the Console level)
2. Ensure all messages have finished loading data

Hopefully this will reduce one of the last major sources of flakiness in these tests.

### DIFF
--- a/packages/replay-next/components/console/MessagesList.tsx
+++ b/packages/replay-next/components/console/MessagesList.tsx
@@ -1,6 +1,7 @@
 import { ForwardedRef, MutableRefObject, ReactNode, forwardRef, useContext, useMemo } from "react";
 
 import Icon from "replay-next/components/Icon";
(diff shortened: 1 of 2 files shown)


markerikson (approved):

bvaughn:
Okay looks like most of the changes are things going _from flaky_ to _no longer flaky_ which is good.

This one change for a search test makes me think we want to add a similar delay after filtering though:

image

image

PR #9233, opened 2023-05-23
Upload all Delta Replay recordings (not just failures)

bvaughn:
Even for tests that "pass" – it can be helpful to look at the Replay to see _why_ a screenshot changed if the change was unexpected.

I'm not sure if I'm doing this correctly :)


bvaughn:
Doesn't look like this worked. cc @ryanjduffy

PR #9234, opened 2023-05-23
Add lots of new loading screen phrases

jonbell-lot23:
The goal here is to add some tone/personality to our loading screen.


jonbell-lot23:
Merging! #biastoaction

We can/will modify this list as we see things to tweak, remove, and add.

PR #9235, opened 2023-05-23
Clicking eye icon in log point panel enables log point

bvaughn:
### [Loom overview of changes](https://www.loom.com/share/4fb38dd747e248e0a3cc0ba71bcfc323)

I think this is mostly a no-brainer good change, but there's an open question in my mind about the UX for the owner of the log point. (I would actually prefer that we remove the pencil icon entirely and just rely on cursor and hover border style changes to indicate "editable" but I realize that's more controversial.)

cc @jasonLaster, @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
+++ b/packages/replay-next/components/sources/log-point-panel/LogPointPanel.module.css
@@ -57,18 +57,16 @@
   align-items: center;
   gap: 1ch;
 }
-.ContentWrapper[data-logging-disabled] {
-  background-color: var(--point-panel-input-disabled-background-color);
-}
-.ContentWrapper:hover {
+.ContentWrapper[data-state-editable="true"]:hover {
   border-color: var(--point-panel-input-border-color-hover);
-}
-.ContentWrapper[data-logging-disabled] {
-  border-color: var(--point-panel-input-border-color);
 }
 .Panel[data-test-state="edit"] .ContentWrapper:focus,
 .Panel[data-test-state="edit"] .ContentWrapper:focus-within {
   border-color: var(--point-panel-input-border-color-focus);
+}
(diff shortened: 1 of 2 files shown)


markerikson (approved):

jasonLaster:
@bvaughn i trust ya

jonbell-lot23:
(Happy to move this conversation wherever we think it works best -- this Linear issue, a new one, here, whatever)

Brian,

You bring up really good points! I agreed with everything you said in your looms, from wanting to remove confusion but also feeling conflicted about some of the approaches. Here's what I'm seeing and what I think we should optimise for:

1. We want new people to "grok" print statements quickly, so subtlety works against us in that scenario
2. But we want confident Replay users to have a powerful/clear experience, so double icons works against us there

Putting aside the grossness of having a "if new user, do x; else do y" bit of code, I think that's probably our best bet. Off the top of my head, I think something like this is a good balance:

1. If a person has never set a print statement, use a pencil (and the eyeball is unlikely to be set at all)
2. After a person has set 1-3 print statements, never show the pencil again

Any thoughts about that? I think zero pencil might be too extreme for new users, but I think the current experience is too clumsy for our most committed users. Think this approach could work?

jonbell-lot23:
cc @bvaughn

bvaughn:
@jonbell-lot23 I think we should file a Linear issue to talk about this rather than talking about it on a closed PR, but in short– 

> If a person has never set a print statement, use a pencil (and the eyeball is unlikely to be set at all)
After a person has set 1-3 print statements, never show the pencil again

We don't track the number of print statements a person has "seen". ("Seen" is kind of an interesting question and not so straight forward, I think. It would have to mean...what– opened source and scrolled within the viewport? That granularity in tracking isn't something we do, and FWIW I would probably advocate against doing it unless we had a _really strong reason_.)

PR #9239, opened 2023-05-23
Support redirecting from /r/:id

jaril:
Quality of life improvement when slinging recording ID's around. 

Eventually it's probably better to use the shortened version (`/r/` instead of `/recording/`) for copied links.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -50,6 +50,10 @@ const baseNextConfig = {
         destination: "/recording/:id",
         permanent: true,
       },
+      {
+        source: "/r/:id",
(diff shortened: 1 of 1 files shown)


bvaughn:
I think this would be better as one or more go/links: https://app.golinks.io/

jonbell-lot23:
Hey @jaril , @bvaughn and I have a little discussion in #eng-frontend on this. Can you use a go/link to get the same results? Right now go/r/[id] goes to our admin app, but maybe we could do a similar redirect for this issue?

PR #9247, opened 2023-05-24
Show flaky tests in the dashboard

jasonLaster:
This is a short-term fix for the dashboard. SCS-1087 

The real fix would involve updating the TestRun Stats which come from hasura and would let us show whether the test run actually failed or had some flaky tests.

Screenshot 2023-05-24 at 11 47 07 AM

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -2,7 +2,7 @@ import { useContext, useMemo, useState } from "react";
 
 import { Recording } from "shared/graphql/types";
 import Icon from "ui/components/shared/Icon";
-import { groupRecordings } from "ui/utils/testRuns";
+import { RecordingGroup, groupRecordings } from "ui/utils/testRuns";
 
 import { TestResultListItem } from "./TestResultListItem";
 import { TestRunOverviewContext } from "./TestRunOverviewContainerContextType";
@@ -18,25 +18,25 @@ export function RunResults() {
 
   return (
     <div className="no-scrollbar flex flex-col overflow-y-auto">
-      <TestStatusGroup recordings={failedRecordings} label="Failed" />
-      <TestStatusGroup recordings={flakyRecordings} label="Flaky" />
-      <TestStatusGroup recordings={passedRecordings} label="Passed" />
+      <TestStatusGroup label="Failed" recordingGroup={failedRecordings} />
+      <TestStatusGroup label="Flaky" recordingGroup={flakyRecordings} />
+      <TestStatusGroup label="Passed" recordingGroup={passedRecordings} />
     </div>
   );
 }
 
-function TestStatusGroup({ recordings, label }: { recordings: Recording[]; label: string }) {
+function TestStatusGroup({
+  label,
+  recordingGroup,
+}: {
+  label: string;
+  recordingGroup: RecordingGroup;
+}) {
   const [expanded, setExpanded] = useState(true);
-  const count = recordings.length;
+  const count = recordingGroup.count;
   if (count == 0) {
     return null;
   }
-
-  const sortedRecordings = recordings.sort((a, b) =>
-    (a.metadata?.test?.file || 0) > (b.metadata?.test?.file || 0) ? 1 : -1
-  );
-
-  console.log("TestStatusGroup", label);
 
   return (
     <div className="flex flex-col">
@@ -57,8 +57,21 @@ function TestStatusGroup({ recordings, label }: { recordings: Recording[]; label
           />
         </div>
       </div>
-      {expanded &&
-        sortedRecordings.map(r => <TestResultListItem recording={r} key={r.id} label={label} />)}
+      {expanded && <Expanded label={label} recordingGroup={recordingGroup} />}
     </div>
   );
+}
+
+function Expanded({ label, recordingGroup }: { label: string; recordingGroup: RecordingGroup }) {
+  const recordings = useMemo(() => {
+    const recordings: Recording[] = [];
+    for (let fileName in recordingGroup.fileNameToRecordings) {
+      recordings.push(...recordingGroup.fileNameToRecordings[fileName]);
+    }
+    return recordings;
+  }, [recordingGroup]);
+
+  return recordings.map(recording => (
+    <TestResultListItem key={recording.id} label={label} recording={recording} />
+  )) as any;
 }
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -11,8 +11,9 @@ import styles from "../../../../Library.module.css";
 export function RunResults() {
   const testRun = useContext(TestRunOverviewContext).testRun!;
 
+  // TODO Don't keep re-computing this; it's expensive
   const { passedRecordings, failedRecordings, flakyRecordings } = useMemo(
-    () => groupRecordings(testRun.recordings),
+    () => groupRecordings(testRun.recordings ?? []),
     [testRun.recordings]
   );
 
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -1,8 +1,8 @@
-import orderBy from "lodash/orderBy";
-import { useContext, useState } from "react";
+import { useContext, useMemo, useState } from "react";
 
 import { Recording } from "shared/graphql/types";
 import Icon from "ui/components/shared/Icon";
+import { groupRecordings } from "ui/utils/testRuns";
 
 import { TestResultListItem } from "./TestResultListItem";
 import { TestRunOverviewContext } from "./TestRunOverviewContainerContextType";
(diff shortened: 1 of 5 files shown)


jasonLaster:
Made a couple of changes:
1. we now show both the passing and failing test for flakes
2. We show all tests in passed/failed. We'll need to group by groups in the future, but at least this is comprehensive

Screenshot 2023-05-24 at 3 40 46 PM

bvaughn:
Nice to see we're re-thinking this UI!

I wonder– can we change the left hand style too? I think that's an even more important place to differentiate between "some flakes but ultimately all passed" vs "some failed"

jasonLaster:
Filed an issue for it https://linear.app/replay/issue/FE-1531/show-flaky-test-count-in-the-list-view

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -55,7 +55,8 @@ function TestStatusGroup({ recordings, label }: { recordings: Recording[]; label
           />
         </div>
       </div>
-      {expanded && sortedRecordings.map((r, i) => <TestResultListItem recording={r} key={i} />)}
+      {expanded &&
+        sortedRecordings.map((r, i) => <TestResultListItem recording={r} key={i} label={label} />)}
```
bvaughn, line 59: nit
```suggestion
        sortedRecordings.map((r) => <TestResultListItem recording={r} key={r.id} label={label} />)}
```

src/ui/utils/testRuns.ts
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+  }
+
+  return {
+    passedRecordings,
+    failedRecordings,
+    flakyRecordings,
+    flakyCount: flakySet.size,
+  };
```
bvaughn, line 71: I think this hook might want to be a bit more prescriptive about the display counts. Unless I'm reading it wrong, right now we'll include flaky tests in both the flaky count and the failed count (because failed count is just `failedRecordings.length`)

Seems like we _don't_ want to do show that on the side bar.

src/ui/utils/testRuns.ts
```diff
@@ -0,0 +1,72 @@
...
+      const recordings = sortedRecordings.filter(r => r.metadata?.test?.file === file);
+      const passed = recordings.filter(r => testPassed(r));
+      const failed = recordings.filter(r => testFailed(r));
+
+      // We currently only show the first failed recording.
+      // What we'd eally want to do is show all the recordings as a group
+      if (passed.length > 0 && failed.length > 0) {
+        flakyRecordings.push(recording);
+        flakySet.add(file);
+      }
```
bvaughn, line 50: Maybe this is the place we'd want to increment either the flaky, passed, or failed counts?

```js
if (passed.length > 0 && failed.length > 0) {
  flakyRecordings.push(recording);
  flakySet.add(file);
} else if (passed.length > 0) {
  passedSet.add(file);
} else {
  failedSet.add(file);
}
```

jasonLaster: This is a bit subtle, but for failed and passed i do want to use `.length` because it's fine to have multiple tests in different groups. Whereas for flakes, those are the same tests that are retried

src/ui/utils/testRuns.ts
jasonLaster: We definitely don't want to double count. I think I was missing an `else if`.

jasonLaster:
Should be ready for a second pass

src/ui/utils/testRuns.ts
bvaughn: Hmm, I hear what you're saying, but I think it's actually pretty confusing UX to count a test as both "flaky" and "failed" (at least in the summary view and/or for how you group it).

bvaughn: In the screenshot of this change, for example:
Screen Shot 2023-05-26 at 8 16 11 AM

The summary row shows a failed status (presumably because `failed.length > 0`) but it shouldn't. It should show a flaky color at worst.

src/ui/utils/testRuns.ts
```diff
@@ -0,0 +1,71 @@
...
+    const file = recording.metadata?.test?.file;
+    if (!file) {
+      return acc;
+    }
+    if (!acc[file]) {
+      acc[file] = 0;
+    }
+    acc[file]++;
+    return acc;
+  }, {} as Record<string, number>);
```
bvaughn, line 33: Performance nit: If you changed the structure of this from a file-to-count map to a file-to-array-of-recordings map, you could avoid having to loop over the entire recordings array to filter by file on line 42 (and just iterate over the map values instead).

bvaughn (commented):
I think the changes in this PR are a big improvement over what we have! 🥳 

I think there are just a few small rough edges (when contrasted to a more familiar UX like GitHub's integration with test runners):

Screen Shot 2023-05-26 at 8 39 15 AM

* On the left-hand side: A test run with no failures (only flakes and passed tests) should be shown as passing

* On the right-hand side: How we display flaky tests on the right:
   * The "failing" test in a flaky should should be styled with a _flaky_ color (not a failure color)
   * The "passing" test should be at the bottom, with the other passing tests (and its number should be counted

src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
```diff
@@ -73,7 +73,7 @@ function FilterByTestButton({ recording }: { recording: Recording }) {
   );
 }
 
-export function TestResultListItem({ recording }: { recording: Recording }) {
+export function TestResultListItem({ recording, label }: { recording: Recording; label: string }) {
```
bvaughn, line 76: Remove `label` prop. Looks like this PR adds it but nothing uses it.

src/ui/utils/testRuns.ts
jasonLaster: Good call. I think the code is simpler too

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -36,6 +36,8 @@ function TestStatusGroup({ recordings, label }: { recordings: Recording[]; label
     (a.metadata?.test?.file || 0) > (b.metadata?.test?.file || 0) ? 1 : -1
   );
 
+  console.log("TestStatusGroup", label);
```
bvaughn, line 39: Don't forget to remove this `console.log`

bvaughn (commented):
Exciting to see a green CI check 😄 

I think there's still an issue that we should address though: [Loom](https://www.loom.com/share/863783084e104a5a89e716eefe74574f)

bvaughn (commented):
Exciting to see a green CI check 😄 

I think there's still an issue that we should address though: [Loom](https://www.loom.com/share/863783084e104a5a89e716eefe74574f)

bvaughn:
I'm going to take a look at this because I'm curious.

src/ui/utils/testRuns.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  for (const file in recordingsMap) {
+    const recordings = recordingsMap[file];
+    if (recordings.length > 1) {
+      for (const recording of recordings) {
+        if (testPassed(recording)) {
+          passedRecordings.push(recording);
+        } else if (testFailed(recording)) {
+          flakyRecordings.push(recording);
+        }
```
bvaughn, line 42: This is why there's never any tests showing up as failing now.

bvaughn:
Note to self that our UI to show the number of comments is broken (not related to this PR just noting it)
Screen Shot 2023-05-30 at 10 12 09 AM

src/ui/utils/testRuns.ts
bvaughn: Fixed this in c25427d

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
bvaughn: Fixed this.

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -11,31 +11,33 @@ import styles from "../../../../Library.module.css";
 export function RunResults() {
   const testRun = useContext(TestRunOverviewContext).testRun!;
 
-  const sortedRecordings = orderBy(testRun.recordings, "date", "desc");
-  const passedRecordings = sortedRecordings.filter(r => r.metadata?.test?.result === "passed");
-  const failedRecordings = sortedRecordings.filter(
-    r => r.metadata?.test?.result && ["failed", "timedOut"].includes(r.metadata?.test?.result)
+  const { passedRecordings, failedRecordings, flakyRecordings } = useMemo(
+    () => groupRecordings(testRun.recordings),
+    [testRun.recordings]
```
bvaughn, line 16: `groupRecordings` isn't super expensive, but it's doing a lot of iterating. Seems like it would be nice if we only called it once, higher up, and passed the values down (rather than calling it in `RunResults` and `RunStats`– and also we'll need to call it in `TestRunListItem` too probably)

bvaughn:
Some nice things GitHub does that we don't (yet) do (but I think we should):
Screen Shot 2023-05-30 at 10 53 39 AM

1. Tests with flakes are shown at a high level (the `TestRunList`) as passing.
2. Retries of flaky (or failed) tests are grouped/collapsed in some way.

bvaughn:
I think this is an improvement, and we're probably blocked from doing anything more on the frontend alone, so let's land it and follow up with FE-1543

PR #9255, opened 2023-05-25
Fix `logpoints-02` save flakiness and clean up Playwright in-browser console logs

markerikson:
This PR:

- Tweaks our Playwright `debugPrint()` util to strip out ANSI formatting when logging messages in-browser, to keep the messages plain and readable
- Ensures that the logpoint panel Lexical typeahead gets hidden if we're not trying to submit an item, to avoid it overlapping the "Save" button and making `logpoints-02` flake

Saw the flakiness locally, seems much more consistent with the hide fix in place

### DIFF
--- a/packages/e2e-tests/helpers/source-panel.ts
+++ b/packages/e2e-tests/helpers/source-panel.ts
@@ -3,7 +3,7 @@ import chalk from "chalk";
 
 import { Badge } from "shared/client/types";
 
-import { type as typeLexical } from "./lexical";
+import { hideTypeAheadSuggestions, type as typeLexical } from "./lexical";
 import { findPoints, openPauseInformationPanel, removePoint } from "./pause-information-panel";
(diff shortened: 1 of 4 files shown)


packages/e2e-tests/helpers/lexical.ts
```diff
@@ -70,5 +70,9 @@ export async function type(page: Page, selector: string, text: string, shouldSub
   if (shouldSubmit) {
     await delay(200);
     await submitCurrentText(page, selector);
+  } else {
+    // The typeahead popup sometimes sticks around and overlaps the save button.
+    // Ensure it goes away.
+    await hideTypeAheadSuggestions(page, selector);
```
bvaughn, line 76: Wow! Could have sworn we had already added that but maybe I"m thinking of the screenshot tests.

markerikson: Looks like putting it here breaks two other logpoints tests.  Moving this to be up a level, right before we try to save.

PR #9259, opened 2023-05-26
Collect Cypress load times

jasonLaster:
FE-1362

I think we could do better here with the right abstraction for caches. Ideally, we would start collecting loading times for several caches like the recording cache and annotation cache as well.

### DIFF
--- a/src/ui/components/TestSuite/suspense/TestItemsCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestItemsCache.ts
@@ -15,6 +15,7 @@ import {
   ProcessedTestStep,
 } from "ui/components/TestSuite/types";
 import validateTestMetadata from "ui/components/TestSuite/utils/validateTestMetadata";
+import { pingTelemetry } from "ui/utils/replay-telemetry";
 
 // Processes TestItems into a format more usable by the TestSuite UI.
 // This cache depends on both Recording (from GraphQL) and annotations data (from the Replay protocol) and may load slowly.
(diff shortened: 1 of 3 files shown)


markerikson:
Where does any of this telemetry data live? How do we view it?

jasonLaster:
It goes into honeycomb and gets hydrated w/ useful user/workspace data. Ryan and i were discussing using this more to measure real user performance yesterday. That's why I added the pref to make it easier to use

src/ui/components/TestSuite/suspense/TestItemsCache.ts
```diff
@@ -128,6 +130,12 @@ export const TestItemsCache = createSingleEntryCache<
         scopePath,
       };
     });
+
+    pingTelemetry("TestItemsCacheLoad", {
```
jasonLaster, line 134: Ideally - this would be baked into caches so that you can just toggle telemetry on or off...

bvaughn: Okay, let's do that then. I'll post a follow up PR.

bvaughn: Maybe something like #9261?

bvaughn:
Closing in favor of a more generic approach (#9261) that will be easier to opt more caches into using.

bvaughn:
#9261 and #9262 add duration telemetry to all Test Suites caches.

bvaughn:
Looking at some preliminary data from Honeycomb that we're now logging– the `TestItemsCache` is taking quite a long time to load data.
Screen Shot 2023-05-26 at 5 56 25 PM

PR #9261, opened 2023-05-26
Suspense cache with telemetry

bvaughn:
Potential idea to expand on #9259

I don't know how we use this telemetry data though, so maybe there's a smarter way to approach this?

Note there are other cache types that this approach doesn't (yet) cover: streaming and interval


jasonLaster (approved):
I like this approach.

PR #9268, opened 2023-05-30
Don't use `loaded` regions to power the loading bar

jcmorrow:
This is a tiny change but has a pretty massive implication for the UI, so let's talk about why this might be a good idea.

Right now loading has three discrete phases from the devtools point of view:

- Creating a session (this is *usually* fast, but if we are currently scaling up it can take a while)
- Hoverboard (I think this blocks on having at least one paint?)
- Indexing/growing the process tree.

This makes no changes to the first two phases, but it almost completely gets rid of the third. I say almost completely because we will still wait for *some* region to get loaded, so that we can be sure that we have successfully downloaded snapshots. Now the user still might see a loading bar, but it will sit at 0% until we have finished RTP'ing to the beginning of the first segment.

Previously it was important to wait until things were loaded, because the backend might throw an error if the queried part of the recording was not loaded yet. Now, that should never happen as long as the request is within the load*ing* region (requests in the regions that are not even attempting to load have always errored and still will).

This *will* have a negative effect on some command times. However, the worst case scenario is that the user requests something in the region that takes the longest to load, and in that case the wait time is exactly as long as it would take us to get to 100%. This is a simple case of tail latency bias - blocking performance for *all* part of the recording gated by the *worst* parts of the recording.

This might have other implications I have not thought about (although it seems to work well locally), so I'm happy to discuss any part of this!

### DIFF
--- a/src/ui/reducers/app.ts
+++ b/src/ui/reducers/app.ts
@@ -222,11 +222,14 @@ export const getIndexedAndLoadedRegions = createSelector(getLoadedRegions, loade
   return overlap(loadedRegions.indexed, loadedRegions.loaded);
 });
 
-// Calculates the percentage of loading regions that have been loaded and indexed.
+// Returns 1 if we have loaded at least one section of the recording. This method has been changed
+// many times and this latest version is really a temporary change while the backend team decides
+// whether to pull out `indexed` and `loaded` events completely in exchange for something more
+// meaningful. We want the client to start sending requests earlier so that we can better gauge
+// how we should change our loading algorithms, and we also want users to feel free to interact
+// with the recording as long as we're somewhat close to being able to respond to their requests.
 //
-// For example:
-// If 80% of the regions have been indexed and 50% have been loaded, this method would return 0.65.
-// If 100% of the regions have been indexed and 50% have been loaded, this method would return 0.75.
+// See https://github.com/replayio/devtools/pull/9268 for more discussion.
 export const getIndexedProgress = createSelector(getLoadedRegions, regions => {
   if (!regions) {
     return 0;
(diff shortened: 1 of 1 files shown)


src/ui/reducers/app.ts
```diff
@@ -249,7 +249,7 @@ export const getIndexedProgress = createSelector(getLoadedRegions, regions => {
     return totalTime + end.time - begin.time;
   }, 0);
 
-  return totalIndexedTime / totalLoadingTime;
+  return totalIndexedTime ? 1 : 0;
```
jasonLaster, line 252: Can you add a comment

bvaughn: Agreed. There's a comment above, that no longer seems relevant:
https://github.com/replayio/devtools/blob/70e3c115ed1a8f6e29bb945137afdb1f4d29e0ef/src/ui/reducers/app.ts#L225-L229

bvaughn: nit: We could probably avoid some unnecessary iteration in this new approach, if we go with it.

```js
return indexed.find(({ begin, end }) => end.time - begin.time > 0) ? 1 : 0;
```

bvaughn:
Honestly, I don't know how to evaluate this change.

cweld510:
> Honestly, I don't know how to evaluate this change.

Another way to look at this change is as a hacky, simple, experimental version of a protocol change we are thinking of making. Long-term, we'd like to stop using `loadedRegions` events for anything except driving the orange progress bar in devtools, and ideally would remove them from the protocol. Instead, we'd like devtools and any other protocol client to consider the recording loaded and fully interactable as soon as the backend has allocated the resources necessary to serve requests in the focus window. As a result, our backend implementation will be less constrained since we won't have to send `loadedRegions` events at the right time, and perceived performance for the user should be better because they won't have to wait as long before the recording is loaded.

However, before committing to any protocol changes, we'd like to see what the UX would be like if we had made them. The above change will simulate the protocol changes - loading a single region is currently the most reliable indicator we have for determining whether or not the backend has allocated enough resources to serve requests in the focus window. Consider this experimental. We're trying to find out whether or not removing `loadedRegions` events would be viable.

I would personally love to make this change as soon as possible. The existence of `loadedRegions` events in the protocol is really constraining our implementation and mental model on the backend, and it's important to know whether or not removing them is a direction we can go in.

cweld510:
I'm happy to sync on this tomorrow, or write a document summarizing what we're trying to accomplish; whatever would be most helpful!

bvaughn:
Thanks for the more detailed explanation, Colin. I don't object to this change, but the code should have some kind of contextual comment I think.

It does look like a lot of Delta tests failed on this PR. Someone should dig into them to rule out whether this change _caused_ those failures or whether they're unrelated.

bvaughn (approved):
🎲🎲

PR #9272, opened 2023-05-31
Remove blocking on 'loaded' state of regions before sending protocol commands.

loganfsmyth:
In the same vein as #9268.

I'm not going to land this yet, because I still need to fix the API to make sure it actually behaves properly without the DevTools doing this, but all of these checks are things that I'd consider backend bugs.

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -156,8 +156,6 @@ export class ReplayClient implements ReplayClientInterface {
   async createPause(executionPoint: ExecutionPoint): Promise<createPauseResult> {
     const sessionId = this.getSessionIdThrows();
 
-    await this._waitForPointToBeLoaded(executionPoint);
-
     const response = await client.Session.createPause({ point: executionPoint }, sessionId);
 
     return response;
@@ -284,14 +282,6 @@ export class ReplayClient implements ReplayClientInterface {
   }> {
     const sessionId = this.getSessionIdThrows();
 
-    // It is important to wait until the range is fully loaded before requesting messages.
-    // It would be better if findMessagesInRange errored when the requested range could not be returned,
-    // or returned the boundaries of what it did successfully load (see BAC-2536),
-    // but right now it will just silently return a subset of messages.
-    // Given that we are extra careful here not to to fetch messages in unloaded regions
-    // because the result might be invalid (and may get cached by a Suspense caller).
-    await this._waitForRangeToBeLoaded(pointRange);
-
     const response = await client.Console.findMessagesInRange({ range: pointRange }, sessionId);
 
     // Messages aren't guaranteed to arrive sorted, but unsorted messages aren't that useful to work with.
@@ -338,12 +328,6 @@ export class ReplayClient implements ReplayClientInterface {
     if (!pointLimits.maxCount) {
       pointLimits.maxCount = MAX_POINTS_TO_FIND;
     }
-
-    await this._waitForRangeToBeLoaded(
-      pointLimits.begin && pointLimits.end
-        ? { begin: pointLimits.begin, end: pointLimits.end }
-        : null
-    );
 
     function onPoints(results: findPointsResults) {
       if (results.findPointsId === findPointsId) {
@@ -577,7 +561,6 @@ export class ReplayClient implements ReplayClientInterface {
     focusRange: PointRange | null
   ) {
     const sessionId = this.getSessionIdThrows();
-    await this._waitForRangeToBeLoaded(focusRange);
     const { hits } = await client.Debugger.getHitCounts(
       { sourceId, locations, maxHits: TOO_MANY_POINTS_TO_FIND, range: focusRange || undefined },
       sessionId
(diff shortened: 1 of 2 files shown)


bvaughn:
Tagging some frontend folks on this. We should be at least following along and kept up-to-date on any changes like this :)

packages/shared/client/ReplayClient.ts
```diff
@@ -284,14 +282,6 @@ export class ReplayClient implements ReplayClientInterface {
   }> {
     const sessionId = this.getSessionIdThrows();
 
-    // It is important to wait until the range is fully loaded before requesting messages.
-    // It would be better if findMessagesInRange errored when the requested range could not be returned,
-    // or returned the boundaries of what it did successfully load (see BAC-2536),
-    // but right now it will just silently return a subset of messages.
-    // Given that we are extra careful here not to to fetch messages in unloaded regions
-    // because the result might be invalid (and may get cached by a Suspense caller).
-    await this._waitForRangeToBeLoaded(pointRange);
```
bvaughn, line 293: If I'm understanding the backend semantics change correctly, I believe we'd want to replace this with a similar `waitForRangeToBeLoading` check?

loganfsmyth: > replace this with a similar waitForRangeToBeLoading check

You mean on the backend side? Yep! The fact that we aren't already doing this is a backend bug as far as I'm concerned.

loganfsmyth: I filed https://linear.app/replay/issue/BAC-3410/ensure-all-focus-window-commands-wait-for-the-windows-regions-to-be for this FYI

bvaughn: > You mean on the backend side?

Yes, but also on the frontend. If I'm understanding the more recent backend behavior, so long as a request falls within a region that's _loading_, the backend will not throw.

I think that means the frontend needs to block on making sure the request falls within a _loading_ region though (whereas before, it blocked for the region to be _loaded_).

loganfsmyth: > I think that means the frontend needs to block on making sure the request falls within a loading region

The frontend should make sure that commands that depend on the focus window aren't going to send commands for data outside the focus window, but that's it. The 'loading/indexed/loaded' state of a region should only be reflected in the timeline and have no other effect.

loganfsmyth:
The backend should now handle all these cases automatically so I pushed a new commit to trigger the tests again. They were all passing even before I deployed though so they aren't much of a positive signal in this case

packages/shared/client/ReplayClient.ts
bvaughn: Where'd we land on this thread?

loganfsmyth: There is no need for the frontend to wait for anything. If it requests a pause at a point, it'll wait for the focus window to be known and then either error because the point it outside the window, or it will create the pause.

bvaughn: The _wait_ was there to guard against race cases– where new data was requested in response to the same action/event that updated the focus window.

Feels like we're going in circles. I think we should just chat about this in realtime.

bvaughn: I'm not sure this specific guard would be relevant to that case, but here's an example of the type of race situation I'm thinking about:
https://discord.com/channels/779097926135054346/1115993272972746873

hbenl (approved):

packages/shared/client/ReplayClient.ts
bvaughn: > I think waiting for it to be loaded implies waiting for it to be loading

Not if I understand the change that's been made on the backend. The backend will wait on a request being made for an in-progress region finished loading, but it will throw if a request is made outside of a loading region (which I think makes sense from the backend's perspective).

I guess @loganfsmyth can confirm that.

bvaughn: > Why should the frontend use backend confirmations of the requests that it has sent instead of keeping track of what it has sent itself?

Because there's no other way to confirm that the backend has received and started processing the request. This isn't that different than what we were doing before (waiting on the backend to tell the frontend it had _loaded_ the range.)

bvaughn: FE-1564 might be an example of this (haven't dug in yet)

hbenl: > Because there's no other way to confirm that the backend has received and started processing the request.

I think we can assume that when we receive the response to `Session.requestFocusRange`, the backend has received and started processing the request.

bvaughn: Yes, we can assume that– but what's actually _waiting_ on that response (on the client) to avoid requesting things in the wrong order? Nothing, as a result of this PR.

This conversation feels pretty circular. Let's talk about it during sync today.

@loganfsmyth , please hold off merging this until Holger and I reach consensus.

hbenl:
Talked with @bvaughn and we decided to replace this PR with one that adds `ReplayClient.waitForRangeToBeRequested()`.

loganfsmyth:
Can you all tag me on the PR for this? I have specific goals with these changes and I'm going to want to help make sure any new changes actually achieve those goals. I think `waitForRangeToBeRequested` sounds like it's probably fine from what I'd guess the implementation is though. Sounds like this is getting at the concerns that drove my filing of https://linear.app/replay/issue/FE-1546/session-loadedregions-data-should-ideally-only-drive-the-timeline

PR #9292, opened 2023-06-01
Test Suites: Support v2 metadata format

bvaughn:
At a high level, this PR updates Test Suites views to use the newly proposed metadata schema (version 3). For background on this proposal, refer to [this Notion doc](https://www.notion.so/replayio/Test-Suites-Proposed-data-format-27ca1e6518af4ce0b77dad5e93f8b038).

At this time, the backend only _partially_ supports the new schema (version 2) so this PR also adds some utility methods (`migrateIncrementalGroupedTestCases` and `migrateIncrementalTestRecording`) to convert the metadata the backend is returning (2) to the one the views now use (3). A similar conversion method (`migrateLegacyTestRunToTestSuite`) is already being used for the Test Suites Library page. All three of the above conversion methods are located in a new package: `shared/test-suites`.

This PR deletes the old, no longer in use metadata types from `shared/graphql/types` and replaces them with new types in `shared/test-suites/types`.

A few follow up comments have been left in place for when the backend changes, linked to Linear issues SCS-1109 and FE-1543.

Note that this PR **should not be landed** until the backend has added the capability to filter which metadata schema the `GET_TEST_RUNS_FOR_WORKSPACE` query should return tests for. As it is currently, the frontend _cannot_ handle schema version 1 any longer. (Supporting both schemas would have been complex and not worth it, given the Test Suites pre-release state.)


packages/replay-next/src/suspense/NetworkRequestsCache.ts
```diff
@@ -0,0 +1,50 @@
...
+  requestInfo: RequestInfo[];
+  responseBodyData: ResponseBodyData[];
+  requestBodyData: RequestBodyData[];
+};
+
+export const networkRequestsCache = createCache<
+  [replayClient: ReplayClientInterface],
+  NetworkRequestsCacheData
+>({
+  debugLabel: "NetworkRequestsCache",
```
bvaughn, line 23: I needed network requests to be loaded into a Suspense cache so that I could fetch them in the test migration code. I've filed a follow up Linear issue to migrate our Network panel over to use this cache as well, and we can remove the redundant network data in Redux.

package.json
```diff
@@ -96,6 +97,7 @@
     "reactjs-popup": "^2.0.5",
     "replay-next": "workspace:*",
     "reselect": "^4.1.5",
+    "semver": "^7.5.1",
```
markerikson, line 100: :hand: Can we use the much smaller https://www.npmjs.com/package/compare-versions lib I used over in the React DevTools?

packages/replay-next/src/utils/array.ts
```diff
@@ -98,6 +99,92 @@ export function findInsertIndex<T>(
   return lowIndex;
 }
 
+export function findSliceIndices<ItemType, TargetType>(
```
markerikson, line 102: :question: what's all this used for?

packages/shared/test-suites/migrateIncrementalTestRecording.ts
```diff
@@ -0,0 +1,348 @@
...
+  isIncrementalIncrementalTestRecording,
+  isTest,
+} from "./types";
+
+export async function migrateIncrementalTestRecording(
+  testRecording: IncrementalTestRecording | TestRecording,
+  annotations: Annotation[],
+  replayClient: ReplayClientInterface
+): Promise<TestRecording> {
+  if (isIncrementalIncrementalTestRecording(testRecording)) {
```
markerikson, line 32: :question: does this really have "incremental" twice? :)

packages/shared/test-suites/types.ts
```diff
@@ -0,0 +1,301 @@
...
+  tests: IncrementalTestRecording[];
+};
+
+export type LegacyTestSuite =
+  | GetTestsRun_node_Workspace_testRuns
+  | GetTestsRunsForWorkspace_node_Workspace_testRuns;
+
+// Type checkers
+
+export function isIncrementalIncrementalTestRecording(
```
markerikson, line 239: :question: Is this meant to have "Incremental" twice?

packages/shared/test-suites/types.ts
```diff
@@ -0,0 +1,301 @@
...
+export type TestEvent = UserActionEvent | NavigationEvent | NetworkRequestEvent;
+
+export type TestError = {
+  name: string;
+  message: string;
+};
+
+// Intermediate types the frontend supports temporarily (until backend data has finished migrating)
+
+export type IncrementalTestRecording = Omit<TestRecording, "events" | "timeStampedPointRange"> & {
```
markerikson, line 205: :question: what makes this "Incremental"?  Is it that it's a type meant to be replaced later?

("Incremental" probably isn't the first name that would come to my mind, but not sure I immediately have a different suggestion. "Placeholder"? "Temporary"?)

src/ui/components/TestSuite/hooks/useShowUserActionEventBoundary.ts
```diff
@@ -0,0 +1,31 @@
...
+  userActionEvent,
+}: {
+  boundary: "before" | "after";
+  userActionEvent: UserActionEvent;
+}) {
+  const { executionPoint: currentPoint } = useContext(TimelineContext);
+
+  const dispatch = useAppDispatch();
+
+  const timeSTampedPoint =
```
markerikson, line 19: :hand: minor typo

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -0,0 +1,97 @@
...
+    let frames: Frame[] | undefined;
+    try {
+      frames = await framesCache.readAsync(client, endPauseId);
+    } catch (error) {
+      // TODO [FE-1555] RUN currently throws in some cases;
+      // degrade gracefully in this case by just disabling the details panel
+      console.error(error);
+    }
+
+    const callerFrameId = frames?.[1]?.frameId;
```
markerikson, line 38: :hand: would be nice to have a comment explaining _why_ we're looking at `[1]`.  I assume we're trying to go back before Cypress or something?

src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
```diff
@@ -1,73 +1,74 @@
...
+      const frames = await framesCache.readAsync(client, pauseId);
 
-            // and extract its sourceId
-            const markerSourceId = frames[markerFrameIndex]?.functionLocation?.[0].sourceId;
-            if (markerSourceId) {
-              // slice the frames from the current to the marker frame and reverse
-              // it so the user frames are on top
-              const userFrames = frames?.slice(0, markerFrameIndex).reverse();
+      if (frames) {
+        if (gte(runnerVersion, "8.0.0")) {
```
markerikson, line 31: :hand: could use `compare-versions` instead here

src/ui/components/TestSuite/suspense/TestSuiteCache.ts
```diff
@@ -0,0 +1,27 @@
+import assert from "assert";
```
markerikson, line 1: :question: is use of the `"assert"` lib intentional? thought we had a couple versions of that util in our codebase

src/ui/components/TestSuite/views/TestRecording/Panel.tsx
markerikson, line 1: :speech_balloon: It'd be nice to have a better file/component name than just "Panel" :)

src/ui/components/TestSuite/views/TetSuiteContext.tsx
markerikson, line 1: :hand: Typo in the filename

markerikson (commented):
Some questions and nits, but nothing blocking

markerikson:
Playing around with this, and I'm seeing a couple differences / potential regressions.

Looking at the Flake workspace and the `adding-spec.ts` test:

- On `main` (recording https://app.replay.io/recording/cypresse2eadding-spects--ea652b3d-76a4-46b4-a226-7d24c55aa14a ):
  - I can select individual steps and see the details pane update
  - There's 2 steps in the `beforeEach` section: `visit` and `title`
  - it shows 3 network request items in the list after the initial `get` step
  - the `eq` steps don't show the arguments (which seems like a bug in `main`)
  
image

- On `FE-1491` ( http://localhost:8080/recording/cypresse2eadding-spects--a8ebd71b-2acf-4c65-9c35-df110924c71c ):
  - selecting individual test steps is _not_ updating the details pane, which seems like a bug
  - there's a third step in the `beforeEach` section, `then null`
  - there are _no_ network requests shown in the step list
  - the `eq` steps _do_ have arguments shown (which seems to fix the bug from `main` )

image

bvaughn:
RE: Differences between main and this branch

The lack of step details sounds like something I should dig into.

The different number of things in the before vs main _could_ just be because all of the non-user-action events were _previously_ being dumped into main (even if they were actually in before/after). This created weird UX edge cases where you might select something in main and see only part of the before each steps highlighted but not the latter half.

packages/shared/test-suites/migrateIncrementalTestRecording.ts
bvaughn: Nah this was just something that went wonky during one of my find-and-replace renames, likely.

packages/replay-next/src/utils/array.ts
bvaughn: I wrote it to more efficiently splice out the network events that were relevant to a specific test, in a recording that contained multiple tests:
https://github.com/replayio/devtools/pull/9292/files#diff-de0bf61f1db9a5c0ca715ab5158845281026d0afd81cdced087f540e38337732R288-R324

package.json
bvaughn: 👍🏼  Sure.

packages/shared/test-suites/types.ts
bvaughn: This is not the final data schema I proposed. It's all that we had time to build before June 12th. So "incremental" indicates that it's...an in between step.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
bvaughn: I didn't write this. I just ported it over from the less usable format here:

https://github.com/replayio/devtools/blob/2406ee28d7aeca034dad142bfed72852d047f183/src/ui/components/TestSuite/suspense/consoleProps.ts#L22-L40

src/ui/components/TestSuite/hooks/useShowUserActionEventBoundary.ts
bvaughn: 👍🏼

src/ui/components/TestSuite/suspense/TestSuiteCache.ts
bvaughn: Uh, not _intentional_ I guess. Only that the`"assert"` lib was already installed and available so I used it. I don't think we have one that we should be using over another, necessarily, btu I'd be happy to see us do a follow up to consolidate!

src/ui/components/TestSuite/views/TestRecording/Panel.tsx
bvaughn: It's the "TestRecording/Panel" hehe

src/ui/components/TestSuite/views/TetSuiteContext.tsx
bvaughn: 👍🏼

bvaughn:
I think I've mostly addressed the PR feedback:
* Replaced "semver" with "compare-versions"
* Fixed various typos

I'll dig into the differences between this branch and main in the morning

src/ui/components/TestSuite/hooks/useShowUserActionEventBoundary.ts
bvaughn: 👍🏼

bvaughn:
I wrote a huge comment responding to Mark's feedback about the differences in those prod vs branch– and Chrome froze and I lost **all** of it.

tl;dr– the changes look like they make sense and the branch is more accurately representing the test metadata than production.

bvaughn:
As for verifying how we handle things–

Here's a recording with the new test metadata:
https://devtools-git-fe-1419-recordreplay.vercel.app//recording/video-comment-works-authenticated--d9493143-09d3-4254-aabb-2ad30d2f616e

Here's one with the old v1 test metadata format; it will look like a normal (non-test) recording:
https://devtools-git-fe-1419-recordreplay.vercel.app//recording/authenticatedlogpoints-01-shared-logpoints-functionality--00e5e4ae-a693-44af-b221-890ca3bafe14

Here's a regular (non-test) recording:
https://devtools-git-fe-1419-recordreplay.vercel.app//recording/cypresse2eadding-spects--a8ebd71b-2acf-4c65-9c35-df110924c71c

bvaughn:
GitHub PR isn't picking up my updates either :( Looks like another GitHub outage.

bvaughn:
This is irritating. Even closing and re-opening the PR doesn't pick up the latest value.

bvaughn:
I think this is ready to ship as soon as the backend is ready: https://github.com/replayio/replay-endpoints/pull/65 and https://github.com/replayio/backend/pull/7935

PR #9294, opened 2023-06-05
Use `hasAnnotationKind` to check for individual annotation types faster

markerikson:
This PR:

- Bumps the protocol package
- Adds `hasAnnotationKind()` to `ReplayClient`
- Swaps the annotation kinds cache to use that instead, so that we don't block on a check for _all_ annotations returning before we get _any_ of them back


markerikson:
Pushed a commit removing the `data/stack` fields, and it seems like it did actually build and deploy, but Vercel bot isn't updating the comment

hbenl (approved):

PR #9295, opened 2023-06-05
Add tests for Passport (FE-1494)

hbenl:


packages/e2e-tests/authenticated/passport-02.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  await openReactDevtoolsPanel(page);
+  await getReactComponents(page).nth(0).click();
+
+  await waitFor(async () =>
+    expect(await isPassportItemCompleted(page, "Inspect React components")).toBeTruthy()
+  );
+
+  // TODO add test for the "Jump to code" item.
+  // Currently we can't make example recordings containing mouse or keyboard events...
```
markerikson, line 61: :question: What's the problem here?

markerikson: I _think_ there's a chance they would?

The left-hand sidebar comes from `Session.findMouse/KeyboardEvents`.  I know those don't show up for Cypress recordings, but I _think_ they do for Playwright recordings, because we're telling a real browser to click?

markerikson:
`logpoints-08` seems flaky - saw it fail on #9294 too

PR #9296, opened 2023-06-05
Catch errors in consoleProps (FE-1555)

hbenl:
@bvaughn FE-1555 seems to be caused by a command error thrown by `framesCache.read()` in `getConsolePropsSuspense()`. I wasn't sure if this is the right place to handle errors from that function, I tried adding `<ErrorBoundary>`s at first but there are several places that use this and at some point I gave up finding them all.
I'm going to log off for today so feel free to merge if you approve.

### DIFF
--- a/src/ui/components/TestSuite/suspense/consoleProps.ts
+++ b/src/ui/components/TestSuite/suspense/consoleProps.ts
@@ -79,7 +79,11 @@ export function getConsolePropsSuspense(client: ReplayClientInterface, step: Ann
         }
       }
     }
-  } catch {}
+  } catch (error) {
+    // TODO [FE-1555] RUN currently throws in some cases;
+    // degrade gracefully in this case by just disabling the details panel
+    console.error(error);
+  }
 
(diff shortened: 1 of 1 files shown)


bvaughn (changes requested):
Hold off on this please. I rewrote this cache as part of FE-1419.

bvaughn:
@hbenl I dunno if `framesCache.read`  throwing is a backend issue or a frontend one or an expected case. Do you?

jasonLaster:
If I had to guess, `getAllFrames` is failing in the runtime which is something we'll need to fix too

bvaughn (commented):
If this is a really bad blocker, then I guess I shouldn't prevent it from landing– but I think we should at least:
1. add error logging (so it goes to Sentry)
2. file a RUN bug for follow up (so the underlying problem gets fixed)

I can deal with the rebase pain later, so long as it's only in this one place.

bvaughn (approved):

PR #9297, opened 2023-06-06
Truncating testsuite urls at 250 characters

jonbell-lot23:
**Summary**

In testsuites, it's possible for URLs to be extremely long, which makes the UX look funny. This change truncates things at 250 characters and adds a little tooltip if you need to see the whole thing.

image

Addresses https://linear.app/replay/issue/DES-690/we-should-truncate-long-urls-in-the-cy-panel

### DIFF
--- a/src/ui/components/TestSuite/views/TestItem/TestSteps/NavigationTestStepRow.tsx
+++ b/src/ui/components/TestSuite/views/TestItem/TestSteps/NavigationTestStepRow.tsx
@@ -9,10 +9,15 @@ export default memo(function NavigationTestStepRow({
 }: {
   navigationTestStep: NavigationTestStep;
 }) {
+  let url = navigationTestStep.data.url;
+  let displayUrl = url.length > 250 ? url.slice(0, 250) + "..." : url;
+
   return (
     <div className={styles.Indented}>
       <div className={styles.Text}>
-        <span className={styles.Name}>{navigationTestStep.data.url}</span>
+        <span className={styles.Name} title={url}>
+          {displayUrl}
+        </span>
       </div>
     </div>
   );


src/ui/components/TestSuite/views/TestItem/TestSteps/NavigationTestStepRow.tsx
```diff
@@ -9,10 +9,15 @@ export default memo(function NavigationTestStepRow({
 }: {
   navigationTestStep: NavigationTestStep;
 }) {
+  let url = navigationTestStep.data.url || "";
+  let displayUrl = url.length > 500 ? url.slice(0, 500) + "..." : url;
```
jasonLaster, line 13: I'm confused. Wouldn't this do 500 when the title says 250?

jonbell-lot23: Whoops, thanks. I committed two things:

1. Limit to 500, then change to 250
2. Check for null

When I did step 2, I mistakenly copied over the 500. Fixed now!

bvaughn (changes requested):
Hold off until #9292 has landed please 🙇🏼 

Alternately I could just pull this change into that PR.

bvaughn:
Regarding truncation, how would you feel about us using middle truncation instead, for this?

So given a URL like:
> https://app.replay.io/recording/cypressintegrationalllicensespects--fabcdce5-42d7-4408-8f75-1e8378f0982b?point=7788445296468053722493621463155185&time=4151&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6Ijc3ODg0NDUyOTY0NjgwNTM3MjI0OTM2MjE0NjMxNTUxODUiLCJ0aW1lIjo0MTUxfSwiZW5kIjp7InBvaW50IjoiNTU0OTI2NzI3NzEyMzA3Njg5NTUzNzA0NTQ4MzM0MzYyNDYiLCJ0aW1lIjoyODIyMn19

Rather than showing:
> https://app.replay.io/recording/cypressintegrationalllicensespects--fabcdce5-42d7-4408-8f75-1e8378f0982b?point=7788445296468053722493621463155185&time=4151&focusRegion=eyJiZWdpbiI6eyJwb2ludCI6Ijc3ODg0NDUyOTY0NjgwNTM3MjI0OTM2MjE0NjMxNTUxODUiLCJ0aW1..

We would show:
> https://app.replay.io/recording/cypressintegrationalllicensespects--fabcdce5-42d7-4408-8f75-1e8378f0982b?point=7788445296468...jE0NjMxNTUxODUiLCJ0aW1lIjo0MTUxfSwiZW5kIjp7InBvaW50IjoiNTU0OTI2NzI3NzEyMzA3Njg5NTUzNzA0NTQ4MzM0MzYyNDYiLCJ0aW1lIjoyODIyMn19

If you like this idea, we could implement it like:
```js
function truncateMiddle(string, maxLength = 250, separator = "...") {
  if (string.length > maxLength) {
    const charIndex = (maxLength - separator.length) / 2;
    return (
      string.slice(0, Math.ceil(charIndex)) +
      separator +
      string.slice(0 - Math.ceil(charIndex))
    );
  }

  return string;
}
```

jonbell-lot23:
Hurray! I am a huge fan of middle truncate for these scenarios. Will implement.

(I designed the first universal version of Office for mobile at Microsoft and middle truncate was deemed too hard because they'd have to implement from scratch. That meant you had a recents docs list on mobile that was basically meaningless because a lot of people put version numbers at the end of long file names. And wrapping didn't really work because filenames often get enormous. Meanwhile, my counterpart on the iOS version of Office could just use Apple's middle truncate because it was built in.)

jonbell-lot23:
Middle truncate has been added. No rush at all, we'll send this out when we're ready.

jonbell-lot23:
Update: #9292 was merged, but there's a small conflict I'll need to address before this is ready.

Reminder: this is a nice PR but of course not a rush. We'll ship when we're ready.

jonbell-lot23:
Turns out this merge conflict is pretty big:

* This file is gone
* My test URL doesn't show the Cypress panel anymore. I saw mention of this, but I'm not sure how to fix it. Will hold off to talk to Brian about next steps.

PR #9298, opened 2023-06-06
Add initial Test Suites E2E tests

markerikson:
This PR stacks on top of #9292 for rewriting the Test Suites data format and panel, and adds some initial E2E tests for the Cypress Test Panel:

- Basics
  - Test Suite panel
    - Panel loads and shows a list of tests with statuses
    - Should summarize total passed/failed/skipped
    - Should list metadata (PR, time, author)
  - Individual test
    - Shows before/after hooks
    - Shows test steps
    - Back button returns to the Test Suite panel
    - Error messages should show up

**UPDATE**

Additional changes:

- E2E tests for test step interactions:
  - Hovering test steps previews the point in time
  - Clicking a test step shows "Before/After" button, and they change time
  - Clicking a test step jumps to time; shows test source line if in DevTools mode, but not in Viewer
  - Clicking a test step of various types loads the step details / "props" section
  - "Jump to Code" button takes you to an app source file if available
  - Various hamburger menu items
- Updated Playwright to 1.35
- Fixed unrelated React warnings about unrecognized DOM attributes because of camel-cased `data-test-someName` attributes
- Deleted the original "jailed" Cypress test
- Added ignore globs to fix Playwright's recent change in "can't import test file A into test file B" errors for the tests in `replay-next/playwright` and their util files


packages/e2e-tests/jail/cypress-01.test.ts
```diff
@@ -13,7 +13,7 @@ import { waitFor } from "../helpers/utils";
 
 const url = "cypress/doc_inspector_styles";
 
-test(`cypress-01: Test basic cypress reporter functionality`, async ({ page }) => {
+test.skip(`cypress-01: Test basic cypress reporter functionality`, async ({ page }) => {
```
jasonLaster, line 16: why are we skipping this? do we want to re-add it eventually? or kill it?

packages/e2e-tests/helpers/index.ts
```diff
@@ -22,21 +22,37 @@ export async function openDevToolsTab(page: Page) {
...
+  const tab = getViewerTab(page);
+  const classes = await getElementClasses(tab);
+  return classes.includes("active");
+}
+
+export async function isDevToolsTabActive(page: Page) {
+  const tab = getDevToolsTab(page);
+  const classes = await getElementClasses(tab);
+  return classes.includes("active");
+}
```
bvaughn, line 54: Seems like these are working so 👍🏼 but if they ever appear flaky, we might have good results from migrating them to the "wait for" pattern (`await waitFor(async () => assertThatMightThrow())`

packages/e2e-tests/helpers/source-panel.ts
```diff
@@ -366,28 +374,32 @@ export async function verifyLogPointContentTypeAheadSuggestions(
...
 
-export async function getSelectedLineNumber(page: Page): Promise<number | null> {
-  let currentLineLocator = null;
+export async function getSelectedLineNumber(
+  page: Page,
+  lineIsCurrentPoint: boolean
+): Promise<number | null> {
+  const highlightLineTestName = lineIsCurrentPoint
+    ? "CurrentExecutionPointLineHighlight"
+    : "ViewSourceHighlight";
```
bvaughn, line 383: Ooo, this is a bit awkward huh.

I think maybe we should change the view to make tests easier to write:
```js
switch (mode) {
  case "search-result": {
    return (
      <div
        className={styles.CurrentSearchResult}
        data-test-name="CurrentLineHighlight"
        data-test-type="search-result"
      />
    );
  }
  case "view-source": {
    return (
      <div
        className={styles.ViewSourceHighlight}
        data-test-name="CurrentLineHighlight"
        data-test-type="view-source"
      />
    );
  }
}
```

What do you think?

packages/e2e-tests/helpers/testsuites.ts
```diff
@@ -1,25 +1,166 @@
...
+
+export function getTestStepBeforeAfterButtons(page: Page) {
+  return getByTestName(page, "TestBeforeAfter");
+}
+
+export function getUserActionEventDetails(page: Page) {
+  return getByTestName(page, "UserActionEventDetails");
+}
+
+export async function getKeyValueEntry(
```
bvaughn, line 107: nit (maybe for follow up, not blocking: Some of these seem like they may be more appropriately put in the object-inspector test helper

packages/e2e-tests/helpers/testsuites.ts
```diff
@@ -1,25 +1,166 @@
...
+  return getByTestName(page, "TestSuiteResultsPassedCount");
+}
+
+export function getTestSuiteResultsFailedCount(page: Page) {
+  return getByTestName(page, "TestSuiteResultsFailedCount");
+}
+
+export function getTestSuiteResultsSkippedCount(page: Page) {
+  return getByTestName(page, "TestSuiteResultsSkippedCount");
+}
```
bvaughn, line 70: IMO these methods should trim `textContent` and return numbers (or `null` if the label isn't visible)

packages/e2e-tests/helpers/testsuites.ts
```diff
@@ -1,25 +1,166 @@
...
 
-export async function getSelectedTestCase(row: Pick<Locator, "locator">) {
-  return row.locator('[data-test-id="TestItemPanelBody"]');
+export function getTestSuiteUser(page: Page) {
+  return getByTestName(page, "TestSuiteUser");
+}
+
+export function getTestSuiteDuration(page: Page) {
+  return getByTestName(page, "TestSuiteDuration");
+}
```
bvaughn, line 86: IMO this method should also trim and return a numeric value

packages/e2e-tests/tests/cypress-02_test-step-timelines.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const hoverPercent = await getTimelineCurrentHoverPercent(page);
+    hoverPercentages.push(hoverPercent);
+  }
+
+  // The first step is 0. All other steps should be greater than 0,
+  // and greater than the step percentage before it.
+  for (let i = 1; i < numSteps; i++) {
+    expect(hoverPercentages[i]).toBeGreaterThan(0);
+    expect(hoverPercentages[i]).toBeGreaterThan(hoverPercentages[i - 1]);
+  }
```
bvaughn, line 77: nit: This could be a single loop

packages/e2e-tests/examples.json
```diff
@@ -39,5 +39,7 @@
   "node/spawn.js": "c9d0539d-fde0-4949-a9f7-acdb5b56d4a6",
   "log_points_and_block_scope.html": "ad8f5490-0b9f-4488-bd3c-4f2824bc21ce",
   "cypress/doc_inspector_styles": "aa6201d2-1e83-408b-8247-cf3d26ac421a",
-  "doc_minified_chromium.html": "5cdb2ffe-beff-4d35-9689-9242bd277f2d"
+  "doc_minified_chromium.html": "5cdb2ffe-beff-4d35-9689-9242bd277f2d",
+  "cypress-realworld/bankaccounts.spec.js": "bbf9dc4f-9d87-4b7e-89e8-1b21e8570de4",
+  "flake/adding-spec.ts": "a8ebd71b-2acf-4c65-9c35-df110924c71c"
```
bvaughn, line 44: Where are the source files for these two new tests?

They should be in `public/test/examples/` so we can regenerate them if necessary.

bvaughn: Ah. Can you drop some documentation somewhere (inline in the examples JSON? in the README?) explaining how we are supposed to update these then? I wouldn't know.

packages/e2e-tests/helpers/source-panel.ts
bvaughn: I think this will make the helper and component easier to write future tests for.

markerikson:
Delta's busted due to Vercel being down

PR #9303, opened 2023-06-08
Use Debugger.newSources events if emitted.

loganfsmyth:
We're not emitting these yet, but we're looking to switch over to them so this patch aims to support both.

Fixes https://linear.app/replay/issue/BAC-3460/use-debuggernewsources-in-devtools-codebase


packages/shared/utils/testing.ts
```diff
@@ -126,6 +128,7 @@ export function createPointDescription({
 }
 
 export type SourceTuple = ["Debugger.newSource", newSource];
+export type SourcesTuple = ["Debugger.newSources", newSources];
 
 export function createSource({
```
loganfsmyth, line 133: @bvaughn This testing stuff is what I wanted to ask you about. I'm not familiar with this code, so while I've done my best to add all this in parallel with what was already here, I'm not sure if there's any issues we should worry about here for the test fixture data, and how we manage that as we swap the the API over to the new event type.

packages/shared/client/ReplayClient.ts
```diff
@@ -417,10 +418,17 @@ export class ReplayClient implements ReplayClientInterface {
...
       sources.push(source);
     };
+    const newSourcesListener = ({ sources }: newSources) => {
+      for (const source of sources) {
+        sources.push(source);
+      }
+    };
 
     client.Debugger.addNewSourceListener(newSourceListener);
+    client.Debugger.addNewSourcesListener(newSourcesListener);
```
bvaughn, line 428: What prevents us from adding every source twice? Is the expectation here that _only_ one of these two events will fire for any given session?

packages/shared/utils/testing.ts
bvaughn: As of #9107, there is no longer any fixture data being used by our screenshot tests.

packages/shared/client/ReplayClient.ts
loganfsmyth: Yeah the API wouldn't emit both, just one or the other. I'm going to delete the other listener as soon as things is deployed, I just have both in this transition period.

packages/shared/utils/testing.ts
loganfsmyth: I don't think I have enough context to know how that relates to the question. Does that mean I'm fine to land this as-is?

packages/shared/client/ReplayClient.ts
```diff
@@ -43,6 +43,7 @@ import {
   getTopFrameResult,
   keyboardEvents,
   navigationEvents,
+  newSources,
```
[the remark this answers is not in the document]
loganfsmyth: No argument there

packages/shared/client/ReplayClient.ts
markerikson: @loganfsmyth note that the routines logic depends on fetching sources as well

loganfsmyth: Yup that should all be updated with the backend conversion https://github.com/replayio/backend/pull/7961

loganfsmyth: Woops wrong PR, this one: https://github.com/replayio/backend/pull/7958

bvaughn (approved):

PR #9306, opened 2023-06-08
Remove the 'point' URL param usage for session creation.

loganfsmyth:
We are currently using this 'point' query param in https://github.com/replayio/devtools/blob/3f188fbc92355d05781d2c5a698a79f1ccd88782/src/ui/utils/environment.ts#L118 but we also read it in https://github.com/replayio/devtools/blob/3f188fbc92355d05781d2c5a698a79f1ccd88782/src/ui/actions/session.ts#L217 

I think this usage is a bit confusing since it means that if you were to view a recording, select an earlier point, and then refresh the page, you'd potentially be viewing a different length of the recording.

@bhackett1024 Would you're fine with us removing this for now? I think overloading this param like this makes the app behavior confusing in a few ways, so I'm in favor of just removing this for now and adding it again when we revisit this usecase.

Fixes https://linear.app/replay/issue/FE-1568/frontend-has-two-conflicting-uses-of-the-point-query-param

### DIFF
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -214,8 +214,6 @@ export function createSocket(
         experimentalSettings.controllerKey = String(Date.now());
       }
 
-      const loadPoint = new URL(window.location.href).searchParams.get("point") || undefined;
-
       const queuedActions: Array<Action | UIThunkAction> = [];
       let flushTimeoutId: NodeJS.Timeout | null = null;
 
@@ -246,7 +244,7 @@ export function createSocket(
 
       const sessionId = await createSession(
         recordingId,
-        loadPoint,
+        undefined,
         experimentalSettings,
         focusRange,
         {


bhackett1024 (approved):

jasonLaster:
I have not looked at the code, so I'm not sure what the trade offs are, but it's nice to be able to share a url and know the user will be paused at that point. 

Cc @bvaughn who's thought about this the most

loganfsmyth:
> it's nice to be able to share a url and know the user will be paused at that point.

That's not what this code is for.

jasonLaster:
Thanks for clarifying.

bvaughn:
> if you were to view a recording, select an earlier point, and then refresh the page, you'd potentially be viewing a different length of the recording.

I don't understand the written description of this issue, and FE-1568 doesn't provide any additional context.

Why would reloading/refreshing the page cause a mismatched behavior like you're describing? Both sections of code should be reading the same URL query param value, since both are called during initialization, right?

If I understand this change correctly, the only practical impact I would expect it to have would be to cause session creation to take a bit longer in some cases.

loganfsmyth:
> If I understand this change correctly, the only practical impact I would expect it to have would be to cause session creation to take a bit longer in some cases.

The `loadPoint` option is not related to setting any concept of a "current" point for the session, which I'd guess is what you're thinking it does.

What the option currently does from an implementation standpoint is that option sets is the the _endpoint_ of a recording session, if the session is still in the process of being uploaded. If the recording has finished uploading, it is entirely a noop. This behavior isn't something that's expected from the `point` query param, and isn't being relied on by anything, so I'd like to remove it. Setting the endpoint of a recording session seems like something that, if we were going to support it, should be a separate query param, and it probably shouldn't vary based on whether the recording is in the process of being uploaded or not, since the user has no insight into that backend state.

bvaughn:
>> If I understand this change correctly, the only practical impact I would expect it to have would be to cause session creation to take a bit longer in some cases.
>
> The loadPoint option is not related to setting any concept of a "current" point for the session, which I'd guess is what you're thinking it does.

My understanding of what this parameter does came from the protocol docs:
> **loadPoint**: Instead of waiting for the recording to be fully received, only wait until enough of the recording has been received that this point can be loaded.

I was just trying to understand what practical effect your change here would have on a user, because I still don't understand the wording of the PR description.

It sounds like I interpreted our docs incorrectly.

I don't think I know enough here to really have a strong opinion on this. If you think it's a good change, go for it. 👍🏼

bvaughn:
Closing this as it seems to have been dropped. If you want to re-open and land it though, it's not blocked from our side.

PR #9307, opened 2023-06-09
Wait for the focus range instead of loaded regions

hbenl:
Replaces #9272, addressing the concern described at the end of https://github.com/replayio/devtools/pull/9272#discussion_r1221696470

### DIFF
--- a/packages/shared/client/ReplayClient.ts
+++ b/packages/shared/client/ReplayClient.ts
@@ -58,10 +58,9 @@ import uniqueId from "lodash/uniqueId";
 import { addEventListener, client, initSocket, removeEventListener } from "protocol/socket";
 import { compareNumericStrings, defer, waitForTime } from "protocol/utils";
 import { initProtocolMessagesStore } from "replay-next/components/protocol/ProtocolMessagesStore";
-import { areRangesEqual } from "replay-next/src/utils/time";
 import { TOO_MANY_POINTS_TO_FIND } from "shared/constants";
 import { ProtocolError, commandError } from "shared/utils/error";
-import { isPointInRegions, isRangeInRegions, isTimeInRegions } from "shared/utils/time";
+import { isPointInRegion, isRangeInRegions } from "shared/utils/time";
 
 import { ReplayClientEvents, ReplayClientInterface, SourceLocationRange } from "./types";
 
@@ -82,6 +81,8 @@ export class ReplayClient implements ReplayClientInterface {
   private _sessionId: SessionId | null = null;
 
   private sessionWaiter = defer<SessionId>();
+
+  private focusRange: TimeStampedPointRange | null = null;
 
   private nextFindPointsId = 1;
   private nextRunEvaluationId = 1;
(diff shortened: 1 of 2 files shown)


packages/shared/client/ReplayClient.ts
```diff
@@ -877,75 +879,51 @@ export class ReplayClient implements ReplayClientInterface {
...
-        }
-      };
-
-      this.addEventListener("loadedRegionsChange", checkLoaded);
-
-      checkLoaded();
-    });
-  }
+    if (focusRange === null) {
+      return Promise.resolve();
```
bvaughn, line 907: This path seems wrong, at a glance.

I think it's relying on the fact that if there's no explicit focus range, the whole recording is "focused" – but that's not true for longer recordings.

I think a more robust fix would be to remove this special case, and update the `getFocusWindow` method to also dispatch the new '"focusRangeChange"` event?

bvaughn (commented):
Lots of e2e tests failing on this branch. Was the backend having troubles or does this indicate a legit regression?

packages/shared/client/ReplayClient.ts
hbenl: > This path seems wrong, at a glance.

This references the `focusRange` parameter instead of the current focus range (i.e. `this.focusRange`), so it's correct. I've renamed the `focusRange` parameter to `range` now to avoid that confusion.

> update the getFocusWindow method to also dispatch the new '"focusRangeChange"` event

Good point! I've added that and also call `getFocusWindow()` in `configure()` and `initialize()` to ensure we always set the initial focus range.

bvaughn (approved):
Cool 👍🏼

hbenl:
@loganfsmyth I already merged this because I was pretty sure this is what you meant when you said "waitForRangeToBeRequested sounds like it's probably fine from what I'd guess the implementation is".

packages/shared/client/ReplayClient.ts
```diff
@@ -290,7 +293,7 @@ export class ReplayClient implements ReplayClientInterface {
     // but right now it will just silently return a subset of messages.
     // Given that we are extra careful here not to to fetch messages in unloaded regions
     // because the result might be invalid (and may get cached by a Suspense caller).
```
loganfsmyth, line 295: This comment is stale now

packages/shared/client/ReplayClient.ts
```diff
@@ -877,75 +883,51 @@ export class ReplayClient implements ReplayClientInterface {
     }
   }
 
-  async _waitForPointToBeLoaded(point: ExecutionPoint): Promise<void> {
+  async waitForPointToBeInFocusRange(point: ExecutionPoint): Promise<void> {
```
loganfsmyth, line 886: Not specific to this PR, and you all may well agree since I'm sure changing this now would be hard, but the fact that this hangs indefinitely if the point isn't in range really seems dangerous to me from an API-design standpoint.

bvaughn: During the chat Holger and I had yesterday, the idea of adding a timeout to this came up. Guess that didn't make it into this PR but seems like it might be worth adding as a follow up.

PR #9308, opened 2023-06-09
Remove annotations from ThreadFront

hbenl:
This also contains fixes for some flaky screenshot tests.

### DIFF
--- a/packages/replay-next/playwright/tests/console/should-evaluate-terminal-expressions-at-an-execution-point.ts
+++ b/packages/replay-next/playwright/tests/console/should-evaluate-terminal-expressions-at-an-execution-point.ts
@@ -1,7 +1,7 @@
 import { test } from "@playwright/test";
 
 import { locateMessage, seekToMessage, toggleProtocolMessage } from "../utils/console";
-import { delay, takeScreenshot } from "../utils/general";
+import { takeScreenshot } from "../utils/general";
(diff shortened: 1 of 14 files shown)


packages/replay-next/playwright/tests/console/should-evaluate-terminal-expressions-at-an-execution-point.ts
```diff
@@ -15,7 +15,7 @@ test("should evaluate terminal expressions at an execution point", async ({ page
   await seekToMessage(page, listItem);
 
   await page.fill("[data-test-id=ConsoleTerminalInput]", "location.hre");
-  await delay(500); // HACK Give the type-ahead data time to load
+  await page.waitForSelector('[data-test-id="ConsoleTerminalInput-CodeTypeAhead"]');
```
bvaughn, line 18: Nice!

packages/shared/client/ReplayClient.ts
```diff
@@ -228,6 +237,13 @@ export class ReplayClient implements ReplayClientInterface {
     return sessionId;
   }
 
+  async findAnnotations(listener: AnnotationListener, kind: string) {
```
bvaughn, line 240: tiny nit: This feels like a backwards order of parameters for this sort of thing

bvaughn (approved):

PR #9309, opened 2023-06-09
Select network event request in Network table on click

bvaughn:
Also fix minor icon color issue on Library page


bvaughn:
This is a small change, and it fixes a (small) visual regression on the test suites page, so i'm going to land it– but I'll be happy to fix any PR notes that come up in a follow up.

PR #9311, opened 2023-06-09
Update playwright plugin to v1.0.1

ryanjduffy:


bvaughn (approved):
I guess this is all that's needed to get us using the new metadata format?

Are the test failures related to this upgrade or part of the intermittent backend failures?

bvaughn:
FWIW I ran one of the failing tests locally (auth comments 1) and it passed, so I think the CI failures are a strong signal in this case.

PR #9319, opened 2023-06-10
Add a nice (subtle) animation to our play button

jonbell-lot23:
Loom of the effect in action: https://www.loom.com/share/d27c6e2b03ea4c8482cf14c11d87a637

PR #9324, opened 2023-06-13
Add subtle animation to testsuites button

jonbell-lot23:
This is the same animation we just deployed on our play icon. ([See that PR here](https://github.com/replayio/devtools/pull/9319))

Here's how it looks in this PR:
buttonPop

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
@@ -1,3 +1,4 @@
+import { motion } from "framer-motion";
 import Link from "next/link";
 
 import Icon from "replay-next/components/Icon";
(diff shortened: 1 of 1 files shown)


src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
```diff
@@ -65,15 +66,22 @@ export function TestResultListItem({
...
+            <MaterialIcon iconSize="2xl" outlined style={{ color }}>
+              {["passed", "flaky"].includes(label) ? "play_circle" : "play_circle_filled"}
+            </MaterialIcon>
+          </motion.div>
         </div>
       </div>
       <div className="flex shrink grow flex-col truncate">
         <div className="block truncate">{title || "Test"}</div>
-        <div className="truncate truncate text-xs text-bodySubColor">{filePath}</div>
+        <div className="truncate text-xs text-bodySubColor">{filePath}</div>
```
bvaughn, line 84: Can we re-add the truncate style? I think keeping the rows uniform height goes a long way to making the page flow well.

jonbell-lot23: I reading on a small screen from bed so it's likely I'm misreading.

But I think truncate is still there? I think the diff shows it was written twice?

bvaughn: Ugh, sorry my bad. The string "truncate" _was_ deleted but only because there were two of them.

bvaughn (approved):

PR #9326, opened 2023-06-13
let users see public replays

jasonLaster:
Fixes FE-1576

### DIFF
--- a/pages/recording/[[...id]].tsx
+++ b/pages/recording/[[...id]].tsx
@@ -161,17 +161,20 @@ function RecordingPage({
       setRecording(rec);
 
       const isTestReplay = rec.metadata?.test && rec.metadata?.source;
+      const isTestWorkspace = rec.workspace?.isTest;
 
-      if (isTestReplay && !rec.workspace?.isTest) {
-        setExpectedError({
-          content: "This recording is not available.",
-          message: "The recording must belong to a test suite",
-        });
-      } else if (!isTestReplay && rec.workspace?.isTest) {
-        setExpectedError({
-          content: "This recording is not available.",
-          message: "The recording cannot be in a test suite",
-        });
+      if (rec.private) {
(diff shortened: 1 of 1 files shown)


pages/recording/[[...id]].tsx
```diff
@@ -161,13 +161,12 @@ function RecordingPage({
...
       const isTestReplay = rec.metadata?.test && rec.metadata?.source;
-
-      if (isTestReplay && !rec.workspace?.isTest) {
+      if (rec.private && isTestReplay && !rec.workspace?.isTest) {
         setExpectedError({
           content: "This recording is not available.",
           message: "The recording must belong to a test suite",
         });
-      } else if (!isTestReplay && rec.workspace?.isTest) {
+      } else if (rec.private && !isTestReplay && rec.workspace?.isTest) {
```
Comment (bvaughn) [inline pages/recording/[[...id]].tsx:169]: nit: I find these conditionals a bit harder to read than (I think) they should be. Maybe we could tidy this up a bit while we're in there?
```js
      if (rec.private) {
        const isTestReplay = rec.metadata?.test != null && rec.metadata?.source != null;
        const isTestWorkspace = rec.workspace?.isTest == true;

        if (isTestReplay !== isTestWorkspace) {
          setExpectedError({
            content: "This recording is not available.",
            message: isTestReplay
              ? "The recording must belong to a test suite"
              : "The recording cannot be in a test suite",
          });
        }
      }
```

Comment (jasonLaster) [reply pages/recording/[[...id]].tsx:169]: Yep - i think that's an improvement

markerikson:
Okay, confirmed that opening `a8ebd71b-2acf-4c65-9c35-df110924c71c` (a "Flake" suite test) in prod as an anonymous user does not let me see it, but https://devtools-git-featuregate2-recordreplay.vercel.app/recording/a8ebd71b-2acf-4c65-9c35-df110924c71c does.

PR #9328, opened 2023-06-13
Remove duration from test run list

ryanjduffy:
## Issue

The duration of a test run is always showing as `0 sec` because we're calculating off the duration of each recording and we're not returning the recordings with the test runs. We could but it slows the query and isn't adding much value.

## Resolution

Removes the duration from the list item but kept it in the detail view (which does have the recordings already).

## Additional Considerations

We could consider calculating the duration as replays are added to the test run like we do for result counts and that would avoid fetching the recordings just to get the duration. But, I'm not sure the duration is adding much value anyway so I'd suggest deferring that for now.


bvaughn (approved):
I wish we had a better way of detecting this case (rather than an empty `recordings` array?)

PR #9332, opened 2023-06-14
github: remove semgrep

jazzdan:


markerikson:
Certainly not _complaining_, but what prompted this change?

jazzdan:
@markerikson now that we're using socket, I don't think we need this. But if you think otherwise we should chat!

PR #9333, opened 2023-06-14
Show loading state for J2C buttons

markerikson:
This PR:

- Updates the Events sidebar and Cypress Panel to try to show a loading state for J2C buttons if we haven't received annotations yet

Honestly I don't think this is going to end up showing up most of the time, because of the timing of when we receive annotations for Cypress and mouse/keyboard events for Events anyway.  But I already had that loading state in the button component, just needed to pass in the right status flag.

### DIFF
--- a/src/ui/components/Events/Events.tsx
+++ b/src/ui/components/Events/Events.tsx
@@ -1,7 +1,7 @@
 import classNames from "classnames";
 import sortedLastIndex from "lodash/sortedLastIndex";
 import { memo, useContext, useMemo } from "react";
-import { useImperativeCacheValue } from "suspense";
+import { STATUS_PENDING, STATUS_RESOLVED, useImperativeCacheValue } from "suspense";
 
 import { getExecutionPoint } from "devtools/client/debugger/src/reducers/pause";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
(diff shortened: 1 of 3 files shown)


src/ui/components/Events/Events.tsx
```diff
@@ -97,6 +97,7 @@ function Events() {
                   currentTime={currentTime}
                   executionPoint={executionPoint!}
                   jumpToCodeAnnotation={jumpToCodeEntriesPerEvent[event.point]}
+                  jumpToCodeLoadingStatus={annotationsStatus === "resolved" ? "success" : "loading"}
```
bvaughn, line 100: Only checking `annotationsStatus === "resolved"` means that we might show "loading" even after the request _fails_.

I think you should swap this (and other places in the file where we have similar checks) to:
```js
annotationsStatus === STATUS_PENDING ? "loading" : "complete"
```

src/ui/components/Events/Events.tsx
```diff
@@ -97,6 +97,7 @@ function Events() {
                   currentTime={currentTime}
                   executionPoint={executionPoint!}
                   jumpToCodeAnnotation={jumpToCodeEntriesPerEvent[event.point]}
+                  jumpToCodeLoadingStatus={annotationsStatus === "pending" ? "loading" : "complete"}
```
bvaughn, line 100: I dunno that it's necessary (maybe TSC is smart enough to check these strings) but "suspense" does export these statuses as constants.

bvaughn (approved):

PR #9334, opened 2023-06-14
Filter Network requests with missing RequestInfo (aka no timestamp points

bvaughn:
Hopefully this won't be required, but pushing for discussion puropses.


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -572,8 +572,11 @@ export async function convertGroupedTestCases(
         const annotationsByTest: Annotation[][] = annotations.reduce(
           (accumulated: Annotation[][], annotation: Annotation) => {
             if (annotation.message.event === "test:start") {
+              // Start a new annotations array for each test
               accumulated.push([annotation]);
-            } else {
+            } else if (accumulated.length > 0) {
+              // Else add new annotations to the current test
+              // (Note some recordings have annotations data before the first "test:start")
```
bvaughn, line 579: This is not related to the rest of the PR, but while testing this change, I noticed a recording that had several annotations in the data _before_ the first `"test:start"` event– which is a case the previous code did not anticipate or handle.

bvaughn:
Going to land this because it fixes an active bug (bad annotation data) in Metabase's latest runs. Happy to make any changes suggested by PR review afterward though

PR #9337, opened 2023-06-15
Add ErrorBoundary to PreviewMarkers (FE-1571)

hbenl:
and allow a `null` fallback in our ErrorBoundary component.


jasonLaster:
What does this look like when it fails?

hbenl:
You see no preview markers in the timeline, as if there weren't any hits for the hovered line.

PR #9338, opened 2023-06-15
Move global ErrorBoundary to _app.tsx

hbenl:
We have 2 `ErrorBoundary` components, one of them shows the "unexpected error" screen and should only be used in `_app.tsx`. But I noticed that it was accidentally used `TestEventDetails.tsx`, so I moved it to `_app.tsx` (and don't export it).
Furthermore I removed some weird loading logic in `EditorPane.tsx` that was left over from codemirror times.


bvaughn (approved):
Nice

PR #9339, opened 2023-06-15
Shorter replay URLs

jaril:


bvaughn:
Seems like we haven't decided to do this so I'm going to close the PR. We can re-open with discussion if there's still support for it.

PR #9349, opened 2023-06-16
Log hit point loading exceptions to Sentry (but don't throw)

bvaughn:


hbenl (commented):
Why do we need to swallow these exceptions? There are only 2 places using this cache and they both handle exceptions (one with an `ErrorBoundary` and the other uses `useImperativeIntervalCacheValues()` which returns instead of rethrowing the exception).

bvaughn:
> The errors are already handled and don't show anything in case of an error, so this PR wouldn't change anything in the UI. However, if we want to show error states in the UI (e.g. changing the background color of the hit counts bar to red) then we should not swallow these errors.

I forced the cache to throw an error prior to making this change and confirmed that our UI did go down with an error boundary.

hbenl:
> I forced the cache to throw an error prior to making this change and confirmed that our UI did go down with an error boundary.

Interesting, that should not have happened anymore (after #9337). Can you make a recording?

bvaughn:
I rebased this branch this morning– and it seems like DEV mode does show an uncaught error dialog, but the source viewer remains useable. So I guess this change isn't needed.

PR #9353, opened 2023-06-18
Add a stress-test script for testing RecordingTestMetadata

jasonLaster:
### ⚠️ This PR includes lots of changes to our protocol that we probably won't want to merge

Just pushed a relatively simple test harness for RecordingTestMetadata. The goal is to be able to exercise the logic with a comprehensive set of real world recordings.

**Here's how it works**
* fetch recording test metadata
* call processGroupedTestCases with a replay client
* saves the metadata, annotations, and result to a fixture file
* on subsequent calls, reads the metadata and annotations from a file and compares the results

This is similar in many ways to our visual snapshot tests and is your basic contract test aka my favorite type of test.

**How to use it**
1. Get the support@replay.io api key ([docs](https://www.notion.so/replayio/Logging-in-as-support-replay-io-960f606f7d7b4b61872ddb6ce7003ced)) and add it to `.env` `STRESS_TEST_API_KEY=<api-key>`
2. run `ts-node packages/shared/test/stress-test.ts`
  2.a  run `ts-node packages/shared/test/stress-test.ts 1bf12a5d` (just one recoring) 
  2.b  run `ts-node packages/shared/test/stress-test.ts mnite` (just one workspace)

**Additional comments**
* [ ] I'd like to call the TestSuites cache directly, but I was not sure what i'd need to do to get GraphQL to work. The fact that I could pass in the replayClient was really nice.
* [ ] I'd like to find an alternative to writing fixture files because I don't want to have to check in all of these files so that they'd be available in CI.
* [ ] I'd like to find a good list of recordings to test with. My first thought is something like the cypress kitchen sink or realworld app could be good.
* [ ] It'd be nice to switch to Jest where we'd benefit from snapshot diffing, jest-cases, and the existing harness


jasonLaster:
There are a couple of commits in here where I made changes to RecordingTestMetadata to prevent it from throwing asserts. I don't plan on landing these. They're just here to get more interesting data for the time being to see what we're missing if we filtered out those the steps..

packages/protocol/socket.ts
```diff
@@ -148,6 +148,10 @@ type SessionCallbacks = {
 
 let gSessionCallbacks: SessionCallbacks | undefined;
 
+export function resetSession() {
+  gSessionCallbacks = undefined;
+}
+
```
jasonLaster, line 154: This was a hack needed to run multiple tests. There's probably a better way

packages/shared/client/ReplayClient.ts
```diff
@@ -244,6 +249,11 @@ export class ReplayClient implements ReplayClientInterface {
     return sessionId;
   }
 
+  releaseSession(): void {
+    client.Recording.releaseSession({ sessionId: this.getSessionIdThrows() });
+    resetSession();
+  }
```
jasonLaster, line 255: needed to test multiple recordings. not sure if this is the best approach or not

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -411,8 +411,10 @@ export async function processCypressTestRecording(
         }
 
         const annotations = userActionEventIdToAnnotations[id];
-
-        assert(annotations != null, `Missing annotations for test event (${command.name})`);
+        if (!annotations) {
+          // console.warn(`No annotations: ${id}`, command);
+          return;
+        }
```
jasonLaster, line 417: helpful for seeing what we'd miss if we didn't throw. I dont plan on landing this.

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -464,41 +466,45 @@ export async function processCypressTestRecording(
...
-          },
-          type: "user-action",
-        });
+            timeStampedPointRange: {
+              begin: beginPoint,
+              end: endPoint,
+            },
+            type: "user-action",
+          });
+        } else {
```
jasonLaster, line 495: same as above

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -578,9 +584,10 @@ export async function processGroupedTestCases(
             }
 
             return accumulated;
-          },
-          []
-        );
+          }, [])
+          .filter(annotations =>
+            annotations.find(annotation => annotation.message.event === "test:end")
+          );
```
jasonLaster, line 590: same as above

packages/shared/test/stress-test.ts
```diff
@@ -0,0 +1,176 @@
...
+    );
+
+    updateFixture(recordingId, { ...fixture, result });
+    await client.releaseSession();
+
+    if (fixture.result) {
+      return isEqual(fixture.result, result);
+    }
+
+    return true;
```
jasonLaster, line 141: it's nice that the test actually fits in a couple lines of code. The rest is just boilerplate

bvaughn:
Cleaning up our Pull Requests today. This is an old draft PR that seems to have been dropped, so I'm going to close it for now. Please re-open if that's a mistake.

PR #9355, opened 2023-06-19
Don't group annotations by test (FE-1601)

hbenl:
We relied on the `test:start` annotations to group the annotations by test. But these annotations are unreliable: they often appear *after* the first test step annotations for that test.
So I removed that grouping and also don't rely on `test:start`/`test:end` annotations anymore to find the point range for a test (I use the points of the first/last step annotations for that test instead).
This is just a quick fix that can be cleaned up later.

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -314,9 +314,6 @@ export async function processCypressTestRecording(
       main: [],
     };
 
-    let beginPoint: TimeStampedPoint | null = null;
-    let endPoint: TimeStampedPoint | null = null;
-
     const navigationEvents: RecordingTestMetadataV3.NavigationEvent[] = [];
 
     // Note that event annotations may be interleaved,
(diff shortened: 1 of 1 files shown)


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -426,6 +409,9 @@ export async function processCypressTestRecording(
         annotations.forEach(annotation => {
           switch (annotation.message.event) {
             case "step:end": {
+              if (!testEndPoint || comparePoints(testEndPoint.point, annotation.point) < 0) {
+                testEndPoint = { point: annotation.point, time: annotation.time };
```
jasonLaster, line 413: Two questions:
1. are these annotations sorted? Would we need a Max function...
2. Are these annotations grouped per test case? I assume so...

jasonLaster (commented):
Thanks.  Recorded two node replays of this PR in action.
1.  [metabase](https://app.replay.io/recording/metabase-ba4396b9--3ec4ebfb-65df-43c6-88d7-f3b73661e79b) which is missing a `step:end`
2. [wandb](https://app.replay.io/recording/wandb-flakes-panelbanksectionspects-7b188dd2--f55dfe50-a84e-4f69-b885-7a89cafc158f) which runs successfully

Will see if they can help me answer my questions

bvaughn (commented):
This was not a safe change. As the inline comments said, a single recording could contain multiple Cypress tests. _The only way_ the frontend knows where one test ends (and another begins) is the "test" annotations data.

If the step data is wrong (before the "test:start") then that's something we should have fixed on the plug-in side.

PR #9359, opened 2023-06-20
Add assertWithTelemetry for reporting to honeycomb

jasonLaster:
This will add the ability to slice and dice the errors we're seeing in the client in Honeycomb and see which issues are effecting which users.

### DIFF
--- a/packages/replay-next/src/utils/telemetry.ts
+++ b/packages/replay-next/src/utils/telemetry.ts
@@ -1,4 +1,13 @@
-export async function recordData(event: string, tags?: Object): Promise<void> {
+import assert from "assert";
+
+let defaultTags = {};
+export function setDefaultTags(tags: Object) {
+  defaultTags = tags;
+}
+
+export async function recordData(event: string, tags: Object = {}): Promise<void> {
+  const eventTags = { ...defaultTags, ...tags };
+
   if (process.env.NODE_ENV !== "development" || process.env.NEXT_PUBLIC_RECORD_REPLAY_TELEMETRY) {
     try {
       const response = await fetch("https://telemetry.replay.io/", {
@@ -8,7 +17,7 @@ export async function recordData(event: string, tags?: Object): Promise<void> {
         },
         body: JSON.stringify({
           event,
-          ...tags,
+          ...eventTags,
         }),
       });
 
@@ -19,4 +28,17 @@ export async function recordData(event: string, tags?: Object): Promise<void> {
       console.error("Telemetry request failed:", error);
     }
   }
+}
+
+export function assertWithTelemetry(
+  assertion: unknown,
+  message: string,
+  type: string,
+  tags: Object = {}
+): asserts assertion {
+  if (!assertion) {
+    recordData(type, tags);
+  }
+
+  return assert(assertion, message);
 }
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -1,4 +1,3 @@
-import assert from "assert";
 import {
   ExecutionPoint,
   RequestId,
@@ -10,11 +9,16 @@ import { satisfies } from "compare-versions";
 import { comparePoints } from "protocol/execution-point-utils";
 import { networkRequestsCache } from "replay-next/src/suspense/NetworkRequestsCache";
 import { findSliceIndices, insert } from "replay-next/src/utils/array";
+import { assertWithTelemetry } from "replay-next/src/utils/telemetry";
 import { ReplayClientInterface } from "shared/client/types";
 import { Annotation } from "shared/graphql/types";
 import { AnnotationsCache } from "ui/components/TestSuite/suspense/AnnotationsCache";
 
 export type SemVer = string;
+
+function assert(value: unknown, message: string, tags: Object = {}): asserts value {
+  return assertWithTelemetry(value, "process-test-metadata", message, tags);
+}
 
 // This type is only minimally supported by the frontend
 export namespace RecordingTestMetadataV1 {
@@ -414,9 +418,20 @@ export async function processCypressTestRecording(
             scope = null,
           } = partialTestEvent.data;
 
-          assert(category, `Test event must have "category" property`);
-          assert(command, `Test event must have "command" property`);
-          assert(id, `Test event must have "id" property`);
+          assert(category, `Test event must have "category" property`, {
+            command: command.name,
+            id,
+          });
+
+          assert(command, `Test event must have "command" property`, {
+            id,
+            category,
+          });
+
+          assert(id, `Test event must have "id" property`, {
+            command: command.name,
+            category,
+          });
 
           // The client does not show certain types of chained events in the list
           // they clutter without adding much value
@@ -430,7 +445,10 @@ export async function processCypressTestRecording(
 
           const annotations = userActionEventIdToAnnotations[id];
 
-          assert(annotations != null, `Missing annotations for test event (${command.name})`);
+          assert(annotations != null, `Missing annotations for test event`, {
+            command: command.name,
+            id,
+          });
 
           let beginPoint: TimeStampedPoint | null = null;
           let endPoint: TimeStampedPoint | null = null;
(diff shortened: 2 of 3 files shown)


packages/replay-next/src/utils/telemetry.ts
```diff
@@ -1,4 +1,15 @@
...
+let defaultTags = {};
+export function setDefaultTags(tags: Object) {
+  defaultTags = tags;
+}
+
+export async function recordData(event: string, tags: Object = {}): Promise<void> {
+  const eventTags = { ...defaultTags, ...tags };
+
+  console.log("recordData", event, eventTags);
+  // TODO: dont send telemetry in preview branches
```
bvaughn, line 12: This TODO comment should be removed?

src/ui/components/DevTools.tsx
```diff
@@ -216,6 +219,27 @@ function _DevTools({
...
+          userId: recording.user?.id,
+          workspace: recording?.workspace,
+          metadata: recording.metadata && {
+            testEnvironment: recording.metadata?.test?.environment,
+          },
+        },
+        user: {
+          id: userId,
+          email: userEmail,
+        },
```
bvaughn, line 238: nit: `user` is kind of ambiguous. (It confused me to see `recording.user` and `user` at first glance.)

Maybe we could name this `session` instead?

```suggestion
        session: {
          userId,
          userEmail,
        },
```

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -412,7 +416,9 @@ export async function processCypressTestRecording(
 
         const annotations = userActionEventIdToAnnotations[id];
 
-        assert(annotations != null, `Missing annotations for test event (${command.name})`);
+        assert(annotations != null, `Missing annotations for test event`, {
+          command: command.name,
+        });
```
bvaughn, line 421: We should probably add the id here too, as is done with the others.
```suggestion
        assert(annotations != null, `Missing annotations for test event`, {
          command: command.name,
          id
        });
```

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -464,14 +470,19 @@ export async function processCypressTestRecording(
...
         });
 
-        assert(beginPoint !== null, `Missing "step:start" annotation for test event ${id}`);
-        assert(endPoint !== null, `Missing "step:end" annotation for test event ${id}`);
-        assert(resultPoint !== null, `Missing "step:end" annotation for test event ${id}`);
+        assert(beginPoint !== null, `Missing "step:start" annotation for test event`, {
+          id,
+          command: command.name,
+        });
+        assert(endPoint !== null, `Missing "step:end" annotation for test event`);
```
bvaughn, line 477: This branch is stale. These two asserts don't exist anymore.

packages/replay-next/src/utils/telemetry.ts
```diff
@@ -20,3 +31,16 @@ export async function recordData(event: string, tags?: Object): Promise<void> {
     }
   }
 }
+
+export function assertWithTelemetry(
+  condition: unknown,
+  event: string,
```
bvaughn, line 37: What is `event` ? It seems significant but there's no explanation about what it is or how you would pick a meaningful name for it.

bvaughn (commented):
Generally I like this direction.

I am a bit confused though. I think that failed assertions already get logged to Sentry (as errors?) and this is just adding them to Honeycomb too?

I think we had also talked about adding some logging for "non fatal" data problems (ones that we could "degrade gracefully"– something I don't love doing). Do you have thoughts about how that would be structured?

src/ui/components/DevTools.tsx
```diff
@@ -216,6 +219,24 @@ function _DevTools({
...
+    if (!userLoading && recording) {
+      setDefaultTags({
+        recording: {
+          id: recording.id,
+          title: recording.title,
+          url: recording.url,
+          userId: recording.user?.id,
+          workspace: recording?.workspace,
+          metadata: recording.metadata && {
+            testEnvironment: recording.metadata?.test?.environment,
```
jasonLaster, line 232: Not sure why `test.environment` is not typed...

Here's the runtime value
```
{
    "errors": [],
    "testRunner": {
        "name": "cypress",
        "version": "12.8.1"
    },
    "pluginVersion": "1.0.1"
}
```

packages/replay-next/src/utils/telemetry.ts
jasonLaster: It's the assert message. What would help here?

jasonLaster:
> Is this just adding it to Honeycomb?

Yes, but that should be a lot more searchable/groupable/filterable etc

> Are we also going to do degredations
I think we should, but I haven't put that in yet.

packages/replay-next/src/utils/telemetry.ts
bvaughn: >> What is `event` ? 
>
> It's the assert message.

Down below it looks like you're passing some other value:

```js
return assertWithTelemetry(value, "process-test-metadata", message, tags);
```

But I'm not sure what the significance of it is. I'm guessing it's some kind of unique type/enum (maybe that's why you chose event?)

> What would help here?

Maybe we could re-order the args to be closer tot he familiar `assert` pattern? Maybe something like this?

```ts
export function assertWithTelemetry(
  assertion: unknown,
  message: string,
  type: string,
  tags: Object = {}
): asserts value {
  if (!assertion) {
    recordData(type, tags);
  }

  return assert(assertion, message);
}
```

"Type" is also kind of vague.

jasonLaster: agreed. will clean up tonight

bvaughn: Sure thing. Also feel free to hand off to me if you'd prefer.

src/ui/components/DevTools.tsx
bvaughn: It is typed, but there are multiple versions of the data that require refinement..

https://github.com/replayio/devtools/blob/01bda21af5a12dd5808ca0ca6b1d3b5f37876401/packages/shared/graphql/types.ts#L249C14-L253

https://github.com/replayio/devtools/blob/01bda21af5a12dd5808ca0ca6b1d3b5f37876401/packages/shared/test-suites/RecordingTestMetadata.ts#L288-L296

jasonLaster: Not sure how to work with this... My TSC are not the best

bvaughn: Metadata is stored differently depending on which version of the plug-in/backend stored it. Those types describe the various formats. If you try to access a property of one of the "any" types you might end up with `any`.

There are a bunch of helper functions further down that file for working with the different formats in a type-safe way:

https://github.com/replayio/devtools/blob/01bda21af5a12dd5808ca0ca6b1d3b5f37876401/packages/shared/test-suites/RecordingTestMetadata.ts#L833-L951C2

bvaughn (commented):
I'm confused. Does this PR _replace_ #9359 or...what?

jasonLaster:
#9392 adds the UI error telemetry on top

bvaughn:
Let me know once you've applied the feedback from this PR and I can review again.

jasonLaster:
@bvaughn i believe I have. Apologies if I missed something

src/ui/components/DevTools.tsx
```diff
@@ -216,6 +220,27 @@ function _DevTools({
     }
   }, [loadingFinished, trackLoadingIdleTime, sessionId]);
 
+  useEffect(() => {
+    if (!userLoading && recording) {
+      const test = recording.metadata?.test;
+      const testEnvironment = test && isGroupedTestCasesV3(test) ? test?.environment : undefined;
```
bvaughn, line 226: I don't think this would actually work currently, because the plug-in/backend are currently sending v2.

I think we should probably add the following to `RecordingTestMetadata`:
```ts
export type TestEnvironment = RecordingTestMetadataV3.GroupedTestCases["environment"];

export function getTestEnvironment(groupedTestCases: AnyGroupedTestCases): TestEnvironment | null {
  if (isGroupedTestCasesV1(groupedTestCases)) {
    return null;
  }
  return groupedTestCases.environment;
}
```

And then you could just use that here:
```js
  const testEnvironment = test ? getTestEnvironment(test) : null;
```

bvaughn (approved):
> @bvaughn i believe I have. Apologies if I missed something

I see. Looks like you have on this one, yup!

My confusion was that you haven't rebased the other one (which I thought was meant to _replace_ this one, since it includes it) https://github.com/replayio/devtools/pull/9392#discussion_r1245468207

I don't think reviewing stacked PRs like this works well in GitHub because it's difficult/cumbersome to rebase and update them all :)

I suggested one minor change but otherwise this looks good.

PR #9360, opened 2023-06-21
Adjust the testsuites error screen

jonbell-lot23:
Here's a Loom describing the rationale for this change:
https://www.loom.com/share/4316c60293214c559c358be25ba3f2c7

### DIFF
--- a/src/ui/components/TestSuite/views/GroupedTestCases/TestRecordingTree.module.css
+++ b/src/ui/components/TestSuite/views/GroupedTestCases/TestRecordingTree.module.css
@@ -18,4 +18,5 @@
   font-size: var(--font-size-medium);
   color: var(--theme-base-60);
   font-weight: bold;
+  text-transform: uppercase;
 }
--- a/src/ui/components/TestSuite/views/TestSuiteErrorFallback.module.css
+++ b/src/ui/components/TestSuite/views/TestSuiteErrorFallback.module.css
@@ -10,8 +10,9 @@
   flex: 0 0 auto;
   display: flex;
   align-items: center;
-  gap: 1ch;
+  gap: 0.75rem;
   padding: 0.75rem;
+  border-bottom: 1px solid var(--theme-border);
 }
 
 .Title {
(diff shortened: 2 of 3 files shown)


src/ui/components/TestSuite/views/GroupedTestCases/TestRecordingTree.module.css
```diff
@@ -18,4 +18,5 @@
   font-size: var(--font-size-medium);
   color: var(--theme-base-60);
   font-weight: bold;
+  text-transform: uppercase;
```
bvaughn, line 21: I don't like that we uppercase user-provided strings (in this case, the names they gave their `describe` blocks). I think it makes the names harder to read/recognize, and since they're user-provided, we should stick with the strings/text the user specified.

src/ui/components/TestSuite/views/TestSuiteErrorFallback.module.css
```diff
@@ -24,30 +25,25 @@
...
-}
-
-.ContactUsMessage {
-  padding: 0.75rem;
+  padding: 0.75rem; 
 }
 
 .Link {
   color: var(--link-color);
   text-decoration: none;
```
bvaughn, line 44: Remove this line
```suggestion
```

src/ui/components/TestSuite/views/GroupedTestCases/TestRecordingTree.module.css
jonbell-lot23: That's fine -- but the design was set differently, including throughout our Figma. When you propose visual changes, please remember to tell me to help design + code stay synced.

bvaughn: I don't think our current process really works that well. You work in Figma, which I rarely even check (or know how or where to check). In the case of Test Suites, it  was handed to me with minimal requirements or explanation (other than "it should look kind of like Cypress") and I kind of just made up the rest as I went along.

This is something we should talk about during the Test Suites post mortem / retro.

bvaughn: I'm feeling a huge amount of stress about the way this project has played out, in general. That spilled over into my response above. Your ask was very reasonable and I should have just said "okay" and left the rest for the retro.

I do think we there's room for improving our process. I do find Figma kind of confusing to navigate or search (and so often I don't even know if there is something I should reference there or not) but that's something we should talk about separately.

jonbell-lot23: I totally understand. It's been a hard slog in a lot of ways. Please don't feel any stress on my account -- simply tagging me on UI changes when you remember is all I need. But I also understand if/when you forget!

PR #9362, opened 2023-06-21
Update more components to use focus window instead of loaded regions

bvaughn:
Open question: Which of these is the better pattern to check with?
```js
// Pattern 1: Only passes if there is a focus region and an execution point and the point falls within the range
const condition = !focusWindow || !isPointInRegion(executionPoint, focusWindow);

// Pattern 2: Passes if there is no focus region or if the point falls within the focus range
const condition = focusWindow && !isPointInRegion(executionPoint, focusWindow);
```

I could see arguments for either. Maybe it's not that important, as `focusWindow` will only _briefly_ be `null` before the Replay Client is initialized/configured (during which it syncs the remote region). For now, I've decided to go with the _stricter_ option (1).

### Before

https://github.com/replayio/devtools/assets/29597/21f652c8-8af9-45df-8480-049bff90ec7e

### After

https://github.com/replayio/devtools/assets/29597/a7e284b6-e14a-4304-906a-4b6ec9401109

### Before

https://github.com/replayio/devtools/assets/29597/992f7915-c22e-410f-aee0-7165a4ccdbb6

### After

https://github.com/replayio/devtools/assets/29597/829bc42e-eb97-4820-89fe-14c0c1b62742

### Before

https://github.com/replayio/devtools/assets/29597/49600422-0e60-43f1-bc4f-1a125b310a94

### After

https://github.com/replayio/devtools/assets/29597/483ce0a0-6bae-489c-aa06-d279be493bfb

### DIFF
--- a/src/ui/components/NetworkMonitor/RequestTable.tsx
+++ b/src/ui/components/NetworkMonitor/RequestTable.tsx
@@ -69,7 +69,7 @@ const RequestTable = ({
             }
 
             const isInLoadedRegion =
-              !focusWindow || isPointInRegion(row.original.point.point, focusWindow);
+              focusWindow !== null && isPointInRegion(row.original.point.point, focusWindow);
 
             prepareRow(row);
 
--- a/src/ui/components/NetworkMonitor/RequestTable.tsx
+++ b/src/ui/components/NetworkMonitor/RequestTable.tsx
@@ -1,12 +1,11 @@
 import classNames from "classnames";
 import { Row, TableInstance } from "react-table";
 
+import { useCurrentFocusWindow } from "replay-next/src/hooks/useCurrentFocusWindow";
 import { useNag } from "replay-next/src/hooks/useNag";
 import { Nag } from "shared/graphql/types";
-import { getLoadedRegions } from "ui/reducers/app";
-import { useAppSelector } from "ui/setup/hooks";
+import { isPointInRegion } from "shared/utils/time";
 import { trackEvent } from "ui/utils/telemetry";
-import { isTimeInRegions } from "ui/utils/timeline";
 
 import { HeaderGroups } from "./HeaderGroups";
 import { RequestRow } from "./RequestRow";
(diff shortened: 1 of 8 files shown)


bvaughn:
The only two test failures on this run are known flakes with Linear issues filed.

bvaughn:
Most recent test run looked good except for the expected failed cypress-01 test. Now that it's been fixed in main, I rebased and pushed again.

src/ui/components/NetworkMonitor/RequestTable.tsx
```diff
@@ -69,9 +68,8 @@ const RequestTable = ({
               firstInFuture = true;
             }
 
-            const isInLoadedRegion = loadedRegions
-              ? isTimeInRegions(row.original.point.time, loadedRegions.loaded)
-              : false;
+            const isInLoadedRegion =
+              focusWindow !== null && isPointInRegion(row.original.point.point, focusWindow);
```
hbenl, line 72: why not use `useIsPointWithinFocusWindow` here?

hbenl (approved):
> Open question: Which of these is the better pattern to check with?
> [...]
> I could see arguments for either. Maybe it's not that important, as focusWindow will only briefly be null before the Replay Client is initialized/configured (during which it syncs the remote region). For now, I've decided to go with the stricter option (1).

I agree with all of that :shrug:

src/ui/components/NetworkMonitor/RequestTable.tsx
bvaughn: You can't call a hook within a loop (`rows.map` in this case)

We could move the is-in-loaded-region check into the `RequestRow` component (and maybe I will) but on first pass, I was just trying to minimize changes.

bvaughn:
Delta had one failure that seems unrelated to this change:
> throw Error("Timed out waiting for loaders to finish");

That one has failed a few times in recent runs, so I'm going to file an issue for it.

PR #9363, opened 2023-06-21
Add a Honeycomb button to Protocol Viewer

jcmorrow:
https://www.loom.com/share/4a1d8ba846d54780b7768ae7fb07c431

### DIFF
--- a/src/ui/components/ProtocolViewer.tsx
+++ b/src/ui/components/ProtocolViewer.tsx
@@ -161,6 +161,7 @@ function ProtocolRequestDetail({
   }
 
   let reportBugLink = null;
+  let honeycombQueryLink = null;
   // If this command failed, or is hung, we might want to report it to the backend team.
   if (isInternalUser) {
     if (error != null || response == null) {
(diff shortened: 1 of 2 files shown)


src/ui/components/ProtocolViewer.tsx
```diff
@@ -174,6 +175,25 @@ function ProtocolRequestDetail({
...
+          breakdowns: ["trace.trace_id"],
+          calculations: [{ op: "COUNT" }],
+          filters: [
+            { column: "name", op: "starts-with", value: "replayprotocol" },
+            { column: "replay.session.id", op: "=", value: window.sessionId },
+            { column: "rpc.replayprotocol.command_id", op: "=", value: request.id },
+          ],
+          filter_combination: "AND",
+          orders: [{ op: "COUNT", order: "descending" }],
+          havings: [],
```
jcmorrow, line 192: Not sure how much of this is actually necessary, but honeycomb includes it in their links, so.

bvaughn (commented):
I'm not sure how this is supposed to work. I tried it out with `listenForLoadChanges` (the only log entry that showed the button) and it loaded [a query](https://ui.honeycomb.io/replay/datasets/backend/result/EBbRHCDraaj) that returned 0 results. Is that expected?

jcmorrow:
> I'm not sure how this is supposed to work. I tried it out with `listenForLoadChanges` (the only log entry that showed the button) and it loaded [a query](https://ui.honeycomb.io/replay/datasets/backend/result/EBbRHCDraaj) that returned 0 results. Is that expected?

`listenForLoadChanges` is a request that we drop all samples of, because it does not give any information about the session, it just runs in the background listening for updates. What was happening when it errored? I guess it's reasonable if that message errors in response to a server-initiated session closure, but that is about the only case where I would maybe expect that method to error...

jcmorrow:
> I'm not sure how this is supposed to work.

The loom gives a pretty clear example I think.

jcmorrow:
I guess I'm just going to merge this, it's an internal-facing change and the E2E tests have all passed, although Delta has been running for 24 hours just about:

CleanShot 2023-06-23 at 07 18 40

PR #9365, opened 2023-06-21
Fix test date text assumption in cypress-01 test

markerikson:
This PR:

- Updates the `cypress-01` test to account for the "relative date since" logic falling back to just the `MM/DD/YYYY` string once it's >14 days since that date


bvaughn (approved):
¯\_(ツ)_/¯

PR #9368, opened 2023-06-22
Use focus window instead of loaded regions

bvaughn:
Update all of the "helper functions" from FE-1546 to use focus window instead of loaded regions:
* getIsInLoadedRegion
* jumpToClickEventFunctionLocation
* getAsyncParentPauseIdSuspense
* getPauseAndFrameId

As well as these components:
* PreviewPopup
* DefaultSelectedFrameContextAdapter
* NewFrames (PauseFrames and Frames sub-components)

And this hook:
* useCurrentPauseIdSuspense

### DIFF
--- a/packages/replay-next/components/console/ConsoleInput.tsx
+++ b/packages/replay-next/components/console/ConsoleInput.tsx
@@ -1,9 +1,4 @@
-import {
-  ExecutionPoint,
-  FrameId,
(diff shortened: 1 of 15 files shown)


bvaughn:
Seems like `node_stepping-01` may be a legit regression from this change. Going to investigate.

hbenl (approved):

PR #9370, opened 2023-06-22
Type user-action events to have null timestamps (for Playwright tests)

bvaughn:
Fixes a Sentry exception I saw logged this afternoon.


bvaughn:
The only failing e2e test now is `breakpoints-04`, which is known to be flaky (and has a Linear issue)

hbenl (approved):

PR #9372, opened 2023-06-23
Support new :id and :attempt Test Suite attributes

bvaughn:
Pairs with [SCS-1165](https://linear.app/replay/issue/SCS-1165/test-suites-support-retries-in-cypress). Should be ready to merge as soon as  https://github.com/replayio/replay-cli/pull/193 lands.

cc @jonbell-lot23 for visual treatment of retries:

Screen Shot 2023-06-23 at 12 03 20 PM

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -163,12 +163,20 @@ export namespace RecordingTestMetadataV3 {
   }
 
   export interface TestRecording {
+    // Useful for identifying retries of a failed test
+    attempt: number;
+
     // An error that occurred for this test that was unrelated to a specific event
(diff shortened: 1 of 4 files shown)


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -615,6 +625,13 @@ export async function processGroupedTestCases(
                     }
                   }
                 }
+
+                const currentTest = partialTestRecordings[currentTestRecordingIndex];
+                assert(
+                  currentTest?.id === annotation.message.testId,
+                  `Test id should match "test:start" annotation testId`
+                );
```
bvaughn, line 633: This isn't strictly _necessary_ but it verifies that things don't get out of sync.

ryanjduffy (approved):

PR #9373, opened 2023-06-23
Improve persistence for per-recording User preferences

bvaughn:
Replays with comments:
* Before: [858c9af5-b4cb-4ff0-8747-63aeaccb513d](https://app.replay.io/recording/fe-1606-recently-selected-tab-is-not-updated-before--858c9af5-b4cb-4ff0-8747-63aeaccb513d)
* After: [fdc666cf-cccd-4453-b187-07f4e8036db8](https://app.replay.io/recording/fe-1606-recently-selected-tab-is-updated-after--fdc666cf-cccd-4453-b187-07f4e8036db8)

Improve the strategy we use to remember the user's most recent view state for each recording by...
* Fixing bad in-memory cached current value
* Memoize to avoid over-updating
* Reduce the debounce delay to avoid missing updates (we could probably remove debounce entirely but I left some small amount in place)

### DIFF
--- a/src/ui/setup/prefs.ts
+++ b/src/ui/setup/prefs.ts
@@ -1,10 +1,12 @@
 import { RecordingId } from "@replayio/protocol";
+import equals from "fast-deep-equal";
 import debounce from "lodash/debounce";
 
 import type { PartialLocation } from "devtools/client/debugger/src/actions/sources";
 import { Tab, getTabs } from "devtools/client/debugger/src/reducers/tabs";
 import { prefs as debuggerPrefs } from "devtools/client/debugger/src/utils/prefs";
 import { persistTabs } from "devtools/client/debugger/src/utils/tabs";
+import { asyncStoreHelper } from "devtools/shared/async-store-helper";
 import { UIStore } from "ui/actions";
 import { getTheme } from "ui/reducers/app";
 import {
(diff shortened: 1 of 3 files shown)


src/ui/setup/prefs.ts
```diff
@@ -25,13 +27,55 @@ export interface ReplaySessions {
...
+    asyncStore.replaySessions = value;
+  }, 250);
+
+  return {
+    readAsyncStore,
+    writeAsyncStore,
+  };
+}
+
+const { readAsyncStore, writeAsyncStore } = initializeAsyncStore();
```
bvaughn, line 70: This ensures a single path to read/write the store, which makes sure that our cached "current" value is always correct. (Previously it was stale after the first update.)

src/ui/utils/prefs.ts
```diff
@@ -98,7 +98,3 @@ export const features = new PrefsHelper("devtools.features", {
 });
 
 export type Features = typeof features;
-
-export const asyncStore = asyncStoreHelper("devtools", {
-  replaySessions: ["Json", "replay-sessions", {} as Record<string, ReplaySession>],
-});
```
bvaughn, line 104: I hope to rewrite this stuff soon but for now, moving this inside of a protected block helps.

hbenl (approved):

PR #9376, opened 2023-06-23
Fix flaky breakpoints-04 e2e test

bvaughn:
Looking at one of the e2e test failures, Replay [bb6f7921-7ef1-4830-8f52-de0d810d6041](https://app.replay.io/recording/bb6f7921-7ef1-4830-8f52-de0d810d6041), shows an interesting thing– the test was clicking away from the "Pause Information" panel, to the "Sources" panel, seemingly while it was _also_ trying to verify the frames and scopes sub-panel contents.

I don't know why this is– (I don't see any missing `await` statements)– but I do know that clicking the "Sources" panel was not necessary. The `addBreakpoint` helper function only does this if the source isn't already open, but for some reason– various helper methods in `source-panel` explicitly called both methods every time. (I could visually confirm this while watching a test run locally.)

Removing the superfluous calls should be enough to fix the flaky behavior of this test, and possibly others.

That test pretty much always fails lately, so let's see what it does on this PR.

### DIFF
--- a/packages/e2e-tests/helpers/pause-information-panel.ts
+++ b/packages/e2e-tests/helpers/pause-information-panel.ts
@@ -110,7 +110,6 @@ export async function openPauseInformationPanel(page: Page): Promise<void> {
   const pane = getBreakpointsAccordionPane(page);
 
   let isVisible = await pane.isVisible();
-
   if (!isVisible) {
     await page.locator('[data-test-name="ToolbarButton-PauseInformation"]').click();
   }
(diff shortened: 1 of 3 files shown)


markerikson (approved):

bvaughn:
Hot dog. It only ran once, and passed.

https://app.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs/3580e494-6f39-4680-adb9-e86b94ccfe17

PR #9377, opened 2023-06-23
Use absolute dates for test panel relative date hover tooltip

markerikson:
This PR:

- Tweaks the hover tooltip for the Test Panel relative date display to show an absolute date instead of the exact same relative date

Before:

image

after:

image

### DIFF
--- a/src/ui/components/TestSuite/views/GroupedTestCases/Panel.tsx
+++ b/src/ui/components/TestSuite/views/GroupedTestCases/Panel.tsx
@@ -34,6 +34,7 @@ export default function Panel() {
   const testTree = useMemo(() => createTestTree(testRecordings), [testRecordings]);
 
   const recording = RecordingCache.read(recordingId);
+  const date = new Date(recording.date);
 
   return (
     <>
@@ -71,6 +72,7 @@ export default function Panel() {
             className={styles.Attribute}
             icon="schedule"
             label={getTruncatedRelativeDate(recording.date)}
+            title={`${date.toLocaleDateString()} ${date.toLocaleTimeString()}`}
             dataTestName="TestSuiteDate"
           />
           <Source />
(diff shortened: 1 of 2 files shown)


src/ui/components/TestSuite/views/GroupedTestCases/Panel.tsx
```diff
@@ -71,6 +71,7 @@ export default function Panel() {
             className={styles.Attribute}
             icon="schedule"
             label={getTruncatedRelativeDate(recording.date)}
+            title={recording.date}
```
bvaughn, line 74: This format isn't the easiest to read. Can we show a _formatted_ full date?

markerikson: yeah, how would you prefer it to be formatted?

bvaughn: Maybe?
```js
`${date.toLocaleDateString()} ${date.toLocaleTimeString()}`
```

> 6/23/2023 5:39:32 PM

markerikson: Yep, that works nicely:

image

bvaughn: Love it.

bvaughn (approved):

jasonLaster:
@markerikson small detail, but is there a nicer way to format this timestamp. `toLocaleStr()` or something...

bvaughn:
"nicer" how? What's wrong with the built-in localized date+time format?

jasonLaster:
Apologies for being vague. 

I was thinking of something like 

June 21st, 3:02pm

Where the date is formatted and the time is in the users timezone.

PR #9383, opened 2023-06-26
Fix the workspace check for test recordings (FE-1612)

hbenl:
Needs https://github.com/replayio/backend/pull/8086 to be deployed first.

### DIFF
--- a/packages/shared/graphql/generated/GetRecording.ts
+++ b/packages/shared/graphql/generated/GetRecording.ts
@@ -150,6 +150,8 @@ export interface GetRecording_recording {
   operations: any | null;
   resolution: any | null;
   metadata: any | null;
+  isTest: boolean;
+  isInTestWorkspace: boolean;
   comments: GetRecording_recording_comments[];
   activeSessions: GetRecording_recording_activeSessions[] | null;
   owner: GetRecording_recording_owner | null;
--- a/pages/recording/[[...id]].tsx
+++ b/pages/recording/[[...id]].tsx
@@ -158,24 +158,21 @@ function RecordingPage({
         return;
       }
 
-      setRecording(rec);
-
-      const isTestReplay = rec.metadata?.test && rec.metadata?.source;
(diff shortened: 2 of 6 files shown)


pages/recording/[[...id]].tsx
```diff
@@ -158,25 +158,22 @@ function RecordingPage({
...
+        setExpectedError({
+          content: "This recording is not available.",
+          message: "The recording must belong to a test suite",
+        });
+        return;
+      } else if (!rec.isTest && rec.isInTestWorkspace) {
+        setExpectedError({
+          content: "This recording is not available.",
+          message: "The recording cannot be in a test suite",
+        });
```
Comment (bvaughn) [inline pages/recording/[[...id]].tsx:171]: I don't think the frontend should be doing this check anymore either.

After BAC-3531, the backend should reject any attempts to open a test recording within a non-test workspace (and vice versa).

packages/shared/graphql/generated/GetRecording.ts
```diff
@@ -150,6 +150,8 @@ export interface GetRecording_recording {
   operations: any | null;
   resolution: any | null;
   metadata: any | null;
+  isTest: boolean;
```
bvaughn, line 153: I think the only thing the frontend should be using this new attribute to do is e.g. show the Test Suites side panel.

pages/recording/[[...id]].tsx
Comment (hbenl) [reply pages/recording/[[...id]].tsx:171]: The backend will refuse to create a session for test recordings in a non-test workspace and vice versa. But I think we should still keep the check here to show the user a more specific error message explaining why the recording can't be viewed.

Comment (bvaughn) [reply pages/recording/[[...id]].tsx:171]: I guess, but I would expect the backend to inform the frontend of what that message should be (with a new error code if needed). It seems like we might guess incorrectly, otherwise.

bvaughn (approved):
I guess I don't feel strongly about the comment I left on the page component, but it does seem like we run the risk of being out of sync.

PR #9391, opened 2023-06-28
Group cypress annotations by test id and attempt

ryanjduffy:
## Issue

Sometimes events will arrive outside of test start/end boundaries from the cypress plugin. The two common cases are:
* events from an `afterEach` handler added by app code that runs after the `afterEach` from the plugin
* failed assertions that arrive after the test has ended because those are handled via log output which can be deferred a bit.

## Resolution

Instead of fencing in the annotations by the test start/end, we can use the `testId` and `attempt` members on the annotation that together uniquely identify a test execution to group annotations by test.


bvaughn:
Cleaning up our Pull Requests today. This is an old draft PR that seems to have been dropped, so I'm going to close it for now. Please re-open and assign a reviewer(s) when it's ready for a look!

ryanjduffy:
It's still in progress but I can re-open it when it's ready to be discussed.

ryanjduffy:
I still need to do some more testing with this change but it checked out on the batch of customer replays shared on SCS-1170 so I think it's promising.

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -660,54 +671,19 @@ export async function processGroupedTestCases(
...
-              // (These are probably beforeAll annotations, which we don't fully support yet)
-              if (currentTestAnnotations) {
-                currentTestAnnotations.push(annotation);
+              const { testId, attempt } = annotation.message;
+
+              assert(testId != null, "Annotation is missing `testId`. Plugin update required.");
+              assert(attempt != null, "Annotation is missing `attempt`. Plugin update required.");
+
+              // beforeAll/afterAll have -1 testId and can be ignored for now
+              if (testId !== -1) {
```
bvaughn, line 680: Can we just _not_ store these events? Seems kind of yuck to have them included with a negative test id.

bvaughn: I guess I get it if they aren't associated with a test, but `null` id seems more expected then.

bvaughn:
I'm not sure what the state of this branch is. Are you hoping to get it actually landed or are you just testing things?

If the former, please let me know and I'll do an actual review.

ryanjduffy:
@bvaughn I am still testing this morning but the intent is to land this.

packages/shared/graphql/types.ts
```diff
@@ -246,7 +246,8 @@ export type CypressAnnotationMessage = {
   logVariable?: "cmd" | "log";
   id?: string;
   url?: string;
-  testId: number | string;
+  testId: number;
```
bvaughn, line 249: This is a nice simplification.

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -350,6 +350,21 @@ export async function processCypressTestRecording(
...
+              time: annotation.time,
+            };
+          }
+
+          if (!endPoint || comparePoints(endPoint.point, annotation.point) < 0) {
+            endPoint = {
+              point: annotation.point,
+              time: annotation.time,
+            };
+          }
```
bvaughn, line 366: This looks like a bit of a hack around bad data. Do we need to do this?

If we do, I'd love if we could add a follow up SCS issue (with a TODO comment referencing it) to remove it later.

packages/shared/test-suites/RecordingTestMetadata.ts
bvaughn, line 680: `testId` should be null for events that don't have a corresponding test. I think that's the idiomatic pattern for attributes like this (at least within JavaScript/web space).

packages/shared/test-suites/RecordingTestMetadata.ts
bvaughn: I think this may also be a bit flaky. If annotations are sorted by execution points, and we hit the `! beginPoint` case, when when we later hit the `"test:start"` case below, we would erase the `beginPoint` we'd set here.

```js
            case "test:start": {
              beginPoint = {
                point: annotation.point,
                time: annotation.time,
              };
              break;
            }
```

To handle that we'd need to also have a `comparePoints` check in `"test:start"`.

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: Maybe but it isn't now. I can update this to handle both with a comment to remove the`-1` handling later.

bvaughn: An anbalagous attribute would be `parentId`, which is `null` when there's no parent command.

> I can update this to handle both with a comment to remove the-1 handling later.

Okay. Let's file a Linear issue for follow up, and link the comment to it? :)

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: This is the crux of the problem. `test:start` and `test:end` annotations are not reliable at this time because the plugin cannot guarantee it runs before and after all other commands. I'll remove the `test:start` and `test:end` handling altogether to avoid confusion.

bvaughn: I'm just not clear why the plug-in can't do a similar type of detection/logic as the frontend is doing here. It collets these annotations, and (eventually ) saves them right? Why can't it do a check _before_ saving them.

I guess I'm concerned that our annotation data has known unreliable cases like this.

bvaughn: > I'll remove the test:start and test:end handling altogether to avoid confusion.

That does seem like it would be an improvement, since (I think) as things currently are– it would break.

ryanjduffy: The annotations are created as events happen in the browser. They're not collected and applied later so it's challenging to know that a new test has started and more events from a previous test aren't still going to come in.

bvaughn: Ah, I see. I'm not very knowledgeable about annotations stuff. I assumed there was a post-processing step, but I guess that's only true if we run a routine and generate derived annotations?

bvaughn (approved):

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: Perhaps? I'm not very knowledgeable about routine stuff :)

bvaughn: I think the high level way it works is that some type of annotation(s) get recorded while the recording is being made, and then a routine runs and loops over them, and stores derived annotations.

In our case, I guess that would mean the raw test+step events (and maybe navigation events) could get pre-processed and normalized before the frontend saw them.

That might be worth considering at some point anyway, for smoothing over differences between Cypress and Playwright. (Not sure how far @hbenl has thought through the Playwright side of this yet. I haven't really.)

ryanjduffy:
@bvaughn updating the two cypress e2e test replay ids here as well. One of the e2e tests is failing because it's on 1.0.0-ish plugin version.

PR #9392, opened 2023-06-28
Send honeycomb events when the error boundary is hit

jasonLaster:
This PR starts sending honeycomb `component-errorboundary` events when we show the error boundary. Better yet, we also include the command id, code, method, as well so that we can see what caused the error

### DIFF
--- a/packages/replay-next/components/ErrorBoundary.tsx
+++ b/packages/replay-next/components/ErrorBoundary.tsx
@@ -1,6 +1,8 @@
 import { captureException } from "@sentry/browser";
 import React, { Component, PropsWithChildren, ReactNode } from "react";
 
+import { recordData } from "replay-next/src/utils/telemetry";
+
 import styles from "./ErrorBoundary.module.css";
 
 type ErrorBoundaryState = {
@@ -11,6 +13,7 @@ type ErrorBoundaryState = {
 type ErrorBoundaryProps = PropsWithChildren<{
   fallback?: ReactNode;
   fallbackClassName?: string;
+  name?: string;
   /**
    * If `resetKey` changes after the ErrorBoundary caught an error, it will reset its state.
    * Use `resetKey` instead of `key` if you don't want the child components to be recreated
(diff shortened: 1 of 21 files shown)


packages/replay-next/src/utils/telemetry.ts
```diff
@@ -1,4 +1,13 @@
...
+import assert from "assert";
+
+let defaultTags = {};
+export function setDefaultTags(tags: Object) {
+  defaultTags = tags;
+}
+
+export async function recordData(event: string, tags: Object = {}): Promise<void> {
+  const eventTags = { ...defaultTags, ...tags };
+  console.log(":: recordData", event, eventTags);
```
bvaughn, line 10: Remove the log :)

bvaughn (commented):
I'm confused. Does this PR _replace_ #9359 or...what?

bvaughn:
This PR needs to be rebased on top of your branch `assert-with-telemetry` (or `main` after that branch is merged) and then I'll take another look.

packages/replay-next/components/ErrorBoundary.tsx
```diff
@@ -43,9 +46,16 @@ export default class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBo
     return null;
   }
 
-  componentDidCatch(error: Error) {
+  componentDidCatch(error: Error, info: React.ErrorInfo) {
     console.error(error);
 
+    recordData("component-errorboundary", {
+      component: this.props.name || "unknown",
```
bvaughn, line 53: nit: May be clearer to just log `undefined` or `null` if no name is specified?

packages/replay-next/components/ErrorBoundary.tsx
```diff
@@ -11,6 +13,7 @@ type ErrorBoundaryState = {
 type ErrorBoundaryProps = PropsWithChildren<{
   fallback?: ReactNode;
   fallbackClassName?: string;
+  name?: string;
```
bvaughn, line 16: Maybe we should just make this prop _required_? It's pretty helpful

packages/shared/utils/error.ts
```diff
@@ -24,10 +33,12 @@ export enum ProtocolError {
   FocusWindowChange = 76,
 }
 
-export const commandError = (message: string, code: number): CommandError => {
+export const commandError = (message: string, code: number, args?: CommandArgs): CommandError => {
   const err = new Error(message) as CommandError;
   err.name = "CommandError";
   err.code = code;
+  err.message = message;
```
bvaughn, line 40: Isn't this prop set automatically by `Error` constructor?

bvaughn (approved):

bvaughn:
I'm going to take this over because it seems like it's going to be dropped otherwise.

bvaughn:
Renamed this branch to reference the Linear issue (so it will do the linking). Re-opened as PR #9466.

PR #9397, opened 2023-06-28
Add jump to code to redux devtools

jasonLaster:
Relatively small PR that gets basic jump to code functionality working

### DIFF
--- a/src/ui/components/shared/JumpToCodeButton.tsx
+++ b/src/ui/components/shared/JumpToCodeButton.tsx
@@ -19,6 +19,7 @@ interface JumpToCodeButtonProps {
   onClick: () => void;
   status: JumpToCodeStatus;
   targetExecutionPoint: ExecutionPoint;
+  className: string;
 }
 
 export function JumpToCodeButton({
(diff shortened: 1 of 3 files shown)


src/ui/components/shared/JumpToCodeButton.tsx
```diff
@@ -27,13 +28,14 @@ export function JumpToCodeButton({
...
   targetExecutionPoint,
+  className = "",
 }: JumpToCodeButtonProps) {
   const [isHovered, setIsHovered] = useState(false);
 
   const jumpToCodeButtonAvailable = status === "not_checked" || status === "found";
 
   const jumpToCodeButtonClassname = classnames(
-    "transition-width flex items-center justify-center rounded-full  duration-100 ease-out h-6",
+    `jump-to-code transition-width flex items-center justify-center rounded-full  duration-100 ease-out h-6 ${className}`,
```
bvaughn, line 38: What is the `jump-to-code` class name supposed to accomplish?

bvaughn: Oh. I see. That's a nested selector style.

I think we should avoid those. Can we do that in a different way here @jasonLaster?

The way we normally do this is to pass in class names that point to CSS module styles.

bvaughn (commented):
I think this landed without review, but there's a style change I think we should make

PR #9400, opened 2023-06-29
Update dependencies

bvaughn:
- [x] Upgrade Babel and Prettier deps to support TypeScript "satisfies" operator
  - [x] Update Trunk to pull in newer Babel also
- [x] Use `@swc/jest` instead of `babel-jest` for unit tests (for the same reason as above)
- [x] Fixed an issue with async-store-helper test mocking


bvaughn:
@SocketSecurity ignore @swc/core@1.3.66

bvaughn:
Everything passed except breakpoints-04, which is known to be flaky

PR #9410, opened 2023-06-29
Analyze React+Redux perf timings using recording analysis

markerikson:
This PR:

- Adds a new "React+Redux Perf" sidebar panel
- Adds a button that, when clicked:
  - Finds all Redux dispatches
  - Finds all calls to `useSelector`
  - Finds all unique selector functions passed to `useSelector`
  - Finds all individual calls to each selector function and calculates their durations
  - Sums up the total execution time for each selector function
  - Finds all calls to notify React-Redux subscribers and checks the time spent notifying subscribers
  - Finds all calls to re-render a React component
  - Buckets the data by `dispatch()` start/end times, and groups all the selector calls, notification calls, and queued re-renders triggered by that dispatch

In other words, this is a major first step towards a useful React+Redux perf analysis tool.

Caveats:

- Right now I'm just logging the grouped data, and haven't tried to display or visualize any of this yet
- It currently takes ~30s to calculate the results for a 2s focus region with ~40 dispatches. The most expensive parts are running evaluations to find all the selector functions, and calling `Debugger.findStepOutTarget` to determine the execution time for a couple thousand function calls.

It's very likely that the execution speed would be faster if we moved this into a backend postprocessing routine, and also added some additional backend APIs to allow requesting step-out results in batches instead of individual requests.

Here's an example of what the output looks like for a given dispatch in our own front-end codebase:

image

That one dispatch:

- did queue a re-render (by scheduling one component to update)
- we spent 15.3ms notifying subscribers
- we cascaded through 10 nested `connect` components notifying their children
- we ran 114 separate selector calls
- of those, only 14 selector calls had a meaningful non-zero duration
- total time of selectors was 0.3ms, which means that there's 15.0ms of subscriber overhead (!)


ryota-murakami:
@markerikson [packages/replay-next](https://github.com/replayio/devtools/pull/9410/files#diff-50db83862d89c7c2c093e7f276ff579f064d404ec8d75d8cbed973babad64dcb) meant Next.js exclusive package?

markerikson:
@ryota-murakami : no. All our app source code originally lived in the `src` folder.  When Brian began rewriting some of the pieces using newer React techniques like Suspense, he created a `packages/bvaughn-architecture` folder and put the new code there. Later, we renamed it to `packages/replay-next`, to indicate that it's "the next generation of Replay app source code".

markerikson:
I just updated this to capture details on the start and end of the related React commit, plus counts of how many function and class components got rendered overall:

image

as well as inspecting the rendered components and determining the details for each unique component type:

image


This gives us a foundation for being able to do real-world React render performance analysis, on _production_ apps, no further profiling setup needed!

ryota-murakami:
@markerikson Great Job!

>  Brian began rewriting some of the pieces

Might be this tweet 😅 https://twitter.com/brian_d_vaughn/status/1674214956668014593

bvaughn:
Is this worth keeping open as a draft or was it just something you wanted to test/share?

markerikson:
Keep it as a draft for now. Like the other "React Render" panel, it's something I intend to merge and feature flag, but I need to do some cleanup before it's ready to at least have in the codebase.

markerikson:
Time to play with this again...

markerikson:
There's a lot of conceptual overlap between what I was doing in the first iteration of this "Redux+React Perf Profiler" PR, and what I did in [the original POC for the "React Render Timeline" panel](https://github.com/replayio/devtools/pull/8915).

Both rely on fetching hit points for internals. Both require fetching points for the current focus window. And my POC for the Redux perf work wanted to know when React had rendered, and when we _queued_ renders, so that we could know if the dispatch did actually result in components rendering.

Over this weekend, I've updated this PR:

- Rebased against `main`
- Extracted the basic Redux `dispatch()` checks and timings as an interval cache
- Made improvements to the function detail formatting logic
- And tonight I rewrote the React Render Timeline to also use an interval cache, use the new `Session.getPointStack` API for stack frames, and rewrote the items as a virtualized list.

The result is that the React Render Timeline is showing up _significantly_ faster (5-10s for several hundred hits), whereas before it would take seemingly _forever_ for values to slowly come in.

I don't have any UI for the Redux data yet, but that should be trivial to add now that I've got the cache for the values.

From there, the next step would be to start adding logic to query for perf stats related to a given dispatch: was a render queued, how long did it take, how many components rendered, how long did _those_ take, which ones were most expensive, and what selectors ran and _their_ timings.  Rather than fetch all of that up front, fetching it for just one render ought to be reasonably faster.

Here's what the render timeline looks like atm:

image

markerikson:
I've finally built out some UI for the Redux Dispatch Perf panel:

image

things I can now do with this info:

- set focus ranges from the timestamps
- determine if any React renders were queued by the notifications
- fetch info about selectors, although that gets trickier because I really need to know what selectors were in use _prior_ to the dispatch

and similarly, with React renders, I could pull together info on:

- which components rendered
- how long those each took

I _know_ there's all kinds of caveats on sync renders vs concurrent rendering, components being thrown away, etc. Not _anywhere_ close to being a "real" profiler.  But I've got demo code (even if commented out atm) that shows I can get plausible interpretations of that info.  So it's a starting point.

markerikson:
Tonight's progress: it's finding the likely start and end time of the next sync render right after the dispatch, and displaying that duration:

image

Also have it displaying how many components _probably_ were triggered to render as a direct result of the dispatch.

(Need to start extracting details on those components to confirm, but the poking around I've done says it's plausible)

markerikson:
Talked with Jon. He likes the general direction of the UI here, including reusing the "details pane locked to the bottom of the panel" approach that we're using on the Protocol Viewer and Test Suites panels.

We'll keep these two panels separate for now, but I'll update the Perf panel to have a Redux icon.

Longer-term, there's a lot of things we _could_ do here (maybe someday having a "popout" button that shows the list in a table format in the area where our source viewer currently lives).

markerikson:
I'm going to call this sufficiently ready that I'd like to get it merged in, behind the existing `feature_reactPanel` experimental flag.



### Current functionality

- React Queued Renders Panel
  - Fetches hits to `scheduleUpdateOnFiber` using an interval cache
  - Fetches stack frames via `getPointStack`, filters via URLs to find a likely candidate call to `setState()` or similar
  - Renders a virtualized list of function names, and has Jump to Code
- Redux Dispatch Perf Panel
  - Fetches hits to Redux `dispatch()` internals
  - Finds all likely Redux dispatches and extracts the action types (even if the Redux DevTools were not enabled)
  - Renders a virtualized list of action types, and has Jump to Code (yes this _highly_ overlaps with our actual Redux DevTools functionality - we'll sort that out later)
  - Calculates durations of the reducer, subscriber notifications, and total dispatch time
  - Uses `scheduleUpdateOnFiber` to estimate how many React components were likely queued to render by this dispatch (by checking `sUOF` calls inside of the subscriber notification start/end)
  - Tries to find a likely React render caused by this dispatch, and if found, displays the estimated total duration of that render

Both panels use Suspense to load, and `react-window` for the virtualized lists.  

### Screenshots

React Render Panel:

image


Redux Perf Panel:

image


### Future Work

- Counting how many function components and class components rendered in a given render pass
- Finding all Redux selectors used in an app
- Counting how many selectors got called during a set of subscriber notifications, and calculating their durations
- Showing list items for React render commits in the React panel (ie, alternating calls to queue a render vs commits)
- Showing the red "you are here" line in the lists


What I've got so far is good enough to at least have in the codebase behind a flag, and we can build on these from here.

jonbell-lot23 (approved):
Happy to do more visual polish later, but would you mind adding a pl-8 over here:

image

We want the "p" in pause/paused to be lined up underneth the disclosure triangle above it.

bvaughn:
Let's chat about this PR and the topic in person at the offsite this coming week?

markerikson:
Superseded by #9775 .  The Redux perf half went into #9776 , but we've decided not to merge that into the codebase for now.  I'll rebase it, then close the PR.

Issue #5186, opened 2022-01-31
Replay Check-in component


jonbell-lot23:
**Background**  
We'd like more signal about how well Replays are solving our users' needs.

**Initial insights**   
Check out [this Discord conversation](https://discord.com/channels/779097926135054346/937458413355532288) where I link to this initial video:
https://user-images.githubusercontent.com/9154902/151868701-40082893-c3c3-492d-9329-242656428b6b.mp4

Key insights/goals:
* The feature needs to benefit the user in a tangible way
*  Asking "how is Replay doing?" isn't super valuable to people
* But "mark as resolved" is a common pattern that can benefit users

**Design proposal**
By using a simple label at the top part of Replay Info, we can provide valuable information while feeling part of the UI:
![image](https://user-images.githubusercontent.com/9154902/151869024-dc6cf23a-0d8d-4222-b6fd-cd33e00b1299.png)

There could be a few options like so:
![image](https://user-images.githubusercontent.com/9154902/151869252-99d363c2-610c-430a-b02e-e6e649258a45.png)

Confetti can sprinkle down when you set something to active:
![image](https://user-images.githubusercontent.com/9154902/151869736-2f10d715-6ce5-42c5-86df-a460373220ec.png)

And after selecting, it can look like this:
![image](https://user-images.githubusercontent.com/9154902/151869312-aa41360d-0692-42bc-8a2d-a7e7c32f9c00.png)

Then the library could look like this:
![image](https://user-images.githubusercontent.com/9154902/151869788-9230b441-bef7-4304-8ee4-754f80f8eaf9.png)

**Status**
Jon's been working [on this branch](https://github.com/jonbell-lot23/devtools/tree/replay-checkin)

Working:
* Confetti animation is looking good
* Dropdown options are coming along

Needs work:
* The dropdown needs to be in the top header of the accordian
* We need to wire it up to track "status" as a property
* We need to make the confetti animation work, not sure how

jcmorrow:
Nice. I think there's also something about like, turning this replay into an Issue/PR/Card somewhere that feels likely relevant? Because in many ways Replay doesn't help you *resolve* your problem, it helps you *define* your problem, and the output of that is some artifact about *how* to resolve it. I don't know if that makes sense or not?

jonbell-lot23:
Thanks for the help today, @jcmorrow ! Here's a quick loom showing how things are working so far:
https://www.loom.com/share/460a787337f0433798376eff2181e73b

[...]

PR #9418, opened 2023-06-30
Removed unused legacy preference "resolveRecording"

bvaughn:
For consideration. I don't think this is still being used, but I could be wrong.

Seems like we should just delete this view, if it's not being used **or** we should delete the preference and always show the view. I don't think we should keep a feature flag like this around forever.

Related: #5186


markerikson (approved):

jasonLaster (approved):
I'm fine killing it. We can reconsider it later if we want to bring it back

jonbell-lot23 (approved):

PR #9442, opened 2023-07-03
Always show recording date, and show as its own row

markerikson:
This PR:

- Updates our "Info" panel to _always_ show the recording date (relative, or absolute for >14 days, same as the test suite info panel), and using the same icon as the test suite panel

Previously it was only showing the date if there was a `recording.user` field, and we don't always have a user (such as QA Wolf recordings).  It was also showing the date half-grayed-out.

Previous display with user:

image

New display with user:

image

New display with no user:

image

### DIFF
--- a/src/ui/components/Events/ReplayInfo.tsx
+++ b/src/ui/components/Events/ReplayInfo.tsx
@@ -4,10 +4,10 @@ import { ConnectedProps, connect } from "react-redux";
 import { OperationsData } from "shared/graphql/types";
 import { getDisplayedUrl } from "shared/utils/environment";
 import * as actions from "ui/actions/app";
+import { getTruncatedRelativeDate } from "ui/components/Library/Team/View/Recordings/RecordingListItem/RecordingListItem";
 import hooks from "ui/hooks";
 import { getRecordingTarget } from "ui/reducers/app";
 import { useAppSelector } from "ui/setup/hooks";
-import { formatRelativeTime } from "ui/utils/comments";
 import { getRecordingId, showDurationWarning } from "ui/utils/recording";
 import useAuth0 from "ui/utils/useAuth0";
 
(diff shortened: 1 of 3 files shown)


jonbell-lot23 (approved):
Looking at these screenshots, I think we should do this:

* When there's a person, don't show a new line. Do what we currently do on production.
* When there isn't a person (in testsuites or otherwise) show the whole new line

With that tweak in place, I think we're good to go.

jonbell-lot23:
Oh, and let's put the date higher in the list, so it'll either look like this:

* Name (with time/date)
* Sharing settings
* URL
* Sensitive data thing

Or like this:

* Date (since no name)
* Sharing settings
* URL
* Sensitive data thing

src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -89,6 +90,15 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
             </Row>
           ) : null}
         </div>
+        <Row>
+          <LabeledIcon
+            iconStyle={{ height: "1.25rem", width: "1.25rem", fontSize: "1.25rem" }}
+            icon="schedule"
+            label={getTruncatedRelativeDate(recording.date)}
```
bvaughn, line 97:
```suggestion
            label={time}
```

src/ui/components/Events/ReplayInfo.tsx
```diff
@@ -89,6 +90,15 @@ function ReplayInfo({ setModal }: PropsFromRedux) {
             </Row>
           ) : null}
         </div>
+        <Row>
+          <LabeledIcon
+            iconStyle={{ height: "1.25rem", width: "1.25rem", fontSize: "1.25rem" }}
```
bvaughn, line 95: Can we use a class name here instead of an inline style?

PR #9051, opened 2023-04-05
Replace runAnalysis with findPoints/runEvaluation (FE-1349)

hbenl:

PR #9446, opened 2023-07-04
Fix getter support in the object inspector

hbenl:


markerikson (approved):

jasonLaster:
Thanks for the test. How'd you find the issue?

hbenl:
> How'd you find the issue?

I noticed it recently while using Replay, then tracked it down to #9051 by bisecting and then made a recording to understand why the getter function was used as the value of the property.

PR #9447, opened 2023-07-05
Remove ThreadFront.evaluate

hbenl:
`ThreadFront.evaluate()` was used in 3 places:
- for `autocomplete-eager.ts` I used the `pauseEvaluationsCache`
- for react and redux devtools I used a copy of the `evaluate()` function that doesn't cache its results: here we're using evaluations to inject or communicate with the react/redux devtools backend, caching these evaluations didn't seem to make sense or may even cause bugs

I also removed the repaint-after-evaluation logic, it was broken anyway (it's meant for user evaluations in the replay console but the console didn't use `ThreadFront.evalute()` anymore). I've created FE-1654 for reimplementing it.

### DIFF
--- a/packages/protocol/thread/thread.ts
+++ b/packages/protocol/thread/thread.ts
@@ -7,19 +7,9 @@
 // performed on the state at different points in the recording. This layer
 // helps with adapting the devtools to the WRP.
 
-import {
-  ExecutionPoint,
-  Frame,
-  FrameId,
-  PauseId,
-  RecordingId,
-  SessionId,
-  Value,
-} from "@replayio/protocol";
+import { ExecutionPoint, Frame, PauseId, RecordingId, SessionId } from "@replayio/protocol";
 
-import { recordingCapabilitiesCache } from "replay-next/src/suspense/BuildIdCache";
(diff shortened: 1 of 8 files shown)


packages/protocol/thread/thread.ts
```diff
@@ -141,62 +131,6 @@ class _ThreadFront {
...
-      return { exception: null, returned: result.returned as unknown as Value };
-    } else if (result.exception) {
-      return { exception: result.exception as unknown as Value, returned: null };
-    } else {
-      return { exception: null, returned: null };
-    }
-  }
 }
 
 export const ThreadFront = new _ThreadFront();
```
bvaughn, line 136: This file is so small now! 🥳

bvaughn (approved):
I like this cleanup 👍🏼

packages/replay-next/src/utils/evaluate.ts
```diff
@@ -0,0 +1,43 @@
+import { FrameId, PauseId } from "@replayio/protocol";
+
+import { ThreadFront } from "protocol/thread/thread";
```
markerikson, line 3: :hand: Can we avoid continuing to reference `ThreadFront` here? I'd be fine with forcing us to pass in `pauseId` all the time.

PR #9451, opened 2023-07-05
Don't open test recordings or Library page in new tab

bvaughn:


src/ui/components/Library/Team/View/TestRuns/Overview/TestResultListItem.tsx
```diff
@@ -63,11 +62,9 @@ export function TestResultListItem({
   }
 
   return (
-    <Link
+    <a
```
bvaughn, line 65: It's important to use `<a>` here instead of `<Link>` or we won't reset Redux state correctly between recordings, which will result in a session error. See [efa80ab2-07b2-4823-abb6-343c8bfe03d0](https://app.replay.io/recording/replay--efa80ab2-07b2-4823-abb6-343c8bfe03d0) and [88819b65-5b8b-42a9-8873-ac8fce6027f1](https://app.replay.io/recording/replay--88819b65-5b8b-42a9-8873-ac8fce6027f1) where I dig into this.

markerikson: Makes sense - we want to reload the entire page with a full navigation in order to recreate the Redux store, vs a SPA router swap that would preserve the store.

markerikson (approved):

PR #9461, opened 2023-07-06
Missing Cypress plug-in (iteration)

bvaughn:
### [View without whitespace](https://github.com/replayio/devtools/pull/9461/files?w=1)

Change client-side detection of missing Cypress plug-in scenario a bit:
* Change test result to `"unknown"` (which also has the effect of skipping annotations parsing).
* Remove fake begin/end annotations in favor of just skipping the annotation processing. (This felt yuck before.)

Note this PR includes @ryanjduffy's fix from #9459 (SCS-1196) so that I could use the status `"unknown"` for these tests (rather than the less accurate `"skipped"`).

### Before
Screen Shot 2023-07-06 at 8 48 26 AM

### After
Screen Shot 2023-07-06 at 8 56 10 AM

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -332,239 +332,245 @@ export async function processCypressTestRecording(
     const navigationEvents: RecordingTestMetadataV3.NavigationEvent[] = [];
 
     // Skipped tests won't contain any annotations (include begin/end point)
-    if (result !== "skipped") {
-      // Note that event annotations may be interleaved,
-      // meaning that we can't step through both arrays in one pass.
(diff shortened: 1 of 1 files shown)


markerikson:
Oops wrong button, sorry :)

markerikson:
Did this change the assumptions in the Cypress E2E tests?

bvaughn:
> Did this change the assumptions in the Cypress E2E tests?

I did not expect it to, but it seems like it may have based on CI. I'll look into that before landing it.

bvaughn:
Looks like our Cypress e2e tests are triggering the new "missing plug-in" error. The reason for this is that I made a silly typo which I'll be correcting shortly.

Thank goodness fro e2e tests :)

PR #9463, opened 2023-07-07
Adjust testsuites dark theme

jonbell-lot23:
**New on left, old on right**

Zoomed out:
image

Zoomed in:
image

* Moved a bunch of tailwind classes into css modules
* Lightened and standardised the color on the secondary line of text

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -226,7 +226,7 @@
   --color-current: #f02d5e;
   --color-default: #ffffff;
   --color-dim: #aaaaaa;
-  --color-dimmer: #666666;
+  --color-dimmer: #5f6977;
   --color-error: #ffb3d2;
   --color-link: var(--color-brand);
   --color-primary-button: #ffffff;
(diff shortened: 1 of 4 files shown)


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-761/testsuites-dark-theme-needs-polish

bvaughn (approved):
I know it's subjective, but I prefer the slightly blueish gray we had before. It seemed like a better hue-match for the background. Any chance you'd consider a lighter version of that hue instead?

| Production | This branch | My suggestion |
| :---: | :---: | :---: |
| Screen Shot 2023-07-07 at 7 02 59 AM | Screen Shot 2023-07-07 at 7 03 14 AM | Screen Shot 2023-07-07 at 7 04 54 AM |

The color shown above is `#5f6977` fwiw.

jonbell-lot23:
I hear you, and I want to reconcile two things:

1. A nice hue (to your point)
2. Not one-offing colors

I'll need to follow this process:

1. Find all the places where var(--color-dimmed) is used
2. Confirm that your proposed color works for all of them
3. Ship it!

That way we get a nicer color ... but without it being a one-off. So I'll do that in the coming days/week.

jonbell-lot23:
Ok, #5f6977 seems to work, so I've made the switch and this PR is good to go!

PR #9464, opened 2023-07-07
Allow passing custom class to jump-to-code button

Mokshit06:
The typescript error was occurring cause `className` was a required property instead of optional.


jasonLaster (approved):

jasonLaster:
Landed in Jump to Code

PR #9467, opened 2023-07-07
Experimenting with scopes rendering for FE-1500

hbenl:
This PR is just an experiment to work on FE-1500.
STR:
- open http://localhost:8080/recording/react-app--cf8417fc-8c82-42df-8c56-94356a172522?point=324518553672943161447660830851097&time=239
- open the "Block" scope - this will take several seconds
- after that you can look at the `rendered` global in the browser console to see how often `Inspector` was rendered for each scope binding


bvaughn:
Just cleaning up some PRs.

PR #9468, opened 2023-07-07
Source line rendering changes [DRAFT]

bvaughn:
- [x] Remove hit point counts conditional. (These are now always-on.)
- [x] Add column highlighting for current execution point. (Mimic Chrome UI.)
- [x] Account for column breakpoint spacing

> **Note** I'd really like to remove column breakpoints UI from the client. I don't think they add much value, and I don't think we plan to encourage their use going forward (rather it would be nicer to reformat the source). As it is, they can really complicated UI features like this.

### Prior art (Chrome)
image

### Replay
Screen Shot 2023-07-07 at 3 33 57 PM

Screen Shot 2023-07-07 at 2 58 06 PM

Screen Shot 2023-07-07 at 3 23 18 PM

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/sources/CurrentLineHighlight.tsx
+++ b/packages/replay-next/components/sources/CurrentLineHighlight.tsx
@@ -100,10 +100,12 @@ function CurrentLineHighlightSuspends({
         columnBreakpointIndex = breakableColumnIndices.findIndex(
           column => column === highlightColumnBegin
         );
-        if (columnBreakpointIndex < breakableColumnIndices.length - 1) {
-          highlightColumnEnd = breakableColumnIndices[columnBreakpointIndex + 1] - 1;
-        } else if (plainText !== null) {
-          highlightColumnEnd = plainText.length - 1;
+        if (columnBreakpointIndex >= 0) {
+          if (columnBreakpointIndex < breakableColumnIndices.length - 1) {
+            highlightColumnEnd = breakableColumnIndices[columnBreakpointIndex + 1] - 1;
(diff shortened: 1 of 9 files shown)


packages/replay-next/components/sources/CurrentLineHighlight.tsx
```diff
@@ -82,15 +94,39 @@ function CurrentLineHighlightSuspends({ lineNumber, sourceId }: Props) {
           );
         }
       });
+      if (match != null) {
+        highlightColumnBegin = match.column;
+
+        columnBreakpointIndex = breakableColumnIndices.findIndex(
+          column => column === highlightColumnBegin
```
hbenl, line 101: I think it's possible to be paused at a location that isn't a breakable location...

hbenl (approved):

packages/replay-next/components/sources/CurrentLineHighlight.tsx
bvaughn: Oh?

markerikson:
Side note: not sure what "better to reformat the source" means in this context

bvaughn:
@markerikson: RE: nicer to reformat the source

@jasonLaster has talked about how column breakpoints are kind of awkward (e.g. you can't add print statements at a column breakpoint) and how it would be a nicer user experience if our source formatting just didn't allow multiple hits per line in the first place.

bvaughn:
Rebased this. Assuming CI is happy I'm going to land it.

I'm not sure I understand @hbenl's comment though.

PR #9469, opened 2023-07-07
Jump to correct point instead of internals

Mokshit06:
This PR fixes the issue in #9397 where the jump to code button would take to internal replay methods instead of user redux code.

### DIFF
--- a/src/ui/actions/eventListeners/eventListenerUtils.ts
+++ b/src/ui/actions/eventListeners/eventListenerUtils.ts
@@ -73,6 +73,11 @@ export const IGNORABLE_PARTIAL_SOURCE_URLS = [
   "__cypress/runner/",
 ];
 
+export const MORE_IGNORABLE_PARTIAL_URLS = IGNORABLE_PARTIAL_SOURCE_URLS.concat(
+  // Ignore _any_ 3rd-party package for now
+  "node_modules"
+);
+
 export function shouldIgnoreEventFromSource(
   sourceDetails?: SourceDetails,
   ignorableURLS = IGNORABLE_PARTIAL_SOURCE_URLS
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -24,6 +24,8 @@ import {
   pointsBoundingTimeCache,
   sessionEndPointCache,
 } from "replay-next/src/suspense/ExecutionPointsCache";
+import { frameStepsCache } from "replay-next/src/suspense/FrameStepsCache";
+import { pauseIdCache } from "replay-next/src/suspense/PauseCache";
 import { screenshotCache } from "replay-next/src/suspense/ScreenshotCache";
 import {
   isTimeStampedPointRangeEqual,
@@ -37,7 +39,9 @@ import {
   isTest,
   updateUrlWithParams,
 } from "shared/utils/environment";
+import { MORE_IGNORABLE_PARTIAL_URLS } from "ui/actions/eventListeners/eventListenerUtils";
 import { getFirstComment } from "ui/hooks/comments/comments";
+import { SourcesState } from "ui/reducers/sources";
 import {
   getCurrentTime,
   getFocusWindow,
@@ -55,6 +59,13 @@ import {
   setPlaybackPrecachedTime,
 } from "ui/reducers/timeline";
 import { FocusWindow, HoveredItem, PlaybackOptions, TimeRange } from "ui/state/timeline";
+import { getPauseFramesAsync } from "ui/suspense/frameCache";
+import {
+  encodeObjectToURL,
+  getPausePointParams,
+  isTest,
+  updateUrlWithParams,
+} from "ui/utils/environment";
 import KeyShortcuts, { isEditableElement } from "ui/utils/key-shortcuts";
 import { trackEvent } from "ui/utils/telemetry";
 import { rangeForFocusWindow } from "ui/utils/timeline";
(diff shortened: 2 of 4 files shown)


src/ui/actions/eventListeners/eventListenerUtils.ts
```diff
@@ -73,6 +73,11 @@ export const IGNORABLE_PARTIAL_SOURCE_URLS = [
   "__cypress/runner/",
 ];
 
+export const MORE_IGNORABLE_PARTIAL_URLS = IGNORABLE_PARTIAL_SOURCE_URLS.concat(
+  // Ignore _any_ 3rd-party package for now
+  "node_modules"
```
jasonLaster, line 78: we might want to add the `react-redux` internal url in there too

src/ui/actions/timeline.ts
```diff
@@ -264,6 +269,44 @@ export function seek(
   };
 }
 
+export function jumpToReduxCode(
```
jasonLaster, line 272: I'm not sure if this makes more sense as a redux action or as a suspense cache that returns the frameStep that we can call `seek` on

src/ui/actions/eventListeners/eventListenerUtils.ts
Mokshit06: we can do that, and then change `filteredPauseFrames[2]` to `filteredPauseFrames[0]`

bvaughn:
There's not a lot of code here, (so in that sense it should be easy to review) but given the complexity of the feature itself, we should really have some end-to-end test coverage for Redux DevTools. (I'm not saying they need to be added with this PR. That would expand scope significantly. I'd like to see them added soon after though.)

jasonLaster:
re e2e tests: I think the first test should be its own issue (it's already in linear) and then from there we can add more coverage with each feature.

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -68,9 +68,10 @@ function ActionItem({
   selectedPoint: ExecutionPoint | null;
   setSelectedPoint: (point: ExecutionPoint | null) => void;
 }) {
+  const sourcesState = useAppSelector(state => state.sources);
```
markerikson, line 71: :hand: This should be moved into the thunk as a `getState()` - the UI doesn't need this part

src/ui/actions/timeline.ts
```diff
@@ -264,6 +269,44 @@ export function seek(
...
+  point: ExecutionPoint,
+  time: number,
+  sourcesState: SourcesState
+): UIThunkAction {
+  return async (dispatch, _, { replayClient }) => {
+    dispatch(framePositionsCleared());
```
markerikson, line 278: :question: Why do we need `framePositionsCleared` dispatched here?

src/ui/actions/timeline.ts
```diff
@@ -264,6 +269,44 @@ export function seek(
...
+    const frameSteps = await frameStepsCache.readAsync(
+      replayClient,
+      pauseId,
+      // The first 2 elements in filtered pause frames are from replay's redux stub, so they should be ignored
+      // The 3rd element is the user function that calls it, and will most likely be the `dispatch` call
+      filteredPauseFrames[2].protocolId
+    );
+
+    if (frameSteps) {
+      const frameStep = frameSteps[0];
```
markerikson, line 303: :hand: It's probably going to be a bit trickier than this. The actual `dispatch()` call ought to be _somewhere_ in this frame, but there's no guarantee it's the _first_ step in the frame. 

I think the `ReactPanel.tsx` code has some logic that tries to find the right framestep for the line of code that led into this call.

src/ui/actions/timeline.ts
```diff
@@ -264,6 +269,44 @@ export function seek(
...
+      // Filter out everything in `node_modules`, so we have just app code left
+      // TODO There may be times when we care about renders queued by lib code
+      // TODO See about just filtering out React instead?
+      return !MORE_IGNORABLE_PARTIAL_URLS.some(partialUrl => source.url?.includes(partialUrl));
+    });
+
+    const frameSteps = await frameStepsCache.readAsync(
+      replayClient,
+      pauseId,
+      // The first 2 elements in filtered pause frames are from replay's redux stub, so they should be ignored
```
markerikson, line 297: :question: Can you paste in an example of what the callstack looks like here?

src/ui/actions/timeline.ts
markerikson: Yeah, the main logic here could be moved into its own cache.

markerikson (changes requested):

src/ui/actions/timeline.ts
Mokshit06: Oh interesting, @jasonLaster mentioned that it should be the first one but yeah `ReactPanel.tsx` does have some logic for finding the right one

src/ui/actions/timeline.ts
Mokshit06: Hm I think we don't need it here since `seek` calls it internally. Looks like I forgot to remove it

src/ui/actions/timeline.ts
Mokshit06: Do you mean the filtered pause frames?

Mokshit06:
I've made the suggested changes

src/ui/actions/timeline.ts
jasonLaster: ooh interesting.

jasonLaster:
@Mokshit06 looks like this needs a rebase

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -59,6 +66,94 @@ export const ReduxDevToolsPanel = () => {
...
+
+      // Filter out everything in `node_modules`, so we have just app code left
+      // TODO There may be times when we care about renders queued by lib code
+      // TODO See about just filtering out React instead?
+      return !MORE_IGNORABLE_PARTIAL_URLS.some(partialUrl => source.url?.includes(partialUrl));
+    });
+
+    // The first 2 elements in filtered pause frames are from replay's redux stub, so they should be ignored
+    // The 3rd element is the user function that calls it, and will most likely be the `dispatch` call
+    const preferredFrame = filteredPauseFrames[2];
```
markerikson, line 103: :hand: I think we need to update the logic here as well.

The issue is that there could be multiple Redux middleware that are in the call stack, on top of the actual `store.dispatch()` call.

As an example, clicking "Jump" for one of the Redux actions in our own FE E2E tests is jumping me to our "context middleware":

image

If we look at the call stack for this case, where we _really_ want to end up is the actual place where `dispatch(someAction())` was called - in this case, one of our store setup files:

image

Similarly, if I jump to the line for a `locationSelected` action, it would be best if we could end up jumping to the `dispatch(locationSelected())` line in this thunk:

image


I don't yet know a good heuristic for figuring out how many frames we need to skip backwards.  We know that a couple frames are part of Redux itself, but there could be multiple middleware frames before that on the call stack too.

Can you do some experimenting with our app and a couple other Redux apps and try to work out a good heuristic?

jasonLaster: Nice find. My hunch is that it'd be nice to land this and then have a more dedicated issue/pr for investigating middleware. 

For instance, I could see this tying into an ignorelist feature for files if we can't come up with the best heuristic.

Also - landing this change will make it easier to see what other cases we want to improve as well

Mokshit06: I agree, tracing the dispatch from middleware will require more work, so it is probably better to make it into a dedicated pr and land this as is. I will try experimenting with babel to find a fix

markerikson: Yeah, landing in the middleware is a step in the right direction, and admittedly not all Redux apps are going to have a userland middleware.

markerikson (approved):

markerikson:
Okay, let's land this. E2E failures are known flaky tests atm.

PR #9470, opened 2023-07-07
Only ignore the redux paths inside node_modules

Mokshit06:
The previous regex was too lose and would also end up grouping redux file paths in user code like `src/redux/utils.js` under Redux in callstacks and frames. Ideally it should only the `node_modules/redux` path should be doing this.


jasonLaster:
Looks good to me. CCing Mark

Mokshit06:
Included them as well.

PR #9472, opened 2023-07-08
Only show actions that are in the focus window

jasonLaster:
fixes FE-1659


markerikson:
Need to re-scope these PRs so that it's just the relevant commit(s), but makes sense otherwise.

PR #9476, opened 2023-07-09
Fix selecting items in quick open modal

Mokshit06:
The actual type of `SearchResult.location` should be `{ begin: Location; end?: Location }` instead of `{ start: Location; end?: Location }`. The latter errors as `start` doesn't exist in the response


jasonLaster:
Thanks! The code looks good, but it'd be nice to add a test to ensure we don't break quick open again.

I think a good test to model this after is `stepping-01.test.ts`, but probably any test that opens a file with a function will do.

PR #9483, opened 2023-07-10
add experimental setting

miriambudayr:


packages/shared/user-data/GraphQL/config.ts
```diff
@@ -43,6 +43,12 @@ export const config = {
     label: "Disable using incremental snapshots",
     legacyKey: "devtools.features.disableIncrementalSnapshots",
   },
+  backend_enableHasAnnotationKindQueryStorage: {
+    defaultValue: Boolean(false),
+    description: "Enable storage of previously generated response to Session.hasAnnotationKind",
+    label: "Enable query-level storage for Session.hasAnnotationKind",
```
miriambudayr, line 49: this is a bit specific for a user-facing experimental setting right?

packages/shared/user-data/GraphQL/config.ts
```diff
@@ -43,6 +43,12 @@ export const config = {
     label: "Disable using incremental snapshots",
     legacyKey: "devtools.features.disableIncrementalSnapshots",
   },
+  backend_enableHasAnnotationKindQueryStorage: {
```
bvaughn, line 46: Nit: Can we keep these keys alpha-sorted? Helps reduce likelihood of merge conflicts and makes it easier to scan code for a specific key.

packages/shared/user-data/GraphQL/config.ts
bvaughn: The label and description are what get displayed in the preferences dialog, yes. Whether or not they show up in "experimental" just depends on where you reference them. (This PR does put them in experimental though, assuming that's what you want.)

It's not clear to me if this should be an internal only setting.

bvaughn (commented):
I'm a little unclear on what the purpose of this change is. Neither the PR nor BAC-3618 have any description.

packages/shared/user-data/GraphQL/config.ts
miriambudayr: Hmm I think it might need to be an internal only setting for now. Did something change recently about how to set those settings from the console? I was having a hard time figuring that out yesterday.

miriambudayr:
> I'm a little unclear on what the purpose of this change is. Neither the PR nor BAC-3618 have any description.
> 
> This kind of change should also have at least one FE member reviewer assigned (when it's ready to be reviewed).

Sure, I can make sure of that

packages/shared/user-data/GraphQL/config.ts
bvaughn: The config objects support an optional `internalOnly` which are automatically hidden from non-Replay users in the preferences panel.

miriambudayr: Thanks! Fixed

bvaughn (approved):
👍🏼  Looks good to me

PR #9484, opened 2023-07-10
Going "back" from a recording will take you to the parent Library

bvaughn:
### [Loom overview](https://www.loom.com/share/6c732c16219e474f86a226d172ea465c)

cc @jonbell-lot23

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -64,8 +64,9 @@ export namespace RecordingTestMetadataV1 {
 export namespace RecordingTestMetadataV2 {
   export type GroupedTestCases = Omit<
     RecordingTestMetadataV3.GroupedTestCases,
-    "source" | "testRecordings"
+    "runId" | "source" | "testRecordings"
   > & {
+    run: { id: string };
     source: {
       path: string;
       title: string | null;
@@ -149,6 +150,8 @@ export namespace RecordingTestMetadataV3 {
 
     // Summarizes result of the test run
     resultCounts: Record<TestResult, number>;
+
+    runId: string;
 
     // Version for test metadata/schema
     schemaVersion: SemVer;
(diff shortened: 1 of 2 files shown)


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -1076,3 +1082,13 @@ export function getTestEnvironment(groupedTestCases: AnyGroupedTestCases): TestE
...
+
+export function getTestRunId(groupedTestCases: AnyGroupedTestCases): string | null {
+  if (isGroupedTestCasesV1(groupedTestCases)) {
+    return null;
+  } else if (isGroupedTestCasesV2(groupedTestCases)) {
+    return groupedTestCases.run.id;
+  } else {
+    return groupedTestCases.runId;
+  }
+}
```
bvaughn, line 1094: @ryanjduffy Is this  safe to rely on?

markerikson (approved):

src/ui/components/Header/Header.tsx
```diff
@@ -169,12 +171,24 @@ export default function Header() {
...
+  if (recording.workspace !== null) {
+    dashboardUrl = `/team/${recording.workspace?.id}`;
+
+    if (recording.isTest && recording.metadata?.test !== undefined) {
+      const runId = getTestRunId(recording.metadata.test);
+      dashboardUrl += `/runs/${runId}`;
+    }
+  } else {
+    dashboardUrl = "/team/me/recordings";
+  }
```
bvaughn, line 190: I don't love how coupled this code is with the combination of (global/Test Suite library) routing.

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: I think every recording with any test metadata will have a test run id (in `metadata.test.run.id`).

hbenl (approved):

PR #9485, opened 2023-07-11
Skip redux middleware frames on jumping to location

Mokshit06:
Experimenting with using babel to figure out which pause frames are of redux middlewares instead of `dispatch`.
It finds the identifier at the column and line where the first user frame paused and matches if the parent functions have the pattern `store => next => action => {}` using babel traversal. If so then skips that frame and moves to the next one. 

This is a very basic test, and needs to check for more cases, like checking at each step that the parent function is returned the child one. eg. in `next => action => {}`, `action => {}` is the child function. It's parent function is `next => {}`, and `next` is returned `action`.


src/ui/utils/redux.ts
```diff
@@ -0,0 +1,34 @@
...
+          }
+        },
+      });
+    }
+
+    return isMiddleware;
+  } catch (e) {
+    return false;
+  }
+}
```
jasonLaster, line 61: it'd be nice to write some jest unit tests for this function

https://github.com/replayio/devtools/blob/7731de4856591274a04b4596dd22d37416ec34f6/src/ui/utils/time.test.ts#L4
https://github.com/replayio/devtools/blob/7731de4856591274a04b4596dd22d37416ec34f6/src/ui/utils/app.test.ts#L4

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -70,6 +79,115 @@ export const ReduxDevToolsPanel = () => {
...
+
+    // One of these locations should match up
+    const matchingFrameStep: PointWithLocation | undefined = pointsWithLocations.find(step => {
+      // Intentionally ignore columns for now - this seems to produce better results
+      // that line up with the hit points in a print statement
+      return (
+        step.location.sourceId === preferredLocation.sourceId &&
+        step.location.line === preferredLocation.line
+      );
+    });
```
jasonLaster, line 168: Now that this function is getting more complex, it might be nice to pull this logic for getting the matching  frame step into a util


```js
function getMatchingFrameStep(frameSteps, preferredLocation) {
  ...
}
```

btw i'm not sure if this is the best API.

Mokshit06:
This should now be able to give more accurate results. It can identify code like `store => next => action => {}` and `store => next => { return action => {} }` and any combination of implicit & explicit returns as redux middleware

markerikson:
Not sure this is working as intended atm.  If I look at https://devtools-git-fork-mokshit06-redux-skip-middleware-recordreplay.vercel.app/recording/breakpoints-01-test-basic-breakpoint-functionality--326d54a3-1523-411a-95ca-018537f5a887 , the jump locations are all still landing inside of the `context` middleware.

Mokshit06:
Oh I know why this example isn't working. Babel isn't able to parse the typescript code. Will need to include typescript plugin for ts/tsx files. Try running it on JS examples

Mokshit06:
Oh that's a good point. How often does the name/path of this file change? Maybe that could be a good way to identify if it's the correct step. Another option could be to go through all the frame steps and based on the `sourceOutlineCache` check if the current pause location is between a function named `applyMiddleware`

markerikson:
The filename shouldn't change, but there _are_ some funky edge cases I've found where there can be more than one copy of `node_modules/redux/es/redux.js` in a project.  In one case, Redux was getting added at the top level as `node_modules/redux/`, but also under `node_modules/redux/@redux-saga/something/node_modules/redux/`.  So, there were two copies of `redux.js`, and only one of them was getting hit.

Was working on some of that logic over in #9410 for the perf analysis POC.  

Want to try grabbing some of that "find the right Redux file" code from there and use it here to identify the right method here?  I think you can actually borrow a bunch of the logic that was already trying to find "the right `dispatch` method inside of `createStore`", except it should be "inside of `applyMiddleware` instead".

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -79,6 +86,41 @@ export const ReduxDevToolsPanel = () => {
   );
 };
 
+async function isInsideApplyMiddlwareFn(
```
markerikson, line 89: :hand: It'll be smoother if we do the `applyMiddleware` function lookup once, then have a separate function that does the "is frame in function declaration" comparison synchronously

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
markerikson: FWIW the _real_ answer here is to implement https://linear.app/replay/issue/BAC-2915/stack-frames-should-come-with-execution-points and stop trying to calculate this ourselves.

Mokshit06: I've moved it into a separate util for now since `ReactPanel.tsx` and `ReduxDevtools.tsx` are doing the same thing but yes having frames have their execution point would make it _alot_ simpler

src/ui/utils/frame.ts
markerikson, line 1: Not sure this is a _good_ place for this code to go, but I also don't have a better suggestion atm :)

markerikson:
Looks reasonable at this point - can you fix up the merge conflict?

markerikson:
Hmm. I just tried it on a Metabase recording, and ended up exactly inside of the original `saveAnnotations()` call:

- https://devtools-git-fork-mokshit06-redux-skip-middleware-recordreplay.vercel.app/recording/e2etestscenariosjoinsreproductions18512-cannot-join-two-saved-questions-with-same-implicit-explicit-grouped-fieldcyspecjs--f455216f-8167-40f4-b1e1-3d3af2a83f30 

I definitely see the `applyMiddleware -> dispatch` method in the call stack, so we should have found that.

Can you do some more looking into this?

markerikson:
Per discussion in Discord, Mokshit is going to try looking at figuring out "are we in a middleware" by looking _just_ at source outline data to determine if this is a triply-nested function.

Mokshit06:
@markerikson This seems to be working now. It tries to search the call stack initially for `applyMiddlware` call. If it finds one then uses that frame, else tries the source outline data, and if both fail then jumps to the third frame (the one after the 2 replay patched steps).

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -133,6 +138,96 @@ function isFrameInDecl(decl: ApplyMiddlewareDecl, frame: PauseFrame) {
...
+}
+
+async function isReduxMiddleware(sourceContents: string, location: Location) {
+  const sourceOutline = await sourceOutlineCache.readAsync(replayClient, location.sourceId);
+  const dispatchFn = findFunctionOutlineForLocation(location, sourceOutline);
+  const functions = sourceOutline.functions;
+
+  if (dispatchFn) {
+    const index = functions.indexOf(dispatchFn);
+    if (index >= 2) {
```
markerikson, line 183: :question: Let me see if I understand the intent here:

- We're looking for a triply-nested function declaration
- We should be paused somewhere inside the center one
- That implies there are at least two function declarations in the file prior to this one
- So, the source outline's `functions` array _ought_ to be listed sequentially, and thus there's no point in looking if this function isn't at least at index 2

That cover it?

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -133,6 +138,96 @@ function isFrameInDecl(decl: ApplyMiddlewareDecl, frame: PauseFrame) {
...
+      const wrapDispatch = functions[index - 1];
```
markerikson: :hand: I can think of an edge case where this fails:

```ts
function myMiddleware(storeApi) {
  const someLookupTable = {};

  function someHelper() {}

  return (next) => {
    const anotherVar = 123;
    function anotherFunction() {}

    return (action) => {
      // pause here or whatever
    }
  }
}
```

These are closures with their own scopes, so it's not uncommon to have variables and possibly more functions in scope to do part of the work.

As a more practical example, see this socket handling middleware: https://gist.github.com/markerikson/3df1cf5abbac57820a20059287b4be58 .

I think the defining characteristics we care about are:

- Is it triply nested
- Do the inner two functions have 1 arg ( `next` and `action`) - I can't think of a time when you wouldn't declare both of those, although the names could easily be different
- The outermost function should either have 1 arg or 0 args (ie, it could be declared as `store`, `storeApi`, `{dispatch, getState}`, or `()` with no arg)

markerikson (changes requested):

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
Mokshit06: > That implies there are at least two function declarations in the file prior to this one

Not sure if I understand how this would always be true. Without this condition wouldn't we need to ensure that `wrapDispatchFn` and `middlewareFn` are not undefined?

markerikson:
One other tweak that could be worth making:

RTK normally installs 3 different middleware in dev, and 1 in production:

- immutability check (dev)
- serializability check (dev)
- thunk (prod)

Also, if you're using RTK Query, you add the RTKQ custom middleware to the store.

In https://devtools-bihivkc27-recordreplay.vercel.app/recording/replay-of-localhost5173--b90a2bd3-563a-4682-9c6e-bdddc61c7042 , I can see the serializability middleware at `http://localhost:5173/node_modules/@reduxjs/toolkit/src/serializableStateInvariantMiddleware.ts` , and similarly `http://localhost:5173/node_modules/@reduxjs/toolkit/src/immutableStateInvariantMiddleware.ts`.  The thunk middleware is at `http://localhost:5173/node_modules/redux-thunk/es/index.js` .

Can we filter out and exclude stack frames with the names `serializableStateInvariantMiddleware`, `immutableStateInvariantMiddleware`, and `redux-thunk`?  Those are definitely not user code, and not a place we'd want to use for J2C.

If you can make that tweak and the middleware closure tweak, I think this is good to go!

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
markerikson: I'm not saying to change anything here, I'm just trying to make sure I understand the intent of _why_ the `index >= 2` check exists

Mokshit06: Sorry, yep that's the intent of the check

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
Mokshit06: The latest commit should fix this edge case. Instead of directly considering the function at the position just before the current one as the parent, it now checks for all functions before it until it finds one that is its parent

markerikson:
Hah, I hate to keep asking you to make more tweaks, but I see one more filename we should exclude: `bindActionCreators` .  Example, see https://devtools-git-fork-mokshit06-redux-skip-middleware-recordreplay.vercel.app/recording/breakpoints-01-test-basic-breakpoint-functionality--326d54a3-1523-411a-95ca-018537f5a887 , and jump to `SET_FOCUSED_SOURCE_ITEM` , which was dispatched by an older-style `connect()`-wrapped component.

markerikson:
Okay, looks good! Just need to fix the merge conflicts and I think this is ready!

markerikson:
Huh, looks like this is missing a `process.env.AUTHENTICATED_TESTS_WORKSPACE_API_KEY` somehow?  that's causing all the E2E tests to fail and bail out.

Mokshit06:
Hm that's weird, I don't see anything in this PR that would change the env variables

markerikson (approved):

markerikson:
Superseded by #9573 .

PR #9486, opened 2023-07-11
Test Suites: Better format flaky tests

bvaughn:
- [x] Flaky tests should sorted so that passing test is at top, with failed attempts nested beneat
- [x] Only display full test title for first attempt (or passing attempt); just show attempt number beneath

| Before | After |
| :---: | :---: |
| Screen Shot 2023-07-11 at 12 46 16 PM | Screen Shot 2023-07-11 at 12 43 47 PM |
| Screen Shot 2023-07-11 at 12 46 22 PM | Screen Shot 2023-07-11 at 12 42 42 PM |

cc @jonbell-lot23 @jasonLaster


markerikson (approved):

jonbell-lot23:
Another welcome and brilliant improvement

bvaughn:
Our Cypress e2e test was failing because it was based on a recording generated by an older version of the Replay+Playwright plug-in (that didn't have `id` or `attempt` attributes). I updated that version, which should fix that test, but the test also changed a bit so I had to update the test slightly too.

PR #9488, opened 2023-07-12
protocolValueToTextHelper properly handles deeply nested objects

bvaughn:
Fixes a bug in `protocolValueToTextHelper` that sometimes caused invalid JS values to be copied to the clipboard via the context menu "Copy object" object. Added new e2e tests to cover the change.

Tests initially failed with:
```
Expected clipboard to contain
  "{"level-1": {"level-2": {"level-3": {"level-4": {"level-5": {"level-6": "[[ Truncated ]]"}}}}}}"

but found
   "{"level-1": {"level-2": {"level-3": {"level-4": {"level-5": {"[[ Truncated ]]"}}}}}}"
```


bvaughn:
Interesting. The new copy deep object test passes for me locally but fails on CI with:
```
Expected clipboard to contain
  "{"level-1": {"level-2": {"level-3": {"level-4": {"level-5": {"[[ Truncated ]]"}}}}}}"

but found
  "{"level-1": {"level-2": {"level-3": {"level-4": {"level-5": {"level-6": "[[ Truncated ]]"}}}}}}"'
```

That's...bizarre, because the test code I committed expects the "found" string, not the "expected" one:
```ts
  await verifyContextMenuCopy(
    page,
    findKeyValues,
    "filter_objectDeep",
    "{level-1: {…}}",
    "Copy object",
    '{"level-1": {"level-2": {"level-3": {"level-4": {"level-5": {"level-6": "[[ Truncated ]]"}}}}}}'
  );
```

Not sure what's going on there.

markerikson (approved):

PR #9489, opened 2023-07-12
Keep existing RDT UI component around and reset tree contents

markerikson:
This PR:

- Reworks our `ReactDevTools.tsx` file to only create the actual RDT component type _once_ and effectively "reset" its contents as the user jumps from point to point, instead of re-creating the entire component instance every time they jump to a point.
   - Ports the RDT operations processing logic I wrote for the backend routine in order to support generating the "remove root" operations on the client
- Patches the `react-devtools-inline` package to turn off a `console.warn("No element found!")` warning that was spewing in certain cases after the logic change (such as having done a search for some components by name, then jumping to a different point).  This was due to the component still having internal state for last identified component IDs that may no longer be relevant.  (Adjusting that behavior would be potential fodder for a future fork of the RDT UI.)

### DIFF
--- a/src/ui/components/SecondaryToolbox/react-devtools/rdtProcessing.ts
+++ b/src/ui/components/SecondaryToolbox/react-devtools/rdtProcessing.ts
@@ -165,9 +165,14 @@ export function parseTreeOperations(treeOperations: number[]): ParsedReactDevtoo
   // We're now going to iterate through just the actual
   // tree operations portion of the original operations array
   let i = 0;
+  let loopCounter = 0;
 
   // Iteration logic and index comments copied from `printOperations.ts`
   while (i < treeOperations.length) {
+    if (++loopCounter > 100000) {
+      throw new Error("TREE_OPERATION_PARSING_INFINITE_LOOP");
+    }
+
     const operation = treeOperations[i];
 
     switch (operation) {
@@ -259,6 +264,8 @@ export function parseTreeOperations(treeOperations: number[]): ParsedReactDevtoo
       case TREE_OPERATION_SET_SUBTREE_MODE: {
         const rootId = treeOperations[i + 1];
         const mode = treeOperations[i + 2];
+
+        i += 3;
 
         const operation: TreeOperationSetSubtreeMode = {
           type: TREE_OPERATION_SET_SUBTREE_MODE,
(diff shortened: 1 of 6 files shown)


hbenl:
When I open our react example recording (https://devtools-git-feature-fe-1392-rdt-jumping-take-2-recordreplay.vercel.app/recording/19fe1893-b9ea-41a5-a90a-90a209064c0b), open the RDT panel and then jump to another execution point, Replay freezes and the browser's CPU utilization goes to 100%.

markerikson:
Hmm. Seems like there's an issue with the tree operation parsing `while` loop...

markerikson:
Couple observations here:

- That recording is an older Gecko recording, and the routine doesn't process Gecko at all. So, these tree operations are coming directly from what we recorded inside Gecko.
- The actual operations array looks like it's malformed in this case. Still figuring out what's up with that.

src/ui/components/SecondaryToolbox/react-devtools/rdtProcessing.ts
```diff
@@ -0,0 +1,535 @@
...
+
+  return {
+    rendererId,
+    rootId,
+    stringTable,
+    treeOperations: parsedTreeOperations,
+  };
+}
+
+export function parseTreeOperations(treeOperations: number[]): ParsedReactDevtoolsTreeOperations[] {
```
bvaughn, line 162: Functions like `parseTreeOperations ` and `reconstructOperationsArray` should have unit tests. Doesn't seem like it would require a lot of effort to make them, and a regression in functions like this could have a large negative impact.

I don't think it's realist to review operations array deconstruction+reconstruction code like this as part of a PR anyway. Mistakes would be way too subtle.

markerikson: Agreed, these are great candidates for unit tests.

Talked to Jason about this. We agreed that I'll merge this now, and write the tests first thing after I get back from PTO:

- https://linear.app/replay/issue/FE-1720/write-tests-for-rdt-routine-processing-logic

markerikson:
Per discussion with Jason, merging this now and will address issues in a follow-up right after I get back.

PR #9490, opened 2023-07-13
Don't show top level preview in redux devtools

### LINKED ISSUE #9480: [FE-1677] [Redux] Don't show top-level preview
The contents pane should not show the top level preview

[Source image](https://github.com/replayio/devtools/assets/254562/dd95ef82-1281-4cc0-a13d-9e50ab07811b)

We pull this trick off elsewhere with the `PropertiesRenderer`

[https://github.com/replayio/devtools/blob/a22628afcc3b7bbb86017320317d5dfa080605f8/packages/replay-next/components/inspector/SourcePreviewInspector.tsx#L129-L131](https://github.com/replayio/devtools/blob/a22628afcc3b7bbb86017320317d5dfa080605f8/packages/replay-next/components/inspector/SourcePreviewInspector.tsx#L129-L131)

<sub>From [SyncLinear.com](https://synclinear.com) | [FE-1677](https://linear.app/replay/issue/FE-1677/[redux]-dont-show-top-level-preview)</sub>

Mokshit06:
Fixes #9480

### DIFF
--- a/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsContents.tsx
+++ b/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsContents.tsx
@@ -7,7 +7,7 @@ import {
 import classnames from "classnames";
 import React, { Suspense, useContext, useEffect, useLayoutEffect, useRef, useState } from "react";
 
-import Inspector from "replay-next/components/Inspector";
+import Inspector from "replay-next/components/inspector";
 import PropertiesRenderer from "replay-next/components/inspector/PropertiesRenderer";
 import Loader from "replay-next/components/Loader";
 import { clientValueCache, objectCache } from "replay-next/src/suspense/ObjectPreviews";
--- a/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsContents.tsx
+++ b/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsContents.tsx
@@ -7,6 +7,7 @@ import {
 import classnames from "classnames";
 import React, { Suspense, useContext, useEffect, useLayoutEffect, useRef, useState } from "react";
 
+import Inspector from "replay-next/components/Inspector";
 import PropertiesRenderer from "replay-next/components/inspector/PropertiesRenderer";
 import Loader from "replay-next/components/Loader";
 import { clientValueCache, objectCache } from "replay-next/src/suspense/ObjectPreviews";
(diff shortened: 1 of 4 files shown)


Mokshit06:
Regarding the change in UI snapshots I think that is because of the change in import of `KeyValueRenderer` from `./KeyValueRenderer` to `./KeyValueRendererWithContextMenu` due to auto import, so that should be fixed now.

As for not making changes to `Inspector`, I think we can just extract this out to a custom component, similar to what the source preview inspector does.

markerikson:
Looks like `PreviewPopup.tsx` uses `<SourcePreviewInspector>`, which in turn ends up rendering a `<PropertiesRenderer>` component instead of `<Inspector>` directly. Can you look into mimicking that approach?  Just trying to avoid changes to `<Inspector>` unless we really need them.

src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsContents.tsx
```diff
@@ -56,6 +63,34 @@ const PanelButton = ({ selected, children, name, onClick }: PanelButtonProps) =>
...
+  path?: string;
+  pauseId: PauseId;
+  protocolValue: ProtocolValue | ProtocolNamedValue;
+}) {
+  const client = useContext(ReplayClientContext);
+  const clientValue = clientValueCache.read(client, pauseId, protocolValue);
+
+  const { objectId } = clientValue;
+
+  const objectWithPreview = objectCache.read(client, pauseId, objectId!, "canOverflow");
```
markerikson, line 80: :hand: One more caveat here. It _is_ possible, even though unlikely, that the top-level state value could be just a simple primitive (ie, `configureStore({reducer: counterReducer})`.  

In that case, there _won't_ be an `objectId` - it'll be a `value` instead.

We should handle that case as well.

Mokshit06: I think this case _should_ be handled now, but I am unable to test it out locally due to some Replay internal issue I seem to be facing.

Mokshit06:
I believe it didn't show up when I ran TS is because of MacOS having a case insensitive file system (`Inspector/` instead of `inspector/`). Surprised this hasn't been fixed yet https://github.com/microsoft/TypeScript/issues/21736

markerikson:
Nice, looks like this works:

image

image

(be nice if we could ditch the `[[Prototype]]` entry while we're at it, but that's purely a nice to have eventually :)  )

Mokshit06:
The easiest way to hide `[[Prototype]]` from redux devtools would be in `PropertiesRenderer.tsx`. Is it alright to make changes to that in this case? https://github.com/Mokshit06/replay-devtools/commit/2450cc3f9f9dc142444e07ff26643801de3bf3fe

markerikson:
That seems reasonable, yeah - throw that up as another PR?

PR #9491, opened 2023-07-13
Fixed column execution point alignment

bvaughn:


Mokshit06:
When I try to run one of the Metabase E2E tests, I still see some issue. Though in this case it's also related to the character length being shorter than intended.
Screenshot 2023-07-13 at 7 49 29 PM

bvaughn:
> When I try to run one of the Metabase E2E tests, I still see some issues.

Replay?

Mokshit06:
https://app.replay.io/recording/column-execution-point-misaligned--5dd5b277-2248-4fbd-89bc-2cd70124e8ee This replay is based on #9485 after applying this change to css

hbenl (approved):

markerikson (approved):

bvaughn:
Regarding the over-all alignment, I think you're right that it's still a bit off in some cases– I think because of font size. (We're trying to offset based on the number of characters in the line number box and the hit count box, but those are both different font sizes.) I don't know how to tackle that off the top and will need to think about it a bit more.

> This replay is based on https://github.com/replayio/devtools/pull/9485 after applying this change to css

FWIW you can try this branch directly using the Vercel deploy @ https://devtools-git-fe-1705-recordreplay.vercel.app/

> In this case it's also related to the character length being shorter than intended.

> Not long enough to highlight entire expression

When you say "the entire expression", I think there's a misunderstanding. The frontend highlights the breakable chunk the backend tells it about. (This chunk is often not a complete valid/parseable  "expression".)

Mokshit06 (approved):

bvaughn:
I've refactored how I'm doing this to work more like the search highlight. In some ways, it's more heavy handed. It avoids the line number/hit counts font size trap though.

PR #9492, opened 2023-07-13
Add fast-forward and rewind to line context menu options

bvaughn:
### [Loom overview](https://www.loom.com/share/a6a48309bef94312936f8d1ff1e9bf0d)

- [x] Upgrade `use-context-menu` package from 0.4.10 to 0.4.11 so that context menus are always shown within _transitions_.
- [x] Replaced the `getHitPointsForLocationSuspense` function with a `hitPointsForLocationCache` so that it works with `use-context-menu` APIs like `useImperativeCacheValue`.
- [x] Updated `PreviewMarkers` to fetch hit point data imperatively to avoid potentially suspending for a long time on recordings like 55943a4c-7521-416d-8d63-cd2136876f54 that have weird performance characteristics.
- [x] Add "fast forward" and "rewind" options to Source context menu (before they were only visible on hover if you pressed the meta key).
- [x] Add e2e test coverage for new context menu items.

---

Screen Shot 2023-07-13 at 10 51 14 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "shared": "workspace:*",
     "slugify": "^1.6.5",
     "suspense": "^0.0.44",
-    "use-context-menu": "^0.4.10"
+    "use-context-menu": "^0.4.11"
   },
   "devDependencies": {
     "@babel/core": "^7.17.8",
--- a/packages/e2e-tests/helpers/context-menu.ts
+++ b/packages/e2e-tests/helpers/context-menu.ts
@@ -24,11 +24,9 @@ export async function selectContextMenuItem(
   } else if (contextMenuTestName) {
     contextMenuLocator = scope.locator(`[data-test-name="${contextMenuTestName}"]`);
   } else if (contextMenuItemText) {
-    contextMenuLocator = scope.locator(
-      `[data-test-name="ContextMenu"]:has-text("${contextMenuItemText}")`
-    );
+    contextMenuLocator = scope.locator(`[data-context-menu]:has-text("${contextMenuItemText}")`);
   } else {
-    contextMenuLocator = scope.locator('[data-test-name="ContextMenu"]');
+    contextMenuLocator = scope.locator("[data-context-menu]");
   }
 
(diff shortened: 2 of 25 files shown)


markerikson (approved):

bvaughn:
Looks like the new Playwright test is failing on CI. Not sure why; it passed for me locally. I'll look.

bvaughn:
Tests are still failing on CI only. Replay controllers crashing when I try to look at the failed recording. Not sure what's going on 😢

bvaughn:
I'm back to head scratching on this one. I run the same `yarn test:ci` script locally and the new test passes, but it seems to fail consistently on CI. I guess it _could_ be a Chromium vs Gecko thing?

bvaughn:
I added a log point to recording dedff89a-e5b9-4d6d-a054-d23f46e84d1f. tl;d it seems like the mouse move event is resetting the meta key in the context (`event.metaKey` ) but I'm not sure why.

hbenl:
FWIW I tried this test locally and for me it consistently fails both with standard and replay chromium. So I guess this is an issue with the linux playwright browser (but not our modifications of it).
There is some flakiness: sometimes it fails at "Start hovering over source "h1" line 14 with meta" and sometimes at "Start hovering over source "h1" line 14 with meta and shift".

hbenl:
I was able to fix this by changing the `hoverOverLine()` util, moving `numberLocator.hover(...)` before `page.keyboard.down(...)`.

hbenl:
>  it seems like the mouse move event is resetting the meta key in the context (event.metaKey ) but I'm not sure why.

That comment tipped me off: I'm seeing this all the time: when I press the Ctrl key and then move the mouse, the `HoverButton` reverts back to its normal state (i.e. the plus sign for adding a log point). Apparently this is only a thing on Linux!?

bvaughn:
> I was able to fix this by changing the `hoverOverLine()` util, moving `numberLocator.hover(...)` before `page.keyboard.down(...)`.

I think that's the same thing I pushed to try earlier this morning as a guess 😅

PR #9495, opened 2023-07-14
Source viewer tweaks

bvaughn:
## Don't syntax highlight lines that are more than 500 characters long

Rendering such lines is very slow, since we don't support horizontal windowing. So it's better (for performance reasons) to just render than plain text for longer lines.

> **Note** that 500 is a somewhat arbitrary length that should protected against extreme cases (minified files) without impacting any pretty-printed sources.

> **Note** that we still _parse_ the entire line, because we can't omit arbitrary lines without breaking the overall syntax. We just don't spend time creating/rendering React elements and HTML elements for the highlighted text.

Here's an example of what that looks like:
Screen Shot 2023-07-14 at 8 55 56 PM

## Jump-to-line also support column index

If you type `":<line-number>"` the `useSourceFileNameSearch` hook will scroll to the line you type. We should also support `":<line-number>:<column-number>"` for the case where we have a very long line. (VS Code does this.)

Screen Shot 2023-07-14 at 4 30 40 PM

Screen Shot 2023-07-14 at 4 30 31 PM

https://github.com/replayio/devtools/assets/29597/a4ecd7b0-238f-4cfd-9e2c-a9bbcde997c9

https://github.com/replayio/devtools/assets/29597/f6511168-9c27-4af4-9650-d2bbfb5cfeb5

---

cc @jonbell-lot23


bvaughn:
Okay added some e2e tests too.

bvaughn:
Same couple of e2e and screenshot tests failing here are also failing on #9505 so I'm guessing either that they're flaky or the backend had a hiccup.

PR #9497, opened 2023-07-14
WIP: Use cloned recording in E2E

callingmedic911:
Part of BAC-3629 and FE-1501.

Early draft. Depends on https://github.com/replayio/backend/pull/8204

### DIFF
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -53,8 +53,12 @@ export async function isDevToolsTabActive(page: Page) {
   return classes.includes("active");
 }
 
-export async function startTest(page: Page, example: string, apiKey?: string) {
-  const recordingId = exampleRecordings[example];
+export async function openRecording(
+  page: Page,
+  recordingId: string,
+  node?: boolean,
+  apiKey?: string
+) {
   const base = process.env.PLAYWRIGHT_TEST_BASE_URL || "http://localhost:8080";
 
   let url = `${base}/recording/${recordingId}?e2e=1`;
@@ -75,11 +79,17 @@ export async function startTest(page: Page, example: string, apiKey?: string) {
   });
 
   // Wait for the recording basic information to load such that the primary tabs are visible.
-  if (example.startsWith("node")) {
+  if (node) {
     await page.locator('[data-panel-id="Panel-SecondaryToolbox"]').waitFor();
   } else {
     // Node recordings don't have the "Viewer/DevTools" toggle
     await page.locator('[data-test-id="ViewToggle-Viewer"]').waitFor();
     await page.locator('[data-test-id="ViewToggle-DevTools"]').waitFor();
   }
+}
+
+export async function startTest(page: Page, example: string, apiKey?: string) {
+  const recordingId = exampleRecordings[example];
+
+  openRecording(page, recordingId, example.startsWith("node"), apiKey);
 }
--- a/packages/e2e-tests/helpers/utils.ts
+++ b/packages/e2e-tests/helpers/utils.ts
@@ -179,3 +179,43 @@ export async function resetTestUser(email: string) {
     },
   });
 }
+
+export async function cloneTestRecording(recordingId: string): Promise<string> {
+  const variables = { recordingId, secret: process.env.AUTOMATED_TEST_SECRET };
+
+  const clonedRecording = await axios({
+    url: config.graphqlUrl,
+    method: "POST",
+    data: {
+      query: `
+        mutation cloneTestRecording($recordingId: String!, $secret: String!) {
+          cloneTestRecording(input: { recordingId: $recordingId, secret: $secret }) {
+            recordingId
+          }
+        }
+      `,
+      variables,
+    },
+  });
+
+  return clonedRecording.data.data.cloneTestRecording.recordingId;
+}
+
+export async function deleteTestRecording(recordingId: string) {
+  const variables = { recordingId, secret: process.env.AUTOMATED_TEST_SECRET };
+
+  return axios({
+    url: config.graphqlUrl,
+    method: "POST",
+    data: {
+      query: `
+          mutation deleteTestRecording($recordingId: String!, $secret: String!) {
+            deleteTestRecording(input: { recordingId: $recordingId, secret: $secret }) {
+              success
+            }
+          }
+        `,
+      variables,
+    },
+  });
+}
--- a/packages/e2e-tests/testFixtureCloneRecording.ts
+++ b/packages/e2e-tests/testFixtureCloneRecording.ts
@@ -1,0 +1,35 @@
+import { Page, test as base } from "@playwright/test";
+
+import { cloneTestRecording, deleteTestRecording } from "./helpers/utils";
+
+type TestIsolatedRecordingFixture = {
+  recordingUrl: string;
+  pageWithMeta: {
+    page: Page;
+    recordingId: string;
+  };
+};
+
+export default base.extend<TestIsolatedRecordingFixture>({
+  recordingUrl: undefined,
+  pageWithMeta: async ({ page, recordingUrl }, use) => {
+    const exampleRecordings = require("./examples.json");
+    if (!exampleRecordings[recordingUrl]) {
+      throw new Error("Invalid recording");
+    }
+
+    console.log("Cloning recording", recordingUrl);
+    const newRecordingId = "testing";
+    // const newRecordingId = await cloneTestRecording(exampleRecordings[recordingUrl]);
+
+    await use({
+      page,
+      recordingId: newRecordingId,
+    });
+
+    console.log("Deleting recording", newRecordingId);
+    // await deleteTestRecording(newRecordingId);
+  },
+});
+
+export { expect } from "@playwright/test";
--- a/packages/e2e-tests/tests/console_dock.test.ts
+++ b/packages/e2e-tests/tests/console_dock.test.ts
@@ -1,6 +1,4 @@
-import test, { expect } from "@playwright/test";
-
-import { openDevToolsTab, startTest } from "../helpers";
+import { openDevToolsTab, openRecording, startTest } from "../helpers";
 import {
   getConsoleDockButton,
   getConsoleDockToBottomButton,
@@ -8,13 +6,14 @@ import {
   getConsoleDockToLeftButton,
   verifyConsoleLayout,
 } from "../helpers/dock";
+import test, { expect } from "../testFixtureCloneRecording";
 
-const url = "doc_rr_basic.html";
+test.use({ recordingUrl: "doc_rr_basic.html" });
 
 test("console_dock: Should show the correct docking behavior for recordings with video", async ({
-  page,
+  pageWithMeta: { page, recordingId },
 }) => {
-  await startTest(page, url);
+  await openRecording(page, recordingId);
   await openDevToolsTab(page);
 
   // Verify default docking position


.github/workflows/test.yml
```diff
@@ -152,6 +152,7 @@ jobs:
           RECORD_REPLAY_WEBHOOK_URL: ${{ secrets.RECORD_REPLAY_WEBHOOK_URL }}
           RECORD_REPLAY_TEST_METRICS: 1
           AUTOMATED_TEST_SECRET: ${{ secrets.AUTOMATED_TEST_SECRET }}
+          AUTHENTICATED_TESTS_WORKSPACE_API_KEY: ${{ secrets.AUTHENTICATED_TESTS_WORKSPACE_API_KEY }}
```
bvaughn, line 155: We'll need to make sure to add a secret for this in the backend repo too.

packages/e2e-tests/helpers/index.ts
```diff
@@ -83,3 +87,9 @@ export async function startTest(page: Page, example: string, apiKey?: string) {
     await page.locator('[data-test-id="ViewToggle-DevTools"]').waitFor();
   }
 }
+
+export async function startTest(page: Page, example: string, apiKey?: string) {
+  const recordingId = exampleRecordings[example];
+
+  openRecording(page, example, recordingId, apiKey);
+}
```
bvaughn, line 95: What's the value of splitting `openRecording` and `startTest` into separate functions?

I think we should merge them back into a single function. It's not clear why or when you would use one vs the other currently.

packages/e2e-tests/tests/console_dock.test.ts
```diff
@@ -1,20 +1,20 @@
...
-const url = "doc_rr_basic.html";
+test.use({ recordingUrl: "doc_rr_basic.html" });
 
 test("console_dock: Should show the correct docking behavior for recordings with video", async ({
-  page,
+  pageWithMeta: { page, recordingId },
+  recordingUrl,
 }) => {
-  await startTest(page, url);
+  await openRecording(page, recordingUrl, recordingId);
```
bvaughn, line 17: Why aren't you calling `startTest` here? I think we should keep a single "start" function.

packages/e2e-tests/helpers/index.ts
```diff
@@ -53,8 +53,12 @@ export async function isDevToolsTabActive(page: Page) {
   return classes.includes("active");
 }
 
-export async function startTest(page: Page, example: string, apiKey?: string) {
-  const recordingId = exampleRecordings[example];
+export async function openRecording(
+  page: Page,
+  example: string,
```
bvaughn, line 58: nit: I know you didn't add this, but as I'm reading this code again I realize that "example" isn't a great name here.

We could probably also make this TypeScript-typed to avoid accidentally passing the wrong thing too?

```diff
diff --git a/packages/e2e-tests/helpers/index.ts b/packages/e2e-tests/helpers/index.ts
index 7ac98befd..74041005f 100644
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -4,12 +4,11 @@ import dotenv from "dotenv";
 
 import { RecordingTarget } from "replay-next/src/suspense/BuildIdCache";
 
+import exampleRecordings from "../examples.json";
 import { debugPrint, getElementClasses } from "./utils";
 
 dotenv.config({ path: "../../.env" });
 
-const exampleRecordings = require("../examples.json");
-
 export async function getRecordingTarget(page: Page): Promise<RecordingTarget> {
   return page.evaluate(async () => {
     return (window as any).app.store.getState().app.recordingTarget;
@@ -53,8 +52,10 @@ export async function isDevToolsTabActive(page: Page) {
   return classes.includes("active");
 }
 
-export async function startTest(page: Page, example: string, apiKey?: string) {
-  const recordingId = exampleRecordings[example];
+type Key = keyof typeof exampleRecordings;
+
+export async function startTest(page: Page, exampleKey: Key, apiKey?: string) {
+  const recordingId = exampleRecordings[exampleKey];
   const base = process.env.PLAYWRIGHT_TEST_BASE_URL || "http://localhost:8080";
 
   let url = `${base}/recording/${recordingId}?e2e=1`;
diff --git a/packages/e2e-tests/tsconfig.json b/packages/e2e-tests/tsconfig.json
index 95323d5c6..f1f470c36 100644
--- a/packages/e2e-tests/tsconfig.json
+++ b/packages/e2e-tests/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "downlevelIteration": true,
     "esModuleInterop": true,
+    "resolveJsonModule": true,
     "types": ["cypress"]
   }
 }
```

packages/e2e-tests/helpers/utils.ts
```diff
@@ -179,3 +179,53 @@ export async function resetTestUser(email: string) {
     },
   });
 }
+
+export async function cloneTestRecording(recordingId: string): Promise<string> {
+  const variables = { recordingId, secret: process.env.AUTOMATED_TEST_SECRET };
```
bvaughn, line 184: Should we throw if `process.env.AUTOMATED_TEST_SECRET` is undefined? Might be clearer that way that it's an environmental/configuration issue.

packages/e2e-tests/helpers/utils.ts
```diff
@@ -179,3 +179,53 @@ export async function resetTestUser(email: string) {
...
+            recordingId
+          }
+        }
+      `,
+      variables,
+    },
+  });
+  var endTime = performance.now();
+  const timeInSecs = (endTime - startTime) / 1000;
+  console.log(`Cloning took ${timeInSecs} seconds.`);
```
bvaughn, line 206: Remove before landing?

```suggestion
```

packages/e2e-tests/helpers/utils.ts
```diff
@@ -179,3 +179,53 @@ export async function resetTestUser(email: string) {
...
+
+  return axios({
+    url: config.graphqlUrl,
+    method: "POST",
+    headers: {
+      Authorization: `Bearer ${process.env.AUTHENTICATED_TESTS_WORKSPACE_API_KEY}`,
+    },
+    data: {
+      query: `
+          mutation deleteTestRecording($recordingId: UUID!, $secret: String!) {
```
bvaughn, line 222: I think we'll also need a backend process that runs periodically and cleans up tests that are older than some threshold (an hour?) because we can't count on JS to always complete.

packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const newRecordingId = await cloneTestRecording(exampleRecordings[recordingUrl]);
+
+    await use({
+      page,
+      recordingId: newRecordingId,
+    });
+
+    console.log("Deleting cloned recording", newRecordingId);
```
bvaughn, line 28: Remove before landing?

```suggestion
```

packages/e2e-tests/authenticated/comments-01.test.ts
```diff
@@ -9,16 +7,18 @@ import {
...
-const url = "authenticated_comments_1.html";
+test.use({ recordingUrl: "authenticated_comments_1.html" });
 
 test(`authenticated/comments-01: Test add, edit, and delete comment functionality`, async ({
-  page,
+  pageWithMeta: { page, recordingId },
+  recordingUrl: url,
 }) => {
-  await startTest(page, url, E2E_USER_1_API_KEY);
+  await openRecording(page, url, recordingId, E2E_USER_1_API_KEY);
```
bvaughn, line 21: Why aren't you calling `startTest` here? I think we should keep a single "start" function.

bvaughn (commented):
Thank you for working on this! 😄 

Left some thoughts.

.github/workflows/test.yml
callingmedic911: https://github.com/replayio/backend/pull/8235

packages/e2e-tests/helpers/index.ts
callingmedic911: Initially kept separate to allow opt-in rather than refactoring all the tests. Now that cloned API seems to be settled, keeping just `startTest`

packages/e2e-tests/helpers/index.ts
callingmedic911: Thanks for the diff.

packages/e2e-tests/helpers/utils.ts
callingmedic911: Done. Though, if we only had public recording this is not required. Note to future PR visitors. 👋

packages/e2e-tests/helpers/utils.ts
callingmedic911: There isn't any observability into tests right now. And in future, if we ever had a big example recording it'd be hard to point if test itself is flow or machine it's running on or the clone.

How about we log if time is greater than 10 sec?

packages/e2e-tests/helpers/utils.ts
callingmedic911: We are keeping track which recording are clones (`cloned_from` field in `recordings` table) but we don't have scheduled deletion implemented yet but as needed we can and execute on these dangling recording.

.github/workflows/test.yml
callingmedic911: Can you please set in devtools repo @bvaughn?

packages/e2e-tests/helpers/utils.ts
bvaughn: Oh, sorry! I meant for my comment to be about `process.env.AUTHENTICATED_TESTS_WORKSPACE_API_KEY`, since (I think) the API clone request would fail if that param was missing.

packages/e2e-tests/helpers/utils.ts
bvaughn: Hm, gotcha. That's fair. Maybe `console.warn` then? 😄

packages/e2e-tests/helpers/utils.ts
bvaughn: Gotcha! Seems like something we will want to add so it's good we have an easy way to identify these.

packages/e2e-tests/helpers/utils.ts
bvaughn: Oh, sorry! I meant for my comment to be about `process.env.AUTHENTICATED_TESTS_WORKSPACE_API_KEY`, since (I think) the API clone request would fail if that param was missing.

callingmedic911: Right, even though I understood you meant this env I relied too much copilot haha. Fixing. Same note applies though, that's an optional header. We can skip whole header if public or pass this if it's private. We have both so passing this to public recording won't have a problem.

bvaughn:
> Value is in 1Password titled "E2E - Clone API Workspace API key"

> Can you please set in devtools repo @bvaughn?

@callingmedic911 I don't see any such string in 1Password so you might want to double check that.

I copied it out of the backend PR though.

callingmedic911:
@bvaughn This is ready for review. All of tests are using recording clone and are completely isolated.

Those two failing I've seen them failing without cloned recording as well and I guess should be tracked separately.

packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -0,0 +1,35 @@
...
+      recordingId: newRecordingId,
+    });
+
+    await deleteTestRecording(newRecordingId);
+  },
+});
+export default testWithCloneRecording;
+export { testWithCloneRecording as test };
+export { expect } from "@playwright/test";
+export type { Page } from "@playwright/test";
```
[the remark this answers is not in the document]
callingmedic911: Added in Linear: https://linear.app/replay/issue/FE-1740/create-eslint-warning-rule-if-we-import-from-playwrighttest (assigned to myself, as soon as possible)

PR #9498, opened 2023-07-15
Hide prototype object preview from redux devtools

Mokshit06:
In addition to #9490

### DIFF
--- a/packages/replay-next/components/inspector/PropertiesRenderer.tsx
+++ b/packages/replay-next/components/inspector/PropertiesRenderer.tsx
@@ -27,15 +27,15 @@ const PROPERTY_BUCKET_SIZE = 100;
 //
 // https://static.replay.io/protocol/tot/Pause/#type-Property
 export default function PropertiesRenderer({
+  hidePrototype = false,
   object,
   path,
   pauseId,
-  hidePrototype = false,
 }: {
+  hidePrototype?: boolean;
   object: ProtocolObject;
   path?: string;
   pauseId: ProtocolPauseId;
-  hidePrototype?: boolean;
 }) {
   const client = useContext(ReplayClientContext);
 
--- a/packages/replay-next/components/inspector/PropertiesRenderer.tsx
+++ b/packages/replay-next/components/inspector/PropertiesRenderer.tsx
@@ -30,10 +30,12 @@ export default function PropertiesRenderer({
   object,
   path,
   pauseId,
+  hidePrototype = false,
 }: {
   object: ProtocolObject;
   path?: string;
   pauseId: ProtocolPauseId;
+  hidePrototype?: boolean;
 }) {
   const client = useContext(ReplayClientContext);
 
@@ -119,7 +121,7 @@ export default function PropertiesRenderer({
   }
 
   let renderedPrototype: ReactNode = null;
-  if (prototype != null) {
+  if (prototype != null && !hidePrototype) {
     const prototypePath = addPathSegment(path, "[[Prototype]]");
     renderedPrototype = (
       <Expandable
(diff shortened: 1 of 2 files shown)


bvaughn (commented):
I'm not sure about this change. Can you explain the motivations?

markerikson:
Redux state and actions are almost always plain JS objects.  Showing a `[[Prototype]]` field is just clutter in that case.

bvaughn:
Hmm. "almost" ? That seems kind of important.

I think a similar argument could be made for Chrome console. People rarely look at prototype properties, but when you need them, they're important to have available.

markerikson:
The only meaningful case I can think of for non-POJO Redux state would be Immutable.js collections.  That's still pretty rare, and I don't think prototypes come into play there either.

Either way, we really don't need prototypes for Redux state as far as I can see.

packages/replay-next/components/inspector/PropertiesRenderer.tsx
```diff
@@ -30,10 +30,12 @@ export default function PropertiesRenderer({
   object,
   path,
   pauseId,
+  hidePrototype = false,
 }: {
   object: ProtocolObject;
   path?: string;
   pauseId: ProtocolPauseId;
+  hidePrototype?: boolean;
```
bvaughn, line 38: Alpha sort props and type attributes :)

bvaughn (approved):
I don't really like us changing the shared/standard inspector UI in this way. Seems like React and Redux should  look and feel like Replay in as many ways as possible (which should in turn look and feel like Chrome devtools in most cases).

If you both feel strongly that this is the right thing to do though, I don't have a _strong objective_ argument. (I do think that this change would be breaking use cases like `Immutable.js` though.)

Mokshit06:
I feel like given that is is focused in the docs and other places to keep the redux state serialisable, we can be pretty certain that it will be a normal JS object. But it looks like Redux + ImmutableJS libraries still do get a [decent number of installs](https://www.npmjs.com/package/redux-immutable), so I'm not sure if it makes sense to break these

packages/replay-next/components/inspector/PropertiesRenderer.tsx
Mokshit06: Should this be in the ESLint config? I saw that `eslint-plugin-sort-keys-fix` was there in the `package.json` but not used in `.eslintrc.json`

bvaughn: Possibly. Can't remember if there was a reason for omitting it (in some cases).

Mokshit06: We can restrict it to `.tsx` files https://www.npmjs.com/package/eslint-plugin-sort-keys-fix#usage. I can open a PR if it makes sense to add it

markerikson:
@bvaughn I do actually feel pretty strongly about this.

This is _really_ just irrelevant clutter in the React and Redux panels, and it's not something we need to show.

jasonLaster:
This topic seems larger than whether or not we show the prototype. 

We should probably table the immutable conversation until we see an example in the wild. At that point we should compare replay to chrome and discuss options.

markerikson:
Just for completeness' sake, I logged an Immutable.js `List` from https://codesandbox.io/s/addpostswithimmutablejs-fe721?file=/src/App.js :

image

The data values are there, but nested down several levels, arranged in tuples for the key/value pairs, and buried amongst a bunch of internal fields for the data structures.

The Immutable.js types _are_ classes with prototypes, but it's all a bunch of methods, and not anything we would care about showing for the actual _data_ in the Redux panel.

PR #9503, opened 2023-07-17
Possible fix for FE-1500

hbenl:
To fix the overrendering of scope values (FE-1500), with this PR they are being prefetched in `ScopesInspector`, i.e. that component suspends until all the object previews for scope values have been received.
Interestingly, that doesn't suffice to fix the issue: I also had to remove `startTransition()` from `Expandable`, so we're not using a transition anymore...
See #9467 for STRs.

### DIFF
--- a/packages/replay-next/components/Expandable.tsx
+++ b/packages/replay-next/components/Expandable.tsx
@@ -64,13 +64,13 @@ export default function Expandable({
     const newIsOpen = !isOpen;
 
     // In case this change triggers a re-render that suspends, it should be in a transition.
-    startTransition(() => {
+    // startTransition(() => {
       onChange(newIsOpen);
       setIsOpen(newIsOpen);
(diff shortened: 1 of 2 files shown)


packages/replay-next/components/Expandable.tsx
```diff
@@ -64,13 +64,13 @@ export default function Expandable({
     const newIsOpen = !isOpen;
 
     // In case this change triggers a re-render that suspends, it should be in a transition.
-    startTransition(() => {
+    // startTransition(() => {
```
bvaughn, line 67: This isn't going to be a working option for us. If something suspends inside of an expandable (likely) during a non-transition update, React will unmount the entire thing and show a fallback.

bvaughn (changes requested):

bvaughn:
Just cleaning up some PRs.

PR #9505, opened 2023-07-17
use unified experimental setting for protocol query cache project

miriambudayr:
Earlier I was going the route of using one experimental setting per command that I was converting to use the cache, but that is proving to be unwieldy and not worth the effort.

https://linear.app/replay/issue/BAC-3649/use-unified-experimental-setting-for-protocol-query-cache-project


bvaughn (approved):

bvaughn:
@miriambudayr Mark's off this week, I think. Tag Holger or I.

miriambudayr:
> @miriambudayr Mark's off this week, I think. Tag Holger or I.

sounds good thanks!

PR #9506, opened 2023-07-17
Fix cypress view source lookup for invocations from within then()

ryanjduffy:
## Issue

When a cypress command is enqueued from with a `then()` callback, the call stack is different and we were not finding the correct frame to point to the invocation of the cypress command

## Resolution

Adjust the logic to not return a frame from within the cypress_runner

### DIFF
--- a/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
@@ -47,14 +47,16 @@ export const TestStepSourceLocationCache = createCacheWithTelemetry<
             const isChaiAssertion = testEvent.data.command.name === "assert";
             // TODO [FE-1419] const isChaiAssertion = testStep.data.name === "assert" && !annotations?.enqueue;
 
-            const frame = userFrames.find((f, i, l) => {
+            const frame = userFrames.find((currentFrame, currentFrameIndex) => {
               if (isChaiAssertion) {
                 // if this is a chai assertion, the invocation was synchronous in this
                 // call stack so we're looking from the top for the first frame that
                 // isn't from the same source as the marker to identify the user frame
                 // that invoke it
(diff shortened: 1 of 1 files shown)


src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
```diff
@@ -54,10 +54,34 @@ export const TestStepSourceLocationCache = createCacheWithTelemetry<
...
+                //     wrapper code like for cy.then in cypress_runner
+                //     __stackReplacementMarker in cypress_runner
+                //   ]
+                //
+                // We find it by looking from the top (the frames array comes in
+                // with the most recent last but it is reversed above) for the
+                // first entry that isn't from the cypress_runner where the
+                // _next_ frame is from the cypress_runner.
+
                 return l[i + 1]?.functionLocation?.some(fl => fl.sourceId === markerSourceId);
```
bvaughn, line 82: I know you didn't newly add these variables in this PR, but I'd prefer we avoid variable names like `l` and `fl`. They're really confusing when you get nested deeply enough to not be able to see where they were defined.

bvaughn (commented):
I'm really concerned about us encoding this level of internal Cypress details (potentially Cypress version specific details at that) in the frontend.

Can we create a Linear issue/project to move this out of the frontend?

ryanjduffy:
> I'm really concerned about us encoding this level of internal Cypress details (potentially Cypress version specific details at that) in the frontend.
> 
> Can we create a Linear issue/project to move this out of the frontend?

We have FE-1354 and i'll add some comments to this effect in it.

src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
ryanjduffy: Sure. Can do!

bvaughn (approved):

PR #9508, opened 2023-07-18
Adjust error color to be less bright

jonbell-lot23:
Old on left, new on right:
image

image

Here's where I was [experimenting in Figma](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?type=design&node-id=17774-308275&mode=design):
image

Addresses https://linear.app/replay/issue/DES-799/our-error-red-is-pretty-bright-align-closer-to-cypress

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -513,8 +513,8 @@
   --testsuites-capsule-error-color: white;
   --testsuites-capsule-success-bgcolor: var(--testsuites-success-color);
   --testsuites-capsule-success-color: rgb(16, 114, 0);
-  --testsuites-error-bgcolor-hover: rgba(237, 36, 36, 0.672);
-  --testsuites-error-bgcolor: rgba(237, 36, 36, 0.672);
+  --testsuites-error-bgcolor-hover: rgba(88, 58, 80, 1);
+  --testsuites-error-bgcolor: rgba(88, 58, 80, 1);
   --testsuites-error-border-color: var(--testsuites-error-icon-bgcolor);
   --testsuites-error-color: #ffe0e0;
   --testsuites-error-icon-bgcolor: rgb(237, 36, 36);
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -513,8 +513,8 @@
   --testsuites-capsule-error-color: white;
   --testsuites-capsule-success-bgcolor: var(--testsuites-success-color);
   --testsuites-capsule-success-color: rgb(16, 114, 0);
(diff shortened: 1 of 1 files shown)


bvaughn (commented):
Making the colors less bright seems like a good goal, but we're changing the hue here too in a way that (I think) does not look good. Our typical error color is a robust red:

Screen Shot 2023-07-18 at 7 30 40 AM

This new color is a _pink_:

Screen Shot 2023-07-18 at 7 30 48 AM

I think what's going on here is that the alpha transparency mixes with the blue-gray background. I also don't think these colors compliment each other.

Compare this to the light theme, where both dark and light error colors are the same hue:

Screen Shot 2023-07-18 at 7 33 00 AM

Screen Shot 2023-07-18 at 7 33 12 AM

jonbell-lot23:
The previous color was picked because it's similar to what people are familiar with when using Cypress, and aligning dark and light theme to the same hue isn't a priority. Dark and light exist in different color spaces, and it's very common (and recommended) to "adjust to taste."

That said, I'm fine with a darker shade of red, #4f1615. What do you think

image

bvaughn (approved):
Thanks for sharing your thinking.

IMO we shouldn't try to match Cypress colors. We should stick within the Replay color scheme. Mismatching hues drives me bonkers. Maybe that's a subjective thing.

jonbell-lot23:
> Thanks for sharing your thinking.
> 
> 
> 
> IMO we shouldn't try to match Cypress colors. We should stick within the Replay color scheme. Mismatching hues drives me bonkers. Maybe that's a subjective thing.

It's subjective, yes, but that doesn't make it any less true! 😂 If I used an Insane Clown Posse illustration rather than our unicorn, some people wouldn't like it. And that number would probably be about 99%! So I always appreciate this feedback.

So in this case, how are you feeling about the new red? It's no longer in the Cypress color space, but it's also not a perfect match with other reds we're using because we're purposely trying to mute the sharpness of the red. What do you think about this iteration?

bvaughn:
Oh sorry. I should have explicitly said this rather than just approving the PR– The new red feels better!

jonbell-lot23:
I was already thinking of doing a fuller alignment with the different reds we're using here (such as the red X icons, which have always been brighter by design), the assert label, and the background of this red.

So maybe the best approach is a) ship this if we're feeling ok about this red b) keep an eye on the family of colors for follow-up issues such as DES-801

jonbell-lot23:
Oh dear, this was a regression from six months ago:
https://app.campsite.design/replay/posts/03ubecsb53f3

image

All's well that ends well

PR #9510, opened 2023-07-18
More precise focus window

bvaughn:
### Code changes
- [x] Update `createSession` to pass `focusRequest` rather than `focusWindow`.
- [x] Replace `Session.requestFocusRange` with `Session.requestFocusWindow`.
- [x] Refine focus window using the _currently displayed_ focus window (see comments on [39e26ef4-5697-4fdb-9e8a-28dd2bc1854c](https://app.replay.io/recording/fe-1632-focus-window-expands-when-dragging--39e26ef4-5697-4fdb-9e8a-28dd2bc1854c)); I think this should fix FE-1632.
- [x] Migrate Redux actions (and React components that use them) to set focus using `TimeStampedPoint` when possible.

### Dependencies
- [x] Upgrade `@replayio/protocol` 0.54.0 -> 0.55.0 for new focus method.
- [x] Downgrade `use-context-menu` 0.4.11 -> 0.4.10 because the transition introduced in 0.4.11 caused the context menu not to open sometimes. (I'm not sure why this is, but I noticed it while testing the log point panel focus menu.)

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -1,4 +1,11 @@
-import { ExecutionPoint, FocusWindowRequestBias, PauseId, ScreenShot } from "@replayio/protocol";
+import assert from "assert";
+import {
+  ExecutionPoint,
+  FocusWindowRequestBias,
+  PauseId,
+  ScreenShot,
+  TimeStampedPointRange,
+} from "@replayio/protocol";
 import clamp from "lodash/clamp";
 
 import { framePositionsCleared, resumed } from "devtools/client/debugger/src/reducers/pause";
@@ -25,6 +32,7 @@ import {
   sessionEndPointCache,
 } from "replay-next/src/suspense/ExecutionPointsCache";
 import { screenshotCache } from "replay-next/src/suspense/ScreenshotCache";
+import { isExecutionPointsLessThan } from "replay-next/src/utils/time";
 import {
   isTimeStampedPointRangeEqual,
   isTimeStampedPointRangeGreaterThan,
@@ -51,7 +59,6 @@ import {
   getShowFocusModeControls,
   getZoomRegion,
   pointsReceived,
-  setFocusWindow,
   setPlaybackPrecachedTime,
 } from "ui/reducers/timeline";
 import { FocusWindow, HoveredItem, PlaybackOptions, TimeRange } from "ui/state/timeline";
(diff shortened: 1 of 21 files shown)


src/ui/actions/timeline.ts
```diff
@@ -564,64 +569,35 @@ export function updateFocusWindow(
...
 
-    let { begin: beginTime, end: endTime } = focusWindow;
+    let { begin, end } = focusWindow;
 
-    const zoomRegion = getZoomRegion(state);
     const prevFocusWindow = getFocusWindow(state);
     const prevBeginTime = prevFocusWindow?.begin.time;
     const prevEndTime = prevFocusWindow?.end.time;
 
-    // Basic bounds check.
```
bvaughn, line 574: The `Focuser` component does this (in React state) now. It's more direct.

src/ui/actions/timeline.ts
```diff
@@ -564,64 +569,35 @@ export function updateFocusWindow(
...
-    if (endTime > zoomRegion.endTime) {
-      endTime = zoomRegion.endTime;
-      if (beginTime > endTime) {
-        beginTime = endTime;
-      }
-    }
-
     // Make sure our region is valid.
-    if (endTime < beginTime) {
-      // If we need to adjust a dimension, it's the most intuitive to adjust the one that's being updated.
```
bvaughn, line 590: The `Focuser` component does this (in React state) now. It's more direct.

bvaughn:
@hbenl This could use a close look over.

hbenl (approved):
We should get the end point from our `sessionEndPointCache`, otherwise LGTM.

src/ui/actions/timeline.ts
```diff
@@ -678,8 +665,13 @@ export function syncFocusedRegion(): UIThunkAction {
...
 
-    const requestedBeginTime = focusWindow ? focusWindow.begin.time : zoomTime.beginTime;
-    const requestedEndTime = focusWindow ? focusWindow.end.time : zoomTime.endTime;
+    const begin = focusWindow
+      ? focusWindow.begin
+      : {
+          point: "0",
+          time: zoomTime.beginTime,
+        };
+    const end = focusWindow ? focusWindow.end : await replayClient.getSessionEndpoint();
```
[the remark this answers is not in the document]
bvaughn: Oh, right. Good catch.

bvaughn:
Thanks for the review, Holger!

PR #9513, opened 2023-07-18
Don't show passport for testsuites

jonbell-lot23:
Addressing https://linear.app/replay/issue/DES-791/disable-passport-for-new-testsuites-users

In time we may add a special tour and/or passport for testsuites, but for now we're concerned it may be confusing.

### DIFF
--- a/src/ui/components/Toolbar.tsx
+++ b/src/ui/components/Toolbar.tsx
@@ -423,7 +423,8 @@ export default function Toolbar() {
         {testRunner !== "cypress" && showTour ? (
           <ToolbarButton icon="tour" name="tour" label="Replay Tour" onClick={handleButtonClick} />
         ) : null}
-        {showPassport ? (
+
+        {showPassport && testRunner === null ? (
           <ToolbarButton
             icon="passport"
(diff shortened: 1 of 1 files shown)


src/ui/components/Toolbar.tsx
```diff
@@ -423,14 +423,16 @@ export default function Toolbar() {
         {testRunner !== "cypress" && showTour ? (
           <ToolbarButton icon="tour" name="tour" label="Replay Tour" onClick={handleButtonClick} />
         ) : null}
-        {showPassport ? (
+
+        {showPassport && testRunner === null ? (
```
bvaughn, line 427: nit: `showPassport` as a name for a variable doesn't make as much sense if it's not the only thing we consider before actually showing the passport.

Maybe we could name this `showPassportPreference` or something?

bvaughn (approved):

bvaughn:
Has this PR been dropped? :)

jonbell-lot23:
Yeah, things got complicated with some other asks. It's not 100% dropped yet but not ready to go either. Will address

Issue #9482, opened 2023-07-10
[FE-1690] [Redux] should be possible to filter by action


jasonLaster:
<sub>From [SyncLinear.com](https://synclinear.com) | [FE-1690](https://linear.app/replay/issue/FE-1690/[redux]-should-be-possible-to-filter-by-action)</sub>

PR #9520, opened 2023-07-20
Filter by action in Redux Devtools

Mokshit06:
Rel #9482 

image

### DIFF
--- a/src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
@@ -27,6 +27,7 @@ export const ReduxDevToolsPanel = () => {
   const client = useContext(ReplayClientContext);
   const [selectedPoint, setSelectedPoint] = useState<ExecutionPoint | null>(null);
   const focusWindow = useAppSelector(getFocusWindow);
+  const [searchValue, setSearchValue] = useState("");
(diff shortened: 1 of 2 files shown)


Mokshit06:
The current style of this panel is similar to the `Elements` Panel, but sure, padding above the input?

jasonLaster:
On second thought - lets add a follow up issue for @jonbell-lot23

src/ui/components/SecondaryToolbox/ReduxDevTools.tsx
```diff
@@ -37,9 +38,12 @@ export const ReduxDevToolsPanel = () => {
   const reduxAnnotations: ReduxActionAnnotation[] = useMemo(() => {
     const annotations = annotationsStatus === "resolved" ? parsedAnnotations : [];
     return annotations.filter(
-      annotation => focusWindow && isPointInRegion(annotation.point, focusWindow)
+      annotation =>
```
markerikson, line 41: :hand: Probably need to check if there even is a non-empty `searchValue` before we do that comparison

Mokshit06: I don't think that's required with `String#match` as it treats the second string as a regex, but on second thought `String#includes` is probably better in this case

PR #9521, opened 2023-07-20
Implement Redux Trace panel

Mokshit06:
image

Uses the same `reduxJumpToLocationCache` for finding frame steps of user code and showing stack traces.


jasonLaster:
Nice! One suggestion, lets call this `Stack` to be consistent with the network monitor and pause panel

markerikson:
UI behavior looks pretty good here. Just to confirm: we're going to use the same point/location that we're using for Jump to Code as the basis for the stack trace, right?  Does this PR need to be rebased on top of #9485 once that's merged?

Mokshit06:
Yep it will need to be rebased as it moves the cache to a separate file

markerikson:
Hmm. I'm glancing at https://devtools-git-fork-mokshit06-redux-stack-trace-recordreplay.vercel.app/recording/breakpoints-01-test-basic-breakpoint-functionality--ab329d84-7429-4571-a7ac-f5bd85017cb5 , and for some reason all the jump locations and stack traces are back to pointing to our `context.ts` middleware file.

I know this PR is about adding the "Stack" tab, and _that_ part seems okay, but can you take one more look at the jump detection behavior to see why that's happening?

markerikson:
Huh, yeah, that's bizarre. Okay, let's go ahead and get this in.

PR #9525, opened 2023-07-20
Show selected frame's location on switch to alternate source (FE-1732)

hbenl:


bvaughn (approved):
🤷🏼

bvaughn:
Should we start adding e2e tests for stuff like this? Seems like alternate/corresponding source stuff bites us kind of often.

PR #9528, opened 2023-07-23
[WIP] MVP protocol sandbox

jasonLaster:
Screenshot 2023-07-23 at 4 49 55 PM

### Requirements
* [x] re-use users login credentials
* [x] input for setting a recording id
* [x] textarea for writing a script that will be eval'd against the protocol
* [x] serialize script to url
* [x] syntax highlighting

### Nice to haves
* [ ] Show list of scripts
* [ ] Show list of recordings
* [ ] Show list of recording title
* [ ] Improve UI
* [ ] Serialize recording ID to URL

### Follow ups
* [ ] autocomplete with TS types
* [ ] show logs on the right


jasonLaster:
**Two reasons**
1. Shared auth / don't need to login twice
2. Shared replayClient. Not sure if we'll expose it, but seemed like a feature

I'd like to ship this and then likely extract when it's earned it.

bvaughn:
> 2. Shared replayClient. Not sure if we'll expose it, but seemed like a feature

You're not using the Replay client (or any shared code, from what I can tell)

> I'd like to ship this and then likely extract when it's earned it.

Ah,. This is kind of the opposite intuition I'd have :) but maybe I'm over-correcting because (I feel) we've shipped too much untested prototype-style code.

Let's run this through a full code review process before we land it :)

jasonLaster:
> You're not using the Replay client (or any shared code, from what I can tell)

I'm using the TokenManager and plan on also re-using the Auth0 provider for SSO. More importantly, I want to share the cookie login state

PR #9531, opened 2023-07-24
Optimize log point panel rendering behavior

bvaughn:
Address a possible **slow** update when adding log points to a Source file (made worse if `findPoints` protocol requests are slow):
- [x] Update `LoggablesContext` so that it maps _points_ (locations in source code) to _hit points_ (execution points) without the need to suspend during render. (See source code comments for how this was done.)
- [x] Update `LogPointPanel` to `useImperativeCacheValue` so that it can render its shell UI before hit points have been fetched.

Helpful in-depth information:
* [Watch this Loom for an explanation](https://www.loom.com/share/3cca3bc345864d83a250ccf50feb87f2)
* [See this Replay for a code walk through](https://app.replay.io/recording/70745841-44ad-4a80-a7f5-91c83600c971)


bvaughn:
Sorry. Marking this as Draft because I want to think about it a bit more.

bvaughn:
I'm going to wait until #9536 lands (and maybe #9534) and then rebase this.

packages/replay-next/src/hooks/useForceUpdate.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback, useState } from "react";
+
+export function useForceUpdate() {
```
bvaughn, line 3: I didn't end up using this because I went with `useSyncExternalStore` instead. Seems like it's maybe worth adding though in case?

packages/replay-next/src/suspense/ExceptionsCache.ts
```diff
@@ -26,10 +33,6 @@ function transformPoint(pointDescription: PointDescription): UncaughtException {
   return { type: "UncaughtException", ...pointDescription };
 }
 
-export const getInfallibleExceptionPointsSuspense = createInfallibleSuspenseCache(
```
bvaughn, line 29: These wrapper functions always bite us. Removing another one.

bvaughn:
e2e test caught a regression with Exceptions (`console_errors.ts`) so I'm going to dig into that.

packages/replay-next/components/console/LoggablesContext.tsx
```diff
@@ -174,58 +174,23 @@ function LoggablesContextInner({
...
-        }
-      }
-    });
+  const { value: exceptions = EMPTY_ARRAY } = useImperativeIntervalCacheValues(
+    exceptionsCache.pointsIntervalCache,
+    focusRange ? BigInt(focusRange.begin.point) : "0",
+    focusRange ? BigInt(focusRange.end.point) : endpoint,
+    client,
+    showExceptions
+  );
```
bvaughn, line 183: e2e test caught a regression caused by this. need to dig into it.

bvaughn: Fortunately this was a very small thing.

packages/replay-next/components/console/hooks/usePointInstances.ts
```diff
@@ -0,0 +1,162 @@
...
+      const pointBehavior = pointBehaviors[point.key];
+      if (pointBehavior?.shouldLog === POINT_BEHAVIOR_ENABLED) {
+        const cachedValue = hitPointsForLocationCache.getValueIfCached(
+          client,
+          toPointRange(focusRange),
+          point.location,
+          point.condition
+        );
+        if (cachedValue != null) {
+          keys.push(point.key);
```
bvaughn, line 155: Delta test (`should-support-conditional-log-points.ts`) caught a small regression in this also. Will address shortly.

packages/replay-next/components/console/hooks/usePointInstances.ts
```diff
@@ -0,0 +1,162 @@
...
+import { ReplayClientContext } from "shared/client/ReplayClientContext";
+import { POINT_BEHAVIOR_ENABLED, Point, ReplayClientInterface } from "shared/client/types";
+import { toPointRange } from "shared/utils/time";
+
+const EMPTY_ARRAY: PointInstance[] = [];
+
+// Transform Points (source location) to PointInstances (hit points / execution points for the source location)
+// Each Point maps to zero or more hit points
+//
+// This information needs to be
```
hbenl, line 15: needs to be what?

packages/replay-next/components/console/hooks/usePointInstances.ts
```diff
@@ -0,0 +1,162 @@
...
+        points,
+      })
+  );
+
+  // Lazily fetch any hit points that weren't already fetched yet
+  // useSyncExternalStore will ensure this component re-renders once values have loaded
+  useEffect(() => {
+    if (focusRange !== null) {
+      points.forEach(point => {
+        hitPointsForLocationCache.readAsync(
```
hbenl, line 73: Shouldn't we filter out the disabled points here as well?

hbenl (approved):
Thanks for the explainer loom, that was very helpful.

packages/replay-next/components/console/hooks/usePointInstances.ts
bvaughn: Uh, oops :)

packages/replay-next/components/console/hooks/usePointInstances.ts
bvaughn: Ah, yes. Good catch.

PR #9533, opened 2023-07-25
Add "processing..." loading page

jonbell-lot23:
If something has already been processed, we use the standard behaviour:

* "Loading..."
* 5 seconds later: (Random message)
* 10 seconds later: "This is taking too long"

image

But with this change, if something has not been processed, we only show one state:

* "Processing..."

image

We can update the look and feel of this in later PRs. This is just to get our "isProcessed" code working.


jasonLaster (commented):
Nice!

jasonLaster (commented):
Nice!

jonbell-lot23:
Also @bvaughn the next thing I want to do is get this same "if isProcessed, do something else" logic working in TestResultListItem.tsx . Here's a one minute Loom where I'm having trouble making it work. Any suggestions? https://www.loom.com/share/a074670f69834a43951893e0c0431ece

src/ui/hooks/recordings.ts
```diff
@@ -282,6 +282,7 @@ export function convertRecording(
     title: "title" in rec ? rec.title : undefined,
     duration: rec.duration,
     private: "private" in rec ? rec.private : undefined,
+    isProcessed: "isProcessed" in rec ? rec.isProcessed : undefined,
     isInitialized: "isInitialized" in rec ? rec.isInitialized : undefined,
```
bvaughn, line 286: What's the diff between processed and initialized?

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -97,31 +97,45 @@ function LoadingScreen({
...
   stalledTimeout = 15000,
-}: PropsFromRedux & { fallbackMessage: string; stalledTimeout?: number }) {
+}: PropsFromRedux & { fallbackMessage: string; isProcessed?: boolean; stalledTimeout?: number }) {
   const longWaitMessage =
     "<div><p>This is taking longer than expected.</p><p><a href='http://replay.io/discord' target='discord'>Contact us on Discord</a></p>";
 
+  const processingMessage = "Processing...";
+
   const [message, setMessage] = useState(fallbackMessage);
+  const [processing, setProcessing] = useState(isProcessed);
```
bvaughn, line 109: Why not immediately show the processing message? (Why wait for an effect to be run?)
```suggestion
  const [message, setMessage] = useState(isProcessed ? fallbackMessage : "Processing...");
  const [processing, setProcessing] = useState(isProcessed);
```

Seems like it would be better for us to avoid the flicker.

bvaughn:
@jonbell-lot23 The GraphQL query fetching data for the Library page is here:

https://github.com/replayio/devtools/blob/f34219cc0d0d12825c39259079fa116b420912c2/src/ui/components/Library/Team/View/TestRuns/graphql/TestRunsGraphQL.ts#L15-L42

Seems like you'll need to add the `isProcessed` attribute to that query if you want to read it from the Library page.

src/ui/hooks/recordings.ts
jonbell-lot23: Not actually sure — can investigate

jonbell-lot23:
Status update of different things happening at once:

* Adi pushed out some changes to how isProcessed works to reduce false negatives
* Brian suggested some changes to this code, but those changes are causing some errors I need to resolve
* Separately I'm working on the same functionality for our library. The code works, but I'm making new visuals. (We were using Material icons, but we want to stylize them in a way that Material can't do.)
* Jaril had done some of this work too, but we marked that as a dupe and I'm taking lead on it

So this PR is going to hang tight while I handle the library stuff (https://linear.app/replay/issue/DES-811/ship-mvp-processing-indicator-for-testsuites-rows#comment-e36e6af8), then I'll knock this out next

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -82,46 +80,47 @@ const phrases = [
...
+      const stalledTimeoutId = setTimeout(() => {
+        setMessage(longWaitMessage);
+      }, stalledTimeout);
+
+      return () => {
+        clearTimeout(phraseTimeout);
+        clearTimeout(stalledTimeoutId);
+      };
+    } else {
+      setMessage("Processing...");
```
bvaughn, line 121: This isn't necessary. State should already be "Processing..."

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -82,46 +80,47 @@ const phrases = [
...
 
 function LoadingScreen({
   uploading,
   awaitingSourcemaps,
   fallbackMessage,
-  stalledTimeout = 15000,
-}: PropsFromRedux & { fallbackMessage: string; stalledTimeout?: number }) {
+  isProcessed,
+  stalledTimeout = 20000,
+}: PropsFromRedux & { fallbackMessage: string; isProcessed?: boolean; stalledTimeout?: number }) {
```
bvaughn, line 98: `isProcessed` should be a _required_ prop to prevent the loading screen from being rendered with an invalid message.
```suggestion
}: PropsFromRedux & { fallbackMessage: string; isProcessed: boolean; stalledTimeout?: number }) {
```

bvaughn: We might be rendering this loader in places where we don't yet have the recording (so we can't know that) in which case I think maybe it would be appropriate to make a higher level change:

1. Move the message logic out of the loading component and make it a prop that's passed in<sup>1</sup> and remove the `isProcessed` prop. (We could make this a custom hook (e.g. `useLoadingMessage(isProcessed: boolean): string`) to encapsulate it a bit more, especially if we need to use it in more than one place.)

```ts
function LoadingScreen({ message }: { message: string }) {
```

2. Pass an empty string (`""`) in for places like `pages/_app.tsx` (where we don't have any info, and we only expect the loader to show for a brief moment)

bvaughn: Hey @jonbell-lot23 The change in 0249270 is not the full thing I mentioned above.

jonbell-lot23:
Thanks for your help, @bvaughn ! Here's [the diff](https://github.com/replayio/devtools/pull/9533/commits/02492700bf23c839d287d00cc72067c93bb0524c) I just did, which changed two things:

* isProcessed is now a required prop
* The unnecessary else statement has been removed

I just checked on two links that I know are processed and unprocessed (these are localhost links):

* [Unprocessed](http://localhost:8080/recording/94d11119-0720-4c1c-b9a3-c436d0894170?e2e=&apiKey=) (string should say "Processing...")
* [Processed](http://localhost:8080/recording/cypressintegrationallawsinstance-rootspects--33557a1d-e70c-4aad-a13d-5d2af796be5e?e2e=&apiKey=) (string should say "Loading...")

They're showing the right strings, and the loading one does the dynamic messages correctly. So I think this is good to go, with some caveats:

* Once this is live, I'm going to turn around and get the [testsuites PR](https://github.com/replayio/devtools/pull/9537) ready to go
* This isn't the final design/copy for the processing screen yet, this is just getting my foot in the door and then we'll iterate on the finer details next
* Watching an unprocessed replay should process it, but I'm not sure it is. That's something I'll bring up with @callingmedic911 
* Thanks for your help!

bvaughn:
@jonbell-lot23 I think there's some misunderstanding here. Did you see this comment? https://github.com/replayio/devtools/pull/9533#discussion_r1274878583

Basically making the prop required just reveals all of the places we need to pass it that we aren't (like `_app`)

jonbell-lot23:
> @jonbell-lot23 I think there's some misunderstanding here. Did you see this comment? [#9533 (comment)](https://github.com/replayio/devtools/pull/9533#discussion_r1274878583)
> 
> Basically making the prop required just reveals all of the places we need to pass it that we aren't (like `_app`). That's why CI is failing the Typecheck task for example.

Sorry for the confusion. I saw your comment but I am hoping we can do the broader logic changes in a follow-up. That was what I was referring to in the last bulletpoint.

bvaughn:
I think there are two things at this point:
1. The PR in its current state is broken. (See the TypeScript CI failure).
2. I'm not really comfortable landing this incremental change, (especially not with a plan to address "later").

If you don't have the bandwidth to make these changes, put it in FE triage?

jonbell-lot23:
> Sorry for the confusion. I saw your comment but I am hoping we can do the broader logic changes in a follow-up. That was what I was referring to in the last bulletpoint.

... but never mind. I was having trouble understanding and implementing your requested changes, but let's assume this PR isn't ready until I figure that bit out.

jonbell-lot23:
> I think there are two things at this point:
> 
> 1. The PR in its current state is broken. (See the TypeScript CI failure).
> 2. I'm not really comfortable landing this incremental change, (especially not with a plan to address "later").
> 
> If you don't have the bandwidth to make these changes, put it in FE triage?

(Our messages passed each other)

I understand. I have the bandwidth, it's just a matter of figuring it out.

jonbell-lot23:
Ok @bvaughn , I think this is ready for review now. I hope I addressed everything well! 🤞

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -44,91 +44,18 @@ export function LoadingScreenTemplate({ children }: { children?: ReactNode }) {
...
-  useEffect(() => {
-    const stalledTimeoutId = setTimeout(() => {
-      // after 15 seconds, switch to stalled message
-      setMessage(longWaitMessage);
-    }, stalledTimeout);
-    return () => clearTimeout(stalledTimeoutId);
-  }, [stalledTimeout]);
-
+  message,
+}: PropsFromRedux & { message: string }) {
```
bvaughn, line 51: This is a nice change!

I wonder how involved it would be for us to remove the rest of the `PropsFromRedux` props. I don't know if they're relevant for most places this Loading panel is used.

🤷🏼  We can do later.

PR #9534, opened 2023-07-25
Update oss dependencies

bvaughn:
- [x] Update `react-resizable-panels` and `suspense` packages to the latest version (to fix a Preconstruct issue where some export conditions were missing files in the bundle).
- [x] Update Next.js to the latest version as well (minor version bump).


bvaughn:
I think this change broke Replay Gecko (`SyntaxError: class static blocks are not currently supported`)

Yuck. Reverted in c12c6f567b3e2447a4c992d4de6aa900350aca94

PR #9536, opened 2023-07-25
Reduce findPoints and getHitCounts protocol requests

bvaughn:
### [View Loom overview](https://www.loom.com/share/953dcf2ea49b416485af26ef81c994b3)

Fix protocol regressions from #9492

See my notes on FE-1750 for a more in-depth accounting of what I looked into and why I changed these two pieces of code.


bvaughn:
I feel good enough about this change and it's positive impact on production that I'm likely going to land it as soon as CI says it's okay. I will happily make follow up changes based on async PR feedback.

PR #9540, opened 2023-07-27
Fix ScopesInspector performance

hbenl:
All I had to do was prefetch the scope values *and move the Suspense boundary inside the Expandable*.
¯\\\_(ツ)\_/¯


packages/replay-next/components/inspector/ScopesInspector.tsx
```diff
@@ -23,23 +25,49 @@ export default function ScopesInspector({
...
+  protocolValues: NamedValue[];
+}) {
+  const client = useContext(ReplayClientContext);
+  suspendInParallel(
+    ...protocolValues.map(value => () => {
+      if (value.object) {
+        objectCache.read(client, pauseId, value.object, "full");
+      }
+    })
+  );
```
bvaughn, line 58: This change baffles me 🤔

The `KeyValueRenderer` and `Value` components which eventually render inside of the `Inspector` should also call `objectCache.read` with either `"full"` or `"canOverflow"` so...I'm not sure why this change would have any impact.

This bit looks a bit sketchy to me:

https://github.com/replayio/devtools/blob/606849734dca3a441b128a908adf1e400ff0a39c/packages/replay-next/components/inspector/ValueRenderer.tsx#L81-L86

`getCachedObject` might return a cached value for `"canOverflow"` or `"none"` even though the `noOverflow` value may not allow that. Given the way our caches work, I _think_ we should remove that `getCachedObject` method entirely and just ask the cache for the type of value we want.

I'm curious if that change would make this one unnecessary? How easily can you repro the original freezing bug?

packages/replay-next/components/inspector/ScopesInspector.tsx
```diff
@@ -23,23 +25,49 @@ export default function ScopesInspector({
...
-      </Suspense>
+      <Expandable
+        className={styles.Expandable}
+        defaultOpen={expandByDefault}
+        header={name}
+        persistenceKey={path}
+      >
+        <Suspense fallback="Loading...">
+          <ScopesInspectorChildren path={path} pauseId={pauseId} protocolValues={protocolValues} />
+        </Suspense>
```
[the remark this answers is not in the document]
hbenl: Actually after some more experimentation I found that this isn't necessary to fix the performance. But previously I had placed the prefetching in `ScopesInspector` and in that case moving the `Suspense` boundary inside the `Expandable` fixed the performance (which I found super baffling).
So while this isn't necessary for performance I want to keep it because the `Expandable` shouldn't disappear while its children are loading.

packages/replay-next/components/inspector/ScopesInspector.tsx
hbenl: By prefetching here we don't try to render the children until the data has arrived, this prevents the overrendering of the children that we were seeing.

> This bit looks a bit sketchy to me: [...]

Agreed, I want to remove that `getCachedObject` call.

> I'm curious if that change would make this one unnecessary?

I just tried and instead of making this change unnecessary it undoes the performance gains of this change :open_mouth:
I'll investigate some more, perhaps I find something that lets me finally understand this issue. 

> How easily can you repro the original freezing bug?

That's easy: go to http://localhost:8080/recording/react-app--cf8417fc-8c82-42df-8c56-94356a172522?point=324518553672943161447660830851097&time=239 and open the bottom scope. For me it takes more than 5 seconds without this fix.

hbenl: So the `ValueRenderer` uses other renderers, some of which (like `ArrayRenderer`) use `ValueRenderer` for properties of the scope values. Those haven't been prefetched but have been cached without a preview, so when I remove the `getCachedObject` call, the `ValueRenderer` suspends for those properties and then we run into the overrendering issue again.
In other words: it's not enough to prefetch the scope values, we'd also have to prefetch the properties of some of those scope values...

packages/replay-next/components/inspector/ScopesInspector.tsx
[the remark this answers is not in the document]
bvaughn: > I want to keep it because the `Expandable` shouldn't disappear while its children are loading.

Yeah, I agree.

packages/replay-next/components/inspector/ScopesInspector.tsx
bvaughn: Thanks for explaining. I think I understand what's going on better now. Essentially it's like a "waterfall" where we fetch 5 things, then those 5 things cause us to fetch 25 more things– and in between, we're rendering and suspending and rendering and suspending. That's definitely not optimal. Changing so that we _suspend_ here to fetch the data before rendering further batches it up so we don't do all of the incremental rendering in between.

bvaughn (approved):
I feel really good about this change after discussing it.

The only thing I'd ask is that we add an inline comment above where we call `suspendInParallel` to explain why we're doing it.

PR #9550, opened 2023-07-31
Migrate Socket to a class

jasonLaster:
This is an in place migration that moves module state into a class. We're still creating a singleton and  exporting individual functions, so it's not a huge win, but a step in the right direction

### DIFF
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -152,26 +152,26 @@ export class ProtocolSession {
       window.addEventListener("beforeunload", () => {
         this.willClose = true;
       });
-    }
 
-    (window as any).disconnect = () => {
-      this.socket?.close();
-    };
-
-    (window as any).protocolClient = this.client;
-
-    (window as any).outstanding = () => {
-      const messages = Array.from(this.messageWaiters.entries()).map(([id, { method }]) => ({
-        id,
-        method,
-      }));
-      return {
-        messages,
-        received: this.receivedBytes,
-        sent: this.sentBytes,
-        time: Date.now() - this.startTime,
+      (window as any).disconnect = () => {
+        this.socket?.close();
       };
-    };
+
+      (window as any).protocolClient = this.client;
+
+      (window as any).outstanding = () => {
+        const messages = Array.from(this.messageWaiters.entries()).map(([id, { method }]) => ({
+          id,
+          method,
+        }));
+        return {
+          messages,
+          received: this.receivedBytes,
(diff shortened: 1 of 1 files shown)


bvaughn:
Ooo, I'm curious– what's the motivation for moving from a module to singleton class? Aren't they kind of the same? (Do you think we'll have multiple socket instances at some point?)

bvaughn:
Can I read more about that idea somewhere? I think there are **a lot of technical hurdles** that we'd have to jump over (beyond the socket) for that to happen. My intuition would not be to try to show them both within a single Replay instance.

Mokshit06:
I'm not sure if it is written in detail anywhere (maybe @jlast can help). Do you have any ideas on how else we could make it easier to inspect tests like these without showing them together?

bvaughn:
Cleaning up the PR queue by closing old/stale PRs.

jasonLaster:
re-opening because mokshit pushed a fix two days ago that makes this PR landable

bvaughn:
@jasonLaster I still don't understand the main point of this PR/change.

markerikson:
@bvaughn : Jason and Mokshit have the idea that we could have 2 (or more) simultaneous active sessions for the same recording, as a starting point for multiplexing some of our UI display and allowing users to compare behaviors.  I believe this is meant as a first step towards that idea.

bvaughn:
>  Jason and Mokshit have the idea that we could have 2 (or more) simultaneous active sessions for the same recording, as a starting point for multiplexing some of our UI display and allowing users to compare behaviors. I believe this is meant as a first step towards that idea.

@markerikson @jasonLaster Yup yup. Mokshit and I discussed that in the thread above.

I believe that would be a **major architectural change for the frontend** though, so if that's the only reason we're making this change, it might not be worth making it at the moment.

jasonLaster:
Thanks for pointing that out. We should first fix the tests and then do the review.

re goals: it'd be great to explore what would be involved in having multiple sessions. Would the socket belong to the replay client? Hard to say... Even without that though, I think this will help remove a code smell and make other projects possible like the testing harness and sandbox i prototyped over the past two months.

bvaughn:
> it'd be great to explore what would be involved in having multiple sessions.

This is probably something we shouldn't even think about that much in h2, IMO.

bvaughn:
Just doing some tidying up and it looks like this PR is stale. Closing for now but if anyone wants to rebase and look into the failing tests, we can re-open it.

PR #9553, opened 2023-07-31
Fix should-render-getters-and-setters-correctly snapshot test

hbenl:


hbenl:
This attempt to fix the test failed.

PR #9556, opened 2023-07-31
remove old experimental settings

miriambudayr:


packages/protocol/socket.ts
```diff
@@ -121,21 +121,18 @@ const noCallerStackTracesForFailedCommands = new Set<CommandMethods>([
...
-  disableUnstableQueryCache?: boolean;
   enableRoutines?: boolean;
-  rerunRoutines?: boolean;
-  profileWorkerThreads?: boolean;
-  disableRecordingAssetsInDatabase?: boolean;
   keepAllTraces?: boolean;
-  disableIncrementalSnapshots?: boolean;
-  disableConcurrentControllerLoading?: boolean;
-  disableProtocolQueryCache?: boolean;
+  listenForMetrics: boolean;
```
miriambudayr, line 133: Sorry there is a bunch of diff noise here --- I wanted to sort the lines because it was bothering me.

bvaughn: yay 🥳  I often do the same, no worries

bvaughn (approved):
The naming of these two flags always confused me 😄  glad they're gone

PR #9577, opened 2023-08-02
Better processing copy

jonbell-lot23:
image

I'm pretty excited about this. It's just a little bit of copy, but despite no progress bar, percentage indicator, or time estimate, it still helps users understand why things may be taking longer than they'd expect.

This only shows for "processing" flows. Our standard loading flows will still look the same.

### DIFF
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -48,17 +48,21 @@ function LoadingScreen({
   uploading,
   awaitingSourcemaps,
   message,
-}: PropsFromRedux & { message: string }) {
+  additionalInfo,
+}: PropsFromRedux & { message: string; additionalInfo: string }) {
   const waitingForMessage =
     awaitingSourcemaps || uploading ? (
       <span>Uploading {Math.round(uploading?.amount ? Number(uploading.amount) : 0)}Mb</span>
     ) : (
-      <span dangerouslySetInnerHTML={{ __html: message }}></span>
+      <>
+        <div className={styles.message} dangerouslySetInnerHTML={{ __html: message }} />
+        <div className={styles.additionalInfo}>{additionalInfo}</div>
+      </>
     );
 
   return (
     <LoadingScreenTemplate>
-      <span className={styles.messageWrapper} dangerouslySetInnerHTML={{ __html: message }}></span>
+      <div className={styles.messageWrapper}>{waitingForMessage}</div>
     </LoadingScreenTemplate>
   );
 }
--- a/src/ui/components/shared/LoadingScreen.tsx
+++ b/src/ui/components/shared/LoadingScreen.tsx
@@ -49,14 +49,14 @@ function LoadingScreen({
   awaitingSourcemaps,
   message,
   additionalInfo,
-}: PropsFromRedux & { message: string; additionalInfo: string }) {
+}: PropsFromRedux & { message: string; additionalInfo?: string }) {
   const waitingForMessage =
     awaitingSourcemaps || uploading ? (
       <span>Uploading {Math.round(uploading?.amount ? Number(uploading.amount) : 0)}Mb</span>
     ) : (
       <>
         <div className={styles.message} dangerouslySetInnerHTML={{ __html: message }} />
-        <div className={styles.additionalInfo}>{additionalInfo}</div>
+        {additionalInfo && <div className={styles.additionalInfo}>{additionalInfo}</div>}
       </>
     );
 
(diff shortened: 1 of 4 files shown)


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-828/slightly-better-processing-text

jasonLaster:
This is an improvement, but I'm also a half measure because awhile could be anything between 30s and 10mins.

Let's land this and aim to do better soon.

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -48,17 +48,21 @@ function LoadingScreen({
   uploading,
   awaitingSourcemaps,
   message,
-}: PropsFromRedux & { message: string }) {
+  additionalInfo,
+}: PropsFromRedux & { message: string; additionalInfo?: string }) {
   const waitingForMessage =
```
bvaughn, line 53: Was this an intentional change too? It isn't mentioned in the PR, but now we'll be using the "uploading" message again. (Before, we weren't.)

src/ui/hooks/useDynamicLoadingMessage.ts
```diff
@@ -53,6 +53,11 @@ export function useDynamicLoadingMessage(
   stalledTimeout: number
 ) {
   const [message, setMessage] = useState(isProcessed ? initialMessage : "Processing...");
+  const [additionalInfo, setAdditionalInfo] = useState(
+    isProcessed
+      ? ""
+      : "This could take a while, depending on the complexity and length of the replay."
+  );
```
bvaughn, line 60: This does not need to be in state. It's never updated. It's just a value derived from a prop.

bvaughn (commented):
"additional info" is kind of an ambiguous term.

Seems like `primaryMessage` and `secondaryMessage` (or `header` and `subHeader`) would be more semantically meaningful prop names.

src/ui/components/shared/LoadingScreen.tsx
jonbell-lot23: Sorry, I'm not seeing the uploading change you're describing. Looks like the last change on that line was 2 months ago?

jonbell-lot23:
> "additional info" is kind of an ambiguous term.
> 
> Seems like `primaryMessage` and `secondaryMessage` (or `header` and `subHeader`) would be more semantically meaningful prop names.

Changed to message and secondaryMessage.

I've been having semantic debates for over 20 years, and I know sometimes it comes down to personal taste. I don't think noun-integer is always necessarily the most semantically sound approach. I do think message + additionalInfo is appropriate for a dialog, moreso than "header" (it's not a header) or "subheader" (it's not a subheader). But this is a subjective opinion and I'm happy to align.

src/ui/components/shared/LoadingScreen.tsx
bvaughn: I'm referring to the `waitingForMessage` change. Before it was being declared but not referenced:

https://github.com/replayio/devtools/blob/856be645dbbe1a0386d4ab1f05cbfae958837d16/src/ui/components/shared/LoadingScreen.tsx#L47-L64

Now it's being referenced:

https://github.com/replayio/devtools/blob/b26e6b013386998f963c9fec3e103936d206a70b/src/ui/components/shared/LoadingScreen.tsx#L53-L68

bvaughn (approved):
> I don't think hierarchy->noun is always necessarily the most semantically sound approach. I do think message + additionalInfo is appropriate for a dialog, moreso than header or subheader. But this is a subjective opinion, as semantic debates usually are, and I'm happy to align.

I want to try explaining my thinking a bit further so it doesn't seem like it's purely a wording nit.

To me, names like "primary and secondary message" clearly convey their use. (Same with "header and sub header".) I don't even have to look at the component to know that (a) they're strings and (b) how they're likely presented, visually.

"additional info" doesn't even sound like it's necessarily text to me. In JavaScript, a term like "info" is often used to refer to an `Object`. (For example, [React's `componentDidCatch` method](https://react.dev/reference/react/Component#componentdidcatch) accepts and `info` param that's an object.)

In a larger context, not specifically focused on parameters, I agree with you that "additional info" as a term would be synonymous.

jonbell-lot23:
That totally makes sense and I appreciate the explanation!

src/ui/components/shared/LoadingScreen.tsx
jonbell-lot23: Ah ha, I see it now. Yeah, I think I probably removed that in a previous ticket without realising, then re-stitched it back together in this one.

PR #9579, opened 2023-08-03
RDT operations unit test

bvaughn:
Proposed changes to testing approach introduced in #9576 for discussion.

I think this approach offers a few advantages:
* The actual test code is easier to read (since it's just normal React code)
* This is more of an "end-to-end" test (in that it uses RDT to generate the original operations arrays)
* If we _did_ end up needing to write more unit tests around RDT, this would put us in a better place to do it more easily
* Should be easier to spot and support future changes to RDT operations array 

It also has a downside:
* The setup and teardown logic is kind of complicated


src/ui/components/SecondaryToolbox/react-devtools/__tests__/rdtProcessing.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      0, // string table (empty)
+
+      TREE_OPERATION_UPDATE_TREE_BASE_DURATION,
+      123,
+
+      TREE_OPERATION_REMOVE_ROOT,
+    ];
+
+    verifySerialization([operationsArray]);
+  });
```
bvaughn, line 93: These are hard to trigger from the outside 😄 

I think we could trigger the `TREE_OPERATION_REMOVE_ROOT` by setting `store.componentFilters` but...that's not really any better than just hard-coding it.

src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts
```diff
@@ -0,0 +1,160 @@
...
+export function verifySerialization(operationsArrays: number[][]) {
+  operationsArrays.forEach(operationsArray => {
+    expect(operationsArray.length > 0).toBeTruthy();
+
+    const deconstructed = deconstructOperationsArray(operationsArray);
+    const reconstructed = reconstructOperationsArray(deconstructed);
+
+    expect(reconstructed).toEqual(operationsArray);
+  });
+}
```
bvaughn, line 39: I think these helper functions could be used in the #9576 approach even if we don't go with this PR.

markerikson:
Played around with this and found some aspects that are missing. The `assertContainsOperation` method isn't actually asserting, just returning a boolean, so those tests were passing incorrectly. The errors test isn't actually generating an errors operation. 

I think this is a very potentially useful technique, and if we need to add additional testing for the logic later I can look at using this approach at that time. For now, though, what I've got in #9576 does what I want in terms of exercising the actual cases.  I'll make the couple tweaks to fix parsing the `profilingFlags` field and cleanup over there, and put this one in my back pocket for later reference.

bvaughn:
FWIW the fix (pushed) was this:
```diff
diff --git a/src/ui/components/SecondaryToolbox/react-devtools/__tests__/rdtProcessing.test.tsx b/src/ui/components/SecondaryToolbox/react-devtools/__tests__/rdtProcessing.test.tsx
index b6e9bcf72..f94c0bf7f 100644
--- a/src/ui/components/SecondaryToolbox/react-devtools/__tests__/rdtProcessing.test.tsx
+++ b/src/ui/components/SecondaryToolbox/react-devtools/__tests__/rdtProcessing.test.tsx
@@ -52,10 +52,13 @@ describe("RDT processing", () => {
   });
 
   test("should support warnings", () => {
+    const ComponentThatWarns = () => {
+      console.warn("This is a warning");
+      return null;
+    };
+
     // Duplicate key warning
-    const operationsArray = getOperations(() =>
-      root.render(<>{[<Component key="C" />, <Component key="C" />]}</>)
-    );
+    const operationsArray = getOperations(() => root.render(<ComponentThatWarns />));
 
     assertContainsOperationType([operationsArray], TREE_OPERATION_UPDATE_ERRORS_OR_WARNINGS);
     verifySerialization([operationsArray]);
diff --git a/src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts b/src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts
index f56bfacda..19ff2d7da 100644
--- a/src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts
+++ b/src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts
@@ -108,6 +108,9 @@ export function reactDevToolsBeforeEach() {
     bridge: backendBridge,
   });
 
+  // Flush activate commands
+  jest.runAllTimers();
+
   // Reset modules so that importing/requiring react-dom below will register with the DevTools hook
   jest.resetModules();
 
```

bvaughn:
For when/if we circle back to this– I replaced the hard-coded `TREE_OPERATION_UPDATE_TREE_BASE_DURATION` and `TREE_OPERATION_REMOVE_ROOT` operations with ones that go through RDT.

I also added `toMatchInlineSnapshot` to verify the contents of the de-structured objects array (so this approach didn't _lose_ any coverage).

bvaughn:
Rebasing

PR #9585, opened 2023-08-04
Use point instead of time to set focus boundaries

bvaughn:


src/ui/components/NetworkMonitor/useNetworkContextMenu.tsx
```diff
@@ -33,15 +33,15 @@ export default function useNetworkContextMenu({ row }: { row: Row<RequestSummary
...
   const { copy: copyAsCURL, state } = useCopyAsCURL(requestSummary);
 
-  const beginTime = requestSummary.start;
-  const endTime = requestSummary.end;
+  const beginTime = requestSummary.start!;
+  const endTime = requestSummary.end!;
 
   const setFocusEnd = () => {
-    dispatch(setFocusWindowEndTime(endTime!, true));
+    dispatch(setFocusWindowEnd({ time: endTime, sync: true }));
```
bvaughn, line 40: I _think_ this is the only place that doesn't have an explicit execution point to pass in. If we could address that, we could _require_ that value and simplify the actions.

markerikson (approved):

PR #9586, opened 2023-08-04
Fill out RDT E2E tests

markerikson:
This PR:

- Renames the existing FF-based RDT E2E test to `react-devtools-01-basic`
- Makes a Chromium-based copy of that RDT-01 basic test
  - Copies the CRA example app `index.html` file so we have a unique filename to use for the test
  - Adds a Chromium recording of that CRA example
  - Also removes a dead Cypress example recording ID
- Updates the test startup logic to allow passing in query params via a `URLSearchParams` instance
- Updates both the FF and Chromium basic tests
  - Updates the Chromium version to use getters instead of properties for reading the DOMRect coordinates
  - Adds more steps for jumping back and forth between points, to test out the client-side "reset the tree whenever you pause" behavior
  - Updates the Chromium version to force re-running the RDT routine every time the test runs, via a preferences query param
- Bumps the timeout in our "wait for stack frames to confirm we're paused" helper from 10s to 15s, just because I see that still happening somewhat frequently
- Updates the React DevTools panel component:
  - Adds logic to reset the Redux "is this node picker active?" state when the RDT tab is hidden, which fixes some timing issues coordinating with the component picker toggle button inside the actual RDT frontend UI component
  - Renames some of the `ReplayWall` internal methods for clarity
- Updates the RDT bundle injection logic to match the routine behavior:
  - Updated the RDT bundle to the current latest used with the routine
  - Copied the logic from the routine for ensuring all current fibers have generated IDs
- Adds another Chromium-based RDT E2E test that exercises additional functionality:
  - Added an example recording ID of one of our FE `breakpoints-01` recordings for use as the target
  - Added a test that always runs the RDT routine, checks for sourcemapped component names, exercises "Jump to Component Source", and checks how the RDT UI handles `<Suspense>` and `<Offscreen>` components


hbenl (approved):

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -521,6 +543,13 @@ export function ReactDevtoolsPanel() {
         },
       };
       nodePickerInstance.enable(actualOpts);
+      // HACK There _may_ be an issue with React's `<Offscreen>` and dispatched Redux actions.
+      // I can see the "picker active" selector re-running afterwards, but the component doesn't re-render.
+      // It seems like it thinks the value is _already_ `true`, when it shouldn't have changed yet.
+      // To work around this, we force a re-render of the component after enabling the picker.
+      // This is incredibly hacky and I don't feel like it actually _solves_ whatever issue
+      // may actually be there... but it does at least let the E2E test pass.
```
bvaughn, line 551: Can you file a Linear issue for follow up?

PR #9593, opened 2023-08-09
Add a dismiss button to the Passport

jonbell-lot23:
Loom walkthrough:
https://www.loom.com/share/f994175ed26444929e96e65b3dbe0a9c

Screenshot:
image

### DIFF
--- a/src/ui/components/shared/Modals/PassportDismissModal.tsx
+++ b/src/ui/components/shared/Modals/PassportDismissModal.tsx
@@ -1,0 +1,68 @@
+import classNames from "classnames";
+import React from "react";
+import { ConnectedProps, connect } from "react-redux";
+
+import { userData } from "shared/user-data/GraphQL/UserData";
+import * as actions from "ui/actions/app";
+import { setSelectedPrimaryPanel } from "ui/actions/layout";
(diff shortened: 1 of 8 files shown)


src/ui/components/shared/Modals/PassportDismissModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+              Cancel
+            </SecondaryButton>
+          </div>
+        </DialogActions>
+      </Dialog>
+    </Modal>
+  );
+}
+
+const connector = connect(
```
markerikson, line 69: :hand: I'll nitpick and say we really shouldn't be writing any more components that use `connect` (we should be using the React-Redux hooks instead), but I realize this was likely a pattern copy-pasted from our other dialogs (most of which have been around for a couple years) and this shouldn't block the PR.

markerikson (approved):

src/ui/components/shared/Modals/PassportDismissModal.tsx
jonbell-lot23: Good to know, thank you! Will follow-up in DM to make sure I learn more about the new way of doing things.

PR #9594, opened 2023-08-09
Use playwright annotations

hbenl:


bvaughn:
Is this PR something we want to pursue? Looks like it's gotten stale.

PR #9595, opened 2023-08-10
Add NUX/Passport messaging for testsuites users

jonbell-lot23:
**Background**

We've long had a NUX flow for traditional replays that starts with "Larry" waving at you and walking you through some steps. Testsuites is a bit different, because we don't want to take over the Cypress/Playwright pane on the side.

**Summary**

image

This explains what the passport is with a little banner that dismisses as soon as you click away.

Loom walkthrough of me testing a few scenarios:
https://www.loom.com/share/dc2730266e064ea3a203cd4c14eaa5f7

### DIFF
--- a/src/ui/components/Passport/Passport.tsx
+++ b/src/ui/components/Passport/Passport.tsx
@@ -29,6 +29,8 @@ type PrimaryPanelName = "events" | "cypress" | string;
 const stepNames = ["step-one", "step-two", "step-three", "step-four"] as const;
 
 const Passport = (props: PropsFromRedux) => {
+  const [showWelcome, setShowWelcome] = useState(localStorage.getItem("passportNUX") !== "true");
+
   const recordingId = useGetRecordingId();
   const { recording } = useGetRecording(recordingId);
   const [selectedIndices, setSelectedIndices] = useState({ sectionIndex: 0, itemIndex: 0 });
@@ -51,6 +53,12 @@ const Passport = (props: PropsFromRedux) => {
   type StepNames = (typeof stepNames)[number];
   const videoExampleRef = useRef<HTMLImageElement>(null);
   const [videoHeight, setVideoHeight] = useState<number | null>(null);
+
+  useEffect(() => {
+    if (showWelcome) {
+      localStorage.setItem("passportNUX", "true");
+    }
+  }, [showWelcome]);
 
   useLayoutEffect(() => {
     const videoExample = videoExampleRef.current;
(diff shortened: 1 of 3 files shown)


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-837/testsuites-passport-create-welcome-screen-for-nux

src/ui/components/Passport/Passport.tsx
```diff
@@ -29,6 +30,10 @@ type PrimaryPanelName = "events" | "cypress" | string;
 const stepNames = ["step-one", "step-two", "step-three", "step-four"] as const;
 
 const Passport = (props: PropsFromRedux) => {
+  const [showTestsuitesPassportWelcome, setshowTestsuitesPassportWelcome] = useState(
+    localStorage.getItem("TestsuitesPassportWelcome") !== "true"
```
markerikson, line 34: :hand: I think we want to avoid using `localStorage` directly at this point, now that Brian has moved all of our settings into the new `UserData` system.

Can you try defining the setting with that system instead? I believe it also has a hook you can use to retrieve and set the value.

markerikson (changes requested):

src/ui/components/Passport/Passport.tsx
jonbell-lot23: Roger Roger

src/ui/components/Passport/Passport.tsx
```diff
@@ -30,8 +32,8 @@ type PrimaryPanelName = "events" | "cypress" | string;
 const stepNames = ["step-one", "step-two", "step-three", "step-four"] as const;
 
 const Passport = (props: PropsFromRedux) => {
-  const [showTestsuitesPassportWelcome, setshowTestsuitesPassportWelcome] = useState(
-    localStorage.getItem("TestsuitesPassportWelcome") !== "true"
+  const [showTestsuitesPassportFirstRun, setShowTestsuitesPassportFirstRun] = useState(
```
markerikson, line 35: :hand: I think the `useState` here can be entirely replaced by the `useGraphQLUserData` hook from `packages/shared/user-data/`.

src/ui/components/Passport/Passport.tsx
```diff
@@ -58,10 +60,10 @@ const Passport = (props: PropsFromRedux) => {
   const [videoHeight, setVideoHeight] = useState<number | null>(null);
 
   useEffect(() => {
```
markerikson, line 62: :question: can you clarify what the intended logic flow is with this `useEffect` ?

jonbell-lot23: Sure – I want to show this banner only a single time. So if you go to the page, you'll see the banner, but then if you switch tabs and return (or hit reload) the tab should disappear.

So this code is meant to do that.

src/ui/components/Passport/Passport.tsx
jonbell-lot23: I'm sure you're right, but when I tried it the banner didn't show up because the logic was:

* Show the banner!
* Ok, now that we did that, userData.set should set this to false!
* ... meaning the banner doesn't appear, even on the first render

Whereas state lets it show on the first load before disappearing. Investigating now.

markerikson: Hmm. Okay, yeah, I think the fact that you are wanting to clear it out right away is the distinction here, and maybe the `useState` is necessary.

jonbell-lot23:
Ok, I think we're good to go! Thanks Mark!

markerikson (approved):

PR #9600, opened 2023-08-11
Use ❌ and ✅ in titles to reflect test result

jonbell-lot23:
image

Addresses https://linear.app/replay/issue/DES-847/replay-test-titles-should-include-a-pass-fail-emoji

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -110,7 +110,7 @@ function Routing({ Component, pageProps }: AppProps) {
       <Head>
         <meta httpEquiv="Content-Type" content="text/html; charset=utf-8" />
         <link rel="icon" type="image/svg+xml" href="/images/favicon.svg" />
-        <title>Replay</title>
+        <title>⏱️ Loading…</title>
       </Head>
       <ErrorBoundary>
         <_App>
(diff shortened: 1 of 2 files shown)


jasonLaster:
I wonder if we should replace the Replay favicon with a test icon? This has the benefit of always being there even with lots of tabs. Also looks nicer.

Should we also address the treatment in the title too or do you think we don't need it if we do the tab? My hunch is that the title still matters because it's in the app, the tab is a bit far to go.

jonbell-lot23:
> I wonder if we should replace the Replay favicon with a test icon? This has the benefit of always being there even with lots of tabs. Also looks nicer.
> 
> Should we also address the treatment in the title too or do you think we don't need it if we do the tab? My hunch is that the title still matters because it's in the app, the tab is a bit far to go.

Good thoughts. I agree that one icon looks nicer than two. But I'm not sure how I feel about logic/branding that says "sometimes we show the Replay icon and sometimes we don't." It feels like it may be a bit too clever and possibly even confusing.

The title is an interesting one. We have a slick feature for hovering over a title to edit it, and right now it's pixel perfect. As soon as we prepend with an emoji, it gets more complicated for users.

--

I think we should ship this as-is to get a quick win, learn things, and iterate from there. Whether we add both ideas, neither of them, or somewhere in between, I think we'll get better instincts for it trying this one for a few days.

jonbell-lot23:
Oh, I meant to talk about the title/tab thing:

To me, this feature addresses a very clear issue: it's hard to tell tabs apart. So I think addressing that is key.

Whereas if we think we can be more clear in the product itself, there are tons of things we can do to make that more clear. It doesn't have to be either/or.

jasonLaster:
Sounds good

markerikson (approved):

PR #9601, opened 2023-08-12
Playwright annotations

jasonLaster:
extends @hbenl's two PRs

1. https://github.com/replayio/devtools/pull/9594/files
2. https://github.com/replayio/playwright-stacks/pull/1/files#diff-841ff7831cd3c0b7ab2f1b287034c13d04afa2a309a09e34e90a2cf5d6cd6444

To show how you can use a Fixture to wrap the instrumentation API similar to what [playwright-test](https://github.com/microsoft/playwright/blob/main/packages/playwright-test/src/index.ts#L273) does and get `onApiBegin` and `onApiCallEnd` events.

We'll never land this PR, but it's a good proof of concept for what we will likely need to do in the playwright plugin side.

What are the next questions
1. What is the role of the reporter if we can get all of the information (stack traces, call params, ...) here and have one ID?
2. How will we setup the fixture? Do we need to tell users to use our test command? ...

Here's a [replay](http://localhost:8080/recording/breakpoints-02-test-unhandled-divergence-while-evaluating-at-a-breakpoint--a17fa476-423f-48a3-afb6-1db7d42af2e8?point=29855706958035288958493902227111936&time=11593) with annotations from the fixture.

Here's a [replay](https://app.replay.io/recording/replay-of-localhost8080--041b414b-3ffe-4c6a-b803-eed1479d85b2?point=28882151300558176690227923302481920&time=12543) of the replay that shows the annotations.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -26,6 +26,14 @@ const baseNextConfig = {
   },
   productionBrowserSourceMaps: true,
 
+  typescript: {
+    // !! WARN !!
+    // Dangerously allow production builds to successfully complete even if
+    // your project has type errors.
+    // !! WARN !!
+    ignoreBuildErrors: true,
+  },
+
   async redirects() {
     return [
       {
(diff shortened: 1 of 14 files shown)


jasonLaster:
It's fine to cancel since it was always a POC and the preview branch is still here

PR #9603, opened 2023-08-13
Add emoji to title when page is loading

jonbell-lot23:
[Loom demonstration](https://www.loom.com/share/916dcc5ea0e74efe8654896ed7c2e316)

image

Another way to make loading more clear for our customers. Pairs well with [DES-812](https://linear.app/replay/issue/DES-812/[explore]-add-a-browser-notification-when-loading-is-complete) (coming) and [DES-847](https://linear.app/replay/issue/DES-847/replay-test-titles-should-include-a-pass-fail-emoji) (shipped last week)

### DIFF
--- a/pages/recording/[[...id]].tsx
+++ b/pages/recording/[[...id]].tsx
@@ -66,11 +66,13 @@ function RecordingHead({ metadata }: MetadataProps) {
     description = "";
   }
 
+  const pageTitle = title ? `⏱️ ${title}` : "⏱️";
+
   const image = `${process.env.NEXT_PUBLIC_IMAGE_URL}${metadata.id}.png`;
 
   return (
     <Head>
-      <title>{title}</title>
+      <title>{pageTitle}</title>
       {/* nosemgrep typescript.react.security.audit.react-http-leak.react-http-leak */}
       <meta property="og:title" content={title} />
       <meta property="og:type" content="website" />
--- a/src/ui/components/Header/Header.tsx
+++ b/src/ui/components/Header/Header.tsx
@@ -1,212 +1,239 @@
-import assert from "assert";
-import { RecordingId } from "@replayio/protocol";
-import classNames from "classnames";
-import { ClipboardEvent, KeyboardEvent, useLayoutEffect, useRef, useState } from "react";
+import Head from "next/head";
+import { useRouter } from "next/router";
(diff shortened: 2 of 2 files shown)


pages/recording/[[...id]].tsx
```diff
@@ -66,11 +66,13 @@ function RecordingHead({ metadata }: MetadataProps) {
     description = "";
   }
 
+  const pageTitle = title ? `⏱️ ${title}` : "⏱️";
```
Comment (jasonLaster) [inline pages/recording/[[...id]].tsx:69]: How do you know that it's loading?

Comment (jonbell-lot23) [reply pages/recording/[[...id]].tsx:69]: This ends up being a placeholder because DevTools.tsx writes the final title once it loads:

```
  const title = recording?.title;
  const testResult = recording?.metadata?.test?.result;
  let emoji = "";
  switch (testResult) {
    case "passed":
      emoji = "✅";
      break;
    case "failed":
      emoji = "❌";
      break;
  }
  ```

**Comment by bvaughn** [comment]
Hey @jonbell-lot23: UI nit about this and the other emoji PR (#9600) – 

It looks weird to me to have an emoji right beside the Replay icon:

Screen Shot 2023-08-14 at 1 25 14 PM

They aren't the same visual style. They don't quite align (vertically)

Could we move these emojis to the right of the text at least? Maybe that would help a bit.

**Comment by jonbell-lot23** [comment]
> Hey @jonbell-lot23: UI nit about this and the other emoji PR (#9600) –
> 
> It looks weird to me to have an emoji right beside the Replay icon:
> 
> Screen Shot 2023-08-14 at 1 25 14 PM
> They aren't the same visual style. They don't quite align (vertically)
> 
> Could we move these emojis to the right of the text at least? Maybe that would help a bit.

I hear ya, but this is out of our control. I can't change the size or position of an emoji inside a title tag, it's handled by the browser. And putting it on the far right would crop it most situations unless you only have a few tabs open.

**Comment by bvaughn** [comment]
I don't think we should have landed this change using an emoji. The misalignment looks bad. Is there some non-emoji character we could use? What about something like `⧖`

**Comment by bvaughn** [comment]
If we expect _loading_ to always be fast, maybe we don't need this except for when we're _uploading_ (processing) in which case maybe we could use something more like this (once we have the progress API– aka later this week)

```
█▒▒▒ Title here
██▒▒ Title here
███▒ Title here
████ Title here
```

PR #9604, opened 2023-08-13
Add 4 UI polish improvements

jonbell-lot23:
[DES-843](https://linear.app/replay/issue/DES-843/testsuites-button-shouldnt-wrap) addresses our testsuites button wrapping awkwardly. Here's a video of the before and after:
https://github.com/replayio/devtools/assets/9154902/9c55a24e-b308-4875-b120-ee098f2f5658

[DES-840](https://linear.app/replay/issue/DES-840/replay-launcher-dialog-is-low-contrast) addresses our low contrast Replay launcher dialog:
image

[DES-841](https://linear.app/replay/issue/DES-841/privacy-modal-has-ugly-white-border) addresses our broken border on our privacy modal:
image

[DES-839](https://linear.app/replay/issue/DES-839/the-share-dialog-is-hard-to-see-in-dark-theme) addresses our share dialog border:
image

### DIFF
--- a/pages/recording/[[...id]].tsx
+++ b/pages/recording/[[...id]].tsx
@@ -66,11 +66,13 @@ function RecordingHead({ metadata }: MetadataProps) {
     description = "";
   }
 
+  const pageTitle = title ? `⏱️ ${title}` : "⏱️";
+
   const image = `${process.env.NEXT_PUBLIC_IMAGE_URL}${metadata.id}.png`;
 
   return (
     <Head>
-      <title>{title}</title>
+      <title>{pageTitle}</title>
       {/* nosemgrep typescript.react.security.audit.react-http-leak.react-http-leak */}
       <meta property="og:title" content={title} />
       <meta property="og:type" content="website" />
--- a/src/ui/components/shared/LaunchBrowserModal.tsx
+++ b/src/ui/components/shared/LaunchBrowserModal.tsx
@@ -20,23 +20,18 @@ function LaunchBrowser({
   }, [path]);
 
   return (
-    <section className="relative m-auto w-full max-w-xl overflow-hidden rounded-lg bg-modalBgcolor text-sm text-bodyColor shadow-lg">
+    <section className="relative m-auto w-full max-w-xl overflow-hidden rounded-lg border border-inputBorder bg-modalBgcolor text-sm text-bodyColor shadow-lg">
       <div className="flex flex-col items-center space-y-9 p-6">
         <div className="place-content-center space-y-3">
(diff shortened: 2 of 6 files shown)


pages/recording/[[...id]].tsx
```diff
@@ -66,11 +66,13 @@ function RecordingHead({ metadata }: MetadataProps) {
     description = "";
   }
 
+  const pageTitle = title ? `⏱️ ${title}` : "⏱️";
```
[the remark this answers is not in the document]
Comment (bvaughn) [reply pages/recording/[[...id]].tsx:69]: Yeah, looks like this branch wasn't created off of `main` (but that other branch). To remove the extra commit(s) @jonbell-lot23, you can rebase. (That being said, it looks like this will merge cleanly so it's probably not a big deal.)

src/ui/components/shared/LaunchBrowserModal.tsx
```diff
@@ -20,23 +20,18 @@ function LaunchBrowser({
...
-            <p>
-              Click <strong>Open Replay</strong> in the dialog shown by your browser
-            </p>
-            {children}
-          </div>
+          <div className="text-lg font-bold">Launching Replay...</div>
+          <div className="space-y-6">{children}</div>
           <div className="w-full" />
-          <div className="flex flex-row text-xs text-gray-500">
+          <div className="flex flex-row text-xs text-gray-400">
```
bvaughn, line 32: gray-400 doesn't really work for light mode

Screen Shot 2023-08-14 at 2 27 53 PM

Screen Shot 2023-08-14 at 2 28 00 PM


I think this is why we should be using CSS vars for stuff like this

jonbell-lot23: Absolutely. 99% of the time I'm using vars and moving things into css modules.

This particular case is an example of light and dark being different by design. Light theme doesn't need (and shouldn't have) a border and dark theme very much needs one. So I could get away with the tailwind variable in this case to get exactly the look I want.

bvaughn: > Light theme doesn't need (and shouldn't have) a border and dark theme very much needs one

I don't really understand. They both look equally low contrast to me without the border :shru

bvaughn: It's fine to land it if you think this looks right

jonbell-lot23: Ah, right. The difference is drop shadows. Ideally borders wouldn't be necessary at all, and in light theme you can use a drop shadow to distinguish between areas. Dark theme can't do that unless you use an electric blue or pink or something. So you end up with borders.

PR #9605, opened 2023-08-14
Remove "Download replay" from NUX flow

jonbell-lot23:
**Background**

Several years ago we built a screen that said "To start making recordings, please download the Replay browser." This was designed for a world where we wanted nearly 100% of new users to download the browser manually.

With testsuites, our product focus has shifted, which has made this screen less relevant. Plenty of invited users go through a flow (sometimes referred to as "New Guest Experience," or NGX) that doesn't require the browser. In fact, the screen is a bit misleading because it implies that the browser is the preferred entry point to recording replays. But for testsuites users, it's not.

This PR removes this screen outright. If someone wants to run the browser, there's an obvious "Launch Replay" button in the top-right corner. But for the vast majority of people, getting dropped into their library directly is the best way to get started. Follow-up: [DES-782](https://linear.app/replay/issue/DES-796/better-library-onboarding)

### DIFF
--- a/pages/recording/[[...id]].tsx
+++ b/pages/recording/[[...id]].tsx
@@ -66,11 +66,13 @@ function RecordingHead({ metadata }: MetadataProps) {
     description = "";
   }
 
+  const pageTitle = title ? `⏱️ ${title}` : "⏱️";
+
   const image = `${process.env.NEXT_PUBLIC_IMAGE_URL}${metadata.id}.png`;
 
   return (
     <Head>
-      <title>{title}</title>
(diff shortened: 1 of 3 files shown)


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-797/welcome-to-replay-screen-appears-even-when-youre-not-new

markerikson (approved):

bvaughn:
Hey @jonbell-lot23, looks like this PR missed the `"download-replay"` enum in `ModalType` – mind posting a follow up to remove that?

PR #9606, opened 2023-08-14
Align grays in testsuites

jonbell-lot23:
Our grays fell out of sync so this is a PR to put things back into tip-top shape.

Old:
image

New:
image


src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.module.css
markerikson, line 1: :hand: Empty file

bvaughn (changes requested):
Personally, I don't like this change. The inputs look **disabled**.

I know it's not up to me but what's motivating this change?

jonbell-lot23:
> Personally, I don't like this change. The inputs look **disabled**.
> 
> I know it's not up to me but what's motivating this change?

The current design is using 4 different shades when 2 are all that's necessary, so I'm tightening them up.

That said, I'd be happy to use a border rather than a background color for the search dialogs. Would that address your concerns?

bvaughn:
> The current design is using 4 different shades when 2 are all that's necessary, so I'm tightening them up.

That seems better to me than a low-contrast input. (I think low-contrast buttons/inputs are often used to indicate _disabled_ state.)

> That said, I'd be happy to use a border rather than a background color for the search fields. Would that address your concerns?

Happy to look?

jonbell-lot23:
Here's a border version:
image

We're probably on the same page, but I wanted to share some insights for later discussions:

**"We use so many colors, what's two more?"**
I think the best way to look at this is per-screen, not across the entire app. This screen uses multiple colors for no reason, and should be standardised. We wouldn't push out a bad bit of code because we use bad code elsewhere, and this is the color version of that.

**It looks disabled**
Agree, that's one of the things about dark theme. My whole life was dark theme when I worked on Windows Phone, and this one of the pitfalls of darktheme. There are two ways to handle it, borders (as I've done above and has precedent) or solid colors (which I started with, and also has precedent).

And of course we can always try for a different shade of gray, but the same overall issue will always exist in dark theme.

bvaughn:
> **It looks disabled**
> Yeah. My whole life was dark theme when I worked on Windows Phone, and this one of the main pitfalls. There are two ways to handle it, borders (as I've done above and has precedent) or solid colors (which I started with, and also has precedent including in our own app: console, network monitor, and library).

This mischaracterizes my concern.

My concern is/was that using a **lighter (lower-contrast) background color** for the inputs makes them look disabled. That's because gray and/or lowered opacity (which comes with decreased contrast) backgrounds are often used for disabled buttons or form controls.

Sitting aside our theme colors (which we could always change) – this input looks disabled to me:
Screen Shot 2023-08-15 at 9 55 50 AM

bvaughn (approved):
The padding adjustments look nice.

I'm not a big fan of the border treatment, but if the choice is between that and the light backgrounds, I'd prefer the borders. :)

jonbell-lot23:
> This mischaracterizes my concern.
> 
> 
> 
> My concern is/was that using a **lighter (lower-contrast) background color** for the inputs makes them look disabled. That's because gray and/or lowered opacity (which comes with decreased contrast) backgrounds are often used for disabled buttons or form controls on the web.

Sorry, I didn't mean to mischaracterise your concerns! But I do think my answer is the same:

* Textfields need to stand out in some way
* You can use a border or a background color
* A very common approach (which is used on iOS and Windows Phone) is to make the text field lighter. The other option is to go darker with textfields. Nothing wrong with darker, but you need to then make sure the rest of the interface is different enough to allow the text field enough contrast. And that means lightening your entire UI. So in my experience they use the "sort of look disabled" approach because it's the least-bad option.


Here's an example in iOS. I agree the fields look disabled, and I agree it's goofy. But in my experience, this is just one of those difficult things about dark theme. 

https://github.com/replayio/devtools/assets/9154902/718db54f-6231-4375-8886-976c2d5e3cd3

bvaughn:
Fair enough!

PR #9613, opened 2023-08-15
[DRAFT] Buggy prototype of browser notifications

jonbell-lot23:
I put together a simple browser notification. This is not production-quality code, but I'm posting it up to get some help from the frontend team.

* When I load a tab in the background, the notification works. But when I hit reload to try it again, it doesn't work. Best I can tell, it's only working once per replay.
* Any advice about how to structure the code is welcome.

Apart from bugs/code, I also have some UX things to consider. Maybe we only show this setting as an opt-in per recording? In my testing, it's pretty annoying having every replay pop up a notification. To experiment and discuss.

### DIFF
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -162,6 +162,47 @@ function _DevTools({
   const [enableLargeText] = useGraphQLUserData("global_enableLargeText");
 
   usePreferredFontSize(enableLargeText);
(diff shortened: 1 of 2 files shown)


bvaughn:
If Replay wants to move forward with investigating browser notifications– (something I'm personally not a fan of)– I think it's the sort of thing that should run through our normal FE process (Linear issue that gets triaged into a cycle.)

PR #9617, opened 2023-08-17
Allow team members to move recordings to other teams

ryanjduffy:
Fixes SCS-1364, SCS-1343, and SCS-1368

* Allows team members to move replays to other teams
* Allows recordings uploaded by a workspace API key to be moved
* Restricts the list of workspaces to match the recording type (test recordings to test workspaces, non-test recordings to non-test workspaces)

### DIFF
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -46,30 +46,6 @@ function DropdownButton({ disabled, children }: { disabled?: boolean; children: 
   );
 }
 
-function filterByWorkspaceType(recording: Recording, workspace: Workspace) {
-  if (workspace.isTest && recording.metadata?.test) {
-    return true;
-  } else if (!workspace.isTest && !recording.metadata?.test) {
-    return true;
-  }
-
-  return false;
-}
-
-function sortWorkspaces(recording: Recording, a: Workspace, b: Workspace) {
-  if (a.id === recording.workspace?.id) {
-    return -1;
-  } else if (a.name && b.name) {
-    return a.name.localeCompare(b.name);
-  } else if (a.name) {
-    return -1;
-  } else if (b.name) {
-    1;
-  }
-
-  return 0;
-}
-
 function useGetPrivacyOptions(
   recording: Recording,
   setExpanded: Dispatch<SetStateAction<boolean>>
@@ -135,8 +111,8 @@ function useGetPrivacyOptions(
       </DropdownItem>,
       <div key="option-team">
         {workspaces
-          .filter(w => filterByWorkspaceType(recording, w))
-          .sort((a, b) => sortWorkspaces(recording, a, b))
+          .filter(w => w.name && !!w.isTest == !!recording.isTest)
+          .sort((a, b) => a.name!.localeCompare(b.name!))
           .map(({ id, name }) => (
             <DropdownItem onClick={() => handleMoveToTeam(id)} key={id}>
               <DropdownItemContent icon="group" selected={!!isPrivate && id === workspaceId}>
--- a/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
+++ b/src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
@@ -46,6 +46,30 @@ function DropdownButton({ disabled, children }: { disabled?: boolean; children: 
   );
 }
 
+function filterByWorkspaceType(recording: Recording, workspace: Workspace) {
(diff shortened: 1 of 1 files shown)


bvaughn:
Hey @ryanjduffy 👋🏼  Mind tagging someone (like me, or whoever's on-call) from FE to review this kind of change too?

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -46,6 +46,30 @@ function DropdownButton({ disabled, children }: { disabled?: boolean; children:
...
 
+function filterByWorkspaceType(recording: Recording, workspace: Workspace) {
+  if (workspace.isTest && recording.metadata?.test) {
+    return true;
+  } else if (!workspace.isTest && !recording.metadata?.test) {
+    return true;
+  }
+
+  return false;
+}
```
bvaughn, line 57:
```suggestion
function filterByWorkspaceType(recording: Recording, workspace: Workspace) {
  return workspace.isTest == recording.metadata?.test;
}
```

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -46,6 +46,30 @@ function DropdownButton({ disabled, children }: { disabled?: boolean; children:
...
+
+function sortWorkspaces(recording: Recording, a: Workspace, b: Workspace) {
+  if (a.id === recording.workspace?.id) {
+    return -1;
```
bvaughn, line 61: This seems odd to me. I think the intention is to always put the workspace a recording is currently assigned to at the top of the list?

I think we should leave the list alpha sorted and allow the browser select control to do its thing. Mixing alpha-sorting like this _feels_ like a weird UX decision.

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
ryanjduffy: Mine is definitely more verbose but I did that intentionally for clarity. The proposed version isn't quite right, I think, because it would be loosely comparing a boolean with an object. (or undefined). My original version was similar to that (`!!workspace.isTest === !!recording.metadata?.test`) but I didn't feel it was as clear.

That said, I also just discovered we have `isTest` and `isInTestWorkspace` on recording so I'll use those instead.

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
ryanjduffy: That's fine. I think the sorting is only a problem for us with many teams anyway.

ryanjduffy:
> Hey @ryanjduffy 👋🏼 Mind tagging someone (like me, or whoever's on-call) from FE to review this kind of change too?

Sure, I can do that.

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
bvaughn: > The proposed version isn't quite right, I think, because it would be loosely comparing a boolean with an object.

The proposed version should be exactly the same as the more verbose version. Both compare truthy value to a falsy value. One is just a lot more verbose. (IMO this doesn't make it "clearer" in my personal opinion but that's subjective.)

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -135,8 +111,8 @@ function useGetPrivacyOptions(
       </DropdownItem>,
       <div key="option-team">
         {workspaces
-          .filter(w => filterByWorkspaceType(recording, w))
-          .sort((a, b) => sortWorkspaces(recording, a, b))
+          .filter(w => w.name && !!w.isTest == !!recording.isTest)
```
bvaughn, line 114: 👍🏼

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
```diff
@@ -135,8 +111,8 @@ function useGetPrivacyOptions(
...
+          .sort((a, b) => a.name!.localeCompare(b.name!))
```
bvaughn, line 115: Can `name` ever be null/undefined? (If so this would throw)

bvaughn (approved):

src/ui/components/shared/SharingModal/PrivacyDropdown.tsx
ryanjduffy: It can't be but I added `w.name` to the filter above to be safe.

PR #9619, opened 2023-08-18
Link to function definition in preview (FE-1825)

Mokshit06:
Adds a link to function location similar to the one in Chrome devtools in the value preview.

Chrome DevTools:
image

Replay DevTools:
Screenshot 2023-08-19 at 12 35 52 AM

This does not modify any logic in `KeyValueRenderer`, just extracts the header of it out to a separate component to be used by `PropertiesRenderer` to render the clickable value.

### DIFF
--- a/packages/replay-next/components/inspector/PropertiesRenderer.tsx
+++ b/packages/replay-next/components/inspector/PropertiesRenderer.tsx
@@ -11,11 +11,17 @@ import { FC, Fragment, ReactNode, Suspense, useContext, useMemo } from "react";
 
 import Expandable from "replay-next/components/Expandable";
 import Loader from "replay-next/components/Loader";
+import { InspectorContext } from "replay-next/src/contexts/InspectorContext";
+import { SourcesContext } from "replay-next/src/contexts/SourcesContext";
 import { objectCache } from "replay-next/src/suspense/ObjectPreviews";
+import { sourcesByIdCache } from "replay-next/src/suspense/SourcesCache";
 import { mergePropertiesAndGetterValues } from "replay-next/src/utils/protocol";
+import { getSourceFileName } from "replay-next/src/utils/source";
+import { getPreferredLocation } from "replay-next/src/utils/sources";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 
 import GetterRenderer from "./GetterRenderer";
+import { KeyValueHeader } from "./KeyValueRenderer";
 import KeyValueRenderer from "./KeyValueRendererWithContextMenu";
 import ValueRenderer from "./ValueRenderer";
 import styles from "./PropertiesRenderer.module.css";
@@ -37,7 +43,9 @@ export default function PropertiesRenderer({
   path?: string;
   pauseId: ProtocolPauseId;
 }) {
+  const { inspectFunctionDefinition } = useContext(InspectorContext);
   const client = useContext(ReplayClientContext);
+  const { preferredGeneratedSourceIds } = useContext(SourcesContext);
 
   // If we have an ObjectPreview already, use it.
   // If we just have an Object, then Suspend while we fetch preview data.
@@ -137,6 +145,41 @@ export default function PropertiesRenderer({
     );
   }
 
+  let renderedFunctionLocation: ReactNode = null;
+  if (preview && object.className == "Function" && preview.functionLocation) {
+    const sourcesById = sourcesByIdCache.read(client);
+    const functionDefinition = preview.functionLocation;
+    const preferredFunctionLocation = getPreferredLocation(
+      sourcesById,
+      preferredGeneratedSourceIds,
+      functionDefinition
+    );
+    const source = sourcesById.get(preferredFunctionLocation.sourceId);
+    const fileName = source ? getSourceFileName(source) : preferredFunctionLocation.sourceId;
+    const fnLocation = `${fileName}:${preferredFunctionLocation.line}`;
+
+    renderedFunctionLocation = (
+      <KeyValueHeader
+        value={
+          <span
+            onClick={() => {
+              if (inspectFunctionDefinition !== null) {
+                inspectFunctionDefinition(functionDefinition);
+              }
+            }}
+            className={styles.SourceLocationLink}
+            data-test-name="ClientValue"
+          >
+            {fnLocation}
+          </span>
+        }
+        nameClass={styles.BucketLabel}
+        layout="vertical"
+        name="[[FunctionLocation]]"
+      />
+    );
+  }
+
   return (
     <>
       <EntriesRenderer containerEntries={containerEntries} pauseId={pauseId} />
@@ -201,6 +244,7 @@ export default function PropertiesRenderer({
         </Suspense>
       )}
 
+      {renderedFunctionLocation}
       {renderedPrototype}
     </>
   );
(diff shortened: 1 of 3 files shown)


bvaughn (commented):
Thanks for tagging me on this! I want to review but, but I'm swamped at the moment. Let me look Monday morning?

packages/replay-next/components/inspector/PropertiesRenderer.tsx
```diff
@@ -137,6 +145,41 @@ export default function PropertiesRenderer({
...
+            {fnLocation}
+          </span>
+        }
+        nameClass={styles.BucketLabel}
+        layout="vertical"
+        name="[[FunctionLocation]]"
+      />
+    );
+  }
+
```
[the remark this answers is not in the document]
bvaughn: What if we did something close to what you're doing, but moving things around a bit?

1. Move the `renderedPrototype` into its own React component e.g. `PrototypeRenderer` (in `replay-next/components/inspector/properties`?) instead of that logic being inline in `PropertiesRenderer`
2. Add a `FunctionRenderer` component in that folder that does the location (and also includes/renders the prototype)

Then we could decide which one to render like so
```tsx
let specialPropertiesRenderer: ReactNode = null;
switch (className) {
  case "Function":
    specialPropertiesRenderer = <FunctionRenderer object={object} />;
    break;
  default:
    specialPropertiesRenderer = <PrototypeRenderer object={object} />;
    break;
}
```

I think we could avoid touching `KeyValueRenderer` at all this way, and keep type-specific logic out of `PropertiesRenderer`

Does that make sense?

bvaughn (commented):
Not strictly necessary but I wonder if we could add a new e2e test for the special function location property? Maybe one that checks that built-in functions _don't_ show this and one that checks that user functions do?

packages/replay-next/components/inspector/PropertiesRenderer.tsx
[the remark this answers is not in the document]
Mokshit06: Yep, working on this

bvaughn:
(FYI I'll wait on an explicit ping on this PR before taking another look. No hurry of course.)

Mokshit06:
@bvaughn I have made the changes you suggested :)

bvaughn:
I see. I didn't expect to still see changes in `KeyValueRenderer` so I assumed it was still a work in progress.

packages/replay-next/components/inspector/values/shared.module.css
```diff
@@ -46,7 +46,7 @@
   color: var(--value-type-undefined);
 }
 
-.IconButton {
+.Iconˀtton {
```
bvaughn, line 49: This is a syntax error.

packages/replay-next/components/inspector/values/PrototypeRenderer.tsx
```diff
@@ -0,0 +1,40 @@
...
+import PropertiesRenderer, { addPathSegment } from "../PropertiesRenderer";
+import styles from "./shared.module.css";
+
+type PropertiesRendererProps = {
+  path: string | undefined;
+  object: ProtocolObject;
+  pauseId: ProtocolPauseId;
+};
+
+export default function PrototypeRenderer({ object, pauseId, path }: PropertiesRendererProps) {
```
bvaughn, line 17: Can we move this into a different folder? It's not a "value" renderer like the other things in here. It's rendering prototype properties.

packages/replay-next/components/inspector/values/FunctionRenderer.tsx
```diff
@@ -73,3 +79,50 @@ export function functionProtocolObjectToString(protocolObject: ProtocolObject) {
 
   return `${functionName}(${functionParameterNames.join(", ")}) {}`;
 }
+
+type FunctionLocationRendererProps = {
+  object: ProtocolObject;
+};
+
+export function FunctionLocationRenderer({ object }: FunctionLocationRendererProps) {
```
bvaughn, line 87: I think this component should be in its own file (in `replay-next/components/inspector/properties`)

packages/replay-next/components/inspector/values/shared.module.css
Mokshit06: The changes to `KeyValueRenderer` are just to extract the Header part of it outside the component itself, so that it can also be used with the `FunctionRenderer`

Mokshit06: Ah, will fix this syntax error

bvaughn: > The changes to KeyValueRenderer are just to extract the Header part of it outside the component itself, so that it can also be used with the FunctionRenderer

I see.

I think the thing that's throwing me about that component is the name.

It isn't really a "key value **header**" – it's not shown "above" the key value component (like a header). It's shown in between an expanded value and its properties list. It seems more like on of the properties?

bvaughn: Really it's just another key-value pair, where the value is a link.

I don't think we really need to share that code.

Mokshit06:
I agree with the changes and have made them. Please have another look

bvaughn (commented):
Overall this looks great. Please revert the change to `packages/replay-next/components/inspector/values/shared.module.css1 before merging though.

packages/replay-next/components/inspector/properties/FunctionLocationRenderer.tsx
```diff
@@ -0,0 +1,61 @@
...
+import styles from "./shared.module.css";
+
+type FunctionLocationRendererProps = {
+  object: ProtocolObject;
+};
+
+export default function FunctionLocationRenderer({ object }: FunctionLocationRendererProps) {
+  const client = useContext(ReplayClientContext);
+  const { inspectFunctionDefinition } = useContext(InspectorContext);
+  const { preferredGeneratedSourceIds } = useContext(SourcesContext) ?? {};
```
bvaughn, line 21: This change looks unexpected. Do we know which instances `SourcesContext` is null?

The `SourcesContextAdapter` (which renders the `SourcesContext` provider) is wrapped around the root of the `DevTools` component's children, right beneath the `SessionContext`.

I'd like to figure this out before we mrege.

bvaughn:
The change seems promising but many tests are still failing 😞 

Looking at one randomly, I'm confused. It looks like the test is hanging because the "wait until indexed" helper is not completing. That helper is driven by the `data-test-progress` attribute, declared here in the source:
https://github.com/replayio/devtools/blob/be8ee99a613f1eaed755ec637890e94d7974b481/src/ui/components/Timeline/Capsule.tsx#L36-L42

But looking at the DOM on one of your tests, that attribute is missing (as is the `data-test-id` attribute):
Screen Shot 2023-08-31 at 4 27 17 PM

Mokshit06:
Ah my bad, the branch wasn't up to date with main, and the changes to `Capsule.tsx` had been made recently

bvaughn:
I thought about that case, but then I thought...if the branch wasn't updated, the e2e tests wouldn't yet be blocking on that helper (which was added at the same time) but 🤷🏼  Let's see! 😆

Mokshit06:
Looks like we're down to only 2 failing tests after rebase, but for some reason I can't open either of them as they're private. But from the reason of failure I also don't think that they are failing because of changes related to this PR

bvaughn:
>  but for some reason I can't open either of them as they're private. 

Ah, I think you're not on the Workspace list. I'll DM you links to join those teams.

> But from the reason of failure I also don't think that they are failing because of changes related to this PR

Agreed they are likely unrelated to this PR.

bvaughn (approved):
This looks good to me 👍🏼  Thanks for sticking with this through all of the rounds of back and forth.

PR #9624, opened 2023-08-21
Improve textfield component to work better with themes

jonbell-lot23:
Addresses [DES-857](https://linear.app/replay/issue/DES-857/fix-new-org-dialog-for-dark-mode)

Old:
image

New:
image

--

We force our onboarding into dark theme. But this means you can follow these steps:

1. Go through onboarding
2. Get to our app
3. Change to light theme
4. Go back to an onboarding page (such as the new team page shown here)
5. Result: a light theme textfield in our dark theme onboarding screens

Fixes/changes

1. Passed a "theme" prop so I can force dark or light theme
2. If nothing is passed, it defaults to the system theme, as expected
3. Removed almost all tailwind and used css modules instead

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -705,7 +705,7 @@
   --theme-sidebar-background: var(--theme-base-100);
   --theme-text-field-bgcolor-hover: var(--theme-base-95);
   --theme-text-field-bgcolor: #f6f6f6;
-  --theme-text-field-color: var(--theme-body-color);
+  --theme-text-field-color: var(--body-color);
   --theme-text-field-placeholder-color: #a9a9b1;
   --team-row-active: #3f434a;
   --tooltip-bgcolor: #38383d;
--- a/src/ui/components/shared/Forms/TextInput.module.css
+++ b/src/ui/components/shared/Forms/TextInput.module.css
@@ -29,7 +29,7 @@
 .dark .textInput {
   background-color: #3e434b; /* Dark: --theme-text-field-bgcolor */
   color: #fff; /* --theme-text-field-color: */
-  border-color: var(--theme-base-85); /* --input-border */
+  border-color: #3e434b; /* Dark: --input-border */
 }
 
 .dark .textInput:hover {
@@ -40,7 +40,7 @@
 
 .light .textInput {
   background-color: #f6f6f6; /* --theme-text-field-bgcolor */
-  color: #fff; /* --theme-text-field-color: */
+  color: #38383d; /* Light: var(--body-color); */
   border-color: var(--cool-gray-200); /* --input-border */
 }
 
--- a/src/ui/components/shared/Forms/TextInput.module.css
+++ b/src/ui/components/shared/Forms/TextInput.module.css
@@ -18,6 +18,9 @@
 
 .default .textInput {
   background-color: var(--theme-text-field-bgcolor);
+  color: var(--theme-text-field-color);
+  border-color: var(--input-border);
+  caret-color: var(--theme-text-field-color);
 }
 
 .default .textInput:hover {
@@ -30,6 +33,7 @@
   background-color: #3e434b; /* Dark: --theme-text-field-bgcolor */
   color: #fff; /* --theme-text-field-color: */
   border-color: #3e434b; /* Dark: --input-border */
+  caret-color: #fff;
 }
 
 .dark .textInput:hover {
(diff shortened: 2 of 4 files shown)


packages/replay-next/pages/variables.css
jonbell-lot23, line 1: Came across this old bug and put it into this PR

src/ui/components/shared/Forms/TextInput.tsx
```diff
@@ -1,27 +1,28 @@
...
     center?: boolean;
+    theme?: "light" | "dark";
     className?: string;
   }
 >(function TextInput(props, ref) {
-  const { textSize, center } = props;
+  const { textSize, center, theme, className, ...otherProps } = props;
+
+  const themeClass = theme || "default";
+  const textSizeClass = textSize ? `text-${textSize}` : "text-sm"; // Tailwind class for textSize
```
ryanjduffy, line 18: `text-${textSize}` may break with tailwind pruning. If we use tailwind, we need to explicitly include all the tailwind strings in the source to avoid unexpected pruning.

src/ui/components/shared/Forms/TextInput.module.css
```diff
@@ -0,0 +1,49 @@
...
+
+.default .textInput:hover {
+  background-color: var(--theme-text-field-bgcolor-hover);
+}
+
+/* Dark theme */
+
+.dark .textInput {
+  background-color: #3e434b; /* Dark: --theme-text-field-bgcolor */
+  color: #fff; /* --theme-text-field-color: */
```
ryanjduffy, line 31: Let's add `caret-color` here too so we can see it in dark mode

src/ui/components/shared/Forms/TextInput.module.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.default .textInput:hover {
+  background-color: var(--theme-text-field-bgcolor-hover);
+}
+
+/* Dark theme */
+
+.dark .textInput {
+  background-color: #3e434b; /* Dark: --theme-text-field-bgcolor */
```
ryanjduffy, line 30: Is it customary to use the explicit color values instead of css vars here?

ryanjduffy (changes requested):

src/ui/components/shared/Forms/TextInput.module.css
jonbell-lot23: Yeah, unfortunately. If we use a css variable, it will register the current theme. Meaning you'll hit this page in light theme, but the page is fixed to dark theme, so you'll see a light theme on a dark background.

I looked for a way to do something like (var--:dark--theme-text-field-bgcolor) but that's not possible. So we've gotta hard-code it.

src/ui/components/shared/TeamOnboarding.tsx
```diff
@@ -121,6 +121,7 @@ function TeamNamePage({
           onChange={onChange}
           textSize="2xl"
           center={true}
+          theme="dark" // we force onboarding into dark theme
```
bvaughn, line 124: Seems like you had to hard-code a bunch of colors to dark theme in the CSS, so what affect does this actually have?

jonbell-lot23: When you don't pass a theme prop, it defaults to the same css variables we've always use, which change based on dark theme or light theme.

The tricky bit was that we need an ability to force dark theme sometimes, or else the css variables will set light theme and do the wrong thing.

So the hard-coding is the way you get around a css variable making the "light theme text field on a dark theme screen" error seen in the screenshot.

src/ui/components/shared/Forms/TextInput.module.css
```diff
@@ -0,0 +1,54 @@
...
```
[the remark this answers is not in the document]
bvaughn: Sorry. I missed the `.dark` and `.light` prefix classes here.

bvaughn (approved):

src/ui/components/shared/Forms/TextInput.tsx
```diff
@@ -1,27 +1,45 @@
...
     textSize?: "base" | "lg" | "2xl";
     center?: boolean;
+    theme?: "light" | "dark";
     className?: string;
   }
 >(function TextInput(props, ref) {
-  const { textSize, center } = props;
+  const { textSize, center, theme, className, ...otherProps } = props;
+
+  const themeClass = theme || "default";
```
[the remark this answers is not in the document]
bvaughn: Can you make sure this works right with the system theme before landing?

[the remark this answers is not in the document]
jonbell-lot23: Oops I almost missed this comment. Yup, will address.

ryanjduffy (approved):

PR #9626, opened 2023-08-22
Use the displayed focus window only in the timeline and the active one everywhere else

hbenl:
- add the active focus window (the one set in the backend with `Session.requestFocusWindow`) to redux and FocusContext
- rename the displayed focus window in redux and FocusContext
- use the active focus window in most places
- remove some unused code (including the `FocusWindow` type because it is identical to `TimeStampedPointRange`)


packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -179,14 +186,22 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
...
       isTransitionPending,
       rangeForDisplay: range,
       range: deferredRange,
       update: updateFocusRange,
       updateForTimelineImprecise,
     }),
-    [deferredRange, isTransitionPending, range, updateFocusRange, updateForTimelineImprecise]
+    [
+      activeRange,
+      deferredRange,
```
bvaughn, line 199: I don't think having "active range", "deferred range", and "range" is good. I think particularly, naming something "deferred range" internally but exposing it as "range" is...confusing. Reading this the first time through– even being familiar with the constraints, I was a bit confused.

I know this is kind of subjective but I think we should try to pick clearer names. I propose:
* rangeForDisplay: Only used by the Timeline UI for display purposes
* rangeForSuspense: Used for anything that might _suspend_ based on the value
* range: Used for things that _don't_ need to suspend

bvaughn (commented):
First round of feedback:

> add the active focus window (the one set in the backend with Session.requestFocusWindow) to redux and FocusContext

I wonder if we can get away from mirroring state in Redux that's already in a context? At best it probably leads to us doing more state updates than necessary, and at worst can lead to syncing issues like the search flickering bug.

hbenl:
Superseded by #9650

PR #9629, opened 2023-08-23
Rewrite "Recorded Protocol Viewer" to use runEvaluation for perf

markerikson:
This PR:

- Makes a few small tweaks
  - Removed a console log from an E2E test helper
  - Updated `hitPointsCache` to stop stripping `frame` info out of hit points
  - Updated `ReplayClient.runEvaluation` options to match what the protocol currently allows
- Rewrites the "Recorded Protocol Viewer" panel to use `runEvaluation` and an interval cache for better perf

The "Recorded Protocol Viewer" panel was _really_ slow, largely because it was trying to make dozens or hundreds of calls to the `logPointAnalysisCache` to extract the arguments to our protocol `onRequest/onResponse/onRequestError` callbacks.  That's the use case `runEvaluation` is meant to solve.  Also, it was rerunning all the requests every time you changed the focus range.

I've rewritten it to use an interval cache, and the body now uses `runEvaluation` to fetch the data.

This did get tricky, because `runEvaluation` still only lets you return 1 depth of object previews, and the response objects can be arbitrarily large.  So, I've resorted to the same trick Logan and I used in the routines: `JSON.stringify()` the actual value in the evaluation, split it into 9999-character chunks to work around our Chromium 10K-character limit, stick all those chunks into one array as the final eval result, then reconstruct the large string and run it through `JSON.parse()` on the other end (whew!).

This actually works!  It can still take a few seconds for it to process several hundred protocol requests, but the data looks good in the UI, and it should now only fetch a given range's protocol requests once.

### DIFF
--- a/packages/replay-next/src/suspense/HitPointsCache.ts
+++ b/packages/replay-next/src/suspense/HitPointsCache.ts
@@ -66,7 +66,10 @@ export const hitPointsCache = createFocusIntervalCacheForExecutionPoints<
         { kind: "locations", locations },
         { begin, end }
       );
-      hitPoints = pointDescriptions.map(({ point, time }) => ({ point, time }));
+      // Note that PointDescriptions include the current frame.
+      // Type-wise we're going to return `TimeStampedPoint`s,
+      // but the extra frame data is there in this case.
+      hitPoints = pointDescriptions;
     }
 
     return hitPoints;
--- a/packages/replay-next/src/suspense/HitPointsCache.ts
+++ b/packages/replay-next/src/suspense/HitPointsCache.ts
@@ -1,4 +1,4 @@
-import { Location, PointRange, TimeStampedPoint } from "@replayio/protocol";
+import { Location, PointDescription, PointRange, TimeStampedPoint } from "@replayio/protocol";
 import { createCache } from "suspense";
 
 import { breakpointPositionsCache } from "replay-next/src/suspense/BreakpointPositionsCache";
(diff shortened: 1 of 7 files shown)


packages/replay-next/src/suspense/HitPointsCache.ts
```diff
@@ -66,7 +66,10 @@ export const hitPointsCache = createFocusIntervalCacheForExecutionPoints<
         { kind: "locations", locations },
         { begin, end }
       );
-      hitPoints = pointDescriptions.map(({ point, time }) => ({ point, time }));
+      // Note that PointDescriptions include the current frame.
+      // Type-wise we're going to return `TimeStampedPoint`s,
+      // but the extra frame data is there in this case.
+      hitPoints = pointDescriptions;
```
hbenl, line 72: This should update the frame data like [here](https://github.com/replayio/devtools/blob/main/packages/replay-next/src/suspense/PauseCache.ts#L105-L112), otherwise we will get bugs due to corresponding sources.
When the frame data is updated, it's also safe to make this "official" by changing the return type from `TimeStampedPoint` to `PointDescription`.

markerikson: Yeah, I can definitely update the frame data, although the issue is that the other half of this cache _just_ gets `TimeStampedPoints`, I think?

markerikson: Ah, looks like those _are_ `PointDescription`s too.

PR #9632, opened 2023-08-24
Improve RDT component inspection and avoid timeouts

markerikson:
- Updated the RDT panel to inject the backend eval earlier (as soon as it pauses, rather than waiting for the first request attempt)
- Added check to avoid passing on messages to the RDT UI if the pause has changed since the request started
- Patched RDT `frontend.js` to increase timeout from 5s -> 15s

This doesn't completely eliminate the chance of an inspection timeout, but given that it looked like inspection results _usually_ come back within a few seconds (and in some cases, less than 500ms after the UI had timed out), I think this should significantly reduce the chances of us hitting a timeout.

(the RDT `.patch` file is ginormous because I hand-edited the sourcemap to match, for completeness, and that's all one giant line.  all I actually changed was `const TIMEOUT = 5000` => `15000`.)

### DIFF
--- a/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
@@ -90,6 +90,7 @@ class ReplayWall implements Wall {
 
   setPauseId(pauseId: string) {
     this.pauseId = pauseId;
+    this.inspectedElements.clear();
   }
 
   // called by the frontend to register a listener for receiving backend messages
@@ -237,7 +238,7 @@ class ReplayWall implements Wall {
     }
   }
 
-  private async ensureReactDevtoolsBackendLoaded() {
+  public async ensureReactDevtoolsBackendLoaded() {
     if (this.recordingTarget === null) {
       this.recordingTarget = await recordingTargetCache.readAsync(this.replayClient);
     }
@@ -250,6 +251,7 @@ class ReplayWall implements Wall {
 
   // send a request to the backend in the recording and the reply to the frontend
   private async sendRequest(event: string, payload: any) {
+    const originalPauseId = this.pauseId;
     const response = await evaluate({
       replayClient: this.replayClient,
       text: ` window.__RECORD_REPLAY_REACT_DEVTOOLS_SEND_MESSAGE__("${event}", ${JSON.stringify(
@@ -261,7 +263,7 @@ class ReplayWall implements Wall {
       assert(this.pauseId, "Must have a pause ID to handle a response!");
       const result: any = await getJSON(this.replayClient, this.pauseId, response.returned);
 
-      if (result) {
+      if (result && this.pauseId === originalPauseId) {
         this._listener?.({ event: result.event, payload: result.data });
       }
       return result;
@@ -629,6 +631,13 @@ export function ReactDevtoolsPanel() {
       }
     };
   }, [wall, store]);
+
+  useEffect(() => {
+    if (wall && currentPoint !== null) {
+      console.log("Starting RDT injection from effect");
+      wall.ensureReactDevtoolsBackendLoaded();
+    }
+  }, [currentPoint, wall]);
 
   if (currentPoint === null) {
     return null;
--- a/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
@@ -72,7 +72,6 @@ class ReplayWall implements Wall {
   private _listener?: (msg: any) => void;
   private inspectedElements = new Set();
   private highlightedElementId?: number;
-  private recordingTarget: RecordingTarget | null = null;
   store?: StoreWithInternals;
   pauseId?: string;
 
@@ -239,11 +238,9 @@ class ReplayWall implements Wall {
   }
 
   public async ensureReactDevtoolsBackendLoaded() {
-    if (this.recordingTarget === null) {
-      this.recordingTarget = await recordingTargetCache.readAsync(this.replayClient);
-    }
+    const recordingTarget = await recordingTargetCache.readAsync(this.replayClient);
 
-    if (this.recordingTarget === "chromium") {
+    if (recordingTarget === "chromium") {
       const pauseId = await ThreadFront.getCurrentPauseId(this.replayClient);
       await injectReactDevtoolsBackend(this.replayClient, pauseId);
     }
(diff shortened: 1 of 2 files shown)


hbenl (approved):

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -261,7 +263,7 @@ class ReplayWall implements Wall {
       assert(this.pauseId, "Must have a pause ID to handle a response!");
       const result: any = await getJSON(this.replayClient, this.pauseId, response.returned);
 
-      if (result) {
+      if (result && this.pauseId === originalPauseId) {
```
bvaughn, line 266: This (race condition type thing) looks like a good thing for us to add test coverage for, though I'm not sure how easy it would be to do.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -630,6 +632,13 @@ export function ReactDevtoolsPanel() {
     };
   }, [wall, store]);
 
+  useEffect(() => {
+    if (wall && currentPoint !== null) {
+      console.log("Starting RDT injection from effect");
```
bvaughn, line 637:
```suggestion
```

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -630,6 +632,13 @@ export function ReactDevtoolsPanel() {
...
+      wall.ensureReactDevtoolsBackendLoaded();
```
bvaughn, line 638: This looks a little racy, because of the implementation of `ensureReactDevtoolsBackendLoaded`. I'll comment more there– but this effect will run multiple times so I think we'll need to make `ensureReactDevtoolsBackendLoaded` more robust to being called multiple times.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -237,7 +238,7 @@ class ReplayWall implements Wall {
     }
   }
 
-  private async ensureReactDevtoolsBackendLoaded() {
+  public async ensureReactDevtoolsBackendLoaded() {
     if (this.recordingTarget === null) {
```
bvaughn, line 242: This function will be called multiple times, and it _looks_ unsafe. (I think it's actually safe because our Suspense caches should prevent meaningful code from being double executed.) Still, I wonder if we should not _rely_ on that and add a better check here to only call it once.

For example, the way we're checking if `this.recordingTarget === null` doesn't prevent that code from executing multiple times (since we're awaiting an async value) and we don't even check before calling `injectReactDevtoolsBackend` again.

I know that neither of these should have any practical impact _now_ but it seems like the sort of thing that _could_ change over time.

bvaughn (commented):
How do you generate these patch files? (Do we have instructions for regenerating them written down somewhere?)

jasonLaster:
This is probably still an improvement, but when I test this on metabase I still see timeouts

https://devtools-git-bugfix-fe-1839-rdt-component-timeouts-recordreplay.vercel.app/recording/e2etestscenariosonboardingsearchrecently-viewedcyspecjs--c27d1526-9d90-4ea3-86bc-ac271d9c8dd7?point=224242320811025954533989985053352943&time=105505.70647931304&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjEzNzI3MTM0ODM0MzkzNTUzNjUxNDM0NDU2MjY1NTc1NjI4OCIsInRpbWUiOjY2ODI3fSwiZW5kIjp7InBvaW50IjoiMjczNTY5MTQxMDExNzA0MDcyNDU4OTIxMDM4Mzg0NzI2MDE2IiwidGltZSI6MTI2Nzg1fX0%253D#

markerikson:
@bvaughn : standard `yarn patch` command, then hand-edit the files that got copied to the temp folder, follow instructions to have Yarn generate the patch file.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
markerikson: Not sure what you mean by "unsafe" in this case.  The worst that can happen here is that two calls to `ensureReactDevToolsBackendLoaded()` would run, the first kicks off the recording target request and waits, and the second one _tries_ to kick off the recording target request but it's already in flight.

markerikson:
@jasonLaster got an actual replay of the timeouts happening? Just tried on that link and component inspection is working okay for me.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
markerikson: Really it's just saving us an additional `await` on later calls.

markerikson:
@jasonLaster clicking around that Metabase recording some more, I saw some evals take 10+ seconds to complete.  That was _just_ fast enough that the RDT UI didn't time out, but if it had, that's a backend problem and there's not much we can do about it (other than cranking up the timeout higher or something).

markerikson:
Worse, some of the evals in that recording are taking 28s (!!!!!!) to complete.

The UI can't solve that.

bvaughn:
I'm a little frustrated that this got merged before I stamped it when I was actively reviewing it and we were discussing concerns I raised.

markerikson:
Sorry - Holger had stamped it earlier, and I thought I'd addressed your concerns.

bvaughn:
I'd asked for us to change something and you asked "why" and then merged before I could respond.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
bvaughn: By "unsafe" I meant that this method calls a function that looks like it could be (and maybe in the future, actually could be) side effectful– `injectReactDevtoolsBackend`. 

In general, I think it's good future-proofing pattern for "init"/"inject" type functions to be defensive against being called more than once.

markerikson: That was actually the point of using the eval cache - to ensure that it only actually runs once per pause.

(I could in theory do a sync check inside that to ask the cache if it has a cache entry for this pause at all, but I don't see how that would be much of an improvement over just waiting for the cache promise to resolve anyway.)

bvaughn: Yeah, that's fair. I think the thing that concerns me is when I see code in one place relying on a particular implementation detail of code in another place that isn't really explicit in the name/api of that other place. But maybe I'm just over-worrying about this.

I agree with you that it shouldn't cause any problems in its current form. 👍🏼  Fair enough.

PR #9636, opened 2023-08-24
Fix text input sizing in settings modal

ryanjduffy:
## Issue

Several text boxes in the settings modal were short instead of taking the whole width. This was a regression from #9624 which added an extra DOM layer that broke some of the existing layout

## Fix

Remove the layer and fix up the associated CSS

Was:
image

Now:
image

### DIFF
--- a/src/ui/components/shared/Forms/TextInput.module.css
+++ b/src/ui/components/shared/Forms/TextInput.module.css
@@ -16,39 +16,39 @@
   text-align: center;
 }
 
-.system .textInput {
+.system.textInput {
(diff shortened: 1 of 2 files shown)


ryanjduffy:
@jonbell-lot23 - I removed the wrapping `<div>` (which I don't _think_ was adding any value) and updated the CSS structure to match it. Let me know if I missed some rationale for it.

bvaughn (approved):

jonbell-lot23 (approved):
Looks good to me!

PR #9637, opened 2023-08-25
Improve passport copy

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-861/passport-improvements after a chat between Jon and Jason

New header
New headings
Slightly different groupings

image

As a follow-up, we're going to add a "jump to network request" bit later. Jumping is a key feature/term for us. We have a whole section for it in the docs! -> https://docs.replay.io/reference-guide/debugging/jumping

### DIFF
--- a/src/ui/components/Passport/Passport.tsx
+++ b/src/ui/components/Passport/Passport.tsx
@@ -124,6 +124,17 @@ const Passport = (props: PropsFromRedux) => {
         "https://replayio.notion.site/Adding-print-statements-18a7d1f85b434706b46af0a8d5d298fe?pvs=4",
       blurb: "When looking at a source, click the plus sign.",
     },
+  ];
+
+  const jumpItems = [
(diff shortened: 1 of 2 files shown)


src/ui/components/Passport/Passport.tsx
```diff
@@ -231,20 +232,20 @@ const Passport = (props: PropsFromRedux) => {
...
-      title: "SWISS ARMY KNIFE",
-      items: swissArmyItems,
+      title: "INSPECTOR",
+      items: inspectionItems,
     },
     {
-      title: "MULTIPLAYER",
-      items: multiplayerItems,
+      title: "ADVENTURER",
+      items: swissArmyItems,
```
jonbell-lot23, line 248: I am using the legacy name here for now because I'm pretty sure we're going to end up changing the user-facing copy 3 more times.

PR #9641, opened 2023-08-25
Add e2e tests for Network panel

bvaughn:
Adds e2e tests for–
- [x] filtering by category and text
- [x] inspecting request/response details
- [x] seeking to a request and syncing with the current time
- [x] navigating selected requests via up/down arrow keys

### DIFF
--- a/packages/e2e-tests/helpers/commands.ts
+++ b/packages/e2e-tests/helpers/commands.ts
@@ -1,7 +1,8 @@
-import { Page } from "@playwright/test";
+import { Page, expect } from "@playwright/test";
 import chalk from "chalk";
 
-import { debugPrint, getCommandKey } from "./utils";
+import { waitForSourceContentsToFinishStreaming } from "./source-panel";
+import { debugPrint, getCommandKey, waitFor } from "./utils";
 
 export async function quickOpen(page: Page, url: string): Promise<void> {
   await debugPrint(page, "Opening quick-open dialog", "quickOpen");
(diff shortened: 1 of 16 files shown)


packages/e2e-tests/helpers/utils.ts
```diff
@@ -84,6 +84,25 @@ export async function mapLocators<T>(
   );
 }
 
+export async function waitForRecordingToFinishIndexing(page: Page): Promise<void> {
```
bvaughn, line 87: Using this new helper method made my tests a lot more stable when the backend was being slow today. Maybe we should consider doing this in more (or every?) test.

bvaughn:
These tests all seem to pass for me locally _if_  the recording finishes indexes in time (which it does about 2/3s of the time?)

So maybe we need to bump up the timeout.

bvaughn:
:( The new Network tests are failing because of timeouts streaming source contents.

bvaughn:
I don't know _why_ Network tests are still seeming sketchy on CI. They pass 100% of the time for me now when running locally, even in headless mode.

bvaughn:
I added some comments to [8c401211-1be5-4538-9de0-2f4abb3c9b06](https://app.replay.io/recording/network-03-should-sync-and-display-the-current-time-in-relation-to-the-network-requests--8c401211-1be5-4538-9de0-2f4abb3c9b06). Looks like it's another mouse-over with meta key issue. Seems to reliably work on MacOS but I guess not on Linux or whatever our CI box is running. Super frustrating.

bvaughn:
Network tests all pass now (that I swapped order on the meta key and mouse over helpers)

Other tests are flaky as hell lately but I think that's related to the cloning recording change. Hopefully adding the "wait until indexed" helper to all tests will help with that. I'll do that next.

PR #9642, opened 2023-08-25
Fix source viewer copy/paste format

bvaughn:
Seems like the extra wrapper element is unavoidable because of the parent using `display: flex`

I tried using `display: block` on the parent instead, but for some reason that resulted in no line breaks being copied at all, even when selecting multiple lines.

### DIFF
--- a/packages/replay-next/components/sources/SourceList.module.css
+++ b/packages/replay-next/components/sources/SourceList.module.css
@@ -1,3 +1,12 @@
 .List {
   overscroll-behavior: none;
 }
+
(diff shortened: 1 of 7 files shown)


hbenl (approved):

bvaughn:
Some subtle impact of the change in this branch causes one of our Delta context menu tests to fail because it tries to right click on a preview popup
Screen Shot 2023-08-28 at 10 36 24 AM

Probably just a minor cursor issue. I'll look into it before merging this.

PR #9643, opened 2023-08-27
Address margins for source viewer search

jonbell-lot23:
**Description**

I'm improving the balance for our search box. It was originally [filed](https://linear.app/replay/issue/FE-1828/find-bar-should-not-have-a-border-radius) with a suggestion to remove the border radius. But that would cause its own composition issues. Better is to give the component its own space, with consistent margins.

Old on left, new on right:
image

Zoomed out view:
image

rem vs px note: we frequently default to .25rem increments, which is great. But there are cases like these where we need more precise than that approach offers.

### DIFF
--- a/packages/replay-next/components/sources/SourceSearch.module.css
+++ b/packages/replay-next/components/sources/SourceSearch.module.css
@@ -1,14 +1,14 @@
 .Outer {
-  width: 100%;
   padding: 0.25rem 0;
   background-color: var(--color-contrast);
+  margin: 1px 5px 4px 5px;
 }
 
 .Container,
 .ContainerFocused {
   width: 100%;
   height: 1.75rem;
-  border: 2px solid var(--background-color-contrast-2);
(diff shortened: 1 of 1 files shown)


bvaughn (changes requested):
Regression on dark mode
Screen Shot 2023-08-28 at 9 54 31 AM

jonbell-lot23:
> This change causes a regression (background color) on dark mode:
> 
> Screen Shot 2023-08-28 at 9 54 54 AM
> 
> 
> 
> 

Can you provide more details? I didn't touch the background color in this diff.

bvaughn:
> I didn't touch the background color in this diff.

I know. Adding margin causes things to break, as shown in the screenshot above.

jonbell-lot23:
Moving to new branch: https://github.com/replayio/devtools/pull/9649

It contains the background color adjustment Brian mentioned earlier.

PR #9648, opened 2023-08-29
Add Tour to Testsuites

jonbell-lot23:
**Background**

This adds our Tour flow to the Testsuites experience, but with different copy to handle the differences with the standard Tour.

**How testsuites will look:**
image

**How standard replays will look:**
image

And as comparison, here's how standard replays looked before these visual adjustments. The most notable change is how I'm making a more clear yellow callout rather than using a single pixel border that looks like a button:
image

I am tracking the text differences in [this Notion document](https://www.notion.so/replayio/Testsuites-Tour-Copy-9aeb9b921b49415b81a739bd0e0f0829?pvs=4). I'm doing Figma work [here](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?type=design&node-id=18482-63785&mode=design).

**Status**

~~I have a few more edgecases I need to track down. First, I don't think we should default to Larry for testsuites users, so I'll need to special-case that. Second, I need to make sure our other paths aren't adversely affected by this change. Stay tuned for that.~~ Ok, I've done my due diligence and this is ready for review.

### DIFF
--- a/src/ui/components/Tour/Tour.module.css
+++ b/src/ui/components/Tour/Tour.module.css
@@ -21,16 +21,24 @@
 }
 
 .stepCallout {
-  border: 1px solid #fff;
-  background-color: rgba(255, 255, 255, 0.05);
-  padding: 0.5em 0.5em;
-  border-radius: 4px;
+  display: flex;
+  align-items: top;
+  color: #ff0;
+}
+
+.stepCallout .Icon {
+  width: 16px;
+  color: #ff0;
+}
+
+.stepCalloutText {
+  padding-top: 1px;
 }
 
 .h1 {
   color: #fff;
   margin-bottom: 20px;
-  font-size: 22px;
+  font-size: 18px;
   font-weight: 500;
   border-bottom: 1px solid rgba(255, 255, 255, 0.3);
   padding-bottom: 1rem;
@@ -40,7 +48,7 @@
 
 .intro {
   color: #fff;
-  font-size: 18px;
+  font-size: 14px;
   font-weight: 300;
   margin-bottom: 1rem;
   height: 100%;
--- a/src/ui/components/Tour/Tour.module.css
+++ b/src/ui/components/Tour/Tour.module.css
@@ -15,7 +15,7 @@
 }
 
 .TourBox {
-  color: #fff;
+  color: var(--tour-body-color);
   padding: 0.5rem;
   height: 100vh;
 }
@@ -23,12 +23,12 @@
 .stepCallout {
   display: flex;
   align-items: top;
-  color: #ff0;
+  color: var(--tour-callout-body-color);
 }
 
 .stepCallout .Icon {
   width: 16px;
-  color: #ff0;
+  color: var(--tour-callout-body-color);
 }
 
 .stepCalloutText {
@@ -36,7 +36,7 @@
 }
 
 .h1 {
-  color: #fff;
+  color: var(--tour-body-color);
   margin-bottom: 20px;
   font-size: 18px;
   font-weight: 500;
@@ -47,7 +47,7 @@
 }
 
 .intro {
-  color: #fff;
+  color: var(--tour-body-color);
   font-size: 14px;
   font-weight: 300;
   margin-bottom: 1rem;
--- a/src/ui/components/Tour/Tour.tsx
+++ b/src/ui/components/Tour/Tour.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect } from "react";
 
+import Icon from "replay-next/components/Icon";
 import { userData } from "shared/user-data/GraphQL/UserData";
 import { setSelectedPrimaryPanel } from "ui/actions/layout";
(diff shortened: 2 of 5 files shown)


src/ui/components/Tour/Tour.module.css
```diff
@@ -21,16 +21,24 @@
...
-  padding: 0.5em 0.5em;
-  border-radius: 4px;
+  display: flex;
+  align-items: top;
+  color: #ff0;
+}
+
+.stepCallout .Icon {
+  width: 16px;
+  color: #ff0;
```
bvaughn, line 31: ⚠️ Hard-coded color

src/ui/components/Tour/Tour.tsx
```diff
@@ -82,17 +99,39 @@ const Tour: React.FC = () => {
...
-      <p>In DevTools, look underneath the video to find the Replay console.</p>
-      <p>
-        <div className={styles.stepCallout}>
-          Hover in the console and click a button to time travel!
-        </div>
-      </p>
-    </div>
-  );
+  const timeTravel =
+    typeOfReplay === "events" ? (
```
bvaughn, line 103:
```js
const typeOfReplay = recording && isTestSuiteReplay(recording) ? "cypress" : "events";
```

nit: This is an odd enum value; "events" is not a type of Replay.

bvaughn (approved):
No strong feelings about this

src/ui/components/Tour/Tour.module.css
jonbell-lot23: This is appropriate because the Tour does not change in different themes. It's always this color.

I understand the argument that everything should have a variables and we can just set the variable to the same color for both dark and light theme, but I disagree with it.

bvaughn: That's kind of an implementation detail and could be broken by a change to either theme– which would not know to scan for hard-coded colors like this. We really shouldn't be adding hard-coded colors to our styles.

jonbell-lot23: I'll change it, but it's important to understand this is a difference of opinion, not objective fact.

This CSS sets an animating background to slightly fade between a dark purple and a deep blue. These are hard-coded to those bgcolor values, and I chose to set the color value the same way. This makes the component and the css self-encapsulating. Theme changes wouldn't affect how this component renders, and I believe there is a benefit to describing single values once rather than having to hunt it down in two places in variables.css.

But happy to disagree and change to your preference here.

src/ui/components/Tour/Tour.tsx
jonbell-lot23: Good call. If I'm reading blame correctly, it looks like you added this 4 months ago:

```
  const typeOfReplay = recording && isTestSuiteReplay(recording) ? "cypress" : "events";
```

So I was following that. Happy to change this, do you want me to say != cypress here instead?

src/ui/components/Tour/Tour.module.css
jonbell-lot23: Ok @bvaughn I've added --tour-body-color and --tour-callout-body-color to address these

src/ui/components/Tour/Tour.tsx
jonbell-lot23: Since there's an approval I am going to merge this to get it live, but happy to fix this in a follow-up

bvaughn:
A ton of automated tests failed for this PR. I wonder why.

src/ui/components/Tour/Tour.tsx
bvaughn: > If I'm reading blame correctly, it looks like you added this 4 months ago

This is no big deal but it got me curious. It was hard to imagine that I wrote that bc the naming choice seemed so strange to me.

The revision from 4 months ago was just a formatting change. The original lines came from PR #8956. 😆

jonbell-lot23: Oh! Got it. That PR is the best PR in the history of PRs.

I'll address this in a follow-up.

jonbell-lot23:
> A ton of automated tests failed for this PR. I wonder why.

Oh, I see that now. Well, feel free to revert if necessary.

bvaughn:
Looking at a few of them at random, I don't think it was related to your change. I think maybe the backend had an issue?

src/ui/components/Tour/Tour.tsx
jonbell-lot23: Hey I'm about to adjust this @bvaughn and I'm curious if you have a naming preference here. I often say "manual replay" or "standard replay." Do you have a term you've used in the code that I can align to like "default" or "standard?"

bvaughn: That's a good question!

Maybe we could just side-step the name entirely by using a boolean (`isTestReplay`?) 🤪

jonbell-lot23: Sounds groovy, thanks!

PR #9650, opened 2023-08-29
Use the displayed focus window only in the timeline and the active one everywhere else

hbenl:


packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -99,8 +99,8 @@ function PointPanel(props: ExternalProps & { focusRange: TimeStampedPointRange }
     hitPointsForLocationCache,
     client,
     { begin: focusRange.begin.point, end: focusRange.end.point },
-    pointForSuspense.location,
-    pointForSuspense.condition
+    pointForDefaultPriority.location,
+    pointForDefaultPriority.condition
```
hbenl, line 103: I've changed this to `pointForDefaultPriority` because since #9531 this component doesn't suspend anymore.

packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -133,27 +140,16 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
...
-      // In this case, React's default high/low priority split is helpful, but we can do more.
-      // Debouncing a little before starting the transition helps avoid sending a lot of unused requests to the backend.
-      if (debounce) {
-        debouncedSetDeferredRange(range);
-      } else {
-        startTransition(() => {
-          setDeferredRange(range);
-        });
-      }
+      setRangeForDisplay(range);
```
hbenl, line 143: This will trigger the effect above which updates `range` and `rangeForSuspense`.

src/ui/reducers/timeline.test.ts
```diff
@@ -236,21 +237,25 @@ describe("Redux timeline state", () => {
...
-              "point": "65000",
-              "time": 65,
-            },
-          }
-        `);
-      });
+      // TODO for some reason the test setup doesn't seem to work: it should create a new mock replayClient
+      // with its focusWindow set to null for every test but for some reason this test gets a mock replayClient
+      // with a focusWindow left over from another test: the test works when run separately but not when run
+      // together with the other tests in this file.
```
bvaughn, line 243: Try adding and then re-enabling this test?
```js
  afterEach(() => {
    jest.resetModules();
  });
```

src/ui/components/FocusContextReduxAdapter.tsx
```diff
@@ -2,31 +2,32 @@ import { TimeStampedPointRange } from "@replayio/protocol";
...
 import { getFocusWindow } from "ui/reducers/timeline";
 import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
-import { rangeForFocusWindow } from "ui/utils/timeline";
 
 // Adapter that reads focus region (from Redux) and passes it to the FocusContext.
 export default function FocusContextReduxAdapter({ children }: PropsWithChildren) {
   const dispatch = useAppDispatch();
-  const focusWindow = useAppSelector(getFocusWindow);
+  const range = useCurrentFocusWindow();
+  const rangeForDisplay = useAppSelector(getFocusWindow);
```
bvaughn, line 21: I think it would be nice to remove focus range from Redux entirely so we could avoid hoops like this. (Not something we need to do _now_, just thinking out loud.)

packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -80,29 +88,28 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
...
+    const timeoutId = setTimeout(async () => {
+      const range = await client.requestFocusWindow({
+        begin: rangeForDisplay.begin,
+        bias,
+        end: rangeForDisplay.end,
+      });
+      setRange(range);
+      startTransition(() => {
+        setRangeForSuspense(range);
+      });
```
bvaughn, line 106: I think there's a small potential race here, if this effect gets destroyed in between the `await` on line 98 and the state updates on lines 103-106.

I think we need something like this to make it safe:
```js
  let cancelled = false;

  const timeoutId = setTimeout(async () => {
    const range = await client.requestFocusWindow({
      begin: rangeForDisplay.begin,
      bias,
      end: rangeForDisplay.end,
    });

    if (!cancelled) {
      setRange(range);
      startTransition(() => {
        setRangeForSuspense(range);
      });
    }
  }, FOCUS_DEBOUNCE_DURATION);

  return () => {
    cancelled = true;

    clearTimeout(timeoutId);
  };
```

bvaughn (approved):
I think this change looks good, although I didn't verify all of the `range` vs `rangeForSuspense` usages :) I trust you verified those.

One race that I think you need to add a guard for, and I think you can reenable that test you commented out, but otherwise 👍🏼

src/ui/reducers/timeline.test.ts
hbenl: That didn't help.

bvaughn: Damn. File a follow up Linear issue?

hbenl: FE-1855

PR #9639, opened 2023-08-25
Put layout options into the menu

jonbell-lot23:
**Review note**

Brian, you're the reviewer so as always I'm looking for ways to make the code better. But please note that we may have design feedback coming through, so this PR may stay open for a bit while we discuss any concerns.

**Background**

This has been a long-term problem with our product. People expect to be able to pin their panes to the side or bottom, but most people don't discover the functionality. This finally makes the feature discoverable and addresses https://linear.app/replay/issue/DES-743/[explore]-put-layout-options-into-our-global-menu. ([Figma](https://www.figma.com/file/mAotNZRv6M5X0zHIFKPzFY/Replay-Design-Doc?type=design&node-id=18487-160698&mode=design&t=cdbbM6HKVfap7sWT-4))

[Loom of the feature in action](https://www.loom.com/share/dad97764c1bc435cae9137657fae97d9)

Screenshot:
image

PR #9651, opened 2023-08-29
Update all e2e tests to wait for loading+indexing to finish before starting

bvaughn:
Hopefully this helps offset the recent increase in flakiness seemingly related to cloning recordings before running tests


bvaughn:
That's good to note.

Ah I bet the other Console e2e test was broken by @jonbell-lot23's recent changes in #9639 and I just didn't notice it at the time because e2e tests have been flaking lately

bvaughn:
#9653 should fix the Console regression. I'll follow up with breakpoint-04 via FE-1860

PR #9660, opened 2023-08-30
Move Video toggle button to app toolbar

bvaughn:
- [x] Update `useLocalStorageUserData` to subscribe to changes made to `localStorage` from other places in the React tree
- [x] Move video toggle button into the app toolbar

Kapture 2023-08-30 at 15 27 35

cc @jonbell-lot23


markerikson (approved):

jonbell-lot23 (commented):
1. I do think keeping the video blue when enabled is an important indicator of set state. This is an opinion, and I won't be surprised if/when others agree, but that's my opinion about that.

2. Can you add a trackEvent for this?

jonbell-lot23:
Lol, missed it by 2 minutes

bvaughn:
> I do think keeping the video blue when enabled is an important indicator of set state. This is an opinion, and I won't be surprised if/when others agree, but that's my opinion about that.

Ah, I misunderstood the mock on the issue. My bad.

bvaughn:
#9661

PR #9662, opened 2023-08-30
Rewrite "Elements" panel to use data caches instead of Redux

markerikson:
This PR rewrites the "Elements" panel data fetching and management logic to speed up fetching and displaying DOM nodes, especially when you select a deeply nested node with the node picker:

- Removed the calls to get CSS rules and computed styles on initial load, as well as all `NodeInfo` fields and `<Node>` logic that depended on those values (graying out hidden elements, `flex/grid` badges, `event` badges)
- DOM node data is now read directly from caches in components, instead of kept in Redux. Ditto for parsed CSS rules, box models, and computed CSS properties.
  - "Expanded" states are still kept in Redux, to facilitate setting them in bulk when we select a deeply nested node
- Selecting deeply nested nodes should now show them way faster (like, immediately, instead of popping in one level at a time)
- I've heavily reworked the Elements panel loading sequence to make it more resilient to race conditions like rapidly clicking between points on the timeline
- Added all of the DOM protocol methods we use to `ReplayClient`
- We avoid highlighting or fetching box models for nodes that can't have a layout, or are hidden

Still need to:

- Sort out issues with the keyboard shortcuts - behavior seems to have changed
- Add E2E tests (which I'll probably do as a separate PR to land first and verify existing behavior before rebasing this on top)


markerikson:
Wrote some thoughts on the complexity of the keyboard shortcuts logic in Discord and thought it's worth pasting here:


- the key shortcut logic is _really_ complex to start with
- it's made more so by the fact that it needs all the node data available synchronously. that was fine when it was all in Redux, _and_ we only updated the list of `children` when a child was finished loading. now that we're loading data async, it got a lot trickier.
- then you've got the "filter out whitespace nodes" aspect. if we had a list ahead of time of what child nodes _were_ renderable, half those shenanigans would go away. but we don't. we have to fetch them all first and filter on the client.
- and to top it off, the Elements tree is written with recursive rendered components, which means you can't just go to one place to look at what nodes are visible.

I'm pretty sure that the RDT Components tree is a flat virtualized list, which also means it knows exactly what component IDs are rendered at any given time, and the up/down behavior would be as simple as `ids[currentIdIdx - 1]`, etc

so at least the inspector tests seem to be passing now, but the RDT tests seem to be failing consistently and I haven't looked into those yet.

src/ui/suspense/nodeCaches.ts
```diff
@@ -173,27 +160,157 @@ export const nodeDataCache: Cache<
...
+// a way to bulk-fetch object previews
+export const renderableChildNodesCache: Cache<
+  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
+  NodeInfo[] | null
+> = createCache({
+  config: { immutable: true },
+  debugLabel: "ChildNodesData",
+  getKey: ([replayClient, pauseId, nodeId]) => `${pauseId}:${nodeId}`,
+  load: async ([replayClient, pauseId, nodeId]): Promise<NodeInfo[] | null> => {
+    if (!pauseId || typeof nodeId !== "string") {
```
bvaughn, line 283: This is unexpected, given the way the types are defined for this cache:
```js
  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
```
Which is wrong?

src/ui/suspense/nodeCaches.ts
```diff
@@ -173,27 +160,157 @@ export const nodeDataCache: Cache<
...
+
+export const ancestorNodesCache: Cache<
+  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
+  NodeInfo[] | null
+> = createCache({
+  config: { immutable: true },
+  debugLabel: "AncestorNodesData",
+  getKey: ([replayClient, pauseId, nodeId]) => `${pauseId}:${nodeId}`,
+  load: async ([replayClient, pauseId, nodeId]): Promise<NodeInfo[] | null> => {
+    if (!pauseId || typeof nodeId !== "string") {
```
bvaughn, line 209: This is unexpected, given the way the types are defined for this cache:
```js
  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
```
Which is wrong?

src/ui/suspense/nodeCaches.ts
```diff
@@ -173,27 +160,157 @@ export const nodeDataCache: Cache<
...
 
+export const processedNodeDataCache: Cache<
+  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
+  NodeInfo | null
+> = createCache({
+  config: { immutable: true },
+  debugLabel: "ProcessedNodeData",
+  getKey: ([replayClient, pauseId, nodeId]) => `${pauseId}:${nodeId}`,
+  load: async ([replayClient, pauseId, nodeId]): Promise<NodeInfo | null> => {
+    if (!pauseId || typeof nodeId !== "string") {
```
bvaughn, line 171: This is unexpected, given the way the types are defined for this cache:
```js
  [replayClient: ReplayClientInterface, pauseId: PauseId, nodeId: string],
```
Which is wrong?

src/devtools/client/inspector/rules/actions/rules.ts
```diff
@@ -1,66 +0,0 @@
-import { assert } from "protocol/utils";
-import type { UIStore } from "ui/actions";
-import { isInspectorSelected } from "ui/reducers/app";
-import { AppStartListening } from "ui/setup/listenerMiddleware";
```
bvaughn, line 4: I think you probably meant to delete `/reducers/rules.ts` and `/selectors/rules.ts` too, but forgot to?

src/devtools/client/inspector/markup/reducers/markup.ts
```diff
@@ -179,6 +149,14 @@ const markupSlice = createSlice({
   },
   extraReducers: builder => {
     // dispatched by actions/timeline.ts, in `playback()`
+    builder.addCase(pauseRequestedAt, () => {
+      console.log("Resetting markup state");
```
bvaughn, line 153:
```suggestion
```

src/devtools/client/inspector/markup/components/TextNode.tsx
```diff
@@ -2,6 +2,7 @@ import React, { PureComponent } from "react";
 
 import NodeConstants from "devtools/shared/dom-node-constants";
 import { assert } from "protocol/utils";
+import { reIsNotWhiteSpace } from "ui/suspense/nodeCaches";
```
bvaughn, line 5: What is this? What is "re"? I'm guessing "regular expression"?

Can we name this something more semantically meaningful?

bvaughn (commented):
Initial pass! Left a few questions.

I only briefly scanned the changes in `src/devtools/client/inspector/markup/actions/markup.ts` bc it's a big file 😅

src/devtools/client/inspector/computed/components/ComputedToolbar.tsx
```diff
@@ -35,19 +37,12 @@ function ComputedToolbar(props: PropsFromRedux) {
...
 
-const mapDispatchToProps = {
-  setComputedPropertySearch,
-  setShowBrowserStyles,
-};
-const connector = connect(undefined, mapDispatchToProps);
-type PropsFromRedux = ConnectedProps<typeof connector>;
-
-export default connector(ComputedToolbar);
+export default React.memo(ComputedToolbar);
```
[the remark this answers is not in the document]
markerikson: Habit, I guess :)

src/devtools/client/inspector/computed/reducers/index.ts
```diff
@@ -3,17 +3,12 @@ import { ComputedAction } from "../actions";
...
 };
 
 const reducers: ReducerObject<ComputedState, ComputedAction> = {
-  set_computed_properties(state, { properties }) {
-    return { ...state, properties };
-  },
-
   set_computed_property_search(state, { search }) {
     return { ...state, search };
   },
```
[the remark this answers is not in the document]
markerikson: Not an oversight, so much as I was trying to limit how much code I touched in the actual components.

src/ui/suspense/styleCaches.ts
```diff
@@ -113,26 +114,154 @@ export const computedStyleCache: Cache<
   },
 });
 
+export function getDeclarationState(declaration: TextProperty, ruleId: string): DeclarationState {
```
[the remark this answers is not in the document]
markerikson: Yeah, I intended to move this, and forgot to clean up the original files.

src/ui/suspense/nodeCaches.ts
markerikson: The argument types, so I should presumably update those (which will have ripple effects).

This is an example of why I was asking for a `skip` option for the `useImperativeCacheValue` hook.  There _are_ cases where we will not have a valid pause ID or node ID, but right now we can't skip calling into the cache.

src/devtools/client/inspector/markup/components/TextNode.tsx
markerikson: I've always prefixed my regex variable names with `re`, yeah. Longstanding habit.  Better suggestions?

bvaughn:
Looks like selecting something with the node picker doesn't scroll the elements panel so the selected thing is visible. It used to.

src/devtools/client/inspector/markup/components/TextNode.tsx
bvaughn: Whoa! "re" means...nothing to me, except as a prefix meaning "again"

Maybe "regex" ?

src/ui/suspense/nodeCaches.ts
bvaughn: Seems like updating the types is appropriate then I guess.

markerikson:
I bet the scrolling behavior is because there's other sibling tree nodes that are getting added a bit later.  So, it scrolls to the right position initially, then the siblings pop in and push it further down in the list...

markerikson:
Okay, added a delay to the scroll behavior.  I _think_ this is good now.

bvaughn:
If you think it's good, and tests are passing, let's do it. We can always iterate more

bvaughn:
FWIW it still isn't scrolling for me, but I don't think that's worth holding up the rest of the changes.

markerikson:
I'd tried having the scroll timer set at 500ms, and I guess that's too slow. Just bumped it to 1s, and that seems to behave better.  It's still a hack and we should come up with a better option, but hopefully sufficient for now? :)

src/devtools/client/inspector/markup/actions/markup.ts
```diff
@@ -74,63 +73,59 @@ export function setupMarkup(store: UIStore, startAppListening: AppStartListening
...
+      // by the time we get here, from the `pauseRequestedAt` action.
+
       const originalPauseId = await ThreadFront.getCurrentPauseId(replayClient);
 
-      // every time we pause, clear the existing DOM node info
-      dispatch(reset());
+      async function throwIfCanceled() {
+        // Cancel-aware - see if this is still the current pause
+        // We'll have this added to RTK at some point
+        await listenerApi.delay(0);
```
bvaughn, line 88: Hey @markerikson – as part of looking into FE-1907, I'm seeing this code and I realize that I don't know what it means. It doesn't look like it "throws" despite the name. Could you help me understand what it does?

PR #9663, opened 2023-08-31
Elements: Rules panel re-write

bvaughn:
### [Loom overview](https://www.loom.com/share/8cb11278bcf9498cb8fc415c9ff30635)

- [x] Rewrite Elements > Rules panel to improve performance
- [x] Delete old (no longer used) Rules components and CSS
  - There are still some old components because of the Layout and Computed tabs; maybe we should kill those too?
- [x] Update e2e test helper methods to use data attributes rather than CSS class names


src/devtools/client/inspector/markup/components/rules/RulesPanel.tsx
```diff
@@ -0,0 +1,78 @@
...
+    deferredSelectedNodeId ?? undefined
+  );
+
+  const canHaveRules = !!node?.isElement;
+
+  const cachedStyles = cssRulesCache.read(
+    replayClient,
+    canHaveRules ? deferredPauseId : undefined,
+    canHaveRules ? deferredSelectedNodeId : undefined
+  );
```
bvaughn, line 45: [This video](https://www.loom.com/share/b1168cda0b674f688a71e2b1a4b3e9f2) explains why I also made this change.

packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -1,5 +1,6 @@
 import { Locator, Page, expect } from "@playwright/test";
 import chalk from "chalk";
+import { string } from "yargs";
```
markerikson, line 3: :hand: Accidental auto-import?

markerikson (approved):
One stray auto-import, but seems reasonable otherwise!

packages/e2e-tests/helpers/elements-panel.ts
bvaughn: Oops, y up

PR #9665, opened 2023-08-31
Test Suites: Re-apply fix for FE-1833

bvaughn:
Fixes FE-1833 (which got reverted) and FE-1875

Refactors the user-action type so that we're tracking three times of time stamped points now:
* One range to set focus so that we can view the test code source (`"step:enqueue"` … `"step:end"`)
* One range to show the before and after action in the UI (`"step:start"` … `"step:end"`)
* One point to view the test code source (`"step:enqueue"` or `"step:end"`, depending on the value of `command.name`)

I used recording 1a098592-1330-4e81-9391-dce178fc1ce2 as a test case for comparing this branch to production. I used recording b6ded8cf-f4b5-4844-a3d6-d311a981d2f3 to verify that this change still addresses the original bug report from FE-1833.

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -558,9 +558,6 @@ export async function processCypressTestRecording(
             assert(stepStartPoint !== null, `Missing "step:start" annotation for test event`, {
               id,
             });
-            assert(stepEnqueuePoint !== null, `Missing "step:enqueue" annotation for test event`, {
-              id,
-            });
             assert(viewSourcePoint !== null, `Missing annotation for test event`, {
               annotationType: isChaiAssertion ? "step:start" : "step:enqueue",
               id,
@@ -585,7 +582,7 @@ export async function processCypressTestRecording(
                 },
               },
               timeStampedPointRange: {
-                begin: stepEnqueuePoint,
+                begin: stepEnqueuePoint ?? stepStartPoint,
                 end: stepEndPoint ?? stepStartPoint,
               },
               type: "user-action",
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -255,10 +255,7 @@ export namespace RecordingTestMetadataV3 {
       parentId: string | null;
 
       // This value comes from annotations and so is only available for Cypress tests (for now)
-      result: {
-        timeStampedPoint: TimeStampedPoint;
-        variable: string;
-      } | null;
+      resultVariable: string | null;
 
       // If an action is somewhere other than the main test body;
       // for example, before/after actions have different scopes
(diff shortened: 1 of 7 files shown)


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -542,12 +555,13 @@ export async function processCypressTestRecording(
...
             });
 
-            assert(beginPoint !== null, `Missing "step:start" annotation for test event`, {
+            assert(stepStartPoint !== null, `Missing "step:start" annotation for test event`, {
               id,
-              isChaiAssertion,
             });
-
-            assert(viewSourceTimeStampedPoint !== null, `Missing annotation for test event`, {
+            assert(stepEnqueuePoint !== null, `Missing "step:end" annotation for test event`, {
```
ryanjduffy, line 561: Should this assert on `stepEndPoint` instead?

src/ui/components/TestSuite/views/TestRecording/useTestEventContextMenu.tsx
```diff
@@ -50,6 +50,7 @@ function JumpToSourceMenuItem({ userActionEvent }: { userActionEvent: UserAction
   });
 
   const onSelect = () => {
+    console.log(userActionEvent);
```
ryanjduffy, line 53: Extra log

bvaughn: 🤡 Thanks

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: On my phone so I can't see easily but will this error on chai assertions that lack r queue? If we're not sure, I can check in the morning

bvaughn: I am so confused by these inconsistencies in the data 😞 

Yes, enqueue is required as of this change because we need it for focus. I did not think it would be missing entirely for "chai assertions" just that it wasn't the point we wanted to use for viewing the source code.

ryanjduffy (approved):
Checked this [flake replay](https://devtools-git-fe-1875-recordreplay.vercel.app/recording/cypresse2ecast-alias-spects--a2912e3b-6edd-4b8f-853f-562eb4a4a422?point=5516815412789020778211717490409825&time=2475&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjU1MTY4MTU0MTI3ODkwMjA3NzgyMTE3MTc0OTA0MDk4MjUiLCJ0aW1lIjoyNDc1fSwiZW5kIjp7InBvaW50IjoiNTUxNjgxNTQxMjgyNTEwMjYwOTYxOTg5MzM3MzM3MDc5MCIsInRpbWUiOjI1MDd9fQ%253D%253D) which includes chai assertions and it checks out.

bvaughn:
Thanks @ryanjduffy

PR #9666, opened 2023-08-31
Add and rework E2E tests for the "Elements" panel

markerikson:
This PR:

- Renames all of the `inspector-*` tests to clarify what they actually do (such as `inspector-rules-02_sourcemapped-rules.test.ts`)
- Adds a Chromium-based "golden recording" for the `doc_stacked.html` file 
  - adds a copy-paste of the existing `stacking.test.ts` for Chromium, but disabled for now - it's definitely not passing and I haven't looked into why yet (not sure if it's due to general Chromium differences, or the specific NodePicker behavior Kannan's been working on)
- Moves the stacking test case definitions and helper utilities into `helpers/elements-panel.ts`
- Adds a test for keyboard selection behavior in the "Elements" tree
- Adds a test that verifies selecting a node via Node Picker or search properly expands all its ancestors


bvaughn:
Nice! Looking forward to reviewing this one in the morning! 🥳

packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -293,7 +297,216 @@ export async function toggleMarkupNode(page: Page, locator: Locator, open: boole
...
+  page: Page,
+  key: string,
+  expectedElement: string
+) {
+  debugPrint(page, `Typing ${key}...`);
+  await page.keyboard.press(key);
+  await waitForSelectedElementsRow(page, expectedElement);
+}
+
+// `doc_stacking.html` data and utils
```
bvaughn, line 316: organizational nit: In our screenshot tests, we put helpers that are specific to a group of tests in a folder with those tests. We don't have to do that here but it does seem a little odd to have test-specific helpers in the general "elements-panel" utils file.

We don't have the same folder organization with these e2e tests so that's not as feasible, but maybe we could at least move these into a `helpers/stack-test-case.ts` file so it's a bit clearer?

Alternately you could just move the `StackingTestCase` and `stackingTestCases` data out somewhere (`data/stacking-test-cases.ts`?) and leave the methods below in this file.

packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -293,7 +297,216 @@ export async function toggleMarkupNode(page: Page, locator: Locator, open: boole
...
+  // children of positioned elements are in front of children of unpositioned ones
+  {
+    id: "r2c3",
+    position: { x: 240, y: 240 },
+    expectedRules: ["element", ".box1"],
+  },
+
+  // The last row has 4 sets of boxes, but some additional checks
+
+  // TODO [FE-626] This lower-left set is finicky and not working right with this  test setup atm.
```
bvaughn, line 409: Is this comment still relevant? FE-626 is over a year old and seems pretty stale.

Looks like it was copied over verbatim from `packages/e2e-tests/tests/stacking.test.ts` – might be worth taking a quick look to see if it's worth copying this over?

packages/e2e-tests/helpers/pause-information-panel.ts
```diff
@@ -45,6 +45,18 @@ export async function expandAllScopesBlocks(page: Page): Promise<void> {
   });
 }
 
+export async function ensureSidePanelClosed(page: Page) {
+  // Clicks that aren't directly on an element can cause the "Comments" pane to open.
```
bvaughn, line 49: I don't understand this comment. What "elements"? Why would clicking on something randomly open the pause panel?

Looking at the places you're calling this (after `warpToMessage` and after `verifyStackingTestCaseSelectedElementUnderCursor` I'm kind of concerned that one of our test helpers is really flaky and we should **fix it** rather than this bandaid.

packages/e2e-tests/tests/inspector-elements-03_nested-node-selection.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  `<div style="left: 200px; top: 200px;"`,
+  `<div style="left: 200px; top: 200px;"`,
+  `<div style="left: 300px; top: 200px;"`,
+  `<div style="left: 0px; top: 300px;"`,
+  `<div style="left: 100px; top: 300px;"`,
+  `<div style="left: 200px; top: 300px;"`,
+  `<div style="left: 300px; top: 300px;"`,
+];
+
+const div0ChildDomNodes = [`<div class="box1"`, `<div class="box2"`];
```
bvaughn, line 54: Not referenced.

packages/e2e-tests/tests/inspector-elements-03_nested-node-selection.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  selectElementsRowWithText,
+  typeKeyAndVerifySelectedElement,
+  waitForElementsToLoad,
+  waitForSelectedElementsRow,
+} from "../helpers/elements-panel";
+import {
+  stackingTestCases,
+  verifyStackingTestCaseSelectedElementUnderCursor,
+} from "../helpers/elements-panel";
+import { toggleToolboxLayout } from "../helpers/layout";
```
bvaughn, line 21: nit A bunch of these imports are unused (Opt+Shift+O will remove)

(This is true for all for all of the new test files fwiw)

bvaughn (commented):
Cool!

packages/e2e-tests/helpers/elements-panel.ts
markerikson: I can check to see if the behavior is any better.

packages/e2e-tests/helpers/pause-information-panel.ts
bvaughn: Ah, that's interesting. Yeah now there's a context menu that you have to explicitly click "add comment" in.

Maybe we can remove this and see if things still pass?

markerikson: Yeah, seems to pass if I just call it once at the start of the test. (I still want it closed in some cases to increase the size of the video/elements pane.)

packages/e2e-tests/helpers/elements-panel.ts
bvaughn: Gotcha. That's a bummer but if you think it's out of scope of this PR, that's understandable

bvaughn (approved):
What's the name "stacking" mean btw? I was wondering that earlier.

Approved but could you rename/update that one helper function before merging?

markerikson:
"Stacking" is about the overlapping of DOM nodes in various style configurations (absolute/relative, z-index, etc), and which node should get picked based on how the browser is overlaying them on top of each other.

bvaughn:
Oh! z-index stacking. Ok that makes sense. I was thinking "stack trace" and "call stack" and neither made sense.

PR #9667, opened 2023-08-31
Add a regression test for FE-1876

jasonLaster:
This PR adds a regression test for FE-1876 to ensure that clicking on a test step and jump to code work as expected.

I expect using public recordings like this will be somewhat controversial, but I'd like to see how it works in practice especially when we can easily replace this recording with a new react hook form one from CI in the future.


markerikson:
> I expect using public recordings like this will be somewhat controversial

Already doing that with the Cypress and Redux tests, actually.

bvaughn:
Which issue is this PR referring to? Doesn't seem like FE-1876 is the right one. I think you meant to link this to FE-1875?

packages/e2e-tests/tests/regression-fe-1875-cypress-jump-to-code.test.ts
```diff
@@ -0,0 +1,36 @@
+import { openViewerTab, startTest } from "../helpers";
+import { waitForPaused } from "../helpers/pause-information-panel";
+import { getSelectedTestCase, getTestCaseSteps, openCypressTestPanel } from "../helpers/testsuites";
+import { waitForTimelineAdvanced } from "../helpers/timeline";
+import test, { expect } from "../testFixtureCloneRecording";
+
+// React Hook Form cypress/e2e/conditionalField.cy.ts
+// When this test is no longer supported, go to
+// https://app.replay.io/team/dzozMDAxY2NkMC1mNGNhLTQ1ZjctYTZkYy0zOGE2MjA2Y2U4M2I=/runs
+// and select the most recent cypress/e2e/conditionalField.cy.ts
```
bvaughn, line 10: 👍🏼

packages/e2e-tests/tests/regression-fe-1875-cypress-jump-to-code.test.ts
```diff
@@ -0,0 +1,36 @@
...
+const recordingId = "122c4fe7-8a58-4c5c-9639-16b35b5f2ef1";
```
bvaughn, line 11: Rather than referring to recording ids directly, tests should use an `exampleKey` (corresponding to `packages/e2e-tests/examples.json`) which should be passed to `test.use()` (see other tests for examples). This ensures recordings get cloned before running tests, to avoid two tests interfering with each other.

bvaughn (commented):
Which issue is this PR referring to? Doesn't seem like FE-1876 is the right one. I think you meant to link this to FE-1875?

> I expect using public recordings like this will be somewhat controversial, but I'd like to see how it works in practice especially when we can easily replace this recording with a new react hook form one from CI in the future.

Regarding this: My only concern is that we have a reliable and easy way to regenerate these recordings when they expire. (If a recording is made from another recordings– which will itself expire– then we'll also need a low-friction way to regenerate the original). So long as that's true, no worries.

Conditionally approve this– but please fix the `test.use()` thing before merging. 😄

PR #9683, opened 2023-09-01
Layout options should not auto-close dropdown menu

bvaughn:
This change on its own would break the stacking tests, but I updated the layout helper to avoid that. (Didn't touch the stacking tests themselves since Mark has moved and rewritten some of them in another open PR.)


bvaughn:
This seems really uncontroversial, and PRs are stacking up a bit, so I'm going to just land this one.

PR #9684, opened 2023-09-01
Delete legacy mock tests and scripts

bvaughn:
This PR also deletes a bug chunk of "mock" junk that I don't think we're using or needing. (There might be more to delete– but I searched for "mock" stuff and cleaned out all I saw.)


bvaughn:
A single test in the "Unit Tests" action failed– `AppModal.test.tsx`

IMO this test isn't really very valuable, and should be an e2e test anyway, rather than a unit test. I'm going to try killing it too.

PR #9685, opened 2023-09-01
Update playwright timeouts + retries

jasonLaster:
I think it would be helpful to see if extending our timeouts would let us use fewer retries.

How will this impact test run times? I don't think it will slow down run times because Playwright will still retry at the same interval. It will help us reduce the instances of backend performance issues causing tests to fail.


bvaughn (approved):
We cab try it and see. I'm not optimistic about the 30s to 60s total timeout change, but the 5s to 10s expect timeout might help a lot.

PR #9686, opened 2023-09-01
Fix network-02

jasonLaster:
Adds "Timings" to the expected list of tabs. Also updates the package.json commands

### DIFF
--- a/packages/e2e-tests/tests/network-02.test.ts
+++ b/packages/e2e-tests/tests/network-02.test.ts
@@ -31,14 +31,14 @@ test(`network-02: should show details for the selected request`, async ({
   await selectRequestRow(page, {
     name: "image/svg+xml",
   });
-  await verifyNetworkDetailsTabsVisible(page, ["Headers", "Response"]);
+  await verifyNetworkDetailsTabsVisible(page, ["Headers", "Response", "Timings"]);
 
   // Select a JavaScript file request
   await filterByText(page, "SpecPatterns");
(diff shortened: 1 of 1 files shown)


packages/e2e-tests/package.json
```diff
@@ -7,9 +7,8 @@
     "test": "tests"
   },
   "scripts": {
-    "test": "playwright test",
-    "test:chromium": "playwright test --project chromium --reporter=@replayio/playwright/reporter,line",
-    "test:replay-chromium": "playwright test --project replay-chromium --reporter=@replayio/playwright/reporter,line",
+    "test": "playwright test --project replay-chromium --reporter=@replayio/playwright/reporter,line",
+    "test:ui": "yarn test --ui",
```
bvaughn, line 11: Hmm, can we land this change separately? I'd like to think about it more.

packages/e2e-tests/tests/network-02.test.ts
```diff
@@ -31,14 +31,14 @@ test(`network-02: should show details for the selected request`, async ({
   await selectRequestRow(page, {
     name: "image/svg+xml",
   });
-  await verifyNetworkDetailsTabsVisible(page, ["Headers", "Response"]);
+  await verifyNetworkDetailsTabsVisible(page, ["Headers", "Response", "Timings"]);
```
bvaughn, line 34: This part looks good 👍🏼

PR #9687, opened 2023-09-01
Source preview popup shows "Loading..." immediately

bvaughn:
### [Loom overview](https://www.loom.com/share/c2f13b9c15d044b5916f94419cafef18)

- [x] Show source preview popup immediately when hovering over an inspectable token. (We used to render a null fallback; now we'll show a "Loading..." popup.)
- [x] Update `Popup` repositioning logic to run synchronously on mount. (We used to rely on `ResizeObserver` only, but it seems not to call our reposition method until after paint– which causes a small layout shift/flicker. This fix is important to land even if we decided not to land the other one for some reason.)

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/Popup.tsx
+++ b/packages/replay-next/components/Popup.tsx
@@ -186,6 +186,10 @@ export default function Popup({
     observer.observe(popover);
     observer.observe(target);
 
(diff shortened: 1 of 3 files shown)


markerikson (approved):

jasonLaster:
Looks great!

PR #9688, opened 2023-09-01
Fix case where applyMiddleware decl can't be found (FE-1847)

Mokshit06:
Fixes [FE-1847](https://linear.app/replay/issue/FE-1847/[pinterest]-redux-j2c-does-not-load-the-source)

The J2C button wasn't working because it previously assumed that `applyMiddleware()` function decl would always be existing and could be found, but in this case it couldn't be and we weren't handling the case where that happened. So accessing a value on the location threw an error, and the code to jump never actually ran.
It should now fallback to searching the source outline for `store => next => action => {}`

### DIFF
--- a/src/ui/suspense/jumpToLocationCache.ts
+++ b/src/ui/suspense/jumpToLocationCache.ts
@@ -24,11 +24,11 @@ interface ApplyMiddlewareDecl {
   sourceId: string;
 }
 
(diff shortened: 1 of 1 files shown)


jasonLaster:
@Mokshit06 could we add a regression test for minified redux apps?

markerikson:
Seems reasonable!

PR #9689, opened 2023-09-01
Update playwright config

jasonLaster:
* yarn test now defaults to chromium
* yarn test:replay is equivalent to `yarn test` but records with Replay
* yarn test:ui adds playwrights UI mode

NOTE: I replaced `--project` with an environment variable because I could not get the playwright CLI to behave. I'd like to file an issue with the playwright project and see why it is failing.


bvaughn:
Closing as stale/abandoned

PR #9693, opened 2023-09-04
Add "Jump to Network Request" to Passport

jonbell-lot23:
As a follow-up to xxx, we are filling out our "jump" section in the passport. This PR adds "Jump to network request":

image

--

- Added passport/nag logic to track this event (note that I used seekToRequestWrapper, let me know if there's a better way to handle that bit)
- New gif/video
- New docs page
- New passport stamp
- Also I spotted a visual regression with our time travel button in light theme. [Short loom showing it here](https://www.loom.com/share/889c6b0b235f4595b61e2c17f823dc2f). This is another situation where hard-coding the foreground is appropriate because the background is also hardcoded, and shouldn't change based on theme.


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-862/add-jump-to-network-request-to-passport

PR #9695, opened 2023-09-05
Increase max log points panel analysis hits from 200 to 500

bvaughn:
Builds on top of #9696 and increases the maximum log points panel analysis hits from 200 to 500.


markerikson (approved):

Andarist:
Thank you ❤️

PR #9697, opened 2023-09-05
Fix `breakpoints-07` flakiness

markerikson:
Per https://app.replay.io/recording/breakpoints-07-rewind-and-seek-using-command-bar-and-console-messages--c129861e-7a6b-410e-a79e-fb02e4f4b207 , `breakpoints-07` sometimes gets stuck because it tries to click on the "DevTools" button to ensure we're on the right tab, but then continues on and shows the "Quick Open" modal immediately without waiting.  

I think there's a couple issues that have been happening:

- The logic for typing into the "Quick Open" modal and selecting a filename wasn't waiting to make sure that the list was loaded and that the _right_ filename was selected
- It was also trying to click on the "DevTools" tab without waiting for the modal to go away (and related to that, it didn't even _need_ to click on the tab in this case)

Updated the logic to more specifically wait for the _selected_ source row with the right text, and `openDevTools/Viewer` to skip clicking if the right tab is already active.

### DIFF
--- a/packages/e2e-tests/helpers/commands.ts
+++ b/packages/e2e-tests/helpers/commands.ts
@@ -2,7 +2,7 @@ import { Page, expect } from "@playwright/test";
 import chalk from "chalk";
 
 import { waitForSourceContentsToFinishStreaming } from "./source-panel";
-import { debugPrint, getCommandKey, waitFor } from "./utils";
+import { debugPrint, delay, getCommandKey, waitFor } from "./utils";
 
 export async function quickOpen(page: Page, url: string): Promise<void> {
   await debugPrint(page, "Opening quick-open dialog", "quickOpen");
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/helpers/commands.ts
```diff
@@ -12,10 +12,21 @@ export async function quickOpen(page: Page, url: string): Promise<void> {
...
+  await waitFor(async () => {
+    const loader = page.getByText("Loading results...");
+    const isVisible = await loader.isVisible();
+    expect(isVisible).toBe(false);
+  });
+
   await debugPrint(page, `Opening file "${chalk.bold(url)}"`, "quickOpen");
   const sourceRow = await page.waitForSelector(
-    `[data-test-name="QuickOpenResultsList-Row"]:has-text("${url}")`
+    `[data-test-name="QuickOpenResultsList-Row"].selected:has-text("${url}")`
```
bvaughn, line 23: That's interesting.

packages/e2e-tests/helpers/commands.ts
```diff
@@ -12,10 +12,21 @@ export async function quickOpen(page: Page, url: string): Promise<void> {
...
   );
+  await waitFor(async () => {
+    const isVisible = await sourceRow.isVisible();
+    expect(isVisible).toBe(true);
+  });
```
bvaughn, line 28: I actually thought `waitForSelector` would do _the same thing_ but... I guess not?

bvaughn (approved):

packages/e2e-tests/helpers/commands.ts
markerikson: Oh, that's what we already _had_. Cool, that simplifies things I guess.

bvaughn (approved):
I think this seems okay 👍🏼  if it makes things less flaky

PR #9701, opened 2023-09-07
Speed up CSS parsing by caching frequently used pieces

markerikson:
This PR attempts to speed up some of the (very legacy!) CSS parsing logic, which is horribly slow and pathological when you inspect some of our own DOM nodes due to the thousands of Tailwind rules:

- Found a bunch of places where we were repeatedly recalculating strings or similar values, and threw in caches
- Turned some hot path array updates from immutable `concat()` to mutable `push()` (in some cases, 50K array copies were being made, totally unnecessarily)
- Moved the calculation of `ComputedPropertyState[]` from the `cssRulesCache` into its own cache, so that we don't do that calculation until you actually select the "Computed" panel

There's still a _lot_ of slowness here, but I think it's somewhat less bad now.

Here's one local dev perf profile before:

image

(yes, that is 7+ seconds of pure CPU, and 1500ms of GCs)

and what ought to be a couple equivalent parts of a perf profile with these changes applied:

image

image

500-1000ms still isn't _good_, but it's less _bad_

### DIFF
--- a/src/devtools/client/inspector/computed/actions/index.ts
+++ b/src/devtools/client/inspector/computed/actions/index.ts
@@ -4,6 +4,8 @@ import CSSProperties from "third-party/css/css-properties";
 import { OutputParser } from "third-party/css/output-parser";
 
 import ElementStyle from "../../rules/models/element-style";
+import RuleModel from "../../rules/models/rule";
+import { ComputedProperty } from "../../rules/models/text-property";
 import { ComputedPropertyState, MatchedSelectorState } from "../state";
 
 type SetComputedPropertySearchAction = Action<"set_computed_property_search"> & { search: string };
@@ -32,6 +34,8 @@ export function setComputedPropertyExpanded(
   return { type: "set_computed_property_expanded", property, expanded };
 }
 
+const cachedParsedProperties = new Map<string, (string | Record<string, unknown>)[]>();
+
 export async function createComputedProperties(
   elementStyle: ElementStyle,
   computed: Map<string, string> | undefined
@@ -41,6 +45,9 @@ export async function createComputedProperties(
   }
 
   const outputParser = new OutputParser(document, CSSProperties);
+
+  const NO_RULES: RuleModel[] = [];
+  const NO_COMPUTEDS: ComputedProperty[] = [];
 
   const properties: ComputedPropertyState[] = [];
   for (const [name, value] of computed) {
@@ -52,7 +59,7 @@ export async function createComputedProperties(
 
     let inheritanceCounter = 1;
     const selectors: MatchedSelectorState[] = [];
-    for (const rule of elementStyle.rules || []) {
+    for (const rule of elementStyle.rules || NO_RULES) {
       if (rule.isUnmatched) {
         continue;
       }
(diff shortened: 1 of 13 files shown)


src/devtools/client/inspector/rules/models/element-style.ts
```diff
@@ -406,11 +406,9 @@ export default class ElementStyle {
...
       // Collect all relevant CSS declarations (aka TextProperty instances).
       if (filterCondition) {
-        for (const textProp of rule.textProps.slice(0).reverse()) {
-          if (textProp.enabled) {
-            textProps.push(textProp);
-          }
-        }
+        const enabledTextProps = rule.textProps.filter(textProp => textProp.enabled);
+        enabledTextProps.reverse();
+        textProps.push(...enabledTextProps);
```
bvaughn, line 411: Especially since this PR is about optimizing things– this new bit looks like it could be optimized too, to avoid creating a throw-away array and having to `.reverse` it and spread it.

```js
for (let index = rule.textProps.length - 1; index >= 0; index--) {
  if (textProp.enabled) {
    textProps.push(textProp);
  }
}
```

src/devtools/client/inspector/rules/models/fronts/rule.ts
```diff
@@ -41,12 +44,24 @@ export class RuleFront {
     return this._rule.cssText;
   }
 
-  get selectorText() {
-    return this._rule.selectorText;
+  get cleanedSelectorText() {
```
bvaughn, line 47: I don't understand why this function is needed, or what a "clean selector text" is.

It looks kind of like this and the getter below only exist to add a space after a comma. Is that...right?

bvaughn (commented):
Love the idea of trying to reduce inefficiencies in our legacy code! I'm not sure I understand some of the change.

src/devtools/client/inspector/rules/models/fronts/rule.ts
markerikson: It's pretty stupid, really:

- We apparently split the original selector text on commas and trim each entry
- But this code wants a string that's effectively the same as the original except with an extra space after each comma.

Honestly I suspect we could probably just use the original selector text string as-is and it'd be essentially the same result, but I was trying to preserve the existing behavior here.

src/devtools/client/inspector/computed/actions/index.ts
```diff
@@ -32,6 +34,8 @@ export function setComputedPropertyExpanded(
   return { type: "set_computed_property_expanded", property, expanded };
 }
 
+const cachedParsedProperties = new Map<string, (string | Record<string, unknown>)[]>();
```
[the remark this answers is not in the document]
markerikson: These are all strings and values that are getting cached in Suspense anyway.  But sure, I can throw an LRU cache in here.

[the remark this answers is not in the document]
markerikson: I was seeing ~1500 items in a couple of the caches, so let's go with 3000 just to give space.

src/devtools/client/inspector/rules/models/fronts/rule.ts
bvaughn: > Honestly I suspect we could probably just use the original selector text string as-is and it'd be essentially the same result, but I was trying to preserve the existing behavior here.

Seems like this would be worth doing.

bvaughn:
Unit tests failing? https://github.com/replayio/devtools/actions/runs/6112862975/job/16591186484

markerikson:
Oh great, this is one of the Sindre Sorhus "ESM only" packages, and Jest is dying on that :(

bvaughn:
FWIW I've used `lru-cache` before and it seems fine

bvaughn:
In fact we're already pulling in `lru-cache` for `suspense` so we should just use it :)

markerikson:
Added the cleanup for the protocol messages on here - not directly part of the CSS perf, but might as well get it in.

PR #9707, opened 2023-09-08
Fix network panel color regression

jonbell-lot23:
Old:
image

New:
image

Another example of a blue background that calls for a white background (similar to our rewind/ff buttons)

### DIFF
--- a/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
+++ b/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
@@ -37,10 +37,10 @@
 .Row[data-selected]:focus,
 .Row[data-selected]:hover {
   background: var(--primary-accent-hover);
-  color: var(--color-default);
+  color: #fff;
 
-  --status-background-color: var(--color-default);
-  --status-color: var(--color-default);
+  --status-background-color: #fff;
+  --status-color: #fff;
 }
 
 /* These styles should stay in sync with NetworkMonitorListHeader */
--- a/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
+++ b/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
@@ -37,10 +37,9 @@
 .Row[data-selected]:focus,
 .Row[data-selected]:hover {
   background: var(--primary-accent-hover);
-  color: #fff;
+  color: var(--primary-accent-foreground-text);
 
-  --status-background-color: #fff;
-  --status-color: #fff;
+  --status-color: var(--primary-accent-foreground-text);
 }
 
 /* These styles should stay in sync with NetworkMonitorListHeader */
--- a/packages/design/AddCommentButton/AddCommentButton.module.css
+++ b/packages/design/AddCommentButton/AddCommentButton.module.css
@@ -23,11 +23,11 @@
   height: 26px;
   padding-left: 3px;
   border-radius: 3rem;
-  color: #fff;
   overflow: hidden;
   transition: width 180ms ease-out;
   width: 26px;
   background-color: var(--primary-accent);
+  color: var(--primary-accent-foreground-text);
   position: absolute;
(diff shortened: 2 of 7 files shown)


bvaughn (changes requested):
I think we end up having some form of "CSS variables vs hard-coded colors" discussion on most frontend PRs 😄

The only piece of UI I'm open to compromise on is Passport since that's intentionally designed to have only one color regardless of the theme. Every other area of the app should use CSS variables for color though.

src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
```diff
@@ -37,10 +37,10 @@
 .Row[data-selected]:focus,
 .Row[data-selected]:hover {
   background: var(--primary-accent-hover);
-  color: var(--color-default);
+  color: #fff;
 
-  --status-background-color: var(--color-default);
-  --status-color: var(--color-default);
+  --status-background-color: #fff;
+  --status-color: #fff;
```
bvaughn, line 43: Probably `--theme-selection-background` and `--theme-selection-color` are the styles we should be using here.

bvaughn, line 43: This change looks like it would make both background and foreground colors white? That doesn't seem right.

Probably `--theme-selection-background` and `--theme-selection-color` are the variables we should be using here.

```suggestion
  --status-background-color: var(--theme-selection-background);
  --status-color: var(--theme-selection-color);
```

src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
```diff
@@ -37,10 +37,10 @@
 .Row[data-selected]:focus,
 .Row[data-selected]:hover {
   background: var(--primary-accent-hover);
-  color: var(--color-default);
+  color: #fff;
```
bvaughn, line 40: This should be a CSS variable.

jonbell-lot23:
> Seems like we end up having some form of "CSS variables vs hard-coded colors" discussion on most frontend PRs 😄
> 
> 
> 
> I'm open to compromise on the Passport UI, since that's intentionally designed to have only one color regardless of the theme. I really think we should hold the line on every other area of the app though (and use CSS variables for colors).

Ha, yeah. To me, the "blue background with white foreground" is a big part of the brand: play button, comments, share, other buttons. So when we were ok with the passport I think I thought "ok! Hard-coded white is ok in these cases!"

What I'll do is audit the places we do this, make it a variable, and always use that variable I GUESS 😂

jonbell-lot23:
Ok, I've introduced var(--primary-accent-foreground-text) that is set to white and is meant to be paired like this in the CSS:

background-color: var(--primary-accent);
color: var(--primary-accent-foreground-text);

bvaughn (approved):
I think these changes look okay, but a bunch of tests failed so I want to make sure we didn't regress here somehow. I restarted the CI actions that failed. Let's make sure they pass before we merge?

jonbell-lot23:
> I think these changes look okay, but a bunch of tests failed so I want to make sure we didn't regress here somehow. I restarted the CI actions that failed. Let's make sure they pass before we merge?

Sorry for the simple question, but to me it looks like 11 tests are still failing, and I struggle to figure out which failures are a problem versus which aren't. I am guessing these 11 are still a concern, but not sure what to do to address them.

bvaughn:
I don't know why a bunch of tests are failing on this PR but not on `main`. Is this PR old? (Like did you branch from an old version of `main`?)

I guess it's fine to just merge it since CSS only changes like this should not cause the behavior we're seeing. Worst case we can revert if needed.

jonbell-lot23:
> I don't know why a bunch of tests are failing on this PR but not on `main`. Is this PR old? (Like did you branch from an old version of `main`?)
> 

Pretty sure it was a clean branch from main

PR #9712, opened 2023-09-08
Retry failed snapshot uploads

hbenl:
Remove `process.exit(1)` from `uploadSnapshot()` and throw instead - the caller of `uploadSnapshot()` will retry and `process.exit(1)` itself if the second attempt fails.


bvaughn (approved):
Seems like a reasonable change

PR #9714, opened 2023-09-11
Open the source for the initial pause point

hbenl:
It may have been a conscious decision to not open that source: it can take quite some time to get the top frame for the initial pause point, by that time the user may already be interacting with the recording and it could feel disruptive if the UI then suddenly opens a source. Furthermore, opening the source manually is easy, just click on the top frame in the frames panel.
@jasonLaster what do you think, should we land this change?


jasonLaster:
@hbenl I agree that we should open the source automatically.

PR #9716, opened 2023-09-11
Add loading screen to redux devtools (FE-1633)

Mokshit06:
Related to [FE-1633](https://linear.app/replay/issue/FE-1633/), adds an indeterminate loader to redux devtools.


Mokshit06:
Ah I see this would fail the current redux devtools tests as the loading indicator is different now. I will fix them

PR #9719, opened 2023-09-12
Improve Source preview popup error messages

bvaughn:
When the user hovers over a value in the source editor that can't be inspected, we used to just show a hard-coded message:
Screen Shot 2023-09-12 at 9 45 38 AM

@Andarist shared feedback that it was confusing the Scopes panel showed something different for these same variables:
Screen Shot 2023-09-12 at 9 46 25 AM

@Andarist suggested we show the same thing in both locations, but after thinking about this more– I think it's better if we make the following two changes to the preview popup:

1. Show the specific error reported by the Protocol. (It will have clearer text than just showing "(unavailable)")
2. Style the popup differently when we're displaying a meta error (to differentiate when we're inspecting an `Error` object)

I don't think it makes sense to change what we show in the Scopes panel to match this, since it's more of a passive panel, but this seems subjective and I'm open to discussion.

# Screenshots

| Dark theme | Light theme |
|:---|:---|
| Screen Shot 2023-09-12 at 9 41 51 AM | Screen Shot 2023-09-12 at 9 42 01 AM |
| Screen Shot 2023-09-12 at 9 41 43 AM | Screen Shot 2023-09-12 at 9 42 08 AM |

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/Popup.module.css
+++ b/packages/replay-next/components/Popup.module.css
@@ -5,6 +5,11 @@
   flex-direction: column;
   padding: 0;
 }
+.Popup[data-style="error"] {
+  --theme-popup-background-color: var(--background-color-error);
+  --theme-popup-border-color: var(--border-color-error);
+  --theme-popup-color: var(--color-error);
+}
 
 .UpArrow,
 .DownArrow {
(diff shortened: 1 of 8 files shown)


markerikson (approved):

hbenl (approved):

Andarist:
chef kiss

bvaughn:
🥳

PR #9723, opened 2023-09-12
Add new React+Redux example app for use in E2E tests

markerikson:
This PR:

- Copies over the source code from the "Redux Fundamentals" example app as a new E2E test example
  - Migrates the code from JS to TS
  - Updates it to work with Vite instead of CRA
  - Reworks the API mocking to use `msw` instead of `miragejs`
  - Commits the built assets to `public/test/examples/`
- Adds a Playwright script that automates the desired user interaction sequence with the app for use in an E2E test "golden recording" (click around, add/delete a couple todos, swap through filters, etc)
- Updates `save-examples.ts` and `record-playwright.ts` to use the latest versions of Playwright (so that we have APIs like `page.getByText()` available), while using the Replay browsers for recording
  - Removes what appear to be dead package deps

Next step after this will be actually writing some tests


markerikson:
welp I broke something :)

markerikson:
Fixed the E2E build issue. Can't have plain `playwright` installed, apparently - just need `@playwright/test`.

bvaughn (approved):
Sure 😄 🤷🏼‍♂️ too much to review

PR #9730, opened 2023-09-14
Remove points from the displayed focus window

hbenl:
- the displayed focus window only needs `time`, so I removed `point`
- I also removed `focusWindowBackup` and use the active focus window instead
- instead of setting the displayed focus window and then syncing the active focus window to it you can now set the active focus window directly with `requestFocusWindow()`, you can use it to set the beginning or the end of the window or both and you can specify the `point`s if you have them or leave them out if not

### DIFF
--- a/packages/replay-next/src/contexts/FocusContext.tsx
+++ b/packages/replay-next/src/contexts/FocusContext.tsx
@@ -1,4 +1,8 @@
-import { FocusWindowRequestBias, TimeStampedPointRange } from "@replayio/protocol";
+import {
+  FocusWindowRequestBias,
+  PointRangeFocusRequest,
+  TimeStampedPointRange,
+} from "@replayio/protocol";
 import {
   PropsWithChildren,
   createContext,
@@ -22,10 +26,15 @@ import {
   imperativelyGetClosestPointForTime,
   preCacheExecutionPointForTime,
 } from "../suspense/ExecutionPointsCache";
-import { TimeRange } from "../types";
+import { TimeRange as TimeRangeArray } from "../types";
 import { SessionContext } from "./SessionContext";
 
 const FOCUS_DEBOUNCE_DURATION = 250;
+
+interface TimeRange {
+  begin: number;
+  end: number;
+}
 
 export type UpdateOptions = {
   bias?: FocusWindowRequestBias;
@@ -41,16 +50,19 @@ export type FocusContextType = {
   // The currently active range - use this for backend requests that don't cause components to suspend
   range: TimeStampedPointRange | null;
   // The range to be displayed in the Timeline - don't use this for backend requests
-  rangeForDisplay: TimeStampedPointRange | null;
+  rangeForDisplay: TimeRange | null;
   // The deferred value of the currently active range - use this for backend requests that may cause components to suspend
   rangeForSuspense: TimeStampedPointRange | null;
 
-  // Set focus window to a range of execution points (or null to clear).
-  update: (value: TimeStampedPointRange | null, options: UpdateOptions) => Promise<void>;
+  // Set focus window to a range of execution points.
+  update: (value: TimeStampedPointRange, options: UpdateOptions) => Promise<void>;
 
-  // Set focus window to a range of times (or null to clear).
+  // Set focus window to a range of times.
   // Note this value is imprecise and should only be used by the Timeline focuser UI.
-  updateForTimelineImprecise: (value: TimeRange | null, options: UpdateOptions) => Promise<void>;
+  updateForTimelineImprecise: (
+    value: TimeRangeArray | null,
+    options: UpdateOptions
+  ) => Promise<void>;
 };
 
 export const FocusContext = createContext<FocusContextType>(null as any);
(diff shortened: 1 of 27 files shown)


src/ui/utils/timeline.ts
```diff
@@ -173,50 +86,14 @@ export function getSecondsFromFormattedTime(formatted: string) {
...
-  );
-}
-
-export function isTimeInRegions(time: number, regions?: TimeStampedPointRange[]): boolean {
-  return !!regions?.some(region => time >= region.begin.time && time <= region.end.time);
-}
-
-export function rangeForFocusWindow(focusWindow: FocusWindow): TimeStampedPointRange {
+export function rangeForFocusWindow(focusWindow: TimeStampedPointRange): TimeStampedPointRange {
   return { begin: focusWindow.begin || { time: 0, point: "0" }, end: focusWindow.end };
```
bvaughn, line 94: This method looks funky:
```js
export function rangeForFocusWindow(focusWindow: TimeStampedPointRange): TimeStampedPointRange {
  return { begin: focusWindow.begin || { time: 0, point: "0" }, end: focusWindow.end };
}
```

Either the types are wrong (I don't think they are) or it doing...nothing.

I think it was unnecessary before but maybe _looked_ necessary because it accepted a `FocusWindow` and returned a `TimeStampedPointRange` (even though those types were the same).

Kill it?

src/ui/components/Timeline/FocusModePopout.tsx
```diff
@@ -43,8 +47,11 @@ export default function FocusModePopout({
       updateFocusWindowThrottled.cancel();
     }
 
-    await dispatch(setFocusWindow(focusWindowBackup));
-    await dispatch(syncFocusedRegion());
+    const currentFocusWindow = replayClient.getCurrentFocusWindow();
+    assert(currentFocusWindow);
+    dispatch(
+      setFocusWindow({ begin: currentFocusWindow.begin.time, end: currentFocusWindow.end.time })
+    );
```
bvaughn, line 54: I like this change.

src/ui/components/FocusContextReduxAdapter.tsx
```diff
@@ -1,37 +1,33 @@
...
   const isPending = rangeForSuspense !== range;
 
   const update = useCallback(
-    async (value: TimeStampedPointRange | null, options: UpdateOptions) => {
+    async (value: TimeStampedPointRange, options: UpdateOptions) => {
       const { sync } = options;
-
-      await dispatch(setDisplayedRange(value));
-
       if (sync) {
```
bvaughn, line 25: The name of this param (`sync`) always confuses me. I know you didn't add it with this PR but since we're in here cleaning things up– I wonder if there's some other, more intuitive name we could think of? Even if we just made it longer...like "sync remote" or something.

Just thinking out loud. No action required.

packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -22,11 +26,16 @@ import {
...
-import { TimeRange } from "../types";
+import { TimeRange as TimeRangeArray } from "../types";
 import { SessionContext } from "./SessionContext";
 
 const FOCUS_DEBOUNCE_DURATION = 250;
 
+interface TimeRange {
+  begin: number;
+  end: number;
+}
```
bvaughn, line 37: We already have a type defined for this in a more shared location. Can we use it instead?

https://github.com/replayio/devtools/blob/95c9148aa9a1676a0bbee2fc0327167fcad43ef4/packages/replay-next/src/types.ts#L4

We also have some redundant ones that would be nice to just delete:

https://github.com/replayio/devtools/blob/95c9148aa9a1676a0bbee2fc0327167fcad43ef4/src/ui/utils/app.ts#L2-L5

https://github.com/replayio/devtools/blob/95c9148aa9a1676a0bbee2fc0327167fcad43ef4/src/ui/state/timeline.ts#L3-L6

packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -88,22 +98,18 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
   // to indicate that what's currently being showed is stale.
   const [isTransitionPending, startTransition] = useTransition();
 
-  // Refine the loaded ranges based on the focus window.
   useEffect(() => {
-    if (rangeForDisplay === null || rangeForDisplay === initialRange) {
+    if (!requestedRange) {
```
bvaughn, line 102: So this change means that we'll always call `client.requestFocusWindow` when this context mounts. I assume that's an intentional part of the changes here just wanted to verify.

packages/replay-next/src/contexts/FocusContext.tsx
```diff
@@ -129,7 +135,7 @@ export function FocusContextRoot({ children }: PropsWithChildren<{}>) {
   }, [loadedRegions]);
 
   const updateFocusRange = useCallback(
-    async (range: TimeStampedPointRange | null, options: UpdateOptions) => {
+    async (range: TimeStampedPointRange, options: UpdateOptions) => {
```
bvaughn, line 138: nit: Changing the `range` param (and the `range` state) to not be nullable means that the conditional not-null checks on line 144 and line 146can be removed.

bvaughn (commented):
Love the simplifying changes in this PR. Couple of changes requested but nothing major.

packages/replay-next/src/contexts/FocusContext.tsx
hbenl: > So this change means that we'll always call client.requestFocusWindow when this context mounts.

I don't think it does because `requestedRange` is initially `undefined`.

packages/replay-next/src/contexts/FocusContext.tsx
bvaughn: > I prefer the object type over the array type

Yeah that makes sense. I would like to just get rid of the array version, if we could.

> And I don't want to import a type from src/ui into replay-next.

Yeah totally, we shouldn't do that.

PR #9731, opened 2023-09-14
Migrate to a fully-updated fork of `react-devtools-inline` with our own customizations

markerikson:
This PR:

- ~~Initially tries updating `react-devtools-inline` from `4.24.7` to `4.28.0`~~

Last time I tried this locally, all of the colors disappeared from the component tree, but this time it seemed to work.

~~If the PR build works okay, then I'm going to take a stab at creating a fork of this and applying the same changes we have in the Yarn patch right now (which will also accomplish FE-1890).~~

Looks good on all accounts!

- The initial test update to `react-devtools-inline@4.28.0` worked okay
- I forked `react-devtools-inline` and republished as `@replayio/react-devtools-inline` (changes here: https://github.com/replayio/react/pull/7 )
- I updated us to use that fork, and replaced all references
- Nuked the ancient pre-RDT-protocol-v2 handling we had
- Deleted the now-dead Yarn patch file


markerikson:
Holy smoke Socket how did you _just now_ detect those packages that have been there for months?

src/ui/components/SecondaryToolbox/react-devtools/__tests__/utils.ts
```diff
@@ -1,5 +1,5 @@
+import type { Store } from "@replayio/react-devtools-inline";
 import { createElement } from "react";
-import type { Store } from "react-devtools-inline";
```
bvaughn, line 2: We could add an entry to the `no-restricted-imports` rule for `"react-devtools-inline"` but I guess it's probably not necessary

bvaughn (approved):

PR #9732, opened 2023-09-14
Change to execution-point based comparison for Redux Devtools (FE-1811)

Mokshit06:
Fixes [FE-1811](https://linear.app/replay/issue/FE-1811)


markerikson (approved):

markerikson:
Hmm. The Redux DevTools test errored, and it does look like it's directly related to this change:

Can you check further?

https://app.replay.io/recording/reduxdevtools-test-redux-devtools--4b039ba0-405d-42c6-832f-303aab3f8a01

I think we need to handle not having a valid `annotation`.

bvaughn:
This commit seems to have broken the jump-to-code e2e test.

PR #9733, opened 2023-09-14
Add ErrorBoundary around Rules panel

bvaughn:
### [Loom overview](https://www.loom.com/share/Elements-greater-Rules-panel-error-boundary-26fa80128bfd4bc6b7b64b35a288f85f)

While I was in there I also removed the unused `react-error-boundary` dependency.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9620,7 +9620,6 @@ __metadata:
     cli-spinners: ^2.7.0
     cypress: ^12.5.1
     log-update: ^4
-    playwright: ^1.37.1
     strip-ansi: ^6.0.0
     ts-node: ^10.7.0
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -14038,17 +14037,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"playwright@npm:^1.37.1":
```
bvaughn, line 14041: Did someone not run `yarn install` after updating `package.json` or is Yarn just being clowny?

markerikson: That one's probably my fault from yesterday, sorry!

bvaughn: No worries!

PR #9736, opened 2023-09-17
Add restart debugging session

jasonLaster:
fixes FE-1848

This is a small feature that can be useful when a backend debuggions session state is corrupted. We should likely put it in the troubleshooting docs when it lands as well.

### DIFF
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -5,7 +5,13 @@ import dotenv from "dotenv";
 import { RecordingTarget } from "replay-next/src/suspense/BuildIdCache";
 
 import exampleRecordings from "../examples.json";
-import { debugPrint, getElementClasses, waitFor, waitForRecordingToFinishIndexing } from "./utils";
+import {
+  debugPrint,
+  getCommandKey,
+  getElementClasses,
+  waitFor,
+  waitForRecordingToFinishIndexing,
+} from "./utils";
 
 dotenv.config({ path: "../../.env" });
 
@@ -75,6 +81,12 @@ export async function startLibraryTest(page: Page, apiKey: string, teamId: strin
 
   await page.locator('[data-test-id="TestRunResults"]').waitFor();
   await page.locator('[data-test-id="TestRunList"]').waitFor();
+}
+
+export async function commandPalette(page: Page, query: string) {
+  await page.keyboard.press(`${getCommandKey()}+K`);
+  await page.keyboard.type(query);
+  await page.keyboard.press("Enter");
 }
 
 export type TestRecordingKey = keyof typeof exampleRecordings;
--- a/packages/e2e-tests/package.json
+++ b/packages/e2e-tests/package.json
@@ -11,6 +11,7 @@
     "test:chromium": "playwright test --project chromium --reporter=@replayio/playwright/reporter,line",
     "test:replay-chromium": "playwright test --project replay-chromium --reporter=@replayio/playwright/reporter,line",
     "test:debug": "playwright test --project chromium --workers 1 --headed",
+    "test:ui": "playwright test --ui",
     "test:install": "playwright install",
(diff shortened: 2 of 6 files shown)


packages/e2e-tests/tests/restart-session.test.ts
```diff
@@ -0,0 +1,17 @@
...
+test(`restart debugging session`, async ({
+    pageWithMeta: { page, recordingId },
+    exampleKey,
+}) => {
+    await startTest(page, recordingId);
+    await openDevToolsTab(page);
+    await commandPalette(page, "restart session");
+    await page.waitForNavigation()
+    await waitForRecordingToFinishIndexing(page);
+});
```
jasonLaster, line 17: The test took longer to write than the feature, but i think it's important to try and write tests for new features. Especially infrequently used features like this.

packages/e2e-tests/package.json
```diff
@@ -30,4 +31,4 @@
     "chalk": "^4",
     "strip-ansi": "^6.0.0"
   }
-}
+}
```
jasonLaster, line 34: how do i remove this again?

bvaughn:
Looks like you didn't assign reviewers but I assume any of us can do this so I'll take a look.

packages/e2e-tests/tests/restart-session.test.ts
bvaughn: Nice!

src/ui/actions/session.ts
```diff
@@ -205,7 +205,15 @@ export function createSocket(
         ),
         disableProtocolQueryCache: userData.get("backend_disableProtocolQueryCache"),
       };
-      if (userData.get("backend_newControllerOnRefresh")) {
+
+      const restartParam = new URL(window.location.href).searchParams.get("restart") || undefined;
```
bvaughn, line 209: Nit: `URLSearchParams.get()` returns `null` for values not in the URL, so `|| undefined` isn't really doing anything meaningful.

```suggestion
      const restartParam = new URL(window.location.href).searchParams.get("restart");
```

src/ui/components/CommandPalette/CommandPalette.tsx
```diff
@@ -47,7 +47,8 @@ export type CommandKey =
   | "jump_to_next_pause"
   | "jump_to_previous_pause"
   | "toggle_protocol_panel"
-  | "toggle_protocol_timeline";
+  | "toggle_protocol_timeline"
+  | "restart_session";
```
bvaughn, line 51: Nit: I think we usually try to alpha sort lists like this (though the "jump_to" keys already didn't do this so 🤷🏼 )

packages/e2e-tests/helpers/index.ts
```diff
@@ -77,6 +83,12 @@ export async function startLibraryTest(page: Page, apiKey: string, teamId: strin
   await page.locator('[data-test-id="TestRunList"]').waitFor();
 }
 
+export async function commandPalette(page: Page, query: string) {
+  await page.keyboard.press(`${getCommandKey()}+K`);
+  await page.keyboard.type(query);
```
bvaughn, line 88: I'm surprised we don't need a wait-for in between these two (for the type ahead to become visible). I assumed that's what you added the new `command-palette` test id for, e.g.

```js
await page.waitForSelector('[data-test-id="command-palette"]`);
```

bvaughn (approved):
Nice!

Love the new test and helper.

packages/e2e-tests/helpers/index.ts
jasonLaster: Same. I like that Playwright is flexible though

bvaughn: I guess we'll see if this ends up being a source of flakiness. (Maybe the default interaction time is just enough for React to render the type ahead)

PR #9737, opened 2023-09-18
Log Point panel properly evaluates inline objects

bvaughn:


jasonLaster (approved):
Likes good. I like the tests

PR #9741, opened 2023-09-19
Use Session.getPointStack to get execution points for frames

hbenl:
Currently we use frame steps to find execution points for parent frames but we may not be able to fetch the frame steps (if there are too many for the given frame) and our logic for selecting a frame step doesn't always select the correct one.


packages/replay-next/src/suspense/FrameStepsCache.ts
```diff
@@ -30,20 +30,3 @@ export const frameStepsCache: Cache<
     }
   },
 });
-
-export async function getFrameStepForFrameLocation(
```
bvaughn, line 34: Love seeing us remove these cache-like-methods

packages/replay-next/src/suspense/PointStackCache.ts
```diff
@@ -0,0 +1,37 @@
...
+      updateMappedLocation(sources, frame.functionLocation);
+      if (frame.point.frame) {
+        updateMappedLocation(sources, frame.point.frame);
+      }
+      return { ...frame, index: index + start };
+    });
+  },
+});
+
+export async function getPointDescriptionForFrame(
```
bvaughn, line 30: I don't love these meta cache methods because we end up being able to use them with other utilities exported by the "suspense" package (the hooks and stuff). I guess in this case it doesn't matter much since the only things using this cache are other async functions, never React components.

packages/replay-next/src/suspense/PointStackCache.ts
```diff
@@ -0,0 +1,37 @@
...
+  debugLabel: "pointStackCache",
+  getPointForValue: pointStackFrame => pointStackFrame.index,
+  getKey: (client, point) => point,
+  load: async (start, end, client, point) => {
+    const pointStack = await client.getPointStack(point, end + 1);
+    const sources = await sourcesByIdCache.readAsync(client);
+    return pointStack.slice(start).map((frame, index) => {
+      updateMappedLocation(sources, frame.functionLocation);
+      if (frame.point.frame) {
+        updateMappedLocation(sources, frame.point.frame);
```
bvaughn, line 23: I don't think I understand the relationship between these two. The Protocol docs say:

> `PointStackFrame`:
> * `functionLocation` – The location of the function containing the point
> * `point.frame` – Location of the topmost frame, omitted if there are no frames on stack

packages/replay-next/src/suspense/PointStackCache.ts
```diff
@@ -0,0 +1,37 @@
+import { ExecutionPoint, PointStackFrame } from "@replayio/protocol";
+import { IntervalCache, createIntervalCache } from "suspense";
+
+import { ReplayClientInterface } from "shared/client/types";
+
+import { updateMappedLocation } from "./PauseCache";
+import { sourcesByIdCache } from "./SourcesCache";
+
+export const pointStackCache: IntervalCache<
```
bvaughn, line 9: The choice to use an interval cache for this was surprising to me at first.

I guess we're always fetching frames 0...N, so the interval cache prevents us from re-fetching the lower frames if we need to step in and fetch the next one(s)?

bvaughn (approved):

PR #9742, opened 2023-09-19
Fix test setup for timeline reducer tests

hbenl:


bvaughn (approved):
Oh so we just had to uncomment it? Haha. Nice

PR #9750, opened 2023-09-20
Fix J2C for Cypress test steps

hbenl:
We're looking for a J2C annotation between the step's `step:enqueue` and `step:end` annotations but `step:enqueue` annotations can come before the `step:start`/`step:end` of subsequent steps, so we may be using the J2C annotation for a different step.
With this PR, we're looking for a J2C annotation between the step's `step:start` and `step:end` annotations.

### DIFF
--- a/packages/e2e-tests/tests/regression-fe-1875-cypress-jump-to-code.test.ts
+++ b/packages/e2e-tests/tests/regression-fe-1875-cypress-jump-to-code.test.ts
@@ -26,10 +26,23 @@ test("fe-1875 :: verify that steps go to the right point in time", async ({
 
   await clickableSteps.nth(0).click();
 
-  const timelinePercent = await waitForTimelineAdvanced(page, 0);
+  let timelinePercent = await waitForTimelineAdvanced(page, 0);
   expect(Math.round(timelinePercent)).toBe(67);
 
   page.click('[data-test-name="JumpToCode"]');
 
   await waitForPaused(page, 658);
+
+  await openCypressTestPanel(page);
+  await clickableSteps.nth(2).click();
+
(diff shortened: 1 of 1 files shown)


bvaughn:
> With this PR, we're looking for a J2C annotation between the step's step:start and step:end annotations.

I don't think that is a safe change. The order of enqueue and start are reversed for "chai assertions"

cc @ryanjduffy

hbenl:
> The order of enqueue and start are reversed for "chai assertions"

But this PR only applies to click and type events, so I guess it should be correct.

bvaughn:
> > The order of enqueue and start are reversed for "chai assertions"
>
> But this PR only applies to click and type events, so I guess it should be correct.

Hmm. It feels kind of fragile to rely on that, because the connection is pretty indirect and not obvious.

jasonLaster (commented):
Can we tighten this test by trying to jump to both the click and type commands? I remember one was failing before, but i accidentally tested the one that worked :)

https://github.com/replayio/devtools/blob/main/packages/e2e-tests/tests/regression-fe-1875-cypress-jump-to-code.test.ts

bvaughn:
Hm. I don't think this was ready to be merged, was it? I thoughts we were still talking about it.

bvaughn:
(Just to be clear, I'm not saying we need to revert this or anything. It's fine. Just talking about broadly.)

jasonLaster:
I asked that we merge the PR so that the demo would work for our 3 prospective customer calls today.

Lets make sure that the Linear issue is not closed until we align on (https://github.com/replayio/devtools/pull/9750#issuecomment-1727840042) though because I think that we likely can improve this further. It's unreasonable for the UI component to know which data source is more accurate most of the time.

PR #9752, opened 2023-09-21
Test playwright plugin with annotations

ryanjduffy:
* Updates `@replayio/playwright` plugin to 2.0.0-alpha.2 to adopt the new fixture integration
* Adds support for using `replay-playwright` annotations to navigate to steps within a test
* Adds support for `stacks` from the `x-replay-playwright` metadata block to show the test-side code triggering a step

To Do:
- [x] Add e2e tests for playwright


ryanjduffy:
cc: @hbenl, @markerikson, @bvaughn - this isn't quite ready to land but letting you know it's around and hopefully getting close

bvaughn:
Just ping us again by name when this is ready to review :)

ryanjduffy:
@hbenl + @bvaughn - I think this is ready for a review.

@hbenl - I've retained most of what you already had in place so hopefully there isn't too much surprising here.

packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -232,7 +232,7 @@ export namespace RecordingTestMetadataV3 {
   export type UserActionEventStack = Array<{
     columnNumber: number;
     fileName: string;
-    functionName: string;
+    functionName?: string;
```
bvaughn, line 235: Why?

packages/shared/graphql/types.ts
```diff
@@ -230,12 +230,22 @@ export type PlaywrightTestSources = {
   [fileName: string]: string;
 };
 
+export type PlaywrightTestStacks = {
+  [fileName: string]: {
+    file: string;
+    line: number;
+    column: number;
+    function?: string;
```
bvaughn, line 238: "function" is a reserved keyword. Can we use `functionName` (also this would be more consistent with other related types)

bvaughn:
TBH I'm super

packages/shared/graphql/types.ts
ryanjduffy: Sadly, this is coming from Playwright as `function` but we can adapt it in the plugin

packages/shared/test-suites/RecordingTestMetadata.ts
ryanjduffy: The playwright type has `function` as `string | undefined` so I've propagated that.

ryanjduffy:
@hbenl - nudge for a review on this please

jasonLaster:
Tire kicking went well. Great job!

ryanjduffy:
@hbenl - Can you take a look at this please? Thanks!

packages/shared/graphql/types.ts
```diff
@@ -231,12 +231,22 @@ export type PlaywrightTestSources = {
   [fileName: string]: string;
 };
 
+export type PlaywrightTestStacks = {
+  [fileName: string]: {
```
hbenl, line 235: This should be `[id: string]`.

packages/e2e-tests/tests/playwright-01_basic-panel.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    await expect(rows).toHaveCount(1);
+  });
+
+  const firstTest = rows.first();
+
+  // displays the nav chevron on hover
+  const chevron = getTestRowChevron(firstTest);
+  await chevron.isHidden();
+  await firstTest.hover();
+  await chevron.isVisible();
```
hbenl, line 60: I guess you're trying to wait until `chevron` is visible but that's not what this line does. You need to use `await expect(chevron).toBeVisible();`. The same goes for `await chevron.isHidden();` at line 58.

packages/e2e-tests/tests/playwright-03_test-step-interactions.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const firstClick = clickableSteps.filter({ hasText: /click/ }).first();
+
+  // Jump to the first `get` step
+  await firstClick.click();
+  prevPercent = await waitForTimelineAdvanced(page, prevPercent);
+
+  // Should show the "Before/After" buttons
+  const beforeAfterButtons = getTestStepBeforeAfterButtons(page);
+  beforeAfterButtons.waitFor({ state: "visible", timeout: 1000 });
```
hbenl, line 76: This needs an `await`.

hbenl (approved):

packages/e2e-tests/tests/playwright-01_basic-panel.test.ts
ryanjduffy: :+1: I'll point out that this test was lifted directly from the cypress tests so those must be wrong too.

PR #9757, opened 2023-09-22
Hide the leave menu if other admins are pending

ryanjduffy:
The backend prevents leaving a team if you are an admin and there are no other non-pending admins but the frontend still shows the Leave menu option.

This adds a check for pending as well when deciding to show the Leave option.

### DIFF
--- a/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
+++ b/src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
@@ -37,7 +37,8 @@ export function WorkspaceMembers({
   }, [members]);
 
   const canLeave = members.length > 1;
-  const canAdminLeave = canLeave && members.filter(a => a.roles?.includes("admin")).length > 1;
+  const canAdminLeave =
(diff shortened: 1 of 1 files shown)


src/ui/components/shared/WorkspaceSettingsModal/WorkspaceSettingsModal.tsx
```diff
@@ -37,7 +37,8 @@ export function WorkspaceMembers({
   }, [members]);
 
   const canLeave = members.length > 1;
-  const canAdminLeave = canLeave && members.filter(a => a.roles?.includes("admin")).length > 1;
+  const canAdminLeave =
+    canLeave && members.filter(a => a.roles?.includes("admin") && !a.pending).length > 1;
```
bvaughn, line 41: nit: Multiple "can leave" variables feels a bit awkward

```js
const isAdmin = member.roles?.includes("admin");
const canLeave = isAdmin
  ? members.filter(
      (member) => member.roles?.includes("admin") && !member.pending,
    ).length > 1
  : members.length > 1;
```

bvaughn (approved):

PR #9758, opened 2023-09-22
Restrict moving replays to appropriate workspace type

ryanjduffy:
## Issue

The bulk edit and per-recording menu in the library allows users to move recordings into unsupported workspaces.

## Resolution

* Restrict the workspace list in the menu to the appropriate type
* Suppress the move option altogether if a user manages to select both types which shouldn't happen because we've prevented this but they might have already and we shouldn't perpetuate it

### DIFF
--- a/src/ui/components/Library/Team/View/Recordings/Header/BatchActionDropdown.tsx
+++ b/src/ui/components/Library/Team/View/Recordings/Header/BatchActionDropdown.tsx
@@ -2,6 +2,7 @@ import { RecordingId } from "@replayio/protocol";
 import classNames from "classnames";
 import React, { useState } from "react";
 
+import { assert } from "protocol/utils";
 import { Recording } from "shared/graphql/types";
 import { useGetTeamIdFromRoute } from "ui/components/Library/Team/utils";
 import { isTestSuiteReplay } from "ui/components/TestSuite/utils/isTestSuiteReplay";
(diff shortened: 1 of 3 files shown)


src/ui/components/Library/Team/View/Recordings/Header/BatchActionDropdown.tsx
```diff
@@ -114,11 +115,16 @@ export default function BatchActionDropdown({
...
   // all the selected recordings.
-  const enableLibrary = selectedIds
-    .map(id => recordings.find(r => r.id === id))
-    .every(recording => userId === recording?.user?.id);
+  const enableLibrary = selectedRecordings.every(recording => userId === recording?.user?.id);
+  const testSuiteReplayFlags = selectedRecordings
+    .filter(r => r != null)
+    .map(r => isTestSuiteReplay(r!));
+  const allTestSuiteReplays = testSuiteReplayFlags.every(isTest => isTest === true);
+  const enableMove = allTestSuiteReplays || testSuiteReplayFlags.every(isTest => isTest === false);
```
bvaughn, line 127: nit: I think we can reduce the amount of looping/mapping and make this more efficient (but feel free to ignore)

```js
let allRecordingsOwnedByCurrentUser = true;
let allRecordingsTestSuites = true;
let noRecordingsTestSuites = true;

selectedIds.forEach((id) => {
  const recording = recordings.find((recording) => recording.id === id);
  assert(recording); // <- This seems worth doing regardless

  if (userId !== recording.user?.id) {
    allRecordingsOwnedByCurrentUser = false;
  }

  if (isTestSuiteReplay(recording)) {
    noRecordingsTestSuites = false;
  } else {
    allRecordingsTestSuites = false;
  }
});

const enableLibrary = allRecordingsOwnedByCurrentUser;
const enableMove = allRecordingsTestSuites || noRecordingsTestSuites;
```

bvaughn (approved):

PR #9762, opened 2023-09-26
Start removing sources from redux

hbenl:
This PR switches most places that were getting sources from redux to the suspense caches. Most of the remaining places are in `createSelector()` and `connect()` calls, where the switch is slightly more involved.


bvaughn (approved):
Just skimmed.

Provided all of our e2e tests pass, seems ok

PR #9775, opened 2023-09-28
Rewrite "React Queued Renders" panel for better perf, and make clear it's experimental

markerikson:
This PR:

- supersedes the original work in #9410
  - splits the changes in that PR into two halves.  The `<ReactPanel>` changes are in this PR. The `<ReactReduxPerfPanel>` changes will go in a follow-up stacked PR.
- Scaffolds a new `packages/replay-experimental` folder by configuring TS and import sorting
- Rewrites the existing `<ReactPanel>` to improve loading perf and rendering behavior:
  - Rewrites the API calls to use `getPointStack` instead of `getAllFrames` for faster stack trace data
  - Adds logic to format point stack frames with source information
  - Adds Suspense interval caches for hits to React internal methods like `scheduleUpdateOnFiber`
  - Uses a virtualized list to render the list items
- Improves the existing logic for formatting function details:
  - Fully switches to using the `sourcesById` cache instead of Redux sources state
  - Handles formatting anonymous functions by looking at the parent function name if available (common with Redux thunks)
  - Copies support for formatting methods in classes such as `render()` from the backend routines logic
- Moves both the `<ReactPanel>` component and all associated Suspense caches over to `replay-experimental`, to clarify that this logic is not production-ready yet

image

### Future Work

- Counting how many function components and class components rendered in a given render pass
- Showing list items for React render commits in the React panel (ie, alternating calls to queue a render vs commits)
- Showing the red "you are here" line in the list


packages/replay-next/src/suspense/SourcesCache.ts
```diff
@@ -187,6 +188,18 @@ export const sourcesByIdCache = createSingleEntryCache<
   },
 });
 
+export const sourcesByPartialUrlCache: Cache<
```
bvaughn, line 191: Kind of a nit: This cache feels a bit unnecessary. Seems like we could `useMemo` or in some other way cache the result of this filter action without adding a whole new cache for it. We're really only using it to look for a single file anyway (react-dom).

src/ui/actions/eventListeners/eventListenerUtils.ts
```diff
@@ -91,24 +99,31 @@ export type FormattedEventListener = EventListenerWithFunctionInfo & {
...
-  sourcesState: SourcesState,
+  locationInFunction: Location | MappedLocation,
   framework?: string
 ): Promise<FormattedEventListener | undefined> => {
-  const { functionLocation } = fnPreview;
-  const sources = await sourcesByIdCache.readAsync(replayClient);
-  updateMappedLocation(sources, functionLocation);
-
-  const location = getPreferredLocation(sourcesState, functionLocation);
+  // const { functionLocation } = fnPreview;
```
bvaughn, line 112:
```suggestion
```

packages/replay-experimental/src/components/ReactPanel.tsx
```diff
@@ -0,0 +1,218 @@
...
+import { UIThunkAction } from "ui/actions";
+import { seek } from "ui/actions/timeline";
+import { JumpToCodeButton, JumpToCodeStatus } from "ui/components/shared/JumpToCodeButton";
+import MaterialIcon from "ui/components/shared/MaterialIcon";
+import { getSourceIdsByUrl } from "ui/reducers/sources";
+import { getCurrentTime } from "ui/reducers/timeline";
+import { useAppDispatch, useAppSelector } from "ui/setup/hooks";
+
+import cardsListStyles from "ui/components/Comments/CommentCardsList.module.css";
+import styles from "ui/components/Events/Event.module.css";
```
bvaughn, line 27: Seems pretty fragile for this component to import styles from other packages. We don't have TypeScript checking us there so it seems we're likely to "break" the React styles by changes made in these other CSS files.

src/ui/components/ReactReduxPanels.module.css
```diff
@@ -0,0 +1,42 @@
...
+  gap: 0.75rem;
+}
+
+.ListContainer {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+  overflow: auto;
+}
```
bvaughn, line 42: Can we delete this file? Seems like nothing is referencing it.

src/ui/suspense/frameCache.ts
```diff
@@ -60,3 +81,99 @@ function createPauseFrames(pauseId: PauseId, frames: Frame[], sourcesState: Sour
...
+export const formattedPointStackCache: Cache<
+  [replayClient: ReplayClientInterface, point: TimeStampedPoint],
+  FormattedPointStack
+> = createCache({
+  debugLabel: "FormattedPointStack",
+  getKey: ([replayClient, point]) => point.point,
+  async load([replayClient, point]) {
+    return formatPointStackForPoint(replayClient, point);
+  },
+});
```
bvaughn, line 130: Nothing referencing this cache?

```suggestion
```

packages/replay-experimental/src/components/ReactPanel.tsx
markerikson: Yep. This was me trying to get something that looked passably decent, as quickly as possible (and an example of why I was saying we could start extracting some reusable components or styles).

src/ui/components/ReactReduxPanels.module.css
markerikson: It gets used in the follow-on PR - got accidentally included here as I was trying to split up the code.

bvaughn: Kill it from this PR?

packages/replay-experimental/src/components/ReactPanel.tsx
bvaughn: I think we should copy/fork the styles for now.

src/ui/suspense/frameCache.ts
markerikson: I added the cache after writing the standalone function.  The other logic should really be using the cache for consistency - I'll update that.

packages/replay-next/src/suspense/SourcesCache.ts
markerikson: It's being used for `react-dom`, but the call is inside of a couple different caches, and I've got a spot in the follow-up PR that I just noticed ought to be using this for Redux files as well.

src/ui/suspense/frameCache.ts
```diff
@@ -60,3 +82,108 @@ function createPauseFrames(pauseId: PauseId, frames: Frame[], sourcesState: Sour
...
+
+export async function formatPointStackForPoint(
```
bvaughn, line 140: Exporting this function separately from the cache seems like a foot gun.

You can always call `cache.readAsync` to get the same behavior of the async function, but exporting the function directly seems to risk computing things that don't get cached.

Is there a reason you're exporting them separately?

It looks like the only thing that imports this function is `reactInternalsCaches.ts` and it doesn't use the import.

packages/replay-experimental/src/suspense/reactInternalsCaches.ts
```diff
@@ -0,0 +1,331 @@
...
+  | ReactSyncUpdatedStarted
+  | ReactRenderCommitted;
+
+export const reactRendersIntervalCache = createFocusIntervalCacheForExecutionPoints<
+  [replayClient: ReplayClientInterface],
+  ReactRenderEntry
+>({
+  debugLabel: "ReactRenders",
+  getPointForValue: data => data.point.point,
+  async load(rangeStart, rangeEnd, replayClient) {
```
bvaughn, line 255: This function has a lot of serial async operations. Looks like the first one is blocking, but the last few could be run in parallel.

src/ui/suspense/frameCache.ts
```diff
@@ -60,3 +82,108 @@ function createPauseFrames(pauseId: PauseId, frames: Frame[], sourcesState: Sour
...
+export interface FormattedPointStack {
+  point: PointDescription;
+  resultPoint: PointDescription;
+  frame?: FormattedPointStackFrame;
+  allFrames: FormattedPointStackFrame[];
+  filteredFrames: FormattedPointStackFrame[];
+  functionName?: string;
+}
+
+export async function formatPointStackFrame(
```
bvaughn, line 104: Same question about this function and only being used in one place. Can we either inline it into the cache or at least not export it?

src/ui/suspense/frameCache.ts
markerikson: In both these cases I wrote the standalone function first, then wrote the cache later, so the code had been left that way.  I can inline these.

PR #9776, opened 2023-09-28
Add a Redux dispatch perf analysis panel

markerikson:
This PR:

- Is stacked on top of #9775 , and also supersedes #9410
- Adds a new "Redux Dispatch Perf" experimental panel and associated logic
  - Adds caches that find the location of the Redux `dispatch` method and calls to `dispatch`.  (This heavily overlaps with the existing Redux DevTools panel, but works even if there are no annotations or Redux DevTools integration.  It _does_ rely on sourcemaps having a `node_modules/redux/redux.js`-type file available)
  - Adds caches that fetch hits for specific lines of code within the Redux dispatch method, and uses those to calculate durations for the reducer execution and subscriber notifications
  - Renders a virtual list of Redux dispatch entries
  - Allows selecting a dispatch entry and showing further details in a bottom panel
- The details panel shows:
  - timing breakdowns for the dispatch
  - estimates for how many components we think were queued to re-render by the dispatch, based on calls to `scheduleUpdateOnFiber` during the subscriber notifications
  - Tries to find a React render that occurred right after the dispatch, and if found, estimates its duration

image

### Future Work

- Finding all Redux selectors used in an app
- Counting how many selectors got called during a set of subscriber notifications, and calculating their durations
- Showing the red "you are here" line in the list

### DIFF
--- a/packages/replay-experimental/src/components/ReactReduxPanels.module.css
+++ b/packages/replay-experimental/src/components/ReactReduxPanels.module.css
@@ -47,3 +47,46 @@
 .eventRow:hover {
   background-color: var(--event-row-hover-background-color);
 }
+
+.Body {
+  flex: 1 1 auto;
+}
+
+.ResizeHandle {
+  height: 0.75rem;
+}
+.ResizeHandleBar {
+  width: 100%;
+  height: 1px;
+  background-color: var(--theme-border);
+}
+
+.DispatchDetailsPanel {
+  min-height: 1.75rem;
+}
+
+.DispatchDetailsHeader {
+  font-size: var(--font-size-large);
+  display: flex;
+  align-items: center;
+  gap: 1ch;
+  cursor: pointer;
+  padding: 0 0.75rem;
+}
+
+.DispatchDetailsContainer {
+  height: 100%;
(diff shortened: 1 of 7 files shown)


bvaughn:
My personal vote is that we should not land this feature.
1. This list is very close to the existing Redux panel to warrant a new, separate section.
2. I don't think the information in the details pane is very useful in its current format. (I know you say it's useful for you Mark, but I don't think it's useful to a wide enough % of our users– and I think the bar for shipping something should be that it's useful for _many/most_ Replay users.)

markerikson:
FWIW I'm not tied to the specific UI location or implementation - most of this was just copy-pasting the setup and layout logic from the "React Render" panel because they're both experimental and it was the easiest way to get something on screen.

I still absolutely think there _is_ value in this info, but I can understand the argument that not enough people would care about it for now.

markerikson:
Okay, I've rebased this on top of `main` as of landing #9775 . The preview branch is here:

https://devtools-git-feature-fe-1631-redux-perf-panel-recordreplay.vercel.app/

I'll go ahead and close this.

PR #9779, opened 2023-09-29
Fix filtering network events for tests

hbenl:
The `findSliceIndices()` util didn't work correctly so I replaced it with our `binarySearch()` util.

### DIFF
--- a/packages/shared/test-suites/RecordingTestMetadata.ts
+++ b/packages/shared/test-suites/RecordingTestMetadata.ts
@@ -7,8 +7,9 @@ import {
 import { satisfies } from "compare-versions";
 
 import { comparePoints } from "protocol/execution-point-utils";
+import { binarySearch } from "protocol/utils";
 import { networkRequestsCache } from "replay-next/src/suspense/NetworkRequestsCache";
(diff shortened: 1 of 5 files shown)


packages/shared/test-suites/RecordingTestMetadata.ts
```diff
@@ -929,13 +930,16 @@ async function processNetworkData(
 
   // Filter by RequestInfo (because they have execution points)
   // then map RequestInfo to RequestEventInfo using ids
-  const [beginIndex, endIndex] = findSliceIndices(ids, begin.point, end.point, (id, point) => {
```
bvaughn, line 932: Was the utility broken or what? It isn't clear from this PR (empty description) why deleting that code and its tests were worth doing. Seems like the "fix" was the changes to the e2e test expectations.

hbenl: Sorry, forgot to add the description - yeah, it was broken. The test expectations needed to be updated because the Cypress panel now shows 3 network events for that test.

bvaughn (approved):

PR #9782, opened 2023-09-29
Group passed recordings of flaky tests with the failed ones

hbenl:
but show them with a "Passed" label so that they can be distinguished from the failing ones for the same test.

### DIFF
--- a/src/ui/utils/testRuns.ts
+++ b/src/ui/utils/testRuns.ts
@@ -17,7 +17,7 @@ export type RecordingGroups = {
   flakyRecordings: RecordingGroup;
 };
 
-function testPassed(recording: Recording) {
+export function testPassed(recording: Recording) {
   const testMetadata = recording.metadata?.test;
   if (testMetadata == null) {
     return false;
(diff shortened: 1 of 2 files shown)


bvaughn:
> Group passed recordings of flaky tests with the failed ones

There should be up to three groups of tests:

1. Failed
2. Flaky
3. Passing

Each test should only appear in one of the groups. (If a test had some passing runs and some failing runs, then ti should be in the flaky group– not the failed group.)

bvaughn (commented):
Can we sort test attempts in the tree by time so that the passing tests always come after/below the flaky ones? It's confusing at the moment I think.

Screen Shot 2023-09-29 at 9 54 51 AM

bvaughn (approved):
I think this is an improvement

tag @jonbell-lot23 for his thoughts though

src/ui/utils/testRuns.ts
```diff
@@ -77,34 +77,45 @@ export function groupRecordings(recordings: Recording[]): RecordingGroups {
...
-            failedRecordings.count++;
-            failedRecordings.fileNameToRecordings[filePath] = [recording];
-          }
+          addToRecordingGroup(failedRecordings, recording);
         }
       }
     }
+
+    if (didAnyTestFail && didAnyTestPass) {
+      flakyRecordings.fileNameToRecordings[filePath].sort((recordingA, recordingB) => {
```
bvaughn, line 109: Why isn't sorting by recording date/time sufficient? (Do we not have that info loaded?)

hbenl: IIRC the recording date/time is the time when the upload started and (again, IIRC) that's not reliably in the same order.

bvaughn: Ooof. That's fair.

PR #9799, opened 2023-10-01
Delete dependabot.yml to hopefully disable it

markerikson:
We tried to turn off Dependabot PRs months ago when we switched to Socket, but it still keeps opening them.  

https://stackoverflow.com/a/74421741/62937 suggests that deleting `dependabot.yml` might help.


bvaughn (approved):
Kill it

PR #9802, opened 2023-10-02
Auto-accept invites to workspaces when opening a recording

ryanjduffy:
## Issue

Our customers have somewhat frequently encountered the issue of believing they do not have access to their team because they first visit app.replay.io to accept an invite before opening a replay (e.g. from a link in their issue tracking system).

## Resolution

Adds a new method, `mayAcceptInviteAndGetRecording`, which tries to get the recording and, if it fails, tries to join the workspace and then get the recording again.

### DIFF
--- a/src/ui/hooks/recordings.ts
+++ b/src/ui/hooks/recordings.ts
@@ -4,7 +4,7 @@ import { useRouter } from "next/router";
 import { useEffect, useMemo } from "react";
 
 import { assert } from "protocol/utils";
-import { query } from "shared/graphql/apolloClient";
+import { mutate, query } from "shared/graphql/apolloClient";
 import {
   AcceptRecordingCollaboratorRequest,
   AcceptRecordingCollaboratorRequestVariables,
@@ -190,6 +190,35 @@ export async function getRecording(recordingId: RecordingId) {
   });
 
   return result.data?.recording ? convertRecording(result.data.recording) : undefined;
+}
+
+export async function maybeAcceptInviteAndGetRecording(recordingId: RecordingId) {
+  const recording = await getRecording(recordingId);
+
+  if (!recording) {
+    const result = await mutate({
+      mutation: gql`
+        mutation MaybeAcceptWorkspaceMembershipForRecording($recordingId: ID!) {
+          maybeAcceptWorkspaceMembershipForRecording(input: { recordingId: $recordingId }) {
+            success
+          }
+        }
+      `,
+      variables: { recordingId },
+      refetchQueries: ["GetRecording"],
+      // we need to ensure GetRecording has been refetched since it will be
+      // immediately called when this method returns if it was successful
+      awaitRefetchQueries: true,
+    });
+
+    const success = !!result?.data?.maybeAcceptWorkspaceMembershipForRecording.success;
+
+    if (success) {
+      return await getRecording(recordingId);
+    }
+  }
+
+  return undefined;
 }
 
 export function useGetRecording(recordingId: RecordingId | null | undefined): {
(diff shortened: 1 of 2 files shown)


markerikson (approved):

src/ui/hooks/recordings.ts
```diff
@@ -192,6 +192,35 @@ export async function getRecording(recordingId: RecordingId) {
   return result.data?.recording ? convertRecording(result.data.recording) : undefined;
 }
 
+export async function maybeAcceptInviteAndGetRecording(recordingId: RecordingId) {
+  const recording = await getRecording(recordingId);
+
+  if (!recording) {
```
bvaughn, line 198: Should we check the error returned here? seems kind of odd to always assume there's a pending invite when the id may be invalid, or there may not be one, etc.

bvaughn (changes requested):
TBH this seems like something the backend should do, not the frontend.

Seems like either a security leak on the frontend (if people are able to infer the existence of a recording by watching the network panel) _or_ a goofy extra couple of meaningless requests if the id is invalid.

ryanjduffy:
> TBH this seems like something the backend should do, not the frontend.
> 
> Seems like either a security leak on the frontend (if people are able to infer the existence of a recording by watching the network panel) _or_ a goofy extra couple of meaningless requests if the id is invalid.

Thanks for the feedback. I don't think there's an opportunity to phish for data since the backend won't respond with anything meaningful to the client. It will be the same "goofy extra requests".

> In either case, neither the user nor the frontend should have to care about this distinction. if there's a pending invite, requesting/opening the recording should accept it.

I agree it's a little strange and I shared the same first instinct that the backend could handle this silently but I don't think a `query` operation should perform a mutation so I elected to split this into separate steps. I've tagged @loganfsmyth for review on the backend side of the change so if there's disagreement, perhaps we should have a discussion when she's back to decide the way forward.

src/ui/hooks/recordings.ts
bvaughn: Yeah, that makes sense 👍🏼  I realized that after writing this comment

ryanjduffy:
> Yeah, to be honest I'm not sure why we require an explicit "accept" phase for this. If a user has been given access to a recording, and then requests that recording...it seems like the recording should just be sent. Do you know the history here?

I don't. Deferring to @jasonLaster, @jaril, or @jonbell-lot23  who might

jonbell-lot23:
I don't recall the specifics, but I do remember how we were erring on the side of caution since security/privacy is so important.

I am very supportive of removing this friction.

bvaughn:
Update: Ryan scheduled a meeting for Monday for us to chat about this.

jasonLaster:
Any of these option is better than today.
1. auto-accepting on the frontend
2. removing the accept feature in the backend 
3. asking the user to first accept

My preference is removing the friction. We can re-add it later if need be.

bvaughn:
Notes from meeting with Brian/Logan/Ryan:
* Brian: You should not have to explicitly accept an invitation to view a recording
* Logan: Don't force add me to a workspace
  * Brian: Wouldn't matter as much if we had better organizational tools for the sidebar
* Ryan: Allowing access to a recording with a pending workspace invite makes sense
  * Logan: Not against that, but need to review code to make sure other places won't fail (e.g. sharing)
* Brian: I think we can find edge cases that don't work well with any approach but we should optimize for streamlining the most common use case (viewing a recording without friction)

Next steps:
* Backend: Audit backend code to make sure nothing breaks in an unexpected way if we allow access to a recording with a pending invite
* Frontend: Make sure we handle the case of going "back" from a recording to a workspace you haven't accepted an invitation to yet; (it should prompt you to accept)

ryanjduffy:
We're likely not pursuing this PR so I'm closing.

PR #9803, opened 2023-10-02
Update `breakpoints-01` example recording ID to (temporarily) fix missing E2E tests

markerikson:
This PR:

- Updates our `breakpoints-01` example recording to a fresh ID to fix the immediate issue of dependent E2E tests not running
- updates the `rdt-02` test to match changed component counts in the new `breakpoints-01` recording, and improves the debug logging for that test
- Improves the error logging for Axios errors when an error is thrown when cloning (just error details, without dozens of internal Axios req/res fields)

I'll try to follow up with a script we can run to move targeted test recordings to a new workspace via GraphQL, so that we can avoid the retention problems.


bvaughn (approved):
When you say "temporarily" – do you mean the recording will be deleted again in a few days and this will break again?

PR #9804, opened 2023-10-02
Add move-recordings script and update breakpoints-01 example again

markerikson:
This PR:

- Adds a new CLI script that moves a recording to a different workspace on command, via GraphQL
- Bumps the `breakpoints-01` example again to one that I've moved into the separate "FE Test Golden Recordings" workspace via that script


bvaughn (approved):
Rubber stamp. Brain is mush

PR #9805, opened 2023-10-02
File search dims results that weren't executed

bvaughn:
- [x] Update file search results to use GenericList/GenericListData
- [x] Dim matches for locations that did not execution (within the focus window)
- [x] Fix bug in `findMatches` (and add test)
- [x] Add e2e test for file search

Screen Shot 2023-10-02 at 7 55 01 PM


jasonLaster (commented):
This feature is such a nice example of every day time travel. 

Also the test looks great!

PR #9807, opened 2023-10-03
New React DevTools [draft]

bvaughn:
I lifted a few pieces of code out of `SecondaryToolbox/ReactDevTools` and into separate files so they could be shared:
* `ReplayWall` class (moved to `react-devtools/ReplayWall` with minimal changes to remove the "no-change" special case)
* `loadReactDevToolsInlineModuleFromProtocol` helper function (moved to `react-devtools/suspense/reactDevToolsInlineModuleCache`)

I also added the new React DevTools as a new usage of `GenericList`/`GenericListData`.

I think the most interesting parts of this PR for code review purposes are:
* The _new_ root RDT component `SecondaryToolbox/react-devtools/components/ReactDevToolsPanel.tsx`
* The new `SelectedElement` component (where you inspect props/state/hook)
* The initialization logic in the new `react-devtools/hooks/useReactDevToolsAnnotations` and `react-devtools/hooks/useReplayWall` hooks
* The new suspense caches in `SecondaryToolbox/react-devtools/suspense`


src/ui/components/SecondaryToolbox/react-devtools/ReactDevToolsListData.ts
```diff
@@ -0,0 +1,92 @@
+import assert from "assert";
```
markerikson, line 1: :hand: intended to be an import of one of our own utils instead?

src/ui/components/SecondaryToolbox/react-devtools/ReactDevToolsListData.ts
```diff
@@ -0,0 +1,92 @@
...
+      this.collapsedFiberIds.add(element.id);
+    } else {
+      this.collapsedFiberIds.delete(element.id);
+    }
+
+    this.store.toggleIsCollapsed(element.id, isCollapsed);
+    this.invalidate();
+  }
+
+  updateStore(store: StoreWithInternals) {
```
markerikson, line 71: :question: Where/when would we need to swap the store instance?

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
```diff
@@ -0,0 +1,280 @@
...
+  children: number[];
+};
+
+export type NodeOptsWithoutBounds = Omit<NodePickerOpts, "onCheckNodeBounds">;
+
+export type StoreWithInternals = Store & {
+  _idToElement: Map<number, ElementWithChildren>;
+};
+
+export class ReplayWall implements Wall {
```
markerikson, line 29: 💬 yeah, this probably shoulda been moved to a separate file a while ago :)

src/ui/components/SecondaryToolbox/react-devtools/components/HooksRenderer.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const { inspectFunctionDefinition } = useContext(InspectorContext);
+
+  const hookObject = objectCache.read(replayClient, pauseId, `${objectId}`, "canOverflow");
+
+  let hookName = findProtocolObjectPropertyValue<string>(hookObject, "name") || "Anonymous";
+  const subHooksProperty = findProtocolObjectProperty(hookObject, "subHooks");
+  const valueProperty = findProtocolObjectProperty(hookObject, "value");
+
+  const subHooks = subHooksProperty?.object
+    ? objectCache.read(replayClient, pauseId, `${subHooksProperty.object}`, "canOverflow")
```
markerikson, line 37: :question: Suspense question I'm still a bit confused on. We've got a whole series of `someCache.read()` calls in here. Doesn't that mean the component will re-suspend N times, at least on the first render?  Does it make sense to do these in parallel?

src/ui/components/SecondaryToolbox/react-devtools/components/ReactDevToolsListItem.tsx
```diff
@@ -0,0 +1,91 @@
+import assert from "assert";
```
markerikson, line 1: :hand: same auto-import thing

src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  const dispatch = useAppDispatch();
+
+  const deferredElement = useDeferredValue(element);
+
+  const { displayName, id } = deferredElement;
+
+  const isPending = element !== deferredElement;
+
+  const [inspectedElement, [, fiberIdsToNodeIds]] = suspendInParallel(
```
markerikson, line 57: :speech_balloon: yeah, this is what I was wondering about in the other file

src/ui/components/SecondaryToolbox/react-devtools/hooks/useHighlightNativeElement.ts
```diff
@@ -0,0 +1,66 @@
...
+  const onMouseMove = (event: MouseEvent) => {
+    if (!wall || !store) {
+      return;
+    }
+
+    let currentTarget: HTMLElement | null = event.target as HTMLElement;
+    while (currentTarget) {
+      if (currentTarget.getAttribute("data-test-name") === "ReactDevToolsListItem") {
+        const id = parseInt(currentTarget.getAttribute("data-element-id") ?? "");
+        if (!isNaN(id)) {
```
markerikson, line 54: :question: stupid question, do we need `Number.isNan` here?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

src/ui/components/SecondaryToolbox/react-devtools/suspense/hookNameCache.ts
```diff
@@ -0,0 +1,115 @@
...
+      location.sourceId
+    );
+    if (!sourceContent) {
+      return fallbackName;
+    }
+
+    let lineIndex = location.line - 1;
+    let columnIndex = location.column;
+
+    // We could probably write a more efficient version of this that doesn't split the string if needed.
```
markerikson, line 66: :speech_balloon: yeah, I was running into a similar question while doing some of the ReactDOM source parsing work for the React Renders panel.

We could consider caching line-split sources separately as an intermediate option, to avoid re-splitting them for each location we are looking up.

src/ui/components/SecondaryToolbox/react-devtools/suspense/inspectedElementCache.ts
```diff
@@ -0,0 +1,57 @@
...
+  debugLabel: "DOMSearchCache",
+  getKey: ([replayClient, bridge, store, replayWall, pauseId, elementId]) =>
+    `${pauseId}:${elementId}`,
+  load: async ([replayClient, bridge, store, replayWall, pauseId, elementId]) => {
+    const rendererID = store.getRendererIDForElement(elementId);
+    const requestID = ++uidCounter;
+
+    // Wait until the backend has been injected before sending a message through the wall/bridge
+    await reactDevToolsInjectionCache.readAsync(replayClient, pauseId);
+    await Promise.resolve();
```
markerikson, line 37: :question: Why the extra tick here?

src/ui/components/SecondaryToolbox/react-devtools/utils/findProtocolObjectProperty.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  NamedValue,
+  Object as ProtocolObject,
+  Property as ProtocolProperty,
+} from "@replayio/protocol";
+
+export function findProtocolObjectProperty(
```
markerikson, line 7: :speech_balloon: I hate that we have to do this :)

src/ui/components/SecondaryToolbox/react-devtools/utils/findProtocolObjectPropertyValue.ts
```diff
@@ -0,0 +1,10 @@
+import { Object as ProtocolObject } from "@replayio/protocol";
```
markerikson, line 1: :question: silly question, but why not have both of these in the same file?

src/ui/components/SecondaryToolbox/react-devtools/utils/findProtocolObjectProperty.ts
bvaughn: You and me both

src/ui/components/SecondaryToolbox/react-devtools/suspense/inspectedElementCache.ts
bvaughn: Oh, something I added while debugging the backend issues I think. Meant to remove it. Good call.

src/ui/components/SecondaryToolbox/react-devtools/hooks/useHighlightNativeElement.ts
bvaughn: I think either is okay. We use `isNaN` more in Replay than `Number.isNaN` FWIW

src/ui/components/SecondaryToolbox/react-devtools/ReactDevToolsListData.ts
bvaughn: No, I meant to import this one. We also have our own `assert()` functions (a few of them). They're all equivalent I think. I'm in the habit of importing this one though 🤷🏼

src/ui/components/SecondaryToolbox/react-devtools/components/HooksRenderer.tsx
bvaughn: Yes, but we can't parallelize the calls in this case (unfortunately) because we have to first fetch the hook object to read it's "subHooks" property (if it has one) to them fetch it. I _think_ we might be able to avoid re-fetching the sub-hooks if our cache pre-caches the object previews but I'm not sure.

src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
bvaughn: In this case it's different because we _can_ fetch these in parallel (we know enough up front to do that)

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -385,53 +103,12 @@ function createReactDevTools(
...
 
   return [ReactDevTools, wall, bridge] as const;
 }
 
-// React DevTools (RD) changed its internal data structure slightly in a minor update.
-// The result is that Replay sessions recorded with older versions of RD don't play well in newer versions.
-// We can work around this by checking RD's "bridge protocol" version (which we also store)
-// and loading the appropriate frontend version to match.
-// For more information see https://github.com/facebook/react/issues/24219
-async function loadReactDevToolsInlineModuleFromProtocol(
```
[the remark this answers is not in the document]
bvaughn: Good call. I'll remove it.

src/ui/components/SecondaryToolbox/react-devtools/ReactDevToolsListData.ts
bvaughn: Good question. I wrote this early on, thinking that maybe we would need to recreate the store rather than update it. I just forgot to remove the method. (430e47596)

src/ui/components/SecondaryToolbox/react-devtools/utils/findProtocolObjectPropertyValue.ts
bvaughn: Your comment is below the `Object` type import but I think you're probably asking about the `findProtocolObjectProperty` utility function?

Just a subjective choice I guess. That function is called directly from other utility functions (like `findProtocolObjectPropertyValue`) so it makes more sense to me, organizationally, to have it in its own module. It's subjective though :)

bvaughn:
I think I have made changes to all of the things pointed out or requested.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -1,4 +1,5 @@
 import { NodeBounds, Object as ProtocolObject } from "@replayio/protocol";
+import { createBridge, createStore, initialize } from "@replayio/react-devtools-inline/frontend";
```
markerikson, line 2: :hand: Hmm. I'm going to half-rethink my previous comment.

I still think we don't need to worry about the RDT backend version at this point, it's always going to be the "modern" version.

But, this does now switch from a lazy import of `react-devtools-inline` (and thus hopefully a separate chunk only loaded on demand) to _always_ loading it.

With Next's build setup, I don't know if that `await import("@replayio/react-devtools-inline/frontend")` was actually getting written to a separate chunk, or if it was getting inlined into the larger frontend bundle.  I'd have to run a bundle analysis to check.

We haven't spent any time lately on trying to optimize JS bundle sizes or page loads.  Honestly this probably is fine for now and we can do bundle size analysis later - I'm just noting that this _is_ a shift in how we import the package which could result in the main app bundle being larger.

markerikson (approved):
We ought to look at our bundle size at some point on general principle and see if `react-devtools-inline` is now getting added to the main bundle, and see if we need to switch back to a dynamic import, but let's go with this for now. No reason to hold this up.

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
bvaughn: Yeah, FWIW we lazily import the root `ReactDevTools` component:
```js
const ReactDevToolsPanel = React.lazy(() => import("./ReactDevTools"));
```

This should decouple all of that logic from the main bundle, so far as I understand. Looking at my fork, I can verify that two JS files (~236kb and ~53kb) are downloaded only once the React tab has been clicked and opened.

bvaughn:
Merged by way of stacked PR #9854

PR #9827, opened 2023-10-16
Optimize RDT component picker setup perf

markerikson:
This PR:

- Adds a new `supportsObjectIdEvaluations` field to our capabilities object
- ~~Tries to rewrite the core `NodePicker` class to centralize some of the behavior that's currently split across multiple files, and cut down on the amount of work done while hovering.  **This is currently pretty broken and needs a lot more thought**.~~
- ~~Experiments with an alternate approach for coming up with a `nodeId->fiberId` map, which makes use of the new `__RECORD_REPLAY__.getProtocolIdForObject()` method available in evaluations as of Chromium 2023-09-16~~
  - ~~Current approach: loop over all renderers and all roots inside the eval, get the root DOM nodes, call `rootNode.querySelectorAll("*")`, then call the simpler `renderer.getFiberIdForNative()` instance that hopefully requires less work inside the RDT bundle~~
  - Additionally, all we care about is the node IDs, not the entire DOM node previews.  We don't currently handle the 1000-property preview limit, and I also suspect serializing all those DOM node previews is expensive.  Switched to a `JSON.stringify(nodeIdsToFiberIds)`, and then having to do the same "chunk this string in the eval and reconstruct it on the other side" trick I'm doing in the RDT routine and in the recorded protocol messages logic.

~~Still _very_ WIP, but I'm tentatively hopeful.  A single hands-on check showed the existing approach took 2000ms to return data on 600 DOM nodes.  My initial attempt of still returning a `Map` ended up hitting the 1000-property limit, and that took 3000ms to return.  BUT, doing the node ID lookups in the eval and returning a record let me return 1100 node IDs in 50ms (!!!).~~

## Update

This should be ready now. Final changes:

- Did add the `capabilities.supportsObjectIdEvaluations` field (but had to fix the build date logic that was broken earlier)
- Dropped the `NodePicker` rewrite entirely, other than switching from passing `nodeIds?: string[]` to `nodeIds?: Set<string>` for slightly faster filtering checks
- Consolidated all the fiber ID eval logic in a cache + one evaluated function.  Both of them handle a fast path (if object ID lookups are available) and a slow path (the existing approach of sending back full DOM node previews).  With the fast path, the eval is coming back in ~150ms rather than 1500ms+.
- Started prefetching the `Map<nodeId, fiberId>` data as soon as we pause.
- Now using the same nodeId/fiberIds cache results to look up what DOM nodes to highlight when hovering over component tree items as well, which avoids further individual evals since we have that data on hand already

### DIFF
--- a/packages/replay-next/src/suspense/BuildIdCache.ts
+++ b/packages/replay-next/src/suspense/BuildIdCache.ts
@@ -16,6 +16,7 @@ export type RecordingCapabilities = {
   supportsNetworkRequests: boolean;
   supportsRepaintingGraphics: boolean;
   supportsPureEvaluation: boolean;
+  supportsObjectIdLookupsInEvaluations: boolean;
 };
 
 // Target applications which can create recordings.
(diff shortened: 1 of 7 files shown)


markerikson:
Note that our RDT E2E tests should indeed try both the "fast path" and "slow path" - an FF recording for the slow path, and a Chromium recording past 2023-09-16 for the fast path.

markerikson:
Vercel's being a pain. Going to close this and open up a duplicate PR to see if that behaves better.

PR #9828, opened 2023-10-17
Refactor the code for resume/step commands

hbenl:
- stop using frame steps in the CommandBar
- add a suspense cache for resume targets
- disable stepping buttons if we can't find resume targets for them


packages/replay-next/src/suspense/FrameCache.ts
```diff
@@ -8,13 +8,16 @@ import { cachePauseData, sortFramesAndUpdateLocations } from "./PauseCache";
 import { sourcesByIdCache } from "./SourcesCache";
 
 export const framesCache: Cache<
-  [replayClient: ReplayClientInterface, pauseId: PauseId],
+  [replayClient: ReplayClientInterface, pauseId: PauseId | undefined],
```
bvaughn, line 11: Suspense caches can be called conditionally, so this change seems unnecessary.

I think it's best to avoid returning undefined from a method whenever possible in terms of type safety. Do we really _need_ this change?

packages/replay-next/src/suspense/ResumeTargetCache.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    let currentPoint = point;
+    if (pauseAndFrameId) {
+      const [pausePoint] = getPointAndTimeForPauseId(pauseAndFrameId.pauseId);
+      if (!pausePoint) {
+        return;
+      }
+
+      const frames = await framesCache.readAsync(replayClient, pauseAndFrameId.pauseId);
```
bvaughn, line 68: I think we should return early here if `pauseAndFrameId.pauseId` is undefined (and not change the param or return types in the frames cache)

bvaughn (approved):
Overall approved, but I think we should change back the params and return type in the frames cache. Let's chat more before merging if that's not possible?

packages/replay-next/src/suspense/FrameCache.ts
hbenl: I'm using `useImperativeCacheValue()` with that cache in `CommandBar.tsx` and also in #9831, so I do need this.
Also, this is only changing the type of a parameter but not the return type.

bvaughn: Ah, right. Okay fair enough.

PR #9831, opened 2023-10-19
Update the pause info panel badge when receiving frames

hbenl:
This PR is on top of #9828 because it depends on a change to the `framesCache` in that PR.


bvaughn (approved):
Nice!

This one's approved once that one lands :)

PR #9834, opened 2023-10-19
Fix Elements resize handle style

bvaughn:
Before

Screen Shot 2023-10-19 at 11 17 11 AM

After

Screen Shot 2023-10-19 at 11 16 42 AM


bvaughn:
This looks pretty bad so I'm going to quick land a fix. Let me know if you want any follow up changes.

PR #9836, opened 2023-10-19
Derive cypress marker frame when missing in call stack

ryanjduffy:
For requests that are dispatched from a Cypress promise handler, we don't always see `__stackReplacemenMarker` so we have to derive the callsite from the sourceIds.

The failing step in recording `9734cbaf-f089-441a-8090-3d0528a19110` shows this behavior (though it's the only case I've ever seen but is also recorded on cypress 13 so it may be new behavior)

### DIFF
--- a/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
@@ -1,5 +1,5 @@
 import assert from "assert";
-import { Location } from "@replayio/protocol";
+import { Frame, Location } from "@replayio/protocol";
 import { compare } from "compare-versions";
 
 import { framesCache } from "replay-next/src/suspense/FrameCache";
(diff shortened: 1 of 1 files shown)


src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
```diff
@@ -8,6 +8,41 @@ import { createCacheWithTelemetry } from "replay-next/src/utils/suspense";
 import { ReplayClientInterface } from "shared/client/types";
 import { GroupedTestCases, UserActionEvent } from "shared/test-suites/RecordingTestMetadata";
 
+function getCypressMarkerFrame(frames: Frame[]) {
```
bvaughn, line 11: I strongly dislike that we have this level of coupling/logic in the frontend

bvaughn (commented):
The trajectory of adding more and more framework-specific complexity and coupling into the frontend seems bad. Do we have concrete plans for moving this stuff out into the Cypress plug-in or at least Cypress-specific routine code?

ryanjduffy:
> This kind of feels like a bandaid fix.
> 
> Our overall trajectory of adding more and more framework-specific complexity and coupling into the frontend seems bad. Do we have concrete plans for moving this stuff out into the Cypress plug-in or at least Cypress-specific routine code?

I disagree that this is a bandaid; it's a fix that operates within the current architecture. This is the bed we've made by building part of our product on top of someone else's that didn't intend for our use case. This logic (or some variation there of) will exist somewhere in our product in order to achieve this goal.

I agree that we need to migrate all of this code into routines to simplify our client logic but I don't think we can stop shipping fixes while that work is pending. I'll link this work into that ticket so we ensure it is accounted for.

PR #9837, opened 2023-10-20
Don't use stepStartPoint for "get" steps in cypress recordings

hbenl:
FE-1944 only asks for the before/after HUD to be hidden for "get" steps (and to always show the "after" screenshot) but I assume that other interactions should also be changed to remain consistent:
- hovering over a "get" step should show the "after" screenshot
- clicking on a "get" step should take you to the "after" point

This is achieved by setting `beforeStep` to `stepEndPoint` for "get" steps.

### DIFF
--- a/packages/e2e-tests/tests/cypress-03_test-step-interactions.test.ts
+++ b/packages/e2e-tests/tests/cypress-03_test-step-interactions.test.ts
@@ -60,12 +60,6 @@ test("cypress-03: Test Step interactions", async ({ pageWithMeta: { page, record
   // Clicking in viewer mode shouldn't switch to DevTools mode
   expect(await isViewerTabActive(page)).toBe(true);
 
-  const firstGet = clickableSteps.filter({ hasText: /get/ }).first();
-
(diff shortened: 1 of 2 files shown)


bvaughn (approved):
It's unfortunate, how this logic is spread throughout the frontend.

Lot of e2e test failures. Presumably unrelated to this change?

hbenl:
> It's unfortunate, how this logic is spread throughout the frontend.

That's why I put it into the generator for `RecordingTestMetadataV3` - I think that's where ideally all the framework-specific logic should live.

> Lot of e2e test failures. Presumably unrelated to this change?

One of the tests was actually testing what this PR disables :laughing:. The other failures were due to poor backend performance.

bvaughn:
> That's why I put it into the generator for RecordingTestMetadataV3 - I think that's where ideally all the framework-specific logic should live.

Yeah, that's fair. I've seen it in other places too recently (#9836) I guess and I wish the client didn't have to care about any of it.

PR #9838, opened 2023-10-21
Use getPointStack for getting pause frames in Redux J2C (FE-1966)

Mokshit06:
Fixes [FE-1966](https://linear.app/replay/issue/FE-1966). The issue was that we were only using `getPointStack` for accessing the frame step rather than the pause frames. Because of this the frame index that we had in the _filtered_ frames didn't match to the actual index for _all_ the frames.


markerikson:
Tested it out on our own FE recordings, and it looks great!

PR #9839, opened 2023-10-23
Update timeline immediately when seeking

hbenl:
We were already [trying to do this](https://github.com/replayio/devtools/blob/bf1bec8037dd4bb67efbea09fbae2273e043c781/src/ui/actions/timeline.ts#L265-L267) but we were using `setTimelineToTime()` which doesn't really do what it says. To avoid similar confusion in the future, I renamed it to `setHoverTime()`.

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -160,7 +160,7 @@ function onPaused({ point, time }: PauseEventArgs): UIThunkAction {
   };
 }
 
-export function setTimelineToTime(time: number | null, updateGraphics = true): UIThunkAction {
+export function setHoverTime(time: number | null, updateGraphics = true): UIThunkAction {
   return async (dispatch, getState) => {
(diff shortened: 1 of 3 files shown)


src/ui/actions/timeline.ts
```diff
@@ -160,7 +160,7 @@ function onPaused({ point, time }: PauseEventArgs): UIThunkAction {
   };
 }
 
-export function setTimelineToTime(time: number | null, updateGraphics = true): UIThunkAction {
+export function setHoverTime(time: number | null, updateGraphics = true): UIThunkAction {
```
bvaughn, line 163: Nice clarification. Thanks!

bvaughn (approved):
Failing Cypress 3 test related? That's been a _little_ flaky on main but mostly passing.

hbenl:
> Failing Cypress 3 test related?

No, it fails in `waitForRecordingToFinishIndexing()`.

PR #9843, opened 2023-10-24
Optimize RDT component picker setup perf

markerikson:
This PR is a duplicate of #9827 to see if Vercel deploys properly (AGAIN!)

Copying description of the changes:

- Did add the `capabilities.supportsObjectIdEvaluations` field (but had to fix the build date logic that was broken earlier in the draft)
- Dropped the `NodePicker` rewrite entirely, other than switching from passing `nodeIds?: string[]` to `nodeIds?: Set<string>` for slightly faster filtering checks
- Consolidated all the fiber ID eval logic in a cache + one evaluated function.  Both of them handle a fast path (if object ID lookups are available) and a slow path (the existing approach of sending back full DOM node previews).  With the fast path, the eval is coming back in ~150ms rather than 1500ms+.
- Started prefetching the `Map<nodeId, fiberId>` data as soon as we pause.
- Now using the same nodeId/fiberIds cache results to look up what DOM nodes to highlight when hovering over component tree items as well, which avoids further individual evals since we have that data on hand already

### DIFF
--- a/packages/replay-next/src/suspense/BuildIdCache.ts
+++ b/packages/replay-next/src/suspense/BuildIdCache.ts
@@ -16,6 +16,7 @@ export type RecordingCapabilities = {
   supportsNetworkRequests: boolean;
   supportsRepaintingGraphics: boolean;
   supportsPureEvaluation: boolean;
+  supportsObjectIdLookupsInEvaluations: boolean;
 };
 
 // Target applications which can create recordings.
@@ -24,6 +25,39 @@ export enum RecordingTarget {
   chromium = "chromium",
   node = "node",
   unknown = "unknown",
+}
+
+/**
+ * Takes a build date (i.e. 'YYYYMMDD') and converts it to a `Date` object.
+ */
+export function buildDateStringToDate(buildDate: string) {
+  const y = buildDate.substring(0, 4);
+  const m = buildDate.substring(4, 6);
+  const d = buildDate.substring(6);
+
+  return new Date(`${y}-${m}-${d}T00:00:00Z`);
+}
+
+export interface BuildComponents {
+  platform: string;
+  runtime: string;
+  date: string;
+}
+
+// Ported from the backend
+export function parseBuildIdComponents(buildId: string): BuildComponents | null {
+  // Add the platform, runtime, and date to the metadata, which we can determine
+  // from the build ID itself.
+  const match = /(.*?)-(.*?)-(.*?)-/.exec(buildId);
+
+  if (match) {
+    return { platform: match[1], runtime: match[2], date: match[3] };
+  }
+
+  // NOTE: Only old builds that are outside our supported range should hit
+  // this case, but this function may well be called for those older builds
+  // still if someone tries to view an older recording.
+  return null;
 }
 
 function getRecordingTarget(buildId: string): RecordingTarget {
@@ -51,10 +85,15 @@ export const recordingTargetCache = createSingleEntryCache<
   },
 });
 
-function getRecordingCapabilities(buildId: string) {
-  const recordingTarget = getRecordingTarget(buildId);
+function getRecordingCapabilities(
+  recordingTarget: RecordingTarget,
+  buildComponents: BuildComponents
+) {
   switch (recordingTarget) {
     case "chromium": {
+      const buildDate = buildDateStringToDate(buildComponents?.date ?? "");
+      const supportsObjectIdLookupsInEvaluations = buildDate >= new Date("2023-09-16");
+
       return {
         supportsEagerEvaluation: false,
         supportsElementsInspector: true,
@@ -62,6 +101,7 @@ function getRecordingCapabilities(buildId: string) {
         supportsNetworkRequests: false,
         supportsRepaintingGraphics: features.chromiumRepaints,
         supportsPureEvaluation: false,
+        supportsObjectIdLookupsInEvaluations,
       };
     }
     case "gecko": {
@@ -72,6 +112,7 @@ function getRecordingCapabilities(buildId: string) {
         supportsNetworkRequests: true,
         supportsRepaintingGraphics: true,
         supportsPureEvaluation: true,
+        supportsObjectIdLookupsInEvaluations: false,
       };
     }
     case "node": {
@@ -82,6 +123,7 @@ function getRecordingCapabilities(buildId: string) {
         supportsNetworkRequests: true,
         supportsRepaintingGraphics: false,
         supportsPureEvaluation: false,
+        supportsObjectIdLookupsInEvaluations: false,
       };
     }
     case "unknown":
@@ -93,6 +135,7 @@ function getRecordingCapabilities(buildId: string) {
         supportsNetworkRequests: true,
         supportsRepaintingGraphics: true,
         supportsPureEvaluation: false,
+        supportsObjectIdLookupsInEvaluations: false,
       };
     }
   }
@@ -106,6 +149,13 @@ export const recordingCapabilitiesCache = createSingleEntryCache<
   debugLabel: "recordingCapabilitiesCache",
   load: async ([replayClient]) => {
     const recordingTarget = await recordingTargetCache.readAsync(replayClient);
-    return getRecordingCapabilities(recordingTarget);
+    const buildId = await buildIdCache.readAsync(replayClient);
+
+    const buildComponents = parseBuildIdComponents(buildId) ?? {
+      platform: recordingTarget,
+      runtime: recordingTarget,
+      date: "",
+    };
+    return getRecordingCapabilities(recordingTarget, buildComponents);
   },
 });
--- a/packages/replay-next/src/suspense/BuildIdCache.ts
+++ b/packages/replay-next/src/suspense/BuildIdCache.ts
@@ -152,8 +152,8 @@ export const recordingCapabilitiesCache = createSingleEntryCache<
     const buildId = await buildIdCache.readAsync(replayClient);
 
     const buildComponents = parseBuildIdComponents(buildId) ?? {
-      platform: recordingTarget,
-      runtime: recordingTarget,
+      platform: "unknown",
+      runtime: "unknown",
       date: "",
     };
     return getRecordingCapabilities(recordingTarget, buildComponents);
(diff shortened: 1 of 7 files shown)


packages/replay-next/src/suspense/BuildIdCache.ts
```diff
@@ -106,6 +149,13 @@ export const recordingCapabilitiesCache = createSingleEntryCache<
...
   load: async ([replayClient]) => {
     const recordingTarget = await recordingTargetCache.readAsync(replayClient);
-    return getRecordingCapabilities(recordingTarget);
+    const buildId = await buildIdCache.readAsync(replayClient);
+
+    const buildComponents = parseBuildIdComponents(buildId) ?? {
+      platform: recordingTarget,
+      runtime: recordingTarget,
+      date: "",
+    };
```
bvaughn, line 158: nit but this feels more honest / less likely to possibly trigger any kind of false positive
```suggestion
    const buildComponents = parseBuildIdComponents(buildId) ?? {
      platform: "unknown",
      runtime: "unknown",
      date: "",
    };
```

packages/replay-next/src/suspense/BuildIdCache.ts
```diff
@@ -51,17 +85,23 @@ export const recordingTargetCache = createSingleEntryCache<
...
-function getRecordingCapabilities(buildId: string) {
-  const recordingTarget = getRecordingTarget(buildId);
+function getRecordingCapabilities(
+  recordingTarget: RecordingTarget,
+  buildComponents: BuildComponents
+) {
   switch (recordingTarget) {
     case "chromium": {
+      const buildDate = buildDateStringToDate(buildComponents?.date ?? "");
+      const supportsObjectIdLookupsInEvaluations = buildDate >= new Date("2023-09-16");
```
bvaughn, line 95: cc @Domiii for another example of how a capabilities API would be useful and the types of things FE is having to do to work around the absence of one 🙇🏼

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -123,6 +124,20 @@ class ReplayWall implements Wall {
...
   }
 
+  // This is called  by the `useEffect` in the React DevTools panel.
+  async setUpRDTInternalsForCurrentPause() {
+    // Preload the React DevTools Backend for this pause
+    await this.ensureReactDevtoolsBackendLoaded();
+    // We should also kick off a request to pre-fetch the `Map<nodeId, fiberId>` lookup for this pause.
+    // That way, we have the data on hand when the user clicks on a node.
+    // This is async, but we won't wait for it here.
+    this.fetchNodeIdsToFiberIdsForPause();
```
bvaughn, line 134: I think I know why this is desirable in terms of speeding up the RDT node picker, but this will fire every time the execution point changes when the RDT component is mounted/visible, which means that in the most common case (where the node picker isn't even used) we'll be doing _a lot_ of "unnecessary" work.

What's the expected impact/overhead of that? Is it worth it?

After the performance improvement, do things feel okay enough without this?

markerikson:
Looking into the RDT-01 failure...

src/ui/components/SecondaryToolbox/react-devtools/injectReactDevtoolsBackend.ts
```diff
@@ -111,5 +142,159 @@ export async function injectReactDevtoolsBackend(
...
+  isObjectIdCapable: boolean
+): Map<HTMLElement, number> | ChunksArray {
+  // Modern: if we have object ID lookup in evals, save just the IDs
+  // This only works with Chromium 2023-09-16+
+  const nodeIdsToFiberIds: Record<string, number> = {};
+  // Legacy: if we don't have object ID lookup in evals, save the full objects
+  // This works with Firefox and older Chromium
+  const domNodesToFiberIds = new Map<HTMLElement, number>();
+
+  const NO_NODES: HTMLElement[] = [];
```
bvaughn, line 175: nit: There isn't much value in defining a "constant" like this within the body of a function.

markerikson:
Looks like a last-minute refactor broke the logic, actually, and the E2E test caught it.  Fixing...

bvaughn:
Testing the new picker on the Vercel deploy (to see how much faster it is) and it's hanging.

bvaughn:
> Looks like a last-minute refactor broke the logic, actually, and the E2E test caught it. (Tried to extract the splitStringIntoChunks function so I don't have to copy-paste it every time, but I think injecting it into the eval string the way I am now isn't actually working.)

I thought that might be the case (bc of how this works in RDT)

You'd have to use a !raw import to get the function as a string, I _think_.

markerikson:
@bvaughn yeah, per my last comment right before your review I did break this at the last minute. Fixing it now.

bvaughn:
> I think the actual problem is actually just the use of an IIFE 

I think the way you're to-stringing that imported function doesn't work (see my above comment)

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
markerikson: I feel good about this, yeah. We're already doing one eval per pause anyway while the RDT is visible to inject the backend logic, this adds a second to pre-fetch the fiber IDs. In my tests against one of our own FE client recordings, that's taking ~150ms.  We do lots of evals all the time anyway, so I don't think this is meaningfully expensive.

bvaughn: What about the slow path though? That's a lot of work, no?

bvaughn: I think it would be better to not do it until/unless the user is likely to be interacting with the node picker. Do we have a way to anticipate that? Maybe only do it if the React panel is focused?

markerikson: We do also use it for the hover behavior now as well, which can happen without any immediate focus.  

If we make the assumption that users are likely to be interacting with recent Chromium recordings (which will be more true over time), this seems acceptable.  (Especially since we know Pinterest is on current Chromium.)

bvaughn: Assuming happy path, I don't think 150ms is too much to pay for an initialization cost TBH. I think that would be better than always doing this.

markerikson:
The `${someFunction}` part does work fine to stringify it.  That implicitly calls `.toString()` on any variable, and we already use that syntax in multiple places.

Latest push should fix the issues - went ahead and reinlined things.

bvaughn:
Nice! I must have been incorrect the in my remembrance of how that works. I think...maybe the thing sticking in my memory was that it was easy to accidentally reference imports that way which would _def_ break things.

bvaughn:
still hanging for me on Vercel.

markerikson:
Tried out the recording Brian made of `react.dev`.  Looks like the `getAllBoundingClientRects` API request is command-failing, likely due to that being in a Chrome for Mac recording.  That call is a prerequisite to being able to use either the DOM or React component pickers, so we're never even making use of the changes in this PR.

I checked that same recording, and the actual eval to collect the fiber/node IDs came back in 58ms:

image

so that part looks like it should work fine - we just need to check against recordings that aren't busted :)

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
bvaughn: > I think it would be better to not do it until/unless the user is likely to be interacting with the node picker

> I checked that same recording, and the actual eval to collect the fiber/node IDs came back in 58ms:

Especially if it's this fast.

packages/replay-next/src/suspense/BuildIdCache.ts
Domiii: @bvaughn - Thanks! I added it [here](https://linear.app/replay/issue/RUN-1148/add-capabilities-metadata-api#comment-14401bf1)

bvaughn:
Okay I just tested this change with one of the e2e tests and it's **much** faster. Nice! 🥳 

My only remaining concern then is the pre-fetching you added (the thread we have above). I think it's wasteful of our computing resources (so it might cost us money) and I think it's unnecessary given how fast it is to fetch the info using the new approach (50-150ms?)

I wonder if someone from Backend might have opinions about the computing/cost concern. Maybe I'm wrong and it's super cheap and we shouldn't care about it? (cc @jazzdan or @jcmorrow)

markerikson:
I'll admit I'm a bit confused by the sudden focus on cost as a concern here.

We run _lots_ of evals from the frontend, for a lot of different things.  Every print statement spawns evals at many points.  I don't see how adding one eval per pause, _if_ you already have the React tab open, is suddenly a reason to be significantly concerned about the potential cost impact.

If you've got the React tab open, it's already an indication you are thinking about interacting with it.  Same with the Cypress panel, and we're now pre-fetching all step details for a test as soon as you click into that test.

I think the "it's fast enough now we _don't_ have to pre-fetch" statement is a decent point, but I don't see how cost of a handful of evals is something we should be worrying about here.

bvaughn:
> I'll admit I'm a bit confused by the sudden focus on cost as a concern here.

Performance/cost is not a new concern for me. I don't think that's a fair statement and I'm a bit surprised that you'd say that.

A chunk of the backend team has done a lot of work over the past couple of months to reduce unnecessary costs. The question I'm raising here is intended to prevent us from unnecessarily _adding_ cost. That is within the scope of a PR. It is worth considering how much cost we're adding and if it's worth adding. It seems we disagree but the discussion is worth having.

Print statements are explicit opt-in, so I don't think they're a valid comparison. The React tab being open is a signal, but not necessarily always. (You may have just left it open as you browse around stepping through the Source viewer or looking at comments.)

bvaughn:
The issue we're discussing is probably not a hill worth dying on. If you want to override my concern and land this, do it. Overall the changes in this PR are a dramatic improvement in user-facing performance.

I'm kind of disappointed with the way this sub-discussion went though. I think we should share the goal of making things as efficient as (reasonably) possible, both in terms of runtime and cost. I'm not sure why it seems that we aren't aligning here.

markerikson:
Following up on the separate discussion here:

Jason told us that the expensive aspects to the backend are having more EC2 instances, which in turn is affected by number of active user sessions (and to _some_ extent, number of unique browser processes that are being managed).  On the flip side, once we have a pause (and thus a browser process available at that point), making additional API calls for that pause is effectively "free" cost-wise.

Brian asked me to double-check what happens if you have the React tab up and hit "Play" for the timeline, to see if we spam more pause requests.  I tried it, and it appears to match my expectation that all we do is fetch more paints until you actually hit "Pause" on the playback, at which point we call `createPause`, save a pause ID, and other parts of the UI may react to that by fetching stack frames or running evals:

image


In this case, it was lots of graphics, eval for injecting the RDT backend, and eval for fetching the fiber IDs per this PR.

I think that addresses all the outstanding questions for this PR, so I'll go ahead and merge it.

PR #9844, opened 2023-10-24
Add more source search filtering options (FE-1964)

Mokshit06:
This PR implements the features mentioned in FE-1964 which include filtering based on: regex, whole word, case sensitivity and open editors.
Screenshot 2023-10-25 at 2 24 26 AM


bvaughn:
Haven't reviewed yet but Trunk is failing because of a formatting problem in `packages/shared/client/ReplayClient.ts` and it looks like a bunch of the automated tests failed because of changes to the DOM (e.g. [here is a failure](https://app.replay.io/recording/display-an-overflow-message--82a9e793-484d-4780-8327-22e6ab5dfb5b) caused by the removal of the node modules checkbox).

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -92,17 +145,69 @@ export default function SearchFiles({ limit }: { limit?: number }) {
...
+          />
+        </div>
+        <div
+          className={styles.InputWrapper}
+          data-error={didError || undefined}
+          data-test-id="ExcludeFiles-SearchInput"
+        >
+          <Icon className={styles.Icon} type="folder-closed" />
+          <input
+            autoFocus
```
bvaughn, line 196:
```suggestion
```
This should not be auto focused.

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -92,17 +145,69 @@ export default function SearchFiles({ limit }: { limit?: number }) {
...
-            checked={includeNodeModules}
-            onChange={() => setIncludeNodeModules(!includeNodeModules)}
+        <div
+          className={styles.InputWrapper}
+          data-error={didError || undefined}
+          data-test-id="IncludeFiles-SearchInput"
+        >
+          <Icon className={styles.Icon} type="folder-open" />
+          <input
+            autoFocus
```
bvaughn, line 174:
```suggestion
```
This should not be auto focused.

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -92,17 +145,69 @@ export default function SearchFiles({ limit }: { limit?: number }) {
...
+            active={useRegex}
+            toggle={() => setUseRegex(!useRegex)}
+            icon="regex"
+            tooltip="Use Regular Expression"
+          />
         </div>
-
-        <div className={styles.CheckboxWrapper}>
-          <Checkbox
-            dataTestId="FileSearch-IncludeNodeModules"
```
bvaughn, line 99: This change broke several of the e2e tests.

packages/replay-next/src/suspense/SearchCache.ts
```diff
@@ -69,18 +71,34 @@ export function assertSourceSearchResultMatch(
...
 
 export const searchCache = createStreamingCache<
-  [replayClient: ReplayClientInterface, query: string, includeNodeModules: boolean, limit?: number],
+  [
+    replayClient: ReplayClientInterface,
+    query: string,
+    defaultFilter: boolean,
+    includedFiles: string,
+    excludedFiles: string,
+    openSourceIds: string[] | null,
```
bvaughn, line 80: nit: When I first read this I was confused about what "open source" meant (like node_modules)

 I wonder if a name like "active sources" would be less confusing/ambiguous than "open sources"?

packages/replay-next/src/suspense/SearchCache.ts
```diff
@@ -69,18 +71,34 @@ export function assertSourceSearchResultMatch(
 }
 
 export const searchCache = createStreamingCache<
-  [replayClient: ReplayClientInterface, query: string, includeNodeModules: boolean, limit?: number],
+  [
+    replayClient: ReplayClientInterface,
+    query: string,
+    defaultFilter: boolean,
```
bvaughn, line 77: I _think_ "default filter" means hiding node_modules and CDN files?

I see VS Code has a setting that this seems to be based off of "use exclude settings and ignored filed" but I think ours is a bit different. VS Code's seems to be based off of things like the `.gitignore` file, where as ours is just hard-coded to filter out NPM packages.

So can we name this something more semantically meaningful than "default"? :)

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -92,17 +145,69 @@ export default function SearchFiles({ limit }: { limit?: number }) {
...
+            onChange={onChange(setExcludedFiles)}
+            placeholder="Files to exclude..."
+            type="text"
+            value={excludedFiles}
+          />
+          <FilterButton
+            active={defaultSrcFilter}
+            toggle={() => setDefaultSrcFilter(!defaultSrcFilter)}
+            icon="settings-off"
+            tooltip="Use Default Exclude Settings"
```
bvaughn, line 208: We aren't doing the same thing as VS Code, so let's make this tooltip clearer about what we _are_ doing?
```suggestion
            tooltip="Exclude NPM modules"
```

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -10,44 +10,97 @@ import {
...
+    (event: ChangeEvent<HTMLInputElement>) => {
+      dispatcher(event.target.value);
+    };
 
   const onKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
     switch (event.key) {
       case "Enter":
       case "NumpadEnter":
         startTransition(() => {
+          console.log({ setting: queryForDisplay });
```
bvaughn, line 103:
```suggestion
```

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -10,44 +10,97 @@ import {
...
   const didError = status === STATUS_REJECTED;
 
-  const onChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setQueryForDisplay(event.target.value);
-  };
+  const onChange =
+    (dispatcher: React.Dispatch<React.SetStateAction<string>>) =>
+    (event: ChangeEvent<HTMLInputElement>) => {
+      dispatcher(event.target.value);
+    };
```
bvaughn, line 96: Neat

packages/replay-next/src/suspense/SearchCache.ts
```diff
@@ -99,13 +117,32 @@ export const searchCache = createStreamingCache<
...
+      }
+
+      if (defaultFilter && (isModuleFromCdn(source) || isNodeModule(source))) {
+        return false;
+      }
+
+      return (
+        (!includedFiles || includedFilesMatcher.match(source.url)) &&
+        !excludedFilesMatcher.match(source.url)
+      );
```
bvaughn, line 138: nit: I think this could be written a little more clearly if you split it up:

```js
if (excludedFiles && excludedFilesMatcher.match(source.url)) {
  return false;
}

if (includedFiles && !includedFilesMatcher.match(source.url)) {
  return false;
}

return true;
```

bvaughn (commented):
This is a nice feature change. Couple of things I think we should change but overall this is really cool work 👍🏼 

One thing I noticed while testing it– we have a significant difference from VS Code in that we aren't running search until the user types Enter _or_ clicks a button. (Compared to VS Code, which automatically searches after a small debounce.)

I think we should either add some kind of pending UI to let the user know that they aren't looking at the latest results yet (since we're so similar to VS Code in other ways) _or_ we should automatically search after a debounce too. (I think we should do the latter TBH.)

packages/replay-next/src/suspense/SearchCache.ts
Mokshit06: I agree, this is something that confused me at first when navigating the codebase as well. I kept it `openSourceIds` as that is what `SourceContext` is also naming it as. Maybe it makes sense changing there as well?

bvaughn: > Maybe it makes sense changing there as well?

👍🏼 Yeah I dig that.

packages/replay-next/playwright/tests/source-search/display-an-overflow-message.ts
```diff
@@ -11,7 +11,7 @@ import { beforeEach } from "./beforeEach";
 beforeEach();
 
 test("display an overflow message", async ({ page }, testInfo) => {
-  await toggleIncludeNodeModulesCheckbox(page, true);
+  await toggleExcludeNodeModulesCheckbox(page, true);
```
bvaughn, line 14: This looks like a mistake. (We were saying toggle-include: true and now we're saying toggle-exclude: true)

packages/replay-next/playwright/tests/source-search/should-expand-and-collapse-sources.ts
```diff
@@ -14,7 +14,7 @@ import { beforeEach } from "./beforeEach";
 beforeEach();
 
 test("should expand and collapse sources", async ({ page }, testInfo) => {
-  await toggleIncludeNodeModulesCheckbox(page, true);
+  await toggleExcludeNodeModulesCheckbox(page, true);
```
bvaughn, line 17: This looks like a mistake. (We were saying toggle-include: true and now we're saying toggle-exclude: true)

Mokshit06: Oops, yeah I missed that. Should be fixed now

packages/replay-next/src/hooks/useDebounce.ts
```diff
@@ -13,5 +13,10 @@ export function useDebounce<Value>(value: Value, delay: number = 100): Value {
     };
   }, [value, delay]);
 
+  return [debouncedValue, setDebouncedValue] as const;
```
bvaughn, line 16: I don't think we should make this change. The purpose of this hook is to debounce a value automatically. Calling the state updater function externally– it's no longer debounced and potentially introduces a race case:

1. New value A passed as param to `useDebouncedState` which starts a debounce timer
2. `setDebouncedValue` state updater is called from an external function with value B
3. Timer fires and overrides with "debounced" value A

I think maybe you want a combination of debounce+transition here. (I don't think we have a shared hook for this, but maybe we don't need it right away?) Something like this (I didn't test this)...

```js
const [valueForDisplay, setValueForDisplay] = useState("");
const [valueForSuspense, setValueForSuspense] = useState("");
const [isPending, startTransition] = useTransition();

useEffect(() => {
  const handler = setTimeout(() => {
    startTransition(() => {
      setDebouncedValue(value);
    });
  }, delay);

  return () => {
    clearTimeout(handler);
  };
}, [valueForSuspense]);
```

Then you'd pass `valueForSuspense` to your Suspense cache and maybe dim the UI or something else when `isPending` is true (to show you're updating results)

Mokshit06: I see, hmm but maybe we don't need to keep "search on enter" functionality either(?) in which case we can just keep the automatically debounced value, without having any need to update this value without skipping the timeout

bvaughn: > maybe we don't need to keep "search on enter" functionality either(?)

Totally agreed. We can drop that.

The addition of `useTransition` would be nice (but you can approximate that by e.g. `const isPending = queryForDisplay !== queryForSearch`)

packages/replay-next/src/hooks/useDebounce.ts
```diff
@@ -1,11 +1,13 @@
...
-export function useDebounce<Value>(value: Value, delay: number = 100): Value {
+export function useDebounce<Value>(value: Value, delay: number = 100) {
   const [debouncedValue, setDebouncedValue] = useState(value);
 
   useEffect(() => {
     const handler = setTimeout(() => {
-      setDebouncedValue(value);
+      startTransition(() => {
+        setDebouncedValue(value);
+      });
```
bvaughn, line 10: I was on the fence about having a `useDebounce` hook automatically wrap the update in a transition, but I guess this change is okay. There's a good chance that we're debouncing a value _to_ Suspend with it, so I guess it's fine :)

packages/shared/client/ReplayClient.ts
```diff
@@ -908,10 +908,16 @@ export class ReplayClient implements ReplayClientInterface {
...
+      useRegex,
+      wholeWord,
+      caseSensitive,
     }: {
       limit?: number;
       query: string;
       sourceIds?: string[];
+      useRegex?: boolean;
+      wholeWord?: boolean;
+      caseSensitive?: boolean;
```
bvaughn, line 920: tiny nit: I generally prefer to alpha-sort keyed properties like this.

packages/replay-next/components/search-files/ResultsList.tsx
```diff
@@ -18,15 +18,17 @@ const EMPTY_ARRAY = [] as SourceSearchResult[];
...
   query: string;
   streaming: StreamingSearchValue;
+  isPending: boolean;
 }) {
   const replayClient = useContext(ReplayClientContext);
 
   const { data, status, value: orderedResults = EMPTY_ARRAY } = useStreamingValue(streaming);
 
-  const isPending = status === STATUS_PENDING;
+  const isPending = status === STATUS_PENDING || searchIsPending;
```
bvaughn, line 31: D'oh this is silly of me. I forgot that we already have an is-pending signal from the Suspense cache. My apologies. We don't need the extra boolean then. Sorry! 🙇🏼

packages/replay-next/components/search-files/SearchFiles.tsx
```diff
@@ -2,57 +2,107 @@ import {
...
-      case "NumpadEnter":
-        startTransition(() => {
-          setQueryForSuspense(queryForDisplay);
-        });
-        break;
       case "Escape":
-        setQueryForDisplay(queryForSuspense);
+        setQueryForDisplay(queryForSearch);
         break;
     }
```
bvaughn, line 107: I think we should just remove this handler entirely. This "escape to cancel" behavior doesn't make sense anymore if we aren't explicitly applying the text with "Enter"

bvaughn (approved):
Small changes requested but otherwise LGTM

packages/replay-next/components/search-files/SearchFiles.tsx
Mokshit06: I agree. I have pushed these changes. Could you take a look? Should be good to merge

bvaughn:
The changes look fine to me.

Lots of e2e test failures though. I assume it's unrelated, but probably best to re-run them and make sure they pass before we merge.

Mokshit06:
Yeah these test failures are unrelated. They also don't occur when I am running the tests locally

bvaughn:
Okay then, free to merge whenever.

jonbell-lot23:
Please remember to tap me on the shoulder for visual updates like this, thanks!

PR #9847, opened 2023-10-27
Ensure all components use the same current point

hbenl:
- we had two current points in redux, which were updated at different times
- the `TimelineContextAdapter` passed on the current point and time with some delay (because they were copied in an effect)

This also fixes possible "tearing" when components use a `pauseId` and a `point` that don't match (i.e. the `pauseId` is for a different `point`).


bvaughn (approved):
Nice catch 👍🏼

PR #9849, opened 2023-10-30
Remove unnecessary timing bar override for selected network requests

markerikson:
This PR:

- Fixes the color of the timing section for failed network requests when selected

### Screenshots:

- Before (unselected):

image

- Before (selected) - note the green color for a failed 500 request:

image

- After (light theme) - note the timing bar is now red:

image

- After (dark theme):

image

I see that the original CSS behavior here was intentional, but I'm not sure why this was necessary.  We wouldn't want to hardcode `--success-color`, and I don't _think_ we need to change the timing bar color in general when it's selected.

### DIFF
--- a/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
+++ b/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
@@ -141,10 +141,6 @@
   min-width: 1px;
 }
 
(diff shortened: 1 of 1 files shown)


hbenl (approved):

bvaughn (approved):
Huh. Hadn't noticed that change. Agreed it doesn't make sense.

PR #9850, opened 2023-10-31
Keep all traces

jasonLaster:
This tells the backend to save all the honeycomb events so that we'll have an event in honeycomb for every protocol request we see in the viewer.

### DIFF
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -109,7 +109,7 @@ export type ExperimentalSettings = {
   disableRecordingAssetsInDatabase?: boolean;
   disableScanDataCache?: boolean;
   enableRoutines?: boolean;
-  keepAllTraces?: boolean;
+  sampleAllTraces?: boolean;
   listenForMetrics: boolean;
   profileWorkerThreads?: boolean;
   rerunRoutines?: boolean;
--- a/packages/shared/user-data/GraphQL/config.ts
+++ b/packages/shared/user-data/GraphQL/config.ts
@@ -80,9 +80,10 @@ export const config = {
     label: "Enable backend processing routines",
     legacyKey: "devtools.features.enableRoutines",
   },
-  backend_keepAllTraces: {
+  backend_sampleAllTraces: {
     defaultValue: Boolean(false),
     internalOnly: Boolean(true),
+    label: "Keep all Honeycomb events",
     legacyKey: "devtools.features.keepAllTraces",
   },
   backend_listenForMetrics: {
--- a/src/ui/actions/session.ts
+++ b/src/ui/actions/session.ts
@@ -197,7 +197,7 @@ export function createSocket(
         enableRoutines: userData.get("backend_enableRoutines"),
         rerunRoutines: userData.get("backend_rerunRoutines"),
         disableRecordingAssetsInDatabase: userData.get("backend_disableRecordingAssetsInDatabase"),
-        keepAllTraces: userData.get("backend_keepAllTraces"),
+        sampleAllTraces: userData.get("backend_sampleAllTraces"),
         disableIncrementalSnapshots: userData.get("backend_disableIncrementalSnapshots"),
         disableConcurrentControllerLoading: userData.get(
           "backend_disableConcurrentControllerLoading"
--- a/src/ui/components/ProtocolViewer/hooks/useHoneycombQueryLink.ts
+++ b/src/ui/components/ProtocolViewer/hooks/useHoneycombQueryLink.ts
@@ -15,15 +15,15 @@ export function useHoneycombQueryLink() {
         JSON.stringify({
           time_range: 3600,
           granularity: 0,
-          breakdowns: ["trace.trace_id"],
-          calculations: [{ op: "COUNT" }],
+          breakdowns: ["trace.trace_id", "name", "rpc.replayprotocol.command_id"],
+          calculations: [{ op: "AVG", column: "duration_ms" }],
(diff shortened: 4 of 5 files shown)


src/ui/actions/session.ts
```diff
@@ -197,7 +197,7 @@ export function createSocket(
         enableRoutines: userData.get("backend_enableRoutines"),
         rerunRoutines: userData.get("backend_rerunRoutines"),
         disableRecordingAssetsInDatabase: userData.get("backend_disableRecordingAssetsInDatabase"),
-        keepAllTraces: userData.get("backend_keepAllTraces"),
+        keepAllTraces: userData.get("sampleAllTraces"),
```
jcmorrow, line 200: this needs to be `sampleAllTraces`

packages/shared/user-data/GraphQL/config.ts
```diff
@@ -80,9 +80,10 @@ export const config = {
     label: "Enable backend processing routines",
     legacyKey: "devtools.features.enableRoutines",
   },
-  backend_keepAllTraces: {
+  sampleAllTraces: {
```
jcmorrow, line 83: I think the FE team prefers that we keep the `backend_` tag here.

src/ui/actions/session.ts
jcmorrow:
```suggestion
        sampleAllTraces: userData.get("backend_sampleAllTraces"),
```

packages/shared/user-data/GraphQL/config.ts
jcmorrow:
```suggestion
  backend_sampleAllTraces: {
```

jcmorrow (commented):
I think the suggestions batch here should set ya straight.

src/ui/components/ProtocolViewer/hooks/useHoneycombQueryLink.ts
```diff
@@ -15,15 +15,15 @@ export function useHoneycombQueryLink() {
...
+          breakdowns: ["trace.trace_id", "name", "rpc.replayprotocol.command_id"],
+          calculations: [{ op: "AVG", column: "duration_ms" }],
           filters: [
             { column: "name", op: "starts-with", value: "replayprotocol" },
             { column: "replay.session.id", op: "=", value: window.sessionId },
             { column: "rpc.replayprotocol.command_id", op: "=", value: request.id },
           ],
           filter_combination: "AND",
-          orders: [{ op: "COUNT", order: "descending" }],
+          orders: [{ op: "AVG", column: "duration_ms", order: "descending" }],
```
jcmorrow, line 26: I think this is probably fine, although I also think that in every case there will be exactly one result here, so sort should not matter.

packages/protocol/socket.ts
```diff
@@ -33,19 +33,19 @@ let gNextMessageId = 1;
...
 export type CommandResponse =
   | {
-      id: number;
-      result: unknown;
-      error?: undefined;
-    }
+    id: number;
+    result: unknown;
+    error?: undefined;
+  }
```
jcmorrow, line 39: I'm not sure what's going on here, seems like somehow your prettier and trunk's prettier disagree?

jcmorrow (approved):
AFAICT, this code looks fine in how it interacts with Honeycomb and the backend. It seems like it has failing tests and formatting issues, and I have no idea if that's expected or not, but probably should not merge without checking with FE as well (but I leave that to your discretion).

jasonLaster:
Thanks.

packages/shared/user-data/GraphQL/config.ts
```diff
@@ -80,9 +80,10 @@ export const config = {
     label: "Enable backend processing routines",
     legacyKey: "devtools.features.enableRoutines",
   },
-  backend_keepAllTraces: {
+  backend_sampleAllTraces: {
     defaultValue: Boolean(false),
     internalOnly: Boolean(true),
+    label: "Keep all Honeycomb events",
     legacyKey: "devtools.features.keepAllTraces",
```
bvaughn, line 87: I guess it isn't _quite_ what the high-risk flag was added for but this setting should probably be flagged as high-risk (since we don't want to it left on in most cases)?

```js
  highRisk: Boolean(true),
```

packages/protocol/socket.ts
jasonLaster: yeah - i think that's true.

packages/shared/user-data/GraphQL/config.ts
jasonLaster: This is a case where i'd like to be able to leave it on for a bit and see what is happening on the backend

src/ui/components/ProtocolViewer/hooks/useHoneycombQueryLink.ts
jasonLaster: yeah - not sure how what the best strategy is.

packages/shared/user-data/GraphQL/config.ts
bvaughn: Isn't that kind of circumstantial? (Would you want more people turning this on and leaving it on?)

The run book you shared says:
> Please do not keep this setting on more than you have to. We almost always operate right at the limit of our honeycomb budget and if we keep all internal traces we will go over our Honeycomb limits and start dropping valuable data. 🙏

To me, that sounds like a reasonable use of the high-risk flag UI.

PR #9855, opened 2023-11-02
Don't request hit points for lines with no hits

bvaughn:
I think this is the change the backend team requested us to make.

cc @jcmorrow

### DIFF
--- a/src/ui/components/Timeline/PreviewMarkers.tsx
+++ b/src/ui/components/Timeline/PreviewMarkers.tsx
@@ -1,4 +1,4 @@
-import type { PointDescription, SourceId } from "@replayio/protocol";
+import type { PointDescription, SourceId, TimeStampedPoint } from "@replayio/protocol";
 import { Suspense, useContext } from "react";
 
 import { binarySearch } from "protocol/utils";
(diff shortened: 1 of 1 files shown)


src/ui/components/Timeline/PreviewMarkers.tsx
```diff
@@ -75,25 +75,38 @@ function PreviewMarkersSuspends({
...
+    firstColumnHitCounts > 0
+  ) {
+    const tuple = hitPointsForLocationCache.read(
+      replayClient,
+      {
+        begin: focusRange ? focusRange.begin.point : "0",
+        end: focusRange ? focusRange.end.point : endpoint,
+      },
+      {
+        column: firstColumnWithHitCounts ?? 0,
```
hbenl, line 100: `?? 0` shouldn't be necessary anymore, right?

hbenl (approved):

src/ui/components/Timeline/PreviewMarkers.tsx
markerikson: Yeah, that's the part they specifically want us to avoid.

PR #9856, opened 2023-11-02
Remove ThreadFront

hbenl:


packages/e2e-tests/tests/breakpoints-04.test.ts
```diff
@@ -35,6 +36,7 @@ test(`breakpoints-04: catch, finally, generators, and async/await`, async ({
 
   async function resumeToBreakpoint(page: Page, lineNumber: number) {
     await addBreakpoint(page, { lineNumber, url: exampleKey });
+    await delay(200);
```
bvaughn, line 39: No objection to this but curious what prompted this change? Seems unrelated to ThreadPoint :)

bvaughn (approved):
Just a quick skim but I trust you on this. So long as e2e tests are happy (or at least no more unhappy than on main)

Congrats on killing the last of ThreadFront 🎊

packages/e2e-tests/tests/breakpoints-04.test.ts
hbenl: This is indeed unrelated and I'm still looking into it - the last successful run of breakpoints-04 was more than a week ago and one reason we didn't notice it is that most recordings of that test fail to upload, so most test runs don't contain a single recording of it.

hbenl: SCS-1540

src/ui/actions/timeline.ts
```diff
@@ -66,17 +63,14 @@ import {
...
 import type { UIStore, UIThunkAction } from "./index";
 
 const DEFAULT_FOCUS_WINDOW_PERCENTAGE = 0.3;
 export const MAX_FOCUS_REGION_DURATION = 60_000;
 export const MIN_FOCUS_REGION_DURATION = 0;
 
 export async function setupTimeline(store: UIStore) {
-  const dispatch = store.dispatch;
-
-  ThreadFront.on("paused", args => dispatch(onPaused(args)));
```
markerikson, line 78: :question: Just to check, where did this logic end up?

src/ui/actions/timeline.ts
```diff
@@ -256,11 +237,15 @@ export function seek({
...
     assert(executionPoint != null, `Could not find execution point for time ${time}`);
 
-    if (pauseId) {
-      ThreadFront.timeWarpToPause({ point: executionPoint, time, pauseId }, openSource);
-    } else {
-      ThreadFront.timeWarp(executionPoint, time, openSource);
-    }
+    dispatch(paused({ executionPoint, time, openSource }));
+
+    const focusWindow = replayClient.getCurrentFocusWindow();
```
markerikson, line 242: :speech_balloon: possibly answering my own question, looks like it ended up down here?

markerikson (approved):
No major comments. Had a couple questions about where some bits of logic moved around, but I think I followed those.  Would be nice to see the E2E tests passing as much as possible before we merge, but yeah, CONGRATS!

src/ui/actions/timeline.ts
hbenl: Exactly.

bvaughn:
Change change seems to have caused a regression and I think we should roll it back for now. After this commit, when the execution point changes– the app will re-render twice: once with a null `PauseId`– which will cause things to unmount unexpectedly (like the new React panel) and lose selected state and other memoized values.

PR #9860, opened 2023-11-03
Fix e2e test regressions related to new Elements panel

bvaughn:
### Change 1: Elements panel fetches 3 levels deep by default (instead of 2)

The old Elements panel used to pre-fetch and expand 3 levels deep in the DOM tree. The new one only fetched 2. This caused the `inspector-computed-01` e2e test to fail, since it relied on the Elements tree "remaining" expanded enough to see `id="maindiv"` between execution points.

This commit also removes several "workarounds" for the new Elements panel (selecting and expanding `<body>` manually) that are no longer necessary.

This e2e regression probably slipped through due to everything else being broken.

### Change 2: Elements search helper utility waits until Elements panel finishes loading

This avoids a race case where the test tried searching the Elements panel too soon and the action was ignored.

### DIFF
--- a/packages/e2e-tests/helpers/elements-panel.ts
+++ b/packages/e2e-tests/helpers/elements-panel.ts
@@ -1,6 +1,7 @@
 import { Locator, Page, expect } from "@playwright/test";
 import chalk from "chalk";
 
+import { clearText } from "./lexical";
 import { debugPrint, delay, waitFor } from "./utils";
 
 type ElementsListRowOptions = {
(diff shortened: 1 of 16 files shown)


packages/replay-next/components/elements/ElementsListData.ts
```diff
@@ -137,7 +137,7 @@ export class ElementsListData extends GenericListData<Item> {
     return index;
   }
 
-  async registerRootNodeId(id: ObjectId, numLevelsToLoad: number = 2) {
+  async registerRootNodeId(id: ObjectId, numLevelsToLoad: number = 3) {
```
bvaughn, line 140: This is the significant change.

hbenl (approved):

markerikson (approved):

packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -315,9 +316,18 @@ export async function searchElementsPanel(page: Page, searchText: string): Promi
...
-  await input.press("Enter");
+
+  await waitFor(async () => {
+    await input.press("Enter");
+
+    // If the Elements panel is still loading, the search won't be handled.
+    // A proxy for confirming that the search has been handled is that a results label will be rendered.
+    const resultsLabel = page.locator('[data-test-id="ElementsPanel-SearchResult"]');
+    await expect(await resultsLabel.count()).toEqual(1);
+  });
```
bvaughn, line 330: This is the key part of the React panel e2e test fix.

PR #9864, opened 2023-11-03
e2e test fixes

bvaughn:
- [x] FE-2030: Avoid race in Elements search helper util; allow search to finish
- [x] FE-2031: Add delay to e2e test to avoid querying DOM too soon after click
- [ ] FE-2032: Fix invalid test util helper params
        Note to self: This test is still failing but now on a different part.

### DIFF
--- a/packages/e2e-tests/tests/cypress-03_test-step-interactions.test.ts
+++ b/packages/e2e-tests/tests/cypress-03_test-step-interactions.test.ts
@@ -11,7 +11,7 @@ import {
   openCypressTestPanel,
 } from "../helpers/testsuites";
 import { waitForTimelineAdvanced } from "../helpers/timeline";
-import { getByTestName, waitFor } from "../helpers/utils";
+import { delay, getByTestName, waitFor } from "../helpers/utils";
 import test, { expect } from "../testFixtureCloneRecording";
 
 test.use({ exampleKey: "flake/adding-spec.ts" });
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/tests/cypress-03_test-step-interactions.test.ts
```diff
@@ -153,6 +153,8 @@ test("cypress-03: Test Step interactions", async ({ pageWithMeta: { page, record
   });
 
   await steps.nth(9).click();
+  await delay(100);
```
markerikson, line 156: 💬 I'm a bit surprised that a delay would help here.  Either I'd expect the other step clicks to have a similar issue, or that this would be okay.

Also, from the triage, it looked like it was having trouble finding a `KeyValue-Header` for "Selected" even though that appeared to be visible on screen.

If this actually does fix things, great, I'm all for that :) but based on what I saw I wouldn't expect this to resolve the failures.

bvaughn: Pretty sure it was still reading DOM for the the wrong/previous thing. Reliably failed before for me and reliably passed after this change.

markerikson: Ah, okay, that makes more sense! Clicking, but the _old_ step details are still there and there's a mismatch.

bvaughn:
I'm not sure why cypress-04 is still failing. It reliably passes for me locally now (and it failed before) but maybe I'll land the other two fixes and follow up on Monday.

PR #9868, opened 2023-11-06
Refactor the code for seeking to a new point and improve UI responsiveness

hbenl:
UI responsiveness improvements:
- update the timeline immediately when seeking but dim the current time indicator until we have a pauseId
- show a dimmed badge for the pause info panel if we have a `pausePreviewLocation` (so we know the pause will have frames) but no frames yet
- I found a couple of places where we could set `pausePreviewLocation` (which is now done by passing the location to `seek()`) but didn't


jasonLaster:
I love this change! 

Could we add some assertions to our e2e tests to ensure we don't regress because these states are easy to break.

packages/replay-next/components/sources/useSourceContextMenu.tsx
```diff
@@ -144,7 +144,11 @@ function FastForwardButton({
 
   const fastForward = () => {
     assert(fastForwardToExecutionPoint != null);
-    update(fastForwardToExecutionPoint.time, fastForwardToExecutionPoint.point, false);
+    update(fastForwardToExecutionPoint.time, fastForwardToExecutionPoint.point, false, {
+      sourceId,
+      line: lineNumber,
+      column: lineHitCounts?.firstBreakableColumnIndex ?? 0,
```
bvaughn, line 150: How does this fallback (`?? 0`) relate to FE-2015?

src/ui/components/Toolbar.tsx
```diff
@@ -275,7 +275,7 @@ function ToolbarButton({
   label: string;
   name: PrimaryPanelName;
   onClick: (name: PrimaryPanelName) => void;
-  showBadge?: boolean;
+  showBadge?: boolean | "dimmed";
```
bvaughn, line 278: This is kid of confusing. Let's add a new prop (`dimmed?: boolean`) for this?

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -183,12 +191,15 @@ const pauseSlice = createSlice({
     pauseHistoryIncremented(state) {
       state.pauseHistoryIndex++;
     },
-    pauseCreationFailed(state, action: PayloadAction<string>) {
-      const executionPoint = action.payload;
+    pauseCreationFailed(state) {
+      const lastPause = state.pauseHistory[state.pauseHistoryIndex];
       Object.assign(state, {
-        id: undefined,
+        seekState: "paused",
```
bvaughn, line 197: Is this the right state after pause failure?

packages/replay-next/components/sources/useSourceContextMenu.tsx
hbenl: We're not treating this location as a breakable location (so we won't request hit points for it). But actually this method should never be called if `lineHitCounts` is `null`, so I'll replace the fallback with an assert to make that clear (there's already `assert(fastForwardToExecutionPoint != null);` in there and `fastForwardToExecutionPoint` would be `null` if `lineHitCounts` was)

src/devtools/client/debugger/src/reducers/pause.ts
hbenl: We're going back to the last successfully created pause (`state.pauseHistory` only contains successfully created pauses, i.e. every entry has a `pauseId`). I thought that was the right thing to do since we don't really handle failed pauses anywhere in the UI, so staying at a failed pause may lead to bugs. But perhaps this (staying at the failed pause) is what we want anyway and we should just start fixing any bugs we may see with that?

bvaughn: I think that's probably right, yeah. Would be confusing at best to have an execution point and pause that aren't in sync, and maybe could cause command failures if we used the pause?

src/ui/actions/timeline.ts
```diff
@@ -207,45 +219,82 @@ export function seek({
...
+      } catch (e) {
+        console.error(e);
+        dispatch(pauseCreationFailed());
+        return;
       }
     }
 
-    assert(executionPoint != null, `Could not find execution point for time ${time}`);
+    assert(executionPoint);
+    dispatch(paused({ executionPoint, time, id: pauseId }));
```
bvaughn, line 272: Hey look! We're using the `pauseId` param again 🥳

bvaughn (approved):

src/ui/actions/timeline.ts
hbenl: Yeah, previously (after my ThreadFront removal) `seek()` would always get the `pauseId` for the given `executionPoint` from `pauseIdCache`. Usually we only have one `pauseId` for any given `executionPoint` (so this shouldn't make a difference), but there is no guarantee for that, so it's good to reconnect the `pauseId` param here.

PR #9869, opened 2023-11-06
Revert "Remove ThreadFront (#9856)"

bvaughn:
This reverts commit c4c5ba3464caca261bec65b0268680cd2cc02198 because of a regression it caused in the new React DevTools panel (see FE-2038)


bvaughn:
Reverted cleanly so I'm going to land this once CI passes.

PR #9871, opened 2023-11-07
Remove ThreadFront

hbenl:
Same as #9856 except for [this additional commit](https://github.com/replayio/devtools/pull/9871/commits/a205d08a4b184560ad8db0018c1c9b63bafb9044)


src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -260,7 +257,7 @@ export function ReactDevtoolsPanel() {
       // Inject the RDT backend and prefetch node IDs
       wall.setUpRDTInternalsForCurrentPause();
     }
-  }, [currentPoint, wall]);
+  }, [currentPoint, currentTime, wall]);
```
bvaughn, line 260: This change seems unnecessary/mistaken.
```suggestion
  }, [currentPoint, wall]);
```

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -252,6 +270,10 @@ export function getPauseHistory(state: UIState) {
 export function getPauseHistoryIndex(state: UIState) {
   return state.pause.pauseHistoryIndex;
 }
+// for components that should not switch to a new point before its `pauseId` is available
+export function getMostRecentPause(state: UIState): PauseHistoryData | undefined {
+  return state.pause.pauseHistory[state.pause.pauseHistoryIndex];
+}
```
bvaughn, line 276: I'd like to better understand this specific change. Maybe we can chat about it during sync today (or now if you're free)?

I'm a bit concerned about us relying on a selector like this because it seems easy to overlook/forget. It's also not obvious to me when or why a component would want to consume the _current_ execution point but the _previous_ pause. Seems like those would need to be kept in sync.

So looking at RDT specifically, maybe we should have a `prevRef` that stores both the previous (non-null?) execution point and pause id (so the two are always in sync, internally)? am I overlooking something?

PR #9874, opened 2023-11-07
Add a way to show multiple views for a test team

jaril:


bvaughn (commented):
The indentation and padding for this looks off:
Screen Shot 2023-11-07 at 1 29 23 PM

Not indented enough (on the left) and too much padding on the bottom.

I also wonder if we should be showing the same "selected" style for both the Workspace name _and_ the sub-section. 

This looks a _little_ better to me:
Screen Shot 2023-11-07 at 1 31 39 PM

changes to top row: no bold

changes to second row:
```css
    padding-left: 1.5rem;
    padding-bottom: 0.25rem;
```

PR #9875, opened 2023-11-07
Fetch test run details along with recordings

ryanjduffy:
Fetch test run along with recordings so that deep linking to a test run that might not have been returned in the list can still show the details along with the recordings. This will help when we limit the number of test runs returned for a team.

Fixes SCS-1516

### DIFF
--- a/packages/shared/graphql/generated/GetTestRunRecordings.ts
+++ b/packages/shared/graphql/generated/GetTestRunRecordings.ts
@@ -11,6 +11,32 @@ export interface GetTestRunRecordings_node_Recording {
   __typename: "Recording";
 }
 
+export interface GetTestRunRecordings_node_Workspace_testRuns_edges_node_results_counts {
+  __typename: "TestRunStats";
+  failed: number;
+  flaky: number;
+  passed: number;
+}
+
+export interface GetTestRunRecordings_node_Workspace_testRuns_edges_node_results {
+  __typename: "TestRunResults";
+  counts: GetTestRunRecordings_node_Workspace_testRuns_edges_node_results_counts;
+}
+
+export interface GetTestRunRecordings_node_Workspace_testRuns_edges_node_source {
+  __typename: "TestRunSource";
+  commitId: string | null;
+  commitTitle: string | null;
+  groupLabel: string | null;
+  isPrimaryBranch: boolean | null;
+  branchName: string | null;
+  prNumber: number | null;
+  prTitle: string | null;
+  repository: string | null;
+  triggerUrl: string | null;
+  user: string | null;
+}
+
 export interface GetTestRunRecordings_node_Workspace_testRuns_edges_node_tests_recordings_comments_user {
   __typename: "User";
   id: string;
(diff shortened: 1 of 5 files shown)


bvaughn:
I wonder if this change is worth doing. Selecting nothing by default (if the run isn't in the list of loaded runs) seems almost better than dispatching details for something on the right that has no corresponding selected row on the left. (I don't feel _strongly_ about this but I wonder if we're fixing something that isn't really broken?)

ryanjduffy:
> I wonder if this change is worth doing. Selecting nothing by default (if the run isn't in the list of loaded runs) seems almost better than dispatching details for something on the right that has no corresponding selected row on the left. (I don't feel _strongly_ about this but I wonder if we're fixing something that isn't really broken?)

It will become broken if we start limiting the number of test runs we return for a workspace which we need to do for our more prolific test suite customers

ryanjduffy:
@bvaughn - can you review this or let me know who else should weigh in on it? Thanks!

bvaughn:
> It will become broken if we start limiting the number of test runs we return for a workspace which we need to do for our more prolific test suite customers

Why would that be any more broken than what currently happens? I don't understand. In either case you'd try to open a deep link to a run, and see more recent runs list instead. That seems okay to me?

> @bvaughn - can you review this or let me know who else should weigh in on it? Thanks!

I didn't get an email notification for your previous comment, sorry. I was reviewing it by saying "I don't think we need to do this". Seems like you think we do, so maybe I'm missing something?

bvaughn:
Thanks for the Loom.

FWIW I think we were both on the same page previously about how things work today (blank details panel). I agree that's not great.

My two concerns with this approach though were:
* Given the pattern of column 2 being details for the selected thing in column 1, I'm concerned that it's pretty confusing to have a disconnect between those two panels (which this PR makes possible). I think you'd argue that it's better than what we currently do. I'm not sure.<sup>1</sup>
* Clicking a "deep link" and seeing nothing in the 2nd column is going to eventually happen regardless (after two weeks, recordings will be deleted and so links will stop working at that point).

<sup>1</sup>  I think maybe there's some way to change the 1st column to show a "fake" entry (at the bottom maybe?) so there's _something_ selected, and this could address my concern.

ryanjduffy:
> Thanks for the Loom.
> 
> FWIW I think we were both on the same page previously about how things work today (blank details panel). I agree that's not great.
> 
> My two concerns with this approach though were:
> 
> * Given the pattern of column 2 being details for the selected thing in column 1, I'm concerned that it's pretty confusing to have a disconnect between those two panels (which this PR makes possible). I think you'd argue that it's better than what we currently do. I'm not sure.1

I agree they should be in sync but showing nothing on the right is bad when we have the record to show. 

> * Clicking a "deep link" and seeing nothing in the 2nd column is going to eventually happen regardless (after two weeks, recordings will be deleted and so links will stop working at that point).

That is a slightly different issue. We should message that the test run isn't available after it's rolled out of the retention period.

This PR (and the limiting backend PR) is trying to manage that we have customers producing > 300 test runs a day so we'd have to return > 4000 test run records in order to have them all in memory to render the details pane for a valid pane.

> 
> 1 I think maybe there's some way to change the 1st column to show a "fake" entry (at the bottom maybe?) so there's _something_ selected, and this could address my concern.

Maybe but I think this PR is an incremental improvement we should consider now.

bvaughn (approved):

PR #9877, opened 2023-11-08
Add e2e test for RDT panel selection when seeking

hbenl:
I wrote this test for the new RDT panel because that was the one showing the regression (FE-2038) and because we will get rid of the old one eventually.

### DIFF
--- a/packages/e2e-tests/helpers/new-react-devtools-panel.ts
+++ b/packages/e2e-tests/helpers/new-react-devtools-panel.ts
@@ -1,0 +1,41 @@
+import { Locator, Page, expect } from "@playwright/test";
+
+import { openConsolePanel, warpToMessage } from "./console-panel";
+import { openReactDevtoolsPanel as openReactDevtoolsPanelWithoutWorkaround } from "./react-devtools-panel";
+import { waitFor } from "./utils";
+
+export function getReactDevToolsPanel(page: Page) {
+  return page.locator('[data-test-id="ReactDevToolsPanel"]');
+}
+
+export function getReactComponents(page: Page) {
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/helpers/new-react-devtools-panel.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function jumpToMessageAndCheckComponents(
+  page: Page,
+  message: string,
+  componentCount: number
+) {
+  await openConsolePanel(page);
+  await warpToMessage(page, message);
+  await openReactDevtoolsPanel(page);
+  await waitForReactComponentCount(page, componentCount);
+}
```
bvaughn, line 29: nit (feel free to ignore just thinking out loud): This seems like it might be too test-specific to be in the generic helpers file

packages/e2e-tests/tests/react_devtools-04-seeking.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // Seek to the first console message and check that the initial tree has 3 components
+  await jumpToMessageAndCheckComponents(page, "Initial list", 3);
+
+  const initialTimePercent = await getTimelineCurrentPercent(page);
+
+  // Select the root component
+  const firstComponent = getReactComponents(page).nth(0);
+  await firstComponent.click();
+  await expect(firstComponent).toHaveAttribute("data-selected", "true");
```
bvaughn, line 30: nit: We could use either a `getSelectedReactComponent` helper or a `verifyReactComponentIsSelected` helper for this.

bvaughn (approved):
> I wrote this test for the new RDT panel because...we will get rid of the old one eventually.

Love this idea.

PR #9879, opened 2023-11-09
Align redux, network monitor, and console rows

jonbell-lot23:
Old:
image

* Row heights are misaligned
* Row borders are inconsistent

New, with these fixes:
image

### DIFF
--- a/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
+++ b/src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
@@ -5,7 +5,7 @@
   flex-direction: row;
   align-items: center;
   background-color: var(--body-bgcolor);
-  border-bottom: 1px solid var(--theme-toolbar-background);
+  border-bottom: 1px solid var(--listitem-row-divider);
   padding: 0 1ch;
 }
 .Row {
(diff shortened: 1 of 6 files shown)


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-974/design-pass-make-sure-network-monitor-console-and-redux-rows-are

jasonLaster:
Much nicer

src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
```diff
@@ -124,7 +124,7 @@
 
 .TimingContainer {
   position: relative;
-  height: 0.75rem;
+  height: 0.8rem;
```
bvaughn, line 127: IMO 0.8 rem is not a very useful value. Might as well use pixels if we aren't sticking to the same increments.

bvaughn (approved):

src/ui/components/NetworkMonitor/NetworkMonitorListRow.module.css
jonbell-lot23: We might want to have a chat about this to get on the same page. I definitely optimise for the final visual, and .25 sometimes causes a misalignment. In this case, .75 was causing the top and bottom margin to feel optically uneven, which lead to us using 27 height (rather than the 26 I was aiming for standardising on)

I'm happy to use whatever units we want to achieve this effect. Are you saying you'd prefer pixels when I need to do this fine-tuning pixel work?

bvaughn: Yes.

PR #9880, opened 2023-11-09
Increased color contrast when buttons overlap selected rows

jonbell-lot23:
We have a long-standing issue where our blue time travel button disappears when it overlaps on a selected row. I've been playing around with ways to make this work better. Not just here, but how the colors can be applied across the system.

Current approach on left, proposed on right:
image
This is a no-brainer.

image
And this is a nice standardization/cleanup in the library.

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -39,8 +39,6 @@
   --secondary-accent-hover: #d72451;
   --secondary-accent-stroke: #f02d5e;
 
-  --tertiary-accent: var(--green-60);
-
   /* used in our codebase, to be refactored */
   --blue-30: #75baff;
   --blue-40: #45a1ff;
@@ -192,7 +190,7 @@
   --theme-text-field-bgcolor: var(--theme-base-85);
   --theme-text-field-color: #fff;
   --theme-text-field-placeholder-color: #babbbc;
-  --team-row-active: var(--theme-text-field-bgcolor);
+  --team-row-active: var(--primary-accent-dimmed);
   --tooltip-bgcolor: var(--theme-base-100);
   --tooltip-color: #fff;
   --tooltip-border: 1px solid rgba(43, 46, 49, 1);
@@ -221,7 +219,6 @@
   --background-color-primary-button: #02acfd;
   --background-color-resize-handle: #242e3d;
   --background-color-secondary-button: #f02d5e;
-  --background-color-tertiary-button: var(--tertiary-accent);
   --background-color-source-search: #25364e;
   --background-color-warning: #42381f;
   --background-color-high-risk-setting: #310f21;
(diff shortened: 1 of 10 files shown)


jasonLaster:
I like this better. Also winter what it would look like with the lighter blue row background and the primary blue button.

jonbell-lot23:
Following up on discord comment from yesterday: yeah, I like the green too but I also think we can stick with blue for now and keep the green in our back pocket. It's coming (probably)! Doesn't need to be rushed out now though.

bvaughn:
I like this change but it only update some parts of our app, so it causes inconsistencies. For example the Network panel looks like this now
Screen Shot 2023-11-09 at 4 46 13 PM

but the Elements and React panels still look like this
Screen Shot 2023-11-09 at 4 47 20 PM
Screen Shot 2023-11-09 at 4 47 25 PM

packages/replay-next/pages/variables.css
```diff
@@ -709,7 +716,7 @@
   --theme-text-field-bgcolor: #f6f6f6;
   --theme-text-field-color: var(--body-color);
   --theme-text-field-placeholder-color: #a9a9b1;
-  --team-row-active: #3f434a;
+  --team-row-active: #163e58; /* the dark version of --primary-accent-dimmed */
```
bvaughn, line 719: btw
```suggestion
  --team-row-active: var(--primary-accent-dimmed);
```

bvaughn:
Noticed light theme has a contrast issue with the duration bubble
Screen Shot 2023-11-09 at 4 50 33 PM

packages/replay-next/pages/variables.css
jonbell-lot23: Usually variables are great. But in this case it's the left-hand column in the library, which always stays dark. Using the variable here means it uses the wrong color when in light theme.

jonbell-lot23:
Oh good -- if you think it works fine in those cases, I'll add them here. I was going to address that in a follow-up because I assumed the HTML colors would cause more issues and edge cases.

(Philosophically there are two ways to look at this -- find everything related and put it into a single PR versus stage things out, but we don't have to get into that here, happy to put this into one)

packages/replay-next/pages/variables.css
bvaughn: Oh I see. Hmm 🤔 That's pretty easy to get mixed up.

jonbell-lot23: Indeed. I run against this every. single. time.

jonbell-lot23:
@bvaughn I am having trouble figuring out what file changes the selected color in the react devtools. I tried deleting everything from variables.css but the blue still comes through, meaning it's not using variables. Then I looked in files like src/ui/components/SecondaryToolbox/react-devtools/components/ReactDevToolsListItem.module.css but I couldn't manage to find the right spot.

jonbell-lot23:
> I like this change but it only update some parts of our app, so it causes inconsistencies. 

Updated these:

image

image

> Noticed light theme has a contrast issue with the duration bubble.

Updated these:

image

image

ryanjduffy (approved):
:+1: Adding some snapshot tests for these changes would be nice too

PR #9881, opened 2023-11-09
Setup E2E fuzz testing for React Devtools (FE-1857)

Mokshit06:
This PR:
- Adds a fuzz test for react devtools for comparing a tree created using the operations generated by the backend to the ones generated by the "real" RDT
- Stores the serialized replay store as a snapshot to be notified in the future if we do anything that causes our structure to change
- It also copies over an implementation of `Store` class from backend , only keeping the functions required for serializing store

### DIFF
--- a/packages/e2e-tests/tests/react_devtoos-04-fuzz-test.test.ts
+++ b/packages/e2e-tests/tests/react_devtoos-04-fuzz-test.test.ts
@@ -1,0 +1,52 @@
+import { openDevToolsTab, startTest } from "../helpers";
+import { warpToMessage } from "../helpers/console-panel";
+import { Store, printStore } from "../helpers/rdt-store";
+import { getReactComponents, openReactDevtoolsPanel } from "../helpers/react-devtools-panel";
+import { delay } from "../helpers/utils";
+import test, { expect } from "../testFixtureCloneRecording";
+
+test.use({ exampleKey: "rdt-react-versions/dist/index.html" });
+
+const versions = ["15.7.0", "16.14.0", "17.0.2", "18.2.0"];
+const messagesToCheck = versions.flatMap(version => [
+  `[${version}] Added an entry`,
+  `[${version}] Removed an entry`,
+]);
+
+test("react_devtools 04: fuzz testing", async ({ pageWithMeta: { page, recordingId } }) => {
+  const queryParams = new URLSearchParams();
+  // // Force this test to always re-run the RDT (and other) routines
+  // // See pref names in packages/shared/user-data/GraphQL/config.ts
+  queryParams.set("features", "backend_rerunRoutines");
+
+  await startTest(page, recordingId, undefined, queryParams);
+  await openDevToolsTab(page);
+
+  for (const message of messagesToCheck) {
+    await warpToMessage(page, message);
+    await openReactDevtoolsPanel(page);
+
+    await delay(500);
+    // wait for components to load (setup is complete)
+    await getReactComponents(page).first().waitFor();
+
+    const [serializedStore, annotations] = (await page.evaluate(`(() => {
+      ${printStore}
+
+      return (${evaluateStore})()
+    })()`)) as [string, any];
+
+    const replayStore = new Store();
+    for (const annotation of annotations) {
+      replayStore.onBridgeOperations(annotation.contents.payload);
+    }
+    const serializedReplayStore = printStore(replayStore);
+
+    expect(serializedStore).toBe(serializedReplayStore);
+    expect(serializedReplayStore).toMatchSnapshot({ name: `fuzz test: ${message}` });
+  }
+});
+
+function evaluateStore() {
+  return [printStore((window as any).app.rdtStore), (window as any).app.rdtAnnotations];
+}
--- a/packages/e2e-tests/tests/react_devtoos-04-fuzz-test.test.ts
+++ b/packages/e2e-tests/tests/react_devtoos-04-fuzz-test.test.ts
@@ -1,10 +1,7 @@
-import chalk from "chalk";
-
-import { printOperationsArray } from "../../../src/ui/components/SecondaryToolbox/react-devtools/printOperations";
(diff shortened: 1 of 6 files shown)


bvaughn (commented):
Do we really need to copy the whole `Store` that feels like it's going to be a hassle to keep in sync. Could we just update the react-devtools-inline package to export the `Store` class? (Maybe I'm overlooking something)

Mokshit06:
`react-devtools-inline` package does export a `createStore` method from frontend entrypoint, but it also accesses `window`, so couldn't be imported from there

bvaughn:
> `react-devtools-inline` package does export a `createStore` method from frontend entrypoint, but it also accesses `window`, so couldn't be imported from there

I wasn't suggesting using the `createStore` method, but updating (our fork of) the `react-devtools-inline` package to export the `Store` class itself. We'd probably need to add a new entry point (maybe `replay`?)

Mokshit06:
Ah, yeah we can do that for sure. I think that `@replayio/react-devtools-inline` is published off of https://github.com/replayio/react/pull/7/, but it isn't merged, so how should I go about modifying it? Maybe this needs to be merged first? @markerikson

markerikson:
First item is that I made some specific edits to that `store.ts` file when I copy-pasted it from the React repo into our backend for use in the routine.  I don't remember _what_ I changed specifically.  I _think_ I at least made one commit that was _just_ "copy the store file", and another that was "start tweaking the logic", so we should be able to look at the backend Git history and see what I changed that way.

I do know that beyond changes to the actual `Store` class, I definitely altered the `printStore()` method in that file to add more detail to the text output, for use in textual tree diffing.

I'm honestly not worried about duplication of code right now.  I'd rather focus on getting the test put together (and also figuring out what is different with the operations tree in this example).

bvaughn:
> I'm honestly not worried about duplication of code right now. I'd rather focus on getting the test put together (and also figuring out what is different with the operations tree in this example).

I appreciate that you're focused on that (and I think that's important) but I would like for us to avoid copying non-trivial chunks of code between repos as well.

@Mokshit06 our fork of React DevTools is published to NPM as [`@replayio/react-devtools-inline`](https://www.npmjs.com/package/@replayio/react-devtools-inline) which I can give you access to (if you'd like) or you could tag one of us to publish an update.

markerikson:
@bvaughn one significant problem is that I converted all of `store.ts` to TS when I ported it.  The `@replayio/react-devtools-inline` package is still Flow.  That will complicate the re-porting work.

bvaughn:
> @bvaughn one significant problem is that I converted all of `store.ts` to TS when I ported it. The `@replayio/react-devtools-inline` package is still Flow. That will complicate the re-porting work.

@markerikson Aren't we talking about just exporting this class?

https://github.com/replayio/react/blob/feature/replay-react-devtools/packages/react-devtools-shared/src/devtools/store.js

markerikson:
Not exactly. That's what I was saying earlier.  Yes, it's that file, but the copy of it that's in the backend was A) ported to TS, and B) has _some_ modifications to both the `Store` class and the `printStore()` function.  It's no longer identical to what's in that `react-devtools-inline` package.

bvaughn:
> Not exactly. That's what I was saying earlier. Yes, it's that file, but the copy of it that's in the backend was A) ported to TS, and B) has some modifications to both the Store class and the printStore() function. It's no longer identical to what's in that react-devtools-inline package.

Isn't that the purpose of these fuzz tests? To make sure that our modifications don't diverge from the "real" RDT package.

markerikson:
Ah, good point. We probably do want to try to use the original `Store` class for the comparison in that case.

Mokshit06:
I see. @bvaughn I've pushed a commit to https://github.com/replayio/react/pull/7 for making `Store` and `printStore` public. You can publish an update or give me the access for it (whichever is easier for you).

packages/e2e-tests/tests/react_devtoos-04-fuzz-test.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const [serializedStore, annotations] = (await page.evaluate(`(() => {
+      ${printStore}
+
+      return (${evaluateStore})()
+    })()`)) as [string, any];
+
+    const replayStore = new Store();
+    for (const annotation of annotations) {
+      replayStore.onBridgeOperations(annotation.contents.payload);
```
markerikson, line 41: :hand: I don't think this is quite testing what we wanted to (and this may actually explain the listed mismatch in FE-2050).

This is _currently_ extracting the routine-generated annotations containing operations, and running those through an RDT `Store` instance.  However, that means we're _not_ asking "what's the difference between what the _routine_ generated and what's _actually_ in the page?".  We're instead asking "if we take the routine-generated output and calculate it twice, is the result the same?", which isn't helpful.  The real goal is to do some checking if what the routine generated is actually correct.  

Also, I think the comparison logic here is flawed.

This is fetching _all_ annotations throughout the _entire_ recording, and feeding _all_ of them into the store.  But, that's going to give us the state of the component tree as it exists at the _end_ of the recording.

However, we're paused at a specific point in time, and even if we were trying to compare based on the annotations, we'd need to compare _just_ the annotations up to this execution point.

Conceptually, what I think we _actually_ want to do here is:

- Expose a new global method from the app code that will make an evaluation call and force "RDT in the paused browser" to calculate an entire operations tree from scratch that describes the _currrent_ React fiber tree as of that pause
- Serialize that operations data from the in-app `Pause` eval, back to the E2E test via a Playwright eval (similar to what you've got on line 33
- Create a fresh RDT `Store` and feed in that "the whole tree as it exists _now_" operations array
- Do the same `printStore(window.app.rdtStore)` eval that you have below to see what the _app_'s store instance thinks the tree is, based on the routine data
- Compare those two outputs

bvaughn:
This is an old PR with a lot of conflicts. I assume it's been abandoned, so I'm going to close it. Rebase and re-open if we want to reconsider.

PR #9883, opened 2023-11-09
Set pause.executionPoint to null instead of undefined

hbenl:
I wish Typescript could typecheck `Object.assign()`...

### DIFF
--- a/src/devtools/client/debugger/src/reducers/pause.ts
+++ b/src/devtools/client/debugger/src/reducers/pause.ts
@@ -146,14 +146,14 @@ const pauseSlice = createSlice({
         location?: Location;
       }>
     ) {
-      const { executionPoint, time } = action.payload;
+      const { executionPoint, location, time } = action.payload;
       Object.assign(state, {
         seekState: executionPoint ? "create-pause" : "find-point",
         id: undefined,
-        executionPoint,
+        executionPoint: executionPoint ?? null,
         time,
         pausePreviewLocation: location ?? null,
-      });
+      } satisfies Partial<PauseState>);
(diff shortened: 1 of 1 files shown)


bvaughn (approved):

src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -150,7 +150,7 @@ const pauseSlice = createSlice({
       Object.assign(state, {
         seekState: executionPoint ? "create-pause" : "find-point",
         id: undefined,
-        executionPoint,
+        executionPoint: executionPoint ?? null,
         time,
         pausePreviewLocation: location ?? null,
       });
```
bvaughn, line 156: You could add a `satisfies` here

bvaughn: Screen Shot 2023-11-09 at 11 16 00 AM

hbenl: Unfortunately `satisfies Partial<PauseState>` won't catch the `undefined` (because that's always allowed for `Partial`). BUT it did show me another pretty stupid bug, so thanks for the suggestion!

hbenl: Yeah, if you say `satisfies PauseState` then it will catch the `undefined`... but then it complains that not all properties are defined (once you have fixed any other bugs).

bvaughn: Yeah, that's an awkward gap. I guess we don't need to be using `Object.assign` here.

PR #9885, opened 2023-11-09
Handle potential missing point stack execution points

markerikson:
This PR:

- bumps the protocol package to 0.63 to pick up the types changes for `getPointStack`
- Tweaks the logic in some caches to handle `pointStackFrame.point` becoming optional

I _think_, based on https://github.com/replayio/backend/pull/8924 and https://github.com/replayio/backend/pull/8925 , that this will only ever occur if the frame just has an `EnterFrame` point and no meaningful frame steps.  However, I _don't_ know often that will occur in practice.

The code compiles, and I _think_ the logic tweaks make sense reading it.

### DIFF
--- a/packages/replay-next/src/suspense/ResumeTargetCache.ts
+++ b/packages/replay-next/src/suspense/ResumeTargetCache.ts
@@ -73,7 +73,16 @@ export const resumeTargetCache = createCache<
 
       if (frameIndex > 0) {
         const pointStack = await pointStackCache.readAsync(0, frameIndex, replayClient, pausePoint);
-        currentPoint = pointStack[frameIndex].point.point;
+        const frame = pointStack[frameIndex];
+        if (!frame.point) {
+          // Avoid stepping in the top frame - we need to step in the _current_ frame.
+          // Note that it's unlikely that we _will_ hit this case.
+          // Per Josh, the most likely reason a stack frame would _not_ have an execution point
+          // is if there's no actual code inside, such as `class B extends A` where B has no constructor.
+          // We _could_ look for the next frame that _does_ have an execution point instead.
+          return;
+        }
+        currentPoint = frame.point.point;
       }
     }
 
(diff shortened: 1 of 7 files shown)


packages/replay-next/src/suspense/ResumeTargetCache.ts
```diff
@@ -73,7 +73,10 @@ export const resumeTargetCache = createCache<
 
       if (frameIndex > 0) {
         const pointStack = await pointStackCache.readAsync(0, frameIndex, replayClient, pausePoint);
-        currentPoint = pointStack[frameIndex].point.point;
+        const frame = pointStack[frameIndex];
+        if (frame.point) {
+          currentPoint = frame.point.point;
+        }
```
hbenl, line 79: We should `return;` if `frame.point` is not set - otherwise the debugger will step in the top frame rather than the selected frame.

hbenl (changes requested):

packages/replay-next/src/suspense/ResumeTargetCache.ts
markerikson: Okay, yeah, that's what I was wondering.

hbenl (approved):

markerikson:
Talked with Josh about this.

One example where this _could_ happen is `class B extends class A`, where `A` has a constructor but `B` does not.  (This occurs in our `doc_rr_error.html` E2E example.)  In that case, there's _technically_ a constructor for `B` on the stack but it has no code associated with it, so you can't pause there.

In practice, any "top" or "bottom" stack frame (no matter which way you visualize the stack in your head) should have an execution point because there should be actual code involved.

For the Redux and React usage cases, we should never run into that scenario, because we're looking at user code frames that should have actual code.

markerikson:
Add further comments clarifying the situation here.

bvaughn:
> One example where this _could_ happen is `class B extends class A`, where `A` has a constructor but `B` does not. (This occurs in our `doc_rr_error.html` E2E example.) In that case, there's _technically_ a constructor for `B` on the stack but it has no code associated with it, so you can't pause there.

This is the kind of thing that's great to capture in unit tests somewhere. I don't have a ton of context (just kind of drive by skimming this thing) but I wonder if there's an opportunity for us to add a test or two here?

PR #9886, opened 2023-11-09
Tighten our UI

jonbell-lot23:
We've always used w-2 to space our elements. This change moves to w-1 for a tighter and more professional appearance. 

Before: (look at the vertical gaps between columns, and the horizontal gaps around the video)
image

After:
image

I've also addressed the timeline to put things on a more consistent grid:

Here's the inconsistent alignment before:
image

Here's how it looks now:
image

### DIFF
--- a/src/ui/components/DevTools.tsx
+++ b/src/ui/components/DevTools.tsx
@@ -106,7 +106,7 @@ function Body() {
   }, [sidePanelCollapsed]);
 
   return (
-    <div className="vertical-panels pr-2">
+    <div className="vertical-panels pr-1">
       <div className="flex h-full flex-row overflow-hidden bg-chrome">
         <Toolbar />
         <PanelGroup autoSaveId="DevTools-horizontal" className="split-box" direction="horizontal">
@@ -122,7 +122,7 @@ function Body() {
             <SidePanel />
           </Panel>
           <PanelResizeHandle
-            className={`h-full ${sidePanelCollapsed ? "w-0" : "w-2"}`}
+            className={`h-full ${sidePanelCollapsed ? "w-0" : "w-1"}`}
             id="PanelResizeHandle-SidePanel"
           />
           <Panel className="flex h-full overflow-hidden" minSize={50}>
--- a/src/ui/components/Library/Team/View/TestRuns/TestRunsPage.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/TestRunsPage.tsx
@@ -117,7 +117,7 @@ function TestRunsContent() {
           </div>
         </Panel>
 
-        <PanelResizeHandle className="h-full w-2" />
+        <PanelResizeHandle className="h-full w-1" />
         <Panel minSize={20} order={2}>
           <div className="h-full w-full overflow-hidden rounded-xl">
             <Suspense fallback={<LibrarySpinner />}>
--- a/src/ui/components/NetworkMonitor/index.tsx
+++ b/src/ui/components/NetworkMonitor/index.tsx
@@ -168,7 +168,7 @@ export default function NetworkMonitor() {
           </Panel>
           {selectedRequestId && (
             <>
-              <PanelResizeHandle className="h-2 w-full" />
+              <PanelResizeHandle className="h-1 w-full" />
               <Panel defaultSize={50}>
                 {selectedRequestId ? (
                   <RequestDetails requests={requests} selectedRequestId={selectedRequestId} />
--- a/src/ui/components/Timeline/PlaybackControls.tsx
+++ b/src/ui/components/Timeline/PlaybackControls.tsx
@@ -38,7 +38,7 @@ export default function PlayPauseButton() {
 
   return (
     <div
-      className="mx-1.5 flex flex-row"
+      className="ml-1 mr-1.5 flex flex-row"
       style={{ width: "32px", height: "32px" }}
       onClick={onClick}
     >
--- a/src/ui/components/Timeline/Timeline.css
+++ b/src/ui/components/Timeline/Timeline.css
@@ -69,7 +69,6 @@
   height: 32px;
   position: relative;
   width: 100%;
-  margin-right: 40px;
 }
 
 .timeline .progress {
--- a/src/ui/components/Timeline/Timeline.css
+++ b/src/ui/components/Timeline/Timeline.css
@@ -6,7 +6,7 @@
   box-sizing: border-box;
   --progressbar-transition: 200ms;
   display: flex;
-  padding: 10px 12px 12px 6px;
+  padding: 10px 0px 12px 6px;
   align-items: center;
   user-select: none;
 
@@ -52,6 +52,7 @@
   display: flex;
   position: relative;
   flex-grow: 1;
+  padding-right: 10px;
 }
 
 .progress-bar-stack {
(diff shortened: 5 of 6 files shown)


bvaughn:
I don't love that the vertical and horizontal margins/gaps are inconsistent after this:
Screen Shot 2023-11-09 at 5 23 43 PM

jonbell-lot23:
> I don't love that the vertical and horizontal margins/gaps are inconsistent after this: Screen Shot 2023-11-09 at 5 23 43 PM

Consistency isn't the goal when laying out a composition, and can sometimes be a mistake. For example, the top and bottom margins on a printed page are rarely identical to the left and right margins. That's what's at play here:

image

On prod, the vertical gap between the header and the columns isn't identical to the horizontal space, by design. But they're sort of close, which actually leads to it looking more sloppy. (Identical would make it even more sloppy because it'd be too tight considering the descenders you need to work around in text.)

In the proposed change, the horizontal gap is very clearly declaring its intent: that the columns are peers (tight horizontal spacing), but the header isn't (wider vertical margins, like a page in a physical book)

jasonLaster:
Project "let it breath" added spacing with the goal of making the UI more approachable. 

It's easy to forget this because we live in the app, but the UI is still intimidating to newcomers today. 

I mention this because I'm all for additional layout improvements that make the ui more understandable, but would be opposed to changes that make it more professional, but could blur the lines between panels.

bvaughn:
> > I don't love that the vertical and horizontal margins/gaps are inconsistent after this:
>
> Consistency isn't the goal when laying out a composition, and can sometimes be a mistake. For example, the top and bottom margins on a printed page are frequently different to the left and right margins.

On the one hand, I wasn't asserting any "goal". I was just saying that (to me) it looks bad. We often vary _padding_ between horizontal and vertical (for example: around text) but I think think, generally, that hard defined gaps like this should be consistent.

I don't think this change buys us much to be honest.

jonbell-lot23:
> I don't think this change buys us much to be honest. Why are we doing it? The PR description says _what_ but not _why_.

That's fair -- when dealing with subjective changes like these, there's no single metric we can point to. I think the new version is an improvement, you don't, so what do we do?

If we were on a design team, we'd take this through design crit. As it stands, I can only assert that I feel strongly that this is an improvement, and that it's not based on opinion but UX best practices.

I think I spend a lot of time being diplomatic and facilitating conversations, but I also think there's a point of diminishing returns and "design by committee." I propose a disagree and commit here.

bvaughn:
> I think I spend a lot of time being diplomatic and facilitating conversations, but I also think there's a point of diminishing returns and "design by committee." I propose a disagree and commit here.

That seems reasonable.

jonbell-lot23:
> That seems reasonable.

Thanks, Brian. Will be pressing the big green button soon.


**Wall of text time**
I think a lot about being a fair partner, and trying to phrase things the right way, etc. As my old mentor used to say "it's not just what you ship, but how." I want people to feel like I'm treating their ideas with respect, even when there's a difference of opinion. And I also know playing the designer card backfires, it's just another way of saying "do you know who I am?!" which never works well.

I'm adding this wall of text to say that I don't care about "winning" or "losing" as much as making sure things are being brought up and discussed (they were), then hashed out and moved forward (they were). So I just want to say thank you for being a fair and reasonable partner, however each particular feature lands, and I hope it feels the same way on your end. If not, I want you to know that making it feel fair and respectful is important to me. I am always fine tuning how to have a point of view without being a butt about it.

Nothing actionable here, just wanted to share. (again)

bvaughn:
Thanks Jon! :)

jasonLaster:
Chiming in from the plane. 

- I don't think we addressed the concerns I raised above (https://github.com/replayio/devtools/pull/9886#issuecomment-1807259622)

- I didn't follow the disagreement above. What aspect of the changes did we agree were good/bad?

jonbell-lot23:
In regards to the "professional vs approachable" items above, I think we all agree with a few things:

1. Replay (or any pro software) can be overwhelming at first
2. We (and all software) needs to be very careful about NUX. Onboarding is fantastically difficult.
3. Rounded edges, spacing, the right colors, etc all can make something feel more approachable

But I'd add one that is rooted in psychology/design: the Aesthetics Effect. People's brains receive information and prejudge a product *extremely quickly* and that's why every tiny little detail in a design matters. [In Defense of Eye Candy](https://alistapart.com/article/indefenseofeyecandy/) is a great article about this proven scientific discovery, with data. It's the same part of our head that makes us prejudge someone based on their appearance, no matter what we know logically.

So if we know people prejudge things quickly, and that details matter, it comes down to figuring out what is more aesthetically pleasing. We used to have 8px (.w-2) gaps, and this change cleans up our grid and changes to 4px (.w-1) gaps. My assertion is that this still provides extremely generous regions for gripping/resizing/distinguishing columns, without feeling "floaty" or "horsey", two design crit terms that mean things don't feel like they're resolved in the composition. I can't objectively prove that the 8px gaps are "horsey" but I can just say that in my experience and judgement, they are. And I am confident if I ran it by 10 designers in a crit, most would say the same. The 8px looks like a mistake. The 4px looks like we designed it.

But we're a small company so the best I can do is try to explain some of the background and promise to keep an open mind if I see data that contradicts my intuition. That means we're agreeing to disagree on this one (aka disagree and commit).

jasonLaster:
Thanks for the write up. Re trusting strong opinions, i'm happy to defer to your judgement here, but in the interest of building testable hypothesis I think it's worthwhile to elaborate the disabgreement.

I believe we went with 8px for information architecture reasons to help distinguish primary panels from sub panels. 

Sometimes our UI can feel like that puzzle where you need to count the number of nested triangles, but in our case it's panels. 

Screenshot 2023-11-14 at 4 39 54 PM

It's hard at our scale to test how well people navigate the UI, but I think that's what we would want to watch for. And of course, we can find other ways to try and help people understand the information hierarchy as well.

jonbell-lot23:
> I believe we went with 8px for information architecture reasons to help distinguish primary panels from sub panels. 

Thanks, Jason! I think you're right that we all agreed to move away from single pixel borders to help distinguish between primary and sub. Then it's just a question of the right amount of space: 10px gaps would feel too floaty, 1px could feel too packed in, then it's a subjective call between those two extremes.

I am absolutely open to fine-tuning this further based on our experience and what other people say. Will keep my ears to the ground with an open mind.

(Also in this PR: our timeline is spaced on the grid again, which I think will feel nice)

jasonLaster:
re timeline: undeniably.

re gap: definitely something we can adjust.

re IA: it'll be nice to run experiments in the future with hover interactions and other treatments that help here too. Part of the challenge is from a marketing perspective. We want a UI that looks good in the Hero and that forces more space too.  Anyways, enough said here, just wanted to align

bvaughn:
While we're talking about space here, there's something that's bugged me for a while :) 

Screen Shot 2023-11-15 at 8 09 32 AM

The bottom of our app wastes so much space that I realize that for months, I've been cropping most it out of my screenshots so it feels more balanced with the top.

In fact, we have 23 pixels of empty space down there.
Screen Shot 2023-11-15 at 8 10 25 AM

Vertical screen realestate is hug, so why do we do this?

jasonLaster:
Let's create a new issue for this. Also not sure what changes with the protocol timeline

bvaughn:
Sure. Filed DES-982.

PR #9887, opened 2023-11-09
Separate concerns for formatting point stacks to speed up Redux J2C perf

markerikson:
This PR:

- Reworks the `formattedPointStackCache` to:
  - stop preemptively formatting function details for every stack frame (which requires fetching source outlines)
  - move all of the "relevant app stack frame" logic into a separate cache (as that is really only needed by the "React Render Panel" component, not the Redux J2C logic)
- Updates the "relevant app frame" logic to be smarter about which frame it finds, by doing a check for sourcemapped stack frames if available
- Updates the `applyMiddlewareCache` to do its own call to the `sourcesById` cache instead of passing `sourcesState` through multiple levels
- Preloads the `applyMiddlewareCache` as soon as the "Redux" panel is selected, since that is needed for all Redux J2C calls

This noticeably speeds up Redux J2C calls - we're now mostly limited by how fast `getPointStack` returns.

(I could imagine pre-fetching point stacks on hover, but we'll save that for another time.)

### DIFF
--- a/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsPanel.tsx
+++ b/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsPanel.tsx
@@ -1,4 +1,4 @@
-import { Suspense, useContext, useMemo, useState } from "react";
+import { Suspense, useContext, useEffect, useMemo, useState } from "react";
 import { PanelGroup, PanelResizeHandle, Panel as ResizablePanel } from "react-resizable-panels";
 import { useImperativeCacheValue } from "suspense";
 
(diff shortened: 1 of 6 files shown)


src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsPanel.tsx
```diff
@@ -60,6 +61,12 @@ export default function ReduxDevToolsPanel() {
     return [filteredAnnotations, firstAnnotationAfterCurrentExecutionPoint];
   }, [annotationsStatus, currentExecutionPoint, focusWindow, parsedAnnotations, searchValue]);
 
+  useEffect(() => {
+    // Preload the cache for where `applyMiddleware` is defined
+    // This will speed up the first click on a Redux "J2C" button
+    applyMiddlewareDeclCache.readAsync(client);
```
hbenl, line 67: nit: `suspense` offers `Cache.prefetch()` for this scenario: the only difference is that `Cache.prefetch()` will catch errors (and doesn't return anything).

hbenl (approved):

src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsPanel.tsx
markerikson: Ah, yes, that _does_ exist!  Okay, will throw that in.

PR #9888, opened 2023-11-10
Source maps debugger

jasonLaster:
Fixes FE-2052


packages/protocol/socket.ts
```diff
@@ -252,11 +253,14 @@ const doSend = makeInfallible(message => {
...
 
   gSessionCallbacks?.onRequest(message);
-
   socket.send(stringified);
 });
 
 function flushQueuedMessages() {
+  if (socket.readyState != 1) {
+    return;
+  }
```
jasonLaster, line 262: These two checks are needed to prevent HMR issues. I think they're fairly safe to land because we only want to send messages when the socket is in fact connected

pages/recording/[id]/sourcemaps/[sourceId].tsx
```diff
@@ -0,0 +1,216 @@
...
+    const sourcesMap: Map<string, Source> = new Map();
+    client.Debugger.addNewSourceListener(source => {
+      sourcesMap.set(source.sourceId, source);
+    });
+    client.Debugger.addNewSourcesListener(({ sources }) => {
+      for (const source of sources) {
+        sourcesMap.set(source.sourceId, source);
+      }
+    });
+    await client.Debugger.findSources({}, sessionId);
```
Comment (jasonLaster) [inline pages/recording/[id]/sourcemaps/[sourceId].tsx:78]: this code is very similar to the source map visualizer, but i felt it was premature to extract.

.trunk/tools
```diff
@@ -0,0 +1 @@
+/Users/jasonlaster/.cache/trunk/repos/27d5659a69425a48e0eb0d117231e40e/tools
```
bvaughn, line 1: whats this

bvaughn (changes requested):
Let's talk about this Monday!

.trunk/tools
jasonLaster: will revert.

jasonLaster:
> Let's talk about this Monday!

Okay

packages/protocol/socket.ts
```diff
@@ -202,8 +202,9 @@ export async function sendMessage<M extends CommandMethods>(
   const id = gNextMessageId++;
   const msg: CommandRequest = { id, method, params, pauseId, sessionId };
   const callerStackTrace = new Error(`Caller stacktrace for ${method}`);
+  console.log(`sendMessage`, method);
```
bvaughn, line 205:
```suggestion
```

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -131,6 +131,25 @@ export function getSourcemapVisualizerURLSuspense(
   return href;
 }
 
+export function getSourcemapsViewerURLSuspense(
```
bvaughn, line 134: Why is this called "Suspense"? It's not suspending.
```suggestion
export function getSourcemapsViewerURL(
```

jasonLaster:
I'm waiting to pick this up with the Pinterest team next week.

bvaughn:
Seems like maybe this work should go into the [`experimental` branch](https://github.com/replayio/devtools/tree/experimental) for now.

PR #9889, opened 2023-11-10
Make the useLineHighlights hook stateless

hbenl:
I've made the hook stateless (see FE-2055) and also removed the handling of corresponding sources, which isn't necessary since we handle them in our caches.
Before: [Replay](https://app.replay.io/recording/line-highlight-not-cleared--53361b73-2c2a-46dd-b869-51d228aafff2)
After: [Replay](https://app.replay.io/recording/line-highlight-cleared--8cfefac0-3e6b-479b-ae76-16bbe8aba4ff)


markerikson (approved):

jasonLaster (commented):
Do we have any E2E tests for source line highlighting?

bvaughn:
> Do we have any E2E tests for source line highlighting?

@hbenl Not sure if you missed it, but this question is worth answering.

hbenl:
I'll add one (FE-2065)

PR #9890, opened 2023-11-10
Fix pause state regressions

hbenl:
- `TimelineContext`: don't fall back to execution point `"0"` - in the past this fallback was only used at the start of the session but now the execution point in redux is set to `null` while we're waiting for the `Session.getPointNearTime` result and we shouldn't jump to `"0"` during that time
- added the `useMostRecentLoadedPause()` hook and used it in the Elements panel (fixing FE-2053), the `CurrentTimeIndicator` in the console (which jumped around while seeking in the timeline) and the RDT panel (which already had its own implementation for getting the most recent loaded pause)

### DIFF
--- a/packages/replay-next/components/console/ConsoleInput.tsx
+++ b/packages/replay-next/components/console/ConsoleInput.tsx
@@ -30,7 +30,7 @@ export default function ConsoleInput({ inputRef }: { inputRef?: RefObject<Impera
 
   let disabledMessage = null;
   let disabledReason = undefined;
-  if (!isPointWithinFocusWindow) {
+  if (executionPoint && !isPointWithinFocusWindow) {
     disabledReason = "not-focused";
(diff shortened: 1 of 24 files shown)


markerikson (approved):
I don't have the nuances of behavior here loaded into my head, but this _seems_ reasonable at first glance

jasonLaster (commented):
Can we add or update an E2E test to make sure we don't regress in the future?

bvaughn:
> Can we add or update an E2E test to make sure we don't regress in the future?

@hbenl Not sure if you missed it, but this seems worth doing.

hbenl:
These regressions were about loading states - to test loading states, we'd need a way to delay backend responses until the test has verified the UI for the loading state, but this is currently not possible in our E2E tests.

PR #9892, opened 2023-11-11
Create a tests view for test workspaces

jaril:
This adds a tests view, as detailed [here](https://www.notion.so/Testsuites-V2-Design-Spec-be3f0201e26e494090ca6b9e62fbd26e). This is for displaying all the tests recorded for a workspace, and surfacing useful information like failure rates over time, passing replays, and failing replays.

In a way, this is not complete yet. We're not able to query performantly yet because some of the data that this view was designed to handle. So added this under a feature flag for now until we're able to turn it on. To enable the test view, go to the console and run `localStorage.setItem("enableTestSuitesTestsView", true)`.

Other notes:
- Failure rates are currently calculated in the frontend. This will be moved to the backend eventually
- We have a time filter that should apply to both 1) tests and 2) errors. That API's not implemented yet so the corresponding dropdowns in the frontend don't do anything right now
- There's a sort by feature for tests as well. Same as above, it's not implemented yet, so the dropdown doesn't do anything for now


jasonLaster:
1. Added a fixture so that data would load quickly 
2. Copied a bunch of the TestRun files into the Tests directory
3. Made a bunch of renaming fixes

jaril:
1. Fixed displaying the test list
2. Pre-computed the failure rate in `src/ui/components/Library/Team/View/TestRuns/suspense/TestsCache.ts`
3. Started displaying the failure rates in the right hand side panel

bvaughn:
What is this PR? No description and a huge delta.

Screen Shot 2023-11-14 at 10 59 14 AM

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // }, [filterByBranch, filterByStatus, filterByText, testRuns]);
+
+  // useEffect(() => {
+  //   if (testRunId == null) {
+  //     // Select the first test run by default if nothing is selected.
+  //     setTestRunId(testRuns[0]?.id ?? null);
+  //   }
+  // }, [router, teamId, testRunId, testRuns]);
+
+  // useSyncTestRunIdToUrl(teamId, testRunId, setTestRunId);
```
bvaughn, line 72: Kill the dead code:
```suggestion
```

src/ui/components/Library/Team/View/Tests/Overview/TestOverviewContainerContextType.tsx
```diff
@@ -0,0 +1,8 @@
+import { ReactNode, createContext } from "react";
+type TestOverviewContainerContextType = {};
+
+export const TestOverviewContext = createContext<TestOverviewContainerContextType>(null as any);
+
+export function TestOverviewContainer({ children }: { children: ReactNode }) {
+  return <TestOverviewContext.Provider value={{}}>{children}</TestOverviewContext.Provider>;
+}
```
bvaughn, line 8: This component doesn't really add much value. Since it's only used in one place (`TestOverviewPage`) can you just inline the context provider there?

src/ui/components/Library/Team/View/Tests/hooks/useTests.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { GraphQLClientContext } from "replay-next/src/contexts/GraphQLClientContext";
+import { TeamContext } from "ui/components/Library/Team/TeamContextRoot";
+import useToken from "ui/utils/useToken";
+
+import { testsCache } from "../../TestRuns/suspense/TestsCache";
+
+const EMPTY_ARRAY: any[] = [];
+
+export function useTests(): any[] {
```
bvaughn, line 12: Replace `any[]` with an actual type.

`useTestRuns()` returns a `TestRun[]`. This should probably return a `Test[]` array (presumably the `Test` type that you added to `packages/shared/test-suites/TestRun`)

src/ui/components/Library/Navigation/TeamButton.tsx
```diff
@@ -43,25 +45,73 @@ export function TeamButton({
...
+
+  const setView = (view: View) => {
+    pushRoute(router, `/team/${teamId}/${view}`);
+  };
+
+  return (
+    <div className="pl-4 pb-2">
+      <div className="flex flex-col">
+        <div
+          className={`px-4 py-1 ${
```
bvaughn, line 96: The margin and alignment for this sub menu looks awkward to me. The "Runs" text is not indented enough to look "nested" beneath the team name:

Screen Shot 2023-11-15 at 10 20 58 AM

And it looks lopsided vertically because of the added bottom padding:

Screen Shot 2023-11-15 at 10 21 12 AM

I suggest changing it like so for now:
```suggestion
          className={`px-8 ${
```

Screen Shot 2023-11-15 at 10 22 55 AM

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,176 @@
...
+import orderBy from "lodash/orderBy";
+import { Dispatch, SetStateAction, useState } from "react";
+import { ContextMenuItem, useContextMenu } from "use-context-menu";
+
+import Icon from "replay-next/components/Icon";
+import { TestExecution } from "shared/test-suites/TestRun";
+
+import { getTruncatedRelativeDate } from "../../Recordings/RecordingListItem/RecordingListItem";
+import styles from "../../../../Library.module.css";
+import testPageStyles from "../TestsPage.module.css";
```
bvaughn, line 11: This feels kind of clunky. Seems like maybe the "drop down trigger" thing could be its own reusable component rather than having views import each others styles like this.

In general, I don't like that pattern because I think it's easy to make (what you think is) a localized style edit that breaks something else (because you don't assume that other components import your style modules).

src/ui/components/Library/Navigation/TeamButton.tsx
```diff
@@ -43,25 +45,73 @@ export function TeamButton({
...
+            view === "runs" ? "font-bold" : ""
+          } hover:cursor-pointer hover:font-bold`}
+          onClick={() => setView("runs")}
+        >
+          Runs
```
bvaughn, line 101: We should only be showing this sub-view for tests suites pages, right? This PR shows it for everything:
https://www.loom.com/share/52299129b46e4430962fac5396e7f588

Also I think we should only show the submenus ("Runs" and "Tests") if the `enableTestView` feature is on, but this PR shows it regardless which is kind of confusing. (Why show only a single option?)

src/ui/components/Library/Team/View/TestRuns/suspense/TestsCache.ts
```diff
@@ -0,0 +1,79 @@
...
+  Test[]
+>({
+  config: { immutable: true },
+  debugLabel: "testsCache",
+  getKey: ([_, __, workspaceId]) => workspaceId,
+  load: async ([graphQLClient, accessToken, workspaceId]) => {
+    const rawTests = await getTestsGraphQL(graphQLClient, accessToken, workspaceId);
+
+    // Compute the failure rate and add it to the stored test in cache
+    // until we're able to do it from the backend
```
bvaughn, line 20: Comments like this are easy to overlook and forget about. Add a TODO that links to a Linear issue for this?

src/ui/components/Library/Navigation/TeamButton.tsx
```diff
@@ -43,25 +45,73 @@ export function TeamButton({
...
+    pushRoute(router, `/team/${teamId}/${view}`);
+  };
+
+  return (
+    <div className="pl-4 pb-2">
+      <div className="flex flex-col">
+        <div
+          className={`px-4 py-1 ${
+            view === "runs" ? "font-bold" : ""
+          } hover:cursor-pointer hover:font-bold`}
```
bvaughn, line 98: I think we should probably also avoid using "bold" style for hover state. Having text resize on mouse over feels jarring.

```suggestion
          } hover:cursor-pointer`}
```

src/ui/components/Library/Navigation/TeamButton.tsx
bvaughn: This sub-menu is also visible under "Your Library"
Screen Shot 2023-11-15 at 10 44 37 AM

And clicking these just shows an empty page (because this isn't a test suite workspace).

packages/shared/user-data/LocalStorage/config.ts
```diff
@@ -10,4 +10,5 @@ export const config = {
   reactDevToolsStateExpanded: Boolean(true),
   replayNextCurrentPanel: "sources" as ReplayNextCurrentPanel,
   replayVideoPanelCollapsed: false,
+  enableTestView: false,
```
bvaughn, line 13: nit: This could use a more descriptive name. Maybe something like `enableTestSuitesTestsView` or something?

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,176 @@
...
+    contextMenu: contextMenuTimeFilter,
+    onContextMenu: onClickTimeFilter,
+    onKeyDown: onKeyDownTimeFilter,
+  } = useContextMenu(
+    <>
+      <ContextMenuItem dataTestId="show-all-runs" onSelect={() => setFilterByTime(null)}>
+        All errors
+      </ContextMenuItem>
+    </>,
+    { alignTo: "auto-target" }
```
bvaughn, line 36: Only adding one option to these menus makes them kind of useless:

Screen Shot 2023-11-15 at 10 46 55 AM

Screen Shot 2023-11-15 at 10 46 58 AM

Screen Shot 2023-11-15 at 10 47 02 AM

There's no way to turn them off.

bvaughn (commented):
Initial batch of feedback.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
jasonLaster: Lets show the other options that are in the mock, but have them be disabled. This will let users see that they're coming

src/ui/components/Library/Team/View/Tests/Overview/TestOverviewContainerContextType.tsx
jaril: Yep — deleted this structure so we can just use the `TestOverviewContent` component directly

src/ui/components/Library/Team/View/TestRuns/suspense/TestsCache.ts
jaril: Sounds good — captured in https://linear.app/replay/issue/SCS-1575/compute-the-failure-rates-in-the-backend-instead-of-the-frontend

bvaughn:
@jaril TypeScript is failing in CI: https://github.com/replayio/devtools/actions/runs/6882322396/job/18720591892?pr=9892

jonbell-lot23 (commented):
1. The capsule text should be #fff
2. Please add  percentage symbol to the end: "50%" rather than "50"

image


4. When the right column is empty, it should have a little note like this: 

image

5. Why are these replays dimmed? 


image

6. 

We shouldn't show a header when there's no content below it

image


7. If something has never failed, I guess we should make it green instead of red: 

image

I don't think any of these are strictly required to push live. Happy to handle these on the design polish side so you can focus on the nuts and bolts.

jaril:
> 1. The capsule text should be #fff
Fixed!

> 2. Please add  percentage symbol to the end: "50%" rather than "50"
Fixed!

> 3. When the right column is empty, it should have a little note like this:
Good catch. Adding.

> 4. Why are these replays dimmed?
The backend doesn't pass those replays to the frontend correctly yet, so we know when they were executed but we don't know what their IDs are. It should be fixed soon:tm:

> 6. If something has never failed, I guess we should make it green instead of red:
Copy!

src/ui/components/Library/LibraryDropdownTrigger.tsx
```diff
@@ -0,0 +1,32 @@
...
+  label,
+  testId,
+}: {
+  onKeyDown: KeyboardEventHandler;
+  onClick: MouseEventHandler;
+  label: string;
+  testId?: string;
+}) {
+  return (
+    <>
```
bvaughn, line 19: nit: Remove the `<>` (fragment) wrapper. It's unnecessary.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,181 @@
...
+}
+
+function ErrorReplays({ executions }: { executions: TestExecution[] }) {
+  const passing = executions.filter(e => e.result === "passed");
+  const failing = executions.filter(e => e.result === "failed");
+  const sortedFailing = orderBy(failing, "createdAt", "desc");
+  const sortedPassing = orderBy(passing, "createdAt", "desc");
+
+  // TODO: show the processed/unprocessed status for the replays -jvv
+  // TODO: add the top commit for the corresponding replay as label -jvv
```
bvaughn, line 127: nit: Nice to link these to a Linear issue (e.g. `// TODO [SCS-#]`)

src/ui/components/Library/Navigation/TeamButton.tsx
```diff
@@ -43,25 +46,62 @@ export function TeamButton({
...
+        onClick={onClick}
+      >
+        <span className="overflow-hidden overflow-ellipsis whitespace-pre">
+          <div className="flex">
+            <LibraryIcon teamType={isTest ? "tests" : "team"} />
+            {label}
+          </div>
+        </span>
+        {isNew ? (
+          <div className={"rounded-md bg-primaryAccent px-3 py-0.5 text-xs text-white"}>New</div>
```
bvaughn, line 70: This is a pretty wide button. Testing it I realized that we don't really handle text overflow well in the side nav:
Screen Shot 2023-11-16 at 8 10 34 AM

I think the above style was meant to address that, but doesn't:
```html
<span className="overflow-hidden overflow-ellipsis whitespace-pre">
```

I think to fix it we'd want to move that wrapper around the `{label}`:
```jsx
<div className="flex overflow-hidden">
  <LibraryIcon teamType={isTest ? "tests" : "team"} />
  <span className="overflow-hidden overflow-ellipsis whitespace-pre">{label}</span>
</div>
```

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <ContextMenuItem disabled onSelect={() => setFilterByTime(1)}>
+        Last day
+      </ContextMenuItem>
+      <ContextMenuItem disabled onSelect={() => setFilterByTime(7)}>
+        Last week
+      </ContextMenuItem>
+      <ContextMenuItem disabled onSelect={() => setFilterByTime(30)}>
+        Last month
+      </ContextMenuItem>
+      <ContextMenuItem onSelect={() => setFilterByTime(null)}>All time</ContextMenuItem>
```
bvaughn, line 44: There's something that feels off to me about showing time filters for both the outer test runs column _and_ the inner top errors column. For example, filtering the outer test runs column to show only failed tests for the past hour and then filtering top errors for the selected test to be for "all time" feels misleading and confusing. Seems like the inner column should mirror the outer setting?

cc @jonbell-lot23 because I realize this is more of a UX comment.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,181 @@
...
+        </div>
+      </div>
+      <div className="flex flex-col gap-2">
+        <div className="overflow-hidden overflow-ellipsis whitespace-nowrap text-lg font-medium">
+          Recent replays of the test passing
+        </div>
+        <div className="flex flex-col gap-2">
+          {sortedPassing.slice(0, MAX_REPLAYS_SHOWN).map((e, i) => (
+            <ErrorReplayList date={e.createdAt} index={i + 1} key={i} status="passed" />
+          ))}
```
bvaughn, line 147: We should probably show something for each of these recent blocks when they're empty. This looks kind of broken to me:
Screen Shot 2023-11-16 at 8 17 26 AM

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
jonbell-lot23: Yeah this is a tough one. The middle mirroring the left can cause other issues. I agree that this can be confusing, but sometimes this granularity is needed and I think the alternative is worse. Will be keeping an eye on this for sure.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
jaril: Recently changed this behavior — for now, I just stopped displaying a block if there are no applicable replays, ie "Recent replays of the test passing" wouldn't show up in the first place.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
jaril: @jonbell-lot23 for that one, in which cases do you imagine the second dropdown will be used?

The more I think about this the less sure I am that we need the second time dropdown since it's redundant. Any narrowing-down action that applies to the second dropdown can just as easily be applied with the first dropdown.

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -0,0 +1,85 @@
...
+        filterByTextForDisplay: filterByText,
+        selectTestId: setTestId,
+        setFilterByTime,
+        setSortBy,
+        setFilterByText,
+        testId: deferredTestId,
+        testIdForDisplay: testId,
+        tests: filteredTests,
+      }}
+    >
```
bvaughn, line 81: nit: Might be worth memoizing this context value to avoid scheduling unnecessary client updates, but it's probably not necessary.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </ContextMenuItem>
+      <ContextMenuItem onSelect={() => setFilterByTime(null)}>All time</ContextMenuItem>
+    </>,
+    { alignTo: "auto-target" }
+  );
+
+  const uniqueErrorCount = Object.entries(errorFrequency).length;
+
+  return (
+    <div>
```
bvaughn, line 52: This needs overflow. As it is now, if the page is too short to show them all, you can't scroll to see what's cut off. (Test by making your browser page shorter.)
```suggestion
        <div class="overflow-y-auto">
```

src/ui/components/Library/Team/View/Tests/Overview/TestOverviewContent.tsx
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+  );
+}
+
+function Stats({ failureRates }: { failureRates: FailureRates }) {
+  const { hour, day, week, month } = failureRates;
+
+  return (
+    <div className="border-b border-themeBorder py-2 px-4">
+      <div>
```
bvaughn, line 52: Pretty sure this needs to be overflow as well
```suggestion
<div class="overflow-y-auto">
```

jaril:
@bvaughn Ready for another look! 

Note there seems to be delta failures, but I'm not sure they're related.

jaril:
@bvaughn Ready for another look!

bvaughn (commented):
Vertical overflow still doesn't scroll...

jaril:
> Vertical overflow still doesn't scroll...

My bad — fixing it now.

jaril:
Double checked and we should be good with the scrolling!

https://github.com/replayio/devtools/assets/15959269/2f3958c1-dcb0-4fda-81c5-3b5a50a81bc7

bvaughn:
RE scrolling: https://www.loom.com/share/f7c6617c7ea74f7889ff2e7462f38b3f cc @jonbell-lot23

bvaughn (approved):

PR #9884, opened 2023-11-09
Better (loading) UX when inspecting an HTML element

bvaughn:
- [x] Replace the mutable ref with a ref-setter in `ElementsPanelAdapter` so that we'll pass through the "selected Node" in Redux after the `ElementsList` renders (if it hasn't rendered yet).
- [x] It can take a while to load the nested elements– during which, the Elements panel doesn't show any kind of loading indicator. Show a loading indicator during this time.

https://github.com/replayio/devtools/assets/29597/62aa583f-015e-4f77-b028-31aa1f9810f6

Note I temporarily copied the new background color style from #9880 for the video about (because it contrasts better with the loading progress bar).

PR #9893, opened 2023-11-14
Better (loading) UX when inspecting an HTML element

bvaughn:
Note this was originally PR #9884 but GitHub closed it after a rebase and wouldn't let me re-open it.

---

- [x] Replace the mutable ref with a ref-setter in `ElementsPanelAdapter` so that we'll pass through the "selected Node" in Redux after the `ElementsList` renders (if it hasn't rendered yet).
- [x] It can take a while to load the nested elements– during which, the Elements panel doesn't show any kind of loading indicator. Show a loading indicator during this time.

https://github.com/replayio/devtools/assets/29597/62aa583f-015e-4f77-b028-31aa1f9810f6

Note I temporarily copied the new background color style from #9880 for the video about (because it contrasts better with the loading progress bar).


bvaughn:
We're back in business after rebasing:

https://github.com/replayio/devtools/assets/29597/a3325286-913e-411d-8907-67836cc48aeb

PR #9895, opened 2023-11-15
Fix the useMostRecentLoadedPause hook

hbenl:


packages/replay-next/src/contexts/TimelineContext.tsx
```diff
@@ -23,22 +29,32 @@ export type TimelineContextType = {
...
+  const update = useCallback(
+    (time: number, executionPoint: ExecutionPoint) => {
+      // Components might suspend in response to the this changing.
+      startTransition(() => {
+        setExecutionPoint(executionPoint);
+        setTime(time);
+      });
+      async function updatePauseId() {
+        const pauseId = await pauseIdCache.readAsync(client, executionPoint, time);
+        setPauseId(pauseId);
```
bvaughn, line 48: Components might suspend in response to the pause changing too, so it should probably also be set inside of the transition?

bvaughn: (Or maybe a separate transition would be okay too?)

I'm a little worried about the "tearing" we might see by splitting these two updates apart.

bvaughn (approved):
Overall approved, with the change we talked about on Hangouts of temporarily setting pause id to `null` before the new one is created (and updating the new one inside of a transition).

packages/replay-next/src/contexts/TimelineContext.tsx
```diff
@@ -23,22 +29,33 @@ export type TimelineContextType = {
...
+    (time: number, executionPoint: ExecutionPoint) => {
+      // Components might suspend in response to the this changing.
+      startTransition(() => {
+        setExecutionPoint(executionPoint);
+        setTime(time);
+        setPauseId(null);
+      });
+      async function updatePauseId() {
+        const pauseId = await pauseIdCache.readAsync(client, executionPoint, time);
+        setPauseId(pauseId);
```
bvaughn, line 49: I think this still needs to be wrapped in a transition since it might cause things to suspend.

PR #9896, opened 2023-11-15
Use cache status to show a loading state for the Redux J2C button

markerikson:
This PR:

- Uses the `useCacheStatus` hook from `suspense` to pre-read whether we have a J2C entry cached or loading, and uses that to set the right loading state for the J2C button

Example:

image

### DIFF
--- a/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsListItem.tsx
+++ b/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsListItem.tsx
@@ -1,11 +1,14 @@
-import { CSSProperties } from "react";
+import { CSSProperties, useContext } from "react";
+import { Status, useCacheStatus } from "suspense";
 
 import { getExecutionPoint } from "devtools/client/debugger/src/selectors";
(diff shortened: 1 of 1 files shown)


bvaughn (commented):
The fe-1875 e2e test is failing on this branch but I don't see another recent failure in the dashboard. Is this a legit regression? (I also see a bunch of snapshot failures but I didn't look at the cause.)

bvaughn (approved):

markerikson:
Hmm. Glancing at those failures:

- The test is related to the Cypress panel, not the Redux panel
- The failures + retries are because the test clicked on the Cypress panel J2C button, and are looking at the frames panel to confirm that we've paused, and never actually found a selected frame in the frames panel
- We did get a _list_ of frames (as opposed to some test failures I've seen where we never got frames and thus were never considered paused), but the current frame isn't actually marked as selected

That passed in both #9887 when I updated it earlier today, and #9885 when I merged it last week.

I'm pretty confused with that, tbh.  It clearly can't have anything to do with _this_ PR, which only changed the Redux list item component.

PR #9897, opened 2023-11-15
Always enable new React DevTools for Chromium

bvaughn:
- [x] Removed new React DevTools feature flag
  * Choose whether to show new or old RDT based on runtime (Chromium = new, Firefox = old)
- [x] Fix light theme changed prop highlight color
- [x] Update e2e tests to run against new RDT


packages/e2e-tests/tests/react_devtools-02-integrations.test.ts
```diff
@@ -141,55 +143,67 @@ test("react_devtools-02: RDT integrations (Chromium)", async ({
...
   // Should jump to the `render()` method in `<SourcesTree>`, which is a class component
   await waitForSelectedSource(page, "SourcesTree.tsx");
   await waitFor(async () => {
     const lineNumber = await getSelectedLineNumber(page, false);
-    expect(lineNumber).toBe(280);
+
+    // TODO Line 98 is the constructor
+    expect(lineNumber).toBe(98);
+    // TODO View source should point to the render() function on line 280
+    // expect(lineNumber).toBe(280);
```
bvaughn, line 175: @markerikson Can we chat about this?

packages/e2e-tests/helpers/new-react-devtools-panel.ts
```diff
@@ -32,6 +105,98 @@ export async function openReactDevtoolsPanel(page: Page) {
...
 }
 
+export async function searchComponents(page: Page, text: string) {
+  const input = getSearchInput(page);
+
+  await waitFor(async () => {
+    await expect(await input.isEnabled()).toBe(true);
+  });
+
+  await waitFor(async () => {
```
hbenl, line 115: I don't think this `waitFor()` makes any sense since we don't `expect()` anything in its body. Also we wouldn't want to run its body more than once because then it would fill the search input with multiple copies of the `text`.

packages/e2e-tests/helpers/new-react-devtools-panel.ts
```diff
@@ -32,6 +105,98 @@ export async function openReactDevtoolsPanel(page: Page) {
...
+
+      if (expectedTotal != null) {
+        expect(actualTotal).toBe(expectedTotal);
+      }
+    });
+  }
+
+  if (expectedName != null) {
+    await waitFor(async () => {
+      await verifySelectedComponentName(page, expectedName);
```
hbenl, line 185: I don't think this `waitFor()` makes sense because `verifySelectedComponentName()` has a `waitFor()` itself.

hbenl (approved):

packages/e2e-tests/helpers/new-react-devtools-panel.ts
bvaughn: Maybe the wait-for is unnecessary. I added in case (for some reason) the search didn't yield results.

> it would fill the search input with multiple copies of the text.

This isn't correct though. This line will clear the input (if there's any text in it from previously, including a previous wait-for iteration):
```js
    await input.press("Escape"); // Clear/reset
```

packages/e2e-tests/helpers/new-react-devtools-panel.ts
bvaughn: That's fair. I think I probably wrote this wait-for before moving that stuff into a helper function.

packages/e2e-tests/helpers/new-react-devtools-panel.ts
bvaughn: > I don't think this waitFor() makes any sense since we don't expect() anything in its body.

Just to be clear though, we do expect something–
```js
    const searchResults = page.locator('[data-test-id="ReactDevTools-SearchResults"]');
    await searchResults.waitFor();
```

The above lines will fail if a search isn't active. (The search results UI won't exist.)

PR #9898, opened 2023-11-15
Update RDT test assertion 

markerikson:
This PR:

- Updates the `rdt-02` test component names assertion to match the latest routine behavior

Here's the current tree for this test:

image


markerikson:
Bingo!

image

markerikson:
Double-checking latest run...

hbenl (approved):

markerikson:
The `rdt-02` test is now failing further down, because there's more components being labeled as "Anonymous" now than there were before. Investigating that behavior...

PR #9899, opened 2023-11-15
Replace async loadNodePathToRoot() with DOM.getParentNodes

bvaughn:


packages/replay-next/components/elements/suspense/DOMParentNodesCache.ts
```diff
@@ -0,0 +1,40 @@
+import { ObjectId, PauseId } from "@replayio/protocol";
+import { createCache } from "suspense";
+
+import { elementCache } from "replay-next/components/elements/suspense/ElementCache";
+import { objectCache } from "replay-next/src/suspense/ObjectPreviews";
+import { cachePauseData } from "replay-next/src/suspense/PauseCache";
+import { sourcesByIdCache } from "replay-next/src/suspense/SourcesCache";
+import { ReplayClientInterface } from "shared/client/types";
+
+export const parentNodesCache = createCache<
```
markerikson, line 10: :speech_balloon: fwiw I had a similar (but expanded) cache in `ui/suspense/nodeCaches.ts` prior to the removal of the old elements panel.  it handled a few different for "fetch some DOM node previews".  May be worth cross-comparing vs that implementation.

bvaughn: I think I remember the previous "nodes cache" thing as being a bit confusing. I just looked at it again and I see three references to "parentNodes" (one of which was the type enum) but I don't see anything it does that this doesn't do? (I could be missing it though, it's late.)

markerikson: Yeah, just looked at the prior implementation, and it does seem basically equivalent for the `parentNodes` case. Never mind!  (didn't immediately remember if I had it doing more logic internally.)

bvaughn: No worries. I'm glad you double checked.

bvaughn:
Given the perf gains, I think this is worth landing regardless of whether we do the more dramatic change discussed on FE-2067.

hbenl (approved):

PR #9900, opened 2023-11-16
Fix the logpoint badge selector in the console context menu

hbenl:


packages/e2e-tests/tests/logpoints-01.test.ts
```diff
@@ -47,6 +50,14 @@ test(`logpoints-01: log-points appear in the correct order and allow time warpin
   await expect(logPointMessages.first()).toHaveText(/Beginning/);
   await expect(logPointMessages.last()).toHaveText(/Ending/);
 
+  await setLogpointBadge(page, logPointMessages.nth(1), "unicorn");
+  await waitFor(async () => {
+    await verifyLogpointBadge(logPointMessages.nth(1), "unicorn");
+    await verifyLogpointBadge(logPointMessages.nth(2), "unicorn");
+  });
+  await verifyLogpointBadge(logPointMessages.first(), null);
+  await verifyLogpointBadge(logPointMessages.last(), null);
```
bvaughn, line 59: I like the idea of adding test coverage for this! I'm not sure this test is actually verifying that the setter works though, since it doesn't check the before state. Can we add an explicit check for that too?

packages/replay-next/components/console/useConsoleContextMenu.tsx
```diff
@@ -112,21 +112,21 @@ export default function useConsoleContextMenu(loggable: Loggable) {
...
-              onSelect={() => setBadge("unicorn")}
+              onClick={() => setBadge("unicorn")}
             />
             {BADGES.map(badge => (
               <div
                 key={badge}
                 className={styles.ColorBadge}
                 data-test-id={`ConsoleContextMenu-Badge-${badge}`}
-                onSelect={() => setBadge(badge)}
+                onClick={() => setBadge(badge)}
```
bvaughn, line 122: ooof.

bvaughn (approved):

PR #9902, opened 2023-11-16
Minimal tests view

jaril:
This builds on top of the [original test view PR](https://github.com/replayio/devtools/tree/1109-tests-view). 

There were two main problems with landing the original PR:
1. The backend cannot yet performantly return the test data we need. It works for smaller teams like Replay E2E test, but bombs with larger ones like Metabase or Pinterest POC.
2. Even if it's able to return the test data, it has difficult (read: cannot) return the recordings associated with each test execution.

This new branch scales back the original PR. Instead of querying all of the workspace's tests, their executions and corresponding recordings, it splits the work into two separate queries. `GetTestsForWorkspace` which is responsible for getting the test names, and `GetTestForWorkspace` which is responsible for getting the execution/replay data for the selected workspace. We're essentially just asking the backend to do the expensive query for one test, instead of all of them.


src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -47,6 +48,8 @@ export function TestErrorList({
   );
 
   const uniqueErrorCount = Object.entries(errorFrequency).length;
+  // We don't bother displaying executions without replays
+  const filteredExecutions = executions.filter(e => e.recording[0] !== null);
```
jaril, line 52: There is an incoming backend fix that will make it so that we don't have to filter by null anymore

src/ui/components/Library/Team/View/Tests/hooks/useTests.ts
```diff
@@ -24,3 +30,92 @@ export function useTests(): Test[] {
   );
   return value;
 }
+
+export function useTest(testId: string) {
```
jaril, line 34: Wasn't sure how to use the existing cache for something that follows a progressively-loaded pattern, so just resorted to a simpler useQuery here for now

markerikson:
When you say it "builds on top of" the other PR... I see this is targeted at the other PR's branch.  Is the intent to land the entire thing together?

src/ui/components/Library/Team/View/Tests/hooks/useTests.ts
```diff
@@ -24,3 +30,92 @@ export function useTests(): Test[] {
...
+  const { teamId } = useContext(TeamContext);
+  const { data, loading } = useQuery(GET_TEST, {
+    variables: { workspaceId: teamId, testId },
+  });
+
+  if (loading) {
```
markerikson, line 40: :hand: This whole section should be wrapped in `useMemo(cb, [data, loading])` so we only recalculate it when this changes.

src/ui/components/Library/Team/View/Tests/hooks/useTests.ts
```diff
@@ -24,3 +30,92 @@ export function useTests(): Test[] {
...
+      executions.filter(
+        e => (Date.now() - new Date(e.createdAt).getTime()) / 1000 / 60 / 60 / 24 < 30
+      )
+    ),
+  };
+}
+
+function getErrorFrequency(executions: TestExecution[]) {
+  return executions.reduce((acc: Record<string, number>, e) => {
+    let newAcc = { ...acc };
```
markerikson, line 99: :hand: no need to spread this. It starts with a brand new empty object at the bottom, so it's safe to mutate all the way through the `reduce()` call.

markerikson (commented):
I'm getting the sense that the entire _other_ PR needs to be reviewed too, not just the couple hundred lines in this PR?

packages/shared/graphql/generated/GetTestsForWorkspace.ts
```diff
@@ -11,19 +11,11 @@ export interface GetTestsForWorkspace_node_Recording {
...
 }
 
-export interface GetTestsForWorkspace_node_Workspace_tests_edges_node_executions {
-  __typename: "TestExecution";
-  errors: string[] | null;
-  createdAt: any;
-  result: string;
-}
-
 export interface GetTestsForWorkspace_node_Workspace_tests_edges_node {
```
bvaughn, line 14: > it splits the work into two separate queries. `GetTestsForWorkspace` which is responsible for getting the test names, and `GetTestForWorkspace` which is responsible for getting the execution/replay data for the selected workspace.

If this is just used to get test names then it shouldn't be named "get tests" because that's confusing.

jaril: Renamed this to `GetTestPreviewsForWorkspace` for now to keep them separate!

jaril:
@bvaughn ready for another look!

packages/shared/test-suites/TestRun.ts
```diff
@@ -53,6 +56,8 @@ export type TestExecution = {
   errors: string[] | null;
   createdAt: string;
   result: string;
+  commitTitle: string;
+  recordings: (Recording | null)[];
```
callingmedic911, line 60:
```suggestion
  recordings: Recording[];
```

`null` Recordings would be filtered now.

jaril: good catch — thanks!

callingmedic911:
I'm not sure if it should be in the scope of this PR but having a loading indicator when listing tests would be nice. Loading from a large workspace can take a bit of time.

src/ui/components/Library/Team/View/Tests/Overview/TestOverviewContent.tsx
```diff
@@ -13,19 +15,9 @@ export function TestOverviewContent() {
...
-              {test.title}
-            </div>
-          </div>
-          <Stats failureRates={test.failureRates} />
-          <TestErrorList errorFrequency={test.errorFrequency} executions={test.executions} />
-        </div>
-      );
+  if (testId && test) {
+    if (!hasFilters) {
+      children = <TestOverview testId={testId} />;
```
bvaughn, line 20: This looks broken to me.

I think this means if you type text to filter the list, then select a test, the 3rd column will show nothing.

Why `!hasFilters`

bvaughn (changes requested):
Just making it clear that this one should not merge.

src/ui/components/Library/Team/View/Tests/Overview/TestOverviewContent.tsx
jaril: This was unintentional — it's fixed now!

bvaughn:
Needs to be rebased.

jaril:
@bvaughn closed this since #9902 and #9908 really should be one PR. Will continue the discussion there!

PR #9904, opened 2023-11-17
Fetch DOM elements using Pause.evaluateInGlobal

bvaughn:
# [Loom overview](https://www.loom.com/share/e2907658fd2f4b8c853b1494a7449651)

* Fetch Elements panel data using [`Pause.evaluateInGlobal`](https://static.replay.io/protocol/tot/Pause/#method-evaluateInGlobal) rather than [`Pause.getObjectPreview`](https://static.replay.io/protocol/tot/Pause/#method-getObjectPreview).
  * Convert DOM to an array of numbers representing
  * Tag names and attributes (the ones we want to send) get de-duped into a string table
* Update Elements panel to consume newer data format
* (Temporarily disable old `ElementsListData` tests as these are no longer relevant)

Example of how quickly data can be fetched this way:

https://github.com/replayio/devtools/assets/29597/7d74e449-d665-4bc8-b3bb-5d8752740e32

---

- [ ] Make sure e2e tests pass with new Elements panel data attributes.
Note that I think they are currently failing because they were recorded with Gecko and so they don't support the `__RECORD_REPLAY_ARGUMENTS__` APIs the Elements panel is now using.


bvaughn:
Closing in favor of #9936

PR #9905, opened 2023-11-17
Add `backend_deprioritizeTraffic` setting and use in E2E tests

markerikson:
This PR:

- Adds a new `backend_deprioritizeTraffic` config setting, which is _not_ exposed through the UI
- Updates our `startTest` helper to always add that param, so that the backend knows to handle our E2E test sessions at a lower priority


markerikson:
Apparently it helps if I remember to commit the config definition first :)

PR #9906, opened 2023-11-17
Experiment with auto-batching Pause.getObjectPreviews requests

bvaughn:
Note this currently fails because the backend doesn't fully implement `Pause.getObjectPreviews`


bvaughn:
This approach is slower over all so I'm going to close the PR for now.

I think what we're seeing is that combining the waterfall (serialized requests) with the debounce delay needed to batch requests actually makes things worse overall. That's not too surprising.

Even if we removed the debounce overhead (by explicitly calling get-previews for groups of children)– I wouldn't expect it to make a big performance improvement. It may still make things a little worse because the slowest child would then hold up the entire group, which may slow further requests down.

PR #9908, opened 2023-11-17
Use the new stats endpoint for displaying the failure rate

jaril:
This PR:
-  handles the backend perf issues by separating the way we fetch all the tests for a workspace (`GetTestPreviewsForWorkspace`), with the way we fetch the executions and recordings for that workspace (`GetTestForWorkspace`)
- calculates the failure rate for any test using the new `stats` endpoint, instead of requiring all of the executions for a given test
- displays working links to passing/failing replays for a selected error for a test (see `TestDetails`)


src/ui/components/Library/Team/View/Tests/Overview/TestErrors.tsx
```diff
@@ -0,0 +1,115 @@
...
+      onClick={() => setSelectedError(msg)}
+    >
+      <div className="flex space-x-3 overflow-hidden ">
+        <div className="flex h-5 w-8 shrink-0 items-center justify-center rounded-md bg-[#F02D5E] text-xs font-bold text-white">
+          {executionCount}
+        </div>
+        <div className="overflow-hidden overflow-ellipsis whitespace-nowrap">{msg}</div>
+      </div>
+      <div className="shrink-0 overflow-hidden overflow-ellipsis whitespace-nowrap opacity-50">
+        ({replayCount})
```
ryanjduffy, line 111: I didn't realize this was the replay count. Leaning towards only showing the replay count (vs the execution count) and filtering out executions that do not have replays.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
```diff
@@ -174,11 +247,13 @@ function ErrorReplayList({
             <Icon className={styles[status]} type="play-unprocessed" />
```
ryanjduffy, line 247: We should be getting the processed state to show the correct icon version.

packages/shared/graphql/generated/GetTestPreviewsForWorkspace.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// ====================================================
+// GraphQL query operation: GetTestPreviewsForWorkspace
+// ====================================================
+
+export interface GetTestPreviewsForWorkspace_node_Recording {
+  __typename: "Recording";
+}
+
+export interface GetTestPreviewsForWorkspace_node_Workspace_tests_edges_node_stats {
```
bvaughn, line 14: I think GraphQL queries, like method names, should have semantic meaning.

What does a "test preview" mean in this case? You're loading these into a thing called a `testCache` which seems like it should contain "tests" but then you have a different GraphQL query called `GetTest...` so what is a "test preview" and why does that go into a "test cache"?

I think the names we choose for this stuff are important because other people will have to understand and maintain this code.

src/ui/components/Library/Team/View/Tests/Overview/ReplayList.tsx
```diff
@@ -0,0 +1,99 @@
...
+    "desc"
+  );
+
+  let children = null;
+
+  if (!sortedReplays.length) {
+    children = <div>No replays found</div>;
+  } else {
+    children = (
+      <>
```
bvaughn, line 25: nit: You don't need `<>` wrappers around a single node or array.

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -76,6 +77,7 @@ export function TestsContextRoot({ children }: { children: ReactNode }) {
         setFilterByText,
         testId: deferredTestId,
         testIdForDisplay: testId,
+        selectedTest: testId ? tests.find(t => t.testId === testId) ?? null : null,
```
bvaughn, line 80: Pretty sure I said this on the other PR but maybe it didn't make it through for some reason.

Wrap this context value in memo (`value = useMemo(({...}), [...])`); this is just a good practice to avoid scheduling unnecessary udpates.

src/ui/components/Library/Team/View/Tests/Overview/TestErrorList.tsx
ryanjduffy: I added this in

packages/shared/graphql/generated/GetTestPreviewsForWorkspace.ts
ryanjduffy: I renamed this to `GetWorkspaceTests`

bvaughn: I appreciate that but I'm not sure it's any clearer. Now we have `GetTestForWorkspace` and `GetWorkspaceTests`.

jonbell-lot23:
Hi @ryanjduffy ! I'm going to note a few UX things for my own notes when I loop around and tweak them in a design PR:

* As you mentioned in https://linear.app/replay/issue/DES-997/passed-and-failed-icons-are-aligned-a-bit-differently, the icons are off
* We moved our gutters from w-2 to w-1
* The right column should have a hover state on each row
* The header on the right column should be a bit taller to match the left column's header height and bottom border
* The unprocessed icons need a tooltip to explain what outline vs not outline means
* We might want a bottom border on each row on the second column, it's a bit floaty and on wide screens it's hard to align the timestamp with the content

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -76,6 +82,7 @@ export function TestsContextRoot({ children }: { children: ReactNode }) {
         setFilterByText,
         testId: deferredTestId,
         testIdForDisplay: testId,
+        selectedTest,
```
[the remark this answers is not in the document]
ryanjduffy: I misunderstood your previous comment. I can memoize the whole thing.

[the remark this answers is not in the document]
bvaughn: Sorry. Maybe it wasn't very clear.

PR #9915, opened 2023-11-20
Revert `backend_deprioritizeTraffic` setting

bvaughn:
This reverts commit a078de15403193c19aded85853b0026ef744d5b9 (#9905)

### DIFF
--- a/packages/e2e-tests/helpers/index.ts
+++ b/packages/e2e-tests/helpers/index.ts
@@ -99,16 +99,7 @@ export async function startTest(
   const base = process.env.PLAYWRIGHT_TEST_BASE_URL || "http://localhost:8080";
 
   const params = new URLSearchParams();
-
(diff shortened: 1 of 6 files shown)


markerikson:
Do we actually need to revert the _entire_ commit? The simplest would be to remove the query param setting from the E2E test setup.

bvaughn:
> Do we actually need to revert the entire commit? The simplest would be to remove the query param setting from the E2E test setup.

I guess, but why? Doesn't seem like much value in keeping around a param that we can't use bc it breaks things. (If we want to re-add it we could revert the revert)

markerikson:
I've always tended towards keeping changes where possible rather than reverting commits, I guess.  If you think it makes more sense to revert the whole thing, go for it - I don't think any of the other tweaks in there explicitly need to be kept around.

bvaughn:
In this case, based on the chat with Collin earlier, I'm not sure if we'll want to re-add the feature, so I'm going to just revert the whole thing.

PR #9916, opened 2023-11-21
Add additional telemetry

jasonLaster:
This PR adds additional telemetry in a couple of places where we had gaps

1. Loading test step details
2. Loading the element tree and an element node. This will likely be replaced soon.
3. Redux jump to code.

One of the common themes with these changes is that we can't track performance at the protocol level because it's many requests. Other performance sensitive areas like the print statement pane or react component picker are a single protocol call like `findPoints` or `getObjectPreview`.

### DIFF
--- a/packages/replay-next/components/elements/ElementsListData.ts
+++ b/packages/replay-next/components/elements/ElementsListData.ts
@@ -10,6 +10,7 @@ import { loadNodeSubTree } from "replay-next/components/elements/utils/loadNodeS
 import { shouldDisplayNode } from "replay-next/components/elements/utils/shouldDisplayNode";
 import { GenericListData } from "replay-next/components/windowing/GenericListData";
 import { ReplayClientInterface } from "shared/client/types";
+import { recordData as recordTelemetryData } from "replay-next/src/utils/telemetry";
 
 import { Item, Metadata } from "./types";
 
@@ -80,6 +81,7 @@ export class ElementsListData extends GenericListData<Item> {
   async loadPathToNode(leafNodeId: ObjectId) {
     let idPath;
     try {
+      const startTime = Date.now()
       this.updateLoadingState(true);
 
       idPath = await parentNodesCache.readAsync(this._replayClient, this._pauseId, leafNodeId);
@@ -132,6 +134,16 @@ export class ElementsListData extends GenericListData<Item> {
 
       const leafNode = this.getMutableMetadata(leafNodeId).element.node;
       const index = this.getIndexForNode(leafNodeId, leafNode);
+
+
+      const stopTime = Date.now();
+      recordTelemetryData("elements-panel-load-node", {
+        duration: stopTime - startTime,
+        params: {
+          nodeId: leafNodeId,
+          pauseId: this._pauseId
+        },
+      });
 
       return index;
     } catch (error) {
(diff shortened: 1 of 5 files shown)


packages/replay-next/components/elements/ElementsListData.ts
```diff
@@ -214,8 +226,7 @@ export class ElementsListData extends GenericListData<Item> {
         const indentation = "  ".repeat(depth);
         const nodeName = element.node.nodeName.toLowerCase();
         rows.push(
-          `${indentation} ${id}:${nodeName}${
-            childrenCanBeRendered ? "" : " *"
+          `${indentation} ${id}:${nodeName}${childrenCanBeRendered ? "" : " *"
```
bvaughn, line 229: Your editor doesn't seem to be properly configured to respect the Trunk/Prettier formatting. This change is a regression which is why [Lint is failing on your PR](https://github.com/replayio/devtools/actions/runs/6938390329/job/18874032684?pr=9916).

packages/replay-next/components/elements/ElementsListData.ts
```diff
@@ -416,7 +427,9 @@ export class ElementsListData extends GenericListData<Item> {
   }
 
   private async loadAndProcessNodeSubTree(relativeRootId: ObjectId, numLevelsToLoad: number = 0) {
+    console.log('load sub tree', relativeRootId == this._rootObjectId)
```
bvaughn, line 430:
```suggestion
```

packages/replay-next/components/elements/ElementsListData.ts
```diff
@@ -439,6 +452,14 @@ export class ElementsListData extends GenericListData<Item> {
...
     this.updateLoadingState(false);
     this.invalidate();
+
+    if (relativeRootId === this._rootObjectId) {
+      const stopTime = Date.now();
+      recordTelemetryData("elements-panel-load-tree", {
+        duration: stopTime - startTime,
+        params: {},
+      });
+    }
```
bvaughn, line 462: I don't think the way you have logging setup in this panel makes sense. For example, logging the same event in both `loadAndProcessNodeSubTree` and `loadPathToNode` – when `loadAndProcessNodeSubTree` is called by `loadPathToNode`.

I'd prefer you remove this logging for now. If we want to add logging here, I think I should probably add it (and I'd prefer to add it after we decide about #9904.)

packages/replay-next/src/suspense/FocusIntervalCache.ts
```diff
@@ -51,7 +53,18 @@ export function createFocusIntervalCache<
...
+        const startTime = Date.now();
+        const response = await options.load(start, end, ...paramsWithCacheLoadOptions);
+        const stopTime = Date.now();
+        if (options.telemetry) {
+          recordTelemetryData("suspense-cache-load", {
+            duration: stopTime - startTime,
+            label: options.debugLabel,
+            params: { start, end, ...paramsWithCacheLoadOptions },
+          });
+        }
```
bvaughn, line 66: This would be a very noisy metric for us to collect, as this cache runs *a lot*. E

bvaughn (changes requested):
I'm not sure that the "suspense-cache-load" makes sense to re-use the way this PR does.

I think we should probably file an FE issue for this and give it some more thought.

PR #9918, opened 2023-11-21
Add concurrency guards to the seek() thunk

hbenl:


src/ui/actions/timeline.ts
```diff
@@ -236,17 +242,23 @@ export function seek({
...
       // If no ExecutionPoint provided, map time to nearest ExecutionPoint
       if (executionPoint == null) {
-        time = await clampTime(replayClient, time);
+        const clampedTime = await clampTime(replayClient, time);
+
+        const nearestEvent = mostRecentPaintOrMouseEvent(clampedTime);
+        const timeStampedPoint = await replayClient.getPointNearTime(clampedTime);
+        if (getTime(getState()) !== time) {
+          return;
+        }
```
bvaughn, line 251: What is the primary point of all of the added "if !== return" checks? Are you trying to account for something else updating Redux state in between these async actions? Seems kind of odd that we're checking different things if that's the case. I think you should probably add some comments to each of these.

hbenl: Yes, this is about `seek()` being called a second time (with different arguments) while it's still running. In each check we use the most accurate information that we have available: first we only have `time`, then `executionPoint` and finally `pauseId`.

bvaughn: Hmm. I think it might be possible that the time check fails, for example...because we refine our point-to-time mapping as we go.

Async actions like this make me a little nervous. :) I wonder if we should use a more explicit lock here?

```js
const lock = {};

setSeekLock(lock);

// Async stuff...

if (getSeekLock(getState()) !== lock) {
  return;
}
```

hbenl: > I think it might be possible that the time check fails, for example...because we refine our point-to-time mapping as we go.

We first set the requested time in redux (so that the timeline is updated immediately) and only update it to the refined time when we have a corresponding execution point.

hbenl: I'll try adding an explicit lock like you suggested.

bvaughn (approved):

PR #9922, opened 2023-11-21
Use bounding client rects for node highlighting in some cases

markerikson:
This PR:

- Updates the `highlightNodes` logic to accept a flag indicating whether to use "real" box models (which results in different highlighting colors for the padding/margin/contents/border), or just the contents (as derived from bounding client rects)
- Updates different parts of the UI to set that flag accordingly:
  - Real box models: "Layout" tab with the box model rectangles display, Cypress Step Details DOM nodes, DOM Node picker
  - Client rects only: React component picker, video context menu

I can tweak these easily if we have opinions on this.  (I'm not convinced that the DOM node picker needs the full box model, but Jason seemed to suggest in Linear that was worth keeping, so it's here.)

### DIFF
--- a/packages/e2e-tests/tests/highlighter.test.ts
+++ b/packages/e2e-tests/tests/highlighter.test.ts
@@ -32,7 +32,10 @@ test("highlighter: element highlighter works everywhere", async ({
     })
     .join(" ");
 
-  // TODO:FE-805 This is a hack to get the test to pass. We should figure out why the path definition is different from the original and not use a hardcoded value.
+  // The important things to check here are 1) the path exists
+  // at all (is the box even visible on screen), and 2) is it
+  // over the iframe element we're trying to highlight.
+  // These coordinates visually match the location of the iframe.
   const pathDefinitionToCompare = `M8,46 L312,46 L312,200 L8,200`;
   expect(normalizedPathDefinition).toBe(pathDefinitionToCompare);
 });
--- a/packages/e2e-tests/tests/highlighter.test.ts
+++ b/packages/e2e-tests/tests/highlighter.test.ts
@@ -33,6 +33,6 @@ test("highlighter: element highlighter works everywhere", async ({
     .join(" ");
 
   // TODO:FE-805 This is a hack to get the test to pass. We should figure out why the path definition is different from the original and not use a hardcoded value.
-  const pathDefinitionToCompare = `M10,48 L310,48 L310,198 L10,198`;
+  const pathDefinitionToCompare = `M8,46 L312,46 L312,200 L8,200`;
   expect(normalizedPathDefinition).toBe(pathDefinitionToCompare);
 });
--- a/src/devtools/client/inspector/boxmodel/components/BoxModelMain.tsx
+++ b/src/devtools/client/inspector/boxmodel/components/BoxModelMain.tsx
@@ -131,7 +131,9 @@ export class BoxModelMain extends React.PureComponent<FinalBMMProps> {
   onShowBoxModelHighlighter = () => {
     const { selectedNodeId, highlightNode } = this.props;
     if (selectedNodeId) {
-      highlightNode(selectedNodeId);
+      // No timer, but use actual box models to show the various sections in the highlight,
+      // rather than just the contents from the bounding rect
+      highlightNode(selectedNodeId, undefined, true);
     }
   };
 
--- a/src/devtools/client/inspector/boxmodel/components/BoxModelMain.tsx
+++ b/src/devtools/client/inspector/boxmodel/components/BoxModelMain.tsx
@@ -133,7 +133,7 @@ export class BoxModelMain extends React.PureComponent<FinalBMMProps> {
     if (selectedNodeId) {
       // No timer, but use actual box models to show the various sections in the highlight,
       // rather than just the contents from the bounding rect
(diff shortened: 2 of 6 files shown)


markerikson:
The highlighter test is failing - "correctly" in that it detected a change of some kind, "incorrectly" in that the replay shows the highlighting generally working.

I looked at the underlying example app using an earlier branch.  In that branch, selecting `<div id="myiframe">` renders the highlighter twice, with these two sets of coords:

```js
[
    "M8,46.400001525878906 L312,46.400001525878906 L312,200.39999389648438 L8,200.39999389648438"
]

[
    "M10,48.400001525878906 L310,48.400001525878906 L310,198.39999389648438 L10,198.39999389648438"
]
```

I don't know _why_ that renders twice in the existing app/test behavior.  The 2px difference is maybe due to borders?

`highlighter.test.ts` then asserts the SVG path contents:

```ts

  const normalizedPathDefinition = pathDefinition
    ?.split(" ")
    .map(part => {
      const match = part.match(/^([M|L])(\d+(?:.\d+)?),(\d+(?:.\d+)?)$/);
      // Note(logan): These matches have subpixel values that fluctuate a bit in CI, so for now we force-round them.
      return match ? `${match[1]}${Math.floor(+match[2])},${Math.floor(+match[3])}` : part;
    })
    .join(" ");

  // TODO:FE-805 This is a hack to get the test to pass. We should figure out why the path definition is different from the original and not use a hardcoded value.
  const pathDefinitionToCompare = `M10,48 L310,48 L310,198 L10,198`;
  expect(normalizedPathDefinition).toBe(pathDefinitionToCompare);
```

In this branch, the highlighter renders once that I can see:

```js
[
    "M8,46.399993896484375 L312,46.399993896484375 L312,200.39999389648438 L8,200.39999389648438"
]
```

Note that those are the same as one of the sets of coords from the original branch, so the values are real, and the "failed" test does show the highlighter visibly showing up with the box in the right approximate spot:

image

Given that, I think this is still sufficiently accurate behavior, and I'm going just update the path assertion to math this.

hbenl (approved):

packages/e2e-tests/tests/highlighter.test.ts
```diff
@@ -33,6 +33,6 @@ test("highlighter: element highlighter works everywhere", async ({
     .join(" ");
 
   // TODO:FE-805 This is a hack to get the test to pass. We should figure out why the path definition is different from the original and not use a hardcoded value.
```
bvaughn, line 35: Is this TODO comment (still) relevant?

bvaughn (approved):
Overall this change looks good 😄 

Taking a step back and looking at at from a higher level though, I wonder if the way the highlight methods/args are structured makes sense with how we're actually using them?

I _think_ our codebase has 8 places that use the node highlighter:

* `ReplayWall.highlightNativeElement`:
```js
highlightNode(nodeId);
```

* `useReplayWall`:
```js
dispatch(highlightNodeAction(nodeId));
```

* `ReactDevTools`:
```js
dispatch(highlightNode(nodeId));
```

* `useVideoContextMenu`:
```js
dispatch(highlightNode(targetNodeId));
```

* `NodePicker`:
```js
dispatch(highlightNode(nodeId));
dispatch(highlightNode(nodeId, undefined, true));
```

* `BoxModelMain`:
```js
highlightNode(selectedNodeId, undefined, true);
```

* `TestSectionRow`:
```js
dispatch(highlightNodes([domNode.id], pauseId, undefined, true));
```

The last usage is the sibling `selectNode` action (also defined in `inspector/markup/actions/markup`). It is the only place that actually passes a timeout value:
```js
dispatch(highlightNode(nodeId, 1000));
```

Given the above usage patterns, I wonder if it would make sense to consider restructuring the arguments:
1. Remove the timeout argument and logic. Since `selectNode` is the only place that uses this, it can set its own "unhighlight" timer.
2. Maybe splitting the exported highlight methods into two:
  * `highlightWithBoxModels(nodeId: string)` (fetches the "real" box models)
  * `highlightWithBoundingRects(nodeId: string)`

packages/e2e-tests/tests/highlighter.test.ts
markerikson: Can probably be removed. Don't remember what the old pre-Playwright test was like, and it's not worth digging back at this point.

markerikson:
I don't think it's worth splitting up by data type, but yeah, I can move the timeout part separately.

markerikson:
Hmm. On the other hand, the logic is set up such that we clear the timer when _any_ additional call to `highlightNodes()` happens, not just around selection.  Given that, maybe it makes sense to keep it as-is.

markerikson:
Yeah, I'll remove the extraneous test comment, but I think I'm going to leave the logic as-is for now.

bvaughn:
> Hmm. On the other hand, the logic is set up such that we clear the timer when any additional call to highlightNodes() happens, not just around selection. Given that, maybe it makes sense to keep it as-is.

The caller would need to check the current node highlight (in state) before clearing, but it still feels goofy to keep an optional param, in the middle before other "more required" params, if it's only used in one place.

bvaughn:
I'm not going to block the PR or anything. I tend to think it's worth tidying up things like this as we go though, since it helps prevent clutter from organically accumulating over time and making the code harder to read/understand/maintain.

PR #9925, opened 2023-11-22
Group Playwright test steps

jasonLaster:
This PR gets a working version of Playwright test steps that you can see here [link](https://www.loom.com/share/f6a7ccb05a1447e8852d00376c35e43e) and play with in the preview branch.

Here are a couple of design decisions:
1. Merge call stacks in sequentially by finding the common frame
2. Drop commands where call stacks do not have a matching common frame

Here are a couple of technical details I learned along the way
- call stacks line and column numbers refer to the frame above
- expects are missing relevant call stacks

Next steps:
- [ ] fix types
- [ ] Show a list of top-level steps as opposed to the test file name section
- [ ] Group the before and after as well
- [ ] Fix the blue left border so that it accurately shows the progress
- [ ] Add a hover state for the functions.


bvaughn:
Closing due to inactivity.

PR #9928, opened 2023-11-25
Updated <Icon> "inspect" type to better match legacy inspect SVG

bvaughn:
Before:
Screenshot 2023-11-25 at 9 17 04 AM

After:
Screenshot 2023-11-25 at 9 22 47 AM


jasonLaster:
Nice. Maybe left align the icon too.

jonbell-lot23 (approved):
Looks good, thanks! As you mentioned in Linear and Jason mentioned in the review, it'd be nice to align these icons better. But it's not as simple as pushing this one to the right, because then it's misaligned with the text below it.

I recommend pushing this out as-is, then I can do a follow-on for additional alignment and other cleanup.

bvaughn:
> But it's not as simple as pushing this one to the right, because then it's misaligned with the text below it.

Right. The "fix" for that is more subjective and complicated. Personally I think the top icon (and maybe buttons?) would be easier to change, but that might be controversial.

PR #9930, opened 2023-11-27
LogPointPanel: don't use condition with pending edits for runEvaluation

hbenl:


packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -103,8 +104,8 @@ function PointPanel(props: ExternalProps & { focusRange: TimeStampedPointRange }
     hitPointsForLocationCache,
     client,
     { begin: focusRange.begin.point, end: focusRange.end.point },
-    pointWithPendingEdits.location,
-    pointWithPendingEdits.condition
+    pointForDefaultPriority.location,
+    pointForDefaultPriority.condition
```
bvaughn, line 108: I think this should be `pointForSuspense`, no?

bvaughn (approved):
I think this should be the right change (but I did not test/verify)

PR #9935, opened 2023-11-27
Add polish to testsuites (attempt #2)

jonbell-lot23:
Addressed these issues from [this ticket](https://linear.app/replay/issue/DES-1002/testsuites-design-polish-pass):

Fixed icon alignment
Adjusted gutters
Added a hover state
Height of headers and rows match now
Added a bottom border
Fixed left padding on left nav

Old:
image

New:
image


jonbell-lot23:
Note for @ryanjduffy because things have gotten backed up with the GitHub outage:

1. First I'm waiting for [this revert](https://github.com/replayio/devtools/pull/9934) to be ready for merge
2. I'm also waiting for a preview link here (things seem to look good on localhost)
3. Once I can confirm 1 and 2 are good, then this will be ready for review

jonbell-lot23:
Ok @ryanjduffy , this looks good to me. Ready for review!

src/ui/components/Library/Library.module.css
```diff
@@ -272,3 +274,45 @@
...
+  justify-content: space-between;
+  gap: 0.5rem;
+  border-bottom: 1px solid var(--chrome);
+  padding: 0.625rem 1rem;
+  cursor: pointer;
+  transition: background-color 0.2s;
+}
+
+.replayListTitle {
+  padding: 0.5rem 1rem 0.08rem;
```
ryanjduffy, line 309: The `0.08rem` is a very specific choice too. Do we do this elsewhere?

ryanjduffy (approved):

src/ui/components/Timeline/Timeline.css
```diff
@@ -9,13 +9,16 @@
...
   --timeline-size: 5px;
 }
 .timeline[data-protocol-timeline-enabled] {
   padding-bottom: 0;
 }
 
+.specialBorder {
+  border: 1px solid red;
+}
+
```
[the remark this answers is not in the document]
jonbell-lot23: Whoops noted

src/ui/components/Library/Library.module.css
```diff
@@ -272,3 +274,45 @@
...
 }
+
+.testTitle {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+  gap: 0.25rem; /* For 'gap-1' */
+  border-bottom: 1px solid var(--theme-border);
+  padding: 1.065em 1rem;
```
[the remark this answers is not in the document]
jonbell-lot23: Here's how it looks without this polish:
image

Here's how it looks with it:
image

If we were using two identical panels with identical structures (for example, two panels with a text header) then you'd expect the same height, and you could probably get by with a more round number. But in this case, we're matching a panel with three dropdown menus with a panel that only has text. The odd numbers aren't a sign that we're being sloppy, but actually the reverse -- the design will look (and often has looked) sloppy when we don't special case things to look aligned on the screen.

src/ui/components/Library/Library.module.css
jonbell-lot23: Same as comments above. Here's how it looks:
image

Here's how it would look if I stuck with round numbers:
image

ryanjduffy: I assumed you had a valid reason for the sizing but I do wonder if this approach addresses the symptom instead of the problem. Should we adjust the sizes or margins of the contents of these containers to be consistent instead?

jonbell-lot23: Fair question, I'm glad you asked!

So if you have two components that share most characteristics, like text inside a tab, then it's easy to keep things the same across height, padding, margin, etc. You still run into optical alignment issues like how descenders make the bottom spacing look off, but it's a fine place to start.

image

The issue is that nearly nothing is this simple. For example, if you need to position a dropdown inside a header or a tab, you're dealing with all sorts of different math. In this example, I have maintained a perfect 8px alignment all the way around the element. But it looks like garbage, because the dropdown needs its own layout needs to be met. This is an example of "consistency" making a worse design.

image

Something will have to be made inconsistent. One idea is to reduce the size of the text, so that's inconsistent but at least our precious margins are consistent! Still looks like garbage.

image

So a better thing to do here is pop out the margins so the text can stay consistent. This means the margins are "inconsistent" but they're optically well-aligned.

image

--

So that's where we are with this. If you look at the dropdowns on the left, that region is 51px tall:
image

Why 51? Because there's 1px for the border, then 8px + 8px for the top and bottom margins, and 34 pixels tall for the drop-down themselves:

image

Which means the item on the right needs to have a fixed font size (14px), aligned optically with the drop-down on the left, and we're trying to reach 51px total height. We also want it to be in REM rather than pixels so it scales correctly with the text. If you pick 1 REM, it's a bit too tall. So we season to taste.

--

In these examples, the key is consistency *in the final design that users see* rather than consistency in the code. Both is ideal, but when they clash (as they do when you're mixing components of different sizes) we need to optimise for the user, not the machine.

Whew! Design class!

bvaughn: > I assumed you had a valid reason for the sizing but I do wonder if this approach addresses the symptom instead of the problem.

This has been my feeling too when seeing odd units like this.

src/ui/components/Library/Library.module.css
```diff
@@ -272,3 +274,45 @@
...
+  gap: 0.5rem;
+  border-bottom: 1px solid var(--chrome);
+  padding: 0.625rem 1rem;
+  cursor: pointer;
+  transition: background-color 0.2s;
+}
+
+.replayListTitle {
+  padding: 0.5rem 1rem 0.08rem;
+  font-size: 1.125rem;
```
bvaughn, line 310: This is another example of a bizarre size increment.

jonbell-lot23: See previous resolved comments with Ryan. These are a natural side effect of polishing a design to make things line up. It's not a flaw in the design, and it will continue to happen.

src/ui/components/Library/Library.module.css
jonbell-lot23: I'd like to resolve this once and for all because frankly it takes a lot of time to explain and re-explain.

It's true that sometimes a strange number points to a larger problem. In our codebase, we've usually swung the other way, where our design looks sloppy because lines just barely don't line up. As said above, we need to optimise for what the user sees, which sometimes requires fine-tuning like this.

bvaughn: I wasn't asking you to re-explain FWIW. We just disagree here.

jonbell-lot23: I'd appreciate fewer comments about my "bizarre" layouts and my "arbitrary" numbers. The fact is our app looks broken when basic layouts don't line up. You could even say it looks "bizarre" to users.

bvaughn: I leave these comments because in my experience of maintaining frontend apps, atypical values are difficult to maintain and the complex relationships between various padding/margin/styles that require those values is often also easy to break.

bvaughn: No worries at all Jon. We all say things sometimes in the heat of the moment. I appreciate your follow up message!

PR #9936, opened 2023-11-27
New Elements panel (Chromium only)

bvaughn:
For recordings made with Chromium we'll use a new Elements panel that fetches DOM information using [`Pause.evaluateInGlobal`](https://static.replay.io/protocol/tot/Pause/#method-evaluateInGlobal) (rather than [`Pause.getObjectPreview`](https://static.replay.io/protocol/tot/Pause/#method-getObjectPreview)). This is more similar to how the React DevTools panel works. (See this [Loom overview](https://www.loom.com/share/e2907658fd2f4b8c853b1494a7449651) for more info.)

For recordings made with Gecko we will continue to use old Elements panel.

Here are some example recordings you can try out on this preview branch:
* New: https://devtools-git-new-chromium-elements-panel-recordreplay.vercel.app/recording/1357d3e1-b50e-4bc2-b683-9fc1add670fd
* Old: https://devtools-git-new-chromium-elements-panel-recordreplay.vercel.app/recording/e1c217c7-e872-4d44-af29-ae4b7a6eae57

Note this diff is much larger than it would otherwise be due to forking the old Elements panel. (See #9904 for a diff if we replaced the old Elements panel instead.) Eventually when we drop Gecko support we can delete the old Elements code.

https://github.com/replayio/devtools/assets/29597/7d74e449-d665-4bc8-b3bb-5d8752740e32

---

Note that many of our e2e tests don't exercise the new Elements panel (because the recordings do not support the object id lookup methods). Some of them do (like `react-devtools_03`) and I've manually copied a couple others (`inspector-elements-01_basic-dom-tree-chromium.test.ts` and `inspector-elements-02_node-picker-chromium.test.ts`) to make sure we have basic coverage of the new panel as well.

---

cc @jonbell-lot23 since this is a very visible user-facing change. Most notable changes for the new Elements panel:
* Loads and displays all dom nodes at once (so you don't have to manually expand them to drill in)
  * Expands the `<body>` tag and all children by default
  * Collapses the `<head>` tag by default
* ~~Does not display text nodes~~
* ~~Only shows `class`, `id`, and `data-*` attributes inline~~

Note that I think we can actually remove the last two constraints (see #9943).

| Old panel | New panel |
| :--- | :--- |
| Screenshot 2023-11-29 at 9 40 40 AM | Screenshot 2023-11-29 at 9 41 30 AM |


packages/replay-next/components/elements-new/ElementsPanel.tsx
```diff
@@ -0,0 +1,206 @@
...
+        event.stopPropagation();
+
+        assert(listData != null);
+
+        if (query === "") {
+          setSearchState(null);
+        } else if (searchState?.query === query) {
+          if (searchState.results.length === 0) {
+            return;
+          }
```
bvaughn, line 102: I should copy this fix to the old Elements panel.

bvaughn:
I think this is ready for review.

packages/replay-next/components/elements-new/utils/serialization.ts
```diff
@@ -0,0 +1,268 @@
...
+    // All node types should pass along id attribute
+    if (id) {
+      attributes.push(2);
+      attributes.push(registerString("id"));
+      attributes.push(registerString(id));
+    }
+
+    // All node types should pass along class attribute
+    // Class names are special; because of style reuse, we serialize individual class names separately
+    if (classList != null && classList.length > 0) {
```
markerikson, line 191: :speech_balloon: yeah, I figured it would be worth reusing classnames

markerikson (approved):

packages/replay-next/components/elements-new/utils/serialization.ts
bvaughn: Given that #9943 is now sending all attributes _and_ text content, this ends up being an overkill optimization.

bvaughn:
This PR was merged as part of #9943.

PR #9943, opened 2023-11-29
New Elements panel w/ all attributes (Chromium only) 

bvaughn:
Builds on top of #9936, but includes all node attributes _and_ as well as text nodes. (Seems like it's still pretty fast if we just send all attributes _and_ text content through.)

| Before (#9936) | After |
| :--- | :--- |
| Screenshot 2023-11-29 at 2 14 05 PM | Screenshot 2023-11-29 at 2 12 44 PM |


packages/replay-next/components/elements-new/utils/serialization.ts
```diff
@@ -179,75 +189,49 @@ export function serializeDOM(rootNode: Document): number[] {
...
-      attributes.push(registerString("class"));
-      classList.forEach(className => {
-        const classNameKey = registerString(className);
-        attributes.push(classNameKey);
-      });
-    }
-
-    // All node types should pass along data-* attributes
     if (typeof domNodeOrText.getAttributeNames === "function") {
       const attributeNames = domNodeOrText.getAttributeNames();
```
markerikson, line 193: :speech_balloon: huh, I did not know that `.getAttributeNames()` existed

jazzdan:
@bvaughn wrong reviewer with @ajdehn-eden I imagine?

bvaughn:
@jazzdan GitHub has a race case with the assignee type-ahead that sometimes causes me to select @ajdehn-eden when going for Mark/Holger. I usually catch it but this time I didn't. Thanks.

markerikson (approved):

PR #9946, opened 2023-11-29
Fix flaky icon color

ryanjduffy:
Since the backend returns `flaky` as a status for tests now, the condition used to select the label (and therefore the CSS class) for the icon needs to be flipped from `testPassed` to `testFailed`. `testPassed` returns true for both passed and flaky tests but we want different classes in these cases. All failed tests are red so its safe to use that check instead.

image

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -17,7 +17,7 @@ import {
   useFileNameTree,
 } from "ui/components/Library/Team/View/TestRuns/Overview/useFileNameTree";
 import { TestRunsContext } from "ui/components/Library/Team/View/TestRuns/TestRunsContextRoot";
-import { TestGroup, testPassed } from "ui/utils/testRuns";
+import { TestGroup, testFailed } from "ui/utils/testRuns";
 
 import { TestResultListItem } from "./TestResultListItem";
 import styles from "../../../../Library.module.css";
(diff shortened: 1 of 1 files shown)


bvaughn:
Should probably browser Frontend E2E Tests project to smoke test these changes.

Now it's flip flopped
Screenshot 2023-11-29 at 2 34 17 PM

ryanjduffy:
@bvaughn Maybe I misunderstood the problem because it's rendering like I expected. What colors are the the flaky tests icons supposed to be?

bvaughn:
They're supposed to look like they did before, yellow for the failed flake and green for the passed flake.

ryanjduffy:
> They're supposed to look like they did before, yellow for the failed flake and green for the passed flake.

Clearly they're supposed to look like they did before ...

bvaughn:
I can't point at an example of what it's supposed to look like because the older revisions of our codebase are no longer showing pass/failed/flakes correctly after the more recent DB changes.

Screenshot 2023-11-29 at 3 01 00 PM
Screenshot 2023-11-29 at 3 00 52 PM

Just imagine these are stacked on top of each other.

In a flaked set of tests, failures come first (yellow) and then the pass (green).

ryanjduffy:
Closing this out because this is the wrong approach to the problem. The real issue is that the backend is returning the test as "flaky" now with multiple recordings for each retry.  From that data alone, we don't know which is the passed and which are the failed.

This worked previously because we didn't roll up the test results and instead determined that a test was flaky only on the the client with the passed and failed attempts.

PR #9949, opened 2023-11-30
Show "recording too long" warnings in the React and Redux panels

markerikson:
This PR:

- Removes a dead `recordingDuration` field from `reducers/app` (the field we _actually_ use is in `reducers/timeline`)
- Adds logic to calculate if a recording is over 3 minutes long. We know that the backend hardcodes a limit of 3 minutes, otherwise it will skip running routines
- Updates `<SecondaryToolbox>` to show the React and Redux tabs if the recording is too long
- Updates the React and Redux panels to show a "Couldn't load data because the recording was too long" warning in that case

### DIFF
--- a/packages/replay-next/src/suspense/BuildIdCache.ts
+++ b/packages/replay-next/src/suspense/BuildIdCache.ts
@@ -9,6 +9,15 @@ export const buildIdCache = createSingleEntryCache<[replayClient: ReplayClientIn
   load: ([replayClient]) => replayClient.getBuildId(),
 });
 
+// The backend has a hardcoded limit it will only load recordings
+// up to 3 minutes long when running routines, otherwise it bails out.
+// If the recording is longer than that, we _won't_ have data for
+// React, Redux, or Jump to Code.
+// This corresponds to `RoutineMaxWindowDurationSeconds` in the backend.
+// The client uses recording duration in milliseconds.
+// Use this to show a warning in the React and Redux panels.
+const MAX_RECORDING_DURATION_FOR_ROUTINES_MS = 3 * 60 * 1000;
+
 export type RecordingCapabilities = {
   supportsEagerEvaluation: boolean;
   supportsElementsInspector: boolean;
@@ -17,6 +26,7 @@ export type RecordingCapabilities = {
   supportsRepaintingGraphics: boolean;
   supportsPureEvaluation: boolean;
   supportsObjectIdLookupsInEvaluations: boolean;
+  maxRecordingDurationForRoutines: number;
 };
 
 // Target applications which can create recordings.
@@ -88,7 +98,7 @@ export const recordingTargetCache = createSingleEntryCache<
 function getRecordingCapabilities(
   recordingTarget: RecordingTarget,
   buildComponents: BuildComponents
-) {
+): RecordingCapabilities {
   switch (recordingTarget) {
     case "chromium": {
       const buildDate = buildDateStringToDate(buildComponents?.date ?? "");
@@ -102,6 +112,7 @@ function getRecordingCapabilities(
         supportsRepaintingGraphics: userData.get("protocol_chromiumRepaints"),
         supportsPureEvaluation: false,
         supportsObjectIdLookupsInEvaluations,
+        maxRecordingDurationForRoutines: MAX_RECORDING_DURATION_FOR_ROUTINES_MS,
       };
     }
     case "gecko": {
@@ -113,6 +124,7 @@ function getRecordingCapabilities(
         supportsRepaintingGraphics: true,
         supportsPureEvaluation: true,
         supportsObjectIdLookupsInEvaluations: false,
+        maxRecordingDurationForRoutines: MAX_RECORDING_DURATION_FOR_ROUTINES_MS,
       };
     }
     case "node": {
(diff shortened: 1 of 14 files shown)


src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -137,6 +140,8 @@ export function ReactDevtoolsPanel() {
     replayClient
   );
 
+  const showRecordingTooLongWarning = !mightHaveRoutines;
```
bvaughn, line 143: Whether a recording has routines or not isn't necessarily related to its length. I think I know why you're using this as a signal (it's all we have) but I'm concerned about false positives here.

src/ui/components/SecondaryToolbox/index.tsx
```diff
@@ -158,18 +160,28 @@ export default function SecondaryToolbox() {
     ? toolboxLayout !== "ide"
     : toolboxLayout === "full";
 
+  // We definitely show these tabs if we have annotations.
+  // If we don't have annotations, we want to show the tabs anyway _if_ the recording
+  // is too long, in which case the panels will show a warning message describing why.
+  const shouldShowReactTab = hasReactAnnotations || !mightHaveRoutines;
+  const shouldShowReduxTab = hasReduxAnnotations || !mightHaveRoutines;
```
bvaughn, line 167: This makes sense but also feels awkward. If a Vue or Angular user records something with Replay and sees a React component tab with this message, that's going to feel kind of janky to them.

Seems like we should ask BAC/RUN for a way to run some kind of scoped, much-smaller routine to determine if React exists at all and use that as a signal for this rather than just the length.

src/ui/components/SecondaryToolbox/react-devtools/components/ReactDevToolsPanel.module.css
```diff
@@ -90,6 +90,7 @@
   gap: 0.5rem;
   padding: 0.5rem;
   border-radius: 0.25rem;
+  font-size: 1rem;
```
bvaughn, line 93: We have variables for font size that we should be using. Probably `--font-size-regular` or `--font-size-regular-monospace`  for this?

src/ui/reducers/timeline.ts
```diff
@@ -140,3 +149,8 @@ export const isMaximumFocusWindow = (state: UIState) => {
     return false;
   }
 };
+
+export const getRecordingMightHaveRoutines = (state: UIState) => {
```
bvaughn, line 153: Naming nit: I think it might make more sense if this was called `getRecordingTooLongToSupportRoutines` then at least the relationship I mentioned above would seem more clear.

src/ui/reducers/timeline.ts
```diff
@@ -10,6 +10,15 @@ import { mergeSortedPointLists } from "ui/utils/timeline";
...
 const initialFocusWindow = getPausePointParams().focusWindow;
 
+// The backend has a hardcoded limit it will only load recordings
+// up to 3 minutes long when running routines, otherwise it bails out.
+// If the recording is longer than that, we _won't_ have data for
+// React, Redux, or Jump to Code.
+// This corresponds to `RoutineMaxWindowDurationSeconds` in the backend.
+// The client uses recording duration in milliseconds.
+// Use this to show a warning in the React and Redux panels.
+export const MAX_RECORDING_LENGTH_FOR_ROUTINES_MS = 3 * 60 * 1000;
```
bvaughn, line 20: If we have to hard-code this on the client, it seems like the sort of thing that might make more sense to put in our capabilities code.

hbenl (approved):

packages/replay-next/src/suspense/BuildIdCache.ts
```diff
@@ -9,6 +9,15 @@ export const buildIdCache = createSingleEntryCache<[replayClient: ReplayClientIn
...
 });
 
+// The backend has a hardcoded limit it will only load recordings
+// up to 3 minutes long when running routines, otherwise it bails out.
+// If the recording is longer than that, we _won't_ have data for
+// React, Redux, or Jump to Code.
+// This corresponds to `RoutineMaxWindowDurationSeconds` in the backend.
+// The client uses recording duration in milliseconds.
+// Use this to show a warning in the React and Redux panels.
+const MAX_RECORDING_DURATION_FOR_ROUTINES_MS = 3 * 60 * 1000;
```
bvaughn, line 19: Thanks 👍🏼 At least we keep this stuff more contained in one place this way.

src/ui/actions/timeline.ts
```diff
@@ -119,8 +120,18 @@ export function jumpToInitialPausePoint(): UIThunkAction<Promise<void>> {
...
+      replayClient
+    );
+
     dispatch(
-      setTimelineState({ currentTime: time, recordingDuration: time, zoomRegion: newZoomRegion })
+      setTimelineState({
+        currentTime: time,
+        recordingDuration: time,
+        zoomRegion: newZoomRegion,
+        maxRecordingDurationForRoutines,
```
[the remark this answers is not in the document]
markerikson: I was trying to keep the actual time calculation isolated to a single Redux selector, and the `recordingDuration` value is already in Redux.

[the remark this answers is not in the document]
bvaughn: > and the `recordingDuration` value is already in Redux

I guess that's fair!

PR #9954, opened 2023-12-01
Group replays in tests view into their test runs

jaril:
Right now in the Tests view, when the user selects a test, we show them a reverse-chronological list of replays for that test. 

This model isn't quite right, because some of these tests are for the same test run. It just so happens that they're for different attempts.

This PR groups the individual replays into test runs, and shows them together.


src/ui/components/Library/Team/View/Tests/hooks/useTest.ts
```diff
@@ -1,7 +1,9 @@
 import { useQuery } from "@apollo/client";
+import groupBy from "lodash/groupBy";
 import { useContext } from "react";
 
 import { GetWorkspaceTestExecutions_node_Workspace } from "shared/graphql/generated/GetWorkspaceTestExecutions";
+import { GroupedTestRun, TestExecution, TestWithExecutions } from "shared/test-suites/TestRun";
```
callingmedic911, line 6:
```suggestion
import { GroupedTestRun } from "shared/test-suites/TestRun";
```
These two are unused.

callingmedic911 (approved):
Looks good. Left a lint comment. Also, can rebase the branch?

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
```diff
@@ -1,13 +1,107 @@
...
+  );
+}
+
+function TestRun({ testRun }: { testRun: GroupedTestRun }) {
+  const executionsWithReplays = testRun.executions.filter(
+    e => e.recordings.length && (testPassed(e) || testFailed(e))
+  );
+
+  if (!executionsWithReplays.length) {
+    return null;
```
bvaughn, line 28: Under what circumstance would this happen?

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
```diff
@@ -1,13 +1,107 @@
...
+      <div className="flex flex-row gap-2 overflow-hidden">
+        <StatusIcon status={execution.result} isProcessed={recording.isProcessed} />
+        <div className="flex flex-col overflow-x-hidden">
+          <div
+            title={title}
+            className="flex-grow overflow-hidden overflow-ellipsis whitespace-nowrap font-bold"
+          >
+            {title}
+          </div>
+          <div className="overflow-hidden overflow-ellipsis whitespace-nowrap">
```
bvaughn, line 95: I think this should include `"shrink-0"`

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
```diff
@@ -1,13 +1,107 @@
...
+          title={
+            i == remainingReplays.length - 1
+              ? "Initial attempt"
+              : `Retry ${remainingReplays.length - 1 - i}`
+          }
+        />
+      ))}
+    </div>
+  );
+}
```
bvaughn, line 54: This is what's being rendered. Is this what it should look like?
Screenshot 2023-12-04 at 3 27 32 PM

Looks a little odd to me.
* Top row has more vertical padding than the others
* Top row renders as a flex column but the others render as flex rows, which I think would look weirder if the title was short:
Screenshot 2023-12-04 at 3 28 40 PM

Making the padding and flex direction more consistent helps a bit I think but then the info seems a bit confusingly presented:
Screenshot 2023-12-04 at 3 31 08 PM

I wonder if it wouldn't be clearer to pull the commit title out so it's separate from the retries?
Screenshot 2023-12-04 at 3 33 17 PM

I haven't seen the UI spec for this but I'm wondering how we landed on the spect that's implemented here.

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
jaril: Test executions past the set retention limit (eg, >7 days) would return the execution, but have no replays attached to it.

For the next batch of work, we're trying to figure out a better model of displaying what should be still-useful data (text execution results) even if it's past the retention limit and doesn't have any replays attached. But for now, this UI/model is replay-centric, so those just get dropped off

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
jaril: The original design:
image

jaril: @jonbell-lot23 I made some (light) alterations here from the original spec, but I think something bigger that I've noticed and Brian's also pointed out is that the way the information's presented here is somewhat confusing.

In this example:
image

It's not clear to me that there were 4 distinct attempts for this test run, and the link to open the latest one is the first row with "Merge 961b…". I think what would be clearer is to group by the test run, then have all four attempts displayed clearly below. For example, for a test run that had 3 attempts, it would look something like

```
-------------------
My commit message goes here
3h ago
-------------------
▶️ Attempt 2
▶️ Attempt 1
▶️ Initial attempt
-------------------
```

jonbell-lot23: Yup, this sounds good to me!

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
bvaughn: Could we filter this stuff out from our GraphQL request (so we aren't fetching data to the client that we don't show)?

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
bvaughn: Some important info seems to have gotten dropped between Jon's mock and the "alterations" –
* The top row should have the user who triggered the run
* The top row should have the the time when the (last?) test attempt was
* Individual rows should have _duration_ not _time ago_ times

jonbell-lot23: Yeah, I should clarify my thumbs up above.

I'm saying the idea of putting metadata up top and attempts below that seem good to me. I have not caught up on the whole discussion and I am sure there are other details that I can find once a do more a more complete NUX pass. Bear with me, I have a lot to catch up on after my concussion.

bvaughn:
Closing due to inactivity.

src/ui/components/Library/Team/View/Tests/Overview/TestDetails.tsx
```diff
@@ -1,13 +1,93 @@
...
+  return (
+    <div className="flex flex-col rounded-md bg-chrome">
+      {sortedExecutions.map((e, i) => (
+        <Execution execution={e} key={i} />
+      ))}
+    </div>
+  );
+}
+
+export function getReplayResult(result: string, index: number, total: number) {
```
jaril, line 51: @ryanjduffy This makes a _lot_ of assumptions, so feel free to flag anything I'm missing here.

ryanjduffy (commented):
I think there are a handful of changes we should make but nothing too significant. I don't want to request changes since I won't be around to re-review so I'll leave this as comments and trust that you, @jonbell-lot23 , and @callingmedic911 will sort it out next week.

tailwind.config.js
```diff
@@ -57,6 +57,8 @@ module.exports = {
         testsuitesErrorBgcolor: "var(--testsuites-error-bgcolor)",
         testsuitesErrorBgcolorHover: "var(--testsuites-error-bgcolor-hover)",
         testsuitesErrorColor: "var(--testsuites-error-color)",
+        testsuitesFailedColor: "var(--testsuites-v2-failed-header)",
+        testsuitesFlakyColor: "var(--testsuites-v2-flaky-header)",
```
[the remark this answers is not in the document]
jonbell-lot23: Yeah, we have a broader tailwind cleanup to do, so this is totally fair game for now. Fixing this now pulls in a much larger job and that's not needed at this point.

PR #9955, opened 2023-12-01
Implement Playwright test step DOM node support

markerikson:
This PR:

- Updates `TestRecording` objects to include the `testRunnerName`, so we don't have to have a separate `TestSuitesCache.read()` call to pass that down
- Updates `RecordingTestMetadata` to calculate a time range for Playwright recordings based on the first and last test step points
- Restructures `TestEventDetailsCache` to segment the logic by test runner
- Updates `ReplayClient` to allow passing in the `preloadedExpressions` option supported by the protocol
- Implements support for fetching DOM nodes for Playwright test steps:
  - Pastes in the `injectedScriptSource.js` bundle that Playwright itself uses to implement the locator selector engines in a real browser in a test
  - Uses the injected selector engine in an eval to parse the locator string for each step, and query the DOM for the relevant DOM nodes
  - Also collects what _should_ be all intermediate DOM nodes for each sub-segment of the locator string, but does not use those yet
  - Extracts the first found DOM node from the results and caches it for use in highlighting
- Updates `TestSectionRow` to use the appropriate execution point for looking up the DOM node to hover-highlight

AND IT WORKS!

Here's an example of me hovering over a Playwright step that tries to select our source panel:

image


hbenl (approved):

bvaughn (commented):
I know this has a ✅ from Holger, and we normally only require one ✅ to land, but I'd like the chance to review too? can you hold off merging until then?

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -275,3 +351,233 @@ async function fetchAndCachePossibleDomNode(
...
+  // Controls injection into window, and some other external pieces
+  const isUnderTest = false;
+  // Playwright can be used from multiple languages
+  const sdkLanguage = "javascript";
+  // data-testid can be customized by the user
+  const testIdAttributeName = "data-testid";
+  // WebKit on Windows needs 5, everything else is 1
+  const rafCountForStablePosition = 1;
+  const browserName = "chromium";
+  const customSelectorEngines: string[] = [];
```
bvaughn, line 382: Some of this stuff feels kind of scary in terms of the coupling, like `rafCountForStablePosition`

markerikson:
Yep, not in a hurry to merge this - just noting that it actually does work as I've intended so far.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -275,3 +351,233 @@ async function fetchAndCachePossibleDomNode(
...
+
+  // Any focus range errors here will bubble up to the parent focus cache,
+  // which will retry the load if needed later.
+  await Promise.all(
+    // TODO We don't currently have a way for `runEvaluation` expressions to have variable behavior
+    // per execution point. The old analysis API put `point` and `time` in scope, and if we had
+    // that we could do _one_ `runEvaluation` with a single expression that would contain a lookup
+    // table mapping execution points to locator strings.
+    // Since we don't have that, for now we'll run a separate `runEvaluation` _per point_. This is horrible
+    // for perf, but it at least will work. (Slowly.)
```
bvaughn, line 408: I think we should maybe hold off on this change for this reason alone.

bvaughn (changes requested):
I think we should talk more about this before landing.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: I have a feature request in to the BAC team to expose the current execution point as a variable in scope in evaluations. The old analysis API did that, and `runEvaluation` doesn't. That would let me do one `runEvaluation`, with an injected lookup table of locator strings per execution point.

We _could_ potentially look at using a routine here, although there was something I thought of while working on this that made me think we'd have issues converting this into a routine - can't remember that atm.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: I looked through the Playwright source. The `rafCountForStablePosition` value specifically is something where Safari on Windows does something weird, and everything else is the same.  So, Safari+Win has a value of 5, and everything else is 1.

src/ui/components/TestSuite/suspense/injectedScriptSource.js
```diff
@@ -0,0 +1,8 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
```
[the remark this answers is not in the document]
markerikson: Agreed. This whole file was a literal copy-paste out of `playwright-core/dist/generated`, but worth documenting that.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
jasonLaster: I'm also concerned about the perf implications because we're bypassing the schedulers. 

If we wanted to land something safe we'd likely need to do it on hover with a denounce to limit the calls.

markerikson:
Updated the PR with requested changes:

- Now uses a single `runEvaluation` and uses a lookup table of execution points -> locator strings, now that we have `__REPLAY_CURRENT_EVALUATION_POINT__` in scope in `runEvaluation`
- Switched to using bounding rects instead of box models for highlighting test step nodes, per FE-2109 (although I'm not sure how much benefit that will have if any)
- Extracted the evaluation logic into a separate TS function, per FE-2093

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -275,3 +358,189 @@ async function fetchAndCachePossibleDomNode(
...
+      evalResults.push(...results);
+    }
+  );
+
+  evalResults.sort((a, b) => compareExecutionPoints(a.point.point, b.point.point));
+
+  // We need to reformat the raw analysis data to extract details on the "step details" object.
+  const processedResults: TestEventDetailsEntry[] = await Promise.all(
+    evalResults.map(async result => {
+      const { pauseId, point: timeStampedPoint, returned, data } = result;
```
bvaughn, line 463: Is this an :any type? Can we use the return type information from `playwrightStepDomNodes.ts` (since it's actually a TS file)?

bvaughn (approved):

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: No, `result` is a protocol `RunEvaluationResult`, and `data` is just a standard `Value`.  (In this case, `data` points to a mixed-contents array, but at the type level all we know is it's a `Value`.)

PR #9956, opened 2023-12-02
Fix icons state for flaky tests

ryanjduffy:
* Updates graphql to fetch executions instead of just recordings since flaky tests contain both passing and failing recordings
* Updated tree view to count "replays" for file nodes and "tests" for path nodes. Fixed E2E tests to match this change.

image


src/ui/components/Library/Team/View/TestRuns/suspense/TestRunsCache.ts
```diff
@@ -61,12 +61,29 @@ export const testRunDetailsCache = createCache<
...
+            // Adapt the execution status to convert "failed" execution status
+            // to "flaky" when the test eventually passes
+            let result = e.result;
+            if (test.result === "flaky") {
+              if (testFailed(e)) {
+                result = "flaky";
+              } else if (testPassed(e)) {
+                result = "passed";
+              }
+            }
```
bvaughn, line 80: Moving the failed vs flaky logic out of the client feels kind of like taking a step forward in one place and a step backward in another place, if we have to deconstruct it like this.

ryanjduffy: I was taking a short cut to show the expected UI for flaky tests by adjusting the status so the UI didn't have to know if the test was "flaky" to show the flaky icon for a failed test. If you'd prefer, I can move a version of this `if` to the view component instead.

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -156,21 +156,29 @@ const FileNodeRenderer = memo(function FileNodeRenderer({
       >
         <Icon className="h-5 w-5 shrink-0" type="file" />
         <div className="truncate">{name}</div>
-        {!expanded && <div className="text-xs text-bodySubColor">({tests.length} tests)</div>}
+        {!expanded && (
+          <div className="text-xs text-bodySubColor">
+            ({nestedRecordingCount} {nestedRecordingCount === 1 ? "replay" : "replays"})
```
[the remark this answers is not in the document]
ryanjduffy: This is probably not right now that I think about it. I was thinking about Cypress' model of 1 recording per spec file so the contents of the nodes would all be replays of the same test. That isn't true of playwright so keeping with tests is fine.

src/ui/components/Library/Team/View/TestRuns/suspense/TestRunsCache.ts
bvaughn: I don't know that I feel strongly about where it lives. Just remarking on the fact that the frontend _used_ to have logic to do this.

PR #9957, opened 2023-12-02
Fix delta tests errors with playwright plugin bump

ryanjduffy:
Updates the playwright plugin to pull in a couple updates that handle some unexpected data from the new fixture code.


jasonLaster:
This is great!

PR #9967, opened 2023-12-05
ReferenceErrors rendering

bvaughn:
Split off from #9965

Note that pending the outcome of RUN-2972, this PR may not be necessary.

---

Chromium returns an unexpected format for reference errors:
```js
{
  result: {
    result: {
      data: {
        objects: [
          {
            className: "ReferenceError",
            objectId: "1",
            // ...
          },
        ],
      },
      exception: {
        unavailable: true,
      },
      returned: {
        object: "1",
      },
    },
  },
  // ...
}
```

So far as I know this should really only happen in two places:
* Log points
* Terminal expressions

This PR updates those two places to better render these errors (by rendering the error message rather than "(unavailable)")

### Log point
| before | after |
| :---: | :---: |
| Screenshot 2023-12-05 at 1 48 24 PM | Screenshot 2023-12-05 at 1 48 28 PM |

### Terminal expression
| before | after |
| :---: | :---: |
| Screenshot 2023-12-05 at 1 48 40 PM | Screenshot 2023-12-05 at 1 48 45 PM |

### DIFF
--- a/packages/e2e-tests/tests/logpoints-03.test.ts
+++ b/packages/e2e-tests/tests/logpoints-03.test.ts
@@ -13,10 +13,7 @@ test(`logpoints-03: should display event properties in the console`, async ({
 
   await addEventListenerLogpoints(page, [{ eventType: "event.mouse.click", categoryKey: "mouse" }]);
 
-  const message = await findConsoleMessage(page, "MouseEvent", "event");
+  const message = await findConsoleMessage(page, "(click)", "event");
 
-  await expect(message).toContainText('type: "click"');
-  await expect(message).toContainText("target: <div");
-  await expect(message).toContainText("clientX: 0");
-  await expect(message).toContainText("clientY: 0");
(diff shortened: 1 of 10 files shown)


hbenl (changes requested):
I think this is a RUN issue: according to [our protocol docs](https://static.replay.io/protocol/tot/Pause/#type-Result), the runtime should set `returned` if the operation returned normally or `exception` if the operation threw. Obviously setting both at the same time doesn't make sense.
I don't know where this protocol message came from but I guess that `exception` should be set to the `ReferenceError` and `returned` should not be set and in this case, the frontend would already show the `ReferenceError` correctly.

bvaughn:
> I don't know where this protocol message came from 

Any log point or console expression that references an undefined value will return this oddly shaped data (when using Chromium). If you think we should file a RUN issue instead, I can do that. It will further hold up our migration away from Gecko though.

PR #9968, opened 2023-12-06
Filter test runs by branch via graphql

ryanjduffy:
Blocked by https://github.com/replayio/backend/pull/9108

Filter test runs by branch via graphql by plumbing the branch name through the cache and to the graphql query

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/TestRunsContextRoot.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/TestRunsContextRoot.tsx
@@ -37,8 +37,6 @@ type FilterByStatus = "all" | "failed";
 export function TestRunsContextRoot({ children }: { children: ReactNode }) {
   const { teamId, testRunId: defaultTestRunId } = useGetTeamRouteParams();
 
-  const testRuns = useTestRuns();
-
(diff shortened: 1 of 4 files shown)


ryanjduffy:
@callingmedic911 - keeping this in draft so it can be reconciled with your test runs PR after it lands

callingmedic911 (changes requested):
Can we also have similar changes for NewTestRuns components?

bvaughn (commented):
I think this change will strictly cause more data to be requested by the frontend, since we default to "all". Currently, switching to "primary" would just be an in-memory filtering, but this would change it to a new request.

What's the motivation here? SCS-1622 just says _what_ but not _why_.

ryanjduffy:
> I think this change will strictly cause more data to be requested by the frontend, since we default to "all". Currently, switching to "primary" would just be an in-memory filtering, but this would change it to a new request.
> 
> What's the motivation here? [SCS-1622](https://linear.app/replay/issue/SCS-1622) just says _what_ but not _why_.

I think you're right, @bvaughn . I had the tests view in my mind when I first picked up this work which needs the stats recalculated by the backend for the branch change but I don't think this is adding much value for the runs view. Thanks for catching this.

PR #9970, opened 2023-12-06
[RUN-2965] Add better error reporting

Domiii:
* Make sure that we understand why and if `registerPlainValue` fails.
* https://linear.app/replay/issue/RUN-2965/fe-throws-could-not-find-object-id#comment-2784c3b8

### DIFF
--- a/packages/replay-next/components/elements-new/utils/serialization.ts
+++ b/packages/replay-next/components/elements-new/utils/serialization.ts
@@ -117,6 +117,8 @@ export function serializeDOM(rootNode: Document): number[] {
         const id = __RECORD_REPLAY_ARGUMENTS__.internal.registerPlainObject(value);
         if (id) {
           return parseInt(id);
+        } else {
+          throw Error(`[RUN-2965] registerPlainObject failed for value: ${JSON.stringify(value, null, 2)}`);
         }
       }
     }


markerikson:
I skimmed the Linear issue discussion. I'm not sure what throwing an error inside the eval is going to do specifically in terms of "error reporting".  Is this _intended_ to bubble up, get caught, and reported to telemetry somewhere? More importantly, _will_ it actually do that?

(legitimately asking - I don't know for sure. My immediate gut reaction is that we'd still get a `res.exception` field in the eval response, and then do nothing with it, unless we actually have logic set up to track and do something with those eval exceptions, and I don't know off the top of my head if we do.)

packages/replay-next/components/elements-new/utils/serialization.ts
```diff
@@ -117,6 +117,8 @@ export function serializeDOM(rootNode: Document): number[] {
         const id = __RECORD_REPLAY_ARGUMENTS__.internal.registerPlainObject(value);
         if (id) {
           return parseInt(id);
+        } else {
+          throw Error(`[RUN-2965] registerPlainObject failed for value: ${JSON.stringify(value, null, 2)}`);
```
bvaughn, line 121: I don't think there's anything that will pick up and report this error. In the case where the evaluation fails, the frontend degrades gracefully and shows "_Elements are not available_" in the Elements panel.

bvaughn: Maybe the backend will automatically log thrown errors? (If so I didn't know that.)

The frontend can't. If an evaluation throws, the response that gets sent to the frontend is:
```js
{
  exception: {
    unavailable: true,
  },
  returned: {
    object: "1",
  },
}
```

I guess the frontend could add some scaffolding of its own to do this but I think that would be clunky.

Domiii: I think you are right, that it should be easier to have RUN triage their own problems.
I guess in this case, considering `registerPlainObject` should be infallable, so making it produce a RUN warning/error if it don't work, would make sense 🤔

bvaughn: > considering registerPlainObject should be infallable, so making it produce a RUN warning/error if it don't work, would make sense

I _think_ you're saying that RUN would do this? (So it would generate its own logging from those failures?)

Sorry I'm not sure I understand. Maybe easier if we chat about this on VC briefly.

PR #9972, opened 2023-12-06
Always highlight the hovered test step's node

hbenl:
No idea why this was disabled for the selected test step.


markerikson:
Looks like it was just an oversight on my part when I was adding the hover highlight logic initially.

bvaughn (approved):

PR #9985, opened 2023-12-07
accept a linux build filename when manually starting a test run

toshok:
There's documentation online for adding fields that GH will synthesize a UI from [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).  This should let us kick off a manual test run where the e2e tests are run with the supplied chromium build.

Starting from this page: https://github.com/replayio/devtools/actions/workflows/test.yml

manually running the `test.yml` workflow from this branch I see:
image

If I then enter the following (from https://buildkite.com/replay/chromium-build/builds/2564)
`linux-chromium-20231207-cde6ff4c667b-ad0834963a29.tar.xz`, when the workflow goes to download the browser, it fetches the correct build.

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,8 @@ on:
   pull_request:
   workflow_dispatch:
     inputs:
-      runtimeBuildId:
-        description: "Runtime build ID (include a leading dash if non-empty)"
+      linuxBuildFile:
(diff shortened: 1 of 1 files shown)


toshok:
might be better to just c&p the entire linux `.tar.xz` filename instead of trying to use a substring.

jasonLaster (approved):
Looks good

PR #9986, opened 2023-12-07
include test executions without recordings

ryanjduffy:
This PR lets the frontend query test executions without recordings. There are cases where this change might confuse/mislead the user, so we've added some messaging for the following cases:

### Runs view

<details>
<summary>Screenshot</summary>
image
</details>

This is shown in the Runs view after the user clicks on a test run, then a test. It addresses cases where we have that executions for that test, but we don't have replays.

[Here's an example of a test run](http://localhost:8080/team/dzoyNTM4ZjRmOC05YmFlLTRiYjYtYjljYi1jOGYzOWUyMjRhZWY=/runs/f136850d-a847-4831-aeb3-6d688817d70e?param=dzoyNTM4ZjRmOC05YmFlLTRiYjYtYjljYi1jOGYzOWUyMjRhZWY%3D&param=runs&param=126406dc-29ba-47a2-b462-83ff18a1f019) where this case applies.

### Tests view

<details>
<summary>Screenshot</summary>
image
</details>

This is shown in the Tests view after the user clicks on a test. The failure rate is calculated from all relevant executions, whether or not they have a replay. Because of this, if the user were to match the failure rate calculation to the replay statuses (pass/fail) displayed on that list, it's possible for them to think that the failure rate is wrong. This disclaimer tries to get ahead of that reasoning.

We don't have working links to the tests view yet, but open [Metabase's runs view](http://localhost:8080/team/dzoyNTM4ZjRmOC05YmFlLTRiYjYtYjljYi1jOGYzOWUyMjRhZWY=/tests?param=dzoyNTM4ZjRmOC05YmFlLTRiYjYtYjljYi1jOGYzOWUyMjRhZWY%3D&param=runs&param=126406dc-29ba-47a2-b462-83ff18a1f019) as normal, switch to the Tests view, and any of the top rows should show the alert.


jonbell-lot23:
Slapping in some visuals to refer to 

image

jaril:
@ryanjduffy @jonbell-lot23  is it just me, or is the test view message unnecessary? 

We're only displaying replay rows, so it's not really helpful to call out that we have executions for which there are no replays. As in, there's no real concept of executions for that column so it's a new concept.

src/ui/components/Library/Team/View/NewTestRuns/Alert.tsx
```diff
@@ -0,0 +1,20 @@
+export function Alert({ children, link }: { children: string; link?: string }) {
```
ryanjduffy, line 1: Let's move this out of `NewTestRuns` since it's used by both the test and test runs views.

It might be helpful to narrow the component a bit by exposing an enum for the alert `reason` rather than the link and message. Not a big deal but it lets us centralize the messaging and URLs. Up to you.

src/ui/components/Library/Team/View/NewTestRuns/TestRunResultList.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { TestRunsContext } from "./TestRunsContextRoot";
+
+export function TestRunResultList({
+  selectedSpecTests,
+}: {
+  selectedSpecTests: TestRunTestWithRecordings[];
+}) {
+  const { testRunId } = useContext(TestRunsContext);
+  const { testRun } = useTestRunDetailsSuspends(testRunId);
+  const hasRecordings = selectedSpecTests.some(test =>
```
ryanjduffy, line 17: This is probably the code I wrote but the logic is a little convoluted to have two `some`'s and then `!hasRecordings`.

Instead, maybe:
```
const noRecordings = selectedSpecTests.all(test => test.executions.all(e => e.recordings.length === 0))
if (noRecordings) {
  return ...
}
```

src/ui/components/Library/Team/View/NewTestRuns/TestRunResultList.tsx
```diff
@@ -0,0 +1,50 @@
...
+    test.executions.some(e => e.recordings.length > 0)
+  );
```
bvaughn, line 19: You can combine this and the map/flatMap operation below into a single, memoized loop to be more efficient.

1. Ignore any entries where `execution.recordings.length === 0`
2. Return the "no replays" message if the resulting flat-mapped collection is `length === 0`

src/ui/components/Library/Team/View/NewTestRuns/Alert.tsx
```diff
@@ -0,0 +1,20 @@
...
+  let externalLink = null;
+
+  if (externalLink) {
+    externalLink = (
+      <a href={link} rel="noreferrer" target="_blank">
+        Learn more
+      </a>
+    );
+  }
+
```
bvaughn, line 11: What's this?
```suggestion
```

bvaughn (approved):
Approval aside from the suggestions.

src/ui/components/Library/Team/View/NewTestRuns/TestRunResultList.tsx
bvaughn: `.some` and `.map` are both iterating over the entire collection. We only need one `.forEach` and then it can conditionally push to a derived array.

jaril: I think that's a bit more readable — went ahead and applied it.

src/ui/components/Library/Team/View/NewTestRuns/Alert.tsx
jaril: good catch, thanks!

jaril: thanks for catching that

src/ui/components/Library/Team/View/NewTestRuns/TestRunResultList.tsx
jaril: Tried this in https://github.com/replayio/devtools/pull/9986/commits/c66d724a8e2f990d94b18ce7bd73c2e97ab47281, but because of how that component accesses that data structure, it does mean that we have to compose the replay a bit weirdly. Let me know if this is what you had in mind!

bvaughn: I don't think the commit you linked to is what I described.

I was suggested you could combine the `.some(condition)` and `.map(transform)` operations into a single `.forEach` operation, essentially

```js
const derived = [];
array.forEach(item =>
  if(condition) {
    derived.push(
      transform(item)
    );
   }
});
```

jaril: Actually I just realized, if this isn't a blocking change do you mind if I merge it without?

We're moving up SCS-1591 and reopening #9954 to work on it next, which will undo a lot of this optimization anyway.

PR #9987, opened 2023-12-08
Debounce loading the selected React component's data while the picker is active

hbenl:


bvaughn (changes requested):
This approach seems better but I think it's still potentially a bit error prone, or at least...it could give a false positive, if the Node picker was active and you used the keyboard event, for instance.

I think we can do this more robustly by just updating the selected element in the normal "high priority" way for mouse clicks and keyboard events, but debouncing the update action when it comes from the node picker. Those all come from different places so it should be easy to isolate.

Below are the call stacks for the three ways the selected item gets updated.

Mouse event:
```
[GenericListData] setSelectedIndex()
[ReactDevToolsListData] selectElement()
[ReactDevToolsListItem] onClick()
```

Keyboard event:
```
[GenericListData] setSelectedIndex()
[GenericList] onKeyDown()
```

Node picker:
```
[GenericListData] setSelectedIndex()
  <<we should add the debounce here>>
[ReactDevToolsList] onSelectFiber()
```

hbenl:
> Node picker:
>
> [GenericListData] setSelectedIndex()
> \<\<we should add the debounce here>>
> [ReactDevToolsList] onSelectFiber()

I didn't want to debounce the entire component selection, only the loading of component details in the right panel. If I had to choose between the component picker always being debounced (so I have to pause every time I want to see where I am in the component tree) or accepting the corner case that loading component details is debounced if you use the keyboard while the component picker is active, I'd choose the latter.

bvaughn:
> I didn't want to debounce the entire component selection, only the loading of component details in the right panel. If I had to choose between the component picker always being debounced (so I have to pause every time I want to see where I am in the component tree) or accepting the corner case that loading component details is debounced if you use the keyboard while the component picker is active, I'd choose the latter.

In that case, we can track the selected row and the details with two separate bits of state.

I feel somewhat strongly that focusing our debouncing to only include the codepath that is triggered by the node picker here is a better way to go.

hbenl:
> In that case, we can track the selected row and the details with two separate bits of state (and debounce the latter).

Since we want to debounce the *loading* of the details (not the details themselves) I think we'd have to track the reason why an element was selected in `GenericListData` - is that what you have in mind?

bvaughn:
> Since we want to debounce the loading of the details (not the details themselves)

TBH I'm not sure what you mean by this.

The thing we don't need/want to debounce is the updating of the selected row in the list. The loading of details should be displayed, and the details panel should probably be dimmed if ever the selected items doesn't match the item we're loading details for. (Does this make sense?)

> I think we'd have to track the reason why an element was selected in GenericListData

I don't think we'd need to do this, would we? Why would we have to do this?

hbenl:
> Why would we have to do this?

We need to decide whether to debounce the loading of the details or not. This PR uses the `isComponentPickerActive` flag for this decision but what we'd really want to use is the reason for the current selection (i.e. whether it came from the component picker or not). Where do we get that from?

bvaughn:
> We need to decide whether to debounce the loading of the details or not. This PR uses the isComponentPickerActive flag for this decision but what we'd really want to use is the reason for the current selection (i.e. whether it came from the component picker or not). Where do we get that from?

What I was trying to suggest [above](https://github.com/replayio/devtools/pull/9987#pullrequestreview-1772442847) was that we should just use a different (debounced) function in `<ReactDevToolsList>` `onSelectFiber()` for this purpose.

hbenl:
> What I was trying to suggest https://github.com/replayio/devtools/pull/9987#pullrequestreview-1772442847 was that we should just use a different (debounced) function in <ReactDevToolsList> onSelectFiber() for this purpose.

But that would debounce the selection, not just the loading of the details. I.e. if you moved your mouse over the screenshot, the component selection (in the left panel of React DevTools) would not get updated until you stopped moving the mouse for a while. I'd like to avoid that.

bvaughn:
> But that would debounce the selection, not just the loading of the details.

Right, which is why I also mentioned this above:

> In that case, we can track the selected row and the details with two separate bits of state (and debounce the latter).

We should just talk about this during the standup. Feels like we're talking past each other.

If you'd like I could put together a patch/POC showing what I have in mind.

bvaughn:
What I had in mind is roughly #10000

PR #9989, opened 2023-12-08
Use cached test runs value while loading

callingmedic911:
Closes SCS-1669.

The issue we want to fix is while switching between time filter toggle, we don't want to show empty state. We already have some cache value after first load and we can render that while it's loading. Here's loom of problem: https://www.loom.com/share/02d89044000d4fa18e9ea04a65b08414

Here's how it's looking after switching: https://www.loom.com/share/cf8fcffb3f204518833ee73c8d3c751b

_____
@ryanjduffy Before I tag frontend team, I want to make sure this is what we expect. The experience feels odd to me. Even though most of values were already in cache, the blank state makes it clear it's loading the new values (while it's done in optimal way: cached range + new range). I'd prefer to keep the existing behavior, or just add a loader than serve cached value.


jonbell-lot23:
Hey @callingmedic911 , what id do is what you mentioned:

1. Add a slight delay, 500ms is good
2. During that time, show a loader centered on the screen

It's just enough to explain the change but still fast and clear

callingmedic911:
I'm strongly against adding a manual delay.
- The subsequent request with smaller range may come back faster than 500ms.
- Or if that request is delayed, we'd have same problem: the list will load with cached data and then when that request succeeds we'd see list change.

I'd much rather keep the existing behavior (without this PR) i.e. keep showing loader while the new range request comes back. What do you think @jonbell-lot23 and @ryanjduffy?

callingmedic911:
On first load, we still see old loader in center: https://www.loom.com/share/e28c2872a0c349da9eeae6389d5f8a58
When switching between toggle - https://www.loom.com/share/69eedfe188b24edfb25639b364f0698c (response is not that slow, I just hardcoded loader so we can actually see here).

jonbell-lot23:
> On first load, we still see old loader in center: https://www.loom.com/share/e28c2872a0c349da9eeae6389d5f8a58 When switching between toggle - https://www.loom.com/share/69eedfe188b24edfb25639b364f0698c (response is not that slow, I just hardcoded loader so we can actually see here).

Looks good! I was thinking it should only be on the single pane it's loading, but this isn't so bad either. I'd change if it's super simple, but it wouldn't be a problem to keep it this way.

ryanjduffy:
The loader is absolutely positioned so we need to add a positioning context via `position: relative` on the runs panel to limit it to that area.

callingmedic911:
Yeah I noticed that but I assumed Jon meant whole screen when he said "top of screen". Anyways moved to top of Panel.

src/ui/components/Library/Team/View/NewTestRuns/hooks/useTestRuns.ts
```diff
@@ -27,7 +28,16 @@ export function useTestRuns(): { testRuns: TestRun[]; status: Status } {
     teamId
   );
 
-  const testRuns = useMemo(() => [...value].reverse(), [value]);
+  useEffect(() => {
+    if (status === STATUS_PENDING) {
+      // To avoid resetting the test runs to an empty array when loading
```
bvaughn, line 33: Seems like we should just be using _suspense_ here for this. It automatically gives you an _is pending_ signal which you can use to e.g. dim the list while you're loading new data (which we should do here btw)

bvaughn (commented):
https://www.loom.com/share/38b6403f19bf4c72b8541dc5d09ff626

src/ui/components/Library/Team/View/NewTestRuns/hooks/useTestRuns.ts
callingmedic911: My understanding of Suspense is, where Suspense is used it needs to have promise inside of it and while this promise is pending Suspense is will fallback to "fallback" component. Please correct me if I'm wrong here.

Now this useTestRuns hook is used at Context level. Even if we use .read here that Suspense is signal is not available when using the value from context deep in child level component. If I'm right so far, maybe we can refactor to NOT use testRuns via Context but it may become little painful passing testRuns value to different components.

bvaughn: You're already working with a Suspense cache. (That's what `testRunsIntervalCache` is.)

> While this promise is pending, Suspense will fall back to a 'fallback' component. Please correct me if I'm wrong.

That's only true if you're mounting something initially or if you suspend during a high priority update. [That's where `useTransition` comes in.](https://react.dev/reference/react/useTransition)

This way of using status to ignore pending updates feels kind of like you're trying to approximate what `useTransition` and Suspense give you already.

PR #9990, opened 2023-12-08
Fix showing all replays for a test execution

ryanjduffy:
Some test executions can include multiple replays (e.g. comments-02 in the FE tests) but we were only showing the first one previously. Now we'll map over all recordings and show each one.


ryanjduffy:
> This change looks fine but... speaking of `comments-02` - where is it on this branch?
> 

No idea ... opened https://linear.app/replay/issue/SCS-1674/comments-02-missing-from-test-run

PR #9991, opened 2023-12-08
Add loaders for new tests and test runs

callingmedic911:
Closes SCS-1624.

Adds loader for both new tests and test runs view.

Tests: https://www.loom.com/share/bcdd5da730144d828372f9d3d441ac3b
Test Runs: https://www.loom.com/share/a17af943103543d4893c328e98621e41 (Notice the loader is little lower than other panel's loader. This is because first panel always has input/filter at top available, this changes vertical offset a bit. I'm open to suggestions if we can improve this, cc: @jonbell-lot23)

### DIFF
--- a/src/ui/components/Library/Team/View/NewTestRuns/TestRunsPage.tsx
+++ b/src/ui/components/Library/Team/View/NewTestRuns/TestRunsPage.tsx
@@ -37,6 +37,7 @@ function TestRunsContent() {
     setFilterByBranch,
     setFilterByStatus,
     setFilterByText,
+    testRunsLoading,
   } = useContext(TestRunsContext);
   const { filterByTime, setFilterByTime } = useContext(TestRunsFilterContext);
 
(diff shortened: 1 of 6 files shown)


src/ui/components/Library/Team/View/NewTestRuns/TestRunsPage.tsx
```diff
@@ -144,9 +145,13 @@ function TestRunsContent() {
...
               data-test-id="TestRunList"
             >
-              <Suspense fallback={<LibrarySpinner />}>
+              {testRunsLoading ? (
+                <div className="flex h-full items-center justify-center">
+                  <LibrarySpinner />
+                </div>
+              ) : (
                 <TestRunList />
-              </Suspense>
```
callingmedic911, line 149: Remove suspense because underlying data (testRuns) is loaded with useImperativeCacheValue in context root, that means no suspense. So I think no point of having that.

bvaughn (approved):
Saw this in your Loom
Screenshot 2023-12-08 at 5 18 06 PM

Doesn't seem related to this PR but any idea what it is?

callingmedic911:
`fetch` to some resource failed. This happens only in dev if I'm too quick to refresh while some resource/api was loading.

PR #9993, opened 2023-12-09
Show flaky rate and add option to sort by it

callingmedic911:
Closes SCS-1648 and SCS-1612.

This PR now shows flaky rate, and allows to sort on flaky rate & alphabetically. https://www.loom.com/share/189d13d23cf54a91be1f1a12ee176396

- I ran GraphQL codegen but I think I missed to do that in earlier PRs, so there are a few unrelated type changes.
- Documenting the design changes I made: Since there's now flaky rate, so I'm pill design from test runs Test Runs Overview panel (flaky and failure) but modified text color to be white for failure. Background red with white was hard to read. This change is reflected both in the original panel and now test list item. https://www.loom.com/share/6ea1979510c842e6b48f7d00e7e822a0 cc: @jonbell-lot23 
- If we have both flaky and failed rate, I'm using a specific order to show. I have commented that below (PR review).


src/ui/components/Library/Team/View/Tests/TestListItem.tsx
```diff
@@ -1,20 +1,44 @@
...
-function Status({ failureRate }: { failureRate: number }) {
-  const status = failureRate > 0 ? "fail" : "success";
-  const displayedFailureRate = Number((failureRate * 100).toFixed(0));
+function Status({ test }: { test: Test }) {
+  const { sortBy } = useContext(TestContext);
 
-  if (failureRate > 0) {
+  // We are showing rate in the pill depending on the sortBy.
+  // If sortBy is failureRate or alphabetical, we show rate in this order of priority: failureRate, flakyRate, success.
+  // Otherwise with sortBy flakyRate, we show rate in this order: flakyRate, failureRate, success.
```
callingmedic911, line 15: Here's how we decide which rate to show.

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
```diff
@@ -1,20 +1,44 @@
-import { useContext } from "react";
+import { useContext, useMemo } from "react";
 
 import { Test } from "shared/test-suites/TestRun";
 import HighlightedText from "ui/components/Library/Team/View/TestRuns/HighlightedText";
 import Icon from "ui/components/shared/Icon";
 
 import { TestContext } from "./TestContextRoot";
-import styles from "../../../Library.module.css";
+import styles from "../NewTestRuns/TestRuns.module.css";
```
callingmedic911, line 8: I'm re-using what I defined for test runs but I saw frontend team deciding on new guideline but I was not sure if this the right way. So just pointing this out.

src/ui/components/Library/Team/View/NewTestRuns/TestRuns.module.css
```diff
@@ -42,7 +42,7 @@
 }
 
 .failedPill {
-  color: var(--theme-base-100);
+  color: var(--primary-accent-foreground-text);
   background-color: var(--testsuites-failed-color);
```
bvaughn, line 46: This looks wrong to me. `--testsuites-failed-color` is meant to be a _foreground_ color, but this is using it as a _background_ color and stacking another arbitrary color on top of it.

Do you mean to be using `--testsuites-capsule-error-bgcolor` and `--testsuites-capsule-error-color` here instead?

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
```diff
@@ -1,20 +1,44 @@
...
+function Status({ test }: { test: Test }) {
+  const { sortBy } = useContext(TestContext);
 
-  if (failureRate > 0) {
+  // We are showing rate in the pill depending on the sortBy.
+  // If sortBy is failureRate or alphabetical, we show rate in this order of priority: failureRate, flakyRate, success.
+  // Otherwise with sortBy flakyRate, we show rate in this order: flakyRate, failureRate, success.
+
+  const { status, rate, classNames } = useMemo(() => {
+    console.log("sortBy", sortBy, test.failureRate, test.flakyRate);
```
bvaughn, line 18:
```suggestion
```

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
ryanjduffy: I think we should move the relevant styles into a specific `.module.css` file for this component.

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
ryanjduffy: It feels unexpected to me to be sorting by flaky rate but sometimes show the failure rate. I'd favor always showing the flaky rate when that's the sort order even if the rate is 0 and there were some failures. My assumption here is the use case for this view is to drive down flake so bringing focus to failures (but not flakes) would be distracting.

ryanjduffy (changes requested):
This is mostly in a good place but I had a couple comments below

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
callingmedic911: That style is used in multiple component. Should we replicate same style for each component or create shared.module.css?

bvaughn: @callingmedic911 Relevant comments @ 
* https://linear.app/replay/issue/FE-2120/add-top-level-testsuites-css#comment-1f7b6338
* https://linear.app/replay/issue/FE-2120/add-top-level-testsuites-css#comment-e334d01e

src/ui/components/Library/Team/View/NewTestRuns/TestRuns.module.css
callingmedic911: @jonbell-lot23 Can you please help here? I think you recently did a pass on improving the background color here. I'm not sure which color do we need to use.

For text color, I want to use white on top of testsuite's red. I'm also not sure which variable to use here.

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
callingmedic911: 🪵: cleaning up.

src/ui/components/Library/Team/View/Tests/TestListItem.tsx
callingmedic911: It'll now show either flaky or failure rate for all tests. Flaky when flaky rate is selected. Failure when either alphabetical or failure rate is selected. Does this seem okay to you?

callingmedic911: Feels bit odd when in flaky sort, we'll never show a green check. 🤔 But maybe that only makes sense in failure rate sort.

src/ui/components/Library/Team/View/NewTestRuns/TestRuns.module.css
bvaughn: > For text color, I want to use white on top of testsuite's red. I'm also not sure which variable to use here.

`--testsuites-capsule-error-bgcolor` and `--testsuites-error-icon-bgcolor` are both solid red, where as `--testsuites-error-color` is pinkish. Not sure why that is. I dunno what you should be using then, but I don't think you should be using what's currently used above.

packages/shared/graphql/generated/GetTestRunRecordings.ts
```diff
@@ -103,7 +103,9 @@ export interface GetTestRunRecordings_node_Workspace {
   testRuns: GetTestRunRecordings_node_Workspace_testRuns | null;
 }
 
-export type GetTestRunRecordings_node = GetTestRunRecordings_node_Recording | GetTestRunRecordings_node_Workspace;
+export type GetTestRunRecordings_node =
+  | GetTestRunRecordings_node_Recording
+  | GetTestRunRecordings_node_Workspace;
```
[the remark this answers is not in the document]
ryanjduffy: It's strange but I think it's right. The `node` member in our schema can be either workspaces or recordings so when it generates the types, it doesn't know at this level which you're using.

You can see the same thing in other places using `node` e.g:

```
export type GetWorkspaceApiKeys_node = GetWorkspaceApiKeys_node_Recording | GetWorkspaceApiKeys_node_Workspace;
```

ryanjduffy (approved):

src/ui/components/Library/Team/View/NewTestRuns/TestRuns.module.css
jonbell-lot23: Sorry @callingmedic911 , I just saw this. Looking now.

jonbell-lot23: 1. The reason --testsuites-error-color is pinkish in dark theme is because it's set on a dark maroon background, and red on near-black is hard to read: 
image

2. @callingmedic911 , you'll want to use --testsuites-capsule-error-color for the color, and --testsuites-capsule-error-bgcolor for the background. See how this class is called "pill?" The aligning naming is "capsule" in the CSS.

callingmedic911: I think similar changes would be required for flakyPill and successPill too - for background colors? Can I request a separate PR to fix the colors too, after we merge this one? I'm copying these classes over from TestRuns.module.css to TestListItemModule.css, I'm going to revert that back to these original values in this PR. To be clear, we'd need to update both of these module in follow up PR.

packages/shared/graphql/generated/GetTestRunRecordings.ts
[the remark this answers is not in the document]
bvaughn: But when would you ever want to treat a _recording_ and a _workspace_ as the same thing? They're not interchangeable.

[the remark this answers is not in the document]
ryanjduffy: I guess it's a deficiency in the type generator? I'd have expected it to be able to figure out that with this graphql:
```
  query GetTestRunRecordings($workspaceId: ID!, $id: String!) {
    node(id: $workspaceId) {
      ... on Workspace {
```
it will only be returning `Workspace` types but apparently it can't.

src/ui/components/Library/Team/View/NewTestRuns/TestRuns.module.css
jonbell-lot23: Filed a task here: https://linear.app/replay/issue/DES-1027/follow-up-color-pr

PR #9995, opened 2023-12-09
Code cleanup

bvaughn:
Split a couple of small things off of #9965 that can land early:

- [x] Fixed the `addVisualComment`
- [x] Updated the `saveBrowserExample` helper to fix a race case that caused it to sometimes missed "ExampleFinished" logs

### DIFF
--- a/packages/e2e-tests/scripts/save-examples.ts
+++ b/packages/e2e-tests/scripts/save-examples.ts
@@ -432,8 +432,10 @@ async function saveBrowserExample({ example }: TestRunCallbackArgs) {
 
   // Shouldn't be "node" by this point
   await recordPlaywright((argv.runtime || example.runtime) as BrowserName, async (page, expect) => {
-    await page.goto(exampleUrl);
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -432,8 +432,10 @@ async function saveBrowserExample({ example }: TestRunCallbackArgs) {
 
   // Shouldn't be "node" by this point
   await recordPlaywright((argv.runtime || example.runtime) as BrowserName, async (page, expect) => {
-    await page.goto(exampleUrl);
-    await playwrightScript(page, expect);
+    const waitForLogPromise = playwrightScript(page, expect);
```
bvaughn, line 435: We need to start listening _before_ loading the page. Otherwise we'll usually miss any logs that weren't put behind a `setTimeout` (and even some of them).

markerikson (approved):

hbenl (approved):

PR #10000, opened 2023-12-11
Debounce React component details when coming from the Node Picker

bvaughn:
- [x] Proposed alternative to #9987; only debounces React element details when the Node Picker is being used.
- [x] Add unit tests for `useDebounceState` hook

### DIFF
--- a/packages/replay-next/components/windowing/GenericListData.ts
+++ b/packages/replay-next/components/windowing/GenericListData.ts
@@ -78,8 +78,10 @@ export abstract class GenericListData<Item> extends EventEmitter<{
   };
 
   setSelectedIndex(value: number | null) {
-    this._selectedIndex = value;
-    this.emit("selectedIndex", value);
+    if (this._selectedIndex !== value) {
+      this._selectedIndex = value;
+      this.emit("selectedIndex", value);
+    }
   }
 
   subscribeToLoading = (callback: (value: boolean) => void) => {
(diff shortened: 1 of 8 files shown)


packages/replay-next/components/windowing/GenericListData.ts
```diff
@@ -78,8 +78,10 @@ export abstract class GenericListData<Item> extends EventEmitter<{
   };
 
   setSelectedIndex(value: number | null) {
-    this._selectedIndex = value;
-    this.emit("selectedIndex", value);
+    if (this._selectedIndex !== value) {
+      this._selectedIndex = value;
+      this.emit("selectedIndex", value);
+    }
```
bvaughn, line 84: We should land this change regardless. Without it we have a potential update loop between this and `useSyncExternalStore`.

src/ui/components/SecondaryToolbox/react-devtools/suspense/inspectedElementCache.ts
```diff
@@ -27,7 +27,7 @@ export const inspectedElementCache = createCache<
   InspectedReactElement
 >({
   config: { immutable: true },
-  debugLabel: "DOMSearchCache",
+  debugLabel: "inspectedElementCache",
```
bvaughn, line 30: We should land this regardless.

bvaughn:
Can't view the failing RDT test because the recording immediately crashes with:
> CommandError: The session was destroyed while the command was in progress

PR #10003, opened 2023-12-13
Configure code coverage reporting for Playwright E2E in CI

markerikson:
This PR:

- drops use of `@replayio/action-playwright`, in order to manually configure the reporters in `playwright.config.ts`
- Generates a JSON-format code coverage file for each Playwright shard, and uploads it as a GH Actions artifact
- adds a new "merge coverage" job that downloads those and collates them, and and generates an HTML format coverage report with all of the merged coverage data. That is also uploaded as a GH Actions artifact.
- Drops the dead "E2E test jail" section (no tests were included in that any more)

So, viewing the results currently requires opening the PR checks summary, downloading the merged coverage artifact zip ( `all-code-coverage-reports-merged` ), and extracting it.

Here's the example output from one of the last runs I did:

[all-code-coverage-reports-merged-18.zip](https://github.com/replayio/devtools/files/13730427/all-code-coverage-reports-merged-18.zip)

We are currently at about 83% code coverage from the E2E tests (regular and auth).  This does _not_ count any coverage we have from the Jest unit tests or the Delta screenshot tests:

image

### DIFF
--- a/packages/e2e-tests/authenticated/logpoints-01.test.ts
+++ b/packages/e2e-tests/authenticated/logpoints-01.test.ts
@@ -1,3 +1,5 @@
+import { addCoverageReport } from "monocart-reporter";
+
 import { openDevToolsTab, startTest } from "../helpers";
 import { E2E_USER_1_API_KEY, E2E_USER_2_API_KEY } from "../helpers/authentication";
 import { disableAllConsoleMessageTypes, verifyConsoleMessage } from "../helpers/console-panel";
@@ -10,19 +12,27 @@ import {
 import { openSource } from "../helpers/source-explorer-panel";
 import { addLogpoint, editLogPoint, removeAllLogpoints } from "../helpers/source-panel";
 import { waitForRecordingToFinishIndexing } from "../helpers/utils";
-import test, { Page, expect } from "../testFixtureCloneRecording";
+import test, { Page, base, expect } from "../testFixtureCloneRecording";
 
 const url = "authenticated_logpoints.html";
 const lineNumber = 14;
 
 async function load(page: Page, recordingId: string, apiKey: string) {
   await startTest(page, recordingId, apiKey);
+  await page.coverage.startJSCoverage();
 
   await openDevToolsTab(page);
   await openSource(page, url);
   await openPauseInformationPanel(page);
 
   await disableAllConsoleMessageTypes(page);
+}
+
+async function close(page: Page) {
+  const jsCoverage = await page.coverage.stopJSCoverage();
+
+  await addCoverageReport(jsCoverage, base.info());
+  await page.close();
 }
 
 test.use({ exampleKey: url });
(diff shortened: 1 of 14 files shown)


bvaughn:
Looks like there was an unusually high number of flaky tests for the [most recent run on this branch](https://app.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs/b291edf4-ea08-4b40-b7fd-afc43520cf3d?param=dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI%3D&param=runs&param=6b178d41-e41c-4b9d-adba-3eac1868fb3e). Do you think this is concerning?

Screenshot 2023-12-21 at 3 46 19 PM

packages/e2e-tests/authenticated/logpoints-01.test.ts
```diff
@@ -25,6 +28,13 @@ async function load(page: Page, recordingId: string, apiKey: string) {
   await disableAllConsoleMessageTypes(page);
 }
 
+async function close(page: Page) {
+  const jsCoverage = await page.coverage.stopJSCoverage();
+
+  await addCoverageReport(jsCoverage, base.info());
+  await page.close();
```
bvaughn, line 35: Too bad we have to remember to do this anywhere we use `context.newPage()` – should we abstract that too? (like we do with `startTest`?)

bvaughn (approved):

markerikson:
I think the most recent run was actually [this one](https://app.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs/0f482a47-e41e-4d83-af3e-9f1345adb438?param=dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI%3D&param=runs&param=b291edf4-ea08-4b40-b7fd-afc43520cf3d), which had _some_ flakiness for 3 auth tests, but not nearly as much.

The flakiness appears to be something related to console message counts - it's trying to find 10 messages with "updated", but one failed to evaluate correctly.  That doesn't seem related to this PR.  (Example: https://app.replay.io/recording/authenticatedlogpoints-01-shared-logpoints-functionality--68b8d7fe-9664-4bd0-87e8-7920d00a117a?e2e=&apiKey= )

packages/e2e-tests/authenticated/logpoints-01.test.ts
markerikson: We've only got two tests that try to do multi-page work atm, so I'm willing to let that be as-is for now.

bvaughn: IMO this is worth addressing even while there are only a few tests that use the functionality, because it seems likely we will overlook this in the future. It also doesn't seem like it would require much effort to address now.

PR #10004, opened 2023-12-13
Debounce React component details when coming from the Node Picker

hbenl:
This is a mixture of #9987 and #10000. Like the latter, it reliably detects where a selection change is coming from. But the logic in `ReactDevtoolsPanel` is simpler and doesn't require the new `useDebounceState` hook.


bvaughn (approved):
I dig it.

PR #10005, opened 2023-12-13
Handle failed logpoint evaluations

hbenl:


packages/replay-next/src/suspense/AnalysisCache.ts
```diff
@@ -18,6 +18,7 @@ export interface AnalysisParams {
 }
 
 export interface RemoteAnalysisResult {
+  failed?: boolean;
```
bvaughn, line 21: nit: I don't like optional attributes. We could mark this one non-optional if we assigned like `failed: result.failed === true`

packages/replay-next/src/suspense/LogPointAnalysisCache.ts
```diff
@@ -19,6 +19,7 @@ import { mappedExpressionCache } from "./MappedExpressionCache";
 
 export type LogPointAnalysisResult = {
   executionPoint: ExecutionPoint;
+  failed?: boolean;
```
bvaughn, line 22: nit: I don't like optional attributes. We could mark this one non-optional if we assigned like `failed: result.failed === true`

bvaughn (approved):

PR #10009, opened 2023-12-13
Use launchdarkly to target specific workspaces with the new test runs view

jaril:
This switches us to using launchdarkly so we can incrementally (and specifically) roll out the new runs view to teams.

https://www.loom.com/share/3553874012784d01b3ec93bfc3a07983

### DIFF
--- a/packages/shared/user-data/LocalStorage/config.ts
+++ b/packages/shared/user-data/LocalStorage/config.ts
@@ -10,6 +10,5 @@ export const config = {
   reactDevToolsStateExpanded: Boolean(true),
   replayNextCurrentPanel: "sources" as ReplayNextCurrentPanel,
   replayVideoPanelCollapsed: false,
-  enableTestSuitesNewRunsView: false,
   enableTestSuitesTestsView: false,
 };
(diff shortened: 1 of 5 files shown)


src/ui/components/Library/Team/utils.ts
```diff
@@ -47,3 +47,10 @@ export function pushRoute(router: NextRouter, relativeURL: string): void {
     search,
   });
 }
+
+export function useGetRealTeamId() {
+  const teamId = useGetTeamIdFromRoute();
+  const match = atob(teamId).match(/w:(.*)/);
+
+  return match ? match[1] : teamId;
```
jaril, line 55: This feels silly but necessary. @bvaughn are you aware of somewhere where we already do this, in case there's some better established pattern?

src/ui/components/Library/Team/utils.ts
```diff
@@ -47,3 +47,10 @@ export function pushRoute(router: NextRouter, relativeURL: string): void {
     search,
   });
 }
+
+export function useGetRealTeamId() {
```
ryanjduffy, line 51: This can be dropped now

ryanjduffy:
We don't really need this if we're putting it on another branch

jonbell-lot23:
@callingmedic911 is tracking this down now, but just flagging that we should document this somewhere

callingmedic911:
https://www.loom.com/share/beb19c53916d4bf39ef6f8e58e196b78

PR #10018, opened 2023-12-14
Adjusting CSS colors

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-1027/follow-up-color-pr

It's a one line change, not doing much. Just better naming.

(Update: added a few more lines)


callingmedic911 (changes requested):
I have copied over this class in TestListItem.module.css, so we probably want to do the same change there too.

jonbell-lot23:
> I have copied over this class in TestListItem.module.css, so we probably want to do the same change there too.

Thanks for the tip! But this is concerning -- there's no way I am going to remember to update CSS in all the places we're squirrelling repeated css classes.  So I'm going to make this change and think of a better way for other commits.

callingmedic911 (approved):

PR #10021, opened 2023-12-14
Add more tests for Playwright panel functionality

markerikson:
This PR:

- [x] Adds a Playwright version of `cypress-04` for testing test step list item interactions
- [x] Adds a Playwright version of `cypress-05` for testing DOM node interactions

### DIFF
--- a/packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
+++ b/packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
@@ -26,11 +26,6 @@ test("playwright-04: Test Step buttons and menu item", async ({
 
   await openPlaywrightTestPanel(page);
 
-  const testPanel = getTestSuitePanel(page);
-
-  const isVisible = await testPanel.isVisible();
-  expect(isVisible).toBe(true);
-
   // has 1 test
   const rows = getTestRows(page);
   await waitFor(async () => {
@@ -81,8 +76,8 @@ test("playwright-04: Test Step buttons and menu item", async ({
   const jumpToSourceMenuItem = page.getByText("Jump to test source");
   expect(await jumpToSourceMenuItem.getAttribute("data-disabled")).toBe("true");
 
-  // click it anyway to hide the context menu
-  await jumpToSourceMenuItem.click({ force: true });
+  // hide the context menu
+  await page.keyboard.press("Escape");
 
   await firstJ2CStep.click();
   const pausedTimelinePercent = await getTimelineCurrentPercent(page);
--- a/packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
+++ b/packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
@@ -1,0 +1,113 @@
+import { Locator } from "@playwright/test";
+
+import { isViewerTabActive, openViewerTab, startTest } from "../helpers";
+import { closeSource, getSelectedLineNumber, waitForSelectedSource } from "../helpers/source-panel";
+import {
+  getSelectedTestCase,
+  getTestCaseSteps,
+  getTestRows,
+  getTestStepBeforeAfterButtons,
+  getTestSuitePanel,
+  getUserActionEventDetails,
+  openPlaywrightTestPanel,
+} from "../helpers/testsuites";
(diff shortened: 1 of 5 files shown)


packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
```diff
@@ -0,0 +1,113 @@
...
+}) => {
+  await startTest(page, recordingId);
+  await openViewerTab(page);
+
+  await openPlaywrightTestPanel(page);
+
+  const testPanel = getTestSuitePanel(page);
+
+  const isVisible = await testPanel.isVisible();
+  expect(isVisible).toBe(true);
```
bvaughn, line 32: nit: `openPlaywrightTestPanel` helper already ensures the test suite panel is visible.

packages/e2e-tests/tests/playwright-05_hover-dom-previews.test.ts
```diff
@@ -0,0 +1,85 @@
...
+}) => {
+  await startTest(page, recordingId);
+  await openViewerTab(page);
+
+  await openPlaywrightTestPanel(page);
+
+  const testPanel = getTestSuitePanel(page);
+
+  const isVisible = await testPanel.isVisible();
+  expect(isVisible).toBe(true);
```
bvaughn, line 26: nit: `openPlaywrightTestPanel` helper already ensures the test suite panel is visible.

packages/e2e-tests/tests/playwright-04_menu-commands.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  // Verify that's disabled.
+  await firstJ2CStep.click();
+  const menuButton = getByTestName(firstJ2CStep, "TestSectionRowMenuButton");
+  await menuButton.click();
+
+  const jumpToSourceMenuItem = page.getByText("Jump to test source");
+  expect(await jumpToSourceMenuItem.getAttribute("data-disabled")).toBe("true");
+
+  // click it anyway to hide the context menu
+  await jumpToSourceMenuItem.click({ force: true });
```
bvaughn, line 85: `page.keyboard.press("Escape")` is a less hacky way of doing this?

(or we could copy the `hideContextMenu` helper from our screenshot tests into the e2e-tests folder and use it, if you think that's cleaner)

packages/e2e-tests/tests/playwright-05_hover-dom-previews.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  const userActionSteps = steps.filter({
+    hasText: /click/,
+  });
+
+  // Hovering over a user action step should show a preview of the DOM node
+  const lastClickStep = userActionSteps.last();
+  await lastClickStep.scrollIntoViewIfNeeded();
+
+  // Note that the ID is dynamically generated based on the piece of the box model shown
+  const highlighter = page.locator("#box-model-content");
```
bvaughn, line 57: What creates this `box-model-content` element? Searching our project for that string, I only see test selectors and a CSS selector.

packages/e2e-tests/tests/playwright-05_hover-dom-previews.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  await waitFor(
+    async () => {
+      // Repeatedly hover over the first step and then the click step, to force the
+      // `onMouseEnter` handler to keep checking if we have a DOM node entry available.
```
bvaughn, line 62: Is there a better way we could wait for this case? I think we typically use data attributes to signal readiness for cases like this (and then we read them, and `delay(...)` before reading again if necessary)

markerikson: I don't think so. We don't use the DOM node cache at all while rendering. It's only checked in the `onHover` logic, synchronously, to see if we have it available.

bvaughn: FWIW I think we could still do what I described here. The `useCacheStatus` hook (in "suspense") lets you peek at a cache status without kicking off a fetch.

PR #10025, opened 2023-12-14
Revert "Stop referring to Library.module.css in Testsuites"

jonbell-lot23:
This looks great in our new testsuites view, but some visual issues snuck into the older testsuites views. Reverting.


jonbell-lot23:
cc @callingmedic911 who I just worked on this with, and @ryanjduffy fyi

PR #10026, opened 2023-12-14
Fixes/docs for running e2e tests against a local backend and chromium.

klochek:
- Updated README
- Allow replay API key to be set via env-var
- new `test:debug_local` target to run with a custom chromium
- new `replay-chromium-custom` to support the above
- Added chromium log output via `RECORD_REPLAY_BROWSER_LOG` envar
- Avoid trying to upload recording twice
- Throw if the recording wasn't uploaded

### DIFF
--- a/packages/e2e-tests/playwright.config.ts
+++ b/packages/e2e-tests/playwright.config.ts
@@ -5,8 +5,20 @@ const { CI, SLOW_MO, SHARD_NUMBER } = process.env;
 
 const projects = [
   {
+    name: "replay-chromium-custom",
+    use: {
+      launchOptions: {
+        executablePath:
+          process.env["REPLAY_BROWSER_BINARY_PATH"] ||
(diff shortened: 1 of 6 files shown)


bvaughn (approved):

markerikson (approved):

packages/e2e-tests/playwright.config.ts
```diff
@@ -4,9 +4,21 @@ import { devices as replayDevices } from "@replayio/playwright";
 const { CI, SLOW_MO, SHARD_NUMBER } = process.env;
 
 const projects = [
+  {
+    name: "replay-chromium-custom",
+    use: {
+      launchOptions: {
+        executablePath:
+          process.env["REPLAY_BROWSER_BINARY_PATH"] ||
```
Domiii, line 12: Can we use `REPLAY_CHROMIUM_EXECUTABLE_PATH` instead, as we are using that already in other places?

PR #10027, opened 2023-12-14
Redirect all recording links from tests.replay.io to app.replay.io

jaril:


src/ui/utils/recording.ts
```diff
@@ -15,9 +16,23 @@ export function getRecordingURL(recording: Recording): string {
...
 
+  // Revert once we merge the test-suites branch into main (see SCS-1723)
+  if (window.location.host === "tests.replay.io") {
+    return `https://app.replay.io/recording/${id}`;
+  }
+
   return `/recording/${id}`;
 }
 
+export function useGetRecordingURLForTest(recordingId: string): string {
```
ryanjduffy, line 27: Will this mean that users will get "stuck" on app.replay.io once they open a recording? E.g. the in app back button will keep them on app.replay.io

jasonLaster: That's a good point. 

If this is a problem in the future we might need to use a NextJS middleware to redirect requests for the library when you're viewing a test workspace. Hopefully this is temporary though and we dont need it.

jaril: @ryanjduffy still ok to just merge so we stop breaking telemetry?

jasonLaster: also probably good to do since test-suites will not have the latest and greatest

ryanjduffy (changes requested):
I tried this locally and it does behave as I guessed above. If all we're trying to do with this is fix sending telemetry, let's fix that problem instead. We need to be merging `main` into this branch regularly anyway so we shouldn't fall behind in functionality or fixes.

PR #10031, opened 2023-12-15
Stop referring to Library.module.css in Testsuites (2nd attempt)

jonbell-lot23:
For a long time we've reused the styles in Library.module.css but the right thing to do is have a Testsuites.module.css file for things that are related to testsuites. (This is try 2 -- there were a few legacy places I needed to patch up)


ryanjduffy:
@jonbell-lot23 - I'm assuming you mean to target this at `test-suites` so I'm going to change it but let me know if this was supposed to update `main`.

ryanjduffy:
Looks like you'll need to rebase this to remove commits from `main`. I can help with that later if you need it.

ryanjduffy (approved):

PR #10038, opened 2023-12-18
Fix ElementsListData parent lookup bug

bvaughn:
Resolves FE-2145


jasonLaster (commented):
I like seeing the new test

bvaughn:
Me too. Designing stuff to be testable up front pays off later.

PR #10039, opened 2023-12-19
Consolidating font-size css variables

jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-1045/fix-our-font-sizes-in-variablescss

image

We have a few straggler css variables that should use more standardised naming, so this PR does that.

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -1,4 +1,4 @@
-/* Text sizes */
+/* Core font sizes */
 :root,
 :root.prefers-default-font-size {
   --font-size-large: 16px;
@@ -24,7 +24,7 @@
   --theme-focus-border-color-textbox: #0675d3;
   --theme-textbox-box-shadow: rgba(97, 181, 255, 0.75);
 
-  /* Other text sizes */
+  /* Other font sizes */
   --root-font-size: 16px;
   --theme-code-line-height: calc(15 / 11);
   --theme-tab-font-size: var(--font-size-large);
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -9,10 +9,10 @@
   --theme-textbox-box-shadow: rgba(97, 181, 255, 0.75);
 
   /* Text sizes */
-  --devtools-base-font-size: 12px;
+  —font-size-regular: 12px;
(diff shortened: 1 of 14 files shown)


packages/replay-next/pages/variables.css
```diff
@@ -9,10 +9,10 @@
   --theme-textbox-box-shadow: rgba(97, 181, 255, 0.75);
 
   /* Text sizes */
-  --devtools-base-font-size: 12px;
+  —font-size-regular: 12px;
   --root-font-size: 16px;
-  --theme-body-font-size: 12px;
-  --theme-code-font-size: 11px;
+  --font-size-regular: 12px;
+  --font-size-regular-monospace: 11px;
```
bvaughn, line 15: These values are already set in the `:root.prefers-default-font-size` and `:root.prefers-large-font-size` blocks:

https://github.com/replayio/devtools/blob/b78b180d8141efc3837975151b086a06fc54a7aa/packages/replay-next/pages/variables.css#L127-L142

I don't think we need to (or should) re-define them here.

```suggestion
```

src/devtools/client/debugger/src/components/Editor/EditorPane.tsx
```diff
@@ -27,9 +27,9 @@ export const EditorPane = () => {
   useLayoutEffect(() => {
     const root = document.querySelector<HTMLElement>(":root")!;
     if (enableLargeText) {
-      root.style.setProperty("--theme-code-font-size", "14px");
+      root.style.setProperty("--font-size-regular-monospace", "14px");
     } else {
-      root.style.setProperty("--theme-code-font-size", "11px");
+      root.style.setProperty("--font-size-regular-monospace", "11px");
```
bvaughn, line 32: I imagine we can delete this entire effect, since these values are already defined at the root.

jonbell-lot23: Adjusted

packages/replay-next/pages/variables.css
jonbell-lot23: Sounds good - font size is so foundational for a CSS layout that it feels strange having it down on line 127, so I pulled it to the top of the document and removed the dupes.

bvaughn (approved):

PR #10044, opened 2023-12-19
Display test run statistics

jaril:


src/ui/components/Library/Team/View/NewTestRuns/Chart.tsx
```diff
@@ -0,0 +1,97 @@
...
+    {
+      id: "build-failures",
+      color: "hsl(21, 70%, 50%)",
+      data: sortedData,
+    },
+  ];
+}
+export const Chart = () => {
+  const { testRuns } = useContext(TestRunsContext);
+  const data = generateChartData(testRuns);
```
jasonLaster, line 49: this should probably be wrapped in a `useMemo`

jasonLaster (approved):
Left one comment. Feel free to merge into test-suites

jasonLaster:
Actually here's some quick feedback

Screenshot 2023-12-19 at 11 55 15 AM

1. Let's try and reduce vertical height by removing the title. Having fewer minor lines (0.1, 0.2, 0.4, ...).
2. Let's use white for the font color for the graph labels. 
3. Lets round the statistics below. No need for that precision

jaril:
> 1. Let's try and reduce vertical height by removing the title. Having fewer minor lines (0.1, 0.2, 0.4, ...).

Are you referring to the "Build failures trend (percentage of builds failed)"? If so I'd prefer if we kept it—the table makes no sense otherwise.

As for the lines, unfortunately there's no way with the current nivo/line API to affect the y-axis tickmark labels for the chart.

> 2. Let's use white for the font color for the graph labels.

Fixed.

> 3. Lets round the statistics below. No need for that precision

Fixed.

> 4. Let's simplify the date. Instead of `2023-12-1`, we can simply do `12/1` which is easier to read.

Fixed.

> 5. Let's make the line red like the stats.

Fixed.

PR #10045, opened 2023-12-19
Aligning our secondary-accent color

jonbell-lot23:
**Background**

From my [Design System Alignment Project](https://www.notion.so/replayio/Design-System-Alignment-Project-87afabecf50d4badb2522b65922fb608?pvs=4) doc: Our variables.css file uses a lot of different colors that are very similar, but not quite. This is a broad project to help us pull our variables.css document into alignment with our Figma and make a system that is less likely to have small visual regressions and inconsistencies.

**Summary of these changes**

image

We have a bunch of hard-coded values in our system that should be referring to a single variable. This can lead to inconsistencies and misalignments. I made changes in these locations: 

1. testsuites-failed-color
2. secondary-accent-stroke
3. background-color-secondary-button
4. color-current
5. point-panel-timeline-label-selected-color
6. testsuites-v2-description
7. nag-background-color
8. testsuites-error-color

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -35,9 +35,9 @@
   --primary-accent: #01acfd;
   --primary-accent-hover: #0194ff;
 
-  --secondary-accent: #f02d5e;
+  --secondary-accent: orange;
   --secondary-accent-hover: #d72451;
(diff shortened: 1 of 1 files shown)


bvaughn (approved):
I like this

PR #10049, opened 2023-12-20
Fix CSS typo

jonbell-lot23:
Dumb typo. I had written:

```
--secondary-accent: var(--secondary-accent);
```

Instead of 

```
--secondary-accent: #d72451;
```

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -48,7 +48,7 @@
   --primary-accent: #01acfd;
   --primary-accent-hover: #0194ff;
 
-  --secondary-accent: var(--secondary-accent);
+  --secondary-accent: #d72451;
   --secondary-accent-hover: #d72451;
   --secondary-accent-stroke: var(--secondary-accent);
 


jonbell-lot23:
It's late on the east coast, this is a regression, and it's a single line revert, so I am going to merge without review once tests pass.

packages/replay-next/pages/variables.css
```diff
@@ -48,7 +48,7 @@
   --primary-accent: #01acfd;
   --primary-accent-hover: #0194ff;
 
-  --secondary-accent: var(--secondary-accent);
+  --secondary-accent: #d72451;
```
bvaughn, line 51: Hmm, this PR didn't get reviewed, and it seems to have set secondary accent and secondary accent hover to the same color.

bvaughn, line 51: Ignore my initial, stupid comment.

Anyway, after this PR both `--secondary-accent` and `--secondary-accent-hover` are the same color, which seems wrong.

jonbell-lot23: Will investigate

PR #10050, opened 2023-12-20
Standardise on "color-default" rather than "body-color"

jonbell-lot23:
**Background**

bvaughn added the concept of a three-colour ramp a while back:

color-default
color-dim
color-dimmer

(as well as a ton of other colors in this namespace, like color-current, color-contrast, etc)

Before that, we had a different naming convention that pivoted on color vs bgcolor:

body-color
body-bgcolor

But of course this method doesn't easily incorporate variations like dimming. So in this example, I'm pointing the legacy body-color stuff to the newer colour-default stuff.

In the future, we can find/replace all the old legacy stuff, but for now a re-route will work well.

### DIFF
--- a/packages/replay-next/pages/variables.css
+++ b/packages/replay-next/pages/variables.css
@@ -780,7 +780,7 @@
   --color-brand-react: #8c65d0;
   --color-contrast: var(--theme-base-100);
   --color-current: var(--secondary-accent);
-  --color-default: #38383d;
+  --color-default: #223344;
   --color-dim: rgba(135, 135, 137, 0.9);
   --color-dimmer: #aaaaaa;
   --color-disabled-button: var(--theme-base-100);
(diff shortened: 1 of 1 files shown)


packages/replay-next/pages/variables.css
```diff
@@ -48,7 +48,7 @@
   --primary-accent: #01acfd;
   --primary-accent-hover: #0194ff;
 
-  --secondary-accent: var(--secondary-accent);
```
bvaughn, line 51: How was this supposed to have worked before 😕

bvaughn (approved):

packages/replay-next/pages/variables.css
jonbell-lot23: Yeah, I fat fingered this in a previous PR and did an emergency one-liner PR when I saw it.

PR #10051, opened 2023-12-20
Handle Cypress elements data nested in a .props field, and highlight multiple nodes per test step

markerikson:
This PR:

- Updates the Cypress step details logic to handle looking for DOM nodes in cases where they are nested inside a `stepDetails.props` field, instead of being at the top level of the `stepDetails` object ( FE-2139 )
- Updates the step details cache to store all found DOM nodes for a step and updates the step row to highlight all of them, instead of just highlighting the first ( FE-2137 )
- Updates the user action row to show count badges for both Cypress and Playwright
- Updates the highlight tests to check for the multiple-DOM-node case

Working for Cypress:

image

and Playwright:

image

### DIFF
--- a/src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
@@ -14,6 +14,10 @@ import { createFocusIntervalCacheForExecutionPoints } from "replay-next/src/susp
 import { objectCache } from "replay-next/src/suspense/ObjectPreviews";
 import { cachePauseData, setPointAndTimeForPauseId } from "replay-next/src/suspense/PauseCache";
 import { Source, sourcesByIdCache } from "replay-next/src/suspense/SourcesCache";
+import {
+  findProtocolObjectProperty,
+  findProtocolObjectPropertyValue,
+} from "replay-next/src/utils/protocol";
 import { compareExecutionPoints, isExecutionPointsWithinRange } from "replay-next/src/utils/time";
 import { ReplayClientInterface } from "shared/client/types";
 import {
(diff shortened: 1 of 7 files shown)


jasonLaster (commented):
Nice! It's great to see the tests as well.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -321,39 +312,57 @@ async function fetchAndCachePossibleCypressDomNode(
...
+      })
+    );
+
+    const uniqueDomNodeIds = new Set<string>();
+
+    // Only want to show elements that are actually in the page,
+    // and remove any duplicates
+    elements = elements.filter(e => {
+      if (e.node.isConnected && !uniqueDomNodeIds.has(e.id)) {
+        uniqueDomNodeIds.add(e.id);
```
bvaughn, line 355: I'm curious about the duplicate element check.

src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
```diff
@@ -55,8 +56,9 @@ export default memo(function UserActionEventRow({
   userActionEvent: UserActionEvent;
 }) {
   const { data } = userActionEvent;
-  const { command, error, parentId, timeStampedPoints, resultVariable } = data;
-  const { result: resultTimeStampedPoint } = timeStampedPoints;
+  const testRunnerName = groupedTestCases.environment.testRunner.name;
+  const { command, error, parentId } = data;
+  const resultPoint = getUserActionTestEventResultPoint(userActionEvent, testRunnerName);
```
bvaughn, line 61: Is this something we could smooth over in the `RecordingTestMetada` component? so we didn't have to make lower level views aware of the implementation detail?

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: Some of the steps, like `eq` have both `"Applied To"` and `"Yielded"`, but the same node appears in each, like: `{"Applied To": ["a", "b", "c", "d"], "Yielded": "a"}`. That led to duplicate component key warnings when the highlighting logic tried to highlight `["a", "b", "c", "d", "a"]`.

markerikson: Hmm. Thinking about that further... maybe it makes more sense to pick _one_ of those.  "Yielded" would presumably be first, because that's what the step settled on.

bvaughn: I think that makes sense.

src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
markerikson: Possibly. It's a mixture of two different checks: does it have a possible execution point that we want to use, _and_ is it a relevant test step.  I could add another field to the test event that saves that value, but not sure if it's worth doing that.

markerikson:
Did several pieces of restructuring to the logic:

- Only fill out `result` point if the test step has the prerequisites for fetching
- Fill out `result` field at event object creation
- Flatten event details caching logic conditional structure
- Prefer "Yielded" for Cypress steps over "Applied To"

That also eliminated the need for that `getUserActionTestEventResultPoint` util.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
bvaughn: Did your thinking here change any or still considering it?

bvaughn (approved):
Not sure what the status is of the open comment thread but approving the rest to unblock.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: Yeah, the last push changed to just pick one of "Yielded" or "Applied To", preferring "Yielded" first.

PR #10052, opened 2023-12-20
Remove and consolidate one-off css colors

jonbell-lot23:
**Sketch I used to track this**
one-offs

**Summary**

There are some strange colors we're only using once. This PR cleaned up a bunch of them and figured out the justification for other ones.

**Removed, usually because they're not used anywhere**

1. cool-gray-500
2. theme-toolbar-highlighted-color (also green-50 and green-60)
3. grey-25
4. testsuites-aside-background-color, testsuites-aside-color, testsuites-aside-border-color
5. theme-highlight-yellow

**Kept**

1. purple-50 – this is our standard purple for the playhead. Maybe we need a better name for this. I don't want purple-50 used willy-nilly other than on the playhead
2. green-70 — I want to see if I can find a more unified green for the whole system
3. A bunch of grays (we need a broader gray audit to do this right)
4. magenta-65 is actually used in a ton of places
5. yellow-65 -- stay tuned

### DIFF
--- a/src/ui/components/shared/ReplayLogo.tsx
+++ b/src/ui/components/shared/ReplayLogo.tsx
@@ -8,8 +8,8 @@ const logoSizes = {
 };
 const logoColors = {
   white: "#FFF",
-  fuschia: "#F02D5E",
-  gray: "#6b7280)",
+  fuschia: "var(--secondary-accent)",
+  gray: "var(--grey-50)",
 };
 
 const Logo = () => (
--- a/src/ui/components/shared/ReplayLogo.tsx
+++ b/src/ui/components/shared/ReplayLogo.tsx
@@ -9,7 +9,7 @@ const logoSizes = {
 const logoColors = {
   white: "#FFF",
   fuschia: "#F02D5E",
-  gray: "var(--cool-gray-500)",
+  gray: "#6b7280)",
 };
 
 const Logo = () => (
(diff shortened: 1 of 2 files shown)


src/ui/components/shared/ReplayLogo.tsx
```diff
@@ -9,7 +9,7 @@ const logoSizes = {
 const logoColors = {
   white: "#FFF",
   fuschia: "#F02D5E",
-  gray: "var(--cool-gray-500)",
+  gray: "#6b7280)",
```
bvaughn, line 12: Revert this specific one.

bvaughn (approved):
approve all except cool-gray-500

src/ui/components/shared/ReplayLogo.tsx
jonbell-lot23: There's no benefit to it being a variable.

* It doesn't change from dark/light theme
* We don't use it anywhere else

If it were a variable named logo, I could see us using it in the future and reserving it. But it's an orphan variable we're trying to remove.

jonbell-lot23: Never mind. I don't want to come up with a short term fake variable yet but I'll do it anyway.

jonbell-lot23:
Fixed the cool-gray-500 thing and threw in an extra variable for good measure

PR #10060, opened 2023-12-21
Fix toggling of video panel

markerikson:
This PR:

- Updates the logic in `<Video>` to more consistently sync the "is video expanded" state with the resizeable panel
- Adds test logic for the video panel collapsing

### DIFF
--- a/packages/e2e-tests/tests/resizable-panels-01.test.ts
+++ b/packages/e2e-tests/tests/resizable-panels-01.test.ts
@@ -1,6 +1,7 @@
-import { startTest } from "../helpers";
-import { delay, waitFor } from "../helpers/utils";
-import test, { Page } from "../testFixtureCloneRecording";
+import { openDevToolsTab, startTest } from "../helpers";
+import { showUserOptionsDropdown } from "../helpers/layout";
+import { debugPrint, delay, waitFor } from "../helpers/utils";
+import test, { Page, expect } from "../testFixtureCloneRecording";
 
 async function waitForPanelSize(page: Page, expectedSize: number) {
   await waitFor(async () => {
@@ -23,7 +24,7 @@ async function waitForPanelSize(page: Page, expectedSize: number) {
 
 test.use({ exampleKey: "doc_rr_basic.html" });
 
-test("resizable-panels-01: Left side Toolbar should be collapsible", async ({
+test("resizable-panels-01: Left side Toolbar and Video should be collapsible", async ({
   pageWithMeta: { page, recordingId },
   exampleKey,
 }) => {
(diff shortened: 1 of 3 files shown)


src/ui/components/Viewer.tsx
```diff
@@ -31,12 +31,10 @@ const Vertical = ({ toolboxLayout }: { toolboxLayout: ToolboxLayout }) => {
...
-      if (videoPanel.isCollapsed() !== videoPanelCollapsed) {
-        if (videoPanelCollapsed) {
-          videoPanel.collapse();
-        } else {
-          videoPanel.expand();
-        }
+      if (videoPanelCollapsed) {
+        videoPanel.collapse();
+      } else {
+        videoPanel.expand();
```
bvaughn, line 37: Something looks odd about this change. Was the panel reporting an incorrect value for `isCollapsed` before?

src/ui/components/Viewer.tsx
```diff
@@ -31,12 +31,10 @@ const Vertical = ({ toolboxLayout }: { toolboxLayout: ToolboxLayout }) => {
   useLayoutEffect(() => {
     const videoPanel = videoPanelRef.current;
     if (videoPanel) {
-      if (videoPanel.isCollapsed() !== videoPanelCollapsed) {
```
bvaughn, line 34: This change seems fishy to me. I think it _should_ work, and the fact that it didn't suggests a bug in react-resizable-panels. Stepping through the code, it looks like the problem is that I don't initialize `collapsedSize` to 0 when it's undefined, so it fails in the library here:

```js
  const isPanelCollapsed = useCallback(panelData => {
    const {
      layout,
      panelDataArray
    } = eagerValuesRef.current;
    const {
      collapsedSize,
      collapsible,
      panelSize
    } = panelDataHelper(panelDataArray, panelData, layout);
    return collapsible === true && panelSize === collapsedSize;
  }, []);
```

I guess this change is probably also okay to make though either way. I'll file a follow up issue.

bvaughn (approved):

src/ui/components/Viewer.tsx
markerikson: Yeah, I didn't go all the way into the lib code, but I wondered if it might be something like that.

Given that this is a "sync external behavior" situation, always calling the relevant `collapse/expand` method seems like reasonable implementation behavior.

packages/e2e-tests/tests/resizable-panels-01.test.ts
```diff
@@ -57,4 +58,39 @@ test("resizable-panels-01: Left side Toolbar should be collapsible", async ({
   // Collapse panel via the side toggle
   await button.click();
   await waitForPanelSize(page, 0);
+
+  // Video panel should be toggleable
```
bvaughn, line 62: Nice 👍🏼

src/ui/components/Viewer.tsx
bvaughn: Fixed upstream and updated our dependency in #10063

PR #10066, opened 2023-12-26
Node picker explicit error state

bvaughn:
Screenshot 2023-12-26 at 2 28 31 PM

Screenshot 2023-12-26 at 2 35 02 PM

Note that all e2e tests passed for this PR except `logpoints-03` (which also fails on main); see [test run](https://tests.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs/3257ec32-0444-4c5a-9c1e-8d6c4747bddf).

Note this PR does not update the legacy React DevTools node picker UI because that component has been deprecated and will be removed soon (once we drop support for Gecko recordings).

### DIFF
--- a/src/ui/components/App.css
+++ b/src/ui/components/App.css
@@ -79,6 +79,9 @@ h2 {
 #command-button-pick.active::before {
   background-color: var(--primary-accent);
 }
+#command-button-pick.errored::before {
+  background-color: var(--theme-icon-error-color);
+}
 
 .app-mask {
   height: 100%;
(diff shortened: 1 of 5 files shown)


src/ui/components/App.css
```diff
@@ -79,6 +79,9 @@ h2 {
 #command-button-pick.active::before {
   background-color: var(--primary-accent);
 }
+#command-button-pick.errored::before {
+  background-color: var(--theme-icon-error-color);
```
bvaughn, line 83: This color looks pink when it should be red. I don't like that, but I think this CSS var is _probably_ the one we should be using.

I think "errors" should always be red and the pinkish red color should be used for _current time_

cc @jonbell-lot23

hbenl (approved):

PR #10068, opened 2023-12-28
Refactor error boundary state and components

bvaughn:
This commit introduces three distinct types of error:

1. **Minor errors** only impact a small part of the app.
   * Show an inline message for these.
   * Looks like the pre-existing inline error boundary fallback UI.
2. **Fatal, expected** errors.
   * Show an actionable message for these (e.g. sign-in, refresh)
   * Looks like the pre-existing error modal dialog.
3. **Fatal, unexpected** errors.
   * Let the user know we've been notified.
   * Prompt them for more repro information if they'd like to share.
   * Looks like the pre-existing "Support" contact form.

It also refactors our error components into a few reusable pieces:

* `replay-next/components/errors/InlineErrorBoundary`
  * This type of boundary should be used to handle **minor errors**.
  * Wrap this around parts of the app that should degrade gracefully in the event of an error.
  * It's in replay-next because a lot of the newer components need to reference it.
* `src/ui/component/Errors/RootErrorBoundary`
  * This type of boundary should be used to handle **fatal errors** (both expected and unexpected).
  * Wrap this around the application root (in `_app`).
  * It's in src because it references a lot of Redux actions (which can't be imported from replay-next)

> ℹ️  **Note** the new `RootErrorBoundary` component is not in the replay-next package because it imports a lot of Redux actions and reducers that are in the legacy project folders. I've added inline comments explaining why the code is structured that way. If we wanted to relax the import restrictions on the `replay-next` module, we could collocate everything, but that would be a larger change.

Example of an unexpected error handled by the new `RootErrorBoundary`:
Screenshot 2023-12-29 at 4 42 50 PM

Example of an expected error handled by the new `RootErrorBoundary`:
Screenshot 2023-12-29 at 5 11 30 PM

(The InlineErrorBoundary UI will remain the same despite a name change.)


bvaughn:
The only new test failure (passport-02) passes locally so I think it's just a fluke.

packages/replay-next/components/errors/UnexpectedErrorForm.module.css
```diff
@@ -0,0 +1,6 @@
+.Details {
+  font-style: italic;
+}
+
+.Boilerplate {
```
markerikson, line 5: :hand: dead style?

markerikson (approved):

packages/replay-next/components/errors/UnexpectedErrorForm.module.css
bvaughn: Not really. The view references this style, there just happens to be no new rules in it. TBH I dunno how we prefer to handle this case. I guess I _could_ just return an empty `<div>` if that's "clearer" (seemed subjective)

bvaughn:
I would have preferred this branch be rebased onto `main` to pull in c9b2aca rather than it being cherry picked, but no big deal.

PR #10070, opened 2023-12-28
Group tests by test id instead of spec

ryanjduffy:
With https://github.com/replayio/backend/pull/9240 in place, the UI needs to be reworked to display the tests by `testId` instead of spec name. This PR includes the following changes:

* Change the test run context to set and pass test id instead of spec. This works with the current data because the backend is generating a spec-based `testId`
* Change `useFileNameTree` to add the file name as a path node and the test as the leaf node. This means that for "prod", we'll show the spec name as the branch and then again as the leaf. This will immediately be replaced by showing the spec as the branch and the individual test names as the leaf when the backend change is deployed
* Removed the frontend-side grouping of flaky tests (duplicate of #10115 targeting `main`)

Before:
image

After:
image

After backend change:
image


src/ui/components/Library/Team/View/NewTestRuns/TestRunsContextRoot.tsx
```diff
@@ -53,7 +53,7 @@ export function TestRunsContextRoot({ children }: { children: ReactNode }) {
   const filterByTextDeferred = useDeferredValue(filterByText);
   const [filterTestsByText, setFilterTestsByText] = useState("");
 
-  const [spec, setSpec] = useState<string | null>(null);
+  const [testId, setTestId] = useState<string | null>(null);
```
jaril, line 56: Ah nice — this is going to make it less janky to store the selected test in the URL instead of serializing the path!

jaril (approved):

PR #10071, opened 2023-12-29
Filter based on test title in test overview panel

callingmedic911:
Closes SCS-1752.

Before: 
Screenshot 2023-12-29 at 10 07 37 PM
After:
Screenshot 2023-12-29 at 10 07 54 PM


src/ui/components/Library/Team/View/NewTestRuns/Overview/RunResults.tsx
```diff
@@ -78,7 +78,7 @@ function TestStatusGroup({
 }) {
   const [expanded, setExpanded] = useState(true);
 
-  const tree = useFileNameTree(testGroup, filterByText);
+  const tree = useFileNameTree(testGroup, filterByText, true);
```
ryanjduffy, line 81: I don't think we need to make this an option. This is the only usage (right?) so it can just be the behavior that we search both the test titles and the file name.

ryanjduffy (approved):
One suggestion below. There's also a trunk error but that'll go away if you take my suggestion. :)

src/ui/components/Library/Team/View/NewTestRuns/Overview/RunResults.tsx
callingmedic911: In old view, do we still want to search just the file name? Because that's what we highlight there.

ryanjduffy: The old view isn't relevant in this branch. It will be replaced when to land this code in main

ryanjduffy (approved):

PR #10073, opened 2023-12-29
Improve RDT test wait logic and view source lookup

markerikson:
This PR:

- Tweaks a couple bits of the RDT-02 test logic to hopefully fix flakes

Latest test run passes, so hopefully that helps.

### DIFF
--- a/src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
+++ b/src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
@@ -147,6 +147,7 @@ export function SelectedElement({
             className={styles.IconButton}
             onClick={viewComponentSource}
             title="Jump to definition"
+            data-test-name="ReactDevTools-JumpToComponentSource"
           >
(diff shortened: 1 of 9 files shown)


src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
```diff
@@ -127,7 +127,7 @@ export function SelectedElement({
 
   return (
     <div className={styles.Panel} data-is-pending={isPending || undefined}>
-      <div className={styles.TopRow}>
+      <div className={styles.TopRow} data-test-id="ReactDevTools-TopRow">
```
bvaughn, line 130: Seems like we don't need this test id.

```suggestion
```

bvaughn (approved):

src/ui/components/SecondaryToolbox/react-devtools/components/SelectedElement.tsx
markerikson: Yeah, I started to add this so we could narrow the locator selection, then decided to just put one right on the button.

PR #10076, opened 2024-01-02
Add pending state for step details and extract Cypress details wrapper

markerikson:
This PR:

- Moves the `useTransition` for "current selected test event" into `TestSuiteContext`
  - Wraps the setter in a `startTransition`
  - adds the pending flag to the context
- Updates `TestSectionRow` and `TestStepDetails` to use that pending flag from context
- Extracts a `CypressUserActionStepDetails` component, to better match the wrapping structure for the Playwright version
- Uses the pending flag to dim opacity for the step details panel
- Fills in a seemingly missing `children` prop type for `<InlineErrorBoundary>` (TS was claiming it couldn't accept children)

image

PR #10080, opened 2024-01-03
Test Suites: Step details loading state

bvaughn:
Supersedes #10076

- [x] Restore #10074 (c78ff0e)
but with the new tabs/elements UI removed (0964878)
- [x] Rebase changes from #10076 (e327e79)


hbenl (approved):

bvaughn:
Had two e2e test failures. Neither seem related to this change and both passed locally, so I believe it's just more flakiness.

Looking at one of the tests, it seems we could be more robust by waiting until the passport side nav button became enabled. I'll add a follow up commit for that.

PR #10081, opened 2024-01-03
Add telemetry for loading the Elements panel(s)

bvaughn:
Log how long it takes us to fetch DOM elements for the new Elements panel to Honeycomb.

For the new, evaluations-based panel, this is done by instrumenting the `DOMCache`. For the legacy (Gecko) Elements panel, this is done by instrumenting the async `loadPathToNode` method. Although that method is called often, I've tried to make this as much of an apples-to-apples comparison as I can by only logging telemetry for the initial load. (This is kind of an imprecise approximation though.)

### DIFF
--- a/packages/replay-next/components/elements/ElementsListData.ts
+++ b/packages/replay-next/components/elements/ElementsListData.ts
@@ -9,6 +9,7 @@ import { isNodeInSubTree } from "replay-next/components/elements/utils/isNodeInS
 import { loadNodeSubTree } from "replay-next/components/elements/utils/loadNodeSubTree";
 import { shouldDisplayNode } from "replay-next/components/elements/utils/shouldDisplayNode";
 import { GenericListData } from "replay-next/components/windowing/GenericListData";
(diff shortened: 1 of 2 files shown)


packages/replay-next/components/elements/ElementsListData.ts
```diff
@@ -33,9 +35,12 @@ export class ElementsListData extends GenericListData<Item> {
     this._pauseId = pauseId;
     this._replayClient = replayClient;
 
-    this._rootObjectIdWaiter = {} as any;
+    this._rootObjectIdWaiter = {
+      resolved: false,
+    } as any;
     this._rootObjectIdWaiter.promise = new Promise(resolve => {
       this._rootObjectIdWaiter.resolve = resolve;
+      this._rootObjectIdWaiter.resolved = true;
```
hbenl, line 43: I guess the intention was to wait for the promise to be resolved before setting `resolved` to `true` but this is setting it immediately when the promise is created.

bvaughn: Geeze, you're right. Thanks for catching.

PR #10090, opened 2024-01-04
Test runs failure chart polish

jaril:
https://www.loom.com/share/a9b95329618c452da4cd0f2608485175


src/ui/components/Library/Team/View/NewTestRuns/TestRunsStats.tsx
```diff
@@ -11,36 +10,16 @@ export function TestRunsStats() {
...
-              <div className={styles.failureRate}>{(testFailureRate * 100).toFixed(0)}%</div>
-              <div>
-                ({testFailuresCount}/{testsCount})
-              </div>
-            </div>
-          </div>
+      <div className="flex flex-col overflow-auto rounded-lg bg-chrome p-4">
+        <Chart />
+        <div>
+          Failure rate: {(buildFailureRate * 100).toFixed(2)}% ({buildFailuresCount}/{buildsCount})
```
ryanjduffy, line 22: Maybe tuck the counts behind a tooltip? I'm not sure they're important to see at a glance

src/ui/components/Library/Team/View/NewTestRuns/Chart.tsx
```diff
@@ -1,7 +1,10 @@
+import { Point, PointTooltipProps } from "@nivo/line";
 import groupBy from "lodash/groupBy";
 import dynamic from "next/dynamic";
-import { useContext, useMemo } from "react";
+import { useContext, useMemo, useState } from "react";
+import { createPortal } from "react-dom";
```
ryanjduffy, line 5: Unused import

ryanjduffy (approved):
* Let's either make the chart fully responsive or set the pane to have a min width so it can't truncate the chart

image

ryanjduffy:
P.S. Maybe we need a slightly different version to show when there are only a few test runs. This would be nice when a new user is getting set up since all of their runs will be from the same day.

image

src/ui/components/Library/Team/View/NewTestRuns/TestRunsStats.tsx
jaril: fixed!

src/ui/components/Library/Team/View/NewTestRuns/Chart.tsx
jaril: fixed!

jaril:
> * Let's either make the chart fully responsive or set the pane to have a min width so it can't truncate the chart
> 
> image

The [panels](https://github.com/bvaughn/react-resizable-panels/tree/main/packages/react-resizable-panels) don't have a way of setting a min-width based on pixels (just percentage). 

> P.S. Maybe we need a slightly different version to show when there are only a few test runs. This would be nice when a new user is getting set up since all of their runs will be from the same day.
> 
> image

You're right this screen doesn't really add any value on day 1, since there's no prior data. 

The chart's purpose is so the user can quickly gauge whether their test runs are doing better/worse. If there were just 6 test runs there and they're all from the same day, then that doesn't really provide any information.

I think the ideal behavior here would be to keep the current view, but also have a way for the user to filter the runs down to the last day, that way we can make sense of the few runs that they do have. And for cases where it's literally their first day, we can prompt them to filter it down to the last 24 hours so that it doesn't show nonsense.

@ryanjduffy probably worth discussing in an issue instead of a PR comment though?

ryanjduffy:
> > P.S. Maybe we need a slightly different version to show when there are only a few test runs. This would be nice when a new user is getting set up since all of their runs will be from the same day.
> > image
> 
> You're right this screen doesn't really add any value on day 1, since there's no prior data.
> 
> The chart's purpose is so the user can quickly gauge whether their test runs are doing better/worse. If there were just 6 test runs there and they're all from the same day, then that doesn't really provide any information.
> 
> I think the ideal behavior here would be to keep the current view, but also have a way for the user to filter the runs down to the last day, that way we can make sense of the few runs that they do have. And for cases where it's literally their first day, we can prompt them to filter it down to the last 24 hours so that it doesn't show nonsense.
> 
> @ryanjduffy probably worth discussing in an issue instead of a PR comment though?

That's fine. I view this as a NUX opportunity more than a "bug" with this implementation.

jonbell-lot23:
I'm in bed so it's harder for me to file a linear ticket, but this is on my design radar. @ryanjduffy would you mind filing a quick ticket and I'll flesh it out?

jonbell-lot23:
Here are design tickets:

https://linear.app/replay/issue/DES-1078/polish-test-run-chart-for-new-teams
https://linear.app/replay/issue/DES-1079/improve-our-testsuites-graph

PR #10093, opened 2024-01-05
Ensure unique keys for test suite recording items

ryanjduffy:
## Issue

When the backend changes to group by test instead of spec file, it may return multiple tests entries for the same recording. Since the previous `key` value was the recording id, we had some odd rendering when switching to a run that had a spec with multiple tests because the `key` wasn't unique.

https://www.loom.com/share/9ef0adb12d1a4c319f82c0168d50d020

## Resolution

Ensure the key is unique

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -171,7 +171,7 @@ const FileNodeRenderer = memo(function FileNodeRenderer({
                 <TestResultListItem
                   depth={depth + 1}
                   filterByText={filterByText}
-                  key={recording.id}
+                  key={test.id + recording.id}
                   label={execution.result}
                   recording={recording}
                   test={test}


ryanjduffy:
@bvaughn - can I get a quick review on this one?

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -171,7 +171,7 @@ const FileNodeRenderer = memo(function FileNodeRenderer({
                 <TestResultListItem
                   depth={depth + 1}
                   filterByText={filterByText}
-                  key={recording.id}
+                  key={test.id + recording.id}
```
bvaughn, line 174: nit: This can still result in conflicts (though unlikely) e.g. test id 123 and recording id 45 vs test id 12 and recording id 345. Given the scheme our recording ids use, I guess this is not possible.

Is there any reason not to just use test id? Not sure I understand what's going on here fully. I guess sometimes there's multiple tests per recording and other times there's multiple recordings per test?

Will there ever be many-to-many?

PR #10094, opened 2024-01-05
Add queue for step commands to support stepping quickly

hbenl:
- add a queue for step commands
- move the action for stepping (formerly known as `executeCommandOperation`) to `src/ui/actions/timeline.ts` (next to the `seek()` action)
- reduce the amount of time that the stepping buttons are disabled (we don't need to wait for frames to be loaded)
- add an e2e test

### DIFF
--- a/src/devtools/client/debugger/src/reducers/pause.ts
+++ b/src/devtools/client/debugger/src/reducers/pause.ts
@@ -2,16 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at <http://mozilla.org/MPL/2.0/>. */
 
-import { AnyAction, PayloadAction, createAsyncThunk, createSlice } from "@reduxjs/toolkit";
+import { AnyAction, PayloadAction, createSlice } from "@reduxjs/toolkit";
 import type { FrameId, Location, PauseId, Value } from "@replayio/protocol";
 
-import { FindTargetCommand, resumeTargetCache } from "replay-next/src/suspense/ResumeTargetCache";
-import { isPointInRegion } from "shared/utils/time";
-import { seek } from "ui/actions/timeline";
-import { SourceDetails, getPreferredLocation, getSelectedSourceId } from "ui/reducers/sources";
+import { FindTargetCommand } from "replay-next/src/suspense/ResumeTargetCache";
+import { SourceDetails } from "ui/reducers/sources";
 import { getContextFromAction } from "ui/setup/redux/middleware/context";
 import type { UIState } from "ui/state";
-import { ThunkExtraArgs } from "ui/utils/thunk";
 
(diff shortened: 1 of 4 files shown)


src/devtools/client/debugger/src/reducers/pause.ts
```diff
@@ -299,6 +258,10 @@ export function getPauseHistoryIndex(state: UIState) {
   return state.pause.pauseHistoryIndex;
 }
 
+export function getNextQueuedCommand(state: UIState) {
+  return state.pause.queuedCommands[0];
+}
```
bvaughn, line 263: Could `dequeueCommand` just remove and return the next command?

Then we wouldn't have to have separate get-next and dequeue methods.

src/ui/actions/timeline.ts
```diff
@@ -229,6 +243,68 @@ function encodeFocusWindow(focusWindow: TimeStampedPointRange | null) {
...
+        location = resumeTarget.frame
+          ? getPreferredLocation(state.sources, resumeTarget.frame)
+          : undefined;
+        if (location) {
+          dispatch(setPreviewPausedLocation(location));
+        } else {
+          dispatch(previewLocationCleared());
+        }
+      } else {
+        dispatch(pauseCreationFailed());
```
bvaughn, line 292: What does it mean for us to return early here with commands still in the queue? Does it leave state in a corrupted state for the next step command?

src/ui/actions/timeline.ts
```diff
@@ -229,6 +243,68 @@ function encodeFocusWindow(focusWindow: TimeStampedPointRange | null) {
...
+        point = resumeTarget.point;
+        selectedFrameId = null;
+        location = resumeTarget.frame
+          ? getPreferredLocation(state.sources, resumeTarget.frame)
+          : undefined;
+        if (location) {
+          dispatch(setPreviewPausedLocation(location));
+        } else {
+          dispatch(previewLocationCleared());
+        }
```
bvaughn, line 290: Would it make sense to only do this for the last command?

I wonder if we're churning things unnecessarily for intermediate states (that we don't expect to be visible for long) in the case where the user has stepped multiple times.

bvaughn (commented):
Some thoughts.

src/devtools/client/debugger/src/reducers/pause.ts
hbenl: The redux action can't return anything, but I can create a thunk that does that.

src/ui/actions/timeline.ts
hbenl: I did this for every command because to me it feels better if I can see every step that the debugger made (even if each individual step is only visible for less than a second).

bvaughn (approved):

PR #10098, opened 2024-01-05
Add a message when there's no test data for a test run

jaril:
When there are no tests for a test run, this:
- hides the filter dropdown and the text filter in RunSummary (3rd column)
- changes the message in the test TestRunSpecDetails (4th column) from "Select a test to see its details here" (which is impossible since there's no test to select), to "No test data available for this test run".

image


ryanjduffy (changes requested):
I'd like to try to vertically align the message in the center with the message on the right. We also need a similar message in the tests view when there aren't any results. The messaging could be the same as we have now when there are no replays for a selected test in the runs view.

jaril:
> I'd like to try to vertically align the message in the center with the message on the right.

This one's difficult the left column has the headers of unknown height. There's a way to do it by positioning the error absolute and centering it within the parent container, but it's pretty janky and I'm not a fan. It's in this commit — lmk what you think.

> We also need a similar message in the tests view when there aren't any results. The messaging could be the same as we have now when there are no replays for a selected test in the runs view.

We already have something there, though I centered it so it looks similar to the other ones

image

ryanjduffy:
> > I'd like to try to vertically align the message in the center with the message on the right.
> 
> This one's difficult the left column has the headers of unknown height. There's a way to do it by positioning the error absolute and centering it within the parent container, but it's pretty janky, requires shifting paddings around and I'm not a fan. [It's in this commit](https://github.com/replayio/devtools/commit/8015107344ee39ec588f5f096c5b2aea989cdecb) — lmk what you think.
> 
> > We also need a similar message in the tests view when there aren't any results. The messaging could be the same as we have now when there are no replays for a selected test in the runs view.
> 
> We already have something there, though I centered it so it looks similar to the other ones
> 
> image

I think you could get there with by adding the following on the test suite message wrapping node.

```
    position: absolute;
    margin: -0.5rem;
```

It already accepts a `className` so you could configure the class in the test view CSS and pass it on to the component

ryanjduffy (approved):

PR #10103, opened 2024-01-08
Improve hover state for runs

jonbell-lot23:
**Summary**

Addresses [DES-1073](https://linear.app/replay/issue/DES-1073/hover-state-should-be-more-clear). Our hover state is current extremely subtle, to the point where you can't see it. This makes it more clear and more consistent with other parts of the system.

I also added some better padding to the hover. Last week we added a hover but it was cut too close.

Old:
image

New:
image

Old, light theme:
image

New, light theme:
image


ryanjduffy:
I don't think this was the PR you meant to open (`test-suites` targeting `main`) so i'm closing :)

jonbell-lot23:
Oh dear. Tried again here: https://github.com/replayio/devtools/pull/10112

PR #10106, opened 2024-01-08
Padding adjustment for testsuites

jonbell-lot23:
When the error starts, there should be a bit more space above it.

Old:
image

New:
image


jonbell-lot23:
Addresses https://linear.app/replay/issue/DES-1068/fix-padding-in-testsuites-devtool

ryanjduffy (approved):

PR #10109, opened 2024-01-08
Reset stackFrameIndex when selecting a different test step

hbenl:
This adds a `key` to force `<PlaywrightUserActionEventDetails>` to be remounted and have its state (i.e. `stackFrameIndex`) reset.
I could have used an effect to reset `stackFrameIndex` but I wanted to avoid the double rendering this would have caused.


markerikson (approved):

markerikson:
Just to check, what issue does this fix?

I had actually intentionally tried to _not_ add a key when I put up my "tabs for step details and source details" PR a couple weeks ago, to help retain the selected tab state (although that was admittedly coming from `localStorage` and probably not as much of a concern).

hbenl:
> Just to check, what issue does this fix?

FE-2164

PR #10114, opened 2024-01-08
Add warning message about deep linking to tests

ryanjduffy:
Fixes SCS-1751

The change to return tests (instead of specs) for cypress means that we'll have multiple links to the same replay in the current prod test suite dashboard. This message acknowledges to the user that they won't land on the specific test when they click it and will have to re-select from the cypress test panel. This will be fixable soon with a plugin change but hasn't been published and adopted yet.

image

### DIFF
--- a/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
+++ b/src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
@@ -17,10 +17,53 @@ import {
   useFileNameTree,
 } from "ui/components/Library/Team/View/TestRuns/Overview/useFileNameTree";
 import { TestRunsContext } from "ui/components/Library/Team/View/TestRuns/TestRunsContextRoot";
-import { TestGroup } from "ui/utils/testRuns";
+import { TestGroup, TestGroups } from "ui/utils/testRuns";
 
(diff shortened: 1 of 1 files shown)


jonbell-lot23 (approved):

hbenl (approved):

src/ui/components/Library/Team/View/TestRuns/Overview/RunResults.tsx
```diff
@@ -17,11 +17,54 @@ import {
   useFileNameTree,
 } from "ui/components/Library/Team/View/TestRuns/Overview/useFileNameTree";
 import { TestRunsContext } from "ui/components/Library/Team/View/TestRuns/TestRunsContextRoot";
-import { TestGroup } from "ui/utils/testRuns";
+import { TestGroup, TestGroups } from "ui/utils/testRuns";
 
 import { TestResultListItem } from "./TestResultListItem";
 import styles from "../../../../Library.module.css";
 
+function hasDuplicateRecordings(fileNameToTests: TestGroup["fileNameToTests"]) {
```
bvaughn, line 25: When we add temporary code, it's nice to have a Linear issue in place (as well as a "TODO <Linear-issue>" comment) to make sure we don't forget to clean it out.

PR #10118, opened 2024-01-09
Extend focus window to contain the entire range of the selected test step

hbenl:
Fixes FE-2141.
We had already fixed a similar issue (FE-1756) but that fix only ensured that the `beforeStep` point is in the focus window. For FE-2141 we need to ensure that the `viewSource` point is also in the focus window.

### DIFF
--- a/packages/shared/utils/time.ts
+++ b/packages/shared/utils/time.ts
@@ -1,7 +1,32 @@
-import { ExecutionPoint, PointRange, TimeStampedPointRange } from "@replayio/protocol";
+import {
+  ExecutionPoint,
+  PointRange,
+  TimeStampedPoint,
+  TimeStampedPointRange,
+} from "@replayio/protocol";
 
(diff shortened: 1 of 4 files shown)


packages/shared/utils/time.ts
```diff
@@ -1,8 +1,33 @@
...
+} from "@replayio/protocol";
 
 const supportsPerformanceNow =
   typeof performance !== "undefined" && typeof performance.now === "function";
 
+export function minTimeStampedPoint(points: (TimeStampedPoint | null)[]): TimeStampedPoint | null {
+  let min: TimeStampedPoint | null = null;
+  for (const point of points) {
+    if (point) {
+      min = min === null ? point : BigInt(point.point) < BigInt(min.point) ? point : min;
```
markerikson, line 15: :question: Can we reuse the existing execution point comparison utils?

markerikson (approved):

PR #10122, opened 2024-01-09
fix: Recording page back button

callingmedic911:
Closes SCS-1775.

Depends on https://github.com/replayio/backend/pull/9278.


src/ui/components/Header/Header.tsx
```diff
@@ -178,25 +180,31 @@ export default function Header() {
...
 
+  const goBack = (e: React.MouseEvent<HTMLAnchorElement, MouseEvent>) => {
+    e.preventDefault();
+
+    if (window.history.length > 1) {
+      router.back();
+    } else {
+      router.push(fallbackUrl);
+    }
+  };
```
ryanjduffy, line 201: I was hoping we could do this but I don't think we can if we can't inspect the history (which, afaict next doesn't provide since it relies on browser state). If we added this and the user direct linked to a replay, clicking the in-app back button would take them out of replay instead of returning to the library.

ryanjduffy (changes requested):

src/ui/components/Header/Header.tsx
callingmedic911: to be able to navigate back (to tests or test run) we can either
1. this approach that takes them out of replay
2. have a referer query param, that takes full path (encoded URI). for example referer=/team/team-id/whatever, and we use this as is.
3. have just tiny query param when linking a recording, for example from=tests, find and return testId from backend for that recording, and form a url for it similarly we do test run.

I think 3 takes more time to implement and is just very rigid and we can have other places we want to link a recording from. 2 should be fine IMO. What do you think?

ryanjduffy: I think using the referrer would be okay. Another alternative would be to store the previous test suites path in context and use it here. I don't really like cramming more data elements into context though so the querystring param is probably better.

ryanjduffy:
The auth tests are failing because the logic to authorize `testRun` on the backend isn't working as expected. Opened https://github.com/replayio/backend/pull/9315 to address it

ryanjduffy (approved):
This should be fine to land once the backend PR lands and is deployed. It works correctly as is for users but our tests will be broken if this is deployed before the backend change.

PR #10128, opened 2024-01-10
Reimplement the "Free Trial Expired" modal

hbenl:
This PR reimplements the modal that was shown for recordings that were recorded after their workspace's trial expired, which had been removed in #10068.
Instead of treating this as a separate kind of error (in addition to `ExpectedError` and `UnexpectedError`), we now use an `ExpectedError` modal.
Furthermore this fixes the `workspaceId` used by the `<TeamBillingButton>` shown in that modal.


bvaughn (approved):
I dig it

PR #10135, opened 2024-01-10
Better wrapping in testsuites (3)

jonbell-lot23:
**Summary**

We're giving the numbers on the left their own column so it's easier to scan.

Old:
image

New:
image


jonbell-lot23:
Ok @ryanjduffy I think the TSC issues are gone and this should be ready.

ryanjduffy:
I don't think the "New" screenshot reflects the changes. Here's one:

image

ryanjduffy (approved):

PR #10136, opened 2024-01-10
Fixing padding in replay list

jonbell-lot23:
Fixes [SCS-1792](https://linear.app/replay/issue/SCS-1792/test-run-details-view-needs-some-padding)

**Summary**

This adds some padding to the hover effect, but that also means pushing other things over to make the lines line up when not hovered.

Examples:
image

image

### DIFF
--- a/src/ui/components/Library/Testsuites.module.css
+++ b/src/ui/components/Library/Testsuites.module.css
@@ -29,6 +29,7 @@
   gap: 0.5rem;
   overflow: hidden;
   border-radius: 0.375rem;
+  padding: 4px;
 }
 
 .linkContent {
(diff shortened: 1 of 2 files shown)


ryanjduffy (approved):

src/ui/components/Library/Testsuites.module.css
```diff
@@ -29,6 +29,7 @@
   gap: 0.5rem;
   overflow: hidden;
   border-radius: 0.375rem;
+  padding: 4px;
```
ryanjduffy, line 32: `px` instead of `rem`?

jonbell-lot23: Yup! We have a Notion page that discusses [when to use rem versus px](https://www.notion.so/replayio/When-to-use-rem-vs-px-f5fbb65a4cad4e0c936ee92581074bb8?pvs=4):

* rem if you want it to resize (think of it like an SVG, designed to scale)
* px if you don't (think of it like a PNG, not designed to scale)

In these three visuals, I am resizing the text but I want the padding to stay the same:
image

And that's why px is the right call! I could allow the capsule padding to get larger with the text, but it has more downsides than benefits.

PR #10138, opened 2024-01-11
Reorganize root error boundary to not drop session/user info

bvaughn:
I can think of a few ways to do this. I wonder what your thoughts are on this one, @hbenl?

Worth noting that if we run into an error during the initial render– it will prevent the session from being created, but even in that scenario we'll get user info after this change.


hbenl:
> I wonder what your thoughts are on this one, @hbenl?

I hadn't thought about it much yet but I think I would have ended up with something very similar.

hbenl (approved):

PR #10139, opened 2024-01-11
Re-record doc_navigate with Chromium

hbenl:
I found one more e2e example recording that can be switched to Chromium. One of the tests needed to be updated because it used a breakpoint location that is not available in the Chromium recording (I didn't see this as a parity issue because I don't think we expect the breakable locations to be exactly the same for different runtimes).


bvaughn (approved):
I think the difference in breakable locations is a regression, if you compare it to Chrome's behavior opening that same file, but I think we've already got a bug for that (BAC-4405) so I guess this is okay to move forward with.

PR #10143, opened 2024-01-11
Update replay cli to fix missing sourcemap upload issue.

klochek:
No idea why this is broken on the older cli, but have confirmed(?) this fix by running with the 0.20 CLI and rolling back to the old 0.12 cli multiple times.  This unlocks a number of tests, so I'm going to put "understanding" aside for know.


bvaughn (approved):
Wow. Nice find!

PR #10144, opened 2024-01-11
Delete old test runs

callingmedic911:
Closes SCS-1794.

- Removed unused files from old TestRuns
- Moved everything still being used to NewTestRuns
- Renamed NewTestRuns -> TestRuns
- Remove old E2E and related files.
- Rename tests path from new-test-suite to test-suite

Diff is really odd because we are reusing same file paths. So 🤷‍♂️


callingmedic911:
True exam of tests we added.

ryanjduffy (approved):

PR #10156, opened 2024-01-12
Consolidate syncing

jaril:
Here's a draft where we make the syncing function even more generic so it can handle both tests and test runs view.
- `testRunId` (runs view) and `testId` (tests view) => `viewId`
- `testId` (runs view) => `subId`


bvaughn:
This is an old PR with a lot of conflicts. I assume it's been abandoned, so I'm going to close it. Rebase and re-open if we want to reconsider.

PR #10158, opened 2024-01-13
Fix native browser navigation with test suite dashboard

ryanjduffy:
## Issue

Navigating back with the native browser buttons after clicking through a couple test runs or tests would break navigation.

## Resolution

Remove the react state <--> URL sync logic and rely just on the URL state

https://www.loom.com/share/d8aea50957cd4a8899d8b2cd1b5fa8d7


src/ui/components/Library/Team/View/TestRuns/TestRunsContextRoot.tsx
```diff
@@ -40,23 +39,21 @@ type TestRunsContextType = {
...
 
 export function TestRunsContextRoot({ children }: { children: ReactNode }) {
-  const { teamId, testOrTestRunId: defaultTestRunId } = useGetTeamRouteParams();
+  const { teamId, testOrTestRunId, testId } = useGetTeamRouteParams();
 
   const { testRuns, status } = useTestRuns();
 
-  const [testRunId, setTestRunId] = useState<string | null>(defaultTestRunId);
-
+  const [localTestRunId, setTestRunId] = useState<string | null>(testOrTestRunId);
```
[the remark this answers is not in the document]
bvaughn: > When I passed it the value from the URL, the URL and deferred value were always the same.

That's because `useDeferredValue` will be the same as whatever value you pass it initially (else it wouldn't work, from a typing perspective), and changes to URL values don't schedule updates with React (unless something is listening and telling React that it needs to update)– which could be done by mirroring the value in state, or you could use the sync external store hook to subscribe to URL changes.

That being said, it's not clear to me what you're trying to do pulling a value from a URL and why you would want/need that to be "deferred". Deferred values are values that get updated as part of a "transition"– which only matters for updates, not mounts. URL values usually only apply to mounts.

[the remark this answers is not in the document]
ryanjduffy: That makes sense, thanks! I haven't worked with suspense enough yet to have learned all of this so I appreciate the info.

I don't think we need the deferred values except that the loading states in the views rely on the difference between the displayed and actual values. That'll need to be handled differently.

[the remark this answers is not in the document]
bvaughn: Maybe you and I can chat about this in realtime sometime this week? Could be easier.

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -76,21 +86,20 @@ export function TestsContextRoot({ children }: { children: ReactNode }) {
...
-      testIdForDisplay: testId,
+      testIdForSuspense: deferredTestId,
+      testId: testId ?? null,
+      testPending: isTestPending,
       selectedTest: testId ? tests.find(t => t.testId === testId) ?? null : null,
       testsLoading: status === STATUS_PENDING,
       tests: filteredTests,
       testsCount: status === STATUS_PENDING ? 0 : tests.length,
     };
   }, [sortBy, filterByText, filterByTextDeferred, deferredTestId, testId, status, tests]);
```
bvaughn, line 102: This lint failure (missing dependencies) needs to be addressed.

src/ui/components/Library/Team/View/Tests/TestContextRoot.tsx
```diff
@@ -76,21 +86,20 @@ export function TestsContextRoot({ children }: { children: ReactNode }) {
       filterByText: filterByTextDeferred,
       filterByTextForDisplay: filterByText,
       selectTestId: (id: string | null) => {
+        router.push(`/team/${teamId}/tests/${id}`);
```
bvaughn, line 89: I think this change is good 👍🏼

src/ui/components/Library/Team/View/TestRuns/hooks/useTestRuns.ts
```diff
@@ -11,33 +11,24 @@ import { testRunsIntervalCache } from "../suspense/TestRunsCache";
 
 const EMPTY_ARRAY: any[] = [];
 
-export function useTestRuns(): { testRuns: TestRun[]; status: Status } {
+export function useTestRuns(): { testRuns: TestRun[] } {
```
bvaughn, line 14: A pattern we've been using in FE has been to name things that suspend explicitly. This makes it clearer that you need to use a Suspense boundary around them, or pass them deferred/transition priority values.

```suggestion
export function useTestRunsSuspends(): { testRuns: TestRun[] } {
```

bvaughn (approved):
Overall this looks like a good change. There is a lint rule that needs fixing, and a hook I think we should rename, but overall 👍🏼 

I'm not sure how I feel about the UX of sometimes maintaining selection in the middle list though:
https://github.com/replayio/devtools/assets/29597/35a87bda-0851-4cb2-9f3f-4cb7d7b1301a

What do you think about that? It's odd that we always reset filter text but we don't reset the selected item.

src/ui/components/Library/Team/View/TestRuns/hooks/useTestRuns.ts
ryanjduffy: I like this :+1:

ryanjduffy:
> What do you think about that? It's odd that we always reset filter text but we don't reset the selected item.
I think I agree. I hadn't noticed the incongruence with the filter and selected test. I've removed retaining the selected test.

bvaughn:
> I've removed retaining the selected test.

Cool 👍🏼 I dig that

PR #10159, opened 2024-01-13
Fix the back button

jaril:


bvaughn:
Seems like #10158 is this fix already?

PR #10163, opened 2024-01-15
[RUN-3006] Enable more object preview tests

Domiii:
* paired w/ https://github.com/replayio/chromium/pull/1079
* https://linear.app/replay/issue/RUN-3006/chromium-object-missing-properties#comment-e7e739a3


Domiii:
This was used for testing only. Closing.

PR #10164, opened 2024-01-16
Branch icon color consistency

jonbell-lot23:
We had special-cased this based on theme, but that's what CSS variables are good at.


src/ui/components/Library/Team/View/TestRuns/BranchIcon.tsx
```diff
@@ -13,11 +13,7 @@ export function BranchIcon({
 }) {
   const theme = useTheme();
```
ryanjduffy, line 14: I think we remove this line and probably the import of `useTheme` too

jonbell-lot23: Good call, fixed

PR #10168, opened 2024-01-16
Add disable cloning flag and process recording automatically

callingmedic911:
Closes SCS-1824.

Discussion around this change: https://discord.com/channels/779097926135054346/1196841641118543994

- Adds env flag to disable cloning for FE Runtime tests.
- Automatically loads recording after saving a new example. Missing two files right now (websocket.ts and protocolClient.ts)

Screenshot 2024-01-17 at 11 48 43 PM

### DIFF
--- a/packages/e2e-tests/scripts/loadRecording.ts
+++ b/packages/e2e-tests/scripts/loadRecording.ts
@@ -1,42 +1,37 @@
-import { loadedRegions as LoadedRegions } from "@replayio/protocol";
-
-import { defer } from "protocol/utils";
-
-import { newProtocolClient } from "./protocolClient/protocolClient";
+import { SimpleProtocolClient } from "@replayio/protocol";
+import WebSocket from "ws";
 
 const DISPATCH_URL =
   process.env.DISPATCH_ADDRESS ||
   process.env.NEXT_PUBLIC_DISPATCH_URL ||
   "wss://dispatch.replay.io";
 
-export const clientPromise = newProtocolClient(DISPATCH_URL);
+const callbacks: any = {
+  onClose: console.log,
+  onError: console.log,
+};
+
+const client = new SimpleProtocolClient(new WebSocket(DISPATCH_URL), callbacks, console.log);
 
 export const loadRecording = async (recordingId: string) => {
-  const client = await clientPromise;
   const { sessionId } = await client.sendCommand("Recording.createSession", {
     recordingId,
   });
+  console.log(`    ⏳ Processing recording ${recordingId} with session ${sessionId}`);
 
-  console.log(`    ⏳ Loading recording ${recordingId} with session ${sessionId}`);
+  try {
+    client.addEventListener(
+      // @ts-expect-error - when we update protocol client to 0.68, it will throw an error - just remove this comment
+      "Recording.processRecordingProgress",
+      ({ progressPercent }: { progressPercent: number }) => {
+        console.log(`    ⏳ Processing recording ${recordingId} ${progressPercent}%`);
+      }
+    );
 
-  const { promise: allLoadedPromise, resolve: allLoadedResolve } = defer<void>();
-  let loadedRegion: LoadedRegions["loaded"][0] | undefined;
-  client.addEventListener("Session.loadedRegions", event => {
-    if (
-      event.loaded.length !== 1 ||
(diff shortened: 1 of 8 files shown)


packages/e2e-tests/scripts/loadRecording.ts
```diff
@@ -0,0 +1,42 @@
...
+import { newProtocolClient } from "./protocolClient/protocolClient";
+
+const DISPATCH_URL =
+  process.env.DISPATCH_ADDRESS ||
+  process.env.NEXT_PUBLIC_DISPATCH_URL ||
+  "wss://dispatch.replay.io";
+
+export const clientPromise = newProtocolClient(DISPATCH_URL);
+
+export const loadRecording = async (recordingId: string) => {
```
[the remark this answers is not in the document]
callingmedic911: when script exits, it should close socket I think. Are you suggesting we should cleanly close it before exit?

packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -632,16 +633,15 @@ async function waitUntilMessage(
...
-        .map(recordingId => ` • ${blue(underline(`https://go/r/${recordingId}`))}`)
-        .join("\n")
+      `${newRecordingIds.length} new recordings have been saved. Loading each recording to ensure it has been pre-processed.`
     );
 
     const { exampleToTestMap } = getStats();
 
+    for (const recordingId of newRecordingIds) {
+      await loadRecording(recordingId);
+    }
```
bvaughn, line 643: We could parallelize this:
```js
await Promise.all(
  newRecordingIds.map(loadRecording);
)
```

If it didn't put too much load on the backend. @cweld510

callingmedic911: Yeah, I didn't do that intentionally. Maybe we can do in batches instead. 🤔

bvaughn: I'll defer.

bvaughn (approved):

callingmedic911:
https://buildkite.com/replay/runtime-fe-end-to-end-tests/builds/159#018d190c-1f75-420e-92b9-771a1820e638
This almost took ~1 minute per recording before it I guess it failed on Internal error. Sequential processing _may_ not work for this pipeline but it's not a fair assessment considering backend is swamped with these request from tests, so I'll hold off for now.

bvaughn:
Any objections if I pick this up and play with it a bit? There are merge conflicts now anyway.

packages/e2e-tests/scripts/save-examples.ts
jasonLaster: We use a simple queue in the test inbox server which you could re-use to control the parallelism. It's probably safe to start at 5 since these are such small replays.

https://github.com/replayio/test-inbox/blob/445925057671f00df20c087c490f9143566a7b2a/src/testInboxServer.ts#L19C7-L19C12

packages/e2e-tests/scripts/loadRecording.ts
```diff
@@ -0,0 +1,41 @@
...
+  const client = new SimpleProtocolClient(new WebSocket(DISPATCH_URL), callbacks, console.log);
+  client.addEventListener(
+    // @ts-expect-error - when we update protocol client to 0.68, it will throw an error - just remove this comment
+    "Recording.processRecordingProgress",
+    onProgress
+  );
+
+  await client.sendCommand("Recording.processRecording", {
+    recordingId,
+  });
```
bvaughn, line 38: I think we need to create isolated clients for each recording, else if we _were_ to try batching this, separate recordings would interfere with each other and the function would resolve prematurely.

bvaughn:
@callingmedic911 I'm not sure I understand the concern about the socket not closing correctly. We don't explicitly close sockets anywhere else (tabs get closed in the browser with an open socket too). But if you think there's a change we should make, I'll review a new PR?

callingmedic911:
I don't think it's _very_ important but I'm also waiting on this to land https://github.com/replayio/backend/pull/9351, so I'll do both together.

callingmedic911:
While writing the comment, it seems like that PR moved from client's experimental setting flag to launchdarkly, so may not need change at all.

PR #10172, opened 2024-01-17
Change email from "help@replay.io" to "support@replay.io"

jonbell-lot23:
We're not sure, but we think help@replay.io might not go anywhere.

### DIFF
--- a/packages/replay-next/components/support/SupportForm.tsx
+++ b/packages/replay-next/components/support/SupportForm.tsx
@@ -212,7 +212,7 @@ export function SupportForm({
           </ExternalLink>
           <ExternalLink
             className={styles.Footer}
-            href="mailto:help@replay.io"
+            href="mailto:support@replay.io"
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):

bvaughn:
> We're not sure, but we think [help@replay.io](mailto:help@replay.io) might not go anywhere.

Who is "we" and how did you arrive at that conclusion?

It is an alias that includes a few people (like me). I mentioned in chat that I was receiving these emails, often duplicates of them.

bvaughn:
Also the Form Carry form is still sending emails to help@replay.io

bvaughn:
I think we should revert this change, because it causes our support contacts to go to two different emails. (Tthe Form Carry form is still sending emails to help@replay.io)

bvaughn:
I am reverting this (#10173) for now until we've had a chance to talk about it and the broader strategy.

PR #10174, opened 2024-01-17
Process recording before creating a session

hbenl:
This shouldn't be merged before [BAC-4477](https://linear.app/replay/issue/BAC-4477/recordingprocessrecording-sometimes-takes-a-very-long-time) is addressed because I had the impression the processing slowness happens more often when using `Recording.processRecording`.


hbenl:
According to [this comment](https://linear.app/replay/issue/BAC-4477/recordingprocessrecording-sometimes-takes-a-very-long-time#comment-0c46e0b3),  using `Recording.processRecording` to trigger processing should not make a difference, so this PR can now be reviewed and merged.

src/ui/state/app.ts
```diff
@@ -93,6 +93,8 @@ export interface AppState {
   modal: ModalType | null;
   modalOptions: ModalOptionsType;
   mode: AppMode;
+  processing: boolean | null;
+  processingProgress: number | null;
```
bvaughn, line 97: I'm not sure why this needed to move from `ReplayClient` to Redux. Generally I would prefer us to be moving things in the other direction (from Redux into Suspense cache or `ReplyClient`).

bvaughn (commented):
I think the footprint of this change would have been smaller if we had stuck with `ReplayClient` and a custom hook that uses `useSyncExternalStore`. Is there a reason you chose to move this stuff into Redux instead?

src/ui/state/app.ts
hbenl: I moved it because we now monitor the processing progress *before* initializing `ReplayClient` (i.e. before we have a session). And I prefer state that changes (i.e. not caches) to be in Redux.

bvaughn: > I moved it because we now monitor the processing progress _before_ initializing `ReplayClient`

`ReplayClient` is instantiated and available (via context and Redux extra args) before the session exists, and it registers the processing event handler. (Doesn't matter when `.configure` is called with the session id. `processingProgress` can still be subscribed to at any time.)

> I prefer state that changes (i.e. not caches) to be in Redux.

Why?

I kind of prefer the opposite.

hbenl: So should I move this back to `ReplayClient`?

bvaughn: > `useAppSelector()` is much simpler than `useSyncExternalStore()`

That's fair, but this would be (was) in a custom hook so we didn't need to write it more than once.

> writing a redux slice feels more straightforward than doing whatever is necessary to make `useSyncExternalStore()` work. But perhaps it's just that I'm more used to redux.

Also fair.

> So should I move this back to `ReplayClient`?

I think it would be better to keep this in `ReplayClient` because it feels kind of related to `waitForSession` and `loadedRegions`. If you hadn't already moved it, I'd definitely say "don't move it" but since it's done I don't feel as strongly.

Do what you think is best here. :)

PR #10178, opened 2024-01-17
Chromium tests

jasonLaster:
This is a collection of changes for our e2e tests
* updates some recordings in examples.json
* removes logpoints-03_firefox since logpoints-03 passes
* updates the buildkite script to output the list of believed failing tests


packages/e2e-tests/scripts/buildkite_run_fe_tests.js
```diff
@@ -127,4 +136,20 @@ function run_fe_tests(CHROME_BINARY_PATH) {
   process.exit(0);
 }
 
+function listFailingTests() {
+  const testDirectory = path.join(__dirname, "../tests");
+  const testFiles = fs.readdirSync(testDirectory);
+  console.log(testFiles);
```
bvaughn, line 142: I don't think we need to log this too.
```suggestion
```

packages/e2e-tests/tests/playwright-01_basic-panel.test.ts
```diff
@@ -97,7 +98,7 @@ test("playwright-01: Basic Test Suites panel functionality", async ({
   const backButton = getTestRecordingBackButton(page);
   await backButton.click();
 
-  // Check if we're back on the main tests panel
+  debugPrint(page, "Check if we're back on the main tests panel");
   const secondRecordingTrees = getTestRecordingTrees(page);
   await waitFor(async () => {
```
bvaughn, line 103: nit: I guess you felt like it added value to include these new logs? Their formatting is kind of inconsistently and might cause some confusion when mixed in with our other logs. 

For example, it means nothing to see "_can open tests_" in the console.

If you want to add these, that's fine, but let's update them to be more semantically meaningful?

PR #10182, opened 2024-01-17
Improve e2e test scripts

bvaughn:
We had a lot of redundant, overlapping data between `examples.json` and the hardcoded `knownExamples` array in `save-examples`. I got tired of trying to update an example that was missing from `knownExamples`. There's no reason to have both.

PR #10184, opened 2024-01-17
Migrate doc_rr_basic.html to chromium

jasonLaster:
This PR moves the 11 or so tests to chromium and creates a new doc_rr_basic_firefox.html fixture for the 4 tests that need firefox.


bvaughn:
Lint fails: https://github.com/replayio/devtools/actions/runs/7565127272/job/20600367719?pr=10184

packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -181,6 +181,12 @@ const knownExamples: TestExampleFile[] = [
     category: "browser",
     runtime: "chromium",
   },
+  {
+    filename: "doc_rr_basic_chromium.html",
+    folder: config.browserExamplesPath,
+    category: "browser",
+    runtime: "chromium",
+  },
```
bvaughn, line 189: I'd like to land #10182 before this (which will obviate the need for this awkward duplication).

PR #10185, opened 2024-01-17
Set a 5 minute start time threshold

jasonLaster:
fixes FE-2181


bvaughn:
I'm not sure it's worth making this change. 5 minutes is a really long time to wait for indexing. (2 and a half minutes is already a long time.) Is there data backing this change?

jasonLaster:
The FE side. I believe the RUN side is hitting this, but lets get the data first.

bvaughn:
Check out this thread: https://discord.com/channels/779097926135054346/1197146238282383441/1197558458749308949

I think we're updating the wrong place here.

jasonLaster:
Yep. Closing this issue

PR #10187, opened 2024-01-18
Migrate console_warp-02.test.ts

jasonLaster:
Fixes FE-2186


packages/e2e-tests/tests/console_warp-02.test.ts
```diff
@@ -53,6 +53,7 @@ test("console_warp-02: support pausing, warping, stepping and evaluating console
   await stepOverToLine(page, 21);
   await verifyPausedAtMessage(page, "1 << 7", "terminal-expression");
 
+  await stepOverToLine(page, 21);
```
bvaughn, line 56: Chromium's stepping behavior is currently broken IMO (see RUN-2958, BAC-4405).

I don't think we should change test stepping behavior to match Chromium because (I think) that would be updating the tests to verify/approve broken behavior. This change does that.

bvaughn (changes requested):

PR #10188, opened 2024-01-18
Migrate doc_control_flow

jasonLaster:
Fixes FE-2187 and FE-2188

Migrates these four gecko tests to chromium
* tests/breakpoints-04.test.ts
* tests/repaint.test.ts
* tests/restart-session.test.ts
* tests/breakpoints-05.test.ts

### DIFF
--- a/packages/e2e-tests/tests/repaint.test.ts
+++ b/packages/e2e-tests/tests/repaint.test.ts
@@ -1,5 +1,10 @@
 import { openDevToolsTab, startTest } from "../helpers";
-import { resumeToLine, rewindToLine, stepOver, waitForPaused } from "../helpers/pause-information-panel";
+import {
+  resumeToLine,
+  rewindToLine,
+  stepOver,
+  waitForPaused,
+} from "../helpers/pause-information-panel";
 import { addBreakpoint } from "../helpers/source-panel";
 import { waitFor } from "../helpers/utils";
 import { Page, test } from "../testFixtureCloneRecording";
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/tests/repaint.test.ts
```diff
@@ -18,8 +23,8 @@ test("repaint: repaints the screen screen when stepping over code that modifies
 
   const prevDataUrl = await getCanvasDataUrl(page);
 
-  await stepOver(page);
-  await waitForPaused(page);
+  await addBreakpoint(page, { lineNumber: 55, url: exampleKey });
+  await resumeToLine(page, 55);
```
bvaughn, line 27: This seems suspicious for the same reasons as I mentioned in #10189 and #10187.


Chromium's stepping behavior is currently broken IMO (see RUN-2958, BAC-4405).

I don't think we should change test stepping behavior to match Chromium because (I think) that would be updating the tests to verify/approve broken behavior. This change does that.

bvaughn (changes requested):
Unless I'm wrong about my comment, I think we should not land this change.

packages/e2e-tests/tests/repaint.test.ts
jasonLaster: I'm okay changing the behavior here because this stepping issue is caught by the stepping tests and in general we should try to test one thing per test. Obviously this is impossible to do in an e2e test, but i think it's a reasonable goal.

bvaughn (approved):
> this stepping issue is caught by the stepping tests and in general we should try to test one thing per test

I'll withdraw my objection then since this is a repaint-focused test.

PR #10189, opened 2024-01-18
Migrate docs_rr_logs.html fixes FE-2189

jasonLaster:


bvaughn (changes requested):
Chromium's stepping behavior is currently broken IMO (see RUN-2958, BAC-4405).

We should not change test stepping behavior to match Chromium because (I think) that would be updating the tests to verify/approve broken behavior.

PR #10190, opened 2024-01-18
Migrate log_points_and_block_scope fixes FE-2191

jasonLaster:
Migrates 
* tests/logpoints-05.test.ts
* tests/logpoints-06.test.ts


bvaughn (approved):
Lots of flakes on CI but I guess that's not necessarily a red flag. Everything passes so 👍🏼

PR #10198, opened 2024-01-18
Fix breakpoints-05 test json

klochek:


ryanjduffy:
This PR failed typechecking and needs to be fixed or reverted.

hbenl:
> This PR failed typechecking and needs to be fixed or reverted.

Fixed in #10200

bvaughn:
This prob should not have landed without review.

PR #10200, opened 2024-01-19
Fix key in examples.json

hbenl:


bvaughn:
How was this working before?

bvaughn:
We have a test, `breakpoints-05`, that uses "doc_debugger_statements.html" as its key.

Then we have several tests that use a recorded run of `breakpoints-05` as their key (`playwright/breakpoints-05`).

We also have an old `manifest.js` entry (what is tihs even used for?) that maps:
```js
{
  example: "doc_debugger_statements.html",
  script: "breakpoints-05.js",
  targets: ["gecko", "chromium"],
},
```

PR #10201, opened 2024-01-19
doc_rr_objects: create Date object using the local timezone

hbenl:
The `Date` object was created using a timestamp, which is interpreted in the UTC timezone. But when it is logged, it is stringified for the local timezone and so the console message depended on the timezone where it was recorded.
With this change, the stringified date will still contain the local timezone (e.g. `Tue Oct 18 2022 00:00:00 GMT+0200 (Central European Summer Time)`) but everything before that (i.e. `Tue Oct 18 2022 00:00:00`) will be independent of the local timezone.
See FE-2195.


bvaughn (approved):
👍🏼

PR #10202, opened 2024-01-19
Partially revert "Migrate doc_control_flow (#10188)"

hbenl:
Bring back the Firefox recording of `doc_control_flow` and revert the changes to `repaint.test.ts`. Since the other tests using `doc_control_flow` work with the Chromium version, I've decided to bring the Firefox version back as `doc_control_flow_firefox`.

### DIFF
--- a/packages/e2e-tests/tests/repaint.test.ts
+++ b/packages/e2e-tests/tests/repaint.test.ts
@@ -1,15 +1,10 @@
 import { openDevToolsTab, startTest } from "../helpers";
-import {
-  resumeToLine,
-  rewindToLine,
-  stepOver,
-  waitForPaused,
-} from "../helpers/pause-information-panel";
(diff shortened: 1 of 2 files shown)


bvaughn (approved):

packages/e2e-tests/tests/repaint.test.ts
```diff
@@ -23,8 +18,8 @@ test("repaint: repaints the screen screen when stepping over code that modifies
 
   const prevDataUrl = await getCanvasDataUrl(page);
 
-  await addBreakpoint(page, { lineNumber: 55, url: exampleKey });
-  await resumeToLine(page, 55);
+  await stepOver(page);
```
bvaughn, line 21: Might be worth adding a comment about why this is critical to the test

PR #10203, opened 2024-01-19
Update file search expectations in screenshot tests

hbenl:
The file search result summary was fixed in #10195, these tests were still expecting the old (broken) behavior.


bvaughn (approved):
Ooof, nice catch.

PR #10209, opened 2024-01-22
Don't fail save-examples if a recording fails to process; RUN: add debug CLI logging for buildkite.

klochek:
Also add a fix if a given example key has no associated tests.

### DIFF
--- a/packages/e2e-tests/scripts/save-examples.ts
+++ b/packages/e2e-tests/scripts/save-examples.ts
@@ -386,7 +386,11 @@ async function waitUntilMessage(
     const { exampleToTestMap } = getStats();
 
     for (const recordingId of newRecordingIds) {
-      await loadRecording(recordingId);
+      try {
+        await loadRecording(recordingId);
+      } catch (e) {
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -386,16 +386,21 @@ async function waitUntilMessage(
...
     console.log("The following tests have been impacted by this change:");
     console.table(
       updatedExamples
         .map(example => {
           const tests = exampleToTestMap[example];
-
-          return ` • ${chalk.yellow(example)}${tests.map(test => `\n   • ${test}`).join("")}`;
+          return tests
+            ? ` • ${chalk.yellow(example)}${tests.map(test => `\n   • ${test}`).join("")}`
+            : "";
```
bvaughn, line 403: Seems like we should log an explicit message in this case?

klochek: Good idea; done.

bvaughn (approved):

PR #10210, opened 2024-01-22
Add client ID header for GraphQL requests

jcmorrow:
👋 Hello! It's been a long time since I have contributed here, so if I'm missing any important PR info please lemme know!

The Problem
---
When debugging GraphQL issues, especially E2E failures, it can get really confusing trying to figure out which backend GraphQL logs correspond to which client.

Proposed Solution
---
- Generate a UUID on the client when we create our GraphQL Client
- Attach that UUID to each GraphQL request as a header
- Teach the backend to extract that header and add it to our log context
- When debugging a test failure, inspect the recorded headers sent to GraphQL
- Use that to filter backend logs down to specific clients and better understand why a given test failed


packages/shared/graphql/apolloClient.tsx
```diff
@@ -115,7 +116,13 @@ export function createApolloCache() {
 }
 
 function createHttpLink(token: string | undefined) {
-  const headers: Record<string, string> = {};
+  // We generate a client ID so that we can easily filter backend requests down to a specific client
```
jcmorrow, line 119: I don't know if this kind of comment is helpful or considered bad style. I think we could safely leave it out if people are likely to `git blame` to understand code, since my commit does a better job of explaining it anyways.

bvaughn: TBH the thing I would want this comment to answer is "why isn't session id used?"

jcmorrow: That's fair! I have no idea what the async flow between apolloClient and session creation looks like, but if it's easy to wait for session creation here, I'd be totally fine with that.

bvaughn: Ah, so this wasn't a security decision or anything, just a convenience one?

I recorded two Replays to compare the sign-in flow (because this is not something I'm super familiar with either)
* Authenticated: https://go/r/baac2125-c979-462d-af0a-8162468c97a2
* Unauthenticated: https://go/r/cd2eb8e5-2d95-4131-91a6-79ea8b31ea1d

tl;dr we initialize GraphQL/Apollo _before_ we have a session id, so maybe this light weight approach you have here is the right trade off.

hbenl: > we initialize GraphQL/Apollo before we have a session id

And we wait for the first GraphQL results (for recording and user metadata) before we even try to create a session.

hbenl (changes requested):
Unfortunately we have more than one GraphQL client: there's [this one](https://github.com/replayio/devtools/blob/24c91549bd6149c96bdc5f28a6c84eb2d0195410/packages/shared/graphql/GraphQLClient.ts#L17-L44) and then there's [this](https://github.com/replayio/devtools/blob/24c91549bd6149c96bdc5f28a6c84eb2d0195410/packages/shared/graphql/User.ts#L38-L51) (which should be using the first one really, I'll file an issue for this).

bvaughn:
FWIW Josh's approach could still be made to work with these three clients. We'd just need to export the "client id" token. But yeah, consolidating our GraphQL clients is still good to do.

bvaughn:
Holger's concern should be addressed after a rebase?

packages/shared/graphql/apolloClient.tsx
jcmorrow: > Ah, so this wasn't a security decision or anything, just a convenience one?

I do think it has one relatively small security implication, which is that it would be possible for the user to change this value, and if on the graphql side we read this value as an authoritative answer as to the user's session we could potentially do something stupid like expose session details to an attacker. I think this is far-fetched and unlikely, but hey, those are usually the attack vectors that end up working! But it was mostly out of convenience.

> And we wait for the first GraphQL results (for recording and user metadata) before we even try to create a session.

Cool, so leaving it arbitrary is probably reasonable then 👍

bvaughn:
I think you forgot to `git add` the new shared file

jcmorrow:
> Holger's concern should be addressed after a rebase?

OK, so I have rebased and made sure that both of the GraphQL clients now use this shared client ID. I ended up pulling it out into a separate shared file. Let me know if that is idiomatic or if there is a better spot for me to put it in!

jcmorrow:
> I think you forgot to git add the new shared file

Huh... you are right. But I *definitely* ran `git add -N .` followed by my usual `git add --patch` 🤔

packages/shared/graphql/clientIdHeader.ts
```diff
@@ -0,0 +1,14 @@
+import { generateUUID } from "devtools/shared/generate-uuid";
+
+// Generate a client ID so that we can easily filter backend requests down to a specific client
+// browser tab. It might seem like we should just be attaching the session ID here, but devtools
+// makes requests before the session is established, so we use an arbitrary ID instead.
+export const GRAPHQL_CLIENT_ID = generateUUID();
+// This constant is also referenced in our GraphQL code.
```
bvaughn, line 7: Is this comment referring to code that runs on the backend?

bvaughn (approved):

packages/shared/graphql/clientIdHeader.ts
jcmorrow: oh yes, sorry let me make that clearer.

bvaughn:
Oops this change broke our imports:
https://github.com/replayio/devtools/actions/runs/7645192638/job/20831182845?pr=10210

jcmorrow:
> Oops this change broke our imports: [replayio/devtools/actions/runs/7645192638/job/20831182845?pr=10210](https://github.com/replayio/devtools/actions/runs/7645192638/job/20831182845?pr=10210)
> 
> Certain parts of FE code (GraphQL parts specifically) are very sensitive to import cycles.

Interesting. I also looked at the content of `generateUUID` as part of investigating this and... it does not generate a UUID 😛 https://github.com/replayio/devtools/blob/2866ec7a5b6fe56a7f40f5512dae065a85b50a36/src/devtools/shared/generate-uuid.js#L1-L7

Any objection to me `yarn add`ing `uuid` to the GraphQL package?

bvaughn:
RE: Where it should live

I think a utility like this could go in either `replay-next` or `shared`

> Any objection to me yarn adding uuid to the shared package?

That package is already a dependency. We use it in `useAddPoint` to generate unique keys for IndexedDB records.

jcmorrow:
> That package is already a dependency. We use it in useAddPoint to generate unique keys for IndexedDB records.

OK, cool! My editor was saying it could not find it, but I'm guessing that's just an issue with where my typescript server is looking, and won't actually break the build.

bvaughn:
I think this looks okay, provided nothing stands out with the e2e tests :)

jcmorrow:
> I think this looks okay, provided nothing stands out with the e2e tests :)

Great! It looks like I can't merge until @hbenl approves since he originally requested changes, so it'll probably have to wait till tomorrow anyways.

jcmorrow:
@hbenl bump on this - could you rereview when you get a chance? Thanks!

hbenl (approved):

packages/shared/graphql/clientIdHeader.ts
```diff
@@ -0,0 +1,14 @@
+import { v4 as uuid } from "uuid";
+
+// Generate a client ID so that we can easily filter backend requests down to a specific client
+// browser tab. It might seem like we should just be attaching the session ID here, but devtools
+// makes requests before the session is established, so we use an arbitrary ID instead.
+export const GRAPHQL_CLIENT_ID = uuid();
+// The header value of this header key will be logged on the backend during GraphQL requests.
+export const GRAPHQL_CLIENT_ID_HEADER_KEY = "Replay-Client-Id";
```
bvaughn, line 8: Why are we exporting these keys like this if they're only being reference in one place (below)

callingmedic911: I could use changes in this PR and taking over. Talked to Josh (here at offsite) and removed the unused exports.

bvaughn: Okedoke.

PR #10214, opened 2024-01-23
Update e2e tests script

bvaughn:
- [x] Add new `--build` flag to the `save-examples` script so that we can more easily target batches of recordings
- [x] Remove (broken) Playwright dynamic import; this didn't work and we won't be updating Gecko recordings anyway


bvaughn:
Loos like this uncovered some new regressions. RUN-3162, RUN-3164, and RUN-3165

PR #10215, opened 2024-01-23
Remove REPLAY_DISABLE_CLONE for tests; add pre-processing to the test startup.

klochek:
We must always be cloning tests, because multiple tests can reference the same recording, and will therefore race with any tests running in parallel, causing data races (and therefore errors) in the tests.  This change removes the preprocessing we were doing in save-examples, and moves it to the startup of the test, immediately following the clone.


bvaughn (changes requested):
> We must always be cloning tests, because multiple tests can reference the same recording

This was always the case, with the possible exception of RUN tests, which created new recordings just for the purpose of testing. Wasn't that why the disable-cloning flag was added in the first place?

> This change removes the preprocessing we were doing in save-examples, and moves it to the startup of the test

I don't think we should do this. I think this could potentially negatively impact the performance of all of FE e2e tests, if I understand the performance characteristics of BAC correct with regard to cloning.

Let's talk before moving forward.

klochek:
> > We must always be cloning tests, because multiple tests can reference the same recording
> 
> This was always the case, with the possible exception of RUN tests, which created new recordings just for the purpose of testing. Wasn't that why the disable-cloning flag was added in the first place?

But only one recording of, for example, `authenticated_comments.html`, is made when running save-examples.  That recording is used by 3 different tests (comments-XXX), which can all run in parallel, altering the same shared data (comments, obviously!)  So, we must always clone.

> > This change removes the preprocessing we were doing in save-examples, and moves it to the startup of the test
> 
> I don't think we should do this. I think this could potentially negatively impact the performance of all of FE e2e tests, if I understand the performance characteristics of BAC correct with regard to cloning.
> 
> Let's talk before moving forward.

Sure thing; my understanding is that the data generated by pre-processing does not get cloned (yet), and so pre-processing must be done on each recording.  It is also my understanding that pre-processing is a fast no-op if processing has already been completed.

bvaughn:
> > > We must always be cloning tests, because multiple tests can reference the same recording
> > 
> > This was always the case, with the possible exception of RUN tests, which created new recordings just for the purpose of testing. Wasn't that why the disable-cloning flag was added in the first place?
> 
> But only one recording of, for example, `authenticated_comments.html`, is made when running save-examples. That recording is used by 3 different tests (comments-XXX), which can all run in parallel, altering the same shared data (comments, obviously!) So, we must always clone.

Got it. That makes sense.

> > > This change removes the preprocessing we were doing in save-examples, and moves it to the startup of the test
> > 
> > I don't think we should do this. I think this could potentially negatively impact the performance of all of FE e2e tests, if I understand the performance characteristics of BAC correct with regard to cloning.
> > Let's talk before moving forward.
> 
> Sure thing; my understanding is that the data generated by pre-processing does not get cloned (yet), and so pre-processing must be done on each recording. It is also my understanding that pre-processing is a fast no-op if processing has already been completed.

My understanding is that data generated by pre-processing the original recording makes the cloned recordings faster to load, but pre-processing a cloned recording does not impact the original.

bvaughn:
> Hmm, I'll double-check with the backend folk about this; if this is the case, then I think the right thing would be to just keep the pre-processing in save-examples in

Right.

> and also include the pre-processing inside the test.

What's the motivation here (or what's the proposed change)? DevTools itself already waits on the pre-processing.

klochek:
> > and also include the pre-processing inside the test.
> 
> What's the motivation here (or what's the proposed change)? DevTools itself already waits on processing.

The idea here is that the waiting for processing is highly variable, and so I'd like to front-load the waiting in a place that's outside of playwright's usual "wait for a UI element to appear" apparatus.  It's true that we are waiting for a UI element to appear, but it's _actually_ the case that we're _really_ waiting for the backend to do processing for us, and so any timeout we see while waiting for that element isn't actually a UI slowness issue, or even a test-data correctness issue, but actually a backend problem, and putting in this explicit check separates-out that piece, making the source of the failure more obvious to the observer.

But maybe I'm over-thinking this?  I dunno--I like seeing "failure to process" in the playwright logs (and doing this via our API) over "waiting for this UI element took too long" (and doing this via playwright's waiting for a UI element.)

bvaughn:
I don't think you're over-thinking it, per say. I agree with what you said. It's just that– my expectation is that the wait time _should be_ very small if we've waited for (the newer implementation of) pre-processing on the golden recording before we clone.

That being said, I guess I don't feel strongly about this aspect of the change, so long as we don't remove the golden recording pre-processing step (in `save-examples`)

packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -23,19 +30,27 @@ const testWithCloneRecording = base.extend<TestIsolatedRecordingFixture>({
...
+      const client = new SimpleProtocolClient(new WebSocket(DISPATCH_URL), callbacks, console.log);
+
+      try {
+        console.log("       Processing recording");
+        await client.sendCommand("Recording.processRecording", {
+          recordingId: newRecordingId,
+        });
+      } catch (e) {
+        console.warn("       Error processing recording; ignoring.");
+      }
```
bvaughn, line 52: Can we re-use the code we already have for this (`loadRecording`)?

klochek: Yup--done.  (I was worried about the little console-animation waiter thingy messing w/ playwright, but it seems fine.)

PR #10221, opened 2024-01-24
Delete a bunch of unused cruft

bvaughn:
Is this safe to do? Was anything actually using this stuff?


test/jest/exportWebsocketMessages.js
markerikson, line 1: I think I'd added this in early 2022 to export some of our messages for testing console behavior before I refactored the messages reducer.

test/jest/jest-e2e-coverage-reporter/data.ts
markerikson, line 1: Also mine, from 2022, when I was experimenting with code coverage for our legacy E2E test setup. Dead.

test/manifest.js
markerikson, line 1: Most of these files in `/test` are also from the legacy test setup, I think

markerikson (approved):

PR #10226, opened 2024-01-25
Fix react_devtools-03 flakiness

hbenl:


packages/e2e-tests/helpers/console-panel.ts
```diff
@@ -478,6 +478,16 @@ export async function warpToMessage(page: Page, text: string, line?: number) {
   await seekToConsoleMessage(page, message, line);
 }
 
+export async function warpToLastMessage(page: Page, text: string, line?: number) {
```
bvaughn, line 481: Might be worth us renaming `warpToMessage` to `warpToFirstMessage` too?

PR #10227, opened 2024-01-25
Remove "basic processing progress" feature

bvaughn:
This feature was a hacky workaround for the problem of loading being "stuck" ([DES-82](https://linear.app/replay/issue/DES-82/sometimes-loading-appears-stuck-at-0percent)). We now have a much better solution in place for this– explicitly wait on the backend to complete processing before loading DevTools UI (FE-2170).

This feature was disabled by default as an _advanced_ setting, so deleting it should have minimal user impact.

Removing this code allows us to also remove the redundant `timeline.paints` and `timeline.points` values from Redux, which has the added benefit of unblocking some [planned refactoring work](https://www.notion.so/replayio/Untangling-legacy-graphics-code-d1d999faf7414f47ab9996ae92f717ff).

### DIFF
--- a/packages/protocol/graphics.ts
+++ b/packages/protocol/graphics.ts
@@ -115,10 +115,6 @@ const gMouseClickEvents: MouseEvent[] = [];
 let gDevicePixelRatio = 1;
 
 function onPaints(paints: PaintPoint[]) {
-  if (typeof onPointsReceived === "function") {
-    onPointsReceived(paints);
-  }
-
   paints.forEach(async ({ point, time, screenShots }) => {
     const paintHash = screenShots.find(desc => desc.mimeType == "image/jpeg")!.hash;
     insertEntrySorted(gPaintPoints, { point, time, paintHash });
@@ -126,10 +122,6 @@ function onPaints(paints: PaintPoint[]) {
 }
 
 function onMouseEvents(events: MouseEvent[]) {
-  if (typeof onPointsReceived === "function") {
-    onPointsReceived(events);
-  }
-
   events.forEach(entry => {
     insertEntrySorted(gMouseEvents, entry);
     if (entry.kind == "mousedown") {
@@ -178,7 +170,6 @@ export function setupGraphics(store: AppStore) {
     client.Graphics.findPaints({}, sessionId).then(async () => {
       initialPaintsReceivedWaiter.resolve(null);
       hasAllPaintPoints = true;
-      onAllPaintsReceived(true);
       maybePaintGraphics();
     });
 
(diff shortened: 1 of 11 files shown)


packages/protocol/graphics.ts
```diff
@@ -626,16 +617,6 @@ export function setPlaybackStatusCallback(callback: typeof onPlaybackStatus): vo
...
 
-let onPointsReceived: (points: TimeStampedPoint[]) => void;
-export function setPointsReceivedCallback(callback: typeof onPointsReceived): void {
-  onPointsReceived = callback;
-}
-
-let onAllPaintsReceived: (received: boolean) => void;
-export function setAllPaintsReceivedCallback(callback: typeof onAllPaintsReceived): void {
-  onAllPaintsReceived = callback;
-}
```
bvaughn, line 637: Removing these listeners is a first step toward removing this module (and it's state) entirely.

src/ui/actions/timeline.ts
```diff
@@ -126,8 +126,10 @@ export function jumpToInitialPausePoint(): UIThunkAction<Promise<void>> {
   return async (dispatch, getState, { replayClient }) => {
     const recordingId = getRecordingId(getState());
     assert(recordingId);
+
     const endpoint = await sessionEndPointCache.readAsync(replayClient);
-    dispatch(pointsReceived([endpoint]));
+    dispatch(setEndpoint(endpoint));
```
bvaughn, line 131: The only place we cared about these points was here, and we only cared about the endpoint. We can just store that in Redux instead.

src/ui/components/SessionContextAdapter.tsx
```diff
@@ -30,8 +30,7 @@ export default function SessionContextAdapter({
   const accessToken = useAppSelector(getAccessToken);
   const sessionId = useAppSelector(getSessionId);
   const duration = useAppSelector(getRecordingDuration)!;
-  const receivedPoints = useAppSelector(getPoints);
-  const endpoint = receivedPoints[receivedPoints.length - 1];
+  const endpoint = useAppSelector(getEndpoint);
```
bvaughn, line 33: The only place we cared about these points was here, and we only cared about the endpoint. We can just store that in Redux instead.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -252,30 +244,6 @@ export default async function setupDevtools(store: AppStore, replayClient: Repla
...
-
-  setAnalysisPointsReceivedCallback(newPoints => {
-    points.push(...newPoints);
-    onPointsReceived();
-  });
-
-  setAllPaintsReceivedCallback((received: boolean) => {
-    store.dispatch(allPaintsReceived(received));
-  });
-
```
bvaughn, line 278: Nice to remove this mutable, redundant state.

markerikson (approved):

PR #10231, opened 2024-01-26
Fix runtime value for save examples script

ryanjduffy:
This fixes sending correct `runtime` value to `getExecutablePath` but the Firefox browser can't launch because our fork only supports 1.19 and the script uses features from a newer version.

### DIFF
--- a/packages/e2e-tests/scripts/save-examples.ts
+++ b/packages/e2e-tests/scripts/save-examples.ts
@@ -153,21 +153,24 @@ async function saveExamples(
       continue;
     }
 
-    const [_, runtime] = buildId.split("-");
+    const [_, buildRuntime] = buildId.split("-");
 
     let category: TestExampleFile["category"];
     let folder: TestExampleFile["folder"];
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -153,21 +153,24 @@ async function saveExamples(
...
+    const [_, buildRuntime] = buildId.split("-");
 
     let category: TestExampleFile["category"];
     let folder: TestExampleFile["folder"];
+    let runtime: TestExampleFile["runtime"];
 
-    switch (runtime) {
+    switch (buildRuntime) {
       case "chromium":
       case "gecko": {
```
bvaughn, line 164: FWIW this was intentional, but we should probably throw instead if "gecko" is passed in as a target.

bvaughn:
Thanks Ryan. Probably not worth landing this change though, since it adds more support back for "gecko" when we're dropping it.

PR #10232, opened 2024-01-26
Fix wrong assumption in HooksRenderer

hbenl:


src/ui/components/SecondaryToolbox/react-devtools/components/HooksRenderer.tsx
```diff
@@ -33,7 +33,9 @@ export function HooksRenderer({
     return null;
   }
 
-  return object.preview.properties.map((property, index) => (
+  const objectProperties = object.preview.properties.filter(property => property.object);
```
Domiii, line 36: I feel like a utility function that filters out only Array elements from an `ObjectPreview` might be better here.

bvaughn: This change needs more description/context via an inline comment.

bvaughn: TBH I don't like this approach we're using (neither here nor the fix yesterday). 

I think we should instead be checking to see if the property is enumerable (via `.flags`)

bvaughn (changes requested):

hbenl:
> I think we should instead be checking to see if the property is enumerable (via .flags)

That would also not be correct, you can add properties to an array that are not elements (and they can be enumerable).

bvaughn:
> > I think we should instead be checking to see if the property is enumerable (via .flags)
> 
> That would also not be correct, you can add properties to an array that are not elements (and they can be enumerable).

That's true, but it feels more semantically correct since we are iterating over properties.

Technically you could push values that aren't the shape we're expecting to the hooks array too, in which case even the number check isn't right. The thing we're guarding against with this PR isn't _unexpected data_, but unexpected protocol properties. (I think "length" used to be in `getterValues` but now is in `properties`?)

bvaughn (approved):
I think using the enumerable attribute seems clearer but I don't feel _super strongly_ about this, so I'm going to remove my blocking mark.

hbenl:
> In this case, we control both ends (the thing creating/writing the array and the thing displaying it) so I don't think we need to guard against unexpected manually-defined properties.

True. But I've moved the check (`isArrayElement()`) into `utils/protocol.ts` now and since it's general-purpose I want to stick to checking whether javascript will see the property as an array element.

packages/replay-next/src/utils/protocol.ts
```diff
@@ -437,3 +430,7 @@ export function findProtocolObjectPropertyValue<Type>(
 ): Type | null {
   return findProtocolObjectProperty(sourceObject, name)?.value ?? null;
 }
+
+export function isArrayElement(property: ProtocolProperty): boolean {
```
bvaughn, line 434: This name confuses me. I know this is a nit, but can you elaborate on why you chose "array element"? I would assume that meant something more like "The value contained is an array"

hbenl: I couldn't think of a better word for an item in an array, I usually call them array elements.

PR #10236, opened 2024-01-26
Update playwright/breakpoints-05 example

hbenl:


packages/e2e-tests/tests/playwright-05_hover-dom-previews.test.ts
```diff
@@ -100,7 +100,7 @@ test("playwright-05: Test DOM node previews on user action step hover", async ({
       await stepWithMultipleNodes.hover({ timeout: 1000 });
       const count = await highlighter.count();
       await highlighter.first().waitFor({ state: "visible", timeout: 1000 });
-      expect(count).toBe(4);
+      expect(count).toBe(3);
```
bvaughn, line 103: Seems like this change maybe got merged prematurely, since RUN-3165 wasn't passed off to FE until this morning.

PR #10241, opened 2024-01-29
Fix object properties ordering

hbenl:
- don't add `getterValues` to the inline preview
- show `getterValues` below `properties` in the expanded preview (together with the special properties for `Promise` and `Proxy`)
- show the names of `getterValues` in grey

### DIFF
--- a/packages/replay-next/components/inspector/KeyValueRenderer.tsx
+++ b/packages/replay-next/components/inspector/KeyValueRenderer.tsx
@@ -26,7 +26,7 @@ export type Props = {
   onContextMenu?: (event: MouseEvent) => void;
   path?: string;
   pauseId: PauseId;
-  protocolValue: ProtocolValue | ProtocolNamedValue;
+  protocolValue: (ProtocolValue | ProtocolNamedValue) & { isGetterValue?: boolean };
 };
 
 // Renders a protocol Object/ObjectPreview as a key+value pair.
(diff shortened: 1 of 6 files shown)


packages/replay-next/components/inspector/KeyValueRenderer.tsx
```diff
@@ -132,6 +132,9 @@ export default function KeyValueRenderer({
       nameClass = styles.VerticalName;
     }
   }
+  if (protocolValue.isGetterValue) {
+    nameClass += " " + styles.GetterValue;
+  }
```
bvaughn, line 137: nit: An alternate way we sometimes do this is with a data attribute (e.g. `data-getter-value`). The primary benefit of this is that the corresponding CSS selector (class + data attribute) is [more specific](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), so it doesn't rely on the order of style names in the CSS file to set the correct color.

bvaughn (approved):

bvaughn:
I think this caused a regression. 4 tests are now failing in `main`, the same four as failed [on this branch](https://app.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs/c1040dbc-4a02-455f-bde7-27d65938fcad?param=dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI%3D&param=runs&param=60318204-9364-4829-b9ab-9b619d0e866b&param=tests&param=dHJ0OmNkNmZmYTMzODIzNTc5YTQxOWM0NjYyMDI3OWM2NDdmOTM5YTExMjk%3D).  Here is one, with comment: http://go/r/e433ab82-f093-4321-bf48-0014d3270e02

PR #10248, opened 2024-01-30
Fix object properties ordering

hbenl:


hbenl:
I'll look into the test failures tomorrow.

packages/replay-next/components/inspector/KeyValueRenderer.tsx
```diff
@@ -164,7 +166,11 @@ export default function KeyValueRenderer({
       {before}
       {name != null ? (
         <>
-          <span className={nameClass} data-test-name="KeyValue-Header">
+          <span
+            className={nameClass}
+            data-getter-value={isGetterValue}
```
bvaughn, line 171: This will simplify the CSS since it won't result in a `data-getter-value` attribute on a non-getter value.
```suggestion
            data-getter-value={isGetterValue || undefined}
```

packages/replay-next/components/inspector/values/ObjectRenderer.tsx
```diff
@@ -15,22 +13,11 @@ const MAX_PROPERTIES_TO_PREVIEW = 5;
...
-    return mergePropertiesAndGetterValues(
-      preview.properties || [],
-      preview.getterValues || [],
-      MAX_PROPERTIES_TO_PREVIEW
-    );
-  }, [preview]);
-
-  const showOverflowMarker = object.preview?.overflow || propertiesWereTruncated;
-
+  const properties = preview?.properties?.slice(0, MAX_PROPERTIES_TO_PREVIEW);
```
bvaughn, line 16: We were previously showing a mix of properties and getter values for collapsed objects.

This change means that objects like this...
Screenshot 2024-01-30 at 12 46 31 PM

Might now be rendered like this...
Screenshot 2024-01-30 at 12 46 34 PM

This seems like a regression. I think it's also why the [`console_async` test](https://tests.replay.io/recording/consoleasync-support-console-evaluations-in-async-frames--aef9d33f-2311-4218-8a2e-3db94a3deecd?e2e=&apiKey=&commentId=&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjI1OTYxNDg0Mjk1MTQ4NTcyODYzNDg0ODM0ODMxNDAzMDciLCJ0aW1lIjoxMjUxfSwiZW5kIjp7InBvaW50IjoiOTM0NjEzNDM0NjQyNzY1MzQ3MDg2OTQ3ODQ4NDU4MDc2MTYiLCJ0aW1lIjo2MTI5M319&point=89891639374026608437481669868407782&primaryPanel=cypress&secondaryPanel=console&time=58427&viewMode=non-dev) failed.

packages/replay-next/components/inspector/values/ObjectRenderer.tsx
```diff
@@ -15,22 +13,11 @@ const MAX_PROPERTIES_TO_PREVIEW = 5;
...
   let propertiesList: ReactNode[] | null = null;
-  if (context !== "nested") {
+  if (context !== "nested" && properties) {
```
bvaughn, line 18: Previously we were merging those two lists. Now that we aren't, if `properties` is empty but `getterValues` is not, this conditional will fail and we'll render "{...}" instead of "{ <list> }"

That's why [the `object_preview-01` test](https://tests.replay.io/recording/objectpreview-01-expressions-in-the-console-after-time-warping--e78b3650-ef3f-42c4-a112-85b6088095bc?e2e=&apiKey=&commentId=&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6Ijg0Mzc0ODIzOTU3OTc3OTQ0MjE4NTEwMDU5OTkxMjAzODQiLCJ0aW1lIjozMzEzfSwiZW5kIjp7InBvaW50IjoiNzM5OTAyMzAyMzY5OTIxOTYyMjczMTc2MzMwMjE2NDIzMDciLCJ0aW1lIjo2NjUwNX19&point=13954297809004117444498180966188422&primaryPanel=cypress&secondaryPanel=console&time=5502&viewMode=non-dev) is failing. What previously rendered as `Proxy{}` is now rendering as `Proxy{…}`

Maybe this change is "expected" and the test would just need to be updated. I'm not sure.

packages/replay-next/components/inspector/values/ObjectRenderer.tsx
hbenl: Both Firefox and Chromium debuggers show only own properties in the inline preview, so I've copied that behavior.
The test is failing because the underlying example recording was made before RUN-3135 was fixed, so the own properties `message` and `stack` didn't show up in `properties` even though they should.

packages/replay-next/components/inspector/values/ObjectRenderer.tsx
hbenl: That wasn't intentional, I'll fix it.

packages/replay-next/components/inspector/values/ObjectRenderer.tsx
bvaughn: > Both Firefox and Chromium debuggers show only own properties in the inline preview, so I've copied that behavior.

Got it 👍🏼 

> The test is failing because the underlying example recording was made before RUN-3135 was fixed, so the own properties message and stack didn't show up in properties even though they should.

Gotcha. I guess we're kind of stuck until that fix has been released then?

I guess we can disable the failing part of the test and file a blocked FE issue to update the recording once RUN-3135 has shipped.

hbenl: RUN-3135 has already been shipped in the `20240126` release but the example was still recorded with the `20240123` release. I've updated most examples in #10249.

bvaughn: I didn't even know we had a `20240126` release 😫 I wish it was easier to keep up with release status

bvaughn (approved):

PR #10251, opened 2024-01-31
Use test scope in passport tests

hbenl:


packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -72,6 +76,9 @@ const testWithCloneRecording = base.extend<TestIsolatedRecordingFixture>({
       if (newRecordingId) {
         await deleteTestRecording(newRecordingId);
       }
+      if (testUser && testScope) {
+        await resetTestUser(testUser, testScope);
```
bvaughn, line 80: Should we log a warning if a `testUser` is supplied without a `testScope`?

Or maybe better yet, can we update the type so it that it requires neither or both?

```ts
type TestIsolatedRecordingFixture = {
  exampleKey: TestRecordingKey;
  pageWithMeta: {
    page: Page;
    recordingId: string;
  };
};

type ScopedTestIsolatedRecordingFixture = TestIsolatedRecordingFixture & {
  testScope: string;
  testUser: string;
};

const testWithCloneRecording = base.extend<
  TestIsolatedRecordingFixture | ScopedTestIsolatedRecordingFixture
>({
```

That way we won't accidentally add any un-scoped authenticated tests?

packages/e2e-tests/authenticated/passport-01.test.ts
```diff
@@ -1,21 +1,27 @@
...
   pageWithMeta: { page, recordingId },
-  exampleKey,
+  testScope,
 }) => {
-  await resetTestUser("frontende2e1@replay.io");
-
-  await startTest(page, recordingId, E2E_USER_1_API_KEY);
+  const queryParams = new URLSearchParams();
+  queryParams.set("testScope", testScope!);
+  await startTest(page, recordingId, E2E_USER_1_API_KEY, queryParams);
```
bvaughn, line 24: Can we update the `startTest` helper methods to _require_ an API key always include a scope argument as well? e.g.

```ts
async function startTest(
  page: Page,
  recordingId: string
): Promise<void>;

async function startTest(
  page: Page,
  recordingId: string,
  apiKey: string,
  testScope: string
): Promise<void>;

async function startTest(
  page: Page,
  recordingId: string,
  apiKey?: string,
  testScope?: string
) {
  // TS will error if this is called with 3 args
}
```

hbenl:
- added the test user emails to `helpers/authentication.ts`
- moved generating the `testScope` to the fixture
- added the ability to reset multiple test users after a test
- moved adding the `testScope` query param to `startTest()`
- changed the `startTest()` signature so that callers can only pass both `apiKey` and `testScope` or none of the two

packages/e2e-tests/helpers/utils.ts
```diff
@@ -197,12 +197,18 @@ export async function locatorTextToNumber(locator: Locator): Promise<number | nu
   return trimmed ? parseInt(trimmed) : null;
 }
 
-export async function resetTestUser(email: string) {
+export async function resetTestUser(email: string, testScope?: string) {
```
bvaughn, line 200: We should probably require `testScope` since we know that calling this method _without_ a scope can break other running tests.

bvaughn (approved):

PR #10253, opened 2024-01-31
update @replayio/* deps

toshok:
only `@replay/cypress` being updated here, since `/playwright` and `/replay` were already updated.

part of RUN-3159

### DIFF
--- a/packages/e2e-tests/package.json
+++ b/packages/e2e-tests/package.json
@@ -21,7 +21,7 @@
   "license": "ISC",
   "devDependencies": {
     "@playwright/test": "^1.37.0",
-    "@replayio/cypress": "^0.5.0",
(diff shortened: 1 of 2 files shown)


bvaughn (approved):
We don't use Cypress in devtools, so I'm not sure what this is actually doing or being used for. I'd prefer to defer to someone else here for this (@ryanjduffy?)

toshok:
yeah, I see references to cypress in packages/e2e-tests (and imports of that particular @replayio npm package), but I can't trace any of it back to actual usage in our tests.  is there some code to remove?

ryanjduffy:
The dependency can be removed. It was used at one point to generate a recording with cypress for e2e tests but that code was removed so it isn't necessary now.

bvaughn:
Closing in favor of #10259

PR #10255, opened 2024-01-31
Rewrite graphics code

bvaughn:
- [x] Update screenshot/graphics code to render using a streaming Suspense cache
  - Added new declarative states for loading (using our standard loading bar) and errored
  - I think the new code, while still complex, is easier to read because it's not spread out in so many places
- [x] Update comments, recorded cursor, and highlighted DOM nodes rendering as well
- [x] Update visual comment rendering to use modern `typeData` instead of legacy `position` attribute.
   - We can remove the legacy attributes now! -> FE-2251
- [x] Replaced our recorded cursor graphic with an SVG, which renders a lot more crisply than it used to

---

Some UI improvements included:

1. Higher resolution cursor and click graphics

| Before | After |
| :--- | :--- |
| image | image |

3. New (standard) loading bar shown while streaming a repaint:

https://github.com/replayio/devtools/assets/29597/0605c7c6-107a-41ac-93c8-5ac14bda11cb

4. Explicit error state if screenshot could not be fetched

Screenshot 2024-02-08 at 2 56 21 PM


bvaughn:
Let's see how the e2e tests look. If they pass, I think this is ready for review. Smoke testing feels okay.

src/ui/components/Video/NodeHighlighter/PreviewNodeHighlighter.tsx
```diff
@@ -0,0 +1,225 @@
...
+      return subscribe(state => {
+        const { localScale, recordingScale } = state;
+        const scale = localScale * recordingScale;
+
+        element.style.transform = `scale(${scale})`;
+        element.style.height = `calc(100% / ${scale})`;
+        element.style.width = `calc(100% / ${scale})`;
+      });
+    }
+  }, []);
```
bvaughn, line 124: This is the only real change to this file. The rest of it was mostly just a `git mv` that it lost track of.

packages/protocol/PaintsCache.ts
```diff
@@ -41,17 +41,18 @@ export async function findFirstMeaningfulPaint() {
...
         );
+        if (value && value.hash) {
+          const { width, height } = await getDimensions(value.hash, value.mimeType);
 
-        const { width, height } = await getDimensions(hash, mimeType);
-
-        // Estimate how interesting the image is by the unique pixel density (given an arbitrary image size)
-        if (data.length > (width * height) / 40) {
-          return paint;
+          // Estimate how interesting the image is by the unique pixel density (given an arbitrary image size)
```
markerikson, line 52: :question: what does "interesting" mean here? What's special about `40`?

packages/protocol/StreamingScreenShotCache.ts
```diff
@@ -0,0 +1,102 @@
...
+  [replayClient: ReplayClientInterface, time: number, executionPoint: ExecutionPoint | null],
+  ScreenShot | undefined,
+  StreamingScreenShotCacheStatus
+>({
+  debugLabel: "StreamingScreenShotCache",
+  getKey: (replayClient, time, executionPoint) => `${time}:${executionPoint}`,
+  load: async ({ update, reject, resolve }, replayClient, time, executionPoint) => {
+    let screenShot: ScreenShot | undefined = undefined;
+
+    update(screenShot, 0, "loading-imprecise");
```
markerikson, line 37: :question: What is "imprecise" in this case?

src/ui/actions/timeline.ts
```diff
@@ -179,37 +163,18 @@ export async function getInitialPausePoint(recordingId: string) {
...
-
-      if (getHoverTime(stateAfterScreenshot) !== time) {
-        return;
-      }
-
-      paintGraphics(screen, mouse);
-    } catch (error) {
-      console.error(error);
-    }
+  return dispatch => {
```
markerikson, line 174: :hand: If this is no longer a thunk, then the entire body of `setHoverTime` could be replaced with just `return setTimeLineState(args)` (ie, it's just a semantic wrapper for the `setTimelineState` action)

src/ui/components/Video/MutableGraphicsState.tsx
```diff
@@ -0,0 +1,74 @@
...
+export function update({ element, scale }: { element: HTMLImageElement; scale?: number }) {
+  const { clientHeight, clientWidth, naturalWidth } = element;
+
+  const { left, top } = element.getBoundingClientRect();
+  const localScale = isNaN(clientWidth / naturalWidth)
+    ? mutableState.localScale
+    : clientWidth / naturalWidth;
+  const recordingScale = scale ?? mutableState.recordingScale;
+
+  if (
```
markerikson, line 55: :hand: Code style thought: given that we're really just creating a new version of the `mutableState` object, feels like this could be:

```ts
const newMutableState = {
  height: clientHeight,
  width: clientWidth,
  left,
  top,
  localScale,
  recordingScale,
}

if (!shallowEqual(mutableState, newMutableState)) {
  Object.assign(mutableState, newMutableState);
}
```

Just thinking of a way to make this less repetitive and easier to read.

markerikson (approved):
No major objections.  A couple spots that would be nice to see some conditional code restructured a bit for readability.

packages/protocol/PaintsCache.ts
bvaughn: You tell me 😄  I didn't write the above code or comment. The code is trying to calculate what % of an image is different pixels, as a proxy for what might be more than just a mostly-blank-screen. That approach– checking the first few frames– doesn't work well though, for apps like Replay, that have a lot of tiny paints as a result of a loading spinner or something. I think we should maybe just ditch this approach, _but_ it seemed out of the scope of this PR so I just left that comment.

packages/protocol/StreamingScreenShotCache.ts
bvaughn: Imprecise means it's the nearest cached paint. We show it first because it's reliably and fast to fetch, then we ask for a (more precise) repaint below (which may be slow and/or may fail to load entirely).

src/ui/actions/timeline.ts
bvaughn: Sure 👍🏼

src/ui/components/Video/MutableGraphicsState.tsx
bvaughn: Interesting. I have a slight preference to avoid the intermediate object creation and `shallowEquals` loop but it's only slight so I'll change it.

packages/protocol/PaintsCache.ts
markerikson: I'm actually kinda confused - if I pull `main` and look at this file, it looks like you added that code just yesterday in #10273 ?

https://github.com/replayio/devtools/pull/10273/files#diff-dfe818b2f76e449a14505ac73061fa1f303cd25d3a8476466eaf94b5663eb259R52

bvaughn: That was a small part of this PR that I landed in advance to make this review smaller. That approach came from here:

https://github.com/replayio/devtools/blob/99212711a2fc7d5b18bd84ae24bfc9a51b52360f/packages/protocol/graphics.ts#L512-L526

It was written three years ago by @jasonLaster (7c98f2a7d27f844e3e9d2f314be5dc2bfc7941f7)

packages/protocol/StreamingScreenShotCache.ts
bvaughn: I struggled a bit to think of the right terminology to use here. I guess I could have used "requesting-cached-paint" and "requesting-repaint"

Do you think those would be clearer? I don't mind changing this. It's only referenced in one place.

bvaughn:
Thanks for the review, Mark. I've pushed changes as a separate commit to make it easier to review them in isolation.

I still need to track down the cause of the failing `react_devtools-01` test before merging, but I'm looking into that next.

hbenl (approved):
Very nice!
I'll look into the e2e failure next. The weird thing is that I can't reproduce it, i.e. when I repeat the test steps manually, devtools always behaves correctly.

bvaughn:
Thanks @hbenl. I can't repro it either manually or locally using chromium, but I can using replay-chromium. I filed RUN-3250 for it.

PR #10265, opened 2024-02-03
Show different error on deleted recording

callingmedic911:
Closes SCS-1874. To distinguish deleted recording from unknown recording/permission issue, I'm changing behaviour in GraphQL, after we merge https://github.com/replayio/backend/pull/9450 it will throw an error with a unique code for deleted recording. This PR handles the change in devtools.
Screenshot 2024-02-02 at 6 11 09 PM


src/ui/setup/index.ts
```diff
@@ -182,7 +182,14 @@ export async function bootstrapApp() {
     const userInfo = await getUserInfo();
     if (userInfo) {
       const recordingId = getRecordingId();
-      const rec = recordingId ? await getRecording(recordingId) : null;
+      let rec = null;
+      if (recordingId) {
+        try {
+          rec = await getRecording(recordingId);
+        } catch (e) {
+          console.warn("Error fetching recording");
```
bvaughn, line 190: Why was this try/catch added? Is it related to the change you have in https://github.com/replayio/backend/pull/9450?

I think all of the places that call `getRecording` will now be try/catching it, except for the `RecordingCache`

bvaughn (commented):
Can you add a new e2e test for this case?

src/ui/setup/index.ts
callingmedic911: Pushed for the other case where it was not handled. But I don't think we go through this code path if we we fail to get the recording from one of earlier getRecording calls.

callingmedic911:
I'll create a separate PR for E2E tests because I'm thinking to merge this first, then deploy GraphQL changes.  E2E test will fail till I deploy GraphQL. If do deploys other way around, devtools will probably start seeing Sentry errors for anyone visits deleted recording page.

bvaughn (approved):
This is a difficult type of review to think through fully while I'm in the middle of another large architectural change. I trust you've thought about this carefully and I don't want to hold you up, so I'm approving this change.

You might want to also cc Holger on any future changes like this btw.

PR #10273, opened 2024-02-06
Incremental graphics refactor– paints cache, array and image utils

bvaughn:
Some standalone changes from #10255 (to de-risk the larger PR and make it easier to review)
- [x] Add new helper methods to `PaintsCache` (with unit tests); these will eventually replace the `protocol/graphics` methods
- [x] Update `waitForTime` helper method to resolve with the actual amount of time it waited for; (this will unblock a later change in #10255)
- [x] Added new `findIndexGTE` and `findIndexLTE` helper array utils method (and tests)
- [x] Added image utility methods `getDimensions` and `fitImageToContainer`

### DIFF
--- a/src/ui/actions/timeline.ts
+++ b/src/ui/actions/timeline.ts
@@ -47,7 +47,7 @@ import {
   paintGraphics,
   previousPaintEvent,
 } from "protocol/graphics";
-import { mostRecentPaint } from "protocol/PaintsCache";
+import { findMostRecentPaint } from "protocol/PaintsCache";
 import { waitForTime } from "protocol/utils";
 import { recordingCapabilitiesCache } from "replay-next/src/suspense/BuildIdCache";
 import {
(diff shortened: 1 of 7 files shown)


src/ui/actions/timeline.ts
```diff
@@ -526,7 +526,7 @@ export function playbackPoints(
       if (currentTime >= nextGraphicsTime) {
         try {
           let maybeNextGraphics = await Promise.race([nextGraphicsPromise, waitForTime(500)]);
-          if (!maybeNextGraphics) {
+          if (typeof maybeNextGraphics === "number") {
```
bvaughn, line 529: I don't know how this worked in `main`. It looks broken to me. 🤷🏼

hbenl: It worked because `waitForTime()` used to return `Promise<void>` (whereas now it returns `Promise<number>`).

hbenl (approved):

src/ui/actions/timeline.ts
bvaughn: Ah, right. That makes sense.

PR #10275, opened 2024-02-07
Fix GraphQLClient headers without a test scope

hbenl:
`GraphQLClient` without a test scope currently sets the header `"replay-test-scope": "undefined"`.

### DIFF
--- a/packages/shared/graphql/GraphQLClient.ts
+++ b/packages/shared/graphql/GraphQLClient.ts
@@ -40,15 +40,19 @@ export class GraphQLClient implements GraphQLClientInterface {
       data.query = data.query.loc?.source.body ?? "";
     }
 
+    const headers: Record<string, string> = {
+      "Content-Type": "application/json",
+    };
+    if (accessToken) {
+      headers["Authorization"] = `Bearer ${accessToken}`;
+    }
+    if (testScope) {
+      headers["replay-test-scope"] = testScope;
+    }
+
     const response = await fetch(this.url, {
       method: "POST",
-      headers: {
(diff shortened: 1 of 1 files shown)


packages/shared/graphql/GraphQLClient.ts
```diff
@@ -40,15 +40,19 @@ export class GraphQLClient implements GraphQLClientInterface {
...
+    if (testScope) {
+      headers["replay-test-scope"] = testScope;
+    }
+
     const response = await fetch(this.url, {
       method: "POST",
-      headers: {
-        ...(accessToken && {
-          Authorization: `Bearer ${accessToken}`,
-          "replay-test-scope": this.testScope,
```
bvaughn, line 48: I would expect this to send the header `"replay-test-scope": undefined` but you say you're seeing `"replay-test-scope": "undefined"`?

bvaughn (approved):

packages/shared/graphql/GraphQLClient.ts
hbenl: That's what I had expected as well (more precisely I expected headers with `undefined` values to not be sent - you can't set an HTTP header to `undefined`). I guess that `fetch()` coerces all header values to `string`.

bvaughn: Interesting. I had tested it in Chrome and it shows the header being sent as `undefined` (not a string). But looking again, I see that Chrome's Network panel only shows quotes for _certain_ strings and hides them for others. That's tricky :)

PR #10276, opened 2024-02-07
Try to fix seekToConsoleMessage() flakiness

hbenl:
Our e2e tests sometimes fail in `seekToConsoleMessage()` with an error message like
```
TimeoutError: locator.click: Timeout 60000ms exceeded.
=========================== logs ===========================
waiting for locator('[data-test-name="Message"]:has-text(\'Done\')').first().locator('[data-test-id="ConsoleMessageHoverButton"]')
  locator resolved to <button data-test-id="ConsoleMessageHoverButton" class="…>…</button>
attempting click action
  waiting for element to be visible, enabled and stable
element was detached from the DOM, retrying
============================================================
```
This sounds like playwright retries clicking the `MessageHoverButton` for 60 seconds but fails. Perhaps it is retrying with the same (detached) DOM element instead of getting a new DOM element for the given locator in every try. If that theory is correct then retrying the click ourselves may fix this flake.

### DIFF
--- a/packages/e2e-tests/helpers/console-panel.ts
+++ b/packages/e2e-tests/helpers/console-panel.ts
@@ -263,10 +263,17 @@ export async function seekToConsoleMessage(
 
   await debugPrint(page, `Seeking to message "${chalk.bold(textContent)}"`, "seekToConsoleMessage");
 
-  await consoleMessage.scrollIntoViewIfNeeded();
-  await consoleMessage.waitFor();
-  await consoleMessage.hover();
-  await consoleMessage.locator('[data-test-id="ConsoleMessageHoverButton"]').click();
+  await waitFor(async () => {
+    if ((await consoleMessage.getAttribute("data-test-paused-here")) === "true") {
+      return;
+    }
+    await consoleMessage.scrollIntoViewIfNeeded();
+    await consoleMessage.waitFor();
+    await consoleMessage.hover();
+    await consoleMessage
+      .locator('[data-test-id="ConsoleMessageHoverButton"]')
+      .click({ timeout: 1000 });
+  });
 
   await waitFor(
     async () =>


bvaughn (approved):
> Perhaps it is retrying with the same (detached) DOM element instead of getting a new DOM element for the given locator in every try.

I don't think that matches the behavior I've seen from Playwright for locators. The [Playwright documentation](https://playwright.dev/docs/api/class-page#page-locator) say:
> Locator is resolved to the element immediately before performing an action, so a series of actions on the same locator can in fact be performed on different DOM elements. That would happen if the DOM structure between those actions has changed.

That being said, explicitly retrying the `.click()` actions seems worth doing I think

hbenl:
> I don't think that matches the behavior I've seen from Playwright for locators.

Agreed, but that was the only theory I could come up with.

hbenl:
Hmm, the failure in [this recording](https://app.replay.io/recording/objectpreview-06-html-elements--692a156e-8a52-4082-96c7-8f5b95ac71a7) looks similar.

bvaughn:
Looking at that recording, it looks like the locator isn't finding the hover button (at least no DOM highlighting in Replay) even though it's in the DOM.
Screenshot 2024-02-07 at 8 57 47 AM

bvaughn:
Could try `click({ force: true })` ?

hbenl:
Aha, shortly after the message is hovered, it shifts down (because a previous message is loaded and becomes bigger) and then the hover button appears on a different message.
Perhaps we need to "rehover" when retrying.

bvaughn:
Oh! Yes. The hover button is visible, but it's not over the "ExampleFinished" log. I totally overlooked that.

hbenl:
Still flaking. But in [this recording](https://app.replay.io/recording/breakpoints-07-rewind-and-seek-using-command-bar-and-console-messages--95ff3cea-3854-46a5-ad17-ea8882414e43) I see another issue: `MessageHoverButton` is actually clicked (and we do seek to that message) but playwright thinks the click failed anyway. And after the click the hover button is not rendered anymore because we're already paused at that message, that's why the retries fail.

PR #10281, opened 2024-02-07
Fix potential source of duplicate recorded mouse events

bvaughn:
I noticed an oversight in the `RecordedEventsCache` that could result in calling `Session.findMouseEvents` twice in parallel, which (because of the way the `@replayio/protocol` API is designed) could result in duplicate events being added to the returned array.

This commit fixes that case by removing the 2nd caller. In general, this is a flaw of the `@replayio/protocol` package though and one we can't totally patch on the frontend.

See [go/r/c1cb4a29-e311-47da-975b-5ce14b10d654](http://go/r/c1cb4a29-e311-47da-975b-5ce14b10d654) for an example of the bug this commit fixes.


bvaughn:
This is a simple fix for a goofy error so I'm going to land it without a review (as soon as CI finishes).

PR #10283, opened 2024-02-08
Add test for delete recording error

callingmedic911:
Add tests for error dialog added in https://github.com/replayio/devtools/pull/10265

### DIFF
--- a/packages/e2e-tests/tests/deleted-recording.test.ts
+++ b/packages/e2e-tests/tests/deleted-recording.test.ts
@@ -1,0 +1,13 @@
+import test, { expect } from "@playwright/test";
+
+import { startTest } from "../helpers";
+
+const deletedRecordingId = "94f1729a-3aab-449b-98d6-10865a5e83c5";
+
+test("deleted-recording: Show error message for deleted recording", async ({ page }) => {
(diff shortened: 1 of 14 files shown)


packages/e2e-tests/tests/deleted-recording.test.ts
```diff
@@ -0,0 +1,13 @@
+import test, { expect } from "@playwright/test";
+
+import { startTest } from "../helpers";
+
+const deletedRecordingId = "94f1729a-3aab-449b-98d6-10865a5e83c5";
```
callingmedic911, line 5: I kept it outside of examples.json because it doesn't really depend on content of recording, just needs to be "deleted" recording.

bvaughn (approved):

bvaughn:
This new test is failing after I rebased my branch on `main`, e.g. https://app.replay.io/recording/deleted-recording-show-error-message-for-deleted-recording--f65c438c-cd21-4cf9-ba76-54409c73dfe0

The dialog is showing but it doesn't have the test id. Looks like somehow I got the new test but not the `ExpectedErrorModal` change after rebasing 😮

PR #10285, opened 2024-02-08
Better e2e error reporting when loading the recording fails

hbenl:


packages/replay-next/components/errors/UnexpectedErrorForm.tsx
```diff
@@ -31,7 +31,11 @@ export function UnexpectedErrorForm({
       currentUserName={currentUserName}
       details={
         <>
-          {details && <div className={styles.Details}>{details}</div>}
+          {details && (
+            <div className={styles.Details} data-test-id="UnexpectedErrorDetails">
+              {details}
+            </div>
+          )}
```
bvaughn, line 38: Putting the test id here seems a bit more fragile than putting it on the root of the unexpected error dialog, because it's possible that our helper would miss an unexpected fatal error if it didn't have a details string.

packages/e2e-tests/helpers/utils.ts
```diff
@@ -92,15 +92,28 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
...
+  } catch {
+    if (await timelineCapsuleLocator.isVisible()) {
+      throw new Error("Recording did not finish loading");
     }
-  );
+
+    const errorDetailsLocator = page.locator('[data-test-id="UnexpectedErrorDetails"]');
+    if (await errorDetailsLocator.isVisible()) {
+      throw new Error(`Recording crashed: ${await errorDetailsLocator.innerText()}`);
+    }
```
bvaughn, line 113: nit: I wonder if it might make more sense to check for this _first_? I think it works either way. Logically I just think of the error check as coming first 😜

bvaughn (approved):
I think the structuring in `waitForRecordingToFinishIndexing` is sound. I'd move the test id to be more certain of catching any/all unexpected errors.

packages/replay-next/components/errors/UnexpectedErrorForm.tsx
hbenl: I'll check for the visibility of the support form first and then extract the details if they are available (and otherwise throw a "Recording crashed" error without details).

bvaughn: Nice.

PR #10294, opened 2024-02-09
Adjust timing of prod_bundle timeouts to ensure sourcemap writes.

klochek:
As per https://linear.app/replay/issue/RUN-3254, I've discovered that our current missing sourcemap issues seem to stem from Chromium closing while we're waiting on a sourcemap fetch request to complete.

Fixing this correctly is hard (need to delay closing chromium if we have outstanding requests, but not indefinitely,) so punting on that for now so we can get some passing tests.

I think I've updated things correctly, though I note that `prod_bundle.js.map` now contains a slightly different path, which worries me.  (`"sources":["webpack:///./bundle_input.js"]` vs. `"sources":["webpack://prod_bundle/./bundle_input.js"]`)


hbenl (approved):
- when updating an example we should also update the corresponding recording in `examples.json` (this doesn't matter in RUN CI because there the examples are rerecorded anyway but it does matter in FE CI). You can use the recording I just made (with `ts-node scripts/save-examples.ts --runtime=chromium -e doc_prod_bundle.html`): f21d0b74-71a0-43ba-8e4a-db2f736ab8d5.
- I wouldn't worry about the changed path in the sourcemap: the example was quite old and I guess webpack's default URL template has changed since then. It doesn't affect the tests.

PR #10295, opened 2024-02-11
[FE-2255] `waitFor` ignores unrecoverable errors, including recorder crashes

Domiii:
* https://linear.app/replay/issue/FE-2255/waitfor-ignores-unrecoverable-errors-including-recorder-crashes

### DIFF
--- a/packages/e2e-tests/helpers/utils.ts
+++ b/packages/e2e-tests/helpers/utils.ts
@@ -95,7 +95,7 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
   try {
     await waitFor(
       async () => {
-        expect(await timelineCapsuleLocator.getAttribute("data-test-progress")).toBe("100");
+        expect(await timelineCapsuleLocator.getAttribute("data-test-progress"), "Recording did not finish processing").toBe("100");
       },
       {
         retryInterval: 1_000,
@@ -107,7 +107,7 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
       if (await page.locator('[data-test-id="SupportForm"]').isVisible()) {
         const errorDetailsLocator = page.locator('[data-test-id="UnexpectedErrorDetails"]');
         if (await errorDetailsLocator.isVisible()) {
-          throw new Error(`Recording did not finish processing: ${await errorDetailsLocator.innerText()}`);
+          throw new Error(`Session failed: ${await errorDetailsLocator.innerText()}`);
         }
       }
     } finally {
--- a/packages/e2e-tests/helpers/utils.ts
+++ b/packages/e2e-tests/helpers/utils.ts
@@ -184,10 +184,15 @@ export async function waitFor(
 
       return;
     } catch (error) {
-      if (!error?.matcherResult && !isString(error)) {
-        // Not an `expect` error and not one of our thrown strings.
-        // → Probably not recoverable.
+      if (error?.message?.includes("crash")) {
+        // We have to resort to heuristics since:
+        // 1. We don't have access to the `Page` object, and
+        // 2. the Error object also has no special properties.
         throw error;
+      }
+      if (!error?.matcherResult) {
+        // Not an `expect` error: → Log it.
+        console.log("ERROR:", error?.stack || error);
       }
 
       if (performance.now() - startTime > timeout) {
--- a/packages/e2e-tests/helpers/utils.ts
+++ b/packages/e2e-tests/helpers/utils.ts
@@ -106,7 +106,7 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
         timeout: 150_000,
       }
     );
-  } catch (err) {
+  } catch (err: any) {
     try {
       if (await page.locator('[data-test-id="SupportForm"]').isVisible()) {
         const errorDetailsLocator = page.locator('[data-test-id="UnexpectedErrorDetails"]');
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -31,34 +31,39 @@ const testWithCloneRecording = base.extend<TestIsolatedRecordingFixture>({
...
         resetOnNavigation: false,
       });
+
+      // Run test.
       await use({
         page,
         recordingId: newRecordingId,
         testScope,
       });
-    } catch (err: any) {
```
Domiii, line 52: This is the error handling for cloning, but it accidentally wrapped the entire test.

packages/e2e-tests/testFixtureCloneRecording.ts
```diff
@@ -31,34 +31,39 @@ const testWithCloneRecording = base.extend<TestIsolatedRecordingFixture>({
...
-      if (axios.isAxiosError(err)) {
-        console.error("Axios error cloning recording: ", {
-          errors: err.response?.data?.errors,
-          details: err.toJSON(),
-        });
-      } else {
-        console.error("Error cloning recording: ", err);
```
Domiii, line 59: This `else` is superfluous, if we are throwing anyway.

packages/e2e-tests/helpers/utils.ts
```diff
@@ -173,8 +175,10 @@ export async function waitFor(
 
       return;
     } catch (error) {
-      if (typeof error === "string") {
-        console.log(error);
+      if (!error?.matcherResult && !isString(error)) {
+        // Not an `expect` error and not one of our thrown strings.
+        // → Probably not recoverable.
+        throw error;
```
hbenl, line 181: :warning: This is not safe! It means we won't retry on some errors and for that we'd need a *reliable* way to determine if the error is "unrecoverable" (FWIW "recoverable" is not a fitting term here since an error doesn't necessarily mean that something went wrong, it usually only means something that we expect to happen hasn't happened yet).
In particular this would break `waitFor()` if we do FE-2256.

hbenl (changes requested):

packages/e2e-tests/helpers/utils.ts
Domiii: → VC?

packages/e2e-tests/helpers/utils.ts
```diff
@@ -95,27 +96,28 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
...
+        }
       }
-      throw new Error("Recording crashed");
-    }
-
-    if (await timelineCapsuleLocator.isVisible()) {
-      throw new Error("Recording did not finish loading");
+    } finally {
+      // Page is in a faulty state or no more info available. Stop trying and only report original error.
+      throw err;
```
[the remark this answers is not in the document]
Domiii: Fixed!

packages/e2e-tests/helpers/utils.ts
Domiii: We now only early-out on `crash` errors.

packages/e2e-tests/helpers/utils.ts
```diff
@@ -92,31 +110,25 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
...
-      const errorDetailsLocator = page.locator('[data-test-id="UnexpectedErrorDetails"]');
-      if (await errorDetailsLocator.isVisible()) {
-        throw new Error(`Recording crashed: ${await errorDetailsLocator.innerText()}`);
+
+  let supportFormErrorDetails: string | null = null;
+  await waitFor(
+    async () => {
+      supportFormErrorDetails = await getSupportFormErrorDetails(page);
+      if (supportFormErrorDetails) {
+        throw new UnrecoverableError(`Session failed: ${supportFormErrorDetails}`);
```
Domiii, line 119: This is an example of what we were talking about, @hbenl: Sometimes, things are unrecoverable, and we just don't want to retry.
But as per your suggestion, I added an explicit way to indicate unrecoverability, rather than assuming that most errors are unrecoverable.

packages/e2e-tests/helpers/utils.ts
```diff
@@ -84,6 +84,24 @@ export async function mapLocators<T>(
   );
 }
 
+export async function getSupportFormErrorDetails(page: Page) {
+  if (await page.locator('[data-test-id="SupportForm"]').isVisible()) {
+    const errorDetailsLocator = page.locator('[data-test-id="UnexpectedErrorDetails"]');
+    return (await errorDetailsLocator.innerText()) || "(support form is visible)";
```
hbenl, line 90: This will throw an error (instead of returning "(support form is visible)") if the `SupportForm` has no error details. You probably want something like `return (await errorDetailsLocator.isVisible()) ? await errorDetailsLocator.innerText() : "(support form is visible";`

hbenl (changes requested):

packages/e2e-tests/helpers/utils.ts
Domiii: Ah! Learned something new.
How is this?

hbenl: Many locator methods work that way: they wait for the located element to become visible and throw a `TimeoutError` if that doesn't happen in time. `locator.isVisible()` is one of those that doesn't wait (and doesn't throw).

hbenl (approved):

hbenl (changes requested):
Don't forget to fix the typescript error.

packages/e2e-tests/helpers/utils.ts
Domiii: Ah whoops... What I meant was: "How is this new version?" - But thanks, it confirms my guesses 😆

hbenl (approved):

PR #10301, opened 2024-02-12
Add back missing fields to GetTestRunRecordings graphql

ryanjduffy:
`index` and `attempt` are expected for the tests within a test run but were dropped from the graphql for fetching test runs


bvaughn (approved):
Wow. These were dropped two and a half months ago (#9956)

How much of our Test Suites UI was that breaking?

ryanjduffy:
> Wow. These were dropped two and a half months ago (#9956)
> 
> How much of our Test Suites UI was that breaking?

There were a couple sorts in the test runs view that were presumably broken. Opening another ticket to add e2e tests to cover that feature.

bvaughn:
Awesome.

PR #10302, opened 2024-02-12
Update "flake/adding-spec.ts" example

hbenl:


packages/e2e-tests/tests/cypress-01_basic-panel.test.ts
```diff
@@ -75,10 +75,10 @@ test("cypress-01: Basic Test Suites panel functionality", async ({
 
   // Relative dates can change over time.
   // Check for either the "X units ago" text, or the literal date.
-  expect(await getTestSuiteDate(page).textContent()).toMatch(/ ago|(7\/14\/2023)/);
+  expect(await getTestSuiteDate(page).textContent()).toMatch(/ ago|(2\/12\/2023)/);
```
bvaughn, line 78: This test just started failing. Seems like we exceeded the relative date threshold:
```
// Show relative time if under 2 weeks, otherwise, use the template below.
if (daysSince > 14) {
  content = formatter.format(new Date(date));
}
```

PR #10308, opened 2024-02-13
Don't request an empty paint point

bvaughn:
This was a silly oversight.

### DIFF
--- a/packages/protocol/StreamingScreenShotCache.ts
+++ b/packages/protocol/StreamingScreenShotCache.ts
@@ -43,7 +43,7 @@ export const StreamingScreenShotCache = createStreamingCache<
 
     const paintPoint = findMostRecentPaint(time);
 
-    if (paintPoint) {
+    if (paintPoint && paintPoint.paintHash) {
       try {
         screenShot = await screenshotCache.readAsync(
           replayClient,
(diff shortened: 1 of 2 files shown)


packages/protocol/StreamingScreenShotCache.ts
```diff
@@ -43,7 +43,7 @@ export const StreamingScreenShotCache = createStreamingCache<
 
     const paintPoint = findMostRecentPaint(time);
 
-    if (paintPoint) {
+    if (paintPoint && paintPoint.paintHash) {
```
jcmorrow, line 46: Out of curiosity, do you usually opt for this syntax over `paintPoint?.paintHash`?

jcmorrow (approved):

packages/protocol/StreamingScreenShotCache.ts
bvaughn: It's subjective, but I think it's more explicit that we're accounting for two cases: no paint found and an empty paint

markerikson (approved):

bvaughn:
Not sure what's going on with the backend. Lots of e2e test failures because of session creation failing.

jcmorrow:
> **Edit** Seems like a spike in `processRecording` on the backend that's triggering session failures at a pretty high rate. Even an hour later, it's still happening.

I'll take a look. This *shouldn't* be happening because our E2E tests are isolated from most processing sessions, but maybe there's a lot of traffic in the Live Tests workspace or something.

bvaughn:
The tests failing on this branch are the exact tests failing on `main`, which is to say, the ones that backend dropped support for due to the age of the runtimes used to record them. So I'm going to land this.

PR #10309, opened 2024-02-13
Remove unneeded comment code

jcmorrow:
Before we can deploy https://github.com/replayio/backend/pull/9495 we have to stop requesting these fields from GraphQL. I tested this in cloud dev and it seemed to work, but I'm not making any strong claims that I did this correctly 😛 . Also, I hand-edited the apollo fixtures, which might be wrong, and I did *not* hand-edit `globalTypes.ts`, but I also do not know how to automatically update it.

### DIFF
--- a/packages/shared/graphql/generated/GetComments.ts
+++ b/packages/shared/graphql/generated/GetComments.ts
@@ -36,16 +36,11 @@ export interface GetComments_recording_comments {
   id: string;
   isPublished: boolean | null;
   content: string;
-  primaryLabel: string | null;
-  secondaryLabel: string | null;
   createdAt: any;
   updatedAt: any;
   hasFrames: boolean;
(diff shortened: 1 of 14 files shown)


jcmorrow:
Oh, interesting, it looks like https://github.com/replayio/devtools/commit/1e0c672638d23577ca90543e01d9c156f4818c2d did most of this, but for some reason my local is still failing even after I pulled. Lemme double check.

bvaughn:
I also rebased on `main` since e2e tests are now passing again.

bvaughn:
I _think_ `globalTypes.ts` might be reflecting all possible attributes, and so wouldn't reflect the removal of these fields untilt he GraphQL server removes them?

bvaughn (approved):
Looks good. Let's do it.

PR #10311, opened 2024-02-13
Manually update unsupported gecko e2e recordings

bvaughn:
Re-record all `macOS-gecko-20220722` e2e test recordings with `macOS-gecko-20230919`


bvaughn:
Looks like this is a duplicate of @hbenl's efforts on #10310

PR #10313, opened 2024-02-13
Fix RDT race case

hbenl:


src/ui/components/SecondaryToolbox/react-devtools/injectReactDevtoolsBackend.ts
```diff
@@ -246,21 +271,10 @@ export const nodesToFiberIdsCache: Cache<
   async load([replayClient, pauseId, store]) {
```
bvaughn, line 271: We should be able to remove the `store` parameter from this cache now.

lol we can probably delete that comment on line 269 as well

bvaughn (approved):

bvaughn:
Hmm. `repaint.test` failed on this PR and on `main` after this was merged.

PR #10315, opened 2024-02-13
Migrate repaint test from Gecko to Chromium

bvaughn:
- [x] Delete old `doc_control_flow_firefox.html` example; this recording is now broken anyway (RUN-3260)
- [x] Update `repaint.test` to use Chromium `doc_control_flow.html` recording instead


hbenl:
This probably won't work since we've seen the same repainting issue in chromium (see also https://linear.app/replay/issue/RUN-2916/[metabase]-repaintgraphics-still-outdated#comment-443f72dd).

bvaughn:
It passes reliably for me locally.

hbenl (approved):

hbenl:
Let's see if this flakes like it did with the new gecko recording (where it mostly failed but sometimes it passed).

PR #10321, opened 2024-02-14
Remove sorting by test.attempt number 

callingmedic911:
Closes SCS-1946. It turns out we don't need to use the test because it's legacy code that doesn't impact current view. We had this sorting line, in the new view, the second panel of the test run. It was trying to sort tests by attempt number but it doesn't make sense anymore because we have unique test entries, and we don't list multiple attempts in the second panel.

Also, GraphQL generates a list of tests and sends them back to the client, the attempt number on the root level is not useful. _I think_ that's built up from the list of test_run_tests table and picks the oldest one (or one at random). That field is useless and should be removed.

Being lazy, used the old commit and reverted that.


ryanjduffy (commented):
Let's also remove it from the graphql query and re-run the type generator so all that is cleaned up too. Thanks for digging into this and sorting it out, @callingmedic911 !

packages/shared/graphql/generated/globalTypes.ts
```diff
@@ -11,13 +11,8 @@ export interface AddCommentInput {
   content: string;
```
callingmedic911, line 11: Unrelated changes in the file but it seem harmless, so adding to the commit.

ryanjduffy (approved):

callingmedic911:
I didn't mean to merge without FE approval. The impact seems limited, I'm not reverting immediately but @hbenl if you disagree (or have any comment) I can address it in a separate PR.

PR #10322, opened 2024-02-14
[RUN-3243] Ignore requiresManualUpdate samples for runtime tests

Domiii:
* https://linear.app/replay/issue/RUN-3243/ignore-requiresmanualupdate-samples

### DIFF
--- a/packages/e2e-tests/scripts/get-stats.ts
+++ b/packages/e2e-tests/scripts/get-stats.ts
@@ -2,6 +2,17 @@ import { lstatSync, readFileSync, readdirSync } from "fs";
 import { join, relative } from "path";
 
 type Examples = typeof import("../examples.json");
+
+/**
+ * Note: typeof does not include optional props, so we need to
+ * cast to this instead.
+ */
+type AllExampleProps = {
+  recording: string;
+  buildId: string;
+  playwrightScript?: string;
+  requiresManualUpdate?: string;
+};
 type Stats = {
   [buildId: string]: {
     numRecordings: number;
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/scripts/get-stats.ts
```diff
@@ -2,6 +2,17 @@ import { lstatSync, readFileSync, readdirSync } from "fs";
...
+
+/**
+ * Note: typeof does not include optional props, so we need to
+ * cast to this instead.
+ */
+type AllExampleProps = {
+  recording: string;
+  buildId: string;
+  playwrightScript?: string;
+  requiresManualUpdate?: string;
```
hbenl, line 14: Why do you treat this as `string`? It's a `boolean` in the `examples.json` file.

packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -116,6 +109,14 @@ async function saveRecording(
     `Saving ${chalk.grey.bold(example)} with recording id ${chalk.yellow.bold(recordingId)}`
   );
 
+  if (!skipUpload) {
+    await uploadRecording(recordingId, {
```
hbenl, line 113: Why did you move this down again? I had moved it up to make it work for node examples.

Domiii: Because of the `console.log` message. I'll try moving them both up - is that Ok?

packages/e2e-tests/scripts/get-stats.ts
Domiii: Mistake

hbenl (approved):

PR #10323, opened 2024-02-14
Make inspector-elements-05 test deterministic. Insert iframe dynamically.

klochek:
There's no reason to assume that halfway through the recording the iframe won't be there, so make that logic dependable.


markerikson:
Did you re-record the example?

klochek:
> Did you re-record the example?

Found the API key :)  Done!

bvaughn:
Love this change. Thanks!

PR #10225, opened 2024-01-24
Enable test reports uploads to Trunk

zaycev:

PR #10324, opened 2024-02-14
Enable Junit uploads to Trunk.io

zaycev:
This PR configures test runners to write Junit XML to `test-results/` and adds a new step to upload Junit reports to [Trunk.io](https://docs.trunk.io/flaky-tests). This will enable our tool to provide flaky tests analytics (we will follow up with more info in Slack).

In order for this to work we would need someone from `replayio` to add Trunk secret: `TRUNK_API_TOKEN`.  You can find Trunk token by navigating to app.trunk.io → Settings → Manage Organization → Organization API Token → View.

- Uploader action: https://github.com/trunk-io/analytics-uploader
- Uploader CLI: https://github.com/trunk-io/analytics-cli

Previous attempt: [#10225](https://github.com/replayio/devtools/pull/10225)


.github/workflows/test.yml
```diff
@@ -217,6 +233,14 @@ jobs:
           YARN_CHECKSUM_BEHAVIOR: "update"
       - name: Run tests
         run: yarn test
+      - name: Upload test results to Trunk
+        uses: trunk-io/analytics-uploader@main
+        with:
+          org-slug: replay
+          junit-paths: test-results/**/*.xml
+          token: ${{ secrets.TRUNK_API_TOKEN }}
```
zaycev, line 241: The secret which we need to configure for this action.

bvaughn:
@zaycev I appreciate you taking the time to post this PR

I think you talked with @jasonLaster about this originally but I don't have much personal context on why we would want to make this update. It seems like something that fell through the cracks.

I feel kind of bad doing this, but I don't like leaving stale PRs open if they aren't going to land, so I'm going to close the PR.

PR #10338, opened 2024-02-18
[RUN-3246] Add playwright meta data to RUN test runs

Domiii:
* https://linear.app/replay/issue/RUN-3246/fix-run-fe-e2e-test-runs-to-produce-playwright-metadata#comment-ab62dde9


bvaughn:
This is not a big deal, but @Domiii  in the future please at least cc Holger and I on all PRs to FE (and preferably wait for them to be approved)

PR #10341, opened 2024-02-20
Add delay at the end of the redux-fundamentals e2e example

hbenl:


packages/e2e-tests/examples/redux-fundamentals/tests/example-script.ts
```diff
@@ -140,4 +140,6 @@ export default async function testFunction(
 
   await filtersPanel.getByText('Green').click()
   await waitForListItemsCount(2)
+
+  await delay(500);
```
bvaughn, line 144: This PR doesn't explain the "why" for this change. Why is this necessary? Why wouldn't it be better for us to wait on something _specific_ rather than a seemingly arbitrary number of milliseconds?

Domiii: ## Missing Linkage
What I always do is:
* I link the Linear issue, and specifically a final `SLN` comment under that issue that explains the rationale behind a change. In this case, the rationale is also exlpained in the issue, but its not linked directly. Having that direct link makes things a lot easier!
* If a change is a hackfix for an issue, I always add the issue as a comment to the hackfix in the code.


## To answer your question
TL;DR: This is to paper over what we believe to be an outstanding (and very hard-to-fix) runtime issue RUN-3258.
Ref: [here](https://linear.app/replay/issue/FE-2286/jump-to-code-01-not-enough-events#comment-edf615b7) is the link 😆

bvaughn: @hbenl If that's what this is for it would be nice to have a comment linking it to that issue. (If I saw this, a random delay at the end of a test, I would assume it was left over from a previous refactor and fair game to delete.)

PR #10351, opened 2024-02-22
Update doc_exceptions example

hbenl:


packages/e2e-tests/tests/logpoints-04.test.ts
```diff
@@ -24,7 +24,7 @@ test.skip(`logpoints-04: should display exceptions in the console`, async ({
   await page.locator('[data-test-id="FilterToggle-exceptions"]').click();
 
   let messages = await findConsoleMessage(page, undefined, "exception");
-  await expect(messages).toHaveCount(20);
+  await expect(messages).toHaveCount(10);
```
hbenl, line 27: The behavior has changed compared to gecko, there seems to be agreement that the changed behavior is what we want, see [this Linear thread](https://linear.app/replay/issue/RUN-3289/exception-only-reported-on-first-frame#comment-e5eef116)

PR #10353, opened 2024-02-22
Bubble session timeouts (or unknown session errors) to the root

bvaughn:
Rather than display inline error fallbacks in the event of a session failure, bubble those errors to the root and show our expected error dialog:
Screenshot 2024-02-22 at 1 34 49 PM

cc @jonbell-lot23 as this impacts UX


jasonLaster (commented):
I like this change. I think it would benefit from an E2E test where we release the session, try to do something, and assert that we show this error modal.

markerikson (approved):

packages/replay-next/components/errors/withSessionTimeoutCheck.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { ProtocolError, isCommandError } from "shared/utils/error";
+
+export function withSessionTimeoutCheck<Props extends object>(component: ComponentType<Props>) {
+  const Wrapper = (props: Props) => {
+    const context = useContext(ErrorBoundaryContext);
+    assert(context, "ErrorBoundaryContext is not available");
+
+    const { didCatch, error } = context;
+    if (didCatch) {
```
bvaughn, line 13: Note to self: This won't handle the case where we show this inline fallback _manually_ (like in the source viewer)

bvaughn:
Thanks for the prompt to add an e2e test for this. Doing that made me realize that there was a more straight forward way to handle this, given the way our `socket` module is currently organized (which maybe we'll eventually clean up some more– but as it is...)

I added an e2e test for this behavior too.

packages/protocol/socket.ts
```diff
@@ -217,11 +217,6 @@ export async function sendMessage<M extends CommandMethods>(
 
     const { code, data, message } = response.error;
 
-    if (method === "Session.listenForLoadChanges" && code === 66) {
-      // We are being disconnected after a timeout, no need to raise
-      return {};
-    }
```
bvaughn, line 223: Now that we're listening for these exceptions explicitly, we shouldn't need to special case them here any more.

bvaughn:
The test failure in this PR is a false positive. It's failed in several other recent branches.

src/ui/setup/helpers.ts
```diff
@@ -28,6 +32,22 @@ export async function setupAppHelper(store: UIStore) {
   const recordingId = getRecordingId();
   const replaySession = recordingId ? await getReplaySession(recordingId) : undefined;
 
+  window.addEventListener("unhandledrejection", function (event) {
+    const { reason } = event;
+
+    if (
+      reason &&
+      (isCommandError(reason, ProtocolError.UnknownSession) ||
+        isCommandError(reason, ProtocolError.SessionDestroyed))
```
hbenl, line 41: We could handle these errors in `socket.ts`, that would also not only handle them when they're uncaught.

bvaughn: Can you be more concrete about what you're proposing?

I thought about handling them in `socket.ts` first, but then I would have to add another module-level event listener for that specific error type, so I just did this instead. I'm not strongly opposed to moving them back into there, but I don't love the `gEventListeners`/`gEventToCallbacksMap` approach and didn't want to lean more into it.

bvaughn: Tell me if you like https://github.com/replayio/devtools/pull/10353/commits/9bd1aed7d72d53ba29b7d7f989cb57d17ea698ea better.

src/ui/setup/dynamic/devtools.ts
```diff
@@ -9,7 +9,9 @@ import * as dbgClient from "devtools/client/debugger/src/client";
 import debuggerReducers from "devtools/client/debugger/src/reducers";
 import * as inspectorReducers from "devtools/client/inspector/reducers";
 // eslint-disable-next-line no-restricted-imports
-import { addEventListener, initSocket, client as protocolClient } from "protocol/socket";
+import { addEventListener, initSocket } from "protocol/socket";
+// eslint-disable-next-line no-restricted-imports
+import { listenForSessionDestroyed, client as protocolClient } from "protocol/socket";
```
bvaughn, line 14: This is goofy, but all 4 imports being in the same statement causes Prettier to re-format the import to be multi-line, which breaks the Lint rule (bc there's no way to ignore it at that point)

src/ui/setup/helpers.ts
hbenl: That additonal listener is unfortunate but I still prefer this to be handled in `socket.ts`.

hbenl (approved):

src/ui/setup/helpers.ts
bvaughn: Cool. I have no strong preference here so let's go with that then.

PR #10354, opened 2024-02-22
Remove protocol viewer experience; always show panel for internal users

bvaughn:


markerikson:
I think Jason once said he wanted to be able to hide it when talking to customers so they don't wonder what the extra button is, but I don't know if that's still relevant (or if he's ever bothered toggling it)

bvaughn:
Ah, interesting.

Mine got toggled off by accident somehow, and it confused me because I couldn't find it. (I didn't even realize/remember that we had a setting for it.)

PR #10355, opened 2024-02-22
Not for review

klochek:
Not for review


packages/e2e-tests/tests/console_warp-02.test.ts
```diff
@@ -56,6 +58,10 @@ test("console_warp-02: support pausing, warping, stepping and evaluating console
   await stepOverToLine(page, 21);
   await verifyPausedAtMessage(page, "1 << 7", "terminal-expression");
 
+  // TODO: extra line
+  if (BrowserName == "chromium") {
+    await stepOverToLine(page, 21);
+  }
```
jasonLaster, line 64: cool. but lets just make this a chromium test and stop using the firefox example.

packages/e2e-tests/tests/stepping-05_chromium.test.ts
```diff
@@ -33,19 +33,14 @@ test(`stepping-05_chromium: Test stepping in pretty-printed code`, async ({
   await resumeToLine(page, 8);
   await stepOverToLine(page, 8);
 
-  // TODO Stepping here hangs
-  // await stepOverToLine(page, 9);
+  // TODO: extra test needed.
+  await stepOverToLine(page, 8);
+  await stepOverToLine(page, 9);
```
bvaughn, line 38: What does that TODO comment mean?

Can you add more explanation and/or link to a Linear task?

bvaughn (commented):
I think `stepping-05_chromium` may have "regressed" due to a BAC bug fix (see [here](https://discord.com/channels/779097926135054346/1210355430484869160))

That sounds different from the other tests in this PR? In which case, can we split that out into its own PR so we can land it now? (It's failing on `main` too)

**Edit** I did that with #10356

PR #10356, opened 2024-02-22
Update stepping-05_chromium.test to match new (fixed) BAC stepping behavior

bvaughn:


markerikson (approved):

bvaughn:
I think we could beef this test out more. I just want to unbreak `main` for now.

PR #10358, opened 2024-02-23
Update prefer-current-time check to account for focus window

bvaughn:
- [x] Update prefer-current-time check to account for focus window
- [x] Add more e2e tests for repaint behavior

### DIFF
--- a/packages/e2e-tests/helpers/screenshot.ts
+++ b/packages/e2e-tests/helpers/screenshot.ts
@@ -1,7 +1,7 @@
-import { Page } from "@playwright/test";
+import { Page, expect } from "@playwright/test";
 import chalk from "chalk";
 
-import { debugPrint } from "./utils";
+import { debugPrint, waitFor } from "./utils";
 
 export async function convertCoordinatesForScreenshot(
   page: Page,
@@ -30,6 +30,10 @@ export async function convertCoordinatesForScreenshot(
 
 export function getGraphicsElement(page: Page) {
   return page.locator("#graphics");
+}
+
+export function getVideoElement(page: Page) {
+  return page.locator("#video");
 }
 
 export async function clickScreenshot(page: Page, xPercentage: number, yPercentage: number) {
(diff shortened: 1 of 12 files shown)


hbenl (approved):

packages/e2e-tests/helpers/screenshot.ts
```diff
@@ -66,9 +70,54 @@ export async function hoverScreenshot(page: Page, xPercentage: number, yPercenta
...
+  return await element.getAttribute("data-status");
+}
+
+export async function getGraphicsTime(page: Page): Promise<number | null> {
+  const element = getVideoElement(page);
+  const value = await element.getAttribute("data-time");
+  return value != null ? parseInt(value) : null;
+}
+
+export async function getGraphicsPixelColor(page: Page, x: number, y: number) {
```
bvaughn, line 96: This helper function isn't _necessary_ but it makes `expect` failures small and easier to read than if we were comparing the full data url.

packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -194,9 +194,13 @@ async function saveExamples(
...
-      throw new Error(
-        `Invalid examples don't exist or require manual recording: ${invalidInputs.join(",")}`
-      );
+      if (argv.target !== "all") {
+        // If no target is specified, this script will check both Chrome and Node examples;
+        // we shouldn't throw in that case (since it's likely that at least one target will contain no matches)
+        throw new Error(
+          `Invalid examples don't exist or require manual recording: ${invalidInputs.join(",")}`
+        );
+      }
```
bvaughn, line 203: This got broken somewhere along the way. Not sure when. Lots of folks were tweaking this file.

packages/protocol/StreamingScreenShotCache.ts
```diff
@@ -42,6 +43,12 @@ export const StreamingScreenShotCache = createStreamingCache<
     await PaintsCache.readAsync();
 
     const paintPoint = findMostRecentPaint(time);
+    if (!paintPoint || !paintPoint.paintHash) {
+      // Don't try to paint (or repaint) if the current time is before the first cached paint
+      update(undefined, 1, "before-first-paint");
+      resolve();
+      return;
+    }
```
bvaughn, line 51: We weren't handling this case very well. The new repaint 04 test highlighted that.

src/ui/components/Video/useSmartTimeAndExecutionPoint.ts
```diff
@@ -16,22 +18,34 @@ export function useSmartTimeAndExecutionPoint() {
...
-  //
-  // A heuristic for detecting this case is comparing how close the two values are
-  const preferCurrentTime = playbackState != null || Math.abs(pauseTime - currentTime) > 0.05;
+  let preferCurrentTime = false;
+  if (playbackState != null) {
+    preferCurrentTime = true;
+  } else if (
+    focusWindow &&
+    !isTimeInRegion(currentTime, focusWindow) &&
+    Math.abs(pauseTime - currentTime) > 0.05
```
bvaughn, line 29: I think this change is sufficient to fix what @hbenl originally reported. (And I believe @hbenl verified that fix on this preview branch.)

I'm not _yet_ sure how to recreate the exact failure that was reported in an e2e test, but I'm going to poke at it a bit more.

bvaughn:
Bunch of e2e tests are failing on this PR but I think it's related to #9594

packages/e2e-tests/scripts/save-examples.ts
Domiii: I remember adding the `if (invalidInputs.length) { ... }` part - Not sure I ever saw the `argv.target` part 👀

PR #10359, opened 2024-02-23
Fix elements-search test flakiness

hbenl:


packages/e2e-tests/helpers/elements-panel.ts
```diff
@@ -423,15 +423,18 @@ export async function searchElementsPanel(
   await input.focus();
   await input.type(searchText);
 
+  const resultsLabel = page.locator('[data-test-id="ElementsPanel-SearchResult"]');
   await waitFor(async () => {
-    await input.press("Enter");
+    // don't press "Enter" again if the search is already running or has just finished
+    if (!page.locator('[data-test-id="ElementsPanel-Searching"]').or(resultsLabel)) {
```
bvaughn, line 429: Huh, I didn't know locators could be changed like that

bvaughn (approved):

PR #10360, opened 2024-02-23
Use error codes for auth errors

callingmedic911:
Closes SCS-1949.

The ultimate goal is to improve the error message for SSO Login error but while we are setting up error codes in Auth0, we are changing it for all of the cases.


ryanjduffy (approved):

callingmedic911:
@ryanjduffy before merging this, we need to double-check check changes in Auth0 Rules are in place.

src/ui/components/shared/Login/Login.tsx
```diff
@@ -175,21 +175,33 @@ function AuthError({ error }: { error: any }) {
     return null;
   }
 
-  if (message === "Invalid state") {
-    // This is usually caused by waiting too long to go through the auth process
-    // and can be fixed by trying again.
-    message = "Your login session expired. Please try logging in again.";
-  } else {
-    // We want to capture any other error so we can investigate further.
+  if (message !== "INVALID_STATE" && message !== "Invalid state") {
```
bvaughn, line 178: Where do these strings come from (e.g. "INVALID_STATE")

bvaughn, line 178: Where do these message strings come from (e.g. "INVALID_STATE")? I'm curious because usually this type of message has some kind of constant we could import/reference from the package we're using (rather than hard-code strings).

Side note: Can you add a comment explaining that "Invalid state" comes from Auth0 and that's why we also check for this different looking message here? Maybe we could link to the docs, e.g. https://auth0.com/docs/customize/integrations/cms/wordpress-plugin/troubleshoot-wordpress-plugin-invalid-state-errors

callingmedic911: This comes from our Auth0 Rule, something we define, except "Invalid state". Ryan and I plan to implement enum-like errors before merging this PR, moving away from string matching on human-readable messages.

callingmedic911: Except "Invalid state", we define every other error type in Auth0 rules.

Adding comment.

bvaughn: Sorry I don't know how I seem to have added the same comment twice 😆

src/ui/components/shared/Login/Login.tsx
```diff
@@ -177,21 +177,37 @@ function AuthError({ error }: { error: any }) {
...
     sendTelemetryEvent("devtools-auth-error-login", {
       errorMessage: message,
     });
+  }
 
-    if (message === "Unable to authenticate user") {
+  // Except of "Invalid state" message that comes from
+  // https://auth0.com/docs/customize/integrations/cms/wordpress-plugin/troubleshoot-wordpress-plugin-invalid-state-errors
+  // we define all the other "message"s in our Auth0 Rule.
+  // See more here: https://auth0.com/docs/customize/rules/raise-errors-from-rules
```
bvaughn, line 189: nit: You might move this comment up, above line 180, where that string is first referenced. Otherwise this is very helpful though 👍🏼

src/ui/components/shared/Login/Login.tsx
```diff
@@ -177,21 +177,37 @@ function AuthError({ error }: { error: any }) {
...
-    }
+      break;
+    case "IDP_CONFIGURATION_ERROR":
+      message =
+        "Failed to log in due to a configuration problem with your Replay account. Support has been notified!";
+    case "IDP_UNEXPECTED_ERROR":
+      message = "Failed to login. Please try logging in with SSO.";
+      break;
+    default:
+      // We want to capture any other error so we can investigate further.
```
bvaughn, line 209: Should we be capturing anything here (in Sentry) btw?

bvaughn (approved):

src/ui/components/shared/Login/Login.tsx
callingmedic911: The comment was misleading. I forgot to move that up where we are capturing telemetry events, for which Ryan gets an alert.

Also from Ryan:
> I don't think it's helpful in Sentry because the cause of the error isn't in devtools. Had we sentry integration in Auth0, it might be more useful there but I think telemetry is the best we can do right now.

src/ui/components/shared/Login/Login.tsx
```diff
@@ -177,21 +177,38 @@ function AuthError({ error }: { error: any }) {
...
+    case "INTERNAL_ERROR":
       // This usually occurs because our auth hook threw an error but the
       // message itself isn't very useful so we show a more friendly message
       message = "We're sorry but we had a problem authenticating you. We're looking into it now!";
-    }
+      break;
+    case "IDP_CONFIGURATION_ERROR":
+      message =
+        "Failed to log in due to a configuration problem with your Replay account. Support has been notified!";
+    case "IDP_UNEXPECTED_ERROR":
```
ryanjduffy, line 207: missing a `break` here

callingmedic911: whoops

PR #10361, opened 2024-02-26
Handle errors from DOM.getAllBoundingClientRects

hbenl:


src/ui/components/NodePickerContext.tsx
```diff
@@ -153,7 +153,16 @@ export function NodePickerContextRoot({ children }: PropsWithChildren<{}>) {
...
+    let boundingRectsPromise: PromiseLike<NodeBounds[]> | NodeBounds[] | undefined;
+    try {
+      boundingRectsPromise = boundingRectsCache.readAsync(replayClient, pauseId);
+    } catch {
+      setState({
+        status: "error",
+        options,
+      });
+      return;
+    }
```
bvaughn, line 165: This reads weird, since catching a promise where it's defined (rather than awaited) won't do anything. Took me a minute to realize that you were catching here in case the cache had already resolved.

I think an easier-to-read way of accomplishing this same thing would be this though:
```suggestion
    boundingRectsCache.prefetch(replayClient, pauseId);
```

This method will start eagerly loading data, but suppress any thrown error.

bvaughn (approved):
I think this change looks okay (preferably with the `prefetch` modification I suggested) but do we know what caused this error? Should we (have we) reported it to RUN?

hbenl:
> Should we (have we) reported it to RUN?

I've created RUN-3322 for this.

PR #10370, opened 2024-02-27
Remove default roles for new members

ryanjduffy:
* Removes the default user roles specified by the FE so the backend can assign the defaults.
* Changes `selectRole` to toggle the appropriate roles instead of setting them so we don't drop the new contributor role being added by the backend

This can be adjusted later if we want to be explicit about the default roles but removing it now will have no user impact and avoid some deploy gymnastics when we add the new contributor role.


bvaughn:
Seems like a reasonable change.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
```diff
@@ -21,12 +21,19 @@ type WorkspaceMemberProps = {
 
 const memberRoleLabels: Record<WorkspaceUserRole, string> = {
   admin: "Admin",
+  // Not presented in the UI yet
+  contributor: "Contributor",
```
bvaughn, line 25: What is this role going to be used for and how will it impact the UI?

I don't see an explanation of this role in SCS-1979 and semantically, I don't know what the difference between a "debugger" and a "contributor" would be.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
```diff
@@ -70,52 +80,64 @@ function WorkspaceMemberRoles({
...
+      case "viewer":
+        updatedRoles.delete("debugger");
+        break;
+      case "contributor":
+      case "debugger":
+        updatedRoles.add(role);
+        break;
+      case "admin":
+        updatedRoles.has(role) ? updatedRoles.delete(role) : updatedRoles.add(role);
+        break;
```
bvaughn, line 100: The way this is written is confusing to me.

Why does "viewer" _delete_ a role from the set, but "debugger" _adds_ one? And why does "admin" _add or remove_ based on the role's presence? 

I think you're trying _not_ to delete one role when adding another, but this looks like an overly complex way to do that. Particularly the "admin" option (which is a checkbox) seems like it should be passing in a true/false flag.

ryanjduffy: It is confusing. I'll do some refactoring to clarify what's happening.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
bvaughn: Ah, thanks for the link.

I think part of what bothers me here is that...these roles seem like they should either be hierarchical (e.g. viewer has read-only permissions, debugger has all of viewer's permissions _plus_ the ability to upload, admin has all of those permissions _plus_ the ability to manage the workspace) _or_ they should be...behaviors that you can mix and match.

Our UI seems to be kind of treating them like both (select implies hierarchy and checkbox implies behavior attributes). Why is it like that?

ryanjduffy: The backend implements these as behaviors but the current UX treats them as a hierarchy (except for Admin which is a behavior in both). This is in part due to how we've chosen to do billing for bug reporting teams (only paying for debugger seats). I've talked with Jon about design options that better reflect the options available so I expect this will UI change in the future but for now it is trying to bridge that gap.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
ryanjduffy: I pushed a refactoring commit with separate handlers for each option with comments. Hopefully that adds a bit of clarity.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
bvaughn: I see. Ideally, I think we should show them in a consistent way (sounds like behaviors) rather than a mix, but as this is long-standing UX, I wouldn't block this PR on it.

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
```diff
@@ -70,52 +80,73 @@ function WorkspaceMemberRoles({
...
+      .finally(() => {
+        setLoading(false);
+        onClick();
+      });
+  };
+
+  const selectViewerRole = (selected: boolean) => {
+    // when switching to viewer role, we remove debugger
+    if (selected) {
+      updateRoles(roles.filter(r => r !== "debugger"));
```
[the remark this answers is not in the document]
ryanjduffy: Correct because the user should always have the viewer role. I can add a safety valve to ensure it's here though.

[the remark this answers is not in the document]
bvaughn: I see. It seems like we're representing our (actual) model in a weird way with this UI. I wonder if it would be better if we did something like...
Screenshot 2024-03-05 at 3 42 16 PM

That being said, I realize that's a more significant change and I won't block this PR on it :)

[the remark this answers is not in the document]
ryanjduffy: Pushed a commit that ensures viewer is there when that role is selected

bvaughn (approved):

src/ui/components/shared/WorkspaceSettingsModal/WorkspaceMember.tsx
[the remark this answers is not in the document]
ryanjduffy: I didn't have your last two comments yet when I last responded so I'll add: Yes! We should improve this. I don't know if there's a DES ticket for this yet but I'll make one and Jon can cancel it if it's a dupe.

PR #10375, opened 2024-02-28
Fix loading request and response bodies

hbenl:
- don't try to load a request or response body if we know it doesn't exist - this was causing protocol errors
- update the streaming value if loading a body succeeds without any events (i.e. we received an empty body) so that the network panel stops showing the body as loading

We could show an error message in the second case but I'm not sure if an empty request/response body is necessarily an error.

STR:
- open one of our FE E2E test recordings
- select one of the requests for `https://vitals.vercel-insights.com/v1/vitals`
- without this PR you can see a failed request for the response body in the protocol viewer and selecting the "Request" tab shows an indefinite loader. With this PR there are no failed protocol requests and the "Request" tab shows an empty body

### DIFF
--- a/packages/replay-next/src/suspense/NetworkRequestsCache.ts
+++ b/packages/replay-next/src/suspense/NetworkRequestsCache.ts
@@ -193,6 +193,10 @@ export const networkRequestBodyCache = createStreamingCache<
       update(requestBodyData);
     });
 
+    if (requestBodyData.length === 0) {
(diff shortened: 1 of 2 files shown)


bvaughn (approved):
> We could show an error message in the second case but I'm not sure if an empty request/response body is necessarily an error.

Agreed, it is not necessarily an error. I've seen empty 200 responses before from APIs.

PR #10379, opened 2024-02-28
Show better user-facing messaging when we fail to parse preview syntax

bvaughn:
While I was working on #10378, I realized (a) that there are probably other syntaxes that our cheap parser doesn't support yet and (b) I don't think we're actually explicitly tracking them, so this PR does a few things:

- [x] Explicitly log failed evaluations to Sentry (so we can learn about other expressions we aren't handling)
- [x] Show better messaging for `SyntaxError` (since they're likely our fault)
- [x] Remove a weird one-off `style` attribute added to the `Popup` component back in #9719 

### Before
Screenshot 2024-02-28 at 11 14 55 AM

### After
Screenshot 2024-02-28 at 1 10 50 PM

cc @jonbell-lot23

### DIFF
--- a/packages/replay-next/components/Popup.tsx
+++ b/packages/replay-next/components/Popup.tsx
@@ -1,4 +1,4 @@
-import { MouseEvent, ReactNode, RefObject, useLayoutEffect, useRef } from "react";
+import { HTMLAttributes, MouseEvent, ReactNode, RefObject, useLayoutEffect, useRef } from "react";
 import { createPortal } from "react-dom";
 
 import styles from "./Popup.module.css";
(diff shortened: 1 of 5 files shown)


packages/replay-next/components/Popup.tsx
```diff
@@ -12,22 +12,19 @@ type Dismiss = () => void;
...
-  dataTestName = "Popup",
   dismiss,
   dismissOnMouseLeave = false,
   horizontalAlignment = "center",
   showTail = false,
-  style = "default",
   target,
-}: {
+  ...rest
+}: Omit<HTMLAttributes<HTMLDivElement>, "onClick"> & {
```
bvaughn, line 24: The changes to this component aren't necessary and I _could_ remove them if anyone objects. Just seemed nice to remove some unused things and simplify this component's interface.

PR #10380, opened 2024-02-28
Be precise about what text/element we select on the elements list.

klochek:
The scripts are inline to the html file, and reference `maindiv`, so just looking for `maindiv` will return the part of the script that contains that text, too.


bvaughn (approved):
NIce

PR #10382, opened 2024-02-29
Replace log point capsule <input> with Lexical plug-in

bvaughn:
This is a prerequisite for FE-2315, as it will allow us to use the strategy described in that issue to prevent any inputs in the log point panel from being selected. (I am not aware of any way to achieve the same with an `<input>` element.)

- [x] Add new `NumberEditor` Lexical component with min/max support (basically a drop-in replacement for `<input type="number">`
- [x] Replace the `<input>` in the log point panel with this new plug-in
- [x] Update e2e test helpers to account for the change from `HTMLInputElement` to Lexical editor
- [x] Revisit decision to disallow temporarily empty input value

### DIFF
--- a/packages/e2e-tests/helpers/lexical.ts
+++ b/packages/e2e-tests/helpers/lexical.ts
@@ -16,13 +16,13 @@ type InputAndTypeAheadOptions =
     };
 
 export async function clearText(page: Page, selector: string) {
-  await focus(page, selector);
-
   const input = page.locator(selector);
 
   // Timing awkwardness;
   // Make sure we clear all of the text (and not just most of it)
   while (((await input.textContent()) || "").trim() !== "") {
+    await focus(page, selector);
+
     await delay(100);
 
     await page.keyboard.press(`${getCommandKey()}+A`);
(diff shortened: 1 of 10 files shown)


packages/replay-next/src/hooks/useNoSelectContentEditable.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export function useNoSelectContentEditable({
+  editable,
+  ref,
+}: {
+  editable: boolean;
+  ref: MutableRefObject<HTMLElement | null>;
+}) {
+  useLayoutEffect(() => {
+    // TODO
```
markerikson, line 11: :hand: missing something here?

bvaughn: Haha, I didn't mean to `git add` this file. It was part of the follow up. Will delete it.

bvaughn:
Hmm, yeah, it doesn't allow you to set an empty/undefined value. I built it that way intentionally because select+delete+retype didn't seem like a common user action (it's more keystrokes than select+retype.

I can revisit that decision I guess.

hbenl:
> select+delete+retype didn't seem like a common user action

Agreed, but it also doesn't work without the select and delete+retype is a common user action. So I think we need to allow setting an empty value (or at least temporarily having an empty value, it could go back to the previous value on blur).

bvaughn:
> > select+delete+retype didn't seem like a common user action
> 
> Agreed, but it also doesn't work without the select and delete+retype is a common user action. So I think we need to allow setting an empty value (or at least temporarily having an empty value, it could go back to the previous value on blur).

Yeah. I'm convinced. Just sorting out how this affects the plug-in's min/max logic.

bvaughn:
Okay. Take it for another spin.

hbenl:
Works for me.

bvaughn:
Great, thanks for the review and feedback. I'm going to wait until #10384 is landable before merging this one.

bvaughn:
Merged as part of #10384

PR #10384, opened 2024-02-29
Fix log point panel `user-select` quirk

bvaughn:
Builds on top of #10382 and fixes `user-select` quirk described in FE-2315

### [This Loom explains how this change works and shows a demo](https://www.loom.com/share/15a0548ec99f4d22bf2e36c69c682239)

- [x] Implement approach described in FE-2315
- [x] Make sure e2e tests pass (I think the _condition_ input may be failing?)
- [x] Verify that this supports _disabled_ and _auto-focus_ inputs
- [x] Verify that this works for @hbenl and @markerikson on Linux and Windows

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
+++ b/packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
@@ -267,7 +267,10 @@ function PointPanelWithHitPoints({
     } else {
       if (!isEditing) {
         startEditing("condition");
+      } else {
+        setEditReason("condition");
       }
+
       editPendingPointText(key, { condition: "" });
     }
   };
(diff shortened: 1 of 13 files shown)


hbenl (approved):
Works on Linux in both Firefox and Chromium.

packages/replay-next/components/sources/log-point-panel/LogPointPanel.tsx
```diff
@@ -267,7 +267,10 @@ function PointPanelWithHitPoints({
     } else {
       if (!isEditing) {
         startEditing("condition");
+      } else {
+        setEditReason("condition");
```
bvaughn, line 271: This was a pre-existing bug that I noticed while testing.

markerikson:
Yep, looks good on Windows!

markerikson (approved):

PR #10385, opened 2024-02-29
Switch stepping-07 to Chromium example

hbenl:


bvaughn (approved):
Feels a little weird to update tests to verify broken behavior but I guess it's a net win to migrate off of Gecko

PR #10387, opened 2024-02-29
Delete Firefox version of doc_rr_basic and switch remaining tests to the Chromium version

hbenl:


bvaughn (commented):
I would expect to see two additional changes in this PR:
* Delete `doc_rr_basic_chromium.html`
* Re-record example with `doc_rr_basic.html` using Chromium (as some tests reference the name of their example file for things)

bvaughn (approved):

PR #10391, opened 2024-03-01
Remove deprecated backend settings

jcmorrow:
This is a follow-up to https://github.com/replayio/backend/pull/9622. I'll be removing these settings altogether shortly so we should also remove them from the UI. Some of them *already* don't do anything, for example we don't turn off snapshots even if the user requests it.


markerikson (approved):

jcmorrow:
> Hmm. That `playwright-04` failure is surprising and might be relevant.

Huh. Dunno, I guess I'll take a look next week.

hbenl:
- in all the `playwright-04` failures the test timed out because `Debugger.findSources` was taking more than 20 seconds
- that test now also fails on other PRs
- but when I open the example recording, `Debugger.findSources` only takes about a second - maybe this is a new issue with cloning recordings

jcmorrow:
> * in all the `playwright-04` failures the test timed out because `Debugger.findSources` was taking more than 20 seconds
> * that test now also fails on other PRs
> * but when I open the example recording, `Debugger.findSources` only takes about a second - maybe this is a new issue with cloning recordings

If you open the recording it was cloned from that should fix the problem

jcmorrow:
Bvaughn has already PRed this, my bad!

bvaughn:
No worries. This PR includes another setting that I didn't know you wanted to remove, so we can just merge this instead. I don't care :)

PR #10393, opened 2024-03-04
Replace declarative Video component with imperative code

bvaughn:
Using a suspense cache for this was a mistake. Renders need to happen too quickly, and mixing rapid updates with suspense leads to starvation problems (renders that don't finish and update the DOM). I rewrote this to use imperative logic, which is how it was before (in a way) though I think this code is more consolidated.

Please give extra thought to the two small changes I made in `actions/timeline` (I left comments) as well as the removal of the `precachedPercent` Redux state. I think these changes are sound but I would like the reviewer to give them consideration as well.


src/ui/components/Timeline/ProgressBars.tsx
```diff
@@ -1,35 +1,27 @@
...
   const percent = getVisiblePosition({ time: currentTime, zoom: zoomRegion }) * 100;
   const hoverPercent = getVisiblePosition({ time: hoverTime, zoom: zoomRegion }) * 100;
-  const precachedPercent = getVisiblePosition({ time: precachedTime, zoom: zoomRegion }) * 100;
 
   return (
     <>
       <div className="progress-line full" />
       <div
         className="progress-line preview-max"
-        style={{ width: `${clamp(Math.max(hoverPercent, precachedPercent), 0, 100)}%` }}
```
bvaughn, line 28: I'm not sure what this "pre-cached percent" state was. I don't _think_ we really need it. Particularly since it's only used for our internal expanded/enhanced timeline UI. So I removed it to simplify things a bit.

src/ui/actions/timeline.ts
```diff
@@ -262,6 +262,11 @@ export function seek({
   location?: Location;
 }): UIThunkAction<Promise<void>> {
   return async (dispatch, getState, { replayClient }) => {
+    const isPlaying = getPlayback(getState()) !== null;
+    if (isPlaying) {
+      dispatch(stopPlayback());
+    }
```
bvaughn, line 268: I think it is a reasonable change. Any objections?

src/ui/actions/timeline.ts
```diff
@@ -412,15 +416,15 @@ export function startPlayback(
 
 export function stopPlayback(updateTime: boolean = true): UIThunkAction {
   return async (dispatch, getState) => {
+    dispatch(setTimelineState({ playback: null }));
```
bvaughn, line 419: This goes along with the other change, to prevent a potential loop.

packages/protocol/PaintsCache.ts
```diff
@@ -41,17 +41,16 @@ export async function findFirstMeaningfulPaint() {
...
-          paint.point
-        );
-        if (value && value.hash) {
-          const { width, height } = await getDimensions(value.hash, value.mimeType);
+        const screenShot = paint.paintHash
+          ? await screenshotCache.readAsync(replayClient, paint.point, paint.paintHash)
+          : undefined;
+
+        if (screenShot && screenShot.hash) {
+          const { width, height } = await getDimensions(screenShot.hash, screenShot.mimeType);
```
hbenl, line 49: This needs to use `screenShot.data`.
```suggestion
          const { width, height } = await getDimensions(screenShot.data, screenShot.mimeType);
```

src/ui/components/Video/imperative/updateGraphics.ts
```diff
@@ -0,0 +1,156 @@
...
+  replayClient: ReplayClientInterface;
+}) {
+  updateState(containerElement, {
+    executionPoint,
+    status: "loading",
+    time,
+  });
+
+  await PaintsCache.readAsync();
+
```
hbenl, line 30: I think we should check `abortSignal` here as well.

hbenl (approved):

src/ui/components/Video/imperative/updateGraphics.ts
bvaughn: Ah, yes. You're right.

packages/protocol/PaintsCache.ts
bvaughn: Ooof, yes.

PR #10395, opened 2024-03-04
Fix support for team settings urls in parser

ryanjduffy:
The query params parser was missing support for `/team/<id>/settings` to support redirect to the settings modal causing the "Update Subscription" button to fail.

### DIFF
--- a/src/ui/components/Library/Team/utils.ts
+++ b/src/ui/components/Library/Team/utils.ts
@@ -12,7 +12,9 @@ export function parseQueryParams(query: ParsedUrlQuery) {
   let testRunId: string | undefined;
   let testId: string | undefined;
 
-  assert(!view || view === "runs" || view === "tests" || view === "recordings");
+  assert(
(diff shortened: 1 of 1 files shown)


bvaughn (approved):

src/ui/components/Library/Team/utils.ts
```diff
@@ -12,7 +12,9 @@ export function parseQueryParams(query: ParsedUrlQuery) {
   let testRunId: string | undefined;
   let testId: string | undefined;
 
-  assert(!view || view === "runs" || view === "tests" || view === "recordings");
+  assert(
+    !view || view === "runs" || view === "tests" || view === "recordings" || view === "settings"
```
bvaughn, line 16: Are these values enumerated somewhere?

ryanjduffy: Not really. I found a `View` type in `src/ui/components/Library/Team/View/ViewContextRoot.tsx` but it doesn't include `settings` because it isn't used the same as the others.

PR #10397, opened 2024-03-05
Only render the React details panel when the wall is updated to the current pauseId

hbenl:
This is a possible fix for FE-2325, it's not the cleanest one but it's simple.


bvaughn (commented):
I wonder...if it would be "safer" to make the wall require a pause id function parameter, then it could always wait until the pause ids matched? (This relies on anything that talks to the wall knowing to _wait_ first)

hbenl:
> I wonder...if it would be "safer" to make the wall require a pause id function parameter, then it could always wait until the pause ids matched?

Yeah, that would be safer (and relatively simple as well). I'll try to do that.

hbenl:
Superseded by #10399

PR #10398, opened 2024-03-06
.github/workflows: switch workflows to use blacksmith

adityamaru:
This change switches replayio's workflows to use blacksmith  (https://blacksmith.sh). It also changes some cache actions to use our collocated caches for much faster read/write times.

### DIFF
--- a/.github/workflows/bvaughn-jest.yml
+++ b/.github/workflows/bvaughn-jest.yml
@@ -14,7 +14,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: blacksmith
     steps:
     - name: Check out repository code
       uses: actions/checkout@v4
@@ -23,7 +23,7 @@ jobs:
       id: yarn-cache-dir-path
       run: echo "{dir}={$(yarn config get cacheFolder)}" >> $GITHUB_OUTPUT
     - name: Restore yarn cache
-      uses: actions/cache@v3
+      uses: useblacksmith/cache@v5
       id: yarn-cache
       with:
         path: |
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   clean:
-    runs-on: ubuntu-latest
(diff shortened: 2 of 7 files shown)


bvaughn:
Ah, I see! I'll re-open until @jasonLaster can weigh in then.

jasonLaster:
Here's the comment in twitter

> Hey Jason, here's the PR - https://github.com/replayio/devtools/pull/10398. For our runners to pickup your jobs you'll have to:
- install our GitHub app
- enable self hosted runners for a public repo
- retry the first set of jobs on this PR once the above steps are done

This is all documented in https://docs.blacksmith.sh/getting-started/quickstart. Let me know if I can help!

jasonLaster:
Thanks @adityamaru - just kicked off the run. I'm curious to see how this will help us with our Playwright performance. Each job has some level of playwright concurrency which could help quite a bit. 

Here's our first error. 

> ✖ Failed to post annotations to github: GitHub Error:403 Body:{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/checks/runs#create-a-check-run"}

adityamaru:
I think you'd have to change the workflow trigger from `pull_request` to `pull_request_target` to allow for PRs from a fork to have access to secrets - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target. This SO article might interest you to weight whether you actually want to do this or not - https://stackoverflow.com/a/71366152

For private repos GitHub seems to have this configuration option under Settings -> Actions -> General that might allow for PRs from forks to run actions:
Screenshot 2024-03-06 at 8 58 16 PM

bvaughn:
I don't think we want to make that change, but I can pull down this branch and push it to our main repo to let CI run.

bvaughn:
@adityamaru Would you mind adding a GitHub _suggestion_ to PR #10408 with the change you're describing?

adityamaru:
done! thank you for working with us on this, we'll also pull down a few more Playwright repos to see why we're not seeing the same level of speedup as some other workflows

bvaughn:
We can close this one in favor of #10408

PR #10399, opened 2024-03-06
Add ReplayWall.sendAtPauseId()

hbenl:
- `ReplayWall.sendAtPauseId()` should be used instead of `ReplayWall.send()` to ensure that the wall has been updated to the desired `pauseId`
- `useHighlightNativeElement` still uses `ReplayWall.send()` because it first gets an ID from the `store` which it then asks to be highlighted, so we don't want to wait for the store to be potentially updated
- I didn't bother updating `ReplayWall.getComponentLocation()` (which calls the private `ReplayWall.sendRequest()`) because it is only used by the legacy RDT panel

### DIFF
--- a/src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
+++ b/src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
@@ -1,4 +1,4 @@
-import { ObjectId } from "@replayio/protocol";
+import { ObjectId, PauseId } from "@replayio/protocol";
 import type { SerializedElement, Store, Wall } from "@replayio/react-devtools-inline/frontend";
 import React from "react";
 
@@ -24,6 +24,11 @@ export type StoreWithInternals = Store & {
   _idToElement: Map<number, ElementWithChildren>;
 };
 
+interface Waiter {
+  promise: Promise<void>;
+  resolve: () => void;
+}
+
 export class ReplayWall implements Wall {
   private disableNodePicker: NodePickerContextType["disable"];
   private dismissInspectComponentNag: () => void;
@@ -34,6 +39,8 @@ export class ReplayWall implements Wall {
   private replayClient: ReplayClientInterface;
   private setProtocolCheckFailed: (failed: boolean) => void;
   private unhighlightNode: () => void;
+
+  private pauseIdWaiters = new Map<PauseId, Waiter>();
 
   // HACK The Wall requires the Store and the Store requires the Wall
   // So this value is set by the caller after initialization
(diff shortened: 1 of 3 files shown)


bvaughn:
Huh... you're still seeing `playwright-04` failures on this PR

hbenl:
> still seeing playwright-04 failures on this PR

Yeah, I think the one PR that was green was a fluke - Josh didn't deploy his fix yet.

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
```diff
@@ -67,6 +74,25 @@ export class ReplayWall implements Wall {
...
+    const waiter = this.pauseIdWaiters.get(pauseId);
+    if (waiter) {
+      await waiter.promise;
+      return;
+    }
+
+    let resolve!: () => void;
+    const promise = new Promise<void>(_resolve => (resolve = _resolve));
+    this.pauseIdWaiters.set(pauseId, { promise, resolve });
+    await promise;
```
bvaughn, line 95: I think what you have is good, but [`Promise.withResolvers()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers) is also a nifty (more recent) API for doing this sort of thing:

```js
const { promise, resolve } = Promise.withResolvers();
```

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
```diff
@@ -96,6 +122,11 @@ export class ReplayWall implements Wall {
     this.listener?.(message);
   }
 
+  async sendAtPauseId(event: string, payload: any, pauseId: PauseId) {
+    await this.waitForPauseId(pauseId);
+    return await this.send(event, payload);
+  }
```
bvaughn, line 128: I guess it would be difficult to ensure that this method is called for sensitive events (at least "inspectElement") but we should probably add a comment above this and `send` below cautioning that it should be used for anything that has pause-specific data (aka `"inspectElement"`)

bvaughn (approved):
Should we also do this for `"highlightNativeElement"`?

bvaughn:
> Yeah, I think the one PR that was green was a fluke - Josh didn't deploy his fix yet.

Ha! That's...oh gosh. That's funny

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
hbenl: TIL, nice!

hbenl: But it's too new, we'd need to add a polyfill for that...

bvaughn: Yeah, agreed. Though the polyfill would be very small/simple :) I was really glad to see that get added because it's so common and awkward an operation otherwise.

PR #10403, opened 2024-03-06
[RUN-3224] Remove `authenticated/comments-03` from blacklist

Domiii:
I think someone paper'ed over the missing favicon for now, so we can re-enable the test.
* https://linear.app/replay/issue/RUN-3224/re-enable-authenticatedcomments-03


Domiii:
Passed!

Domiii:
Oh shoot... Tagging @hbenl because you asked to at least always be tagged!
Let me know if I should back it out again!

PR #10405, opened 2024-03-06
Don't reset page title after a root-level error

bvaughn:
[Loom demo and explanation](https://www.loom.com/share/2f6e61f8d5c843b2ba678d2de5651cf4)

I am open to other ways of doing this. I don't really like this approach, nor the way that our some of our code is currently organized so that it can't reference other parts.

I am not sure how I feel about this PR to be honest. Posting it for discussion purposes. An alternative route would be to remove this default title from the parent layout:

https://github.com/replayio/devtools/blob/ec3b9e7cddb555368b55c2b11f31e9ecbe96d762/pages/_app.tsx#L106-L112

Although we would have to be sure to manually add the title to individual pages (like Library). I think I might prefer _that_ fix.

Note to test this change I did...
```diff
diff --git a/packages/protocol/socket.ts b/packages/protocol/socket.ts
index 540701f38..e08f5fc00 100644
--- a/packages/protocol/socket.ts
+++ b/packages/protocol/socket.ts
@@ -143,7 +143,11 @@ export async function createSession(
 ) {
   const { sessionId } = await sendMessage("Recording.createSession", {
     recordingId,
-    experimentalSettings,
+    experimentalSettings: {
+      ...experimentalSettings,
+      inactivityTerminationTimeout: 0.2,
+      inactivityNotificationTimeout: 0.1,
+    },
     focusRequest: focusWindow,
   });
 
```

### DIFF
--- a/src/ui/components/Errors/RootErrorDocumentTitle.tsx
+++ b/src/ui/components/Errors/RootErrorDocumentTitle.tsx
@@ -1,0 +1,14 @@
+import Head from "next/head";
+
+import { useGetRecordingTitle } from "shared/graphql/Recording";
+import { getRecordingId } from "shared/utils/recording";
+
+export function RootErrorDocumentTitle() {
+  const { title = "Replay" } = useGetRecordingTitle(getRecordingId());
+
+  return (
+    <Head>
+      <title>{title}</title>
+    </Head>
+  );
+}
--- a/packages/shared/graphql/Recording.ts
+++ b/packages/shared/graphql/Recording.ts
@@ -1,0 +1,40 @@
+import { gql, useQuery } from "@apollo/client";
+import { RecordingId } from "@replayio/protocol";
+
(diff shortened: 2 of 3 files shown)


src/ui/components/Errors/RootErrorDocumentTitle.tsx
```diff
@@ -0,0 +1,14 @@
+import Head from "next/head";
+
+import { useGetRecordingTitle } from "shared/graphql/Recording";
+import { getRecordingId } from "shared/utils/recording";
+
+export function RootErrorDocumentTitle() {
+  const { title = "Replay" } = useGetRecordingTitle(getRecordingId());
```
bvaughn, line 7: One flaw of this is, I guess, if GraphQL goes down, our error boundary will also crash.

bvaughn:
Closing in favor of #10406

PR #10407, opened 2024-03-06
Redirect mobile clients to warning page

bvaughn:
cc @jonbell-lot23 if you want to tweak the UI of this page

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -1,0 +1,25 @@
+import type { NextRequest } from "next/server";
+import { NextResponse, userAgent } from "next/server";
+
+export function middleware(request: NextRequest) {
+  const { ua } = userAgent(request);
+
+  if (isMobile(ua)) {
+    // If the user is attempting to visit Replay on a mobile device for the first time,
+    // show them a message that it has not been optimized for mobile
+    // If they have already confirmed this message (detectable via a cookie) then let them through
+    const cookie = request.cookies.get("mobile-warning-dismissed");
+    if (cookie == null && request.nextUrl.pathname !== "/mobile-warning") {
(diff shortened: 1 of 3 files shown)


jonbell-lot23 (approved):

jonbell-lot23:
I notice that this screen appears after the login page, rather than before. It's better to make it before. Failing fast is better than "Please go through the log in process, oops, actually the site won't look right, but we didn't tell you first."

bvaughn:
> I notice that this screen appears after the login page, rather than before. It's better to make it before. Failing fast is better than "Please go through the log in process, oops, actually the site won't look right, but we didn't tell you first."

That's a good catch. I am not sure how our auth is configured. I'll have to look into it and see if I can override it per-route like this.

middleware.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function middleware(request: NextRequest) {
+  const { ua } = userAgent(request);
+
+  if (isMobile(ua)) {
+    // If the user is attempting to visit Replay on a mobile device for the first time,
+    // show them a message that it has not been optimized for mobile
+    // If they have already confirmed this message (detectable via a cookie) then let them through
+    const cookie = request.cookies.get("mobile-warning-dismissed");
+    if (cookie == null && request.nextUrl.pathname !== "/mobile-warning") {
```
bvaughn, line 12: The `pathname` check is probably unnecessary, since the middleware doesn't respond to that route.

hbenl (approved):

bvaughn:
I see an e2e test failing on this branch that I am pretty sure is unrelated to this change. I filed FE-2346 to follow up.

PR #10408, opened 2024-03-07
[draft] .github/workflows: switch workflows to use blacksmith

jasonLaster:
Creating a new branch which will have env vars. the source of truth is still https://github.com/replayio/devtools/pull/10398

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,11 +73,11 @@ jobs:
         shard: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]
     timeout-minutes: 20
     steps:
-        - uses: alexellis/setup-arkade@master
-        - name: Install vmmeter
-           run: |
-            sudo -E arkade oci install ghcr.io/openfaasltd/vmmeter:latest --path /usr/local/bin/
-        - uses: self-actuated/vmmeter-action@master
+      - uses: alexellis/setup-arkade@master
+      - name: Install vmmeter
+         run: |
+          sudo -E arkade oci install ghcr.io/openfaasltd/vmmeter:latest --path /usr/local/bin/
+      - uses: self-actuated/vmmeter-action@master
       - uses: actions/checkout@v4
       - name: Setup node
         uses: useblacksmith/setup-node@v5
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -75,15 +75,15 @@ jobs:
     steps:
```
adityamaru, line 75:
```suggestion
    steps:
        - uses: alexellis/setup-arkade@master
        - name: Install vmmeter
           run: |
            sudo -E arkade oci install ghcr.io/openfaasltd/vmmeter:latest --path /usr/local/bin/
        - uses: self-actuated/vmmeter-action@master
```

bvaughn: Seems like the e2e tests didn't run at all after this change?

adityamaru:
Screenshot 2024-03-07 at 2 28 54 PM

I think it might be an indentation issue in the yaml file. The run needs a backspace I believe

bvaughn:
I see. I just committed the suggestion directly. I guess I should have looked closer.

.github/workflows/test.yml
```diff
@@ -73,11 +73,11 @@ jobs:
...
     timeout-minutes: 20
     steps:
-        - uses: alexellis/setup-arkade@master
-        - name: Install vmmeter
-           run: |
-            sudo -E arkade oci install ghcr.io/openfaasltd/vmmeter:latest --path /usr/local/bin/
-        - uses: self-actuated/vmmeter-action@master
+      - uses: alexellis/setup-arkade@master
+      - name: Install vmmeter
+         run: |
```
adityamaru, line 78: I think you'll need to un-indent the run to be in the same line as name. Apologies, the suggestion seems to have messed this up.

bvaughn: Ah. Whitespace sensitive structures are irritating :)

bvaughn: Apologies. That's what I get for using GitHub's in-browser editor for this.

adityamaru:
I took a look at the vmmeter output and while the 1min CPU peak for some of them was pretty high, overall it does not seem that throwing more CPUs at these tests will speed them up. We're going to pull down some Playwright repos and try using https://playwright.dev/docs/trace-viewer-intro#recording-a-trace to understand why our faster single core performance/caching is not having the same speedup we see with other build systems. We don't have a good way to measure off-cpu time in these workflows as well.

All to say that I'm happy to come back to y'all once I better understand some of these things, and we can close this for now since the perf is comparable to GitHub. I appreciate you trying us out!

PR #10409, opened 2024-03-07
Don't try to load computed properties if no element is selected

hbenl:
This was causing "Bad parameter type" protocol errors because we were trying to load the preview for objectId `null`.


src/devtools/client/inspector/computed/components/ComputedProperties.tsx
```diff
@@ -49,8 +72,8 @@ function ComputedProperties() {
   const { value: element, status: elementStatus } = useImperativeCacheValue(
     elementCache,
     replayClient,
-    pauseId!,
-    selectedNodeId!
```
bvaughn, line 53: "!" bites again

bvaughn (approved):

PR #10410, opened 2024-03-07
Show picker loading state until bounding rects are loaded

hbenl:
`NodePickerContext` set its state from `"initializing"` to `"active"` as soon as `await initializer()` finished. The React component picker didn't load the bounding rects in the initializer, so the picker went into `"active"` state before they finished loading.
Now `NodePickerContext` will wait for `boundingRectsCache.readAsync()` in parallel to `initializer()`.

### DIFF
--- a/src/ui/components/NodePickerContext.tsx
+++ b/src/ui/components/NodePickerContext.tsx
@@ -28,7 +28,7 @@ export type NodePickerOptions = {
 };
 
 export type NodePickerContextType = {
-  enable: (options: NodePickerOptions, initializer: () => Promise<void>) => Promise<void>;
+  enable: (options: NodePickerOptions, initializer?: () => Promise<void>) => Promise<void>;
   disable: () => void;
   status: NodePickerStatus;
   type: NodePickerType | null;
@@ -68,7 +68,11 @@ export function NodePickerContextRoot({ children }: PropsWithChildren<{}>) {
   });
 
   const enable = useCallback(
-    async (options: NodePickerOptions, initializer: () => Promise<void>) => {
+    async (options: NodePickerOptions, initializer?: () => Promise<void>) => {
+      if (!pauseId) {
+        return;
+      }
+
       const { status } = committedStateRef.current;
 
       switch (status) {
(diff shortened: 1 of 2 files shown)


src/ui/components/NodePickerContext.tsx
```diff
@@ -28,7 +28,7 @@ export type NodePickerOptions = {
 };
 
 export type NodePickerContextType = {
-  enable: (options: NodePickerOptions, initializer: () => Promise<void>) => Promise<void>;
+  enable: (options: NodePickerOptions, initializer?: () => Promise<void>) => Promise<void>;
```
bvaughn, line 31: Is there any value in keeping the initializer param? The only place calling this function has been updated so as not to pass it.

src/ui/components/NodePickerContext.tsx
```diff
@@ -152,9 +160,6 @@ export function NodePickerContextRoot({ children }: PropsWithChildren<{}>) {
       return;
     }
 
-    // Start loading rects eagerly; we'll need them in the hover/click handlers
-    boundingRectsCache.prefetch(replayClient, pauseId);
```
bvaughn, line 156: Just curious, why'd you remove this?

src/ui/components/NodePickerContext.tsx
```diff
@@ -68,7 +68,11 @@ export function NodePickerContextRoot({ children }: PropsWithChildren<{}>) {
   });
 
   const enable = useCallback(
-    async (options: NodePickerOptions, initializer: () => Promise<void>) => {
+    async (options: NodePickerOptions, initializer?: () => Promise<void>) => {
+      if (!pauseId) {
+        return;
+      }
```
bvaughn, line 74: I think this is the only substantial change?

bvaughn (commented):
I think I'm missing something. This change looks like you just moved `await boundingRectsCache.readAsync` from a callback into the context, but you're still calling it at the same time and waiting on the exact same thing to happen before updating the state.

bvaughn (approved):

PR #10413, opened 2024-03-07
Redirect mobile clients to warning page

bvaughn:
Resolves FE-2342 and FE-2346


bvaughn:
Not sure why the GitHub bot comment says we have a test failure but the Test Suites page shows success so...
Screenshot 2024-03-07 at 3 38 13 PM

PR #10416, opened 2024-03-08
Don't override null fallback in InlineErrorBoundary

hbenl:
In FE-2292 / BAC-4627, the default `<InlineErrorFallback>` was shown for a failing component even though [its error boundary](https://github.com/replayio/devtools/blob/fc64bb6774c32427f62badeeb3331e394760f229/src/ui/components/Timeline/PreviewMarkers.tsx#L33) has a `fallback={null}` prop.
`<InlineErrorBoundary>` should only show the `<InlineErrorFallback>` if its `fallback` prop is `undefined`.


bvaughn (approved):
Good catch.

PR #10418, opened 2024-03-08
Remove cloning and user-reset from e2e tests

bvaughn:
Resolves FE-2347 and BAC-4754.

This was originally added in #9497 as a way to avoid parallel tests from interfering with each other but has since proved to be difficult to maintain on the backend side. Our tests will now use unique "scopes" instead to isolate comments, points, and nags between runs.

I've tested this locally by...
- [x] Running tests in parallel, multiple times
- [x] Killing tests in the middle (leaving stranded comments) and then re-running them


packages/e2e-tests/testFixture.ts
```diff
@@ -23,30 +22,16 @@ const testWithCloneRecording = base.extend<TestIsolatedRecordingFixture>({
   exampleKey: undefined,
   testUsers: undefined,
   pageWithMeta: async ({ page, exampleKey, testUsers }, use) => {
-    const exampleRecordings = require("./examples.json");
```
bvaughn, line 26: Why were we using `require` here before?

Using `import` gives up TSC hints for the JSON structure.

hbenl:
I think we may want to keep using `resetTestUser` because otherwise the test user's nags array fills up with scoped nags.

bvaughn:
> I think we may want to keep using resetTestUser because otherwise the test user's nags array fills up with scoped nags.

cc @jcmorrow about this. If that's the case, then supporting scopes for nags seems kind of pointless. (I vaguely recall us talking about some periodic cleanup of scoped nags but maybe I'm remembering incorrectly?)

bvaughn:
> It isn't, the scoped nags from one test will be invisible in other tests. 

But we already kind of had that ability with the URL param that overrode nags, didn't we? After all, we've been testing passport functionality for months (before the scopes stuff existed)

bvaughn:
I just read the PR description for https://github.com/replayio/backend/pull/9399 and it looks like you're right, and Josh intended for us to continue using the reset-test-user mutation. I guess that makes sense. I'll push an update and I guess we'll wait and see why the mutation has started failing now.

hbenl (approved):

bvaughn:
Quite a few e2e test failures but I don't think any are related to this change so..

PR #10426, opened 2024-03-11
Harden Timeline current time/pause state

bvaughn:
[Loom overview/demo](https://www.loom.com/share/f97b09cfb7834c36bdf993ea77cc8d4f?from_recorder=1&focus_title=1)

- [x] Fixed a logic error in `stopPlayback` that caused it to null out playback state _before_ checking if it was null (🤡)
  * I think this was the core of the bugfix
- [x] Move Timeline move-move and mouse-up handlers to the `body` element, to ensure that dragging past the timeline's boundaries updates the current time appropriately
  * This was making it difficult to scrub the playback time to the beginning of a recording, for example
- [x] Remove no-longer-used "dragging" state from Redux
- [x] Remove no-longer-used `updateTime` param to Redux `stopPlayback` action

Please give this a close look. Our timeline logic is kind of complicated, and I just want to be double checked. :)

### DIFF
--- a/src/ui/components/Timeline/Timeline.tsx
+++ b/src/ui/components/Timeline/Timeline.tsx
@@ -1,4 +1,4 @@
-import { MouseEvent, Suspense, useContext, useLayoutEffect, useRef, useState } from "react";
+import { Suspense, useContext, useEffect, useLayoutEffect, useRef, useState } from "react";
 
 import { FocusContext } from "replay-next/src/contexts/FocusContext";
 import { useGraphQLUserData } from "shared/user-data/GraphQL/useGraphQLUserData";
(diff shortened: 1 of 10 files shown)


src/ui/components/Timeline/Timeline.tsx
```diff
@@ -103,30 +116,42 @@ export default function Timeline() {
     }
   };
 
-  const onMouseUp = (event: MouseEvent) => {
```
bvaughn, line 106: This event handler was attached too low in the DOM before. If you moused-out and then released, this code would never execute.

src/ui/components/Video/imperative/subscribeToMutableSources.ts
```diff
@@ -54,37 +54,45 @@ export function subscribeToMutableSources({
...
-        abortController.abort();
-        abortController = null;
+    } else {
+      if (didStopPlaying) {
+        if (abortController != null) {
+          abortController.abort();
+          abortController = null;
+        }
+
+        // TODO [FE-2353] Pause near wherever playback stopped
```
bvaughn: This was just a whitespace change mostly but I think it reads better.

jasonLaster:
@bvaughn are there any e2e tests we could write for the timeline / video code?

bvaughn:
There are several e2e tests already for this. In general, writing e2e tests that simulate drag operations is difficult in my experience.

jasonLaster:
Totally get it. Just curious. And yes - i like repaint-05

hbenl (approved):
LGTM

PR #10428, opened 2024-03-12
Update TypeScript dep from 5.3.3 to 5.4.2 (latest)

Andarist:
There is no immediate need to land this. I'm just looking around and poking things so I thought I might do it and solve any issues that could arise from it - but there were none.


bvaughn (approved):
No objections here.

PR #10429, opened 2024-03-12
Make it a bit easier to see what to expect when looking at repaints.

klochek:
It was very helpful to have this during my debugging process with repaints (so that you can just rewind to a console message and know exactly what the image should look like.)


bvaughn (approved):

bvaughn:
I'm going to look into FE-2357 and see if I can create a Replay of the bug. I'm going to beef up the paint tests as a result, which will include something along these lines– so I'm going to just close this PR as redundant. Hope that's okay.

PR #10432, opened 2024-03-13
Fix handling of Control key in KeyboardModifiersContext

hbenl:
`KeyboardModifiersContext` treats the Control key as another Meta key (which is useful on Windows where the Meta key always opens the Windows menu). But the `onMouseMove` event handler didn't do that so `isMetaKeyActive` was reset to `false` when the mouse was moved while the Control key is pressed.

### DIFF
--- a/packages/replay-next/src/contexts/KeyboardModifiersContext.tsx
+++ b/packages/replay-next/src/contexts/KeyboardModifiersContext.tsx
@@ -57,7 +57,7 @@ export function KeyboardModifiersContextRoot({ children }: PropsWithChildren<{}>
     };
 
     const onMouseMove = (event: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


bvaughn:
> `KeyboardModifiersContext` treats the Control key as another Meta 

Where? I don't see any reference to `ctrlKey` in that context. 

Seems like maybe there are a couple of places we don't treat both interchangeably, though I haven't looked at individual spots to see if these make sense.

| | |
| :--- | :--- |
| Screenshot 2024-03-13 at 10 23 37 AM | Screenshot 2024-03-13 at 10 23 44 AM |

hbenl:
> Where? I don't see any reference to ctrlKey in that context.

In the `onKeyDown`/`onKeyUp` handlers, where `event.key === "Control"` is interpreted as a meta key.

PR #10437, opened 2024-03-13
Revert "Add switch accounts interstitial to login flow"

ryanjduffy:
Reverts replayio/devtools#10431

Caused a regression when launching login from the Replay browser. I didn't see it in testing because the browser was always launching to app.replay.io for login so the new local code wasn't triggered.

https://www.loom.com/share/001be5bac6af48cd957e32948feb9f83?sid=6940f0ab-9937-4058-8254-d5ff27151562


bvaughn:
I don't really understand the implications of what you're doing here so I'm just kind of rubber stamping the sign-in-flow PRs.

bvaughn (approved):

PR #10439, opened 2024-03-14
Revert "Update TypeScript dep from 5.3.3 to 5.4.2 (latest) (#10428)"

Andarist:
This reverts commit f5dc68b73fce8e16d6a640d9d8704b4866934978.


bvaughn (approved):
Approved! (Generally you don't need an approval for a revert like this, FWIW)

PR #10447, opened 2024-03-15
Dont swallow error when recording e2e tests

Andarist:
An extra safe-guard against failures here. This should stop the whole script from executing. it would potentially help out people investigating issues like the ones araised from https://github.com/replayio/devtools/pull/10428 . I tested this out locally to confirm that this would fail the script.

### DIFF
--- a/packages/e2e-tests/scripts/record-playwright.ts
+++ b/packages/e2e-tests/scripts/record-playwright.ts
@@ -41,9 +41,7 @@ export async function recordPlaywright(
   });
   const page = await context.newPage();
   try {
-    return await script(page as any, expect);
-  } catch (err) {
-    console.error("PLAYWRIGHT ERROR", err);
+    return await script(page, expect);
   } finally {
     await page.close();
     await context.close();


packages/e2e-tests/scripts/record-playwright.ts
```diff
@@ -41,9 +41,7 @@ export async function recordPlaywright(
   });
   const page = await context.newPage();
   try {
-    return await script(page as any, expect);
```
Andarist, line 44: for context - there are 2 timeouts at play here, 30s timeout in `defaultPlaywrightScript` (within `waitUntilMessage`) and the 60s timeout based on the `CONFIG.recordingTimeout` value. When the latter happens it already throws nicely but when the former one gets thrown... this was caught here and swallowed

bvaughn (approved):

packages/e2e-tests/scripts/record-playwright.ts
bvaughn: I _think_ the original idea here was to allow some tests to re-record and upload even if others failed (but the fact that the script (a) never re-threw an error later and (b) actually continued to upload a potentially failed recording was just _broken__

PR #10449, opened 2024-03-18
Improve paint/repaint behavior

bvaughn:
Makes several small changes to graphics related code to improve interruption and better handle caching edge cases. I would appreciate a very close review of `updateGraphics`. It _looks_ like it could be simplified but I don't think it can without missing some cases.

I wrote a new, more low-level repaint test (`repaint-06`) but it currently fails in many places due to a suspected RUN issue which Klochek is looking into (RUN-3397). For now, I have _disabled_ that test and filed a follow up issue (FE-2363).

### DIFF
--- a/packages/e2e-tests/tests/repaint-06.test.ts
+++ b/packages/e2e-tests/tests/repaint-06.test.ts
@@ -7,18 +7,19 @@ import { expect, test } from "../testFixture";
 
 test.use({ exampleKey: "doc_control_flow.html" });
 
-test.skip("repaint-06: repaints the screen screen when stepping over code that modifies the DOM", async ({
+test("repaint-06: repaints the screen screen when stepping over code that modifies the DOM", async ({
   pageWithMeta: { page, recordingId },
   exampleKey,
 }) => {
   await startTest(page, recordingId);
   await openDevToolsTab(page);
 
+  // TODO [FE-2363] Some of these points fail repaint, so they're disabled until RUN-3397 has been fixed
   const printedStrings = [
-    "catch",
-    "afterCatch",
+    // "catch",
+    // "afterCatch",
     "finally",
-    "yield 1",
+    // "yield 1",
     "generated 1",
     "yield 2",
     "generated 2",
@@ -29,17 +30,13 @@ test.skip("repaint-06: repaints the screen screen when stepping over code that m
     "within iterator",
   ];
 
-  const paints: {
-    [key: string]: {
-      before: string | null;
-      after: string | null;
-    };
+  const textToBase64DataMap: {
+    [text: string]: string | null;
   } = {};
 
   {
     await debugPrint(page, "Testing repaint graphics");
 
-    let graphicsDataUrl: string | null = null;
     let previousGraphicsDataUrl: string | null = null;
 
     for (let index = 0; index < printedStrings.length; index++) {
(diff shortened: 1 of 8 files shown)


packages/protocol/RepaintGraphicsCache.ts
```diff
@@ -11,6 +12,20 @@ export const RepaintGraphicsCache: Cache<
...
+    const result = await replayClient.repaintGraphics(pauseId);
+
+    let { description, screenShot } = result;
+
+    // The backend won't return a screenshot for a given hash more than once; see FE-2357
+    if (screenShot) {
+      paintHashCache.cacheValue(screenShot, description.hash);
+    } else {
+      screenShot = paintHashCache.getValueIfCached(description.hash);
+    }
```
bvaughn, line 24: [`DOM.repaintGraphics` API docs](https://static.replay.io/protocol/tot/DOM/#method-repaintGraphics) say this about the `screenShot` value:
> Contents of the screen shot. If an identical screen shot was previously returned, this will be omitted.

packages/replay-next/src/suspense/ScreenshotCache.ts
```diff
@@ -10,5 +10,18 @@ export const screenshotCache = createCache<
   config: { immutable: true },
   debugLabel: "ScreenshotCache",
   getKey: ([client, point, paintHash]) => paintHash,
-  load: ([client, point]) => client.getScreenshot(point),
+  load: async ([client, point]) => {
+    const screenShot = await client.getScreenshot(point);
+
+    paintHashCache.cacheValue(screenShot, screenShot.hash);
+
+    return screenShot;
```
bvaughn, line 18: I'm not sure if this is prone to the same case mentioned in the repaint cache. Seems like it could be, depending on how RUN keys screenshots (by point or by hash)

src/ui/components/Video/imperative/MutableGraphicsState.tsx
```diff
@@ -48,25 +50,33 @@ export const state = createState<State>({
...
     didResize?: boolean;
     executionPoint: ExecutionPoint | null;
     screenShot: ScreenShot | null;
+    screenShotType: ScreenShotType | null;
     status: Status;
     time: number;
   }> = {}
 ) {
+  const localLock = new Object();
+  lock = localLock;
```
bvaughn, line 71: There was a potential race here before when e.g. scrubbing the timeline.

src/ui/components/Video/imperative/updateGraphics.ts
```diff
@@ -31,27 +31,22 @@ export async function updateGraphics({
...
+  const isBeforeFirstCachedPaint = !paintPoint || !paintPoint.paintHash;
+  if (!isBeforeFirstCachedPaint) {
+    promises.push(
+      fetchPaintContents({
+        abortSignal,
+        containerElement,
+        replayClient,
+        time,
+      })
+    );
```
bvaughn, line 47: Even if we're before the first paint, we can still request a repaint.

jasonLaster:
Love seeing the new repaint test!

packages/e2e-tests/tests/repaint-06.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      await stepOver(page);
+      await stepOver(page);
+
+      await waitForRepaintGraphicsToLoad(page);
+
+      graphicsDataUrl = await getGraphicsDataUrl(page);
+      expect(
+        graphicsDataUrl,
+        `Expected graphics before "${string}" not to match previous after graphics`
+      ).not.toEqual(previousGraphicsDataUrl);
```
hbenl, line 62: I don't think this is correct: why should the screenshot change in between updates?

src/ui/components/Video/imperative/MutableGraphicsState.tsx
```diff
@@ -86,6 +100,10 @@ export async function updateState(
       imageHeight: naturalHeight,
       imageWidth: naturalWidth,
     });
+    if (lock !== localLock) {
+      return;
```
hbenl, line 104: Nit: `fitImageToContainer()` isn't async so we can remove the `await` and this check.

hbenl (approved):

packages/e2e-tests/tests/repaint-06.test.ts
bvaughn: Yeah, you're right about that.

PR #10457, opened 2024-03-21
Ensure the test step details panel doesn't get stuck in loading state

hbenl:
See FE-2377


src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -125,6 +125,22 @@ export const testEventDetailsIntervalCache = createFocusIntervalCacheForExecutio
       testEventDetailsResultsCache.cacheValue(processedResult, processedResult.point);
     }
 
+    for (const event of filteredEvents) {
```
bvaughn, line 128: Seems like this could use an inline comment explaining why we're doing it.

bvaughn (approved):

PR #10463, opened 2024-03-26
[RUN-3274] Re-enable `repaint-01`

Domiii:
* https://linear.app/replay/issue/RUN-3274/re-enable-repaint-01

I think this is fixed, is it?


bvaughn (approved):
Is this ready to be re-enabled from the RUN side?

bvaughn:
This seems have been abandoned.

PR #10467, opened 2024-03-27
Clear showHoverTimeGraphics after dragging or saving the focus window

hbenl:
- clear `hoverTime` and `showHoverTimeGraphics` when leaving focus edit mode
- clear `showHoverTimeGraphics` after dragging the focus window, its beginning or end
- remove `updateFocusWindowThrottled()`: this is no longer necessary (because we don't sync the "real" focus window to the displayed one during focus edit mode anymore) and was causing timing issues

### DIFF
--- a/src/ui/components/Timeline/FocusModePopout.tsx
+++ b/src/ui/components/Timeline/FocusModePopout.tsx
@@ -5,8 +5,7 @@ import { Button } from "replay-next/components/Button";
 import { useNag } from "replay-next/src/hooks/useNag";
 import { ReplayClientContext } from "shared/client/ReplayClientContext";
 import { Nag } from "shared/graphql/types";
-import { DebouncedOrThrottledFunction } from "shared/utils/function";
-import { exitFocusMode, requestFocusWindow } from "ui/actions/timeline";
+import { exitFocusMode, requestFocusWindow, setHoverTime } from "ui/actions/timeline";
 import {
   getFocusWindow,
   getShowFocusModeControls,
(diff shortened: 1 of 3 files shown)


src/ui/components/Timeline/FocusModePopout.tsx
```diff
@@ -43,16 +35,14 @@ export default function FocusModePopout({
...
-      updateFocusWindowThrottled.cancel();
-    }
-
     const currentFocusWindow = replayClient.getCurrentFocusWindow();
     assert(currentFocusWindow);
     dispatch(
       setFocusWindow({ begin: currentFocusWindow.begin.time, end: currentFocusWindow.end.time })
     );
 
+    dispatch(setHoverTime(null, false));
```
bvaughn, line 44: nit: We could use a comment explaining why we do this. (Your explanation in the Linear issue makes sense but glancing at the code, it's maybe not obvious.)

src/ui/components/Timeline/Focuser.tsx
```diff
@@ -18,45 +22,24 @@ function stopEvent(event: MouseEvent) {
...
 
-  // Mirror focus state so we can re-render immediately and dispatch throttled Redux updates
-  const [displayedFocusWindow, setDisplayedFocusWindow] = useState({
-    beginTime: focusWindow?.begin ?? zoomRegion.beginTime,
-    endTime: focusWindow?.end ?? zoomRegion.endTime,
-  });
-  const displayedFocusWindowRef = useRef(displayedFocusWindow);
-  useEffect(() => {
-    displayedFocusWindowRef.current = displayedFocusWindow;
-  });
```
bvaughn, line 58: This is a nice simplification

bvaughn (approved):

PR #10470, opened 2024-03-29
ci: draw lines in honeycomb when frontend tests start and end

jazzdan:
We suspect these tests may be causing issues in the backend. This should help us see that more clearly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,25 @@ jobs:
           script: return require("uuid").v4()
     outputs:
       testRunId: ${{ steps.uuid.outputs.result }}
+  start-e2e-marker:
+    runs-on: ubuntu-latest
+    needs: [preview-branch]
+    outputs:
+      id: ${{ steps.start-e2e-marker.outputs.id }}
+    steps:
+      - name: Honeycomb Start Marker for E2E Tests
+        id: start-e2e-marker
+        uses: pawelros/gha-honeycomb-marker@v1.0.0
+        with:
+          api-key: ${{ secrets.HONEYCOMB_API_KEY }}
+          dataset: "backend"
+          operation: "create"
+          type: "e2e_test"
+          message: "E2E Tests - Run #${{ github.run_number }}"
+          url: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+      - name: Log Outputs
+        run: |
+          echo "ID Output: ${{ steps.start-e2e-marker.outputs.id }}"
   e2etest:
     name: End-to-end tests (${{ matrix.shard }})
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -189,6 +203,20 @@ jobs:
         with:
           api-key: ${{ env.RECORD_REPLAY_API_KEY }}
           public: true
+  end-e2e-marker:
```
toshok, line 206: I think this job needs `if: always()` so it'll run even if the needed e2e tests fail.

.github/workflows/test.yml
```diff
@@ -62,6 +62,20 @@ jobs:
...
+    steps:
+      - name: Honeycomb Start Marker for E2E Tests
+        uses: pawelros/gha-honeycomb-marker@v1.0.0
+        id: e2e_start_marker
+        with:
+          api-key: ${{secrets.HONEYCOMB_API_KEY}}
+          dataset: "backend"
+          operation: "create"
+          type: "e2e_test_start"
+          message: "E2E Tests Start - Run #${{ github.run_number }}"
```
toshok, line 77: probably don't want "Start" here, since it'll be a range.

.github/workflows/test.yml
```diff
@@ -189,6 +203,20 @@ jobs:
...
+    steps:
+      - name: Honeycomb End Marker for E2E Tests
+        uses: pawelros/gha-honeycomb-marker@v1.0.0
+        with:
+          api-key: ${{secrets.HONEYCOMB_API_KEY}}
+          operation: "update"
+          id: ${{needs.start-e2e-marker.outputs.e2e_start_marker_id}}
+          dataset: "my-dataset"
+          type: ${{needs.e2etest.result == 'success' && 'e2e_test_complete' || 'e2e_test_failed'}}
+          message: "E2E Tests Start - Run #${{ github.run_number }} - ${{needs.e2etest.result == 'success' && 'OK' || 'FAILED'}}"
```
toshok, line 218: same here.  🪓  "Start"?

toshok (approved):
lgtm.  we'll need something similar in the e2e buildkite pipeline too, right?

jazzdan:
@toshok is this what you meant by getting rid of "Start"?

toshok:
@jazzdan yep!

bvaughn (approved):
🤷🏼  I guess this seems fine

bvaughn:
Looks like CI failed:
> Error: Honeycomb marker `id` is required for update operation type.

jazzdan:
Yup, I'm on it. Thanks @bvaughn

bvaughn:
@jazzdan We going to land this? :)

jazzdan:
@bvaughn this currently isn't doing what I want. I'm hoping to get back to it today with a fresh mind. Thanks for the ping!

bvaughn:
This seems have been abandoned.

jazzdan:
Yup, sorry, and thanks @bvaughn!

PR #10471, opened 2024-03-29
Fix ObjectRenderer bug that caused ellipsis to sometimes not render for truncated properties lists

bvaughn:
Looks like this regressed in a refactor PR #10248

I added a new e2e test for this edge case


bvaughn:
Holger's on PTO and Mark is on RCA so this might sit for a while. Going to merge and release it. Happy to follow up later.

PR #10472, opened 2024-03-29
Fix ObjectRenderer bug that caused ellipsis to sometimes not render for truncated properties lists

bvaughn:
Same as #10471 but for Gecko


bvaughn:
Holger's on PTO and Mark is on RCA so this might sit for a while. Going to merge and release it. Happy to follow up later.

PR #10476, opened 2024-04-03
Bump playwright plugin version

ryanjduffy:
Bumps the replay playwright plugin to the latest version after the yarn migration

### DIFF
--- a/packages/e2e-tests/package.json
+++ b/packages/e2e-tests/package.json
@@ -21,7 +21,7 @@
   "license": "ISC",
   "devDependencies": {
     "@playwright/test": "^1.37.0",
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I verified that we have annotations for both replays for the comments-02 tests. I thought this was regressing the two failing tests but it sounds like that's a broader problem that just happened to coincide with this PR.

bvaughn (approved):
Whoa, v2 alpha 9 to v3 alpha 3 😅 YOLO 😆

ryanjduffy:
I should have used `yolo` as the prerelease prefix ... `3.0.0-yolo.1` has a nice ring to it ;)

PR #10479, opened 2024-04-08
Remove leading "$" from comment tooltip

bvaughn:
Likely because this was originally a string (`${fileName}:${lineNumber}`) and then later got refactored to JSX text (`<div>{fileName}:{lineNumber}</div>`) and one of the "$" didn't get removed.

### Before
Screenshot 2024-04-08 at 10 19 45 AM

### After
Screenshot 2024-04-08 at 10 19 32 AM


Andarist (approved):

hbenl (approved):

bvaughn:
There are a lot of e2e test failures, but no way this PR broke those things so I'm going to land it and follow up with Product on-call

PR #10480, opened 2024-04-08
Fix Session.mayDestroy event listener

hbenl:
I forgot to import `addEventListener` from `socket.ts` and so this file used the DOM `addEventListener` instead.


bvaughn (commented):
Oh geeze.

bvaughn (approved):

PR #10483, opened 2024-04-10
Fetch breakpoint positions in buckets and split up position caches

markerikson:
This PR:

- Fixes a `CursorPosition` type that had been `any`
- Updates our breakpoint position caches:
  - Splits into two separate caches, an interval cache (for fetching chunks of per-line data), and a cache that returns a `Map` keyed by lines for that chunk (and relies on the interval cache for its data, to avoid duplicate fetching)
  - Adds a "bucket these line numbers" util that reuses the same logic we did for hit counts, but with a different bucket size. I've opted for 5000 lines for now - that is more than big enough to still fetch all positions for most source files in a single request, but small enough that we don't fetch an entire bundle's worth at once (and thus avoid the backend's serialization cost issue). Can totally tweak this if desired.
  - Updated all usage locations to use the new caches appropriately, including dropping one or two usages of the `Map`

At this point there's only one section of the client that still relies on the `Map` cache, and that's in the source viewer components. That gets passed through 3-4 levels of list components, so I didn't want to mess with that for now. Realistically we could probably change it to be fetching the breakpoints bucket for the current line and be passing that down through context, and get rid of the `Map` entirely.

### DIFF
--- a/src/devtools/client/debugger/src/utils/sourceVisualizations.ts
+++ b/src/devtools/client/debugger/src/utils/sourceVisualizations.ts
@@ -17,8 +17,10 @@ import { getPauseFrameSuspense } from "ui/suspense/frameCache";
 import { PauseAndFrameId } from "../reducers/pause";
 import { isBowerComponent, isNodeModule } from "./source";
 
-// TODO
-type CursorPosition = any;
+export type CursorPosition = {
+  readonly column: number;
+  readonly line: number;
+};
 
 export function getSourceIDsToSearch(
   sourcesById: Map<string, SourceDetails>,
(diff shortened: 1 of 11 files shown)


src/devtools/client/debugger/src/utils/sourceVisualizations.ts
```diff
@@ -196,40 +199,28 @@ function getAlternateSourceIdForPositionSuspense(
...
-  if (firstBreakableLineAfterPosition === null) {
+  if (!firstBreakableLineAfterPosition) {
     return undefined;
   }
 
-  const breakableLineLocations = breakablePositionsByLine.get(firstBreakableLineAfterPosition)!;
-
   // Now we can ask the backend for alternate locations with that line.
-  let breakableColumn = breakableLineLocations.columns[0];
+  let breakableColumn = firstBreakableLineAfterPosition.columns[0];
```
bvaughn, line 219: This always guaranteed to be defined?

bvaughn (commented):
I don't have the mental bandwidth to really dig into this right now :( Sorry

src/devtools/client/debugger/src/utils/sourceVisualizations.ts
markerikson: Yeah, there's a bailout just before this if it isn't.

PR #10491, opened 2024-04-18
Retry saving breakpoints in tests when type ahead hides the button

ryanjduffy:
Added comments to https://app.replay.io/recording/logpoints-02-conditional-log-points--48d35edf-4c2c-4763-a91c-2454e47a399d. The cause is a race condition in which the type-ahead box will sometime show up after we try to clear it but before we click the save button.

If the logic for managing the save button were specific to this test, we could wait on the dialog and then clear it. Instead, the logic is shared for any add or edit logpoint so i'm adding retries to ensure we handle this more cleanly.


bvaughn (approved):
Worth a shot.

~~I think 2s is a bit long though.~~ Disregard. I read this as the interval at first.

PR #10492, opened 2024-04-22
Fix over-render caused by useTooltip

bvaughn:
Resolves PRO-237 and PRO-240.

See my comments in `fac34a14-0fe1-48d2-90de-9978320df955` for further info.


hbenl (approved):

bvaughn:
End to end tests seemingly passed, but all of the CI shards seem to have failed after upload, e.g.
> Error: The operation was canceled.

Why was the operation cancelled? What does that mean?

PR #10493, opened 2024-04-22
Remove old library code

hbenl:


src/ui/utils/time.ts
```diff
@@ -9,3 +10,64 @@ export function formatTimestamp(ms: number) {
   const seconds = Math.round(ms / 1000.0);
   return `${Math.floor(seconds / 60)}:${padStart(String(seconds % 60), 2, "0")}`;
 }
+
+export function getDurationString(durationMs: number | null | undefined) {
```
bvaughn, line 14: I don't think anything outside of the library uses this method. Why are we preserving it?

src/ui/utils/time.ts
```diff
@@ -9,3 +10,64 @@ export function formatTimestamp(ms: number) {
...
+export function getTruncatedRelativeDate(date: string, addSuffix = true) {
+  return getRelativeDate(date, true, addSuffix);
+}
+
+export function getRelativeDate(date: string, truncate: boolean = false, addSuffix = true) {
+  let content = formatDistanceToNow(new Date(date), { addSuffix });
+
+  if (truncate) {
+    content = shortenRelativeDate(content);
+  }
```
bvaughn, line 45: nit: Nothing calls `getRelativeDate` directly, only `getTruncatedRelativeDate`, and nothing overrides the default param values.

```suggestion
export function getRelativeDate(date: string) {
  let content = shortenRelativeDate(
    formatDistanceToNow(new Date(date), { addSuffix: true })
  );
```

bvaughn (approved):

PR #10496, opened 2024-04-22
Force earlier CLI version for replay upload in CI

ryanjduffy:
The latest CLI is working but causing GH to hang when used with
action-upload. Rolling back to an earlier version resolves the
issue for this repo until we figure out the root cause and fix it.


bvaughn:
Wow. Nice find

PR #10498, opened 2024-04-23
Wait for the logpoint to be saved in e2e tests

hbenl:
I think this may have been the cause of some recent `logpoints-02` test failures.


bvaughn (approved):
Ah this would for sure cause a problem.

PR #10499, opened 2024-04-23
[TT-1001] Loosen Runtime CI assumptions on devtools behavior

Domiii:
* https://linear.app/replay/issue/TT-1001/e2e-tests-fail-for-chromium-because-of-connection-tester-assumption


jazzdan (approved):
Should be fine?

bvaughn (approved):
Sure why not :)

PR #10500, opened 2024-04-23
Fixed frame steps cache key

Andarist:
When debugging one of the test failures I noticed that this cache key gets computed as something like
> e6ee6c7c-97d2-453c-a03c-0bde20c7c290:0:[object Object]-[object Object]

I confirmed that this protocol type hasn't changed since 0.27.0 (the first released version of `@replayio/protocol`)


packages/replay-next/src/suspense/FrameStepsCache.ts
```diff
@@ -9,7 +9,7 @@ import { sourcesByIdCache } from "./SourcesCache";
 
 function getKey([client, pauseId, frameId]: [ReplayClientInterface, PauseId, FrameId]) {
   const focusWindow = client.getCurrentFocusWindow();
-  return `${pauseId}:${frameId}` + (focusWindow ? `:${focusWindow.begin}-${focusWindow.end}` : "");
+  return `${pauseId}:${frameId}` + (focusWindow ? `:${focusWindow.begin.point}-${focusWindow.end.point}` : "");
```
Andarist, line 12: `TimeStampedPoint` has `.point` and `.time`. From what I understand they both describe the same thing - just in a different format. So we can use one of them, please correct me if I'm wrong about this one

bvaughn (approved):

bvaughn:
Need to fix lint

packages/replay-next/src/suspense/FrameStepsCache.ts
bvaughn: Yeah. Point is more precise so using that seems right.

bvaughn:
Hey @Andarist can we land this?

PR #10501, opened 2024-04-23
[TT-1006] update the runtime tests to start both servers.

toshok:
Following instructions [here](https://github.com/replayio/devtools?tab=readme-ov-file#local-development).

works for me locally (forcing `runInCI` to true)

### DIFF
--- a/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
+++ b/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
@@ -280,7 +280,7 @@ export default async function run_fe_tests(
 
     // make sure the servers are up and running.
     console.log("waiting for dev servers to start up");
-    await testHttpConnection("http://localhost:8080/");
+    await testHttpConnection("http://localhost:8081/");
     console.log("dev servers up, continuing with test");
   }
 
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
```diff
@@ -234,32 +234,60 @@ export default async function run_fe_tests(
...
       stdio: "inherit",
     });
 
-    // make sure the dev server is up and running.
-    console.log("waiting for dev server to start up");
-    await testHttpConnection("http://localhost:8080/");
-    console.log("dev server up, continuing with test");
+    // make sure the servers are up and running.
+    console.log("waiting for dev servers to start up");
+    await testHttpConnection("http://localhost:8081/");
```
bvaughn, line 289: I think we could check `http://localhost:8081/login` instead if we want to get a 200 without a possible redirect

bvaughn (approved):
I think this looks reasonable at a glance

packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
toshok: ah yeah I meant to add that :+1:

bvaughn: I like this URL because it will not load at all if DevTools isn't running yet and it will return a 500 error if Dashboard isn't running. Otherwise it will always return a 200 status (regardless of authentication)

bvaughn:
Seems like the test failures are unrelated to your change. (See PRO-281 for more details)

bvaughn:
Hey @toshok should we land this or has it been replaced by something else?

hbenl:
This has been replaced by #10503 so I'll close this.

PR #10502, opened 2024-04-23
Always render `FrameTimeline` for the latest `selectedFrameId`

Andarist:
I verified that (by mocking this situation in code) this would prevent https://github.com/replayio/devtools/blob/853232ad9058009c3a96ef89ac5ac9ccb6783e41/packages/e2e-tests/tests/node_stepping-01.test.ts#L17 from failing when the first `Pause.getFrameSteps` fails to be resolved.

It doesn't fix the underlying issue (the request can still fail/timeout) but at least the `FrameTimeline` won't get completely stuck if this happens - which is an improvement in itself.


jasonLaster (commented):
can we write an E2E test that verifies this behavior?

Andarist:
@jasonLaster this is quite time-sensitive. The recording that captured the issue had to meet those 2 conditions:
- the first `Pause.getFrameSteps` had to timeout
- the first render of `FrameTimeline` had to be already rendered with existing `selectedFrameId` (at times the first render was with `null`/`undefined` and in such a scenario the test *didn't* fail)

I don't see how to write e2e test for this that wouldn't be super brittle

src/devtools/client/debugger/src/components/SecondaryPanes/FrameTimeline.tsx
```diff
@@ -261,8 +268,10 @@ function FrameTimeline() {
 }
 
 export default function FrameTimelineSuspenseWrapper() {
+  const selectedFrameId = useAppSelector(getSelectedFrameId);
   return (
     <Suspense
+      key={selectedFrameId ? `${selectedFrameId.pauseId}:${selectedFrameId.frameId}` : null}
```
bvaughn, line 274: Was this `key` change part a necessary part of the fix? (Can you elaborate on why it was needed, if so?)

Andarist: Yes, this is the core of the fix. We were suspending *within* `FrameTimeline` and we were waiting for the awaited value to resolve there. That never happened because the relevant requested timedout~ (we found a failed assert on BE but no error/timeout was actually returned to FE). So we never had a chance to start listening for a response to a new request - that would get dispatched if we'd rerender with the new source information (`selectedFrameId`). In other words, we were indefinitely waiting for a stale data to be resolved.

bvaughn: It surprises me though that you needed to hard-reset the component like this (via `key`)

I would have expected React to bail on the old suspended IO if props/state changed

Andarist: This component has no props and I guess that subscriptions (for state changes) might not even be set up when the component is suspended. So there is no trigger for it to try again in a situation like this.

bvaughn: Hmm. It _had_ no props, but now it does (`selectedFrameId `) so...I'm not sure I get it

Andarist: you are right: https://github.com/replayio/devtools/pull/10505

PR #10505, opened 2024-04-24
Remove redundant `key` from `FrameTimeline`

Andarist:
As @bvaughn observed [here](https://github.com/replayio/devtools/pull/10502#discussion_r1578079781) the fix wasn't strictly related to adding a `key` but rather just to triggering a new render here. That's already covered by passing new props to `FrameTimeline` so the `key` is redundant. 

I tested this locally in the same way as I did with the previous PR - by mocking this situation in code.


bvaughn (approved):
Thanks for the follow up! I'm relieved to hear that this wasn't some bizarre React Suspense bug.

Andarist:
yeah, I focused on the `key` after recognizing the core issue and then I just happened to have to add props for the structure to make sense here and I didn't realize that that just made `key` redundant ;p

PR #10509, opened 2024-04-26
Make sure that Scopes display loading state when seek happens

Andarist:
While investigating a test failure I noticed that it's possible to see such a state
Frames panel displays a loading state while Scopes display that we are not paused on a point with scopes

This PR improves this by making sure that Scopes also display a loading state - just like Frames - in this situation

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
@@ -233,7 +233,7 @@ function Frames({ panel, point, time }: FramesProps) {
   );
 }
 
-export default function FramesSuspenseWrapper(props: FramesProps) {
+export default function NewFrames(props: FramesProps) {
   return (
     <Suspense
       fallback={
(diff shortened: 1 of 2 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -233,7 +233,7 @@ function Frames({ panel, point, time }: FramesProps) {
   );
 }
 
-export default function FramesSuspenseWrapper(props: FramesProps) {
+export default function NewFrames(props: FramesProps) {
```
Andarist, line 236: just matching this name to the filename and how it's called in other files

bvaughn (approved):
This change makes sense to me. Would be nice to get @hbenl's sign off too though.

src/devtools/client/debugger/src/components/SecondaryPanes/NewScopes.tsx
```diff
@@ -17,15 +17,25 @@ import { getPreferredGeneratedSources } from "ui/reducers/sources";
...
 function ScopesRenderer() {
   const replayClient = useContext(ReplayClientContext);
   const sourcesById = sourcesByIdCache.read(replayClient);
   const preferredGeneratedSources = useAppSelector(getPreferredGeneratedSources);
   const selectedFrameId = useAppSelector(getSelectedFrameId);
+  const seekLock = useAppSelector(getSeekLock);
+
+  if (seekLock) {
+    return <LoadingInfo />;
+  }
```
Andarist, line 37: conceptually it could make sense to have a similar check in the Frames but:
- I couldn't make those 2 to desynchronize right now
- I'm not quite sure where exactly I would squeeze it there (things are rendered recursively there)
- it's not worth putting more effort into this rn ;p

hbenl (approved):

PR #10510, opened 2024-04-26
Fix accordion border color

bvaughn:
I noticed that our border colors in the accordion component had sometime recently gotten broken:
image

This PR fixes them.
Screenshot 2024-04-26 at 9 26 02 AM

### DIFF
--- a/packages/accordion/index.tsx
+++ b/packages/accordion/index.tsx
@@ -53,12 +53,13 @@ function ResizeHandle({
   return (
     <div
       className={classNames(
-        "absolute w-full border-chrome hover:border-blue-400",
+        "absolute w-full hover:border-blue-400",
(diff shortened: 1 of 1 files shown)


packages/accordion/index.tsx
```diff
@@ -53,12 +53,13 @@ function ResizeHandle({
   return (
     <div
       className={classNames(
-        "absolute w-full border-chrome hover:border-blue-400",
+        "absolute w-full hover:border-blue-400",
```
bvaughn, line 56: Any Tailwind border color style I tried declaring– even built-in colors– got dropped. I don't know why this is. It does not seem worth spending much time debugging right now though.

Andarist (approved):

PR #10513, opened 2024-04-26
Remove legacy auth code

hbenl:


src/ui/utils/browser.ts
```diff
@@ -1,15 +1,3 @@
-export function setUserInBrowserPrefs(user: { sub: string } | null) {
```
hbenl, line 1: This isn't actually supported in our browser anymore (I get "no such channel" errors when using it) so I removed it completely.

pages/[id]/sourcemap/[sourceId].tsx
```diff
@@ -7,13 +7,12 @@ import { client, initSocket } from "protocol/socket";
...
 import renderSourcemap from "third-party/sourcemap-visualizer/sourcemapVisualizer";
 import { UIStore } from "ui/actions";
-import { setAppMode } from "ui/actions/app";
+import { getAccessToken, setAppMode } from "ui/actions/app";
 import { getAccessibleRecording } from "ui/actions/session";
 import { ExpectedErrorModal } from "ui/components/Errors/ExpectedErrorModal";
 import LoadingScreen from "ui/components/shared/LoadingScreen";
 import { useGetRecordingId } from "ui/hooks/recordings";
 import { useAppDispatch, useAppStore } from "ui/setup/hooks";
-import tokenManager from "ui/utils/tokenManager";
```
Comment (bvaughn) [inline pages/[id]/sourcemap/[sourceId].tsx:16]: Good riddens!

pages/_app.tsx
```diff
@@ -131,18 +92,31 @@ const App = ({ apiKey, ...props }: AppProps & AuthProps) => {
...
       setToken({ token });
     }
+
+    async function getTokenFromBrowser() {
+      const token = await Promise.race([
+        new Promise<string>(resolve => listenForAccessToken(resolve)),
+        // if the user is not logged in, the listenForAccessToken callback
+        // is never called, so we add this timeout to continue without a token
+        new Promise<null>(resolve => setTimeout(() => resolve(null), 500)),
+      ]);
```
bvaughn, line 111: I _guess_ this makes sense but it seems kind of iffy. Is 500ms enough time to always be safe?

bvaughn (approved):

pages/_app.tsx
hbenl: Yeah, I'd prefer if the browser sent us a "no access token" response. Anyway in my tests the browser responded within a few milliseconds, so 500ms should be enough.

PR #10516, opened 2024-04-29
Replace middleware with rewrite config

hbenl:
I had only written the middleware because I thought that "all paths that don't start with `/recording`" couldn't be expressed in the next config but I was wrong.
Since Vercel charges for middleware invocations, we shouldn't use a middleware if we can avoid it.


bvaughn (commented):
Nice cleanup!

This is funny. The Vercel support engineer literally emailed a similar suggestion overnight:

```js
basePath: "/recording",
  redirects: async () => {
    return [
      {
        source: "/((?!recording).*)",
        destination: "https://your-prod-domain.com/",
        basePath: false,
        permanent: true,
      },
    ];
  }
```

bvaughn (approved):

PR #10522, opened 2024-05-04
[POC] dub minifier

jasonLaster:
Adds a dub link minifier which we can use in two places

1. in the share modal for getting a short link that is nicer to share
2. getting a url for cli

You give it an ID and URL and you get back a short link e.g. https://replay.run/0gR7Y1z

```js
(async () => {
  const replays = [
    {
      url: "https://app.replay.io/recording/cannot-toggle-video--cc968cff-245d-4916-9db4-a36a99d07417?commentId=&focusWindow=&point=&primaryPanel=comments&secondaryPanel=console&time=0&viewMode=non-dev",
      id: "cc968cff-245d-4916-9db4-a36a99d07417",
    },
    {
      id: "763b4c78-300b-4516-a092-c462af80ccfa",
      url: "https://app.replay.io/recording/cypresse2especcyts--763b4c78-300b-4516-a092-c462af80ccfa?focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiMjI3MTYyOTg3NjA5OTgyMzU4NzgxODU3ODAwNjI5NzgwNDgiLCJ0aW1lIjoxMDM5M319&point=2920666983521556235400266624532820&primaryPanel=cypress&secondaryPanel=console&time=1750&viewMode=dev&commentId=",
    },
    {
      id: "06e0e381-d89a-490b-b4e9-313b6702c487",
      url: "https://app.replay.io/recording/can-buy-board--06e0e381-d89a-490b-b4e9-313b6702c487?focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiMjgyMzMxMTQxNzA0Njc2MjU1NTY4MzE0MDYzMTE2MDQyMjQiLCJ0aW1lIjoxMDk4N319&point=2920666983183140339074523438514314&primaryPanel=cypress&secondaryPanel=console&time=1324&viewMode=non-dev#",
    },
  ];

  const response = await fetch("http://localhost:8080/recording/api/createLink", {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: JSON.stringify(replays[2]),
  });

  const data = await response.json();
  console.log(data);
})();
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "date-fns": "^2.28.0",
     "design": "workspace:*",
     "dotenv": "^16.0.0",
+    "dub": "^0.26.0",
     "error-stack-parser": "^2.1.4",
     "escape-html": "^1.0.3",
     "fluent-ffmpeg": "^2.1.2",
@@ -94,7 +95,8 @@
     "shared": "workspace:*",
     "slugify": "^1.6.5",
     "suspense": "^0.0.53",
-    "use-context-menu": "0.4.13"
+    "use-context-menu": "0.4.13",
+    "zod": "^3.23.6"
   },
   "devDependencies": {
     "@babel/core": "^7.17.8",
--- a/pages/api/createLink.ts
+++ b/pages/api/createLink.ts
@@ -1,0 +1,40 @@
+import { Dub } from "dub";
+import { NextApiRequest, NextApiResponse } from "next";
+
+const dub = new Dub({
+  token: process.env.DUB_API_KEY,
+  workspaceId: process.env.DUB_WORKSPACE_ID,
+});
+
+async function doesLinkExist(id: string) {
+  try {
+    const found = await dub.links.get({
+      domain: "replay.run",
+      externalId: `ext_${id}`,
+    });
+    return found;
+  } catch (error: any) {
+    const dubError = error.error;
+    if (dubError.code != "not_found") {
+      return null;
+    }
+  }
+}
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  const { id, url } = req.body;
+
+  const found = await doesLinkExist(id);
+  if (found) {
+    return res.status(200).json(found);
+  }
+
+  const newLink = await dub.links.create({
+    url,
+    domain: "replay.run",
+    externalId: id,
+    tagIds: ["clvrlf0y40001u5txsd0p0efa"],
+  });
+
+  return res.status(200).json(newLink);
+}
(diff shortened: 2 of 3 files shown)


bvaughn:
Hey @jasonLaster I filed a Linear issue for this (PRO-388). I think we should treat this as a feature in terms of prioritizing.

I don't know anything about how Dub works (where it stores the map between links and full URLs). I'd personally want to be really clear about that before we moved forward. Also want to make sure we keep the CLI and Devtools "Share" dialogs in sync.

PS Love the replay.run domain!

jasonLaster:
Let's close this pr until we want to revaluate. By the way part of the rationale is that I'm a fan of Steven tey's. I'm sure he'll be happy to discuss when we're ready to evaluate options.

bvaughn:
> Let's close this pr until we want to revaluate

Makes sense!

jasonLaster:
> where it stores the map between links and full URLs

It is okay for vendors to have user data e.g. honeycomb. The important piece is that it is documented. This would be applicable if the user passed along the short title and og image.

jasonLaster:
Re-opening the short links PR with the goal of making it easy to create a short link from devtools in either the share modal or the command palette.

https://www.loom.com/share/69bfd87dfbdf48c280109077da996e2d

The API endpoint itself is minimal, so i think the only concern will be around whether we use the hosted version, self-hosted version, or roll our own. I don't see a good reason at the moment to opt for the self-hosted version which we'd have to maintain. There's no sensitive data and if at some point we wanted to host it we could. Rolling our own version is similar. If at some point we wanted to bring this in house it would be fairly easy to migrate the existing shortLinks.

bvaughn:
To my knowledge, nothing has changed since we last talked about this. Why the push to reopen and land this?

bvaughn:
Or put another way, why are not prioritizing this work using our normal process? We have a Linear issue for this (PRO-388) which got backlogged when we last looked at this relative to other priorities.

jasonLaster:
Taking this discussion offline because I think it’s easier for the two of us to discuss.

package.json
```diff
@@ -94,7 +95,8 @@
     "shared": "workspace:*",
     "slugify": "^1.6.5",
     "suspense": "^0.0.53",
-    "use-context-menu": "0.4.13"
+    "use-context-menu": "0.4.13",
+    "zod": "^3.23.6"
```
bvaughn, line 99: Btw it looks like you added an unused dependency here (`zod`)

jasonLaster: That's something that dub requires. It might need to be a peer dep or something.

PR #10528, opened 2024-05-13
Don't show processing estimation for recordings with a duration of 0

bvaughn:
Looking at other recordings, I thought this would be null/undefined or a valid value. Seems like it could also be 0 in which case, we shouldn't show an estimated processing duration.

cc @miriambudayr 

| before | after |
| :---: | :---: |
| Screenshot 2024-05-13 at 1 38 02 PM | Screenshot 2024-05-13 at 1 37 43 PM |

### DIFF
--- a/src/ui/components/DevToolsProcessingScreen.tsx
+++ b/src/ui/components/DevToolsProcessingScreen.tsx
@@ -36,7 +36,7 @@ export function DevToolsProcessingScreen() {
 
   if (showStalledMessage) {
     secondaryMessage = "There may be a problem. This is taking much longer than expected.";
-  } else if (recording?.duration != null) {
+  } else if (recording?.duration) {
     const durationLabel = formatEstimatedProcessingDuration(recording?.duration);
     secondaryMessage = `Based on the length of this Replay, we expect processing will take about ${durationLabel}`;
   }


src/ui/components/DevToolsProcessingScreen.tsx
```diff
@@ -36,7 +36,7 @@ export function DevToolsProcessingScreen() {
 
   if (showStalledMessage) {
     secondaryMessage = "There may be a problem. This is taking much longer than expected.";
-  } else if (recording?.duration != null) {
+  } else if (recording?.duration) {
```
Andarist, line 39: So for `recording.duration === 0` we'd show now: "This could take a while, depending on the complexity and length of the replay.", right? Would it make sense to display a prediction for `1` instead of that? Are we talking about duration in seconds here?

Andarist (approved):

src/ui/components/DevToolsProcessingScreen.tsx
bvaughn: I'm chatting with Miriam right now about a user report of a 20 minute recording that (for some reason) has duration:0 in the DB. I think it's safest not to estimate anything in that case. (We _were_ estimating 1 minute, but that was way off.)

bvaughn: `formatEstimatedProcessingDuration` rounds up to the nearest min for what it's worth, so if we did nothing (current behavior) we'd predict 1 minute.

PR #10530, opened 2024-05-13
Fixed Fast Refresh complaining about socket re-initialization

Andarist:
When debugging I noticed an error to be thrown:
> Session callbacks can only be set once

and I narrowed it down to Fast Refresh setting this effect once again.


packages/replay-next/components/Initializer.tsx
Andarist, line 1: as far as I can tell this was unused

src/ui/components/DevTools.tsx
```diff
@@ -225,7 +225,13 @@ function _DevTools({
...
   });
 
+  const didInitializeSocketRef = useRef<boolean>(false);
   useEffect(() => {
+    if (didInitializeSocketRef.current) {
+      return;
+    }
+    didInitializeSocketRef.current = true;
+
     async function createSocketWithToken() {
```
Andarist, line 235: I'm not completely sure if this is never meant to get called again (like with a new `accessToken`). Based on a quick reading of the code that would always~ trigger the mentioned error so I assume it shouldn't be called more than once

Andarist: + from what I can tell `setAccessToken` could also be removed, I can't find any dispatch site for it

Andarist:
Actually, I can still see how this error gets triggered at times - I confirmed that sometimes `<DevTools/>` are completely remounted and when that happens the ref trick doesn't help. This effect should probably get hoisted closer to the root or something. I'll take another look at this later.

bvaughn (approved):

bvaughn:
Ping @Andarist. Should we land this or close it?

Andarist:
It does improve the situation for edits within the `src/ui/components/DevTools.tsx` but it doesn't solve everything. Let's close this for now and revisit later

PR #10532, opened 2024-05-15
Bypass UploadScreen if the user has no workspaces

hbenl:


pages/upload.tsx
```diff
@@ -12,15 +13,44 @@ function UploadScreenWrapper({ onUpload }: { onUpload: () => void }) {
...
+        variables: { recordingId, title: recording?.title || "Untitled", workspaceId: null },
+      });
+      onUpload();
+    }
+    if (
+      recording?.isInitialized === false &&
+      !initializing &&
+      !workspacesLoading &&
+      workspaces.length === 0 &&
+      getRecordingTarget(recording.buildId ?? "unknown") !== RecordingTarget.gecko
```
bvaughn, line 38: Wish we had a better, capabilities based way of detecting this. (Although maybe we want to also bypass this screen for Chromium recordings that are streamed in the background while being recorded?)

bvaughn (approved):
I think this looks okay. I didn't smoke test it though.

PR #10533, opened 2024-05-16
Fix the retry logic in the editLogPoint test helper

hbenl:
The `logpoints-02` e2e test flaked again for the same reason we had tried to fix in the past: a logpoint couldn't be saved because the type-ahead box hides the logpoint's save button ([PRO-444](https://linear.app/replay/issue/PRO-444/e2e-test-failure-logpoints-02-at-pointpanel-savebutton) and [PRO-238](https://linear.app/replay/issue/PRO-238/locatorclick-timeout-60000ms-exceeded-waiting-for-locator[data-test)).
In #10491 we added retry logic so that the type-ahead box is cleared if it showed up after we tried to clear it for the first time but before the save button is clicked. However, that logic didn't work: the click action had a default timeout of 60 seconds but the surrounding retry logic had a timeout of 2 seconds, so if the click action failed it was never retried.

### DIFF
--- a/packages/e2e-tests/helpers/source-panel.ts
+++ b/packages/e2e-tests/helpers/source-panel.ts
@@ -337,24 +337,21 @@ export async function editLogPoint(
     // The typeahead popup sometimes sticks around and overlaps the save button.
     // Sometimes, it will show up after we check the first time (PRO-238) so we
     // retry a couple times to ensure that we can clear it and move forward.
-    await waitFor(
-      async () => {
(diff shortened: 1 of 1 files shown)


bvaughn (approved):
> the click action had a default timeout of 60 seconds but the surrounding retry logic had a timeout of 2 seconds, so if the click action failed it was never retried.

Ooof! Wish there was a way for our `waitFor` function to override the default timeout values for things executed within its callbacks.

PR #10534, opened 2024-05-17
Don't allow uploading a recording to an expired workspace

hbenl:
Implements parts 1 and 2 of the proposal in [PRO-57](https://linear.app/replay/issue/PRO-57/expired-workspaces-shouldnt-block-you-from-seeing-the-replay).


bvaughn:
Does this change still make sense to land now that there is consensus on enforcing this on the backend?

bvaughn:
We've reached consensus that PRO-455 is the better way to approach this (disallow on the backend) so we should close this PR.

bvaughn:
That's fair. Thanks for sharing the context.

PR #10535, opened 2024-05-21
Update `@replayio/playwright`

Andarist:


.github/workflows/test.yml
```diff
@@ -97,8 +97,8 @@ jobs:
         env:
           YARN_CHECKSUM_BEHAVIOR: "update"
           PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
-      - name: Install Replay Playwright
-        run: npx @replayio/playwright install
+      - name: Install Replay Chromium
+        run: npx replayio install
```
Andarist, line 101: this won't work yet because we havent yet released a version with this alias of the `update` command

PR #10536, opened 2024-05-22
fix e2e script after @replayio/playwright upgrade

toshok:
`getExecutablePath` lost its parameter


bvaughn (approved):
I'm surprised this broke anything but the change looks okay

PR #10539, opened 2024-05-25
On `save-examples` failure: Retry with `RECORD_REPLAY_VERBOSE` enabled

Domiii:
* https://linear.app/replay/issue/TT-1261/recorder-crashes-from-save-examplests-are-not-uploaded#comment-4b77231f


packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -446,8 +463,7 @@ async function waitUntilMessage(
 
     process.exit(0);
   } catch (error) {
-    console.error("Unexpected error: ", error);
-
+    console.error("☠☠☠ save-examples failed - ", error?.stack || error);
```
bvaughn, line 466: Why this change?

Logging `error` will print both the error message _and_stack which seems nicer?

packages/e2e-tests/scripts/save-examples.ts
```diff
@@ -242,28 +243,44 @@ async function saveBrowserExample({ example }: TestRunCallbackArgs) {
...
+      if (config.useExampleFile && recordingId) {
+        await saveRecording(example.filename, config.replayApiKey, recordingId, true);
+      }
 
-  if (config.useExampleFile && recordingId) {
-    await saveRecording(example.filename, config.replayApiKey, recordingId, true);
+      if (recordingId) {
+        removeRecording(recordingId);
+      }
+      break;
```
bvaughn, line 273: What happens if recording fails the first time but then succeeds on retry? Seems like this would still throw and say that recording failed.

Am I reading it wrong? Seems like we would want a retry to succeed.

Seems like the most straight forward way to do that would be to return here rather than _breaking_.
```suggestion
      return;
```

bvaughn (commented):
[CI failure](https://github.com/replayio/devtools/actions/runs/9235731964/job/25410826668?pr=10539) is saying you need to run `yarn prettier` to fix formatting issues.

packages/e2e-tests/scripts/save-examples.ts
Domiii: I don't want a retry to succeed. The root cause here is likely/often going to be a recorder crash, and we want to show no leniency toward those! I am only retrying to get a verbose error output 🙄

packages/e2e-tests/scripts/save-examples.ts
Domiii: * Hum... I'm always doing `error.stack`, because for the longest time, I guess I had middleware that did not print errors correctly. Not sure if old Node might also have had problems with it, all I know is that it was a real problem.
* Also note that `Error.stack` includes `message`, so this gets us the same thing.

But I cannot find any evidence of this being a problem here, so I can revert it.

packages/e2e-tests/scripts/save-examples.ts
bvaughn: Ah, I see. In that case, I think the "recordingAttempts" variable is kind of misleading (at least to me). You're not really _attempting_ some number of times to get a usable recording. You're trying once, and if it fails, you're re-running in verbose mode to get more error info. Seems like that may be more clear if you just hard-coded the retry logic and used a boolean (e.g. `retryingWithVerboseMode`) to track the current case.

callingmedic911:
I merged `main` to run the test suite. We were debugging the cause of missing tests in the dashboard and broke the comment bot. Sorry for the random commits.

bvaughn:
@Domiii FYI this is still failing for the same reason (need to run `npm run lint` locally)

bvaughn:
Closing this for now since it seems to have been abandoned. Happy to look again if someone wants to reopen.

PR #10541, opened 2024-05-27
Avoid spamming telemetry with repeated `reactdevtools.node_not_found` events

Andarist:
I noticed a lot of repeated network requests going out when scrolling through the Network/Elements panels and I linked them them to this call site.

### DIFF
--- a/src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
+++ b/src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
@@ -168,7 +168,10 @@ export class ReplayWall implements Wall {
           // Get the first node ID we found for this fiber ID, if available.
           const [nodeId] = fiberIdsToNodeIds.get(fiberId) ?? [];
           if (!nodeId) {
-            sendTelemetryEvent("reactdevtools.node_not_found", payload);
+            // skip on hitting this again for the same fiberId (maybe combo it with pauseId?)
+            sendTelemetryEvent("reactdevtools.node_not_found", payload, {
+              id: fiberId,
+            });
             return;
           }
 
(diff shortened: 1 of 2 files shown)


src/ui/utils/telemetry.ts
```diff
@@ -112,11 +112,26 @@ export function setTelemetryContext({ id, email, internal }: TelemetryUser) {
   }
 }
 
-export function sendTelemetryEvent(event: string, tags: any = {}) {
+const sentEvents = new Set<string>();
+
+type TelemetryOptions = {
+  /** It can be used to avoid sending repeated events. It's always combined with `event` for that purpose */
+  id?: string;
```
bvaughn, line 119: I think "id" might be a bit too ambiguous. I wonder if something like `distinctOn` might read more intuitively at the call sight?

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
```diff
@@ -168,7 +168,10 @@ export class ReplayWall implements Wall {
           // Get the first node ID we found for this fiber ID, if available.
           const [nodeId] = fiberIdsToNodeIds.get(fiberId) ?? [];
           if (!nodeId) {
-            sendTelemetryEvent("reactdevtools.node_not_found", payload);
+            // skip on hitting this again for the same fiberId (maybe combo it with pauseId?)
```
bvaughn, line 171: Mixed feelings here, but I think we should either remove the "maybe" comment or make it more concrete.

Replay's persistent Fiber ids are supposed to be the same for the entire recording, so that would be a reason to only key on the id as you're doing. However, if we can't find one at a particular execution point, that could indicate something unexpected which would be a reason to include the pause id in the data at least (even if not distinct on that id)

bvaughn (approved):

src/ui/utils/telemetry.ts
Andarist: i like that 👍

src/ui/components/SecondaryToolbox/react-devtools/ReplayWall.ts
Andarist: Sorry about that, this was a comment I added when I noticed the problem last week and it was by no means meant to be committed 🤦‍♂️

bvaughn: No worries :)

PR #10542, opened 2024-05-27
Add visibility badge to header

bvaughn:
When clicked, this badge opens the share modal dialog

https://github.com/replayio/devtools/assets/29597/bbabe90a-65d9-4331-a361-7d6a8e71538c


src/ui/components/Header/Header.tsx
```diff
@@ -210,6 +235,9 @@ export default function Header() {
         ) : (
           <div className={styles.ReadOnlyTitle}>Recordings</div>
         )}
+        <button className={styles.VisibilityIconAndText} onClick={onClick}>
+          <Icon className={styles.VisibilityIcon} type={iconType} /> {visibilityLabel}
+        </button>
```
Andarist, line 240: I wonder, wouldnt it make sense to get rid of the Share button with this in place?

Andarist (approved):

src/ui/components/Header/Header.tsx
Andarist: I know Share might call out explicitly what it is for and this button doesn't exactly scream that. It feels weird to have 2 trigger buttons for the same action side by side like this. Maybe they could be consolidated into a single element that would somehow convey both informations? This is probably a @jonbell-lot23 question though

hbenl (approved):

src/ui/components/Header/Header.tsx
jonbell-lot23: Yup, but the side-by-side configuration should be less common than something like this:
image

That gap between elements helps with [the law of proximity](https://lawsofux.com/law-of-proximity/) ("Elements in close proximity are perceived to share similar functionality or traits.") But it's true, there's a strange tradeoff when the screen is squished tight like this and two entry points go to the same place. 

--

One option is to use a segmented dropdown like Google does:
image

The problem with that is that it prioritises cute icons over clarity. I much prefer being as clear as possible, right in the header, with the word "public". Security/privacy/secrets/etc are too important to Replay's reputation to not make this crystal clear, like so:
image

--

I can see us revisiting this in the future, especially as we make the share dialog itself more clear. This is purely in the realm of design tradeoffs rather than one approach being objectively right or wrong.

Andarist: > Fortunately that's less common for devs doing DevTools work, who will usually have a non-squished screen.

Oh, I missed that they are only close to each other on squished screens. Given that, I feel much less concerned about those 2 elements triggering the same modal

bvaughn: > I wonder, wouldnt it make sense to get rid of the Share button with this in place?

I don't think the visibility badge is as clear of a call to action as a "Share" button so I would be reluctant to consider this a viable _replacement_ for that button.

PR #10545, opened 2024-05-28
Log the url of the session that failed

Domiii:
* This is a CI/test change only.
* [This test run](https://buildkite.com/replay/runtime-fe-end-to-end-tests/builds/1383#018fbf4e-09d4-4d03-9671-c8786cfd16e2) is an example of where it matters.
* Implementation: When we report `Session failed`, make sure to also give us the info we need to debug it, i.e. the Url of the recording whose session is failing. We need that to look up the relevant crash.

Sidenote: This message is usually encountered when the "golden recording" RTP crashed. The Runtime E2E test pipeline is more likely to encounter this problem, as we re-record golden recordings all the time, possibly on broken builds.

### DIFF
--- a/packages/e2e-tests/helpers/utils.ts
+++ b/packages/e2e-tests/helpers/utils.ts
@@ -131,12 +131,16 @@ export async function waitForRecordingToFinishIndexing(page: Page): Promise<void
     async () => {
       const supportFormErrorDetails = await getSupportFormErrorDetails(page);
       if (supportFormErrorDetails) {
-        throw new UnrecoverableError(`Session failed: ${supportFormErrorDetails}`);
+        throw new UnrecoverableError(
+          `Session failed (visiting ${page.url()}): ${supportFormErrorDetails}`
+        );
       }
 
       const expectedErrorDetails = await getExpectedErrorDetails(page);
       if (expectedErrorDetails) {
-        throw new UnrecoverableError(`Session failed: ${expectedErrorDetails}`);
+        throw new UnrecoverableError(
+          `Session failed (visiting ${page.url()}): ${expectedErrorDetails}`
+        );
       }
 
       if (await timelineCapsuleLocator.isVisible()) {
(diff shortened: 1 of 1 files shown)


bvaughn:
> Yes, but I find sifting through thousands of lines of logs to be a rather error-prone process.

Hmm. It's generally more like a couple dozen lines per test, and the logs should be pretty close together.

> Also, it requires intricate understanding of how and where we are logging what as it is implemented right now, and that is bound to change over time. This also makes things even harder for anyone not used to these logs.

I don't think it's likely to change over time FWIW. Fair point if it's less obvious to someone unfamiliar with the tests– but then I'm also not sure how that person would be intending to _use_ the URL that's logged. Can you share more about that?

hbenl (approved):

bvaughn (approved):
I don't think I fully understand the _why_ here (and I'd like to) but I'm not really opposed to the _what_

I think I'd rearrange the error messages so the URL came _last_ (and on a separate line) because it's lone and it may be clearer if we don't put it _before_ the error message.

Domiii:
> Hmm. It's generally more like a couple dozen lines per test, and the logs should be pretty close together.

I find that the logs are a total mess, and I can't navigate them. I only search for things, but every test name has > 50 results. I end up scrolling through those many results to find the sample name, and then again need to search the sample name for the crashing/failing recording id. While my process might be sub-optimal, having a clear answer for the failure just eliminates that search.

bvaughn:
> While my process might be sub-optimal, having a clear answer for the failing recording/url just eliminates that search.

I asked before though, can you tell me how you plan to _use the URL_ in this scenario?

Domiii:
Final update - I synced with Brian on this yesterday:

1. I fixed the formatting according to Brian's suggestion
2. I updated the description to better explain what is going on here.
3. To answer the last question: We need the url/recordingId to get to the crash report.

PR #10546, opened 2024-05-28
Warn about possible timeout loading React annotations

bvaughn:
## [Loom](https://www.loom.com/share/326724ac62914eca91382a263b558deb)

### DIFF
--- a/packages/replay-next/src/suspense/TimeoutCache.ts
+++ b/packages/replay-next/src/suspense/TimeoutCache.ts
@@ -1,0 +1,23 @@
+import { createStreamingCache } from "suspense";
+
+export const TimeoutCache = createStreamingCache<
+  [eventName: string, timeoutAfterMs: number],
+  boolean
+>({
+  debugLabel: "TimeoutCache",
+  getKey: (eventName: string, timeoutAfterMs: number) => `${eventName}:${timeoutAfterMs}`,
+  load: async (options, _, timeoutAfterMs: number) => {
+    const { resolve, signal, update } = options;
+
+    update(false);
+
+    const timeout = setTimeout(() => {
+      update(true);
+      resolve();
+    }, timeoutAfterMs);
+
+    signal.addEventListener("abort", () => {
+      clearTimeout(timeout);
+    });
+  },
+});
--- a/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
+++ b/src/ui/components/SecondaryToolbox/ReactDevTools.tsx
@@ -1,14 +1,54 @@
+import { PropsWithChildren, ReactNode, Suspense } from "react";
+import { useStreamingValue } from "suspense";
+
 import { InlineErrorBoundary } from "replay-next/components/errors/InlineErrorBoundary";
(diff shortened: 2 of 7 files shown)


hbenl (approved):

packages/replay-next/src/suspense/TimeoutCache.ts
```diff
@@ -0,0 +1,23 @@
+import { createStreamingCache } from "suspense";
+
+export const TimeoutCache = createStreamingCache<
```
Andarist, line 3: q: what's the lifetime of this cached value? at what times it's created and disposed?

src/ui/components/SecondaryToolbox/ReactDevTools.tsx
```diff
@@ -1,14 +1,54 @@
...
+
+function SuspenseWithLoadingStalledMessage({ children }: PropsWithChildren) {
+  let message: ReactNode | undefined = undefined;
+
+  const { value: showShortMessage } = useStreamingValue(
+    TimeoutCache.stream("react:annotations-loading-stalled:short", 30_000)
+  );
+  const { value: showLongMessage } = useStreamingValue(
+    TimeoutCache.stream("react:annotations-loading-stalled:long", 300_000)
+  );
```
Andarist, line 33: pretty neat stuff

Andarist (approved):

packages/replay-next/src/suspense/TimeoutCache.ts
bvaughn: Lifetime is forever (in memory) rather than being tied to a specific component. It's created when the React DevTools panel renders for the first time (and tries to fetch annotations). Both of those are intentional in this case, since a user might switch away to the Viewer tab (or some other Devtools tab) and come back.

Andarist: ah, ok - so the signal here is handled because it's the right thing to do but this one won't actually ever be aborted because that requires an explicit `.abort(...params)` call. got it 👍

bvaughn: Yeah. This specific one doesn't receive an abort bc it's only used in a suspense boundary so it will go away on its own once the annotations data has loaded. The only things further down in the tree that can suspend (inspected elements panel) have lower-level suspend boundaries that catch them.

That being said, we could try doing an explicit abort here. It's a little weird bc the two sites are separated a bit, but if we wanted to we could probably use the `useCacheStatus` hook to listen for annotation loading status and then abort in an effect.

Andarist: yeeeah, I wouldn't bother since this is just a timeout and we can safely ignore it switching the state after the annotations already resolved (and resumed the suspended tree)

bvaughn:
You know, another approach we could have taken here would be to have a suspend fallback boundary that had a built-in timer itself (maybe one that could be configured by props) that did the "Loading..." -> "This is taking longer than we expected" transition

PR #10549, opened 2024-05-30
Typing Escape after adding a new comment should cancel the comment

bvaughn:
### Note I am on PTO until Monday the 3rd; feel free to merge this without me if this is approved.

---

This interaction has nagged me for a while also. @jcmorrow recently pointed it out after dogfooding so I took the opportunity to fix it. Rules are:
* Typing Escape into a new (not yet saved) comment will delete it.
* Typing Escape into an existing (already saved) comment will restore the previous content.
* Erasing all of a comment's text and typing Enter will delete it.

I can imagine us adding e2e tests for this functionality, but I don't have time to do this today due to planning meetings.


callingmedic911:
I merged `main` to run the test suite. We were debugging the cause of missing tests in the dashboard and broke the comment bot.

Andarist (approved):

callingmedic911 (approved):

PR #10550, opened 2024-05-30
[TT-189] Add platform-specific block lists and a bit more  logic

toshok:
* switch from black/white to block/force since it's a little clearer what they do.
* add output for when a test is blocked but wouldn't be run anyway, or forced when it would be run anyway.
* add platform specific block/force lists (so I can block the long-time macos failing `tests/inspector-elements-02_node-picker.test.ts` - added this to TT-189.)

Honestly I think we should probably move away from a block list and more to an expected-failure/XFAIL list.   Then we can see when it starts passing in CI.  for sometime soon.

### DIFF
--- a/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
+++ b/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
@@ -20,48 +20,76 @@ const CONFIG = {
 let TestFileOverrideList = [];
 
 // Disable some tests that we know to be problematic.
-const TestFileBlackList = new Set([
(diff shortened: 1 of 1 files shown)


jazzdan (approved):

Domiii:
I think this got the wrong linkage. Would be good to have a Linear issue for this patch. But it certainly is not TT-189.

toshok:
while there could be a linear issue re: platform blocklists (I'm not sure there needs to be?), TT-189 is "re-enable FE tests that are blocked" which this PR also added.

PR #10554, opened 2024-06-08
Use graphics element, not container element for resize listener

jcmorrow:
The best explanation is this loom:

https://www.loom.com/share/a123d237c47b476db00bc79a0aee0bb8

### DIFF
--- a/src/ui/components/Video/RecordedCursor.tsx
+++ b/src/ui/components/Video/RecordedCursor.tsx
@@ -27,14 +27,14 @@ export function RecordedCursor() {
           const originalHeight = height / localScale;
           const originalWidth = width / localScale;
 
-          const mouseX = (mouseEvent.clientX / originalWidth) * width;
-          const mouseY = (mouseEvent.clientY / originalHeight) * height;
+          const mouseX = (mouseEvent.clientX / originalWidth) * 100;
+          const mouseY = (mouseEvent.clientY / originalHeight) * 100;
 
           const cursorScale = Math.min(1, Math.max(0.25, localScale * recordingScale));
 
           element.style.display = "block";
-          element.style.left = `${mouseX}px`;
-          element.style.top = `${mouseY}px`;
+          element.style.left = `${mouseX}%`;
+          element.style.top = `${mouseY}%`;
           element.style.transform = `scale(${cursorScale})`;
           element.style.setProperty("--click-display", shouldDrawClick ? "block" : "none");
         } else {
--- a/src/ui/components/Video/imperative/subscribeToMutableSources.ts
+++ b/src/ui/components/Video/imperative/subscribeToMutableSources.ts
@@ -35,7 +35,7 @@ export function subscribeToMutableSources({
     });
 
     // When the graphics element resizes, the mutable state should recalculate the its layout
-    updateState(containerElement, { didResize: true });
+    updateState(graphicsElement, { didResize: true });
   });
 
   resizeObserver.observe(containerElement);


markerikson (approved):

bvaughn:
When reviewing something like this, I would find it really helpful if there was a link to a Replay or something that demonstrated the broken behavior. I am unable to reproduce the misaligned behavior on production, looking at test suite recordings _and_ manual recordings.

jcmorrow:
> When reviewing something like this, I would find it really helpful if there was a link to a Replay or something that demonstrated the broken behavior. I am unable to reproduce the misaligned behavior on production, looking at test suite recordings _and_ manual recordings.
> 
> **Update**: I scrubbed through the Loom until I was able to see enough of the URL to copy the recording id down (67c3b2ef-9077-4b06-b0f4-55f5a3e0d2a2). Comparing the behavior of this branch to production for that recording still doesn't show any differences to me though.
> 
> Here's a question for @jcmorrow: [loom.com/share/9c2e935245b74eb99e29d73588fd409f](https://www.loom.com/share/9c2e935245b74eb99e29d73588fd409f)

https://app.replay.io/recording/cursor-bug-w-zoom--eaafaf12-1704-45c5-a3ce-32511175f6d3?commentId=YzpkYzU5MmU2NC03OThkLTQ1MmMtOTQ3YS04NmI4YjU3MzI5ODA%3D&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiMTA1NzkzMDQ4NTA1MzI5NjAyMjc1OTY0NTg5MTM5MDM0MTEyIiwidGltZSI6NTA2ODd9fQ%253D%253D&point=83725786855250810169943861598617600&primaryPanel=comments&secondaryPanel=console&time=37751.12093023256&viewMode=non-dev

As I say in the comments, I don't know if zoom is required or not, it seems like it might be?

bvaughn:
> As I say in the comments, I don't know if zoom is required or not, it seems like it might be?

Hmm, interesting. FWIW zooming in or out to the max amount doesn't seem to change the observed behavior for me. (Production looks fine without this change.)

What's your `devicePixelRatio` btw? I've tested this on DPR 1 and 2.

jcmorrow:
> What's your devicePixelRatio btw? I've tested this on DPR 1 and 2.

Mine is 2.

I understand why you are concerned that you can't repro the bug on your machine. That said - it clearly *is* a bug, and this change fixes it and also makes sense AFAICT (we should be scaling all graphics to the inner container, not the outer one).

So, should this question be considered blocking? If not I'll merge this.

bvaughn:
Hm. No, I'm not going to block this from landing.

I'm just a little uneasy because I don't understand why the fix is necessary :)

jcmorrow:
> Hm. No, I'm not going to block this from landing.
> 
> I'm just a little uneasy because I don't understand why the fix is necessary :)
> 
> In the past when that has happened, it sometimes might point to something unexpected (like a browser extension that caused unexpected behavior) so I'd feel better about it if I understood it that's all :)

Yep, that all makes sense. I think that I do understand at least one key piece of the puzzle, let's talk about it when we chat in about 10 minutes, and I'll merge this after that.

jcmorrow:
Yep, I figured out the crucial piece to repro: it happens when the graphics container scales beyond the *original* size of the graphics. It should be easier to repro with this recording: https://app.replay.io/recording/smol-graphics--607bcbad-151f-45bb-bda9-9678fa38825e?commentId=&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiNDUxMDgwNzg5NzU0OTM3NTYxMTA4NjYyMTQ0NzgxNTE2ODAiLCJ0aW1lIjoxNzUxNn19&point=26286002855529899327030372442573536&primaryPanel=events&secondaryPanel=console&time=10469&viewMode=dev

Because I intentionally made that recording window really small.

bvaughn:
Oh wow! Okay that was not something I considered trying 🤪

bvaughn (approved):

PR #10556, opened 2024-06-10
Look up DOM nodes for Cypress "click" steps at the before point

markerikson:
This PR:

- Updates the Cypress-specific logic in `TestEventDetailsCache.ts` to do additional lookups when the test step is a `"click"` event, by looking for the corresponding DOM node at the `beforeStep`.  This ensures the DOM node _should_ still be in the DOM and we have something to highlight.

The logic is ugly, but in general;

- We have DOM node previews from the `result` step. These include a persistent ID
- We do an eval at the `beforeStep` point, and pass in all of the target node persistent IDs
- We use a modified version of the DOM tree walking logic used by the Elements panel, and loop over the whole tree looking for a DOM node with the same persistent ID. When we find it, we return the _per-pause_ object ID.
- We look up the actual DOM node previews from that pause now that we know the object IDs
- We cache the "before" versions using the "_result_" point as the key, which is fine - that's what the row component uses as the lookup, but what really matters is that we've saved a pause ID and DOM node IDs, and the pause we saved for highlighting doesn't have to be a pause for the result point.

### DIFF
--- a/src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
@@ -13,6 +13,7 @@ import { assert } from "protocol/utils";
 import { Element, elementCache } from "replay-next/components/elements/suspense/ElementCache";
 import { createFocusIntervalCacheForExecutionPoints } from "replay-next/src/suspense/FocusIntervalCache";
 import { objectCache } from "replay-next/src/suspense/ObjectPreviews";
+import { pauseEvaluationsCache } from "replay-next/src/suspense/PauseCache";
 import {
   cachePauseData,
   pauseIdCache,
@@ -190,7 +191,11 @@ interface FindDOMNodeObjectIdForPersistentIdResult {
 }
 
 const domNodeIdForPersistentIdCache: Cache<
-  [replayClient: ReplayClientInterface, point: ExecutionPoint, persistentIds: (string | number)[]],
+  [
+    replayClient: ReplayClientInterface,
+    point: TimeStampedPoint,
+    persistentIds: (string | number)[]
+  ],
   { pauseId: string; domNodeIds: FindDOMNodeObjectIdForPersistentIdResult[] }
 > = createCache({
   debugLabel: "DomNodeIdForPersistentIdCache",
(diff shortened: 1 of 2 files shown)


src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -346,7 +436,59 @@ async function fetchAndCachePossibleCypressDomNode(
...
+        objectCache.readAsync(replayClient, beforeStepPauseId, domNodeId, "canOverflow")
+      )
+    );
+
+    // Note that in this case, we'll have the DOM nodes and pause ID from the before step,
+    // but we're going to put them in the cache with the result point as the key.
+    // This is okay! The test step row consistently uses the result point as the key,
+    // and this _is_ the DOM node and pause ID that relate to the step.
+    // `highlightNodes()` just needs these values, so it's fine if they're at a different
+    // point in time than the step itself.
```
hbenl, line 478: Does this mean we'll show the screenshot for the `result` point with a highlighter for the `beforeStep` point? We shouldn't be doing this, after all we're adding all this logic because the node may not be part of the DOM at the `result` point anymore, so we may be highlighting a node that doesn't exist in the screenshot. We should show the screenshot for the `beforeStep` point when hovering over click events instead.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -346,7 +436,59 @@ async function fetchAndCachePossibleCypressDomNode(
...
+    possibleDomNodes,
+    pauseId,
+  };
+
+  // [PRO-599] If this is a click event, we need to find the DOM node as it existed
+  // _before_ the step
+  if (testEvent.data.command.name === "click") {
+    // If this is a click event, we need to find the DOM node as it
+    // existed _before_ the step, as it may have been disconnected
+    // during the step handling.
```
hbenl, line 449: Duplicated comment

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -171,6 +184,83 @@ export const testEventDomNodeCache: ExternallyManagedCache<
...
+    }
+
+    const evalResults: RunEvaluationResult[] = [];
+
+    // The runtime expects these to be numbers, so we need to convert them.
+    const persistentIdsAsNumbers = persistentIds.map(id => Number(id));
+
+    const sources = await sourcesByIdCache.readAsync(replayClient);
+
+    await replayClient.runEvaluation(
```
hbenl, line 210: Any particular reason why you're not using a regular evaluation here?

bvaughn: Agreed. I think it might be easier to just use the `pauseEvaluationsCache` for this (with a Pause id for the `beforeStep` execution point)

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
bvaughn: > Does this mean we'll show the screenshot for the `result` point with a highlighter for the `beforeStep` point?

I don't _think_ so. I think that execution point is just used as a _key_ to the cached data, but it doesn't actually _seek_ or update the timeline. It just tries to highlight DOM elements with the same id:

https://github.com/replayio/devtools/blob/96a68a9a1d104658d3078b65f18dd58709baefbc/src/ui/components/TestSuite/views/TestRecording/TestSectionRow.tsx#L148-L175

Which...come to think of it, since ids aren't globally unique outside of a pause, that seems like a potential problem.

src/ui/components/TestSuite/utils/cypressStepDomNodes.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  function getObjectId(value: any) {
+    if (typeof __RECORD_REPLAY_ARGUMENTS__ !== "undefined" && __RECORD_REPLAY_ARGUMENTS__ != null) {
+      // TODO [FE-2005][FE-2067] With persistent DOM ids, presumably we should switch to using this API?
+      // if (__RECORD_REPLAY_ARGUMENTS__.getPersistentId) {
+      //   const id = __RECORD_REPLAY_ARGUMENTS__.getPersistentId(value);
+      //   if (id) {
+      //     return id;
+      //   }
+      // }
```
bvaughn, line 43: This comment isn't relevant here, I think. We already have a handle on the DOM element (and its persistent id). That's what we're using to match up with the regular `ObjectId`

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: The Playwright logic below uses `runEvaluation`, so I was being consistent.

Thinking about it, the Playwright code relies on `fullPropertyPreview: true` to get additional data being returned, but in this case I'm serializing the data as a string, so a plain evaluation would work okay.

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
markerikson: The details cache provides both the pause ID and the DOM node IDs, so there shouldn't be any collisions - we're targeting that specific pause for the highlighting.

markerikson:
All comments addressed

src/ui/components/TestSuite/suspense/TestEventDetailsCache.ts
```diff
@@ -171,6 +185,64 @@ export const testEventDomNodeCache: ExternallyManagedCache<
...
+      // Implicit return
+      stringifiedResults;
+    `;
+
+    const pauseId = await pauseIdCache.readAsync(replayClient, point.point, point.time);
+    const result = await pauseEvaluationsCache.readAsync(replayClient, pauseId, null, expression);
+
+    const { returned } = result;
+    assert(returned?.value, "Expected a value to be returned from the eval.");
+    assert(typeof returned.value === "string", "Expected the value to be a string.");
```
bvaughn, line 235: Oh I missed this the first time through but it's a nit so you can skip it if you don't think it's worthwhile.
```suggestion
    assert(typeof returned?.value === "string", "Expected a string to be returned from the eval.");
```

src/ui/components/TestSuite/utils/cypressStepDomNodes.ts
```diff
@@ -0,0 +1,111 @@
...
+  // especially since Cypress tests always run in an iframe.
+  let objectId: string | null = null;
+  let queue: (ChildNode | Document | Element)[] = [document];
+
+  // This loop modified from the version in `serialization.ts`,
+  // which is used to serialize the entire DOM tree.
+  // In our case, we want to _loop_ over the entire DOM tree,
+  // but we're only looking to find a single DOM node by persistent ID.
+  // It's important that we be able to drill down through iframes,
+  // especially since Cypress tests always run in an iframe.
```
bvaughn, line 65: Duplicate comment
```suggestion
```

bvaughn (approved):
Approve to unblock landing but with two small requested changes

PR #10557, opened 2024-06-12
Improve message shown when async stack can't be loaded

bvaughn:
This message is shown all the time when it shouldn't be (because there's no focus window):
Screenshot 2024-06-12 at 8 35 22 AM

This PR changes it to (a) not show it when there's no focus window
Screenshot 2024-06-12 at 8 35 14 AM

and (b) if we do show it, make the wording less certain (since it's often wrong)
Screenshot 2024-06-12 at 8 35 00 AM

### DIFF
--- a/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
+++ b/src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
@@ -10,9 +10,11 @@ import {
   getSelectedFrameId,
   getThreadContext,
 } from "devtools/client/debugger/src/selectors";
+import { isFocusWindowApplied } from "devtools/client/debugger/src/utils/focus";
 import { InlineErrorBoundary } from "replay-next/components/errors/InlineErrorBoundary";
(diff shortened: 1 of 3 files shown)


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -44,12 +46,9 @@ function FramesRenderer({
   const replayClient = useContext(ReplayClientContext);
   const sourcesState = useAppSelector(state => state.sources);
   const { rangeForSuspense: focusWindow } = useContext(FocusContext);
+  const { endpoint } = useContext(SessionContext);
   const dispatch = useAppDispatch();
 
-  if (focusWindow === null) {
-    return null;
-  }
```
bvaughn, line 51: I think this check was in place just to make TypeScript happy, but the `getAsyncParentPauseIdSuspense` function _already_ had a check for `focusWindow === null` so I just updated its type to accept a null param.

(If this ever _did_ return `null`, it hid the stack entirely which isn't what we'd want anyway.)

hbenl (approved):

PR #10561, opened 2024-06-14
Jump-to-source falls back to parent user event if marker frame cannot be found

bvaughn:
### [Loom overview](https://www.loom.com/share/6d697c9136274fb8892325d39283fdac)

I don't know if this is a good idea or a bad one because I'm not super familiar with the Cypress code or constraints.

Looking at [go/r/bb82f5c3-af5c-4e20-9f1c-5efa4ee6bd29](https://app.replay.io/recording/appreplayio--bb82f5c3-af5c-4e20-9f1c-5efa4ee6bd29), I observed that the `getCypressMarkerFrame` lookup fails to find a "marker frame" because there are only two sources in the call stack– the Replay plugin itself and Cypress. That method expects to find user code in between for some reason, so this breaks the jump-to-source behavior.

In this case, the source is:
```js
cy.contains('Product added to cart!').should('be.visible');
```

That gets rendered in the UI as this:

Screenshot 2024-06-13 at 10 46 16 PM

Clicking on the failed _assert_ (or trying to "jump to source") would ideally jump to the `should('be.visible')` but it fails because of the "marker frame" issue I mentioned above. The idea occurred to me that in this case, jumping to the parent command (`contains('Product added to cart!')`) is probably better than nothing. (In this case, they're even on the same line, but that won't always be true.)

I wish the architecture for this stuff was different. I wish we did this kind of processing in a routine so we could detect cases like this ahead of time and not show "jump to source" buttons that aren't going to work, but I don't even really have an idea of the effort that would be required to rearchitect some of this stuff at the moment.)

### DIFF
--- a/src/ui/components/TestSuite/hooks/useJumpToSource.ts
+++ b/src/ui/components/TestSuite/hooks/useJumpToSource.ts
@@ -19,11 +19,13 @@ import { AwaitTimeout, awaitWithTimeout } from "ui/utils/awaitWithTimeout";
 export function useJumpToSource({
   groupedTestCases,
   testEvent,
+  testEvents,
   testRecording,
   openSourceAutomatically = false,
 }: {
   groupedTestCases: GroupedTestCases;
   testEvent: TestEvent;
+  testEvents: TestEvent[];
   testRecording: TestRecording;
   openSourceAutomatically: boolean;
 }) {
@@ -57,7 +59,8 @@ export function useJumpToSource({
       const locationPromise = TestStepSourceLocationCache.readAsync(
         replayClient,
         groupedTestCases,
-        testEvent
+        testEvent,
+        testEvents
       );
 
       let location;
--- a/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
+++ b/src/ui/components/TestSuite/suspense/TestStepSourceLocationCache.ts
@@ -1,12 +1,18 @@
 import assert from "assert";
 import { Frame, Location } from "@replayio/protocol";
 import { compare } from "compare-versions";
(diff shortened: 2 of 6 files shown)


filiphric:
I think the decision to go into parent command is great. There’s a concept of parent/child/hybrid commands in Cypress. For child commands, you will always have a parent command, so in a way, jumping to the parent command oftentimes makes more sense than jumping to the child command itself. 

Couple of things that might be good to know, or simply just add more context:

1. chained commands are not always on the same line. you could have a situation like this:

```js
cy.contains('Add to cart')
  .should('be.visible')
 ```
 
 In fact, this is how I mostly write my commands, but it’s mostly for visibility reasons when I do workshops or tutorials.
 
 2. There’s a case, which I don’t see too often, where you have a custom function or a custom command returning (yielding) some context. in this situation, you might have `.should()` and the thing it is asserting on in different files.
 
 ```ts
 // findbutton.ts
 export const findButton = () => {
   return cy.contains('Add to cart')
 }
 ```
 
 ```ts
 // spec.cy.ts
 import findButton from '@commands/findbutton.ts'
 
 it('tests stuff', () => {
   cy.visit('/')
   findbutton().should('be.visible')
 })
 ```
 
I’m assuming that with this fix, we’d jump to `findbutton.ts` instead of `spec.cy.ts`. But I don’t think this would be a problem, I think it’s better. Unless someone tells me otherwise, I’m pretty sure it makes more sense.

filiphric (approved):

hbenl (approved):

bvaughn:
Thanks for the detailed response, Filip. Much appreciated.

PR #10567, opened 2024-06-14
Support `redux-promise` package

bvaughn:
Not sure if this is something we should do or not. Sharing with @markerikson.

Rambly video [here](https://www.loom.com/share/e7ce5558d1584e0cb6088149e129bb36).

### DIFF
--- a/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsListItem.module.css
+++ b/src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsListItem.module.css
@@ -1,6 +1,7 @@
 .ListItem {
   line-height: 1.5rem;
   padding: 0 1ch;
+  white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
   border-bottom: 1px solid var(--listitem-row-divider);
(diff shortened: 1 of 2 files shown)


src/ui/components/SecondaryToolbox/redux-devtools/ReduxDevToolsListItem.module.css
```diff
@@ -1,6 +1,7 @@
 .ListItem {
   line-height: 1.5rem;
   padding: 0 1ch;
+  white-space: nowrap;
```
bvaughn, line 4: This is not technically related but I noticed it while I was in there.
### Before
Screenshot 2024-06-14 at 2 30 47 PM
### After
Screenshot 2024-06-14 at 2 31 04 PM

markerikson:
I see this PR and the associated issue, but don't have the brainpower to think through the implications atm.  Let me look it over on Monday?

bvaughn:
Oh for sure. No hurry.

markerikson:
Read this PR and watched the Loom.   The explanation is really helpful, thanks!

I think this is ultimately a too-narrow approach, and _if_ we want to make changes here it should be done to handle the general case of "there are 2+ dispatches in a single synchronous call stack, jump to the appropriate one even if it's in a middleware", rather than something that targets the `redux-promise` library specifically.

Wrote up a longer explanation in an `#eng-product` thread in Discord.

bvaughn:
Makes sense!

PR #10568, opened 2024-06-14
Log point panel should never show 1/0

bvaughn:
Screenshot 2024-06-14 at 3 57 18 PM

### DIFF
--- a/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
+++ b/packages/replay-next/components/sources/log-point-panel/Capsule.tsx
@@ -116,7 +116,7 @@ export default function Capsule({
               defaultValue={closestHitPointIndex + 1}
               disableSelectionWhenNotFocused
               editable={!tooManyPointsToFind}
-              maxValue={Math.max(1, hitPoints.length)}
+              maxValue={hitPoints.length}
               minValue={1}
               onSave={onSave}
               ref={editorRef}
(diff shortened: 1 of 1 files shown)


packages/replay-next/components/sources/log-point-panel/Capsule.tsx
```diff
@@ -104,20 +104,24 @@ export default function Capsule({
...
+            <NumberEditor
+              className={styles.CurrentIndex}
+              data-exact={currentHitPoint !== null || undefined}
+              data-test-name="LogPointCurrentStepInput"
+              data-too-many-points-to-find={tooManyPointsToFind || undefined}
+              data-value={closestHitPointIndex + 1}
+              defaultValue={closestHitPointIndex + 1}
+              disableSelectionWhenNotFocused
+              editable={!tooManyPointsToFind}
+              maxValue={Math.max(1, hitPoints.length)}
```
Andarist, line 119: do we have to keep this `Math.max` here? 🤔

Andarist (approved):

packages/replay-next/components/sources/log-point-panel/Capsule.tsx
bvaughn: Nah, it would be okay to remove it now. I had left it there because it wasn't hurting anything but I don't mind.

PR #10573, opened 2024-06-18
Fix detection of async parent pauses outside of the focus window

hbenl:
#9368 broke the detection: before that PR `getAsyncParentPauseIdSuspense()` would return `undefined` if there is no async parent pause or `null` if it is outside the focus window, after that PR it always returned `null`.
With this PR `getAsyncParentPauseIdSuspense()` will return a `boolean` in these cases instead of abusing `null`/`undefined`, which obviously caused confusion leading to a regression.


Andarist:
How hard would it be to add tests for both scenarios? I'm not familiar with the test setup here so I'm just genuinely curious how much effort is required to add new tests

hbenl:
> How hard would it be to add tests for both scenarios?

That shouldn't be too hard and we should get back into the habit of writing tests so I'll create one.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -64,30 +65,25 @@ function FramesRenderer({
...
-                your debugging window
-              </span>
-              .
-            </>
-          )}
+          This part of the call stack is unavailable because it is outside{" "}
+          <span className="cursor-pointer underline" onClick={() => dispatch(enterFocusMode())}>
+            your debugging window
+          </span>
+          .
```
bvaughn, line 77: I think we need to re-add the messaging changes from #9368. We should never show this message if the focus window is the full recording.

src/ui/suspense/util.ts
```diff
@@ -6,36 +6,36 @@ import { pauseIdCache } from "replay-next/src/suspense/PauseCache";
...
     }
 
     const executionPoint = steps[0].point;
     if (
       focusWindow === null ||
       !isExecutionPointsWithinRange(executionPoint, focusWindow.begin.point, focusWindow.end.point)
     ) {
-      return null;
+      return true;
     }
```
bvaughn, line 34: Returning `true` here if `focusWindow === null` seems funky?

I don't think `focusWindow` window is ever `null` (it's either the full recording or a subset) _but_ if it were null, that would be _the same as_ the full recording, wouldn't it? In which case, returning `true` here seems like it would always be misleading.

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
hbenl: > Unless you're 100% confident that we'll never have a false positive here

Yes, I'm confident (especially after changing it to *not* return true if `focusWindow` is `null`, as suggested below).

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/NewFrames.tsx
```diff
@@ -46,13 +44,16 @@ function FramesRenderer({
   const replayClient = useContext(ReplayClientContext);
   const sourcesState = useAppSelector(state => state.sources);
   const { rangeForSuspense: focusWindow } = useContext(FocusContext);
-  const { endpoint } = useContext(SessionContext);
   const dispatch = useAppDispatch();
 
+  if (focusWindow === null) {
+    return null;
+  }
```
bvaughn, line 51: Returning `null` here still feels odd. I guess this will only ever happy briefly during initialization so maybe it doesn't matter.

bvaughn (approved):
I dig the new test!

PR #10576, opened 2024-06-18
Improve heuristics for finding the user frame that dispatched a Redux…

markerikson:
This PR:

- Reworks the Redux Jump-to-Code feature to improve the heuristics we use to find the right userland stack frame to jump to.
- Updates to a new `breakpoints-01` recording just so we've got a more recent example
  - Updates the expectations in the `react_devtools-02` test to match the new `breakpoints-01` recording
- Extracts a reusable test util for verifying that a given J2C button click jumped to the right file and time
- Adds a new E2E test that exercises the Redux J2C behavior by checking that several different dispatched actions end up on the right lines / stack frames

The original logic was iterating backwards across the point stack frames looking for an applyMiddleware frame, and going backwards one more frame from there. This is problematic because I'm seeing variance in whether or not applyMiddleware actually shows up in stack frames. For some reason it shows up in Metabase recording stack traces, but not in our own E2E tests (and we're both using RTK).  That meant that the existing logic could keep right on walking backwards long past the frame it "should" have found.

What I've settled on is using the logic we had that tries to identify "is this function in a Redux middleware" based on source outlines (looking for the triply-nested function signature of a middleware definition), and stopping at the first frame that appears to be not a middleware and not applyMiddleware.  This seems to be producing pretty reasonable results.

As an example, try opening https://app.replay.io/recording/breakpoints-01-test-basic-breakpoint-functionality--4f74bac8-d3f5-4e17-aeca-d670e002eabf and then the same recording in this preview, and see how the PR version more consistently lands on the `dispatch()` call in app code rather than inside of `applyMiddleware`. Same for a Metabase recording.

### DIFF
--- a/packages/e2e-tests/helpers/source-panel.ts
+++ b/packages/e2e-tests/helpers/source-panel.ts
@@ -442,6 +442,35 @@ export async function getSelectedLineNumber(
   return parseInt(textContent, 10);
 }
 
+export async function verifyJumpToCodeResults(
+  page: Page,
+  filename: string,
+  lineNumber: number,
+  expectedHits?: { current: number; total: number }
+) {
+  await waitForSelectedSource(page, filename);
+  // Should highlight the line that ran
+  await waitFor(async () => {
+    const lineNumber = await getSelectedLineNumber(page, true);
+    expect(lineNumber).toBe(lineNumber);
+  });
+
+  if (expectedHits) {
+    // Should also have jumped in time. Since this can vary (slightly different progress %
+    // based on timing differences), we'll add a log statement and verify _which_ hit we're at.
+    await addLogpoint(page, {
+      url: filename,
+      lineNumber,
+    });
+
+    const { current, total } = expectedHits;
+
+    // Should have paused on the handler for the first valid keystroke
+    await verifyLogpointStep(page, `${current}/${total}`, { url: filename, lineNumber });
+    await removeLogPoint(page, { url: filename, lineNumber });
+  }
+}
+
 export function getSourceLocator(page: Page, sourceId: string): Locator {
   return page.locator(getSourceSelector(sourceId));
 }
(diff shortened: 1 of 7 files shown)


bvaughn:
> Adds a new E2E test that exercises the Redux J2C behavior by checking that several different dispatched actions end up on the right lines / stack frames

Doesn't look like you pushed all of the code described in the PR summary.

markerikson:
Yeah, push failed because I'd reset the branch tip.  Hopefully that should work this time!

packages/e2e-tests/helpers/source-panel.ts
```diff
@@ -442,6 +442,35 @@ export async function getSelectedLineNumber(
...
+    // based on timing differences), we'll add a log statement and verify _which_ hit we're at.
+    await addLogpoint(page, {
+      url: filename,
+      lineNumber,
+    });
+
+    const { current, total } = expectedHits;
+
+    // Should have paused on the handler for the first valid keystroke
+    await verifyLogpointStep(page, `${current}/${total}`, { url: filename, lineNumber });
```
bvaughn, line 469: These values are optional according to your TS types, so we should optionally verify them (or change the type to make them required)

bvaughn (approved):
Spot checked a few things on this branch vs prod and this does look better!

packages/e2e-tests/helpers/source-panel.ts
markerikson: Clarify what you're saying? I've got an `if (expectedHits)` check just above this.

markerikson:
Looks like updating to a newer `breakpoints-01` recording broke the RDT test, looking into it

packages/e2e-tests/helpers/source-panel.ts
bvaughn: I somehow missed that. Sorry.

markerikson:
Comment bot is a bit slow here, but the PR run is green!

PR #10577, opened 2024-06-19
Remove LogRocket from Devtools

bvaughn:
- Uninstall dependencies
- Remove "logrocket" library usage
- Delete LogRocket privacy setting
- Remove LogRocket-specific `data-private` attributers
- Remove `Redacted` component (and styles)


pages/_document.tsx
```diff
@@ -26,24 +26,21 @@ const csp = (props: any) => {
...
+    `script-src 'self' 'unsafe-eval' https://cdn.lr-ingest.io https://cdn.lr-in.com https://js.stripe.com ${hash}`,
     `form-action https://${authHost}`,
 
     // From vercel's CSP config and Google fonts
     `font-src 'self' data: https://fonts.gstatic.com`,
     // Google fonts
     `style-src-elem 'self' 'unsafe-inline' https://fonts.gstatic.com`,
-    // Required by LogRocket
-    `child-src 'self' blob:`,
-    `worker-src 'self' blob:`,
```
bvaughn, line 46: I assume LogRocket was the only thing needing these settings? Do we know?

markerikson: I've never dealt with CSP concepts before, but the LogRocket docs do specifically mention needing to add these:

https://docs.logrocket.com/v1.0/docs/troubleshooting-sessions#due-to-content-security-policy

bvaughn: Turns out we need the worker-src one for log point analysis cache.

PR #10582, opened 2024-06-23
Add option to notify user when processing is complete

bvaughn:
- [x] Add option to processing screen to be notified when recording has finished processing
- [x] Manually test a bit more (once I'm off airplane wifi)

### DIFF
--- a/packages/replay-next/src/hooks/useNotification.ts
+++ b/packages/replay-next/src/hooks/useNotification.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useState, useSyncExternalStore } from "react";
 
 export type Permission = NotificationPermission | PermissionState;
 export type RequestPermission = () => Promise<boolean>;
@@ -9,28 +9,24 @@ export function useNotification(): {
   requestPermission: RequestPermission;
   supported: boolean;
 } {
-  const [permission, setPermission] = useState<Permission>(Notification.permission);
+  const permission = useSyncExternalStore(
+    function subscribe(change: () => void) {
+      let permissionStatus: PermissionStatus;
 
-  useEffect(() => {
-    let permissionStatus: PermissionStatus;
+      (async () => {
+        permissionStatus = await navigator.permissions.query({ name: "notifications" });
+        permissionStatus.addEventListener("change", change);
+      })();
 
-    const onChange = () => {
-      if (permissionStatus) {
-        setPermission(permissionStatus.state);
-      }
-    };
-
-    (async () => {
-      permissionStatus = await navigator.permissions.query({ name: "notifications" });
-      permissionStatus.addEventListener("change", onChange);
-    })();
-
-    return () => {
-      if (permissionStatus) {
-        permissionStatus.removeEventListener("change", onChange);
-      }
-    };
-  }, []);
+      return () => {
+        if (permissionStatus) {
+          permissionStatus.removeEventListener("change", change);
+        }
+      };
+    },
+    () => Notification.permission,
+    () => Notification.permission
+  );
 
   const [requested, setRequested] = useState(false);
 
--- a/packages/replay-next/src/hooks/useNotification.ts
+++ b/packages/replay-next/src/hooks/useNotification.ts
@@ -1,0 +1,60 @@
+import { useCallback, useEffect, useState } from "react";
+
+export type Permission = NotificationPermission | PermissionState;
+export type RequestPermission = () => Promise<boolean>;
+
+export function useNotification(): {
+  permission: Permission;
+  requested: boolean;
+  requestPermission: RequestPermission;
+  supported: boolean;
+} {
+  const [permission, setPermission] = useState<Permission>(Notification.permission);
+
+  useEffect(() => {
+    let permissionStatus: PermissionStatus;
+
+    const onChange = () => {
+      if (permissionStatus) {
+        setPermission(permissionStatus.state);
+      }
+    };
+
+    (async () => {
+      permissionStatus = await navigator.permissions.query({ name: "notifications" });
+      permissionStatus.addEventListener("change", onChange);
+    })();
+
+    return () => {
+      if (permissionStatus) {
+        permissionStatus.removeEventListener("change", onChange);
+      }
+    };
+  }, []);
+
+  const [requested, setRequested] = useState(false);
+
+  const requestPermission = useCallback(async () => {
+    if (!supported) {
+      return false;
+    }
+
+    setRequested(true);
+
+    let permission = Notification.permission;
+    if (permission !== "granted") {
+      permission = await Notification.requestPermission();
+    }
+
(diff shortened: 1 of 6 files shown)


packages/replay-next/src/hooks/useNotification.ts
```diff
@@ -0,0 +1,60 @@
...
+    (async () => {
+      permissionStatus = await navigator.permissions.query({ name: "notifications" });
+      permissionStatus.addEventListener("change", onChange);
+    })();
+
+    return () => {
+      if (permissionStatus) {
+        permissionStatus.removeEventListener("change", onChange);
+      }
+    };
```
Andarist, line 32: There is a small race condition here. It's possible for the cleanup of this effect to be called before the query resolves. In such a scenario, we'd still call `addEventListener` and never clean it up.

I'd introduce a `disposed` variable, set it in the cleanup and check in that `async` function to handle this scenario

packages/replay-next/src/hooks/useNotification.ts
```diff
@@ -0,0 +1,60 @@
...
+export type Permission = NotificationPermission | PermissionState;
+export type RequestPermission = () => Promise<boolean>;
+
+export function useNotification(): {
+  permission: Permission;
+  requested: boolean;
+  requestPermission: RequestPermission;
+  supported: boolean;
+} {
+  const [permission, setPermission] = useState<Permission>(Notification.permission);
```
Andarist, line 12: this is fine but since technically this is a mutable value maybe we should wrap this in a lazy init to make that more apparent?

```suggestion
  const [permission, setPermission] = useState<Permission>(() => Notification.permission);
```

src/ui/components/DevToolsProcessingScreen.tsx
```diff
@@ -45,5 +81,43 @@ export function DevToolsProcessingScreen() {
     message = `Processing... (${Math.round(processingProgress)}%)`;
   }
 
-  return <LoadingScreen message={message} secondaryMessage={secondaryMessage} />;
+  const onChange = async (event: ChangeEvent<HTMLInputElement>) => {
+    if (event.target.checked) {
+      const granted = await requestNotificationPermission();
+      if (!granted) {
+        event.target.checked = false;
+        event.target.blur();
```
Andarist, line 89: I'm not an a11y expert, so take that with a grain of salt, but I consider `.blur()` to be an anti-pattern. Focus management should be intentional. If we want to change the currently focused element that's fine but to do that we need some kind of a target. IIRC, it's preferred to always have *some* element focused and `.blur()` breaks that rule.

src/ui/components/shared/LoadingScreen.tsx
```diff
@@ -1,86 +1,65 @@
...
+    </>
   );
 }
 
-const connector = connect((state: UIState) => ({
-  uploading: getUploading(state),
-  awaitingSourcemaps: getAwaitingSourcemaps(state),
-}));
-type PropsFromRedux = ConnectedProps<typeof connector>;
+const colorOptions: Array<"blue" | "green" | "red"> = ["blue", "green", "red"];
```
Andarist, line 60: `as const` would help you to keep those literal strings without duplicating them in the type:
```suggestion
const colorOptions = ["blue", "green", "red"] as const;
```

It changes the type to be a readonly tuple though so sometimes u might actually not want to use this.

src/ui/components/DevToolsProcessingScreen.tsx
```diff
@@ -45,5 +81,43 @@ export function DevToolsProcessingScreen() {
...
+                  : undefined
+              }
+            >
+              <input
+                disabled={notificationPermission === "denied"}
+                className={styles.Checkbox}
+                onChange={onChange}
+                type="checkbox"
+              />{" "}
+              Notify me when this tab loads
```
Andarist, line 116: I wonder if people wouldn't expect to get a **push** notification here but we don't support this. Maybe we should somehow warn them that they can't close this tab to get those?

src/ui/components/DevToolsProcessingScreen.tsx
```diff
@@ -30,8 +53,21 @@ export function DevToolsProcessingScreen() {
...
   }, [processingProgress]);
 
+  useEffect(() => {
+    return () => {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      const { permission, permissionsRequested } = notificationPermissionStateRef.current;
+      if (permissionsRequested && permission === "granted" && recording) {
+        const progress = getProcessingProgress(store.getState());
+        if (progress === 100) {
+          new Notification(`"${recording.title}" has loaded`);
```
Andarist, line 63: we could add a `tag` option to avoid creating multiple notifications when the user would have that recording open multiple times in their tabs. The tag could contain the recording's ID and the status so notification related to different recordings wouldn't cancel each other

src/ui/components/DevToolsProcessingScreen.tsx
```diff
@@ -30,8 +53,21 @@ export function DevToolsProcessingScreen() {
     }
   }, [processingProgress]);
 
+  useEffect(() => {
+    return () => {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      const { permission, permissionsRequested } = notificationPermissionStateRef.current;
+      if (permissionsRequested && permission === "granted" && recording) {
+        const progress = getProcessingProgress(store.getState());
+        if (progress === 100) {
```
Andarist, line 62: to avoid creating notifications when the user is looking at our app we could combine this with something like:
```suggestion
        if (progress === 100 && !document.hasFocus()) {
```

Andarist (approved):

packages/replay-next/src/hooks/useNotification.ts
bvaughn: Yeah, this was a little half-assed. I should have been using `useSyncExternalStore` for this. Will fix.

packages/replay-next/src/hooks/useNotification.ts
bvaughn: Using `useSyncExternalStore` will fix this.

src/ui/components/shared/LoadingScreen.tsx
bvaughn: Oh nice. The changes in this file were mostly just cleanup so I dig this follow up.

src/ui/components/DevToolsProcessingScreen.tsx
bvaughn: Great suggestion! I was not aware of the tag API.

src/ui/components/DevToolsProcessingScreen.tsx
bvaughn: I added this initially bc I didn't want focus to be left on a _disabled_ looking input, but since I also disabled the input then this would be unnecessary anyway. At the same time, it's effectively the same thing to my knowledge (the input will be blurred).

PR #10583, opened 2024-06-25
Better view element icon

jonbell-lot23:
[Quick Loom walking through the difference](https://www.loom.com/share/16ba88e4a65545f0845dd02577675b46)

Shout out to @filiphric , who pointed out that the current icon for view element isn't very universal. But the select element is.

**Old:**
image

**New:**
image


filiphric (approved):

jasonLaster:
@jonbell-lot23 this is a relatively small detail, but i _believe_ that the old icon was what both chrome and firefox used for jump to element.

PR #10584, opened 2024-06-26
Enable Asserts

Domiii:
* We need these to fix RUN crashes. We saw 3 crashes on the latest build.

### DIFF
--- a/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
+++ b/packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
@@ -255,6 +255,7 @@ export default async function run_fe_tests(
   }
   // process.env.RECORD_REPLAY_DIRECTORY =
 
+  process.env.REPLAY_ENABLE_ASSERTS = process.env.RECORD_REPLAY_ENABLE_ASSERTS = "1";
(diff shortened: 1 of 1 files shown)


Andarist (approved):

packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
```diff
@@ -255,6 +255,7 @@ export default async function run_fe_tests(
   }
   // process.env.RECORD_REPLAY_DIRECTORY =
 
+  process.env.REPLAY_ENABLE_ASSERTS = process.env.RECORD_REPLAY_ENABLE_ASSERTS = "1";
```
bvaughn, line 258: Do we really need to specify both `REPLAY_ENABLE_ASSERTS` and `RECORD_REPLAY_ENABLE_ASSERTS`? (Is there anything that doesn't support `REPLAY_ENABLE_ASSERTS`?)

bvaughn (approved):

packages/e2e-tests/scripts/buildkite_run_fe_tests.ts
Domiii: We have not done the rename yet. Its right here: https://linear.app/replay/issue/TT-1281/rename-record-replay-enable-asserts-env-variable

PR #10586, opened 2024-06-26
Handle lack of `point.frame`

Andarist:
I don't know how to repro this but types confirm that this shouldn't lead to crashes

### DIFF
--- a/src/ui/components/ProtocolViewer/suspense/recordedProtocolMessagesCache.ts
+++ b/src/ui/components/ProtocolViewer/suspense/recordedProtocolMessagesCache.ts
@@ -86,10 +86,9 @@ export const recordedProtocolMessagesCache = createFocusIntervalCacheForExecutio
           return [];
         }
 
-        const preferredLocation = getPreferredLocation(
-          sourcesState,
-          pointDescriptions[0].frame ?? []
-        );
+        const preferredLocation =
+          pointDescriptions[0].frame &&
+          getPreferredLocation(sourcesState, pointDescriptions[0].frame);
 
         if (!preferredLocation) {
           return [];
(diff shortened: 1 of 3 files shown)


src/ui/suspense/frameCache.ts
```diff
@@ -123,7 +123,8 @@ export const formattedPointStackCache: Cache<
     const formattedFrames = pointStack.map(frame => {
       updateMappedLocationForPointStackFrame(sourcesById, frame);
       const functionLocation = getPreferredLocation(sourcesById, [], frame.functionLocation);
-      const executionLocation = getPreferredLocation(sourcesById, [], frame.point?.frame ?? []);
```
Andarist, line 126: `?? []` couldn't work correctly here before because that would just always lead to a failed assert in `getPreferredLocation`

src/ui/components/ProtocolViewer/suspense/recordedProtocolMessagesCache.ts
```diff
@@ -86,10 +86,9 @@ export const recordedProtocolMessagesCache = createFocusIntervalCacheForExecutio
           return [];
         }
 
-        const preferredLocation = getPreferredLocation(
-          sourcesState,
-          pointDescriptions[0].frame ?? []
-        );
+        const preferredLocation =
+          pointDescriptions[0].frame &&
+          getPreferredLocation(sourcesState, pointDescriptions[0].frame);
```
bvaughn, line 91: I wonder if it would be easier/safer for us to update `getPreferredLocation` to support either `MappedLocation | undefined` as the third parameter?

bvaughn (approved):

markerikson (approved):

src/ui/components/ProtocolViewer/suspense/recordedProtocolMessagesCache.ts
bvaughn: I don't have a strong preference. I can imagine us adding a new callsite and not knowing to avoid the `?? []` pattern here is the only reason I asked. Probably fine.

PR #10587, opened 2024-06-26
Improve video cursor display, especially for Cypress tests

markerikson:
This PR:

- Extracts the current logic that determines "is this a click or keyboard test step?" out of `UserActionEventRow.ts` (where it's currently used to determine "_can_ we have a Jump-to-Code result?") and turns it into standalone utils in `RecordedTestMetadata.ts`
- Updates `TestSectionRow.ts` to use the current mouse event (if available) as the hover time for click steps
  - Previously, we were always using the `beforeStep` time as the hover time for user interaction test steps.  However, for a click step, the mouse event occurs in the _middle_ of the step (ie, `mouseEvent.time >= beforeStep.time && mouseEvent.time <= afterStep.time`).  The `RecordedCursor` logic tries to find "the most recent mouse event <= the current display time".  That meant we were finding a mouse event from _before_ this test step, which resulted in an off-by-one style display as you'd hover or click on test steps.
  - Now, if the current test step is a click, we do a search for a mouse click event that occurred _during_ the test step, and use that event's timestamp as the hover time. This should give more consistent results and ensure that the "correct" event is displayed for the click step.
- Updates the `RecordedCursor.ts` logic to be more narrow in what it displays:
  - Limits the possible search range for "most recent mouse event" to use the focus window. Previously, if you selected a Cypress test halfway through a recording, we would find and show the last mouse click in the previous test because that was "the most recent", even though it was conceptually irrelevant for _this_ test. Since we always zoom in focus on the current test, limiting "most recent mouse event" to the current focus window prevents that.
  - Updates the time that we still show the "click" mouse cursor circle to be one frame (16.67ms), instead of 200ms, to narrow down the appearance of when a click happened.

Note that this is just about mouse event _timing_. We've also found that there are bad synthetic `clientX/Y` values ( see [this summary comment from PRO-598](https://linear.app/replay/issue/PRO-598/cypress-fake-mouse-events-are-not-displayed-in-the-video-panel#comment-52439c6e) and the actual issue in [TT-1474](https://linear.app/replay/issue/TT-1474/fix-x-y-coordinates-of-synthetic-mouse-events).  That means that even though this PR _improves_ the behavior significantly, we also need the runtime fix to make the locations be correct.

Some examples from recording https://app.replay.io/recording/cypresstestsuiauthspects--8ebc83cd-ea98-49d4-bbf6-643708ff9482 , and Clicked on test "Should allow a visitor to sign up":

### Jumping to a later test in the recording

Before this PR, an incorrect cursor left over from the previous test in the recording:

image

After, no cursor at the start of the test:

image

#### Step 04, clicking "signup"

Before, still showing the really old cursor prior to this test:

image

After, showing a cursor that is at least at the right timestamp and in the vicinity, although the X/Y values are still wrong:

image

#### Step 30, clicking "Submit"

Before, no click indicator, and it's still an old cursor. It's actually _closer_ to the button, but it's the wrong event:

image

After, showing a click indicator, and it's the _right_ event even if the coords are still off:

image

#### Step 45, hovering `wait @loginuser`

Before, we're showing the "right" event because it's well past the click step, but we keep showing a click indicator for an extended time even though the page is now blank:

image

After, we're also showing the same correct event, but we're no longer showing the click indicator because we're more than 1 frame past the click:

image

### DIFF
--- a/packages/protocol/RecordedEventsCache.ts
+++ b/packages/protocol/RecordedEventsCache.ts
@@ -83,14 +83,25 @@ export const RecordedNavigationEventsCache = createSingleEntryCache<[], Navigati
   },
 });
 
-export function findMostRecentClickEvent(time: number) {
+export function findMostRecentClickEvent(time: number, startTime: number = 0) {
   const events = RecordedClickEventsCache.getValueIfCached() ?? [];
   const index = findIndexLTE(events, { time } as MouseEvent, (a, b) => a.time - b.time);
-  return index >= 0 ? events[index] : null;
+  if (index < 0) {
+    return null;
+  }
+  const event = events[index];
+  return event.time >= startTime ? event : null;
 }
 
-export function findMostRecentMouseEvent(time: number) {
+export function findMostRecentMouseEvent(time: number, startTime: number = 0) {
   const events = RecordedMouseEventsCache.getValueIfCached() ?? [];
   const index = findIndexLTE(events, { time } as MouseEvent, (a, b) => a.time - b.time);
-  return index >= 0 ? events[index] : null;
+  if (index < 0) {
+    return null;
+  }
+  if (index < 0) {
+    return null;
+  }
+  const event = events[index];
+  return event.time >= startTime ? event : null;
 }
--- a/packages/protocol/RecordedEventsCache.ts
+++ b/packages/protocol/RecordedEventsCache.ts
@@ -83,25 +83,14 @@ export const RecordedNavigationEventsCache = createSingleEntryCache<[], Navigati
   },
 });
 
-export function findMostRecentClickEvent(time: number, startTime: number = 0) {
+export function findMostRecentClickEvent(time: number) {
   const events = RecordedClickEventsCache.getValueIfCached() ?? [];
   const index = findIndexLTE(events, { time } as MouseEvent, (a, b) => a.time - b.time);
-  if (index < 0) {
-    return null;
-  }
-  const event = events[index];
-  return event.time >= startTime ? event : null;
+  return index >= 0 ? events[index] : null;
 }
 
-export function findMostRecentMouseEvent(time: number, startTime: number = 0) {
+export function findMostRecentMouseEvent(time: number) {
   const events = RecordedMouseEventsCache.getValueIfCached() ?? [];
   const index = findIndexLTE(events, { time } as MouseEvent, (a, b) => a.time - b.time);
-  if (index < 0) {
-    return null;
-  }
-  if (index < 0) {
-    return null;
-  }
-  const event = events[index];
-  return event.time >= startTime ? event : null;
+  return index >= 0 ? events[index] : null;
 }
--- a/src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
+++ b/src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
@@ -14,6 +14,8 @@ import {
   TestSectionName,
   UserActionEvent,
   isUserActionTestEvent,
+  isUserClickEvent,
+  isUserKeyboardEvent,
 } from "shared/test-suites/RecordingTestMetadata";
 import { jumpToKnownEventListenerHit } from "ui/actions/eventListeners/jumpToCode";
 import { seek } from "ui/actions/timeline";
@@ -94,7 +96,6 @@ export default memo(function UserActionEventRow({
     annotationsStatus === STATUS_RESOLVED ? parsedAnnotations : NO_ANNOTATIONS;
 
   const [canShowJumpToCode, jumpToCodeAnnotation] = findJumpToCodeDetailsIfAvailable(
-    groupedTestCases,
     userActionEvent,
     jumpToCodeAnnotations
   );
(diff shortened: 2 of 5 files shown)


jcmorrow:
> Previously, we were always using the beforeStep time as the hover time for user interaction test steps.

Ah, this makes so much sense. This is an amazing PR description, thank you thank you thank you!

packages/protocol/RecordedEventsCache.ts
```diff
@@ -83,14 +83,25 @@ export const RecordedNavigationEventsCache = createSingleEntryCache<[], Navigati
...
 
-export function findMostRecentMouseEvent(time: number) {
+export function findMostRecentMouseEvent(time: number, startTime: number = 0) {
   const events = RecordedMouseEventsCache.getValueIfCached() ?? [];
   const index = findIndexLTE(events, { time } as MouseEvent, (a, b) => a.time - b.time);
-  return index >= 0 ? events[index] : null;
+  if (index < 0) {
+    return null;
+  }
+  if (index < 0) {
```
hbenl, line 102: Checking this once should be enough.

bvaughn:
> - Limits the possible search range for "most recent mouse event" to use the focus window. Previously, if you selected a Cypress test halfway through a recording, we would find and show the last mouse click in the previous test because that was "the most recent", even though it was conceptually irrelevant for this test. Since we always zoom in focus on the current test, limiting "most recent mouse event" to the current focus window prevents that.

Leaving a note for myself before digging into the code here, but I am concerned about this proposed change. It makes sense for the test suites case (as showing a cursor from a previous test run doesn't mean much) but I think it will break UX for a "real world app" case if you set a focus window and as a result, the mouse cursor disappears (even though the user didn't move the cursor in the time you've focused on).

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -2,12 +2,16 @@ import assert from "assert";
 import { useLayoutEffect, useRef } from "react";
 
 import { findMostRecentClickEvent, findMostRecentMouseEvent } from "protocol/RecordedEventsCache";
+import { useCurrentFocusWindow } from "replay-next/src/hooks/useCurrentFocusWindow";
 import { state } from "ui/components/Video/imperative/MutableGraphicsState";
 
-const CLICK_TIMING_THRESHOLD_MS = 200;
+// One frame seems reasonable for a click to be considered
+// recent enough to be drawn on the screen
+const CLICK_TIMING_THRESHOLD_MS = 16.67;
```
bvaughn, line 10: I think this change makes good sense 👍🏼

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -18,8 +22,8 @@ export function RecordedCursor() {
       ({ currentTime, graphicsRect, localScale, recordingScale }) => {
         const { height, width } = graphicsRect;
 
-        const mouseEvent = findMostRecentMouseEvent(currentTime);
-        const clickEvent = findMostRecentClickEvent(currentTime);
+        const mouseEvent = findMostRecentMouseEvent(currentTime, focusWindow?.begin.time);
+        const clickEvent = findMostRecentClickEvent(currentTime, focusWindow?.begin.time);
```
bvaughn, line 26: I think this change should be reverted because of its impact on all non-Cypress recordings.

Thinking about other ways we might convey the information of "the cursor hasn't moved in a while" that wouldn't _break_ the interaction for non-Cypress cases: Perhaps we could add a threshold (similar to `CLICK_TIMING_THRESHOLD_MS`) of a couple of seconds after which we _dimmed_ the cursor icon? (Not sure if that would be intuitive enough though.)

hbenl (approved):

src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
```diff
@@ -215,65 +216,24 @@ function Badge({
...
+
+  if (timeStampedPoints.beforeStep !== null && timeStampedPoints.afterStep !== null) {
+    canShowJumpToCode = isUserClickEvent(userActionEvent) || isUserKeyboardEvent(userActionEvent);
+    jumpToCodeAnnotation = jumpToCodeAnnotations.find(a =>
+      isExecutionPointsWithinRange(
+        a.point,
+        timeStampedPoints.beforeStep!.point,
+        timeStampedPoints.afterStep!.point
+      )
+    );
```
bvaughn, line 236: I like this change.

bvaughn (changes requested):
I like most of this! I think the `RecordedCursor` changes should not land though.

src/ui/components/Video/RecordedCursor.tsx
markerikson: You've got the better product sense here, but I thought about this while I was working on it and "hide the cursor any time it happened before the focus window" feels like a _potentially_ valid thing to do.

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -2,12 +2,16 @@ import assert from "assert";
...
 import { state } from "ui/components/Video/imperative/MutableGraphicsState";
 
-const CLICK_TIMING_THRESHOLD_MS = 200;
+// One frame seems reasonable for a click to be considered
+// recent enough to be drawn on the screen
+const CLICK_TIMING_THRESHOLD_MS = 16.67;
 
 export function RecordedCursor() {
   const elementRef = useRef<HTMLDivElement>(null);
+  const focusWindow = useCurrentFocusWindow();
```
bvaughn, line 14: We aren't using `focusWindow` anymore so we can remove this hook and the dependency below
```suggestion
```

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -47,7 +51,7 @@ export function RecordedCursor() {
     return () => {
       unsubscribe();
     };
-  }, []);
+  }, [focusWindow]);
```
bvaughn, line 54:
```suggestion
  }, []);
```

src/ui/components/Video/RecordedCursor.tsx
bvaughn: Yeah it's plausible for sure. It's not crazy. (Especially in the context of Cypress.) I just think it's the _more_ confusing of the two directions we could go.

Thanks for being flexible 🙇🏼

bvaughn (approved):

markerikson:
Between the timing fixes here and Domi's x/y coord fixes, this looks great now!

image

PR #10588, opened 2024-06-27
Update the Cypress steps test to check for hover cursor position

markerikson:
This PR:

- Updates the `cypress/bankaccounts.spec.js` test recording to a fresh run (which I've set to "Never Delete" in admin)
- Updates `RecordedCursor.ts` to have data attributes for cursor details that we can check in the test
- Updates the `cypress-05` test to check the cursor location for a click and after-click step and assert its coords and click display are correct

### DIFF
--- a/src/ui/components/Video/RecordedCursor.tsx
+++ b/src/ui/components/Video/RecordedCursor.tsx
@@ -39,18 +39,20 @@ export function RecordedCursor() {
           element.style.top = `${mouseY}%`;
           element.style.transform = `scale(${cursorScale})`;
           element.style.setProperty("--click-display", shouldDrawClick ? "block" : "none");
-          // Testing support
 
-          element.dataset.cursordisplay = "true";
-          element.dataset.clickdisplay = shouldDrawClick ? "true" : "false";
-          element.dataset.clientx = `${mouseEvent.clientX}`;
-          element.dataset.clienty = `${mouseEvent.clientY}`;
+          // Set data attributes describing the cursor state for use in our E2E tests
+          element.dataset.cursorDisplay = "true";
+          element.dataset.clickDisplay = shouldDrawClick ? "true" : "false";
+          element.dataset.clientX = `${mouseEvent.clientX}`;
+          element.dataset.clientY = `${mouseEvent.clientY}`;
         } else {
           element.style.display = "none";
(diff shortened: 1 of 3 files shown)


src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -39,9 +39,18 @@ export function RecordedCursor() {
           element.style.top = `${mouseY}%`;
           element.style.transform = `scale(${cursorScale})`;
           element.style.setProperty("--click-display", shouldDrawClick ? "block" : "none");
+          // Testing support
+
```
bvaughn, line 43:
```suggestion

          // Support synthetic events from Cypress tests
```

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -58,6 +67,11 @@ export function RecordedCursor() {
         display: "none",
         position: "absolute",
       }}
+      data-test-name="recorded-cursor"
+      data-cursordisplay="false"
+      data-clickdisplay="false"
+      data-clientx="0"
+      data-clienty="0"
```
bvaughn, line 74: nit: Would be more consistent with how we name data attributes elsewhere in the codebase if you separated by "-", maybe

```suggestion
      data-client-x="0"
      data-client-y="0"
      data-show-click="false"
      data-show-cursor="false"
      data-test-name="recorded-cursor"
```

src/ui/components/Video/RecordedCursor.tsx
```diff
@@ -39,9 +39,18 @@ export function RecordedCursor() {
...
+          element.dataset.cursordisplay = "true";
+          element.dataset.clickdisplay = shouldDrawClick ? "true" : "false";
+          element.dataset.clientx = `${mouseEvent.clientX}`;
+          element.dataset.clienty = `${mouseEvent.clientY}`;
         } else {
           element.style.display = "none";
           element.style.setProperty("--click-display", "none");
+
+          element.dataset.cursordisplay = "false";
+          element.dataset.clickdisplay = "false";
```
bvaughn, line 53: Maybe it doesn't matter too much but should we reset the data client x and y values to 0 in this case?

bvaughn (approved):

PR #10593, opened 2024-07-02
Clone protocol messages before adding them to redux

hbenl:


Andarist (approved):

src/ui/reducers/protocolMessages.ts
```diff
@@ -83,7 +84,7 @@ const protocolMessagesSlice = createSlice({
           }
           case "error": {
             const { value: error } = message;
-            state.idToErrorMap[error.id] = error;
+            state.idToErrorMap[error.id] = cloneDeep(error);
```
bvaughn, line 87: This makes sense.

Is it possible this could cause problems anywhere that may rely on the data to stay in-sync after a mutation? (Do we read source stuff from Redux and use it for the related sources logic? Or does that always come from Suspense caches?

markerikson:
I'm actually not sure this is the right answer.

This is still going to result in `updateMappedLocation` trying to update fields inside of objects that Immer has frozen.

src/ui/reducers/protocolMessages.ts
hbenl: This redux slice is only used for the protocol viewer, I don't think it expects any updates.

> Do we read source stuff from Redux and use it for the related sources logic?

We're still using the sources from redux in some places but that is independent of this redux slice.

hbenl:
> This is still going to result in `updateMappedLocation` trying to update fields inside of objects that Immer has frozen.

`updateMappedLocation` is called for locations in messages taken from the suspense caches, not the clones that we put into this redux slice.

src/ui/reducers/protocolMessages.ts
bvaughn: Would probably be good for us to stop putting this data in Redux and just use our suspense caches directly (since it would avoid the potential for being out of sync here)

PR #10594, opened 2024-07-03
Use a proxy for mixpanel

hbenl:


src/ui/utils/mixpanel.ts
```diff
@@ -142,7 +142,9 @@ let mixpanelDisabled = true;
 const enableMixpanel = () => (mixpanelDisabled = false);
 
 export function initializeMixpanel() {
-  mixpanel.init("ffaeda9ef8fb976a520ca3a65bba5014");
+  mixpanel.init("ffaeda9ef8fb976a520ca3a65bba5014", {
+    api_host: "https://mixpanel-proxy.replay.io",
```
Andarist, line 146: If I were writing an adblocker I'd block all calls to any URL containing "mixpanel" in the URL 🤪 I recall having to change a script's name from `tracking.js` to something else cause it was being blocked by a blunt regex-based rule

PR #10595, opened 2024-07-03
Auto-scroll to the last failed test step

hbenl:
I also updated the logic for auto-selecting the failed step to use the *last* failed step if there is more than one.

### DIFF
--- a/packages/e2e-tests/tests/react_devtools-02-integrations.test.ts
+++ b/packages/e2e-tests/tests/react_devtools-02-integrations.test.ts
@@ -140,7 +140,7 @@ test("react_devtools-02: RDT integrations (Chromium)", async ({
   await searchComponents(page, "Anonymous"); // Search and select 1st result
   await verifySearchResults(page, {
     currentNumber: 1,
-    totalNumber: 16,
+    totalNumber: 17,
   });
 
   await componentSearchInput.focus();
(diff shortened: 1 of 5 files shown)


markerikson (approved):

hbenl:
That test failure is interesting: we are now getting 17 "Anonymous" React components instead of 16. This has nothing to do with this PR, it now also [happens on main](https://app.replay.io/team/dzowNDAyOGMwYS05ZjM1LTQ2ZjktYTkwYi1jNzJkMTIzNzUxOTI=/runs?testRunId=4e422e87-4847-446d-aab6-694b7c9ca632&testId=dHJ0OmE3MTQzZjkyY2EwODUwNmUzMzU4YTJkNTVlOWM5M2YxYWE3ZjEzYzU%3D). I guess this changed when the new RDT bundle was deployed to the backend (note that we have `backend_rerunRoutines` enabled for this test).
I'll update the test.

bvaughn:
> That test failure is interesting: we are now getting 17 "Anonymous" React components instead of 16. 

That's interesting. That didn't fail on Mark's commit to frontend (06b89f444f9630cea0afef4ff67624e3cf84f1aa) so...was it caused by the backend commit (06b89f444f9630cea0afef4ff67624e3cf84f1aa)?

hbenl:
> so...was it caused by the backend commit

I think so (that's why I mentioned `backend_rerunRoutines` - without it I think the backend commit should not have made a difference).

bvaughn:
That seems like a potentially significant weakness in our current sync process

markerikson:
Oh lovely :(

PR #10597, opened 2024-07-05
Don't show item count for cy.contains() events

hbenl:


src/ui/components/TestSuite/views/TestRecording/TestRecordingEvents/UserActionEventRow.tsx
```diff
@@ -118,7 +118,10 @@ export default memo(function UserActionEventRow({
     dispatch(jumpToKnownEventListenerHit(onSeek, jumpToCodeAnnotation));
   };
 
-  const showBadge = isSelected && resultPoint !== null;
+  const isCypressContainsEvent =
+    groupedTestCases.environment.testRunner.name === "cypress" &&
+    userActionEvent.data.command.name === "contains";
```
bvaughn, line 123: I dislike that we need this level of test-runner-specific special casing in Devtools.

PR #10600, opened 2024-07-05
HACK DO NOT SUBMIT.  Integration example for replay videos.

klochek:
A (possible/plausible?) integration of the new getVideos functionality.  Missing is needing to account for multiple videos (due to resolution changes), and probably some new logic to fall back to the old .pngs if no videos are present.


markerikson:
Does this need a protocol package update to make `Video` available?

PR #10601, opened 2024-07-05
Removed left/right breakpoint arrows from Pause Info panel

bvaughn:
This is my reasoning:
- These button confuse all of us. (Domi, Josh, Mark, and I just chatted about this in Discord.)
- Mixpanel data shows that [almost no one uses the buttons](https://mixpanel.com/project/2349904/view/2894968/app/insights#2NaEPayGGLnB).
- The rewind/fast-forward buttons (available via hover click and context menu) are easier to use. (After all, Replay is fundamentally different from a traditional browser/debugger. Let's lean into that and simplify our UI!)

Screenshot 2024-07-05 at 6 00 02 PM


Andarist (approved):

markerikson:
Note that we _do_ use a `breakpoints-01` recording as the "Golden" example for some of our _other_ E2E tests.  If we remove it, we're stuck with the current recording ID in `examples.json`, forever.

bvaughn:
> Note that we _do_ use a `breakpoints-01` recording as the "Golden" example for some of our _other_ E2E tests. If we remove it, we're stuck with the current recording ID in `examples.json`, forever.

We can use a different test.

markerikson:
@bvaughn except this PR also deleted `breakpoints-02` and _all_ of the `breakpoints-*` tests, so none of those will be useable in the future as E2E examples.

bvaughn:
Oof. That was silly. I mixed up breakpoints-02 and logpoints-02. I'll post a follow up.

PR #10606, opened 2024-07-09
Auto-expand framework frames when one of them is selected

hbenl:


src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.tsx
```diff
@@ -56,6 +55,11 @@ export function Group({
     copyStackTrace,
   });
 
+  const containsSelectedFrame = group.some(
+    frame =>
+      selectedFrameId?.pauseId === frame.pauseId && selectedFrameId?.frameId === frame.protocolId
+  );
+  const [expanded, setExpanded] = useState(containsSelectedFrame);
```
Andarist, line 62: we could compute this lazily to denote more prominently that it's only used as the initial value here
```suggestion
  const [expanded, setExpanded] = useState(
    frame =>
      selectedFrameId?.pauseId === frame.pauseId && selectedFrameId?.frameId === frame.protocolId  
  );
```

Andarist (approved):

src/devtools/client/debugger/src/components/SecondaryPanes/Frames/Group.tsx
hbenl: IMO the code is slightly more readable the way it is (in particular I'd like to keep the `containsSelectedFrame` variable because it also serves as documentation).

PR #10608, opened 2024-07-10
Stop gathering element's CSS rules at the nearest HTML element

Andarist:
Traversing elements from outer frames could lead to wildly confusing results for elements within iframes

**Before**
image

**After**
image


src/devtools/client/inspector/rules/models/element-style.ts
```diff
@@ -153,12 +153,16 @@ export default class ElementStyle {
...
-            this._maybeAddRule(rule, elemNodeWithId);
+            this._maybeAddRule(rule, parentNodeWithId);
           }
         }
       }
 
-      parentNodeId = elem.parentNode;
+      if (parentObject.preview?.node?.nodeName === "HTML") {
+        break;
+      }
```
Andarist, line 163: This is the fix, the rest is just renaming things. 

At first, I started playing with comparing this against data received from:
```ts
const ownerDocumentRequest = this.replayClient.getDocument(this.pauseId).catch(() => null);
```
but then I realized I could access this more directly.

bvaughn (approved):
Nice change!

PR #10609, opened 2024-07-10
Display matched selector priority

Andarist:
This is a follow-up to the investigation that led to https://github.com/replayio/devtools/pull/10608 . We can see there that the non-top rule was originally left as the "matching one". 

While it shouldn't even be displayed there in the first place (which got fixed by https://github.com/replayio/devtools/pull/10608 ), it wasn't apparent *why* it got matched. The specificity wasn't higher. It turns out that it was an `!important` rule but we never surfaced that in the UI.

### DIFF
--- a/src/devtools/client/inspector/rules/components/DeclarationValue.tsx
+++ b/src/devtools/client/inspector/rules/components/DeclarationValue.tsx
@@ -2,6 +2,7 @@ import React from "react";
 
 import { COLOR, FONT_FAMILY, URI } from "third-party/css/output-parser";
 
+import { Priority } from "../models/text-property";
 import Color from "./value/Color";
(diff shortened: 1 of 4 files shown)


src/devtools/client/inspector/rules/components/DeclarationValue.tsx
```diff
@@ -10,12 +10,13 @@ interface DeclarationValueProps {
   colorSpanClassName: string;
   colorSwatchClassName: string;
   fontFamilySpanClassName: string;
+  priority?: string;
```
Andarist, line 13: Are there places where we'd like to render this without the priority? Chrome doesn't render it in its Computed tab

hbenl (approved):

PR #10612, opened 2024-07-16
Dependency Graph UI/UX changes

bvaughn:
https://www.loom.com/share/ce2d6b56d4914ee2afd81fa8a1bf2a9f

- [x] Split component stack and dependency graph into two accordion tabs (labeled "experimental")
- [x] Make tabs look and feel more like our Call Stack info (no "jump" button, just click the row to jump to code)
- [x] Auto-load dependency graph paths when accordion is expanded
- [x] Hide JSON viewer (this info is available in the Protocol panel)
- [x] Hide React-only dependency graph (the info was redundant)

| Before | After |
| :--- | :--- |
| Screenshot 2024-07-16 at 1 57 57 PM | Screenshot 2024-07-16 at 1 58 16 PM |


markerikson:
I like the changes in general, but have some thoughts / concerns:

- Like with call stacks, I think there needs to be a separation between "the point we loaded the data for" and "where we're currently paused". What's trickier in this case is that with call stacks, we jump you to the other frame's line of code, but don't change the pause point.  With the "jump" buttons I've got here, that _does_ change the pause point, but in my current mental / behavior model I want to retain the dep graph data for the original point.  (Not sure exactly how you changed it yet, just riffing off the description.)  I could see a potential argument for _not_ directly exposing the "jump to that earlier point" button, similar to how we do with call stacks, and just selecting the line of code.  I guess I'm trying to figure out how you actually _would_ jump to the earlier point if all we do is show you the line, and with call stacks the answers are "add a print statement on the earlier line", or "right-click the frame and 'Jump to Frame' ".
- There _is_ different data that comes back for the "normal" Dep Graph vs the "React" dep graph, which was part of why I had both of those being displayed - to visualize the different data for the same point, and get a sense of what data _is_ coming back.

bvaughn:
> * Like with call stacks, I think there needs to be a separation between "the point we loaded the data for" and "where we're currently paused".

With this feature, we're always loading data from the "top frame" (the current point) which is why I have it highlighted as selected (matching our Call Stack UI)

markerikson:
No, there were definitely differences in the results.

In particular, try going to a component that rendered multiple times over the course of the app.  Pick a hit point that was triggered by something like a local `setState`, then load the dep graph.

The normal dep graph focuses on this _one_ current event and how we got here. What you should see is that the "normal" dep graph shows, say, one entry that's `Discrete Input click` or something like that, whereas the "React" dep graph should show all of the ancestor component `ReactElementCreated` and `ReactRendered` entries, at the most recent times that those ancestor renders lead to _this_ component - in other words, the "component stack" data, even though the ancestor components did not render as part of _this_ render pass.

markerikson:
and yeah, with the "jump" buttons, I was treating it slightly differently than call stacks - ie, going to both the earlier line _and_ the point we hit that line.  Definitely reasonable for either approach. I'm fine with the change here to match the similarity with call stack behavior.

bvaughn:
> No, there were definitely differences in the results.

Got it. I'll restore those separate stacks then.

bvaughn:
Interestingly sometimes the differences in these stacks look _weird_. For example, [here](https://experimental.replay.io/recording/async-stack-should-detect-async-stacks-outside-the-focus-window--4a4214e2-f367-400a-abf5-cd3efbfe62e3?commentId=&focusWindow=eyJiZWdpbiI6eyJwb2ludCI6IjAiLCJ0aW1lIjowfSwiZW5kIjp7InBvaW50IjoiNDU3NTcxMTYwNjk1Mzk1Njk5MzI1NzYwNjkyMDI3MzkyMDAiLCJ0aW1lIjoyMjU4OX19&point=6165852519807576545047785857089538&primaryPanel=debugger&secondaryPanel=react-components&time=2831.5365425158116&viewMode=dev)

I'm paused inside of a React component that's rendering:
Screenshot 2024-07-16 at 3 42 56 PM

But the React dep graph looks less helpful than the normal one:
Screenshot 2024-07-16 at 3 42 50 PM

markerikson:
Yep, agreed, especially in that case. I've been showing bhackett cases like this and asking him to investigate.  (Who knows how much longer that will be feasible.)

bvaughn:
I guess the thing to do for now will be...I don't even know what. File Linear bugs with Replays attached? Will those even be looked at?

bhackett1024:
> I guess the thing to do for now will be...I don't even know what. File Linear bugs with Replays attached? Will those even be looked at?

We need to discuss this at the company / leadership level.  I'm happy to continue working on this but only if we can decide that it's important enough to pull attention away from the other work going on.

bvaughn:
> I'm happy to continue working on this but only if we can decide that it's important enough to pull attention away from the other work going on.

I was under the impression that the dependency graph feature was pretty central to possible Replay's AI work.

cc @jasonLaster

bhackett1024:
> I was under the impression that the dependency graph feature was pretty central to possible Replay's AI work. Maybe the expectation is that this kind of work shouldn’t require any help from the TT team?

OK, I synced with @jasonLaster to get caught up.  We do want to continue this work, the plan we talked about is that @bvaughn go ahead and file linear issues around problems with the dependency graph and its react data.  @markerikson can focus on fixes and I'll be available to support and help however I'm needed.

PR #10619, opened 2024-10-22
Fixed an issue with `SupportForm` not being submittable

Andarist:


packages/replay-next/components/errors/UnexpectedErrorForm.tsx
```diff
@@ -61,7 +61,6 @@ function UnexpectedErrorFormSuspends({ details, replayClient, title, unexpectedE
   return (
     <ModalFrame
       dataTestId="UnexpectedErrorDetails"
-      onDismiss={noop}
```
Andarist, line 64: this was acting as the "outer one" as per the comment I added:
```ts
// TODO: this doesn't work correctly when multiple stacked modals are open
// they are unaware of each other and the global listeners added by them compete between each other
// in a way that the "outer" can perform its action and call `.preventDefault` - preventing the "inner" one from closing
```

markerikson (approved):

PR #10652, opened 2026-04-17
Update jest config and add more unit test coverage

striderwilsonreplay:


markerikson (approved):

packages/protocol/execution-point-utils.ts
```diff
@@ -3,7 +3,7 @@ import { ExecutionPoint } from "@replayio/protocol";
 import { compareNumericStrings } from "protocol/utils";
 
 export function pointEquals(p1: ExecutionPoint, p2: ExecutionPoint) {
-  p1 == p2;
+  return p1 === p2;
```
striderwilsonreplay, line 6: Always returned undefined. Fixed to return boolean

PR #10653, opened 2026-04-23
PRO-2372: capture ad attribution and forward on login

michaelfward:
captures ad click IDs (`li_fat_id`, `twclid`, `rdt_cid`) and UTM params from the URL on every app.replay.io page load, then forwards them as query params on the `/login?...` redirect so the dashboard can thread them into auth0's authorizationParams. ticket: PRO-2372.

## how it works

1. new `src/ui/utils/adAttribution.ts` exports `captureAdAttribution` / `readAdAttribution` / `clearAdAttribution`. first-touch semantics: once set in localStorage, we don't overwrite.
2. `pages/_app.tsx` calls `captureAdAttribution()` in a `useEffect` on mount, so it runs regardless of auth state on every page load. SSR-safe via `typeof window` guard.
3. `src/ui/utils/auth.ts` `login()` reads cached attribution and appends it to the `/login?origin=...&returnTo=...` redirect URL.
4. `logout()` clears attribution so a second user on the same browser doesn't inherit the previous signup's click ID.

## rollout order

1. depends on the dashboard PR to pick up the params and forward to auth0, and the backend PR to persist them on first auth. shipping devtools standalone is harmless - attribution just ends up in a URL param the dashboard ignores until its own PR ships.

## test plan

- [x] `npx tsc --noEmit` - clean
- [ ] hit `https://app.replay.io/?li_fat_id=smoke&utm_source=linkedin`
- [ ] devtools console: `localStorage.getItem('replay_ad_attribution')` shows the captured JSON
- [ ] click Sign Up: outbound `/login?...` URL has `li_fat_id=smoke&utm_source=linkedin` appended
- [ ] log out: `localStorage.replay_ad_attribution` is cleared

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -19,6 +19,7 @@ import { ConfirmProvider } from "ui/components/shared/Confirm";
 import LoadingScreen from "ui/components/shared/LoadingScreen";
 import useAuthTelemetry from "ui/hooks/useAuthTelemetry";
 import { bootstrapApp } from "ui/setup";
+import { captureAdAttribution } from "ui/utils/adAttribution";
 import { listenForAccessToken } from "ui/utils/browser";
 import { useLaunchDarkly } from "ui/utils/launchdarkly";
 import { InstallRouteListener } from "ui/utils/routeListener";
@@ -81,6 +82,10 @@ const App = ({ apiKey, ...props }: AppProps & AuthProps) => {
   useAuthTelemetry();
   const router = useRouter();
   let head: React.ReactNode;
+
+  useEffect(() => {
+    captureAdAttribution();
+  }, []);
 
   // HACK: Coordinates with the recording page to render its <head> contents for
   // social meta tags. This can be removed once we are able to handle SSP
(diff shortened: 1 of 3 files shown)


michaelfward:
closing. re-analyzed today: devtools only serves \`/recording/*\` (next.js basePath). ads never deep-link to a specific shared recording, so the capture logic in \`_app.tsx\` almost never fires from a real ad flow. app.replay.io/ (dashboard) is where users actually land from app-level signup CTAs, and replay.io/ (landing-page) is where they land from ads per tom.

moving the capture to where it actually happens:
- new landing-page PR: capture on first touch, write \`.replay.io\` cookie with 90d max-age (matching ad CVR window)
- reworked dashboard #195: read the cookie + belt-and-suspenders capture for users landing directly on app.replay.io
- backend #12932 stays as-is (with segment event name fixed)
- extension and cli PRs closed, same reason

if shared-recording deep-link campaigns ever become a real channel, we'd re-open a devtools PR that reads the shared \`.replay.io\` cookie rather than maintain duplicate localStorage capture.

PR #10656, opened 2026-08-18
Allow jump-to-code-01 nav seek to round to 1%

Domiii:
* `jump-to-code-01` treats a navigation-event seek as start-of-recording.
* `Math.round` of timeline percent expected `0`.

## Timeline percent rounding

* CI re-records the example each run.
* Observed seek: `13 / 2496` ms → `0.52%` → `Math.round` → `1`.
* Same test already treats later J2C percents as timing-dependent.
* Assert now `toBeLessThanOrEqual(1)`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/e2e-tests/tests/jump-to-code-01_basic.test.ts
+++ b/packages/e2e-tests/tests/jump-to-code-01_basic.test.ts
@@ -136,7 +136,7 @@ test(`jump-to-code-01: Test basic jumping functionality`, async ({
   await navigationEvent.click();
   await waitFor(async () => {
     const timelinePercent = await getTimelineCurrentPercent(page);
-    expect(Math.round(timelinePercent)).toBe(0);
+    expect(Math.round(timelinePercent)).toBeLessThanOrEqual(1);
   });
 
   debugPrint(page, "Checking that the first keypress J2C jumps to the correct line");


packages/e2e-tests/tests/jump-to-code-01_basic.test.ts
```diff
@@ -136,7 +136,7 @@ test(`jump-to-code-01: Test basic jumping functionality`, async ({
   await navigationEvent.click();
   await waitFor(async () => {
     const timelinePercent = await getTimelineCurrentPercent(page);
-    expect(Math.round(timelinePercent)).toBe(0);
+    expect(Math.round(timelinePercent)).toBeLessThanOrEqual(1);
```
Andarist, line 139:
```suggestion
    expect(Math.floor(timelinePercent)).toBe(0);
```

Andarist (approved):

packages/e2e-tests/tests/jump-to-code-01_basic.test.ts
Domiii: Mine has more wiggle room (allows going up to 1.49999) and I don't mind that one bit :smiling_face_with_tear:

